PR #293, opened 2018-04-09
rename HTTPHeaders.getCanonicalForm to subject and improve efficiency

weissi:
Motivation:

HTTPHeaders had an unusual API for Swift: `getCanonicalForm(String)` which is
better suited as a subscript `[canonicalForm: String]`.
Also the implementation was needlessly inefficient.

Modifications:

- renamed HTTPHeaders.getCanonicalForm to HTTPHeaders[canonicalForm]
- improved efficiency by replacing anti-pattern
  Array.map { ...  }.reduce(+, []) by flatMap

Result:

more Swift in both meanings

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -286,6 +286,11 @@ public struct HTTPHeaders: CustomStringConvertible {
             buffer.write(staticString: crlf)
         }
     }
+    
+    @available(*, deprecated, message: "getCanonicalForm has been changed to a subscript: headers[canonicalForm: name]")
+    public func getCanonicalForm(_ name: String) -> [String] {
+        return self[canonicalForm: name]
+    }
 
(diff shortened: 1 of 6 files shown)


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -294,15 +299,15 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    public subscript(canonicalForm name: String) -> [String] {
         // It's not safe to split Set-Cookie on comma.
         let queryName = name.lowercased()
         if queryName == "set-cookie" {
             return self[name]
         }
 
         if let result = storage[queryName] {
-            return result.map { tuple in tuple.1.split(separator: ",").map { String($0.trimWhitespace()) } }.reduce([], +)
+            return result.flatMap { tuple in tuple.1.split(separator: ",").map { String($0.trimWhitespace()) } }
```
weissi: this is the only non-renaming change

PR #294, opened 2018-04-09
fix issues found by TSan

### LINKED ISSUE #292: investigate master TSan issues
see title

weissi:
- thanks @vlm for pointing this out (fixes #292)
- because TSan doesn't run cleanly with Dispatch on Linux we can't put this in CI at the moment :(
- for some reason I was thinking that `lifecycleManager` is immutable which is obviously totally bonkers as its sole purpose is to track the lifecycle _changes_. Silly me 🙈

Motivation:

NIO 1.3.1 was TSan clean but `master` regressed. In total I could see
two breakages both proudly presented by `SocketChannelLifecycleManager`.
Sorry :| :

1. accessing the `isActiveAtomic` from arbitrary threads:
   this was clearly a bug because `SocketChannelLifecycleManager` is
   usually held mutabe. So `isActiveAtomic` needs to be injected into
   `SocketChannelLifecycleManager` rather than contstructed there
2. accessing the `ChannelPipeline` through `SocketChannelLifecycleManager`:
   same problem as above really.

Modifications:

inject both the ChannelPipeline and the isActiveAtomic in the
appropriate places.

Result:

NIO much more correct, TSan happy

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -30,10 +30,10 @@ private struct SocketChannelLifecycleManager {
     }
 
     // MARK: properties
+    private let eventLoop: EventLoop
     // this is queried from the Channel, ie. must be thread-safe
-    internal let isActiveAtomic = Atomic(value: false)
+    internal let isActiveAtomic: Atomic<Bool>
     // these are only to be accessed on the EventLoop
-    internal let channelPipeline: ChannelPipeline
     private var currentState: State = .fresh {
         didSet {
             assert(self.eventLoop.inEventLoop)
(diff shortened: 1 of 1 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -276,9 +279,9 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
 
     // This is `Channel` API so must be thread-safe.
     public final var pipeline: ChannelPipeline {
-        return self.lifecycleManager.channelPipeline
+        return self._pipeline
     }
-
+
```
Lukasa: Sneaky sneaky whitespace.

weissi: whaat, I thought I have git configured to auto-remove them

weissi: ha, that was on my other laptop. /me should sync dotfiles :)

weissi:
@Lukasa please re-review, fixed now

PR #295, opened 2018-04-09
Enable our one disabled integration test.

### LINKED ISSUE #132: end2end tests 12 & 14 are disabled at the moment, fix them
see title, for @weissi to do as he promised that a while ago and then forgot :|

Lukasa:
Motivation:

We have one integration test left disabled, but actually it basically
already passes: it just expects the wrong format of error. We should
have all our integration tests enabled.

Modifications:

Enabled the test.
Changed the wording of the error it expected.

Result:

No disabled integration tests.

Resolves #132.


Lukasa:
Argh dictionaries are the worst, one sec.

IntegrationTests/tests_01_http/test_14_strict_mode_assertion.sh
```diff
@@ -20,13 +20,19 @@ start_server "$token"
...
+fi
+if ! grep -q 'content-length: 0' "$tmp/actual"; then
+    fail "couldn't find content-length in response"
+fi
+if ! grep -q 'connection: close' "$tmp/actual"; then
+    fail "couldn't find connection: close in response"
+fi
+
+linecount=$(wc "$tmp/actual")
+if [ $linecount -ne 4 ]; then
```
weissi: just for consistency, would you mind using `[[ ... ]]` ?

Lukasa:
ACK, done.

PR #296, opened 2018-04-10
Stop calling read(...) if we were not able to fill the whole buffer.

normanmaurer:
Motivation:

If we not filled the whole buffer with read(...) we should stop reading and wait until we get notified again. Otherwise chances are good that the next read(...) call will either read nothing or only a very small amount of data.
Also this will allow us to call fireChannelReadComplete() which may give the user the chance to flush out all pending writes.

Modifications:

Stop reading and wait for next wakup when we not fill the whole buffer.

Result:

Less wasted read calls and early chance of flushing out pending data.

### DIFF
--- a/Sources/NIO/SocketChannel.swift
+++ b/Sources/NIO/SocketChannel.swift
@@ -136,12 +136,19 @@ final class SocketChannel: BaseSocketChannel<Socket> {
 
                     assert(self.isActive)
                     pipeline.fireChannelRead0(NIOAny(buffer))
-                    if mayGrow && i < maxMessagesPerRead {
+                    result = .some
+
+                    if buffer.writableBytes > 0 {
+                        // If we did not fill the whole buffer with read(...) we should stop reading and wait until we get notified again.
+                        // Otherwise chances are good that the next read(...) call will either read nothing or only a very small amount of data.
+                        // Also this will allow us to call fireChannelReadComplete() which may give the user the chance to flush out all pending
+                        // writes.
+                        return result
+                    } else if mayGrow && i < maxMessagesPerRead {
                         // if the ByteBuffer may grow on the next allocation due we used all the writable bytes we should allocate a new `ByteBuffer` to allow ramping up how much data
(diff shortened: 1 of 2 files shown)


normanmaurer:
This also gives a small perf improvement for HTTP benchmarks.

Sources/NIO/SocketChannel.swift
```diff
@@ -136,12 +136,19 @@ final class SocketChannel: BaseSocketChannel<Socket> {
 
                     assert(self.isActive)
                     pipeline.fireChannelRead0(NIOAny(buffer))
-                    if mayGrow && i < maxMessagesPerRead {
+                    result = .some
+
+                    if buffer.writableBytes > 0 {
+                        // If we not filled the whole buffer with read(...) we should stop reading and wait until we get notified again.
```
weissi: grammar nit: 'If we did not fill the whole buffer' ...

normanmaurer: fixed

normanmaurer:
@weissi never! Fixed it 👍

Tests/NIOTests/ChannelTests.swift
```diff
@@ -1963,12 +1972,14 @@ public class ChannelTests: XCTestCase {
...
-                XCTFail("channelReadComplete unexpected")
+                if self.state != .read {
+                    XCTFail("channelReadComplete unexpected")
+                }
                 self.state = .readComplete
             }
 
             func errorCaught(ctx: ChannelHandlerContext, error: Error) {
-                XCTAssertEqual(.read, self.state)
+                XCTAssertTrue(.read == self.state || .readComplete == self.state)
```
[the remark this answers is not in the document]
normanmaurer: actually I think  `.readComplete ==` is the only one we need to test... let me fix

normanmaurer:
@Lukasa addressed

Lukasa:
Rebase and merge when ready.

PR #297, opened 2018-04-10
TailChannelHandler does not need to implement _ChannelOutboundHandler

normanmaurer:
Motivation:

TailChannelHandler does not implement any _ChannelOutboundHandler methods and so has no need to conform to it.

Modifications:

Not implement _ChannelOutboundHandler for TailChannelHandler

Result:

More correct code


normanmaurer:
@Lukasa @weissi noticed this while looking through our pipeline impl.

PR #299, opened 2018-04-10
Reuse detected keepalive of http_parser as long as possible

normanmaurer:
Motivation:

To detect if keepalive is used we need to search the headers (if HTTP/1.1 is used). This may be expensive depending how many headers are present. http_parser itself detects already if keep alive should be used and so we can re-use this as long as the user did not modify the headers.

Modifications:

Reuse keepalive parsed by http_parser as long as the headers were not modified.

Result:

Less overhead as long as the headers are not modified.

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -15,6 +15,13 @@
 import NIO
 import CNIOHTTPParser
 
+private extension UnsafeMutablePointer where Pointee == http_parser {
+    /// Returns the `KeepAliveState` for the current message that is parsed.
+    var keepAliveState: KeepAliveState {
+        return c_nio_http_should_keep_alive(self) == 0 ? .close : .keepAlive
+    }
+}
+
 private struct HTTPParserState {
     var dataAwaitingState: DataAwaitingState = .messageBegin
     var currentNameIndex: HTTPHeaderIndex?
@@ -214,7 +221,7 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
     private func newRequestHead(_ parser: UnsafeMutablePointer<http_parser>!) -> HTTPRequestHead {
         let method = HTTPMethod.from(httpParserMethod: http_method(rawValue: parser.pointee.method))
         let version = HTTPVersion(major: parser.pointee.http_major, minor: parser.pointee.http_minor)
-        let request = HTTPRequestHead(version: version, method: method, rawURI: state.currentURI!, headers: HTTPHeaders(buffer: cumulationBuffer!, headers: state.currentHeaders))
+        let request = HTTPRequestHead(version: version, method: method, rawURI: state.currentURI!, headers: HTTPHeaders(buffer: cumulationBuffer!, headers: state.currentHeaders, keepAliveState: parser.keepAliveState))
         self.state.currentHeaders.removeAll(keepingCapacity: true)
         return request
     }
@@ -222,7 +229,7 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
     private func newResponseHead(_ parser: UnsafeMutablePointer<http_parser>!) -> HTTPResponseHead {
         let status = HTTPResponseStatus(statusCode: Int(parser.pointee.status_code), reasonPhrase: state.currentStatus!)
         let version = HTTPVersion(major: parser.pointee.http_major, minor: parser.pointee.http_minor)
-        let response = HTTPResponseHead(version: version, status: status, headers: HTTPHeaders(buffer: cumulationBuffer!, headers: state.currentHeaders))
+        let response = HTTPResponseHead(version: version, status: status, headers: HTTPHeaders(buffer: cumulationBuffer!, headers: state.currentHeaders, keepAliveState: parser.keepAliveState))
         self.state.currentHeaders.removeAll(keepingCapacity: true)
         return response
     }
@@ -383,7 +390,8 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
             handler.state.complete(state: handler.state.dataAwaitingState)
             handler.state.dataAwaitingState = .messageBegin
 
-            let trailers = handler.state.currentHeaders.isEmpty ? nil : HTTPHeaders(buffer: handler.state.cumulationBuffer!, headers: handler.state.currentHeaders)
+            // Just use unknown for trailers as there is no point for anything else.
+            let trailers = handler.state.currentHeaders.isEmpty ? nil : HTTPHeaders(buffer: handler.state.cumulationBuffer!, headers: handler.state.currentHeaders, keepAliveState: .unknown)
             handler.state.currentHeaders.removeAll(keepingCapacity: true)
             switch handler {
             case let handler as HTTPRequestDecoder:
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -17,6 +17,18 @@ import NIO
 let crlf: StaticString = "\r\n"
 let headerSeparator: StaticString = ": "
 
+private let connectionUtf8 = "connection".utf8
+
+// Keep track of keep alive state.
+internal enum KeepAliveState {
+    // We know keep alive should be used.
+    case keepAlive
+    // We know we should close the connection.
+    case close
+    // We need to scan the headers to find out if keep alive is used or not
+    case unknown
+}
+
 /// A representation of the request line and header fields of a HTTP request.
 public struct HTTPRequestHead: Equatable {
     private final class _Storage {
@@ -175,15 +187,7 @@ extension HTTPRequestHead {
     /// Whether this HTTP request is a keep-alive request: that is, whether the
     /// connection should remain open after the request is complete.
     public var isKeepAlive: Bool {
-        guard let connection = headers["connection"].first?.lowercased() else {
-            // HTTP 1.1 use keep-alive by default if not otherwise told.
-            return version.major == 1 && version.minor == 1
-        }
-
-        if connection == "close" {
-            return false
-        }
-        return connection == "keep-alive"
+        return headers.isKeepAlive(version: version)
     }
 }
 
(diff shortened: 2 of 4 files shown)


normanmaurer:
Also @tanner0101 

This gives some perf win in my benchmarks. Let me know what you think about this idea @tanner0101 @weissi @Lukasa

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -220,7 +234,22 @@ public struct HTTPHeaders: CustomStringConvertible {
...
-    private var continuous: Bool = true
+
+    fileprivate var state: HTTPHeadersModificationState = .notmodified
+
+    /// Update the current state of the headers if needed.
+    private mutating func updateStateIfNeeded(newState: HTTPHeadersModificationState) {
+        switch newState {
+        case .modified where self.state == .notmodified:
+            self.state = newState
+        case .modifiedNonContinuous:
```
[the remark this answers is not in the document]
normanmaurer: good idea :)

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -220,7 +234,20 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    /// Update the current state of the headers if needed.
+    private mutating func updateStateIfNeeded(newState: HTTPHeadersModificationState) {
+        switch newState {
+        case .modified where self.state == .notModified, .modifiedNonContinuous:
+            self.state = newState
+        case .notModified:
+            fatalError("Should never be called with \(newState)")
+        default:
+            break
+        }
```
Lukasa: I tend to think this kind of method is nicer on the enum. Also no need to say "ifNeeded" in the name.

Lukasa: In fact, it may be nicer just to have functions that explain what they are, e.g. `state.appendedHeader()` and `state.removedHeader()`. Then you can have some computed vars that use the right names, e.g `state.continuous` and `state.wasModified`.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -202,6 +211,34 @@ private extension UInt8 {
...
 }
 
+fileprivate enum HTTPHeadersModificationState {
+    case modifiedNonContinuous
+    case modified
+    case notModified
+
+    mutating func addedHeader() {
+        switch self {
+        case .notModified:
```
BasThomas: `if case . notModified = self {`?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -38,13 +38,17 @@ public struct HTTPRequestHead: Equatable {
     /// The header fields for this HTTP request.
     public var headers: HTTPHeaders
 
+    /// The keepAlive that may be parsed by http_parser before. This is only safe to be used as long as
+    /// the user did not modify the headers yet.
```
Lukasa: Let's update this comment slightly:

```
/// The keepAlive value that was detected by http_parser. This can only be used safely
/// as long as the user has not made modifications to this header block.
```

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -202,6 +211,31 @@ private extension UInt8 {
     }
 }
 
+fileprivate enum HTTPHeadersModificationState {
```
Lukasa: Mind adding some quick doc comments here?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -202,6 +211,31 @@ private extension UInt8 {
...
+    case notModified
+
+    mutating func addedHeader() {
+        if case .notModified = self {
+            self = .modified
+        }
+    }
+
+    mutating func removedHeader() {
+        if self != .modifiedNonContinuous {
```
Lukasa: Why doesn't this use `if case`?

normanmaurer:
@Lukasa addressed

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -202,6 +211,41 @@ private extension UInt8 {
     }
 }
 
+/// Represent the current modification state of the HTTPHeaders.
+fileprivate enum HTTPHeadersModificationState {
+    /// Was modified and the underlying storage is not continous anymore.
+    case modifiedNonContinuous
+    /// Was modified and the underlying storage is continous.
+    case modified
+
```
weissi: nit: why a space here?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -38,13 +38,17 @@ public struct HTTPRequestHead: Equatable {
     /// The header fields for this HTTP request.
     public var headers: HTTPHeaders
 
+    /// The keepAlive value that was detected by http_parser. This can only be used safely
+    /// as long as the user has not made modifications to this header block.
+    private let keepAlive: Bool?
```
weissi: I don't really like having a state variable here that is only valid when also consulting another variable. I'd say the whole state should live in the headers. Couldn't we architect it this way: We allow (as an `internal` constructor) to pre-seed the headers with a keep alive state. Whenever the headers change, we invalidate that.

So the keep-alive state in the headers could be one of the following:

- `keepAlive`: no header scan needed, I know I have a `Connection: keep-alive` header
- `close`: no header scan needed, I know I have a `Connection: close` header
- `default`: no header scan needed, I know I had neither `Connection: keep-alive` nor `Connection: close`
- `needsScan`: header scan needed

so internally we could pre-seed it with either `keepAlive`, `close`, or `default`, depending on what http_parser tells us. Then on any modification we change it to `needsScan`.

normanmaurer:
@Lukasa @weissi PTAL again

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -202,6 +197,69 @@ private extension UInt8 {
...
+}
+
+// Keep state of keep alive state.
+internal enum KeepAliveState {
+    // We know keep alive should be used.
+    case keepAlive
+    // We know we should close the connection.
+    case close
+    // We need to scan the headers to find out if keep alive is used or not
+    case needScan(continuous: Bool)
```
Lukasa: Why is there a continuous flag here? You have a `HTTPHeadersModificationState` that tracks this.

normanmaurer: Because I forgot to remove the modification state... it’s all tracked via this one now

normanmaurer: ok removed

Lukasa:

Sources/NIOHTTP1/HTTPTypes.swift
Lukasa: I don't like this much. I'd rather this bool was moved out of this enum: it just has nothing to do with what this enum is tracking.

If we do that, we can then name the third case of this enum `unknown`, and then the implementation of `HTTPHeaders.isKeepAlive` gets a lot nicer because it can just return `.unknown` when it doesn't know the answer instead of being passed the HTTP version.

normanmaurer:
@Lukasa @weissi need to investigate as it actually makes things slower now ?!?!?

weissi:
@normanmaurer ping, what’s up with this friend?

normanmaurer:
@weissi I need to investigate as it actually resulted in a slowdown after rebasing

weissi:
@normanmaurer you probably want to use the isolated (no network) perf test to verify that. Otherwise there’ll be too much noise I suspect

tomerd:

tomerd:

tomerd:

normanmaurer:
Alright updated please take a look again @weissi @Lukasa 

Before:

```
 wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://127.0.0.1:8888
Running 10s test @ http://127.0.0.1:8888
  4 threads and 256 connections
    Latency   104.83ms  209.95ms   2.00s    90.27%
    Req/Sec   132.15k    31.57k  346.95k    88.41%
  5234833 requests in 10.04s, 249.62MB read
  Socket errors: connect 0, read 162, write 0, timeout 17
Requests/sec: 521428.30
Transfer/sec:     24.86MB
```

with this pr:

```
wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://127.0.0.1:8888
Running 10s test @ http://127.0.0.1:8888
  4 threads and 256 connections
    Latency   105.65ms  215.06ms   1.99s    91.27%
    Req/Sec   137.61k    32.64k  329.69k    89.59%
  5406252 requests in 10.03s, 257.79MB read
  Socket errors: connect 0, read 137, write 4, timeout 24
Requests/sec: 538977.98
Transfer/sec:     25.70MB
```

normanmaurer:
And @tanner0101

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -15,6 +15,14 @@
 import NIO
 import CNIOHTTPParser
 
+private extension UnsafeMutablePointer where Pointee == http_parser {
+
```
Lukasa: No need for this whitespace.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -17,6 +17,20 @@ import NIO
 let crlf: StaticString = "\r\n"
 let headerSeparator: StaticString = ": "
 
+private let connectionUtf8 = "connection".utf8
+
+
+// Keep state of keep alive state.
```
Lukasa: Nit: keep track, not keep state.

normanmaurer:
@Lukasa yep will add some tests just to ensure we correctly set it to "unknown" when we add / remove connection header.

normanmaurer:
@Lukasa added some tests.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -286,6 +290,26 @@ private extension UInt8 {
...
 
+/* private but tests */ internal extension HTTPHeaders {
+    func isKeepAlive(version: HTTPVersion) -> Bool {
+        switch self._storage.keepAliveState {
+        case .close:
+            return false
+        case .keepAlive:
+            return true
+        case .unknown:
+            guard let connection = self["connection"].first?.lowercased() else {
```
weissi: whilst you're working on this: worth fixing this implementation of follow-up?

normanmaurer: Let me do a follow up

PR #304, opened 2018-04-12
Introduce HTTPHeaderName for faster header operations

normanmaurer:
Motivation:

Most HTTP header names are well known and so we can optimize for them.

Modification:

- Introduce HTTPHeaderName and also expose common used names via static lets.
- Add unit test

Result:

Less overhead when acting on headers

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -128,7 +128,7 @@ extension HTTPRequestHead {
     /// Whether this HTTP request is a keep-alive request: that is, whether the
     /// connection should remain open after the request is complete.
     public var isKeepAlive: Bool {
-        guard let connection = headers["connection"].first?.lowercased() else {
+        guard let connection = headers[.connection].first?.lowercased() else {
             // HTTP 1.1 use keep-alive by default if not otherwise told.
             return version.major == 1 && version.minor == 1
         }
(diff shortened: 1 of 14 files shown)


normanmaurer:
Before:

```
➜  ~  wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://localhost:8888
  4 threads and 256 connections
    Latency   106.00ms  188.23ms   1.97s    92.33%
    Req/Sec   114.50k    39.96k  296.79k    79.02%
  4410634 requests in 10.07s, 210.32MB read
  Socket errors: connect 0, read 133, write 3, timeout 42
Requests/sec: 438012.73
Transfer/sec:     20.89MB
```

With this PR:
```
➜  ~  wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://localhost:8888
  4 threads and 256 connections
    Latency   106.23ms  196.92ms   1.99s    91.03%
    Req/Sec   116.80k    23.14k  251.81k    87.72%
  4650158 requests in 10.04s, 221.74MB read
  Socket errors: connect 0, read 151, write 0, timeout 52
Requests/sec: 463370.93
Transfer/sec:     22.10MB
```

normanmaurer:
As a side-note... One all the outstanding HTTP PRs are merged I think we should move HTTPHeaders related code to another file as its a lot of code now ;)

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+
+extension HTTPHeaderName: CustomStringConvertible {
+    public var description: String {
+        return self.value
+    }
+}
+
+/// Defines commonly used HTTP header names.
+public extension HTTPHeaderName {
+    public static let aIM = HTTPHeaderName("A-IM")
```
normanmaurer: @weissi @Lukasa I wonder if we should use "lower-case" for all of the header names because of HTTP2 tho ... WDYT ?

Lukasa: Yes, I think we should. TitleCase is silly.

normanmaurer: ok will do :)

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+/// Known HTTP header name. These should be created once and re-used as instances are memory intensive and "expensive" to create.
+public struct HTTPHeaderName {
+
+    /// Holds the bytes of the name + ': '
+    fileprivate let bytesWithSeperator: ContiguousArray<UInt8>
+    /// Holds the lowercased bases (masked) of the name.
+    fileprivate let lowerCaseBytes: ContiguousArray<UInt8>
+
+    /// Lowercase version of the name.
+    fileprivate let lowerCaseName: String
```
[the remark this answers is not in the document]
normanmaurer: I just used it for `Equatable`

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+    public static let wwwAuthenticate = HTTPHeaderName("WWW-Authenticate")
+    public static let wantDigest = HTTPHeaderName("Want-Digest")
+    public static let warning = HTTPHeaderName("Warning")
+    public static let xDeviceAccept = HTTPHeaderName("X-Device-Accept")
+    public static let xDeviceAcceptCharset = HTTPHeaderName("X-Device-Accept-Charset")
+    public static let xDeviceAcceptEncoding = HTTPHeaderName("X-Device-Accept-Encoding")
+    public static let xDeviceAcceptLanguage = HTTPHeaderName("X-Device-Accept-Language")
+    public static let xDeviceUserAgent = HTTPHeaderName("X-Device-User-Agent")
+    public static let xRequestedWith = HTTPHeaderName("X-Requested-With")
+    public static let xFrameOptions = HTTPHeaderName("X-Frame-Options")
```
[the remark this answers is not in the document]
normanmaurer: Vapor ;) Have any good list of the ones we want to include ?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+    func equalCaseInsensitiveASCII(nameType: HTTPHeaderNameType, at index: HTTPHeaderIndex) -> Bool {
+        switch nameType {
+        case .headerName(let name):
+            return self.equalCaseInsensitiveASCII(name: name, at: index)
+        case .string(let string):
+            return self.equalCaseInsensitiveASCII(view: string.utf8, at: index)
+        }
+    }
+
+    private func equalCaseInsensitiveASCII(name: HTTPHeaderName, at index: HTTPHeaderIndex) -> Bool {
```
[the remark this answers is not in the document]
normanmaurer: @Lukasa yeah that may be "cleaner". Let me do this.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+    }
+
+    var maskedLowercase: UInt8 {
+        return self & 0xdf
+    }
+}
+
+
+/// Known HTTP header name. These should be created once and re-used as instances are memory intensive and "expensive" to create.
+public struct HTTPHeaderName {
```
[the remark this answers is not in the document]
Lukasa: Sure, boxing it is fine too.

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
Lukasa: Why? I think `ContiguousArray` has an `Equatable` conformance.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+    /// Holds the bytes of the name + ': '
+    fileprivate let bytesWithSeperator: ContiguousArray<UInt8>
+    /// Holds the lowercased bases (masked) of the name.
+    fileprivate let lowerCaseBytes: ContiguousArray<UInt8>
+
+    /// Lowercase version of the name.
+    fileprivate let lowerCaseName: String
+
+    /// Original version of the name.
+    private let value: String
```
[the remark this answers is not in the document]
Lukasa: Eh, I think throw it away, save ourselves the bytes and use a lower-case description.

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
normanmaurer: @Lukasa good point!

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
Lukasa: Sadly, not really. That said, we're missing "Content-Security-Policy", "Content-Security-Policy-Report-Only", "X-XSS-Protection", & "X-Content-Type-Options" as some choices.

Also I just kinda noticed that this isn't in alphabetical order. It *almost* is, but not quite.

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
normanmaurer: @Lukasa we can also recreate the String for the array... Let me do this for now (as this will also be inline with what the user used to construct it).

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
normanmaurer: @Lukasa actually I can just make it a class and move on... WDYT ?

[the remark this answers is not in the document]
Lukasa: Ping @weissi, but I'm fine with this being a class.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+        return self & 0xdf
+    }
+}
+
+
+/// Known HTTP header name. These should be created once and re-used as instances are memory intensive and "expensive" to create.
+public struct HTTPHeaderName {
+
+    /// Holds the bytes of the name + ': '
+    fileprivate let bytesWithSeperator: ContiguousArray<UInt8>
```
Lukasa: Nit: separator.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
+    /// Holds the lowercased bases (masked) of the name.
+    fileprivate let lowerCaseBytes: ContiguousArray<UInt8>
```
weissi: why isn't this a slice on `bytesWithSeparator`?

weissi:
@normanmaurer what Swift version did you do the perf tests on? Swift _should_ AFAIK now store ASCII Strings as contiguous ASCII blocks so I'm unsure why there's such a perf difference

normanmaurer:
4.1... also the perf win may be related to the fact that we also include the separator in the byte array

weissi:
@normanmaurer ok, fair enough. We should analyse this.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -381,13 +779,26 @@ public struct HTTPHeaders: CustomStringConvertible {
     ///     - name: The name of the header
     //  - returns: `true` if a header with the name (and value) exists, `false` otherwise.
     public func contains(name: String) -> Bool {
+        return self.contains(nameType: .string(name))
+
```
weissi: nit: empty line

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -176,33 +176,332 @@ struct HTTPHeader {
...
             return true
         }
     }
-}
 
-
-private extension UInt8 {
-    var isASCII: Bool {
-        return self <= 127
+    private func equalCaseInsensitiveASCII(view: String.UTF8View, at index: HTTPHeaderIndex) -> Bool {
```
weissi: both those `equalCaseInsensitiveASCII` should be expressible as functions on `<S: Sequence>(bytes: S) where S.Element == UInt8`

normanmaurer: I think this lacks .count

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
normanmaurer: Ok fair enough

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: Because the other may not be lowercase?

Sources/NIOHTTP1/HTTPTypes.swift
weissi: @normanmaurer then on `Collection`

Sources/NIOHTTP1/HTTPTypes.swift
weissi: @normanmaurer ah right! What about using `bytesWithSeparator.dropLast(2).lazy.map { $0.makeLowerCase() }` that should just lazily make them lowercase when needed. Sure, it's a tiny computation overhead but we're comparing them one by one anyway... Not sure we should be storing all headers twice...

weissi: `makeLowerCase` would turn a character lower case, ie. just setting a bit in ASCII

weissi: @normanmaurer actually your `maskedLowercase` is exactly that already

normanmaurer: @weissi honestly I think its fine to pre-compute all of this as this is really meant to be created once and reused. So I would trade memory for performance. WDYT ?

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
normanmaurer: done

normanmaurer:
@Lukasa @weissi PTAL again

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -174,17 +174,323 @@ struct HTTPHeader {
...
+            return string.lowercased()
+        }
+    }
+}
+
+/// Extensions dealing with `HTTPHeaderNameType`s.
 private extension ByteBuffer {
-    func equalCaseInsensitiveASCII(view: String.UTF8View, at index: HTTPHeaderIndex) -> Bool {
-        guard view.count == index.length else {
+    /// Writes the `HTTPHeaderNameType` and also the separator (`: `). It returns the length of the name (without the seperator).
```
Lukasa: Why does this return that strange number?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -174,17 +174,323 @@ struct HTTPHeader {
...
+    mutating func writeWithSeperator(nameType: HTTPHeaderNameType) -> Int {
```
Lukasa: Nit: "separator". My mnemonic for this from when I was a kid is "there's *a rat* in sep*arat*e"

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -405,10 +814,25 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    /// one line are returned as separate entries. Also respects the fact that Set-Cookie should not
+    /// be split in this way.
+    ///
+    /// - Parameter name: The header field `HTTPHeaderName` whose values are to be retrieved.
+    /// - Returns: A list of the values for that header field name.
+    public subscript(canonicalForm name: HTTPHeaderName) -> [String] {
+        return self.getCanonicalFrom(nameType: .headerName(name))
+    }
+
+    private func getCanonicalFrom(nameType: HTTPHeaderNameType) -> [String] {
```
Lukasa: Should this be a private subscript?

weissi:
paging @gottesmm isn't this exactly the kind of operation that should be fast enough with the new Swift 4.1 `String`s? I was under the impression that for guaranteed ASCII-only strings that we don't really do any operations to (except `str.utf8`), the new `String` should be as good as `[UInt8]`?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -174,17 +174,323 @@ struct HTTPHeader {
...
+        /// Holds the lowercased bases (masked) of the name.
+        let lowerCaseBytes: ContiguousArray<UInt8>
+
+        let lowerCaseName: String
+
+        init(_ value: String) {
+            precondition(!value.utf8.contains(where: { !$0.isASCII }), "name must be ASCII")
+            var array = ContiguousArray(value.utf8)
+            array.append(contentsOf: ": ".utf8)
+            self.bytesWithSeparator = array
```
weissi: @gottesmm this is what I mean, we have a guaranteed ASCII string (append `: `) and store it as `[UInt8]` here. I would have hoped that this isn't necessary anymore.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -174,17 +174,323 @@ struct HTTPHeader {
...
+            self.lowerCaseBytes = ContiguousArray(value.utf8.map { $0.maskedLowercase })
```
weissi: @gottesmm and this I would've hoped we can hold as a stored property that returns `string.utf8.lazy.map { $0.maskedLowercase }` so no extra storage needed.

weissi: actually, if we do store lower case, we can just use `memcmp`

normanmaurer:
@weissi again the perf win may be even because we can write the name + separator in one go... Should I remove this optimization and test again ?

weissi:
@normanmaurer but with `String` we could also just append `": "` and write them in one go

normanmaurer:
@weissi but this would case a copy no ?

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: lol... thats a good one :)

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: yep sure why not :)

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: @Lukasa just because thats what the is returned when `String` is used instead and so we not need to special case it.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -174,17 +174,323 @@ struct HTTPHeader {
     let value: HTTPHeaderIndex
 }
 
+
+private extension UInt8 {
+    var isASCII: Bool {
+        return self <= 127
+    }
+
+    var maskedLowercase: UInt8 {
```
weissi: should we name this `maskedACIILowercase`?

Sources/NIOHTTP1/HTTPTypes.swift
weissi: @normanmaurer I know it's only invoked very few times but you're missing a `.lazy` here. Also this would be quicker:

    self.lowerCaseBytes = ContiguousArray(self.bytesWithSeparator.lazy.dropLast(2).map { $0.maskedLowercase })

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -405,10 +814,25 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    public subscript(canonicalForm name: HTTPHeaderName) -> [String] {
+        return self[canonicalForm: .headerName(name)]
+    }
+
+    private subscript(canonicalForm nameType: HTTPHeaderNameType) -> [String] {
+        let result = self[name: nameType]
 
         // It's not safe to split Set-Cookie on comma.
-        guard name.lowercased() != "set-cookie" else {
+        guard nameType.lowercased() != "set-cookie" else {
```
weissi: whilst we're here I think `lowercased` allocates, can you please check?

Sources/NIOHTTP1/HTTPTypes.swift
weissi: this is also for @milseman

weissi:
closing this as we very much need to reconsider this for Swift 5

PR #309, opened 2018-04-12
fix a crash when async connect failed (#302)

### LINKED ISSUE #302: New assert failure with 1.4.1
I've been working on rebasing Kitura on top of swift-nio. Things were all good with 1.3.1. However, with 1.4.1, I see a new assert failure at `Sources/NIO/BaseSocketChannel.swift, line 384`. I tried switching back to 1.3.1 and the assert failure goes away. 

This is the assert: 
`assert(self.lifecycleManager.isActive)`

I'm not sure what causes the failure. Can someone please explain, from an API usage standpoint, what could lead to this? Thanks!

weissi:
Motivation:

When an asynchronous connect failed we would crash because we didn't
correctly close the connection to start tearing everything down.
That correctly made the `SocketChannelLifecycleManager` trip.

This was filed as #302.

Modifications:

Correctly deregister and close the connection if an async connect fails.

Result:

Fewer crashes, fixes #302.


Tests/NIOTests/ChannelTests.swift
```diff
@@ -2019,4 +2019,86 @@ public class ChannelTests: XCTestCase {
...
+            }
+            private var state: State = .fresh
+
+            init(allDone: EventLoopPromise<Void>) {
+                self.allDone = allDone
+            }
+            deinit {
+                XCTAssertEqual(.unregistered, self.state)
+            }
+            func channelActive(ctx: ChannelHandlerContext) {
```
normanmaurer: nit: But an empty line between func`s so its easier to read ?

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2019,4 +2019,86 @@ public class ChannelTests: XCTestCase {
         try allDone.futureResult.wait()
         XCTAssertNoThrow(try sc.syncCloseAcceptingAlreadyClosed())
     }
+
+    func testSocketFailingAsyncCorrectlyTearsTheChannelDownAndDoesntCrash() throws {
+        // regression test for #302
+        enum DummyError: Error { case dummy }
+        class SocketFailingAsyncConnect: Socket {
+            private var firstReadHappened = false
```
Lukasa: This isn't used for anything.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2019,4 +2019,86 @@ public class ChannelTests: XCTestCase {
...
+                self.allDone.succeed(result: ())
+            }
+        }
+        let group = MultiThreadedEventLoopGroup(numThreads: 2)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let serverEL = group.next()
+        let clientEL = group.next()
+        precondition(serverEL !== clientEL)
```
Lukasa: I don't think we need the two threads here: it's totally fine for them to run on the one thread.

Tests/NIOTests/ChannelTests.swift
weissi: @normanmaurer can do if you want, I removed them because I thought it's  soo many lines that are all boring. totally up to you though

Tests/NIOTests/ChannelTests.swift
weissi: totally unpicking my copy&paste here :D

Tests/NIOTests/ChannelTests.swift
weissi: @normanmaurer proposed a new version with spaces

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2019,4 +2019,81 @@ public class ChannelTests: XCTestCase {
...
+                case unregistered
+            }
+            private var state: State = .fresh
+
+            init(allDone: EventLoopPromise<Void>) {
+                self.allDone = allDone
+            }
+            deinit { XCTAssertEqual(.unregistered, self.state) }
+
+            func channelActive(ctx: ChannelHandlerContext) { XCTFail("should never become active") }
```
weissi: @normanmaurer / @Lukasa  what do you think about this, 'folding' the boring lines?

Lukasa: Totally file with that.

PR #313, opened 2018-04-13
Use power of two for underlying storage of CircularBuffer

normanmaurer:
Motivation:

We should use power of two for the underlying storage to allow us to make use of bitmasking which is faster then modulo operations.

Modifications:

Use power of two and replace module with bitmasking.

Result:

Faster implementation of CircularBuffer.

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -26,62 +26,61 @@ public protocol AppendableCollection: Collection {
 public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     private var buffer: ContiguousArray<E?>
 
-    /// The capacity of the underlying buffer
-    private var bufferCapacity: Int
-
     /// The index into the buffer of the first item
-    private var startIdx = 0
+    private(set) /* private but tests */ internal var headIdx = 0
 
     /// The index into the buffer of the next free slot
-    private var endIdx = 0
+    private(set) /* private but tests */ internal var tailIdx = 0
 
-    /// The number of items in the ring part of this buffer
-    private var ringLength = 0
+    /// Bitmask used for calculating the tailIdx / headIdx based on the fact that the underlying storage
+    /// has always a size of power of two.
+    private var mask: Int {
+        return self.buffer.count - 1
+    }
 
     /// Allocates a buffer that can hold up to `initialRingCapacity` elements and initialise an empty ring backed by
     /// the buffer. When the ring grows to more than `initialRingCapacity` elements the buffer will be expanded.
     public init(initialRingCapacity: Int) {
-        self.bufferCapacity = initialRingCapacity
-        self.buffer = ContiguousArray<E?>(repeating: nil, count: Int(initialRingCapacity))
+        let capacity = Int(UInt32(initialRingCapacity).nextPowerOf2())
+        self.buffer = ContiguousArray<E?>(repeating: nil, count: capacity)
+        assert(self.buffer.count == capacity)
     }
 
     /// Append an element to the end of the ring buffer.
     ///
     /// Amortized *O(1)*
     public mutating func append(_ value: E) {
-        let expandBuf: Bool = self.bufferCapacity == self.ringLength
+        self.buffer[self.tailIdx] = value
+        self.tailIdx = (self.tailIdx + 1) & self.mask
 
-        if expandBuf {
+        if self.headIdx == self.tailIdx {
+            // No more room left for another append so grow the buffer now.
             var newBacking: ContiguousArray<E?> = []
-            let newCapacity = Swift.max(1, 2 * self.bufferCapacity)
+            let newCapacity = self.buffer.count << 1 // Double the storage.
+            precondition(newCapacity > 0, "Can't double capacity of \(self.buffer.count)")
+            assert(newCapacity % 2 == 0)
+
             newBacking.reserveCapacity(newCapacity)
-            newBacking.append(contentsOf: self.buffer[self.startIdx..<self.bufferCapacity])
-            if startIdx > 0 {
-                newBacking.append(contentsOf: self.buffer[0..<self.startIdx])
+            newBacking.append(contentsOf: self.buffer[self.headIdx..<self.buffer.count])
+            if self.headIdx > 0 {
+                newBacking.append(contentsOf: self.buffer[0..<self.headIdx])
             }
             newBacking.append(contentsOf: repeatElement(nil, count: newCapacity - newBacking.count))
+            self.tailIdx = self.buffer.count
+            self.headIdx = 0
             self.buffer = newBacking
-            self.startIdx = 0
-            self.endIdx = self.ringLength
-            self.bufferCapacity = newCapacity
-            precondition(self.bufferCapacity == self.buffer.count)
         }
-
-        self.buffer[self.endIdx] = value
-        self.ringLength += 1
-        self.endIdx = (self.endIdx + 1) % self.bufferCapacity
     }
 
     /// Remove the front element of the ring buffer.
     ///
     /// *O(1)*
     public mutating func removeFirst() -> E {
-        precondition(self.ringLength != 0)
+        precondition(!self.isEmpty)
 
-        let value = self.buffer[self.startIdx]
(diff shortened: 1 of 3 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -20,6 +20,28 @@ public protocol AppendableCollection: Collection {
...
+
+        var n = self
+
+        n -= 1
+        n |= n >> 1
+        n |= n >> 2
+        n |= n >> 4
+        n |= n >> 8
+        n |= n >> 16
+        n |= n >> 32
```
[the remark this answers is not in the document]
normanmaurer: yeah thats a good point... Let me just use UInt32 and move on. That should be plenty of space. WDYT ?

normanmaurer:
@weissi I see no differences with our network benchmarks (which is not really unexpected atm). But I think we all know that modulo is considered slow compared to bit masking.

Sources/NIO/CircularBuffer.swift
```diff
@@ -41,8 +63,9 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     /// Allocates a buffer that can hold up to `initialRingCapacity` elements and initialise an empty ring backed by
     /// the buffer. When the ring grows to more than `initialRingCapacity` elements the buffer will be expanded.
     public init(initialRingCapacity: Int) {
-        self.bufferCapacity = initialRingCapacity
-        self.buffer = ContiguousArray<E?>(repeating: nil, count: Int(initialRingCapacity))
+        let capacity = initialRingCapacity.nextPowerOf2()
```
weissi: do the tests really pass with this? The need a careful look if the pass as they definitely try to test the edge cases on the edge of extension

normanmaurer: yep they pass

weissi:
@normanmaurer sure, I thought a micro benchmark but it's unlikely to be worse so maybe it's fine

Sources/NIO/CircularBuffer.swift
weissi: @normanmaurer then they will most likely need adjustment to still test the edge cases (unless they happen to have picked only powers of 2). Did you check the code coverage?

normanmaurer: nope not yet

normanmaurer:
@Lukasa @weissi PTAL again... I also did a few more changes to make the whole data structure a bit smaller.

Sources/NIO/CircularBuffer.swift
```diff
@@ -129,17 +123,17 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
...
     /// Returns whether the ring is empty.
     public var isEmpty: Bool {
-        return self.ringLength == 0
+        return self.startIdx == self.endIdx
     }
 
     /// Returns the number of element in the ring.
     public var count: Int {
-        return self.ringLength
+        return (self.endIdx - self.startIdx) & self.mask
```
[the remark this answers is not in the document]
weissi: yes, that's correct. Can we also have a test case for that please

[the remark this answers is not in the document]
normanmaurer: I think it is actually correct ( the impl) let me add a unit test for it and either prove me or you wrong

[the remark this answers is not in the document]
normanmaurer: ok added test for it... works as expected (thanks to bitwise ops ;))

normanmaurer:
Will add more tests tomorrow

Sources/NIO/CircularBuffer.swift
```diff
@@ -26,62 +26,59 @@ public protocol AppendableCollection: Collection {
...
-        let expandBuf: Bool = self.bufferCapacity == self.ringLength
+        self.buffer[self.endIdx] = value
+        self.endIdx = (self.endIdx + 1) & self.mask
 
-        if expandBuf {
+        if startIdx == endIdx {
+            // No more room left for another append so grow the buffer now.
             var newBacking: ContiguousArray<E?> = []
-            let newCapacity = Swift.max(1, 2 * self.bufferCapacity)
+            let newCapacity = self.buffer.count << 1 // Double the storage.
```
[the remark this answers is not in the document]
normanmaurer: @weissi IMHO its not really guaranteed its as fast... generally speaking bitshifting should be faster.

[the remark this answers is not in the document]
Lukasa: Any optimising compiler in the land will trivially see that multiply-by-two is the same as bit shift left by one.

Tests/NIOTests/CircularBufferTests.swift
```diff
@@ -207,4 +207,70 @@ class CircularBufferTests: XCTestCase {
...
+
+
+
+
+
+
+
+
+
+
```
Lukasa: Um?

Sources/NIO/CircularBuffer.swift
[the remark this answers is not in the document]
normanmaurer: @Lukasa I learned the hard way to not trust these ;) That said if you are both strong I can remove it... that said I think its not really hard to understand as it is.

Sources/NIO/CircularBuffer.swift
```diff
@@ -26,62 +26,60 @@ public protocol AppendableCollection: Collection {
 public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     private var buffer: ContiguousArray<E?>
 
-    /// The capacity of the underlying buffer
-    private var bufferCapacity: Int
-
     /// The index into the buffer of the first item
-    private var startIdx = 0
+    private(set) /* private but tests */ internal var startIdx = 0
```
Lukasa: If you're going to expose these as internal I think these variables need different names. They are dangerously close to shadowing the variables `startIndex` and `endIndex` that come from `Collection`, and yet they behave quite differently.

Sources/NIO/CircularBuffer.swift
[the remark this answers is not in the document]
Lukasa: 👍

Sources/NIO/CircularBuffer.swift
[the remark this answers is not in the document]
Lukasa: I mean, this is a trivial peephole optimisation. :wink:

Tests/NIOTests/CircularBufferTests.swift
normanmaurer: more empty lines ftw... seriously let me cleanup while adding more tests

Sources/NIO/CircularBuffer.swift
Lukasa: Up to @weissi.

Tests/NIOTests/CircularBufferTests.swift
normanmaurer: I wonder if I felt asleep while holding the enter key 🤷‍♂️

Sources/NIO/CircularBuffer.swift
weissi: I like `headIdx` and `tailIdx`, then it's also more obvious that head can be larger than tail

normanmaurer:
@weissi @Lukasa I enhanced the tests a bit and addressed comments. Can you let me know what kind of other tests you would like to see (if any)?

Sources/NIO/CircularBuffer.swift
```diff
@@ -26,62 +26,59 @@ public protocol AppendableCollection: Collection {
...
-    private var startIdx = 0
+    private(set) /* private but tests */ internal var headIdx = 0
 
     /// The index into the buffer of the next free slot
-    private var endIdx = 0
+    private(set) /* private but tests */ internal var tailIdx = 0
 
-    /// The number of items in the ring part of this buffer
-    private var ringLength = 0
+    private var mask: Int {
```
Lukasa: Can you add a doc comment for `mask`?

normanmaurer: addressed

normanmaurer:
@weissi @Lukasa PTAL again

PR #315, opened 2018-04-16
Allow to specify the max websockets frame size when using the upgrader.

normanmaurer:
Motivation:

At the moment its not possible to adjust the max websockets frame size when using the upgrader while its possible when directly use the WebSocketFrameDecoder.

Modifications:

Allow to specify max frame size via an init argument (with default value).

Result:

More flexible way to use the upgrader.


normanmaurer:
also @vlm

normanmaurer:
@Lukasa done

Sources/NIOWebSocket/WebSocketFrameDecoder.swift
```diff
@@ -221,7 +221,7 @@ public final class WebSocketFrameDecoder: ByteToMessageDecoder {
     public var cumulationBuffer: ByteBuffer? = nil
 
     /// The maximum frame size the decoder is willing to tolerate from the remote peer.
-    private let maxFrameSize: Int
+    public let maxFrameSize: Int
```
normanmaurer: we can also keep this internal if we like

Sources/NIOWebSocket/WebSocketUpgrader.swift
```diff
@@ -73,10 +74,21 @@ public final class WebSocketUpgrader: HTTPProtocolUpgrader {
...
+    ///         possible to even allocate a buffer large enough to handle this size), so we
+    ///         set a lower one. The default value is the same as the default HTTP/2 max frame
+    ///         size, `2**14` bytes. Users may override this to any value up to `UInt32.max`.
+    ///         Users are strongly encouraged not to increase this value unless they absolutely
+    ///         must, as the decoder will not produce partial frames, meaning that it will hold
+    ///         on to data until the *entire* body is received.
     public init(shouldUpgrade: @escaping (HTTPRequestHead) -> HTTPHeaders?,
-                upgradePipelineHandler: @escaping (Channel, HTTPRequestHead) -> EventLoopFuture<Void>) {
+                upgradePipelineHandler: @escaping (Channel, HTTPRequestHead) -> EventLoopFuture<Void>, maxFrameSize: Int = 1 << 14) {
+        precondition(maxFrameSize <= UInt32.max, "invalid overlarge max frame size")
```
weissi: if you're already preconditioning here, mind checking it's positive too?

Sources/NIOWebSocket/WebSocketFrameDecoder.swift
weissi: my vote would go to let's keep it internal :)

Lukasa: Yup, I think we should keep it internal too.

Lukasa: It's always easier to make things more public than to make them less public.

Sources/NIOWebSocket/WebSocketUpgrader.swift
Lukasa: Is there any point in this precondition? I think it just duplicates logic that exists in the web socket frame decoder.

Sources/NIOWebSocket/WebSocketUpgrader.swift
```diff
@@ -73,10 +74,21 @@ public final class WebSocketUpgrader: HTTPProtocolUpgrader {
...
+    ///         this is an objectively unreasonable maximum value (on AMD64 systems it is not
+    ///         possible to even allocate a buffer large enough to handle this size), so we
+    ///         set a lower one. The default value is the same as the default HTTP/2 max frame
+    ///         size, `2**14` bytes. Users may override this to any value up to `UInt32.max`.
+    ///         Users are strongly encouraged not to increase this value unless they absolutely
+    ///         must, as the decoder will not produce partial frames, meaning that it will hold
+    ///         on to data until the *entire* body is received.
     public init(shouldUpgrade: @escaping (HTTPRequestHead) -> HTTPHeaders?,
-                upgradePipelineHandler: @escaping (Channel, HTTPRequestHead) -> EventLoopFuture<Void>) {
+                upgradePipelineHandler: @escaping (Channel, HTTPRequestHead) -> EventLoopFuture<Void>, maxFrameSize: Int = 1 << 14) {
```
Lukasa: Can I recommend that instead of making this change you just add a new constructor that this one calls, with `maxFrameSize` earlier in the argument list? Trailing closures are really nice, and it's a shame to lose them.

Sources/NIOWebSocket/WebSocketUpgrader.swift
normanmaurer: @Lukasa imho its better to fail fast... WDYT ?

Sources/NIOWebSocket/WebSocketUpgrader.swift
```diff
@@ -73,10 +74,44 @@ public final class WebSocketUpgrader: HTTPProtocolUpgrader {
...
+    ///         websocket protocol. This only needs to add the user handlers: the
+    ///         `WebSocketFrameEncoder` and `WebSocketFrameDecoder` will have been added to the
+    ///         pipeline automatically.
+    public init(maxFrameSize: Int, shouldUpgrade: @escaping (HTTPRequestHead) -> HTTPHeaders?,
+                upgradePipelineHandler: @escaping (Channel, HTTPRequestHead) -> EventLoopFuture<Void>) {
+        precondition(maxFrameSize <= UInt32.max, "invalid overlarge max frame size")
         self.shouldUpgrade = shouldUpgrade
         self.upgradePipelineHandler = upgradePipelineHandler
+        self.maxFrameSize = maxFrameSize
     }
```
vlm: Curious, what is the motivation for providing two constructor rather than a constructor with the default value?

Sources/NIOWebSocket/WebSocketFrameDecoder.swift
```diff
@@ -221,7 +221,7 @@ public final class WebSocketFrameDecoder: ByteToMessageDecoder {
     public var cumulationBuffer: ByteBuffer? = nil
 
     /// The maximum frame size the decoder is willing to tolerate from the remote peer.
-    private let maxFrameSize: Int
+    /* private but tests */ let maxFrameSize: Int
```
vlm: I think it make even more sense to test behavior (e.g., ability _and_ inability to transfer more than the limit when the limit is appropriately changed) rather than checking whether we set the `maxFrameSize`. It'll have a nice side effect of eliminating this comment.

normanmaurer: @vlm I will add more tests as a followup to test max frame size handling. But I think this is ok to test what the change did.

Sources/NIOWebSocket/WebSocketUpgrader.swift
normanmaurer: @vlm https://github.com/apple/swift-nio/pull/315#discussion_r181697306

Lukasa: @vlm Short answer is "to preserve the trailing closure".

PR #316, opened 2018-04-16
reduce allocations & test we don't regress

weissi:
Motivation:

The ping/pong test should only do 4 allocations:
- readFromSocket does unconditionally allocate one ByteBuffer
  (allocating one ByteBuffer does 2 allocations)
- we have a ping and a pong side, each call readFromSocket

so 2 * 2 = 4 allocations per message. Before adding the `@_inlineable`
to the read/write/get/set integer methods in ByteBuffer however I was
seeing way more allocations. So looked at them and fixed them.

Turns out that massively reduces the number of allocations for HTTP too.

On Linux I was seeing this change:

1000_reqs_1_conn: 282000 -> 70000
1_reqs_1000_conn: 1287000 -> 907000

Modifications:

- made the ByteBuffer integer methods inlineable
- added a test that makes sure we don't regress on the simplest possible
  use of NIO: a ping/pong server

Result:

After your change, what will change.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
@@ -50,6 +50,81 @@ private final class SimpleHTTPServer: ChannelInboundHandler {
     }
 }
 
+private struct PingPongFailure: Error, CustomStringConvertible {
+    public var description: String
+
+    init(problem: String) {
+        self.description = problem
+    }
+}
+
+private final class PingHandler: ChannelInboundHandler {
+    typealias InboundIn = ByteBuffer
+    typealias OutboundOut = ByteBuffer
+
+    private var pingBuffer: ByteBuffer!
+    private let numberOfRequests: Int
+    private var remainingNumberOfRequests: Int
+    private let allDone: EventLoopPromise<Void>
+    public static let pingCode: UInt8 = 0xbe
+
+    public init(numberOfRequests: Int, eventLoop: EventLoop) {
+        self.numberOfRequests = numberOfRequests
+        self.remainingNumberOfRequests = numberOfRequests
+        self.allDone = eventLoop.newPromise()
+    }
+
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pingBuffer = ctx.channel.allocator.buffer(capacity: 1)
+        self.pingBuffer.write(integer: PingHandler.pingCode)
+
+        ctx.writeAndFlush(self.wrapOutboundOut(self.pingBuffer), promise: nil)
+    }
+
+    public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        var buf = self.unwrapInboundIn(data)
+        if buf.readableBytes == 1 &&
+            buf.readInteger(as: UInt8.self) == PongHandler.pongCode {
+            if self.remainingNumberOfRequests > 0 {
+                self.remainingNumberOfRequests -= 1
+                ctx.writeAndFlush(self.wrapOutboundOut(self.pingBuffer), promise: nil)
+            } else {
+                ctx.close(promise: self.allDone)
+            }
+        } else {
+            ctx.close(promise: nil)
+            self.allDone.fail(error: PingPongFailure(problem: "wrong buffer received: \(buf.debugDescription)"))
+        }
+    }
+
+    public func wait() throws {
+        try self.allDone.futureResult.wait()
+    }
+}
+
+private final class PongHandler: ChannelInboundHandler {
+    typealias InboundIn = ByteBuffer
+    typealias OutboundOut = ByteBuffer
+
+    private var pongBuffer: ByteBuffer!
+    public static let pongCode: UInt8 = 0xef
+
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pongBuffer = ctx.channel.allocator.buffer(capacity: 1)
+        self.pongBuffer.write(integer: PongHandler.pongCode)
+    }
+
+    public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        var buf = self.unwrapInboundIn(data)
+        if buf.readableBytes == 1 &&
+            buf.readInteger(as: UInt8.self) == PingHandler.pingCode {
+            ctx.writeAndFlush(self.wrapOutboundOut(self.pongBuffer), promise: nil)
+        } else {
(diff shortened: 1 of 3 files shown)


IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
```diff
@@ -50,6 +50,85 @@ private final class SimpleHTTPServer: ChannelInboundHandler {
...
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pingBuffer = ctx.channel.allocator.buffer(capacity: 4)
+        self.pingBuffer.write(staticString: "ping")
+
+        ctx.writeAndFlush(self.wrapOutboundOut(self.pingBuffer), promise: nil)
+    }
+
+    public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        var buf = self.unwrapInboundIn(data)
+        if buf.readableBytes == 4 &&
```
normanmaurer: This is not correct... as tcp is Stream based and can be fragmented we need to use a ByteToMessageDecoder here and keep
Buffering if the readable bytes are smaller then 4

normanmaurer: Or just keep state by ourself

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
```diff
@@ -50,6 +50,85 @@ private final class SimpleHTTPServer: ChannelInboundHandler {
...
+    private var pongBuffer: ByteBuffer!
+
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pongBuffer = ctx.channel.allocator.buffer(capacity: 4)
+        self.pongBuffer.write(staticString: "pong")
+    }
+
+    public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+        var buf = self.unwrapInboundIn(data)
+        if buf.readableBytes == 4 &&
```
Lukasa: Same note applies here.

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -13,7 +13,8 @@
 //===----------------------------------------------------------------------===//
 
 extension ByteBuffer {
-    private func toEndianness<T: FixedWidthInteger> (value: T, endianness: Endianness) -> T {
+    @_inlineable @_versioned
```
Lukasa: Is both right here?

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
weissi: Oh sure but I mean it’s loopback and only ever 4 bytes in transit. I thought about just making it 1 byte to not have this problem but then thought that’s over the top as that’ll actually not happen.
But happy to change it to 1 byte.
Buffering won’t work as we count every allocation and if we were to buffer we’d add allocations.
I think 4 bytes as is is fine but can change it to one byte if you want.

Sources/NIO/ByteBuffer-int.swift
weissi: Yes, needs to be versioned so it can be used from inlineable but it can also be inlineable. Just gives the compiler more options.

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
normanmaurer: I think 1 byte would be a safer bet here.

weissi: fair enough. Didn't doo it because I thought sending `ping` and `pong` around is much more fun than `x` and `y`. Can't even send emoji around now 😢

weissi:
@normanmaurer / @Lukasa updated now, only sending 1 byte around

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
```diff
@@ -50,6 +50,81 @@ private final class SimpleHTTPServer: ChannelInboundHandler {
...
+    public static let pingCode: UInt8 = 0xbe
+
+    public init(numberOfRequests: Int, eventLoop: EventLoop) {
+        self.numberOfRequests = numberOfRequests
+        self.remainingNumberOfRequests = numberOfRequests
+        self.allDone = eventLoop.newPromise()
+    }
+
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pingBuffer = ctx.channel.allocator.buffer(capacity: 4)
```
normanmaurer: nit: this could be 1 now

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
```diff
@@ -50,6 +50,81 @@ private final class SimpleHTTPServer: ChannelInboundHandler {
...
+
+private final class PongHandler: ChannelInboundHandler {
+    typealias InboundIn = ByteBuffer
+    typealias OutboundOut = ByteBuffer
+
+    private var pongBuffer: ByteBuffer!
+    public static let pongCode: UInt8 = 0xef
+
+    public func channelActive(ctx: ChannelHandlerContext) {
+        self.pongBuffer = ctx.channel.allocator.buffer(capacity: 4)
```
normanmaurer: nit this could be 1 now

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
weissi: fixd

IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
weissi: fixed

weissi:
@normanmaurer addressed

Lukasa:
LGTM.

PR #321, opened 2018-04-17
Allow ChannelCore implementors to unwrap NIOAny.

Lukasa:
Motivation:

When implementing a custom ChannelCore, you will probably need access
to the data inside a NIOAny. It should be possible to unwrap that.

Modifications:

Added an extension to ChannelCore to allow unwrapping a NIOAny.
Added @_versioned to almost all of NIOAny.

Result:

ChannelCore is implementable outside NIO


Sources/NIO/NIOAny.swift
```diff
@@ -77,6 +77,7 @@ public struct NIOAny {
     /// Try unwrapping the wrapped message as `ByteBuffer`.
     ///
     /// returns: The wrapped `ByteBuffer` or `nil` if the wrapped message is not a `ByteBuffer`.
+    @_versioned
```
weissi: @Lukasa why not `@_inlineable` as well?

Lukasa: Not allowed on public methods according to my compiler.

Lukasa: The current reason is that `storage` is `private`. Do we think that's worth changing?

weissi: @Lukasa not for this particular one but possibly for the ones that take generics.

Lukasa: Yeah, I'm a bit on the fence here. I suppose making the storage internal won't be too big a deal.

weissi: me too tbh. I'll make it your call. If you make the storage internal, please underscore it

Lukasa: Ok, made all of these `@_inlineable`.

Sources/NIO/Channel.swift
```diff
@@ -244,6 +244,28 @@ public extension Channel {
     }
 }
 
+public extension ChannelCore {
+    /// Unwraps the given NIOAny as a specific concrete type.
```
normanmaurer: back-ticks around `NIOAny`

Lukasa: Applied.

weissi:
test failure should be addressed by #326

weissi:

PR #323, opened 2018-04-17
[2.0 change] Hard-fail on invalid write types in core channels.

Lukasa:
Motivation:

There is relatively little benefit in allowing the user to send invalid
data types on a given channel. Currently we tolerate this by simply
failing the promise, but this is not really ideal: if the user never sets
a promise here they never even find out that we bailed on sending their
data. Given that we attempt to be type safe everywhere else, we should
just blow up immediately if the user attempts to write an invalid data
type.

Modifications:

Removed the nice gentle type casts, replaced them with explody ones.

Result:

Faster failure in error cases.


Sources/NIO/SocketChannel.swift
```diff
@@ -290,11 +290,7 @@ final class SocketChannel: BaseSocketChannel<Socket> {
             return
         }
 
-        guard let data = data.tryAsIOData() else {
-            promise?.fail(error: ChannelError.writeDataUnsupported)
```
normanmaurer: I think we should also remove `ChannelError.writeDataUnsupported` completely ?

Lukasa: ACK, done.

weissi:
@Lukasa this conflicts now

Lukasa:
Updated.

weissi:

weissi:
@normanmaurer are you happy with this?

normanmaurer:
I am not strong on this honestly... I am fine either way.

weissi:
@Lukasa for you to decide

PR #286, opened 2018-04-05
eagerly process input EOF/connection resets (fixes #277)

weissi:
Motivation:

We had a number of problems:

1. We wanted to lazily process input EOFs and connection resets only
   when the user actually calls `read()`. On Linux however you cannot
   unsubscribe from `EPOLLHUP` so that's not possible.
2. Lazily processing input EOFs/connection resets wastes kernel
   resources and that could potentially lead to a DOS
3. The very low-level `Selector` interpreted the eventing mechanism's
   events quite a lot so the `EventLoop`/`Channel` only ever saw
   `readable` or `writable` without further information what exactly
   happened.
4. We completely ignored `EPOLLHUP` until now which on Unix Domain
   Socket close leads to a 100% CPU spin (issue #277)

Modifications:

- made the `Selector` interface richer, it now sends the following
  events: `readable`, `writable`, `readEOF` (input EOF), `reset`
  (connection reset or some error)
- process input EOFs and connection resets/errors eagerly
- change all tests which relied on using unconnected and unbound sockets
  to user connected/bound ones as `epoll_wait` otherwise would keep
  sending us a stream of `EPOLLHUP`s which would now lead to an eager
  close

Result:

- most importantly: fixes #277
- waste less kernel resources (by dealing with input EOFs/connection
  resets eagerly)
- bring kqueue/epoll more in line

PR #325, opened 2018-04-17
propagate socket errors on .reset

weissi:
Motivation:

Since #286, we do eagerly detect connection resets which is good. To
resolve the situation we first try to read everything off the socket. If
something fails there, everything is good and the client gets the
correct error. If that however doesn't lead to any errors, we would
close the connection with `error: ChannelError.eof` which the client
would then see for example on a connect that happened to get
`ECONNRESET` or `ECONNREFUSED`. That's not right so these situations are
fixed here.

Modifications:

- on reset, try to get the socket error if any
- if no socket error was detected, we send a `ECONNRESET` because we
  know we have been reset but not why and didn't encounter any errors
- write tests for these situations

Result:

connection resets that don't hit any other errors will now see the
correct error

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -798,7 +798,27 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
     // the `reset` event.
     final func reset() {
         self.readEOF()
-        self.close0(error: ChannelError.eof, mode: .all, promise: nil)
+
+        if self.socket.isOpen {
+            assert(self.lifecycleManager.isRegistered)
+            let error: IOError
+            // if the socket is still registered (and therefore open), let's try to get the actual socket error from the socket
+            do {
+                let result: Int32 = try self.socket.getOption(level: SOL_SOCKET, name: SO_ERROR)
+                if result != 0 {
+                    // we have a socket error, let's forward
+                    // this path will be executed on Linux (EPOLLERR) & Darwin (ev.fflags != 0)
+                    error = IOError(errnoCode: result, reason: "connection reset (error set)")
+                } else {
+                    // we don't have a socket error, this must be connection reset without an error then
+                    // this path should only be executed on Linux (EPOLLHUP, no EPOLLERR)
+                    error = IOError(errnoCode: ECONNRESET, reason: "connection reset (no error set)")
+                }
+                self.close0(error: error, mode: .all, promise: nil)
+            } catch {
+                self.close0(error: error, mode: .all, promise: nil)
+            }
+        }
         assert(!self.lifecycleManager.isRegistered)
     }
 
--- a/Tests/NIOTests/ChannelTests.swift
+++ b/Tests/NIOTests/ChannelTests.swift
@@ -2214,6 +2214,32 @@ public class ChannelTests: XCTestCase {
         XCTAssertNoThrow(try sc.closeFuture.wait())
         XCTAssertNoThrow(try sc.syncCloseAcceptingAlreadyClosed())
     }
+
+    func testConnectWithECONNREFUSEDGetsTheRightError() throws {
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let serverSock = try Socket(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+        // we deliberately don't set SO_REUSEADDR
+        XCTAssertNoThrow(try serverSock.bind(to: SocketAddress(ipAddress: "127.0.0.1", port: 0)))
+        let serverSockAddress = try! serverSock.localAddress()
+        XCTAssertNoThrow(try serverSock.close())
+
+        // we're just looping here to get a pretty good chance we're hitting both the synchronous and the asynchronous
+        // connect path.
+        for _ in 0..<64 {
+            do {
+                _ = try ClientBootstrap(group: group).connect(to: serverSockAddress).wait()
+                XCTFail("just worked")
+            } catch let error as IOError where error.errnoCode == ECONNREFUSED {
+                // OK
+            } catch {
+                XCTFail("unexpected error: \(error)")
+            }
+        }
+    }
+
 }
 
 fileprivate class VerifyConnectionFailureHandler: ChannelInboundHandler {
--- a/Sources/NIO/IO.swift
+++ b/Sources/NIO/IO.swift
@@ -53,17 +53,17 @@ public struct IOError: Swift.Error {
 ///
 /// - parameters:
 ///       - errorCode: the `errno` that was set for the operation.
-///       - function: the function / syscall that caused the error.
+///       - reason: what failed
 /// -returns: the constructed reason.
-private func reasonForError(errnoCode: Int32, function: StaticString) -> String {
+private func reasonForError(errnoCode: Int32, reason: String) -> String {
     if let errorDescC = strerror(errnoCode) {
         let errorDescLen = strlen(errorDescC)
         return errorDescC.withMemoryRebound(to: UInt8.self, capacity: errorDescLen) { ptr in
             let errorDescPtr = UnsafeBufferPointer<UInt8>(start: ptr, count: errorDescLen)
-            return "\(function) failed: \(String(decoding: errorDescPtr, as: UTF8.self)) (errno: \(errnoCode)) "
+            return "\(reason): \(String(decoding: errorDescPtr, as: UTF8.self)) (errno: \(errnoCode)) "
         }
     } else {
-        return "\(function) failed: Broken strerror, unknown error: \(errnoCode)"
+        return "\(reason): Broken strerror, unknown error: \(errnoCode)"
     }
 }
 
@@ -75,9 +75,9 @@ extension IOError: CustomStringConvertible {
     public var localizedDescription: String {
         switch self.reason {
         case .reason(let reason):
-            return reason
+            return reasonForError(errnoCode: self.errnoCode, reason: reason)
         case .function(let function):
-            return reasonForError(errnoCode: self.errnoCode, function: function)
+            return reasonForError(errnoCode: self.errnoCode, reason: "\(function) failed")
         }
     }
 }
--- a/Tests/NIOTests/ChannelTests+XCTest.swift
+++ b/Tests/NIOTests/ChannelTests+XCTest.swift
@@ -67,6 +67,7 @@ extension ChannelTests {
                 ("testCloseSocketWhenReadErrorWasReceivedAndMakeSureNoReadCompleteArrives", testCloseSocketWhenReadErrorWasReceivedAndMakeSureNoReadCompleteArrives),
                 ("testSocketFailingAsyncCorrectlyTearsTheChannelDownAndDoesntCrash", testSocketFailingAsyncCorrectlyTearsTheChannelDownAndDoesntCrash),
                 ("testSocketErroringSynchronouslyCorrectlyTearsTheChannelDown", testSocketErroringSynchronouslyCorrectlyTearsTheChannelDown),
+                ("testConnectWithECONNREFUSEDGetsTheRightError", testConnectWithECONNREFUSEDGetsTheRightError),
            ]
    }
 }


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -798,7 +798,29 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
     // the `reset` event.
     final func reset() {
         self.readEOF()
-        self.close0(error: ChannelError.eof, mode: .all, promise: nil)
+
+        if self.lifecycleManager.isRegistered {
+            // if the socket is still open, let's try to get the actual socket error from the socket
```
normanmaurer: @weissi hmm this comment is confusing as we not really check for open but for registered above. Can you clarify ?

weissi: good point. We close the socket when unregistering it but I should probably do something like this:

    if self.socket.isOpen {
        assert(self.lifecycleManager.isRegistered)
        ...

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -798,7 +798,29 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
...
+            let result: Int32
```
Lukasa: Does this need to be declared outside the `do`? It isn't used elsewhere.

Sources/NIO/BaseSocketChannel.swift
weissi: fixed

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -798,7 +798,30 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
...
+                } else {
+                    // we don't have a socket error, this must be connection reset without an error then
+                    // this path should only be executed on Linux (EPOLLHUP, no EPOLLERR)
+                    self.close0(error: IOError(errnoCode: ECONNRESET,
+                                               reason: "connection reset (no error set)"),
+                                mode: .all, promise: nil)
+                }
+            } catch {
+                self.close0(error: error, mode: .all, promise: nil)
+            }
```
Lukasa: This code could be cleaned up, I think, either by stack allocating an existential `Error` that each branch sets, and then calls `close0` with it, or by having the top two branches stack allocate a concrete `IOError`.

Either way we should aim to have fewer of these calls to `close0`.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -798,7 +798,30 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
     // the `reset` event.
     final func reset() {
         self.readEOF()
-        self.close0(error: ChannelError.eof, mode: .all, promise: nil)
+
+        if self.socket.isOpen {
+            assert(self.lifecycleManager.isRegistered)
+            // if the socket is still open, let's try to get the actual socket error from the socket
+            let result: Int32
```
Lukasa: This doesn't need to be declared outside the block, it's not used outside it.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,98 @@ public class ChannelTests: XCTestCase {
...
+        let serverSockSyncQ = DispatchQueue(label: "serverSock syncQ")
+        let serverSock = try serverSockSyncQ.sync { () -> Socket in
+            let sock = try Socket(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+            XCTAssertNoThrow(try sock.bind(to: SocketAddress(ipAddress: "127.0.0.1", port: 0)))
+            XCTAssertNoThrow(try sock.withUnsafeFileDescriptor { fd in
+                try Posix.listen(descriptor: fd, backlog: 1)
+            })
+            return sock
+        }
+        let serverSockAddress = serverSockSyncQ.sync { try! serverSock.localAddress() }
```
Lukasa: These two synchronous queue operations appear to be unnecessary.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,98 @@ public class ChannelTests: XCTestCase {
...
+        defer {
+            XCTAssertNoThrow(try serverSockSyncQ.sync { try serverSock.close() })
+        }
+        let g = DispatchGroup()
+        let q = DispatchQueue(label: "q")
+        do {
+            // spawn 64 connects, likely some will already get ECONNREFUSED as the backlog isn't large enough
```
Lukasa: Nit: think you're looking for ECONNRESET here.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,98 @@ public class ChannelTests: XCTestCase {
...
+            q.async(group: g) { () -> Void in
+                while true {
+                    do {
+                        // try to accept one client
+                        let accepted = try serverSockSyncQ.sync { () -> CInt? in
+                            try serverSock.withUnsafeFileDescriptor { fd in
+                                var pfd = pollfd(fd: fd, events: Int16(POLLIN | POLLERR | POLLHUP), revents: 0)
+                                // wait a maximum of 100ms to accept a new client, just to be guaranteed we don't get stuck here.
+                                if poll(&pfd, 1, 100) == 1 {
+                                    return accept(fd, nil, nil)
```
Lukasa: Any reason not to just block in `accept` here?

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,98 @@ public class ChannelTests: XCTestCase {
...
+                                } else {
+                                    return nil
+                                }
+                            }
+                        }
+                        if let accepted = accepted {
+                            // set SO_LINGER to 0, which will cause most TCP stacks to send an RST packet
+                            var lin = linger(l_onoff: 1, l_linger: 0)
+                            setsockopt(accepted, SOL_SOCKET, SO_LINGER, &lin, socklen_t(MemoryLayout.size(ofValue: lin)))
```
Lukasa: If `SO_LINGER` can get a `RST` packet, why do we need 64 connections?

Tests/NIOTests/ChannelTests.swift
weissi: yeah, that would hang forever so we need the 100ms timeout here

Tests/NIOTests/ChannelTests.swift
weissi: @Lukasa the `SO_LINGER` works well for Darwin, the 64 connections work well for Linux ;)

Tests/NIOTests/ChannelTests.swift
Lukasa: In what circumstance will it hang forever?

Tests/NIOTests/ChannelTests.swift
Lukasa: Ah, can you update the comments to make that clear?

Sources/NIO/BaseSocketChannel.swift
weissi: reducing to 2 as we chatted about

Tests/NIOTests/ChannelTests.swift
weissi: @Lukasa well, it calls `accept` on a socket. That would only ever return if there was someone connecting. But if there are no further connections (remember the 64 client connections each might either fail or (at least partially) succeed)

weissi:
@Lukasa feedback addressed

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -798,7 +798,27 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
...
+        if self.socket.isOpen {
+            assert(self.lifecycleManager.isRegistered)
+            let error: IOError
+            // if the socket is still registered (and therefore open), let's try to get the actual socket error from the socket
+            do {
+                let result: Int32 = try self.socket.getOption(level: SOL_SOCKET, name: SO_ERROR)
+                if result != 0 {
+                    // we have a socket error, let's forward
+                    // this path will be executed on Linux (EPOLLERR) & Darwin (ev.fflags != 0)
+                    error = IOError(errnoCode: result, reason: "connection reset (error set)")
```
Lukasa: Rather than passing `reason` can you pass `function`? That'll give us better debug strings here.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,108 @@ public class ChannelTests: XCTestCase {
...
+                        } else {
+                            break
+                        }
+                    } catch let error as IOError where error.errnoCode == EBADF {
+                        // totally fine as we just close the server socket from the main thread if we have seen
+                        // the error we're waiting for.
+                        break
+                    } catch {
+                        XCTFail("unexpected error \(error)")
+                    }
```
Lukasa: Here's my question: is there any reason we can't do this with a regular server socket channel? Create it, set the child channel options, turn off auto-read so that we only accept a maximum of one.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2200,4 +2200,108 @@ public class ChannelTests: XCTestCase {
...
+    func testConnectWithECONNREFUSEDGetsTheRightError() throws {
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let serverSock = try Socket(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+        // we deliberately don't set SO_REUSEADDR
+        XCTAssertNoThrow(try serverSock.bind(to: SocketAddress(ipAddress: "127.0.0.1", port: 0)))
+        XCTAssertNoThrow(try serverSock.withUnsafeFileDescriptor { fd in
+            try Posix.listen(descriptor: fd, backlog: 1)
```
Lukasa: Why are you calling `listen`?

Sources/NIO/BaseSocketChannel.swift
Lukasa: Just make one up. ;) The key advantage is that we get the textual description of the underlying errno, rather than your hard-coded guess.

weissi: that's not how it's intended. The reason is there so it gives you a _better_ message. I acknowledge that we had a bug here where that was worse. In case where you don't want to provide a reason as it would be `"connect failed, errno \(errno)"`, you can use `function: #function` as that's shorter.

Tests/NIOTests/ChannelTests.swift
weissi: @Lukasa I actually will scrap this whole test. It just isn't causing the `ECONNRESET` reliably enough on all the platforms and there's always room for random failure of the test (unless I don't assert that at least one actually gets `ECONNRESET`). But tbh I think pretty much everything is covered by the `ECONNREFUSED` test already...

Tests/NIOTests/ChannelTests.swift
weissi: true, actually don't need that

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2214,6 +2214,30 @@ public class ChannelTests: XCTestCase {
...
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let serverSock = try Socket(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+        // we deliberately don't set SO_REUSEADDR
+        XCTAssertNoThrow(try serverSock.bind(to: SocketAddress(ipAddress: "127.0.0.1", port: 0)))
+        let serverSockAddress = try! serverSock.localAddress()
+        XCTAssertNoThrow(try serverSock.close())
+
+        for _ in 0..<64 {
```
Lukasa: Do we need the loop? We fail in all cases other than the throw, which suggests we always throw.

weissi: @Lukasa in my tests it made the probability that we also hit the async connect path a lot higher (as in pretty much always). Therefore I thought that's a good idea :)

PR #329, opened 2018-04-18
Synchronous connection failures must close channels.

### LINKED ISSUE #322: Synchronous connect errors do not properly unregister their selector.
The Kitura test suite that runs with swift-nio 1.3.1 in 3-4 seconds, takes 133-134 seconds with 1.4.2. 

I drilled down a bit and found each `ClientBootstrap.connect().wait()` waiting longer  with 1.4.2(~0.25s) as compared to 1.3.1 (~0.01s). We're using only loopback connections.

Came across this change: https://github.com/apple/swift-nio/pull/221  Not sure if that causes the problem.

I'll try to write up a small benchmark.

Lukasa:
Motivation:

When a connect() call returns an error other than EINPROGRESS, we
currently leave the FD registered and don't close the channel. This is
wrong, we should take this as a signal to close the socket immediately,
rathern than letting the selector tell us this FD is dead: after all,
we know it's dead.

Modifications:

Call `close0()` in synchronous connect failures.

Result:

Faster failures!
Resolves #322.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -902,7 +902,6 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         }
 
         do {
-            assert(self.lifecycleManager.isRegistered)
             if try !connectSocket(to: address) {
                 // We aren't connected, we'll get the remote address later.
                 self.updateCachedAddressesFromSocket(updateLocal: true, updateRemote: false)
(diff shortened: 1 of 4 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -917,7 +917,10 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
                 becomeActive0(promise: promise)
             }
         } catch let error {
-            promise?.fail(error: error)
+            assert(self.isOpen)
```
weissi: in other places we opted to go for `assert(self.lifecycleManager.isRegistered)`. Maybe you also want to `assert(!self.lifecycleManager.isActive)` because we should be registered but not active here.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -83,6 +83,41 @@ class ChannelLifecycleHandler: ChannelInboundHandler {
     }
 }
 
+fileprivate class VerifyConnectionFailureHandler: ChannelInboundHandler {
```
weissi: would you mind putting this class at the bottom of the file? I don't really like when I need to search for the beginning of the actual test class as that is where Xcode puts the 'run all tests in this class' button. I think our test files should all start with the actual test class and not the helpers.

Sources/NIO/BaseSocketChannel.swift
Lukasa: That assertion is unnecessary because the `do` block covers it. I suppose I could assert `isRegistered && !isActive`, if you prefer.

Lukasa:
Ok @weissi, feedback addressed.

Lukasa:
ACK, feedback addressed.

PR #330, opened 2018-04-18
Ensure correct ordering of promise notification when connect is still…

normanmaurer:
… pending.

Motivation:

When a pending connect is still in process and the Channel is closed we need to ensure we use the correct ordering when notify the promises and ChannelHandlers in the pipeline.

The ordering should be:

- Notify pending connect promise
- Notify close promise
- Call channelInactive(...)

Modifications:

- Correct ordering of notification
- Add unit test

Result:

Ensure correct ordering when connect is still pending

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -677,6 +677,11 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         // Transition our internal state.
         let callouts = self.lifecycleManager.close(promise: p)
 
+        if let connectPromise = self.pendingConnect {
+            self.pendingConnect = nil
+            connectPromise.fail(error: error)
+        }
+
         // Now that our state is sensible, we can call out to user code.
         self.cancelWritesOnClose(error: error)
         callouts(self.pipeline)
@@ -686,11 +691,6 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
             self.pipeline.removeHandlers()
 
             self.closePromise.succeed(result: ())
-        }
-
-        if let connectPromise = pendingConnect {
-            pendingConnect = nil
-            connectPromise.fail(error: error)
         }
     }
 
--- a/Tests/NIOTests/SocketChannelTest.swift
+++ b/Tests/NIOTests/SocketChannelTest.swift
@@ -350,4 +350,100 @@ public class SocketChannelTest : XCTestCase {
 
         try serverChannel.close().wait()
     }
+
+    public func testPendingConnectNotificationOrder() throws {
+
+        class NotificationOrderHandler: ChannelDuplexHandler {
+            typealias InboundIn = Never
+            typealias OutboundIn = Never
+
+            private var connectPromise: EventLoopPromise<Void>?
+
+            public func channelInactive(ctx: ChannelHandlerContext) {
+                if let connectPromise = self.connectPromise {
+                    XCTAssertTrue(connectPromise.futureResult.isFulfilled)
+                } else {
+                    XCTFail("connect(...) not called before")
+                }
+            }
+
+            public func connect(ctx: ChannelHandlerContext, to address: SocketAddress, promise: EventLoopPromise<Void>?) {
+                XCTAssertNil(self.connectPromise)
+                self.connectPromise = promise
+                ctx.connect(to: address, promise: promise)
+            }
+
+            func handlerAdded(ctx: ChannelHandlerContext) {
+                XCTAssertNil(self.connectPromise)
+            }
+
+            func handlerRemoved(ctx: ChannelHandlerContext) {
+                if let connectPromise = self.connectPromise {
+                    XCTAssertTrue(connectPromise.futureResult.isFulfilled)
+                } else {
+                    XCTFail("connect(...) not called before")
+                }
+            }
+        }
+
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
+        defer { XCTAssertNoThrow(try group.syncShutdownGracefully()) }
+
+        let serverChannel = try ServerBootstrap(group: group).bind(host: "127.0.0.1", port: 0).wait()
+        defer { XCTAssertNoThrow(try serverChannel.close().wait()) }
+
+        let eventLoop = group.next()
+        let promise: EventLoopPromise<Void> = eventLoop.newPromise()
+
+        class ConnectPendingSocket: Socket {
+            let promise: EventLoopPromise<Void>
+            init(promise: EventLoopPromise<Void>) throws {
+                self.promise = promise
+                try super.init(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+            }
+
+            override func connect(to address: SocketAddress) throws -> Bool {
+                // We want to return false here to have a pending connect.
+                _ = try super.connect(to: address)
+                self.promise.succeed(result: ())
+                return false
+            }
+        }
+
+        let channel = try SocketChannel(socket: ConnectPendingSocket(promise: promise), parent: nil, eventLoop: eventLoop as! SelectableEventLoop)
+        let connectPromise: EventLoopPromise<Void> = channel.eventLoop.newPromise()
+        let closePromise: EventLoopPromise<Void> = channel.eventLoop.newPromise()
+
+        closePromise.futureResult.whenComplete {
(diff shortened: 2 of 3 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -677,6 +677,11 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         // Transition our internal state.
         let callouts = self.lifecycleManager.close(promise: p)
 
+        if let connectPromise = pendingConnect {
```
Lukasa: Now that we're touching this, can we add the missing `self` clauses?

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -297,4 +297,81 @@ public class SocketChannelTest : XCTestCase {
...
+                try super.init(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
+            }
+
+            override func connect(to address: SocketAddress) throws -> Bool {
+                return false
+            }
+        }
+
+        class NotificationOrderHandler: ChannelDuplexHandler {
+            typealias InboundIn = Any
```
Lukasa: Better to make these `Never`?

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -297,4 +297,82 @@ public class SocketChannelTest : XCTestCase {
...
+                } else {
+                    XCTFail("connect(...) not called before")
+                }
+            }
+        }
+
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
+        defer { XCTAssertNoThrow(try group.syncShutdownGracefully()) }
+
+        let channel = try SocketChannel(eventLoop: group.next() as! SelectableEventLoop, protocolFamily: PF_INET)
```
Lukasa: This isn't the constructor you want: right now you're using an actual socket, not the `ConnectPendingSocket`.

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -297,4 +297,82 @@ public class SocketChannelTest : XCTestCase {
...
+
+        XCTAssertNoThrow(try channel.pipeline.add(handler: NotificationOrderHandler()).then {
+            channel.register()
+        }.wait())
+
+        let connectPromise: EventLoopPromise<Void> = channel.eventLoop.newPromise()
+        let closePromise: EventLoopPromise<Void> = channel.eventLoop.newPromise()
+
+        channel.connect(to: try! SocketAddress.init(ipAddress: "127.0.0.1", port: 9999), promise: connectPromise)
```
weissi: extra `.init`

weissi: move `try` to the beginning of the line and add `XCTAssertNoThrow(...)`

normanmaurer: @weissi I just removed the `!` which should be good enough as it will fail the test.

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -297,4 +297,90 @@ public class SocketChannelTest : XCTestCase {
...
+            override func connect(to address: SocketAddress) throws -> Bool {
+                self.promise.succeed(result: ())
+                return false
+            }
+        }
+
+        let channel = try SocketChannel(socket: ConnectPendingSocket(promise: promise), parent: nil, eventLoop: eventLoop as! SelectableEventLoop)
+
+        XCTAssertNoThrow(try channel.pipeline.add(handler: NotificationOrderHandler()).then {
+            channel.register()
```
Lukasa: You can't do this, it'll cause the channel to be insta-closed because we'll register the channel and then immediately get EPOLLHUP.

normanmaurer: Hmm what you suggest to do here to test this ?

normanmaurer: Also it works me be preferable to make this consistent on Mac and Linux as on Mac it works

Lukasa: We cannot make this consistent on Mac and Linux, really, because to do it requires enormously annoying insight into the socket state. Specifically, it would require that the kqueue selector can check for all FDs registered with it whether they're connected or not before it enters the selector, which is obviously untenable.

The other solution is to have the epoll selector refuse to actually register anything until something registers for either readable or writable. That's more tenable, but the code sucks.

normanmaurer: @Lukasa hmm ok... so any idea for writing a test that works on mac and linux ?

normanmaurer: I think the fix is correct and the test shows it on Mac but would love to have a test that works on both ;)

normanmaurer:
@weissi @Lukasa I think this should work now... thanks again for the help :)

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -301,4 +301,103 @@ public class SocketChannelTest : XCTestCase {
...
+        XCTAssertNoThrow(try channel.eventLoop.submit {
+            channel.register().then { () -> EventLoopFuture<()> in
+                channel.connect(to: serverChannel.localAddress!, promise: connectPromise)
+                return channel.eventLoop.newSucceededFuture(result: ())
+            }.then { () -> EventLoopFuture<()> in
+                XCTAssertFalse(connectPromise.futureResult.isFulfilled)
+                // The close needs to happen in the then { ... } block to ensure we close the channel
+                // before we have the chance to register it for .write.
+                channel.close(promise: closePromise)
+                return channel.eventLoop.newSucceededFuture(result: ())
```
Lukasa: Why are you using `then` and returning succeeded futures? Just use `map` with a void return type. :smile:

normanmaurer: 🤦‍♂️

normanmaurer:
@Lukasa addressed and rebased,

normanmaurer:
@Lukasa huh ? https://github.com/apple/swift-nio/pull/330/files#diff-570b896bf499f64d3df962dd1026c797R432 ? I changed the `then` to `map`

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -350,4 +350,103 @@ public class SocketChannelTest : XCTestCase {
...
+        connectPromise.futureResult.whenComplete {
+            XCTAssertFalse(closePromise.futureResult.isFulfilled)
+        }
+
+        XCTAssertNoThrow(try channel.pipeline.add(handler: NotificationOrderHandler()).wait())
+
+        // We need to call submit {...} here to ensure then {...} is called while on the EventLoop already to not have
+        // a ECONNRESET sneak in.
+        XCTAssertNoThrow(try channel.eventLoop.submit {
+            channel.register().then { () -> EventLoopFuture<()> in
```
Lukasa: There's a then right here.

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -350,4 +350,103 @@ public class SocketChannelTest : XCTestCase {
...
+                channel.connect(to: serverChannel.localAddress!, promise: connectPromise)
+                return channel.eventLoop.newSucceededFuture(result: ())
+            }.map { () -> () in
```
Lukasa: Also our style is to spell the second parenthesis pair here as `Void` (that is, `() -> Void`), and if you've done that you don't need the explicit `return` statement.

Lukasa:
LGTM, we can wait for @weissi.

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -350,4 +350,100 @@ public class SocketChannelTest : XCTestCase {
...
+        XCTAssertNoThrow(try channel.eventLoop.submit {
+            channel.register().map { () -> Void in
+                channel.connect(to: serverChannel.localAddress!, promise: connectPromise)
+            }.map { () -> Void in
+                XCTAssertFalse(connectPromise.futureResult.isFulfilled)
+                // The close needs to happen in the then { ... } block to ensure we close the channel
+                // before we have the chance to register it for .write.
+                channel.close(promise: closePromise)
+            }
+        }.wait())
```
weissi: this needs an extra `.wait() as Void` (see #332)

normanmaurer: @weissi fixed please check again

PR #331, opened 2018-04-18
testConnect() declared a custom Socket but never used it.

normanmaurer:
Motivation:

testConnect() should use its custom Socket implementation.

Modifications:

Make the test use the custom Socket

Result:

Correct test

### DIFF
--- a/Tests/NIOTests/SocketChannelTest.swift
+++ b/Tests/NIOTests/SocketChannelTest.swift
@@ -195,36 +195,41 @@ public class SocketChannelTest : XCTestCase {
             }
         }
 
-        class ConnectSocket: Socket {
-            init() throws {
-                try super.init(protocolFamily: PF_INET, type: Posix.SOCK_STREAM)
-            }
-
-            override func connect(to address: SocketAddress) throws -> Bool {
-                return true
-            }
-        }
-
         let group = MultiThreadedEventLoopGroup(numThreads: 1)
         defer {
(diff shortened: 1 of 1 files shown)


Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -195,36 +195,41 @@ public class SocketChannelTest : XCTestCase {
...
+        let connectPromise: EventLoopPromise<Void> = eventLoop.newPromise()
+
+        let channel = try SocketChannel(socket: ConnectSocket(promise: connectPromise), parent: nil, eventLoop: eventLoop as! SelectableEventLoop)
         let promise: EventLoopPromise<Void> = channel.eventLoop.newPromise()
 
         XCTAssertNoThrow(try channel.pipeline.add(handler: ActiveVerificationHandler(promise)).then {
             channel.register()
         }.then {
-            channel.connect(to: serverChannel.localAddress!)
+            channel.connect(to: try! SocketAddress(ipAddress: "127.0.0.1", port: 9999))
```
weissi: `XCTAssertNoThrow`

weissi: ignore please

normanmaurer:
I think the test failure is actually related to the same problem as @Lukasa stated here: https://github.com/apple/swift-nio/pull/330#discussion_r182429701 (as we also do register first and then call connect)

PR #335, opened 2018-04-18
Refactor http response status equatable

pedantix:
Refactored `extension HTTPResponseStatus: Equatable`, no change to the efficacy of the code.

### Motivation:

Code readability is essential to a long-lasting maintainable project, the code this PR changes was "readable" however this PR will enhance the readability by removing excess lines of code.

### Modifications:

Changed a long case statement, to a much shorter one.

### Result:

There will be fewer lines of code, no change to the efficacy of product should occur. I am not familiar enough with the compiler to know if this will have a deleterious effect on performance I don't think it should be based on the way `public var code: UInt { get }` is implemented.


pedantix:
@helje5 pulling this convo in from #327. 
```
so custom(418) is equal to .teapot, but custom(418,”earl gray”) differs to custom(418,”whatever”) 🤔
```

are you stating that custom anything lhs/rhs should be cased out, I can see the motivation for that. It is still much shorter then long case statement. I will make change.

normanmaurer:
I think what we have here is mimic the same as we had before.  If we want to change this another PR should be done IMHO.

normanmaurer:

helje5:
@pedantix IMO using a closed enum in a open API is deeply flawed in the first place, as demonstrated by the teapot (apart from the SemVer major source incompatibility you introduce by adding a case, the ugliness of custom, and all the bugs revolving around the .custom pattern).
Wrt the specific question: yes, the reason *must* be ignored. Why? Because it is already is stripped by the parser! I.e. you don't actually get the original reason for "builtin codes" and hence can't know whether it matches some other reason you got.

BasThomas:
What could be done is to group all the `case`s that `return true` to one, and have only one line that returns true, without resorting to a `default` case.

pedantix:
@Lukasa does the code in my PR not keep the current behavior? As I read it, there is an ever-growing number of cases that pattern matches on if custom compares stored values, if not compare match on equal symbols(which is the source of the growth) if not default to false. I may be missing a swift subtlety but my PR is not aimed at changing functionality just reducing growth. There is definitely a weakness with using a default case.  As @BasThomas says all the true pairs could be reduced to a single case, I am not sure if this is any better in terms of readability.

Lukasa:
Sorry, you're right. :)

weissi:

normanmaurer:
@pedantix thanks!

pedantix:
@normanmaurer @weissi @Lukasa  thank you! -119 lines of code as a net gain I think is just cool. This project is so great to make use of thanks!!!

PR #336, opened 2018-04-19
Use XCTAssertNoThrow in tests added by 338cfb541b98636264a21989ef2dea…

normanmaurer:
…6032151dc4

Motivation:

We should use XCTAssertNoThrow whenever possible.

Modifications:

Make use of XCTAssertNoThrow in tests.

Result:

More consistent and correct tests.


Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -322,16 +322,15 @@ public class SocketChannelTest : XCTestCase {
         let clientSock = try Socket(protocolFamily: AF_INET, type: Posix.SOCK_STREAM)
         let connected = try clientSock.connect(to: serverChannel.localAddress!)
         XCTAssertEqual(connected, true)
-        let clientChannelFuture = try clientSock.withUnsafeFileDescriptor {
+        let clientChannel = try clientSock.withUnsafeFileDescriptor {
             ClientBootstrap(group: group).withConnectedSocket(descriptor: dup($0))
-        }
+        }.wait()
         try clientSock.close()
-        let clientChannel = try clientChannelFuture.wait()
```
Lukasa: This fairly substantially changes the semantic of this test, why are we doing it?

Tests/NIOTests/SocketChannelTest.swift
```diff
@@ -340,14 +339,11 @@ public class SocketChannelTest : XCTestCase {
...
         try serverSock.bind(to: SocketAddress(ipAddress: "127.0.0.1", port: 0))
-        let serverChannelFuture = try serverSock.withUnsafeFileDescriptor {
+        let serverChannel = try serverSock.withUnsafeFileDescriptor {
             DatagramBootstrap(group: group).withBoundSocket(descriptor: dup($0))
-        }
-        try serverSock.close()
-        let serverChannel = try serverChannelFuture.wait()
-
+        }.wait()
+        XCTAssertNoThrow(try serverSock.close())
```
Lukasa: Same here, why are we changing the semantic of this test?

Tests/NIOTests/SocketChannelTest.swift
normanmaurer: @Lukasa I can revert it if you want but for the test it not matters really...

normanmaurer: I just thought it makes sense to remove local vars that are not really needed.

Lukasa: It changes the behaviour quite dramatically. I'm not sure the behavioural change is worse, it may in fact be better (as right now the `close` races with the `withConnectedSocket` call). But it does change the behaviour.

normanmaurer: yeah... maybe we should loop in @vlm ... WDYT @vlm  ?

weissi: it does change behaviour so I'm not sure about this one but if @vlm is happy, I'm too

weissi:
ping @normanmaurer

weissi:
ping @normanmaurer

weissi:
ping @normanmaurer again

weissi:
closing this, please re-open if you want

PR #337, opened 2018-04-19
tests: wait fully in tests that use EventLoop.submit

weissi:
Motivation:

It's quite easy to misuse `eventLoop.submit` as if you do some
asynchronous operation (will return `EventLoopFuture<T>`) the result of
the submit will be `EventLoopFuture<EventLoopFuture<T>>`. So if you just
do

    try eventLoop.submit {
        return somethingAsynchronous
    }.wait()

then you will have only waited for the asynchronous operation to have
started (it might still be running). Fortunately, the compiler usually
warns you about that unless it returns `Void`. So it naturally tells you
to write

    try eventLoop.submit {
        return somethingAsynchronous
    }.wait().wait()

which will wait for both. Unfortunately, `XCTAssertNoThrow` will just
eat away any type so there it's advisable to write

    XCTAssertNoThrow(try eventLoop.submit { ... }.wait.wait() as Void)

to be sure it's actually a void. The Swift compiler however has an
annoying bug where it won't always realise that that's a problem
(https://bugs.swift.org/browse/SR-7478). Fortunately the 4.1 release
behaves okay so CI will catch it.

Modifications:

- replace the bad pattern
    _ = try el.submit { ... }
- added `as Void`s inside `XCTAssertNoThrow`s that do submit
- added missing `.wait()`s

Result:

tests should be more correct.


normanmaurer:

normanmaurer:

tomerd:

weissi:
@normanmaurer / @Lukasa please re-review, this test wan't actually testing what it was supposed to test because it didn't fully wait.

weissi:
```
Fatal error: Posix wrapper adds more than 20% overhead (with wrapper: 0.832322955131531, without: 0.686237931251526
```

I didn't change the syscall wrappers. Very much looks like some cohabited process used too much CPU, re-running the tests.

weissi:

weissi:

weissi:
(timed out)

tomerd:

Lukasa:
Yup, @tomerd, same issue here: seems like our CI doesn't support parallel runs. :(

tomerd:

PR #341, opened 2018-04-19
don't deliver events for unregistered fds

weissi:
Motivation:

Since forever we had a major bug in the Selector: In this condition:

- kqueue/epoll had many events
- in one of the earlier events we unregister a Channel whose fd is on of
  the later events
- we subsequently (still in the same event loop tick) register a new
  channel which gets the same fd as the previously closed one

then we would deliver an event that was meant for a previous channel to
a newly opened one.

Thanks to @mcdappdev for hitting this bug, helping us debug it and also
providing a repeatedly working repro.

Modifications:

if during event delivery any fd gets unregistered, we stop delivering
the remaining events and rely on the selector to redeliver them
again next time.

Result:

we don't deliver events for previously closed channels to new ones.


Sources/NIO/Selector.swift
```diff
@@ -540,7 +543,8 @@ final class Selector<R: Registration> {
             Int(try KQueue.kevent(kq: self.fd, changelist: nil, nchanges: 0, eventlist: events, nevents: Int32(eventsCapacity), timeout: ts))
         }
 
-        for i in 0..<ready {
+        self.deregistrationsHappened = false
+        for i in 0..<ready where !self.deregistrationsHappened {
```
normanmaurer: as noted offline I think while this should work we should treat this as workaround as this depends highly on the "level-triggered" semantics of kqueue / epoll.

weissi: Totally agreed. Let’s get this in (I actually found another bug (and a missing assertion) whilst fixing it) and then fix properly. I agree we shouldn’t rely on it being level triggered here.

normanmaurer: Works for me

Sources/NIO/Selector.swift
```diff
@@ -259,6 +259,7 @@ final class Selector<R: Registration> {
     private var eventsCapacity = 64
     private var events: UnsafeMutablePointer<EventType>
     private var registrations = [Int: R]()
+    private var deregistrationsHappened: Bool = false
```
Lukasa: Can you put a doc comment on this guy indicating that it's a temporary fix?

Sources/NIO/Selector.swift
```diff
@@ -500,7 +502,8 @@ final class Selector<R: Registration> {
             ready = Int(try Epoll.epoll_wait(epfd: self.fd, events: events, maxevents: Int32(eventsCapacity), timeout: -1))
         }
 
-        for i in 0..<ready {
+        self.deregistrationsHappened = false
```
Lukasa: I actually don't think we need to do this for epoll. I realised this this morning, but epoll generally only returns each FD once, with a mask of all event flags for that FD.

This means that our epoll code is immune to this bug: the channel is concretely held to dispatch all the events to it. This also suggests a sensible long-term fix: we should coalesce the kqueue notifications so that each channel has events dispatched to it only once.

normanmaurer: That’s actually a very good point 👍

weissi: that's not true because (just like the test does), if you close lots of fds and then open lots of fds, there might very well be events for file descriptors that are no longer valid.

Lukasa: Yeah, you're right.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -776,6 +781,9 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
     }
 
     final func readEOF() {
+        print("readEOF \(self)")
```
Lukasa: Print statement left behind.

Sources/NIO/Selector.swift
Lukasa: ^

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                    ctx.close(promise: nil)
+                }
+            }
+
+            func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+                // we expect these channels to be get data only a while after the re-connect event loop tick as it's
+                // impossible to get a read notification in the very same event loop tick that you got registered
+                XCTAssertTrue(self.hasReConnectEventLoopTickFinished.value)
+
+                // TODO: check the bytebuffer
```
Lukasa: TODO

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                XCTAssertTrue(self.hasReConnectEventLoopTickFinished.value)
+                XCTAssertTrue(self.didRead)
+                if !self.didRead {
+                    self.didReadPromise.fail(error: DidNotReadError.didNotReadGotInactive)
+                    ctx.close(promise: nil)
+                }
+            }
+
+            func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+                // we expect these channels to be get data only a while after the re-connect event loop tick as it's
```
Lukasa: s/to be get/to get/

Sources/NIO/Selector.swift
normanmaurer: doh... true :(

Sources/NIO/BaseSocketChannel.swift
normanmaurer: @weissi remove

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                    precondition(halfClosureAllowed,
+                                 "the test configuration is bogus: half-closure is dis-allowed which breaks the setup of this test")
+                }
+                self.allChannels.value.append(ctx.channel)
+            }
+
+            func userInboundEventTriggered(ctx: ChannelHandlerContext, event: Any) {
+                // this is the `.readEOF` that is triggered by the `ServerHandler`'s `close` calls because our channel
+                // supports half-closure
+                if self.allChannels.value.count == SelectorTest.testWeDoNotDeliverEventsForPreviouslyClosedChannels_numberOfChannelsToUse {
```
Lukasa: This would be cleaner as `guard`.

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
         }
         XCTAssertEqual(1, readyCount)
     }
-}
 
+    private static let testWeDoNotDeliverEventsForPreviouslyClosedChannels_numberOfChannelsToUse = 10
+    func testWeDoNotDeliverEventsForPreviouslyClosedChannels() {
+        /// We use this class to box mutable values, generally in this test anything boxed should only be read/written
+        /// on the event loop `el`.
```
normanmaurer: @weissi should we use a computed property and check `MultiThreadedEventLoopThread.currentEventLoop` returns what we expect ?

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                self.value = value
+            }
+            var value: T
+        }
+        enum DidNotReadError: Error {
+            case didNotReadGotInactive
+            case didNotReadGotReadComplete
+        }
+
+        /// This handler is inserted in the channels that are re-connected. So we're closing a bunch of channels and
```
normanmaurer: ... is inserted in the `ChannelPipeline` of the channels that are re-connected

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                                               The good news is: Not everything is lost, this test should also work
+                                               if you instead open a regular file (in O_RDWR) and just use this file's
+                                               fd with `ClientBootstrap(group: group).withConnectedSocket(fileFD)`.
+                                               Sure, a file is not a socket but it's always readable and writable and
+                                               that fulfills the requirements we have here.
+                                               I still hope this change will never have to be done.
+                                               Note: if you changed anything about the pipeline's handler adding/removal
+                                               you might also have a bug there.
+                                               """)
+                                return ()
```
Lukasa: Shouldn't need the return statement.

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+        /// will end up using the _same_ fds and this test ensures that we're not getting the outdated events. In this
+        /// case the outdated events are all `.readEOF`s which manifest as `channelReadComplete`s. If we're delivering
+        /// outdated events, they will also happen in the _same event loop tick_ and therefore we do quite a few
+        /// assertions that we're either in or not in that interesting event loop tick.
+        class HappyWhenReadHandler: ChannelInboundHandler {
+            typealias InboundIn = ByteBuffer
+
+            private let didReadPromise: EventLoopPromise<Void>
+            private var didRead: Bool = false
+            private let hasReConnectEventLoopTickFinished: Box<Bool>
```
normanmaurer: can we group the `let` together ?

Tests/NIOTests/SelectorTest.swift
```diff
@@ -96,5 +97,257 @@ class SelectorTest: XCTestCase {
...
+                    XCTAssertEqual(SelectorTest.testWeDoNotDeliverEventsForPreviouslyClosedChannels_numberOfChannelsToUse, numberOfConnectedChannels.value)
+                    self.allServerChannels.value.forEach { c in
+                        c.close(promise: nil)
+                    }
+                }
+            }
+        }
+        let el = MultiThreadedEventLoopGroup(numThreads: 1).next()
+        defer {
+            try! el.syncShutdownGracefully()
```
normanmaurer: XCTAssertNoThrow(...) ?

Tests/NIOTests/TestUtils.swift
```diff
@@ -56,6 +56,21 @@ func withTemporaryFile<T>(content: String? = nil, _ body: (NIO.FileHandle, Strin
...
+    var templateBytes = template.utf8 + [0]
+    let templateBytesCount = templateBytes.count
+    templateBytes.withUnsafeMutableBufferPointer { ptr in
+        ptr.baseAddress!.withMemoryRebound(to: Int8.self, capacity: templateBytesCount) { (ptr: UnsafeMutablePointer<Int8>) in
+            let ret = mkdtemp(ptr)
+            XCTAssertNotNil(ret)
+        }
+    }
+    templateBytes.removeLast()
+    let udsTempDir = String(decoding: templateBytes, as: UTF8.self)
```
normanmaurer: nit: just return.. not net to create the let first.

Tests/NIOTests/SelectorTest.swift
weissi: why not

Tests/NIOTests/SelectorTest.swift
weissi: sure why not

weissi:
@normanmaurer / @Lukasa all addressed

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -619,18 +624,23 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         }
     }
 
-    private func registerForReadable() {
+    private final func registerForReadable() {
         assert(eventLoop.inEventLoop)
         assert(self.lifecycleManager.isRegistered)
 
+        guard !self.lifecycleManager.hasSeenEOFNotification else {
```
Lukasa: Should this be a `guard` or an `assert`?

Tests/NIOTests/EchoServerClientTest.swift
```diff
@@ -779,4 +779,55 @@ class EchoServerClientTest : XCTestCase {
...
+        let countingHandler = ByteCountingHandler(numBytes: numBytes, promise: promise)
+
+        // we're binding to IPv4 only
+        let serverChannel = try ServerBootstrap(group: group)
+            .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+            .childChannelInitializer { channel in
+                channel.pipeline.add(handler: countingHandler)
+            }
+            .bind(host: "127.0.0.1", port: 0)
+            .thenIfError {
```
Lukasa: Surely `bind` isn't going to error.

weissi: indeed

PR #343, opened 2018-04-20
don't crash if non-socket is provided to withConnectedSocket

weissi:
CC @vlm I can't add you as a reviewer for some reason but it'd be great if you could have a look.

Motivation:

If the registration of a file descriptor passed to `withConnectedSocket`
failed, we previously crashed in an assertion as we couldn't activate
the channel.

Modifications:

Check if registration succeeds or fails and only activate if
registration succeeded.

Result:

Less crashes for users that try odd stuff.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -732,8 +732,15 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         assert(self.eventLoop.inEventLoop)
         assert(self.isOpen)
         assert(!self.lifecycleManager.isActive)
-        register0(promise: nil)
-        becomeActive0(promise: promise)
+        let registerPromise: EventLoopPromise<Void> = self.eventLoop.newPromise()
+        if let promise = promise {
+            registerPromise.futureResult.cascadeFailure(promise: promise)
+        }
+        register0(promise: registerPromise)
+        if self.lifecycleManager.isRegistered {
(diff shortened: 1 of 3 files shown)


Lukasa:
@normanmaurer The problem with that is that this is likely to be a kqueue-specific issue: I bet `epoll` will just tolerate this behaviour, as there is no point at which you assert to `epoll` "I believe this FD is a socket".

Does anyone know off the top of their head if there's a syscall we can make to detect whether an FD is a socket? If there is, I agree with Norman that we'd probably prefer to crash hard here, but if there isn't then I think @weissi is right and we probably need to tolerate this.

normanmaurer:
@Lukasa I think it should be possible with `stat` and check via `S_IFSOCK`:

https://linux.die.net/man/2/stat

/cc @weissi

Lukasa:
Nah, we can do better than that and get a nice side benefit along the way.

We should instead call `getsockopt(SOL_SOCKET, SO_ERROR)`. This will fail with `ENOTSOCK` for anything that isn't a socket, and will also conveniently tell us if the socket we're about to create a channel for is already dead, saving us a lot of setup cost. My suggestion would be that the `with{Connected,Bound}Socket` methods should call this method on the provided FD and fail fast if it fails, or if it returns an error.

normanmaurer:
@Lukasa sounds like a plan

normanmaurer:
@weissi @Lukasa so what we want to do with this one ?

weissi:
closing this

PR #346, opened 2018-04-23
Ensure localAddress / remoteAddress are still accessible in channelIn…

normanmaurer:
…active / handlerRemoved

Motivation:

Often its useful to be still be able to access the local / remote address during channelInactive / handlerRemoved callbacks to for example log it. We should ensure its still accessible during it.

Modifications:

- Fallback to slow-path in ChannelHandlerContext.localAddress0 / remoteAddress0 if fast-path fails to try accessing the address via the cache.
- Clear cached addresses after all callbacks are run.
- Add unit test.

Result:

Be able to access addresses while handlers are notified.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -1017,11 +1017,29 @@ public final class ChannelHandlerContext: ChannelInvoker {
     }
 
     public var remoteAddress: SocketAddress? {
-        return try? self.channel._unsafe.remoteAddress0()
+        do {
+            // Fast-path access to the remoteAddress.
+            return try self.channel._unsafe.remoteAddress0()
+        } catch ChannelError.ioOnClosedChannel {
+            // Channel was closed already but we may still have the address cached so try to access it via the Channel
+            // so we are able to use it in channelInactive(...) / handlerRemoved(...) methods.
+            return self.channel.remoteAddress
+        } catch {
+            return nil
+        }
     }
 
     public var localAddress: SocketAddress? {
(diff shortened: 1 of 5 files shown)


normanmaurer:
@ldewailly PTAL as well

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1017,11 +1017,29 @@ public final class ChannelHandlerContext: ChannelInvoker {
     }
 
     public var remoteAddress: SocketAddress? {
-        return try? self.channel._unsafe.remoteAddress0()
+        do {
+            // Fast-path access to the remoteAddress.
+            return try self.channel._unsafe.remoteAddress0()
+        } catch let err as ChannelError where err == .ioOnClosedChannel {
```
Lukasa: No need for this verbose syntax, just do `catch ChannelError.ioOnClosedChannel`.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1017,11 +1017,29 @@ public final class ChannelHandlerContext: ChannelInvoker {
...
+            return nil
+        }
     }
 
     public var localAddress: SocketAddress? {
-        return try? self.channel._unsafe.localAddress0()
+        do {
+            // Fast-path access to the localAddress.
+            return try self.channel._unsafe.localAddress0()
+        } catch let err as ChannelError where err == .ioOnClosedChannel {
```
Lukasa: Same here.

normanmaurer:
@Lukasa addressed

Issue #279, opened 2018-04-04
Enhance CircularBuffer with insertFirst and popFirst/Last


helje5:
I'd like to abuse the `CircularBuffer` as a CoW data structure w/ linked list behaviour (better index-0 insert/remove behaviour than `ContiguousArray`, which I currently use). Since the ivars are all private, I can't extend `CircularBuffer` with the operations I need. I think it would be those (also available in the stdlib, excluding insertFirst):
```swift
public mutating func popFirst() -> Element?
public mutating func popLast() -> Element?
public mutating func insertFirst(Element)
```
Would be great to get those (and a CoW linked list ;-) )

weissi:
sure, we can totally just implement the missing methods. @Lukasa / @normanmaurer opinions?

Lukasa:
Eh, sure, why not?

normanmaurer:
If we can without any perf hit for our use-case I am in

normanmaurer:
@helje5 btw we love PRs :)

helje5:
Every one loves PRs! But then @Lukasa wants you to write tests for perfectly working code and a 2 minute patch becomes hours and hours. Been there, done that ;-)

But yes, I consider doing a PR. I need to fully understand it to do the `insertFirst` right, the only reason I didn't approach that yet (apart from the fact that it would probably take @weissi 2 minutes to do those 😉)

Lukasa:
I'm a monster, I know. And yet I feel so little guilt!

Lukasa:
(Also I'd definitely make @weissi write the tests too.)

normanmaurer:
@Lukasa maybe I should write the patch then as it seems I will not need to write tests 🤷‍♂️  ;)

normanmaurer:
I will take a stab on this once https://github.com/apple/swift-nio/pull/313 is merged

PR #347, opened 2018-04-23
Add last / addFirst to CircularBuffer

normanmaurer:
Motivation:

Its often useful to be able to add / retrieve elements on both sides of a CircularBuffer.

Modifications:

- Add last / addFirst to be able to act on both ends.
- Add tests

Result:

Fixes https://github.com/apple/swift-nio/issues/279


normanmaurer:
FYI @helje5

helje5:
Very nice! 👯 Should that also do the power-of-2 buffer size increases instead of doubling? For Redi/S to be fast, I had to do manual capacity increases.

helje5:
Oh, I just noticed you already did that in #313 .

Sources/NIO/CircularBuffer.swift
```diff
@@ -89,11 +116,31 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     ///
     /// *O(1)*
     public var first: E? {
-        if self.isEmpty {
-            return nil
-        } else {
-            return self.buffer[self.headIdx]
-        }
+        // As self.buffer is filled with `nil` this will correctly return `nil` if the CircularBuffer is empty.
```
Lukasa: Is this change worth it? The path of checking `isEmpty` is going to be faster, because it saves us the additional level of indirection into the buffer.

normanmaurer: @Lukasa not sure I think it is a nice change but I am not strong on it...

Lukasa: @weissi for the tie-break!

weissi: feels to clever to me, I'd leave the old code

normanmaurer: @weissi achievement unlocked... @weissi called me clever =P 

Let me revert this part

normanmaurer:
@weissi @Lukasa PTAL again

Sources/NIO/CircularBuffer.swift
```diff
@@ -50,36 +50,63 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     ///
     /// Amortized *O(1)*
     public mutating func append(_ value: E) {
+        return self.addLast(value)
+    }
+
+    /// Add an element to the end of the ring buffer.
+    ///
+    /// Amortized *O(1)*
+    public mutating func addLast(_ value: E) {
```
weissi: why do we have this method? We have `append` already which matches `Array`'s `append`.

Sources/NIO/CircularBuffer.swift
```diff
@@ -50,36 +50,63 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
...
-            self.headIdx = 0
-            self.buffer = newBacking
+            self.doubleCapacity()
+        }
+    }
+
+    /// Add an element to the front of the ring buffer.
+    ///
+    /// Amortized *O(1)*
+    public mutating func addFirst(_ value: E) {
```
weissi: Swift doesn't really seem to use `add...`. I would suggest `prepend` maybe as it's the opposite of `append`? CC @lorentey, @moiseev, @airspeedswift

Sources/NIO/CircularBuffer.swift
```diff
@@ -50,36 +50,63 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
...
     /// Remove the front element of the ring buffer.
     ///
     /// *O(1)*
     public mutating func removeFirst() -> E {
-        precondition(!self.isEmpty)
-
         let value = self.buffer[self.headIdx]
-        self.buffer[headIdx] = nil
+
+        precondition(value != nil, "CircularBuffer is empty")
```
weissi: this should be the first instruction I think, just check `isEmpty`

Sources/NIO/CircularBuffer.swift
```diff
@@ -96,6 +123,32 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
...
     }
 
+    /// Remove the last element of the ring buffer.
+    ///
+    /// *O(1)*
+    public mutating func removeLast() -> E {
+        let idx = (self.tailIdx - 1) & self.mask
+        let value = self.buffer[idx]
+
+        precondition(value != nil, "CircularBuffer is empty")
```
weissi: should be first

Sources/NIO/CircularBuffer.swift
weissi: also: what protocols should this auto expanding ring buffer implement?

Lukasa: Appending to the front gets us to `BidirectionalCollection`. Now that we're at that, we can actually hop the next level up to `RandomAccessCollection`. Given the way this behaves, we could also give it `RangeReplaceableCollection` too, which is a pretty easy cousin of `RandomAccessCollection`.

So I'd recommend adding conformances to those two, though we may want to do them in a different PR.

weissi: @Lukasa isn't `RangeReplaceableCollection` a thing where you can replace any range by any other range? So for example put 5 elements in a space where there used to be 3 ?

Like `thing[1...3] = ["foo", "bar", "buz", "qux", "quux"]` ?

Lukasa: Yeah, and our buffer can do that. It's no less efficient at it than a regular array.

weissi: sure but we have not implementation for that just yet so I'd do that whenever...

Lukasa: Oh agreed, I think both of the new conformances should come in separate PRs to reduce scope. :smile:

Sources/NIO/CircularBuffer.swift
normanmaurer: @weissi I added it for consistent naming ;)

Sources/NIO/CircularBuffer.swift
normanmaurer: @weissi sure I can do this but I would argue that this way we it may be a bit faster...

Sources/NIO/CircularBuffer.swift
normanmaurer: See https://github.com/apple/swift-nio/pull/347#discussion_r183466837

Sources/NIO/CircularBuffer.swift
normanmaurer: @weissi `prepend` sounds good to me as well... Waiting for @lorentey, @moiseev, @airspeedswift tho

Sources/NIO/CircularBuffer.swift
Lukasa: If you want a speedup, the better thing to do is to do this:

```swift
guard let value = self.buffer[headIdx] else {
    preconditionFailure("Circular buffer is empty")
}
self.buffer[headIdx] = nil
self.headIdx = (self.headIdx + 1) & self.mask
return value
```

This checks the optional only once, not twice.

normanmaurer: fine with me as well...(would use fatalError tho) @weissi  ?

Sources/NIO/CircularBuffer.swift
moiseev: I vote for `prepend`/`append`.

moiseev: Thinking more about the right standard protocols for this type, looks like it should be possible to make `CircularBuffer` into a slice type. This way (after the `BidirectionalCollection` requirements are implemented) `removeLast`, `removeLast(_:)`, and `popLast()` default implementations will become automagically available.

moiseev: Ahh... No. You can't do that, because `CircularBuffer.startIndex` is always `0`. How attached are we to zero-based indexing for the `CircularBuffer`? ;-)

helje5: I guess that (startIndex=0) wouldn't matter much. (though Redi/S lists also have that getrange-family of operations, but that could be mapped in the frontend).

The thing I'm wondering is whether there should also be a `prepend(contentsOf:)` which reserves the proper capacity.

moiseev: From what I can see, `CircularBuffer` does not have even an `append(contentsOf:)` right now.

I had this idea for a moment, but then I quickly hit a wall trying to name `prepend(contentsOf:)` correctly and specify the behavior when `tailIdx` reaches the `headIdx`, so I decided not to =)

Going back to indices. FWIW, in the collection types provided by the standard library there are no guarantees about the values of indices, so one should always use `startIndex`/`endIndex` combo and `index` family of methods, which might be annoying, but is the only guaranteed way of writing generic code over collections, even when `Index: FixedWidthInteger` or `Index == Int`.

Sources/NIO/CircularBuffer.swift
weissi: happy with either

Sources/NIO/CircularBuffer.swift
```diff
@@ -96,6 +115,31 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
         }
     }
 
+    /// Remove the last element of the ring buffer.
+    ///
+    /// *O(1)*
+    public mutating func removeLast() -> E {
+        let idx = (self.tailIdx - 1) & self.mask
+        guard let value = self.buffer[idx] else {
+            fatalError("CircularBuffer is empty")
```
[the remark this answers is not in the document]
weissi: sure, happy with either

Sources/NIO/CircularBuffer.swift
weissi: @moiseev Many thanks! I'm not personally attached to the 0 based indexing and I think we can change it as SemVer patch I think. What do @normanmaurer  & @Lukasa  think?

weissi: and obvs I'm fine with prepend

normanmaurer: @weissi sounds good to me... That said I think we can pull in the PR as it is now... WDYT ?

normanmaurer:
Created https://github.com/apple/swift-nio/issues/350 as a followup.

PR #348, opened 2018-04-23
fix recursive channelReads in pipelining handler

weissi:
Motivation:

The pipelining handler made the assumption that `channelRead` is never
called recursively. That's mostly true but there is at least one
situation where that's not true:
- pipelining handler seen a response .end and delivers a .head (which is
  done in `channelRead`)
- a handler further down stream writes and flushes some response data
- the flushes fail which leads to us draining the receive buffer
- if the receive buffer contained more requests, the pipelining
  handler's `channelRead` is called again (recursively)

The net result of that was that the new request parts from the receive
buffer would now jump the queue and go through the channel pipeline
next, before other already buffered messages.

Modifications:

made the pipelining handler buffer if a `channelRead` comes in from the
pipeline and there is already at least one message buffered.

Result:

the ordering of the incoming messages should now be respected which is
very important...

### DIFF
--- a/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
+++ b/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
@@ -45,6 +45,7 @@ private final class ReadRecorder: ChannelInboundHandler {
 
     func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
         self.reads.append(.channelRead(self.unwrapInboundIn(data)))
+        ctx.fireChannelRead(data)
     }
 
     func userInboundEventTriggered(ctx: ChannelHandlerContext, event: Any) {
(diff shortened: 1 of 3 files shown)


Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -356,4 +357,100 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+
+                case done
+            }
+            var state: State = .req1HeadExpected
+
+            func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+                let req = self.unwrapInboundIn(data)
+                switch req {
+                case .head(let head):
+                    var sendEnd = true
```
Lukasa: This doesn't look right. It may *be* right, but it's probably better to place this in the relevant switch cases.

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -356,4 +357,100 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+            XCTAssertNoThrow(try self.channel.writeInbound(HTTPServerRequestPart.end(nil)))
+        }
+
+        // now we should have delivered the first request, with the second and third buffered
+        XCTAssertEqual(.req2HeadExpected, handler.state)
+
+        // finish 1st request, that will send through the 2nd one which will then write the 'req_boom' request
+        XCTAssertNoThrow(try channel.writeAndFlush(HTTPServerResponsePart.end(nil)).wait())
+
+        XCTAssertEqual(.done, handler.state)
```
Lukasa: Is this end case right? It doesn't look right, why are we expecting to see all four requests when only one `end` has been sent through?

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
weissi: I think it is. We only delay the `.end` for `/req_1`

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
weissi: all of them are going through, only `req_1` is delayed

PR #349, opened 2018-04-23
make enums containing ByteBuffer fit in three words

weissi:
CC @tanner0101 & @helje5 

only really makes sense together with #351

Motivation:

NIO's dynamic pipeline comes with one performance caveat: If your type
isn't specialised in NIOAny (only types in the NIO module can be) and
it's over 3 words (24 bytes on 64-bit platforms), you'll suffer an
allocation for every time you box it in NIOAny. Very unfortunately, both
ByteBuffer and FileRegion were exactly 3 words wide which means that any
enum containing those would be wider than 3 words. Worse, both
HTTP{Request,Response}Head contained types that were wider than 3 words.
The best solution to this problem is to shrink ByteBuffer and FileRegion
to just under 3 words and that's exactly what this PR is doing. That is
slightly tricky as ByteBuffer was already bit packed fairly well
(reader/writer indices & slice begin/end were stored as a UInt32). The
trick we employ now is to store the slice beginning in a UInt24 and the
file region reader index in a UInt56. That saves one byte for both
ByteBuffer and FileRegion with very moderate tradeoffs: The reader index
in a file now needs to be within 64 PiB (peta bytes) and a byte buffer
slice beginning must start within 16 MiB (mega bytes). Note: The
reader/writer indices as well as slice ends are _not_ affected and can
still be within 4 GiB. Clearly no one would care about the restrictions
for FileRegions in the real world but we might hit the ByteBuffer slice
beginning limit in which case the slice would be copied out. But
given that slices are mostly used to slice off headers in network
protocols, 16 MiB should be _plenty_.

Norman was kind enough to measure the perf differences:

master (before this PR):
```
$ wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://localhost:8888
  4 threads and 256 connections
    Latency    93.02ms  158.13ms   1.97s    93.07%
    Req/Sec   112.79k    26.05k  258.75k    84.50%
  4501098 requests in 10.04s, 214.63MB read
  Socket errors: connect 0, read 111, write 0, timeout 89
Requests/sec: 448485.61
Transfer/sec:     21.39MB
```

after this PR:
```
$ wrk -c 256 -d 10s -t 4 -s ~/Downloads/pipeline-many.lua -H "X-Host: SomeValue" -H "Host: swiftnio.io" -H "ThereAreEvenMoreHeaders: AndMoreValues" http://localhost:8888
  4 threads and 256 connections
    Latency   107.53ms  206.56ms   1.99s    90.55%
    Req/Sec   124.15k    26.56k  290.41k    89.25%
  4952904 requests in 10.03s, 236.17MB read
  Socket errors: connect 0, read 161, write 0, timeout 22
Requests/sec: 493852.65
Transfer/sec:     23.55MB
```

so we see a nice 10% improvement and allocations go down by 2, too:

```
23:21:44 info: 1000_reqs_1_conn: allocations not freed: 0
23:21:44 info: 1000_reqs_1_conn: total number of mallocs: 67832
```

from

```
19:08:57 info: 1000_reqs_1_conn: allocations not freed: 0
19:08:57 info: 1000_reqs_1_conn: total number of mallocs: 69834
```

Modifications:

- shrank ByteBuffer by 1 byte, making it 23 bytes in total
- shrank FileRegion by 1 byte, making it 23 bytes in total
- CoW boxed parts of HTTP{Request,Response}Head to make them fit within 2 words
- added @_inlineable to NIOAny where appropriate to not suffer boxed existentials in different places
- added tests for the new edge cases

Result:

- more speed
- fewer allocations

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -29,6 +29,43 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
     }
 #endif
 
+extension _ByteBufferSlice: Equatable {
+    static func ==(_ lhs: _ByteBufferSlice, _ rhs: _ByteBufferSlice) -> Bool {
+        return lhs._begin == rhs._begin && lhs.upperBound == rhs.upperBound
+    }
+}
+
+/// The slice of a `ByteBuffer`, it's different from `Range<UInt32>` because the lower bound is actually only
+/// 24 bits (the upper bound is still 32). Before constructing, you need to make sure the lower bound actually
+/// fits within 24 bits, otherwise the behaviour is undefined.
+@_versioned
+struct _ByteBufferSlice {
+    @_versioned var upperBound: ByteBuffer.Index
+    @_versioned var _begin: _UInt24
+    @_versioned var lowerBound: ByteBuffer.Index {
+        return UInt32(self._begin)
+    }
+    @_inlineable @_versioned var count: Int {
+        return Int(self.upperBound - self.lowerBound)
+    }
+    init() {
+        self._begin = 0
+        self.upperBound = 0
+    }
+    static var maxSupportedLowerBound: ByteBuffer.Index {
+        return ByteBuffer.Index(_UInt24.max)
+    }
+}
+
+extension _ByteBufferSlice {
+    init(_ range: Range<UInt32>) {
+        self = _ByteBufferSlice()
+
+        self._begin = _UInt24(range.lowerBound)
+        self.upperBound = range.upperBound
+    }
+}
+
 /// The preferred allocator for `ByteBuffer` values. The allocation strategy is opaque but is currently libc's
 /// `malloc`, `realloc` and `free`.
 ///
@@ -163,15 +200,15 @@ public struct ByteBufferAllocator {
 /// for doing so. In any case, if you use the `get` prefixed methods you are responsible for ensuring that you do not reach into uninitialized memory by taking the `readableBytes` and `readerIndex` into
 /// account, and ensuring that you have previously written into the area covered by the `index itself.
 public struct ByteBuffer {
-    typealias Slice = Range<Index>
+    typealias Slice = _ByteBufferSlice
     typealias Allocator = ByteBufferAllocator
     typealias Index = UInt32
     typealias Capacity = UInt32
 
+    @_versioned private(set) var _storage: _Storage
     @_versioned private(set) var _readerIndex: Index = 0
     @_versioned private(set) var _writerIndex: Index = 0
     @_versioned private(set) var _slice: Slice
-    @_versioned private(set) var _storage: _Storage
 
     // MARK: Internal _Storage for CoW
     @_versioned final class _Storage {
@@ -184,7 +221,7 @@ public struct ByteBuffer {
             self.bytes = bytesNoCopy
             self.capacity = capacity
             self.allocator = allocator
-            self.fullSlice = 0..<self.capacity
+            self.fullSlice = _ByteBufferSlice(0..<self.capacity)
         }
 
         deinit {
@@ -210,7 +247,7 @@ public struct ByteBuffer {
                             allocator: self.allocator)
         }
 
-        public func reallocSlice(_ slice: Slice, capacity: Capacity) -> _Storage {
+        public func reallocSlice(_ slice: Range<ByteBuffer.Index>, capacity: Capacity) -> _Storage {
             assert(slice.count <= capacity)
             let new = self.allocateStorage(capacity: capacity)
             self.allocator.memcpy(new.bytes, self.bytes.advanced(by: Int(slice.lowerBound)), slice.count)
@@ -223,7 +260,7 @@ public struct ByteBuffer {
             ptr.bindMemory(to: UInt8.self, capacity: Int(capacity))
             self.bytes = ptr
             self.capacity = capacity
-            self.fullSlice = 0..<self.capacity
+            self.fullSlice = _ByteBufferSlice(0..<self.capacity)
         }
 
         private func deallocate() {
@@ -282,7 +319,7 @@ public struct ByteBuffer {
             } while newCapacity < index || newCapacity - index < capacity
 
             self._storage.reallocStorage(capacity: newCapacity)
-            self._slice = _slice.lowerBound..<_slice.lowerBound + newCapacity
+            self._slice = _ByteBufferSlice(_slice.lowerBound..<_slice.lowerBound + newCapacity)
         }
     }
 
@@ -486,8 +523,19 @@ public struct ByteBuffer {
         }
         let index = _toIndex(index)
         let length = _toCapacity(length)
+        let sliceStartIndex = self._slice.lowerBound + index
+
+        guard sliceStartIndex <= ByteBuffer.Slice.maxSupportedLowerBound else {
+            // the slice's begin is past the maximum supported slice begin value (16 MiB) so the only option we have
+            // is copy the slice into a fresh buffer. The slice begin will then be at index 0.
+            var new = self
+            new._moveWriterIndex(to: sliceStartIndex + length)
+            new._moveReaderIndex(to: sliceStartIndex)
+            new._copyStorageAndRebase(capacity: length, resetIndices: true)
+            return new
+        }
         var new = self
-        new._slice = self._slice.lowerBound + index ..< self._slice.lowerBound + index+length
+        new._slice = _ByteBufferSlice(sliceStartIndex ..< self._slice.lowerBound + index+length)
         new.moveReaderIndex(to: 0)
         new._moveWriterIndex(to: length)
         return new
--- a/Sources/NIO/FileRegion.swift
+++ b/Sources/NIO/FileRegion.swift
@@ -28,11 +28,23 @@ public struct FileRegion {
     /// The `FileHandle` that is used by this `FileRegion`.
     public let fileHandle: FileHandle
 
+    private let _endIndex: UInt64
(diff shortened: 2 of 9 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -29,6 +29,86 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
     }
 #endif
 
+struct SadUInt24: ExpressibleByIntegerLiteral {
```
normanmaurer: Haha... `Sad...` ;)

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -29,6 +29,86 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
...
+    @_versioned var b3: UInt8
+
+    init(integerLiteral value: UInt16) {
+        self.b1 = UInt8(truncatingIfNeeded: (value & 0xff      ) >> 0 )
+        self.b2 = UInt8(truncatingIfNeeded: (value & 0xff_00   ) >> 8 )
+        self.b3 = 0
+    }
+
+    init?(_ value: UInt32) {
+        guard value & 0xff_00_00_00 == 0 else {
```
normanmaurer: TIL that you can use `_` to make it more readable

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -19,11 +19,52 @@ let headerSeparator: StaticString = ": "
 
 /// A representation of the request line and header fields of a HTTP request.
 public struct HTTPRequestHead: Equatable {
+    private class _Storage {
```
normanmaurer: I guess it not really makes a difference but use `final` (for all internal classes) ?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -19,11 +19,52 @@ let headerSeparator: StaticString = ": "
...
+        var version: HTTPVersion
+
+        init(method: HTTPMethod, rawURI: URI, version: HTTPVersion) {
+            self.method = method
+            self.rawURI = rawURI
+            self.version = version
+        }
+
+        func copy() -> _Storage {
+            return _Storage(method: self.method, rawURI: self.rawURI, version: self.version)
```
normanmaurer: Can you either always use `.init` or always `_Storage` just to make it consistent ?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -486,8 +584,19 @@ public struct ByteBuffer {
         }
         let index = _toIndex(index)
         let length = _toCapacity(length)
+        let sliceBeginIndex = self._slice.lowerBound + index
+
+        guard sliceBeginIndex <= ByteBuffer.Slice.maxSupportedLowerBound else {
+            // the slice's begin is past the maximum supported slice begin value (16 MiB) so the only option we have
+            // is copy the slice into a fresh buffer and the slice begin will then be at index 0.
```
normanmaurer: nit: s/and the/. The/

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -29,6 +29,97 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
...
+        newValue |= Int(value.b3) << 16
+        self = newValue
+    }
+}
+
+extension _UInt24 {
+    init(_ value: UInt32) {
+        precondition(value & 0xff_00_00_00 == 0, "value \(value) too large for _UInt24")
+        self.b12 = UInt16(truncatingIfNeeded: (value &    0xff_ff) >> 0 )
+        self.b3  =  UInt8(truncatingIfNeeded: (value & 0xff_ff_00) >> 16)
```
Lukasa: You can save yourself a masking operation here. You already precondition that the top byte is zero, so with that knowledge in place you can just shift right by 16 bits and store into `UInt8`.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -29,6 +29,97 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
     }
 #endif
 
+@_versioned
+struct _UInt24: ExpressibleByIntegerLiteral {
```
Lukasa: Can we get a doc comment here please?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -29,6 +29,97 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
...
+}
+
+extension _ByteBufferSlice: Equatable {
+    static func ==(_ lhs: _ByteBufferSlice, _ rhs: _ByteBufferSlice) -> Bool {
+        return lhs._begin == rhs._begin && lhs.upperBound == rhs.upperBound
+    }
+}
+
+@_versioned
+struct _ByteBufferSlice {
```
Lukasa: Similarly, can we get a doc comment here please?

Sources/NIO/ByteBuffer-core.swift
Lukasa: Also as noted in the chat, the mask isn't really right anyway.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -486,8 +577,19 @@ public struct ByteBuffer {
         }
         let index = _toIndex(index)
         let length = _toCapacity(length)
+        let sliceBeginIndex = self._slice.lowerBound + index
+
+        guard sliceBeginIndex <= ByteBuffer.Slice.maxSupportedLowerBound else {
```
Lukasa: Nit: the word "begin" here is odd. Start might be better.

Sources/NIO/FileRegion.swift
```diff
@@ -12,6 +12,68 @@
 //
 //===----------------------------------------------------------------------===//
 
+struct _UInt56: ExpressibleByIntegerLiteral {
```
Lukasa: Might be nice to consolidate our helper integers into a single source file.

Sources/NIO/FileRegion.swift
```diff
@@ -12,6 +12,68 @@
...
+
+    static let min: _UInt56 = 0
+}
+
+extension _UInt56 {
+    init(_ value: UInt64) {
+        precondition(value & 0xff_00_00_00_00_00_00_00 == 0)
+        self.init(b1234: UInt32(truncatingIfNeeded: (value &          0xff_ff_ff_ff) >> 0 ),
+                  b56:   UInt16(truncatingIfNeeded: (value &    0xff_ff_00_00_00_00) >> 32),
+                  b7:     UInt8(truncatingIfNeeded: (value & 0xff_00_00_00_00_00_00) >> 48))
```
Lukasa: Given your earlier precondition, this mask is unnecessary.

Sources/NIO/ByteBuffer-core.swift
weissi: oops! thanks

Sources/NIO/FileRegion.swift
weissi: fair enough, will do that

Sources/NIO/IntegerTypes.swift
```diff
@@ -0,0 +1,137 @@
...
+        var newValue: Int = 0
+        newValue  = Int(value.b12)
+        newValue |= Int(value.b3) << 16
+        self = newValue
+    }
+}
+
+extension _UInt24 {
+    init(_ value: UInt32) {
+        precondition(value & 0xff_00_00_00 == 0, "value \(value) too large for _UInt24")
```
normanmaurer: @weissi should this be an assert as there is no way for the user to actual construct this ? I think we usually use `assert` for things that would only happen because of a bug in NIO itself

Sources/NIO/IntegerTypes.swift
```diff
@@ -0,0 +1,137 @@
...
+    static var max: _UInt56 {
+        return .init(b1234: .max, b56: .max, b7: .max)
+    }
+    
+    static let min: _UInt56 = 0
+}
+
+extension _UInt56 {
+    init(_ value: UInt64) {
+        precondition(value & 0xff_00_00_00_00_00_00_00 == 0)
```
normanmaurer: @weissi should this be an assert as there is no way for the user to actual construct this ? I think we usually use `assert` for things that would only happen because of a bug in NIO itself

Also consider adding a message as well

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1214,6 +1214,87 @@ class ByteBufferTest: XCTestCase {
...
+        var buf = self.allocator.buffer(capacity: totalBufferSize)
+        for _ in 0..<5 {
+            var fresh = fourMBBuf
+            buf.write(buffer: &fresh)
+        }
+        XCTAssertEqual(totalBufferSize, buf.readableBytes)
+
+        let offset = Int(_UInt24.max) + 1
+        let expectedReadablebytes = totalBufferSize - offset
+        let slice = buf.getSlice(at: offset, length: expectedReadablebytes)!
```
Lukasa: Minor nit: it might be better if we sliced the same size as the test above, to isolate the test differences to only being the slice start index, rather than the slice sizes.

weissi: addressed and improved test (by marking some 'important' positions with different byte values)

normanmaurer:
@weissi please merge once CI completes

PR #351, opened 2018-04-24
shrink HTTP{Request,Response}Head by CoW boxing them

weissi:
only really makes sense with #349 

Motivation:

HTTP{Request,Response}Head were too large even with a less than 3 word
ByteBuffer, this shrinks them box CoW boxing them. Should also reduce
ARC traffic when passing around a bunch.

Modifications:

CoW box HTTP{Request,Response}Head

Result:

fewer existential containers created, more performance


weissi:
tests failed because allocations went up which is expected until #349 lands, retrying then

weissi:
#349 now merged, so tests should now pass

normanmaurer:
@weissi yep it did... so I merged this ;)

PR #356, opened 2018-04-25
Fix warning during compiling tests.

normanmaurer:
Motivation:

When compiling the tests the following warning is printed:

swift-nio/Tests/NIOTests/SelectorTest.swift:361:43: warning: result of call to 'wait()' is unused
        }.wait().forEach { return try! $0.wait() } as Void)

Modifications:

Ignore return value to silent the compiler warnings.

Result:

No more warnings.

### DIFF
--- a/Tests/NIOTests/SelectorTest.swift
+++ b/Tests/NIOTests/SelectorTest.swift
@@ -358,7 +358,7 @@ class SelectorTest: XCTestCase {
                         return channel
                     }
                 }
-        }.wait().forEach { return try! $0.wait() } as Void)
+        }.wait().forEach { XCTAssertNoThrow(try $0.wait()) } as Void)
         XCTAssertNoThrow(try everythingWasReadPromise.futureResult.wait())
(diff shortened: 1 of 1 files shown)


tomerd:

tomerd:

Tests/NIOTests/SelectorTest.swift
```diff
@@ -358,7 +358,7 @@ class SelectorTest: XCTestCase {
                         return channel
                     }
                 }
-        }.wait().forEach { return try! $0.wait() } as Void)
+        }.wait().forEach { _ = try! $0.wait() } as Void)
```
Lukasa: Rather than just suppress the return value, why not wrap this in a `XCTAssertNoThrow` and remove the `try!`? That way we have `.forEach { XCTAssertNoThrow(try $0.wait) }`, which would be a bit better.

normanmaurer: sure why not...

normanmaurer:
@Lukasa addressed

PR #359, opened 2018-04-26
improve ByteBuffer documentation regarding CoW on slices

weissi:
Motivation:

We got a very valid question on Twitter
(https://twitter.com/karim_elna/status/989516133421199361) regarding
ByteBuffers 'sharing storage'. Our docs almost sound like a ByteBuffer
and its slices share storage even after a write. Especially given that
this is a difference to Netty, we should document this better.

Modifications:

Added clarifying documentation.

Result:

Hopefully the docs are more clear now.


normanmaurer:

PR #362, opened 2018-04-26
Not do any precondition range checks when moving the reader / writer …

normanmaurer:
…index from within our methods of ByteBuffer.

Motivation:

We should not need to do any precondition range changes when we change the reader / writer index within our ByteBuffer methods as we know the index is valid. Beside this we already have asserts in place.

Modifications:

Add missing _ methods to adjust the reader / writer index and make use of these.

Result:

Less range checks within our ByteBuffer methods.


tomerd:

normanmaurer:
Let me merge as @weissi and @Lukasa approved :)

PR #363, opened 2018-04-26
Not cause a CoW when calling ByteBuffer.discardReadBytes() while it wa…

normanmaurer:
…s fully consumed.

Motivation:

We should not need to trigger any copy when the ByteBuffer was fully consumed and discardReadBytes() is called.

Modifications:

- If fully consumed just reset reader and writer Index.
- Add unit test

Result:

Less memory copies


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -550,6 +550,13 @@ public struct ByteBuffer {
             return false
         }
 
+        if self._readerIndex == self.writerIndex {
```
Lukasa: Why are we hitting different indices here, one underscored and one not?

normanmaurer: because I have fat fingers 🤤 ... Let me fix it.

normanmaurer: fixed

PR #367, opened 2018-04-27
make BB.debugDescription print xxd -r -p compatible format

weissi:
Motivation:

When debugging, I often use ByteBuffer.debugDescription which dumps
the bytes. Unfortunately, the format wasn't compatible with what
`xxd -r -p` expects. Now you can just copy some output, for example

    echo '04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11 12' | xxd -r -p

and xxd -r -p will correctly parse the byte values.

Modifications:

made BB.debugDescription prefix byte values under 0x10 with a 0 instead
of a space.

Result:

easier debugging with the standard tools


normanmaurer:
@weissi can you please rebase and resolve the conflicts ?

weissi:
@normanmaurer done

PR #369, opened 2018-04-27
[2.0] Remove decodeLast.

### LINKED ISSUE #108: ByteToMessageDecoder behaves gracelessly when channelInactive is triggered while decode() is on the stack.
### Expected behavior

When a `ByteToMessageDecoder` calls `ctx.close()` during its processing, and that `close` completes synchronously, receiving `channelInactive` will not cause ByteToMessageDecoder to re-parse previously parsed data.

### Actual behavior

When this happens, `ByteToMessageDecoder.channelInactive` calls `decodeLast` with a fresh copy `self.cumulationBuffer`. By default, `decodeLast` is the same as `decode`, so this essentially calls `decode` in almost all cases. This means that the call stack ends up looking like this:

1. `decode()`
2. `ctx.close()`
...
3. `decodeLast()`
4. `decode()`

Worse, this second call to `decode` does it with an out of date copy of the `cumulationBuffer`, which it necessarily must do because we always operate on a copy of the `cumulationBuffer` and save it off after our calls to `decode`.

```swift
final class BufferLengthReporter: ByteToMessageDecoder {
    typealias InboundIn = ByteBuffer
    typealias InboundOut = Int

    var cumulationBuffer: ByteBuffer?

    func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
        ctx.fireChannelRead(self.wrapInboundOut(buffer.readableBytes))
        buffer.moveReaderIndex(forwardBy: 4)
        ctx.close(promise: nil)
        return .needMoreData
    }

    func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
        ctx.fireChannelRead(self.wrapInboundOut(buffer.readableBytes))
        return .needMoreData
    }
}

class Test: XCTestCase {
    func testDontRepeatedlyParse() throws {
        defer {
            _ = try? channel.finish()
        }

        XCTAssertNoThrow(try channel.pipeline.add(handler: BufferLengthReporter()).wait())

        var buffer = channel.allocator.buffer(capacity: 5)
        buffer.write(staticString: "hello")

        XCTAssertTrue(try channel.writeInbound(buffer))
        XCTAssertFalse(channel.isActive)

        XCTAssertEqual(5, channel.readInbound())
        XCTAssertEqual(1, channel.readInbound())
        XCTAssertEqual(nil as Int?, channel.readInbound())
    }
}
```

The second `XCTAssertEqual` in the final block fails, reporting `5` instead of `1`.

1.1.0

#### 1. Remove `decodeLast`.

`decodeLast` is a strange function. It exists for protocols for whom EOF is a meaningful signal, but those protocols are not that common. More importantly, in the case of those protocols they could simply implement `channelInactive` themselves to handle the case. Right now, for the default protocol case, `decodeLast` is simply a waste of CPU: the protocol couldn't decode the buffer before, no reason to assume it can now.

The advantage of removing `decodeLast` is that it becomes clear that this is a thing users need to handle themselves. The disadvantage is that it doesn't solve this problem for handlers that treat EOF as meaningful: they still need a way to work out whether they're synchronously receiving EOF from the remote peer.

#### 2. Change the default behaviour of `decodeLast`.

A smaller-scale version of (1), this would just change `decodeLast` to be a no-op. This avoids the current problem, but still provides the hook. The downside of this is, again, we end up letting users fall back into this trap.

#### 3. Manage state better in `ByteToMessageDecoder`.

This is probably my preferred option, but it makes B2MD substantially more complex. Ultimately the core issue is that B2MD will *eagerly* dispatch `channelInactive`, which means that it's partway through parsing when that message is processed. That's just a bad idea. So we could add state to `B2MD` to record that it has seen `channelInactive`, but to not dispatch it while we're inside `decode`. Instead, we'd let `decode` return and then spot the change, and transition to calling `decodeLast`.

This is likely to behave better in the long-term.

#### 4. Prevent `channelInactive` firing synchronously.

This is not a great fix, but most users do not realise that the moment they call out to third-party code they are at risk of being re-entrantly called by all kinds of events. We could, in principle, prevent `close` re-entrantly firing `channelInactive`, but frankly I don't think that's a very good idea. I just wanted to note it here.

All-in-all I think I prefer (3), but I'd like to hear what others think.

Lukasa:
Motivation:

decodeLast is nothing more or less than a bug magnet. It by default
assumes that it is safe to re-enter decode synchronously, despite its own
execution model ensuring that this is not the case. Additionally, it does
this quietly, without notifying most naive users of B2MD that the method
exists at all.

This is done to ensure that parsers successfully parse everything, but the
implementation is silly. In particular, it assumes that this uncommon
behaviour (needing to drain the parser on EOF) is sufficiently common
to justify burdening all implementations with it. Most protocol parsers
do not need to do anything special on EOF, because EOF is not a meaningful
signal for their protocol.

On top of this, B2MD does not even properly handle this behaviour becuase
it is not half-close aware. If a TCP half-close is received, B2MD will
not call decodeLast. Of course, protocols that treat EOF as semantic
are also the protocols that are most-likely to benefit from half-closure
support, meaning that the protocol types B2MD is working to support here
need custom code to work properly anyway.

Given that, and the long sequence of bugs caused by decodeLast, we should
remove it. In the extremely unusual case that EOF is semantic for the
protocol, users can reimplement something like decodeLast using
channelInactive and userInboundEventReceived.

Modifications:

Deleted decodeLast.

Result:

Fewer bugs. Happier users.

Resolves #108.

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -52,15 +52,6 @@ public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == B
     //             again once more data is present in the `ByteBuffer`.
     func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState
 
-    /// This method is called once, when the `ChannelHandlerContext` goes inactive (i.e. when `channelInactive` is fired)
-    ///
-    /// - parameters:
-    ///     - ctx: The `ChannelHandlerContext` which this `ByteToMessageDecoder` belongs to.
-    ///     - buffer: The `ByteBuffer` from which we decode.
-    /// - returns: `DecodingState.continue` if we should continue calling this method or `DecodingState.needMoreData` if it should be called
-    //             again when more data is present in the `ByteBuffer`.
-    func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws  -> DecodingState
-
     /// Called once this `ByteToMessageDecoder` is removed from the `ChannelPipeline`.
     ///
     /// - parameters:
@@ -121,21 +112,6 @@ extension ByteToMessageDecoder {
         }
     }
 
-    /// Call `decodeLast` before forward the event through the pipeline.
-    public func channelInactive(ctx: ChannelHandlerContext) {
-        if var buffer = cumulationBuffer {
-            ctx.withThrowingToFireErrorAndClose {
-                // Running decodeLast method until either the user returned `.needMoreData` or an error occurred.
-                while try decodeLast(ctx: ctx, buffer: &buffer)  == .`continue` && buffer.readableBytes > 0 { }
-            }
-
-            // Once the Channel goes inactive we can just drop all previous buffered data.
-            cumulationBuffer = nil
-        }
-
-        ctx.fireChannelInactive()
-    }
-
     public func handlerAdded(ctx: ChannelHandlerContext) {
         decoderAdded(ctx: ctx)
     }
@@ -149,11 +125,6 @@ extension ByteToMessageDecoder {
         }
         cumulationBuffer = nil
         decoderRemoved(ctx: ctx)
-    }
-
-    /// Just call `decode`. Users may implement their own logic.
-    public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
-        return try decode(ctx: ctx, buffer: &buffer)
     }
 
     /// Do nothing by default.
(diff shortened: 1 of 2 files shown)


Lukasa:
By the way, I think it's hugely meaningful that this patch is only deletions. There are exactly zero tests and zero protocols relying on the behaviour of `decodeLast`.

weissi:
@Lukasa shouldn't we remove the definitions of `decodeLast` too because that code won't ever be run now, right?

Lukasa:
In the master that existed when I wrote this patch there were no user definitions of `decodeLast`. There is now one, in the WS decoder, but the HTTP decoder doesn't implement `decodeLast` because it was literally impossible for it to have a correct implementation. It implements `channelInactive` instead. ;)

That said, yes, I'll rebase on the current master and remove the WS implementation.

weissi:
We fixed this in a different way, B2MD2 is not done yet but the bug-magnetiness of decodeLast is gone

PR #370, opened 2018-04-28
Add better guards again re-entrance in ByteToMessageDecoder

normanmaurer:
Motivation:

When the ByteToMessageDecoder re-entranced in its decode(...) / decodeLast(...) methods it could be possible that the ordering of processing and so the seen bytes are mixed.

Modifications:

- Refresh the buffer on each loop iteration and update the cumulationBuffers indicies.

Result:

More robust code.

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -38,6 +38,13 @@ public enum DecodingState {
 /// If you move the reader index forward, either manually or by using one of `buffer.read*` methods, you must ensure
 /// that you no longer need to see those bytes again as they will not be returned to you the next time `decode` is called.
 /// If you still need those bytes to come back, consider taking a local copy of buffer inside the function to perform your read operations on.
+///
+/// The `ByteBuffer` passed in as `buffer` is a slice of a larger buffer owned by the `ByteToMessageDecoder` implementation. Some aspects of this buffer are preserved across calls to `decode`, meaning that any changes to those properties you make in your `decode` method will be reflected in the next call to decode. In particular, the following operations are have the described effects:
+
+/// 1. Moving the reader index forward persists across calls. When your method returns, if the reader index has advanced, those bytes are considered "consumed" and will not be available in future calls to `decode`.
+///    Please note, however, that the numerical value of the `readerIndex` itself is not preserved, and may not be the same from one call to the next. Please do not rely on this numerical value: if you need
+///    to recall where a byte is relative to the `readerIndex`, use an offset rather than an absolute value.
+/// 2. Mutating the bytes in the buffer will cause undefined behaviour and likely crash your program
 public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == ByteBuffer {
     /// The cumulationBuffer which will be used to buffer any data.
     var cumulationBuffer: ByteBuffer? { get set }
@@ -96,64 +103,92 @@ private extension ChannelHandlerContext {
 
 extension ByteToMessageDecoder {
 
+    /// Decode in a loop until there is nothing more to decode.
+    private func decodeLoop(ctx: ChannelHandlerContext, decodeFunc: (ChannelHandlerContext, inout ByteBuffer) throws -> DecodingState) throws {
+        while var slice = self.cumulationBuffer?.slice(), slice.readableBytes > 0 {
+            // Needed to later calculate how much we need to advance the readerIndex of the cumulationBuffer.
+            let sliceReadable = slice.readableBytes
+            let sliceWriterIndex = slice.writerIndex
+
+            // We fetch the writerIndex of the cumulationBuffer to make a good guess about if the cumulationBuffer changed in between due re-entrant call to
+            // channelRead after we called decodeFunc(...).
+            let writerIndex = self.cumulationBuffer!.writerIndex
+            let result = try decodeFunc(ctx, &slice)
+
+            guard self.cumulationBuffer != nil else {
+                // The cumulationBuffer was set to nil by either removing the decoder or closing the channel, just break the loop.
+                break
+            }
+
+            precondition(slice.writerIndex == sliceWriterIndex, "Writing to the buffer is not allowed")
+
+            self.cumulationBuffer!.moveReaderIndex(forwardBy: sliceReadable - slice.readableBytes)
+
+            // If the user told us more data is needed we also need to ensure the writerIndex did not change in between.
+            // If the writerIndex changed we need to retry as there is more data delivered via re-entrance maybe.
+            if result == .needMoreData && self.cumulationBuffer!.writerIndex == writerIndex {
+                break
+            }
+        }
+    }
+
     /// Calls `decode` until there is nothing left to decode.
     public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
-        var buffer = self.unwrapInboundIn(data)
+        // Either merge the received data into the existing cumulationBuffer or use it as the cumulationBuffer if none exists yet.
         if self.cumulationBuffer != nil {
+            var buffer = self.unwrapInboundIn(data)
             self.cumulationBuffer!.write(buffer: &buffer)
-            buffer = self.cumulationBuffer!
         } else {
-            self.cumulationBuffer = buffer
+            self.cumulationBuffer = self.unwrapInboundIn(data)
         }
 
         ctx.withThrowingToFireErrorAndClose {
-            // Running decode method until either the user returned `.needMoreData` or an error occurred.
-            while try decode(ctx: ctx, buffer: &buffer) == .`continue` && buffer.readableBytes > 0 { }
+            try self.decodeLoop(ctx: ctx, decodeFunc: self.decode)
         }
 
-        if buffer.readableBytes > 0 {
-            if self.shouldReclaimBytes(buffer: buffer) {
-                buffer.discardReadBytes()
-            }
-            cumulationBuffer = buffer
-        } else {
-            cumulationBuffer = nil
+        // Discard the cumulationBuffer or discard read bytes if needed.
+        guard let buffer = self.cumulationBuffer, buffer.readableBytes > 0 else {
+            self.cumulationBuffer = nil
+            return
+        }
+
+        // Check if we should reclaim some bytes and if so do it.
+        if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) {
+            self.cumulationBuffer!.discardReadBytes()
         }
     }
 
     /// Call `decodeLast` before forward the event through the pipeline.
     public func channelInactive(ctx: ChannelHandlerContext) {
-        if var buffer = cumulationBuffer {
+        if self.cumulationBuffer != nil {
             ctx.withThrowingToFireErrorAndClose {
-                // Running decodeLast method until either the user returned `.needMoreData` or an error occurred.
-                while try decodeLast(ctx: ctx, buffer: &buffer)  == .`continue` && buffer.readableBytes > 0 { }
+                try self.decodeLoop(ctx: ctx, decodeFunc: self.decodeLast)
             }
-
             // Once the Channel goes inactive we can just drop all previous buffered data.
-            cumulationBuffer = nil
+            self.cumulationBuffer = nil
         }
 
         ctx.fireChannelInactive()
     }
 
     public func handlerAdded(ctx: ChannelHandlerContext) {
-        decoderAdded(ctx: ctx)
+        self.decoderAdded(ctx: ctx)
     }
 
     public func handlerRemoved(ctx: ChannelHandlerContext) {
-        if let buffer = cumulationBuffer as? InboundOut {
+        if let buffer = self.cumulationBuffer as? InboundOut {
             ctx.fireChannelRead(self.wrapInboundOut(buffer))
         } else {
             /* please note that we're dropping the partially received bytes (if any) on the floor here as we can't
                send a full message to the next handler. */
         }
-        cumulationBuffer = nil
-        decoderRemoved(ctx: ctx)
+        self.cumulationBuffer = nil
+        self.decoderRemoved(ctx: ctx)
     }
 
     /// Just call `decode`. Users may implement their own logic.
     public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
-        return try decode(ctx: ctx, buffer: &buffer)
+        return try self.decode(ctx: ctx, buffer: &buffer)
     }
 
     /// Do nothing by default.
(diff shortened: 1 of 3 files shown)


normanmaurer:
This needs more tests and we may want to split it in two PRs (while the second needs the first)...

That said here are some numbers for allocations:


Before: `1000_reqs_1_conn.total_allocations: 61113`
After: `1000_reqs_1_conn.total_allocations: 58116`

And benchmark via wrk:

Before:
```
  4 threads and 256 connections
    Latency   104.15ms  206.11ms   1.99s    91.22%
    Req/Sec   132.16k    26.40k  324.24k    89.22%
  5258220 requests in 10.03s, 250.73MB read
  Socket errors: connect 0, read 154, write 0, timeout 22
Requests/sec: 524473.37
Transfer/sec:     25.01MB
```

After:
```
  4 threads and 256 connections
    Latency   103.21ms  204.95ms   1.98s    88.99%
    Req/Sec   135.99k    25.68k  304.99k    88.75%
  5425784 requests in 10.03s, 258.72MB read
  Socket errors: connect 0, read 157, write 0, timeout 12
Requests/sec: 540810.81
Transfer/sec:     25.79MB
```

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -186,14 +185,29 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
 
-    fileprivate var pendingInOut: [NIOAny] = []
     fileprivate var state = HTTPParserState()
 
     fileprivate init(type: HTTPMessageT.Type) {
         /* this is a private init, the public versions only allow HTTPClientResponsePart and HTTPServerRequestPart */
         assert(HTTPMessageT.self == HTTPClientResponsePart.self || HTTPMessageT.self == HTTPServerRequestPart.self)
     }
 
+    deinit {
```
normanmaurer: I used a `deinit { }` for it as `decoderRemoved` may be called while still in the `decode(...)` and so I can not modify the `parser`.

Sources/NIO/Codec.swift
```diff
@@ -96,59 +96,83 @@ private extension ChannelHandlerContext {
 
 extension ByteToMessageDecoder {
 
+    /// Decode in a loop until there is nothing more to decode.
+    private func decodeLoop(ctx: ChannelHandlerContext, decodeFunc: (ChannelHandlerContext, inout ByteBuffer) throws -> DecodingState) {
+        assert(self.cumulationBuffer != nil)
+        ctx.withThrowingToFireErrorAndClose {
```
Lukasa: Probably better to hoist this up to the caller.

Sources/NIO/Codec.swift
```diff
@@ -96,59 +96,83 @@ private extension ChannelHandlerContext {
...
+            while true {
+                var buffer = self.cumulationBuffer!
+                let readable = buffer.readableBytes
+                let writerIndex = buffer.writerIndex
+                let result = try decodeFunc(ctx, &buffer)
+                if self.cumulationBuffer != nil {
+                    self.cumulationBuffer!.moveReaderIndex(forwardBy: readable - buffer.readableBytes)
```
Lukasa: Is this right? We’re passing a local copy of the buffer, but then mutating the cumulation buffer. Why not just preserve the buffer returned from the call to decode, especially as we’re bothering to pass it `inout`.

Sources/NIO/Codec.swift
```diff
@@ -96,59 +96,83 @@ private extension ChannelHandlerContext {
...
-            while try decode(ctx: ctx, buffer: &buffer) == .`continue` && buffer.readableBytes > 0 { }
-        }
+        self.decodeLoop(ctx: ctx, decodeFunc: self.decode)
 
-        if buffer.readableBytes > 0 {
-            if self.shouldReclaimBytes(buffer: buffer) {
-                buffer.discardReadBytes()
+        if self.cumulationBuffer != nil {
+            if self.cumulationBuffer!.readableBytes > 0 {
+                if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) && self.cumulationBuffer!.discardReadBytes() && self.cumulationBuffer!.readableBytes == 0 {
```
Lukasa: Let’s make this clearer.

Sources/NIO/Codec.swift
normanmaurer: I think it’s right but I am all ears for better ideas. Can you show me some code with your suggestion as I don’t understand what exactly you suggest here

Sources/NIO/Codec.swift
normanmaurer: You mean let the error bubble up ?

normanmaurer:
Also tests pass in docker locally... we may need to update timeouts

Sources/NIO/Codec.swift
normanmaurer: Ok let me break up the if statement and add some comments

Sources/NIO/Codec.swift
Lukasa: Let `decodeLoop` throw, place the `ctx.thenThrowingToFireErrorAndClose` in the function that calls `decodeLoop`.

Sources/NIO/Codec.swift
Lukasa: Basically I'm just saying why not just do `self.cumulationBuffer = buffer`?

Lukasa: Ah, I know why: the `cumulationBuffer` may have been mutated. That's awkward.

Lukasa: Hrm, I'm inclined to say that we should probably pass `&buffer` as a slice instead, and then come up with some custom logic to reconcile the changes from that slice with the parent.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -186,14 +185,29 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+        // Remove the stored reference to ChannelHandlerContext
+        parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
+
+        // Set the callbacks to nil as we dont need these anymore
+        settings.on_body = nil
```
weissi: this should have the same effect:

    settings = http_parser_settings()

tomerd:

normanmaurer:
Let me update the PR to only include the `ByteToMessageDecoder` changes.

normanmaurer:
@weissi @Lukasa PTAL again... Will add some more tests as well.

normanmaurer:
@Lukasa haha you were faster then me (and I did not refresh yet). Yep will add a few tests.

weissi:
@normanmaurer awesome!

```
18:30:29 info: 1000_reqs_1_conn: total number of mallocs: 46730
18:30:29 info: 1_reqs_1000_conn: allocations not freed: 0
18:30:29 info: 1_reqs_1000_conn: total number of mallocs: 775997
18:30:29 info: ping_pong_1000_reqs_1_conn: allocations not freed: 0
18:30:29 info: ping_pong_1000_reqs_1_conn: total number of mallocs: 4513
18:30:29 info: bytebuffer_lots_of_rw: allocations not freed: 0
18:30:29 info: bytebuffer_lots_of_rw: total number of mallocs: 3011
18:30:29 info: future_lots_of_callbacks: allocations not freed: 0
18:30:29 info: future_lots_of_callbacks: total number of mallocs: 99001
```

Please edit the docker compose files and lower the limits once again.

Sources/NIO/Codec.swift
```diff
@@ -96,64 +106,95 @@ private extension ChannelHandlerContext {
...
-            // Running decode method until either the user returned `.needMoreData` or an error occurred.
-            while try decode(ctx: ctx, buffer: &buffer) == .`continue` && buffer.readableBytes > 0 { }
+            try self.decodeLoop(ctx: ctx, decodeFunc: self.decode)
         }
 
-        if buffer.readableBytes > 0 {
-            if self.shouldReclaimBytes(buffer: buffer) {
-                buffer.discardReadBytes()
+        // Discard the cumulationBuffer or discard read bytes if needed.
+        guard self.cumulationBuffer != nil else {
```
Lukasa: Any reason not to use `guard let` here?

Sources/NIO/Codec.swift
```diff
@@ -96,64 +106,95 @@ private extension ChannelHandlerContext {
...
+            return
+        }
+
+        guard self.cumulationBuffer!.readableBytes > 0 else {
+            self.cumulationBuffer = nil
+            return
+        }
+
+        if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) && self.cumulationBuffer!.discardReadBytes() {
+            if self.cumulationBuffer!.readableBytes == 0 {
```
Lukasa: Any reason not to just hoist this check up? It won’t be changed by reclaiming bytes.

Sources/NIO/Codec.swift
```diff
@@ -96,64 +106,95 @@ private extension ChannelHandlerContext {
 
 extension ByteToMessageDecoder {
 
+    /// Decode in a loop until there is nothing more to decode.
+    private func decodeLoop(ctx: ChannelHandlerContext, decodeFunc: (ChannelHandlerContext, inout ByteBuffer) throws -> DecodingState) throws {
+        while var slice = self.cumulationBuffer?.slice(), slice.readableBytes > 0 {
```
normanmaurer: @weissi @Lukasa so I started to write some unit tests and notice we can not do this.. The problem here is that it is possible that `decodeLast` will be triggered from within `decodeFunc` (due a `close(...)` call for example). In this case `decodeLast` will by default call `decode(...)` which will then see the same bytes again even if `decode(...)` before increased the `readerIndex` as we had no chance to replicate these changes to the `cumulationBuffer` yet :(

Lukasa: Yeah, there's no way to entirely prevent re-entrancy without removing decodeLast.

normanmaurer: yeah... so I wonder what we should do here in the case... any suggestions ?

Lukasa: I mean, I'm inclined to sit on it and wait until 2.0, when decodeLast is removed. In the meantime we can recommend users override decodeLast or channelInactive.

Sources/NIO/Codec.swift
normanmaurer: Because we want to act on the `self.cumulationBuffer` on the following lines and not on a different copy ?

Sources/NIO/Codec.swift
normanmaurer: @Lukasa you are right this can be removed.

Sources/NIO/Codec.swift
Lukasa: I mean, it doesn't matter what you do, the issue is still the same: `decodeLast` is not safe from avoid reentrancy. So you may as well keep using the slice.

Sources/NIO/Codec.swift
```diff
@@ -96,64 +103,94 @@ private extension ChannelHandlerContext {
...
+            let writerIndex = self.cumulationBuffer!.writerIndex
+            let result = try decodeFunc(ctx, &slice)
+
+            guard self.cumulationBuffer != nil else {
+                // The cumulationBuffer was set to nil by either removing the decoder or closing the channel, just break the loop.
+                break
+            }
+
+            guard slice.writerIndex == sliceWriterIndex else {
+                fatalError("Writing to the buffer is not allowed")
```
Lukasa: Minor nit, but I tend to prefer these to be `preconditionFailure`.

Tests/NIOTests/CodecTest.swift
```diff
@@ -237,6 +237,95 @@ public class ByteToMessageDecoderTest: XCTestCase {
...
+            XCTAssertNoThrow(try channel.finish())
+        }
+
+        class TestDecoder: ByteToMessageDecoder {
+
+            typealias InboundIn = ByteBuffer
+            typealias InboundOut = ByteBuffer
+
+            var cumulationBuffer: ByteBuffer?
+            var triggeredReentrace: Bool = false
```
Lukasa: nit: "reentrance"

Tests/NIOTests/CodecTest.swift
```diff
@@ -237,6 +237,95 @@ public class ByteToMessageDecoderTest: XCTestCase {
...
+                    XCTAssertTrue(self.triggeredReentrace)
+                    return .needMoreData
+                }
+                self.decode = true
+                defer {
+                    self.decode = false
+                }
+
+                if !self.triggeredReentrace {
+                    self.triggeredReentrace = true
```
Lukasa: Rather than set this here, it might be better to let the re-entrant call set it (where you currently have an `XCTAssertTrue`) and then check it at the end of the test. Otherwise this test could pass if you never re-entered at all.

Sources/NIO/Codec.swift
normanmaurer: sure why not... I dont care and I think at the end it makes no difference as long as it crashes :)

Sources/NIO/Codec.swift
```diff
@@ -38,6 +38,13 @@ public enum DecodingState {
...
+///
+/// The `ByteBuffer` passed in as `buffer` is a slice of a larger buffer owned by the `ByteToMessageDecoder` implementation. Some aspects of this buffer are preserved across calls to `decode`, meaning that any changes to those properties you make in your `decode` method will be reflected in the next call to decode. In particular, the following operations are have the described effects:
+
+/// 1. Moving the reader index forward persists across calls. When your method returns, if the reader index has advanced, those bytes are considered "consumed" and will not be available in future calls to `decode`.
+///    Please note, however, that the numerical value of the `readerIndex` itself is not preserved, and may not be the same from one call to the next. Please do not rely on this numerical value: if you need
+///    to recall where a byte is relative to the `readerIndex`, use an offset rather than an absolute value.
+/// 2. Mutating the bytes in the buffer will cause a `fatalError` and so is not allowed. You are only allowed to move the readerIndex forward or consume from the buffer.
```
weissi: should we say 'mutating the bytes or the readerIndex will cause undefined behaviour and likely crash your program' or something?

Sources/NIO/Codec.swift
```diff
@@ -96,64 +103,94 @@ private extension ChannelHandlerContext {
...
+            // channelRead after we called decodeFunc(...).
+            let writerIndex = self.cumulationBuffer!.writerIndex
+            let result = try decodeFunc(ctx, &slice)
+
+            guard self.cumulationBuffer != nil else {
+                // The cumulationBuffer was set to nil by either removing the decoder or closing the channel, just break the loop.
+                break
+            }
+
+            guard slice.writerIndex == sliceWriterIndex else {
```
weissi: why not just `precondition(slice.writeIndex == sliceWriterIndex, "...")`?

PR #371, opened 2018-04-28
Refactored UInt32 and UInt64 methods nextPowerOf2

adamnemecek:
Refactored nextPowerOf2 methods.

### Motivation:

Make them shorter and somewhat faster.

### Modifications:

Changed the implementations of nextPowerOf2

### Result:

Methods are shorter.

### DIFF
--- a/Sources/NIO/ByteBuffer-int.swift
+++ b/Sources/NIO/ByteBuffer-int.swift
@@ -97,22 +97,10 @@ extension ByteBuffer {
 extension UInt64 {
     /// Returns the next power of two.
     public func nextPowerOf2() -> UInt64 {
-        guard self > 0 else {
+        if self == 0 {
             return 1
         }
-
-        var n = self
-
-        n -= 1
-        n |= n >> 1
-        n |= n >> 2
-        n |= n >> 4
-        n |= n >> 8
-        n |= n >> 16
-        n |= n >> 32
-        n += 1
-
-        return n
+        return 1 << (64 - (self - 1).leadingZeroBitCount)
     }
 }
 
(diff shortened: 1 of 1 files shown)


Lukasa:

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -97,22 +97,10 @@ extension ByteBuffer {
...
-        n |= n >> 1
-        n |= n >> 2
-        n |= n >> 4
-        n |= n >> 8
-        n |= n >> 16
-        n |= n >> 32
-        n += 1
-
-        return n
+        return 1 << (64 - (self - 1).leadingZeroBitCount)
```
weissi: @adamnemecek ah nice, didn't know we had `leadingZeroBitCount`, hope that gets lowered to the `LZCNT` instruction.

You can actually get the number of bits a type has from `Self.bitWidth`. Then the code for all unsigned integer types will actually be the same. You can probably just put the extension on `UnsignedInteger` and you'll get it for all.

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -97,22 +97,10 @@ extension ByteBuffer {
 extension UInt64 {
     /// Returns the next power of two.
     public func nextPowerOf2() -> UInt64 {
-        guard self > 0 else {
+        if self == 0 {
```
weissi: I don't think we should take this change. Happy to make it `guard self != 0 else { ... }` though. But the `guard` tells/should tell the compiler that this is the less likely case and it can optimise for that.

Sources/NIO/ByteBuffer-int.swift
adamnemecek: It uses bsr https://godbolt.org/g/A9RjhW

weissi: @adamnemecek awesome, thank you! TIL that godbolt supports Swift, how cool's that!

adamnemecek: You can guess who added the support. His name starts with A and ends with dam Nemecek.

tomerd:

tomerd:

weissi:

Lukasa:

PR #373, opened 2018-04-30
Add UnsafeEmbeddedAtomic and use it.

Lukasa:
Motivation:

Many uses of atomics restrict themselves to using atomic variables within
objects that already have reference semantics (i.e. classes) and have
appropriate lifetime management. In this case, the current NIO Atomic
performs an unnecessary heap allocation in order to get itself a reference
count that it can use to free the underlying heap memory.

This is an unnecessary allocation that can often be elided, especially in
situations where many such atomics may be needed. Thus, we should provide
the ability to manually manage the lifetime of the heap-allocated memory for
the atomic, and change some uses of it in our codebase to reduce overhead.

Modifications:

1. Moved most of the code of Atomic to a new struct, UnsafeEmbeddedAtomic.
2. Reimplemented Atomic in terms of UnsafeEmbeddedAtomic.
3. Updated EventLoopFuture to use an UnsafeEmbeddedAtomic for isFulfilled.

Result:

1/3 as many allocations per EventLoopFuture.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_allocation_counts_for_http1.sh
+++ b/IntegrationTests/tests_04_performance/test_01_allocation_counts_for_http1.sh
@@ -58,7 +58,7 @@ cd ..
 "$swift_bin" run -c release | tee "$tmp/output"
 )
 
-for test in 1000_reqs_1_conn 1_reqs_1000_conn ping_pong_1000_reqs_1_conn bytebuffer_lots_of_rw; do
+for test in 1000_reqs_1_conn 1_reqs_1000_conn ping_pong_1000_reqs_1_conn bytebuffer_lots_of_rw future_lots_of_callbacks; do
     cat "$tmp/output"  # helps debugging
     total_allocations=$(grep "^$test.total_allocations:" "$tmp/output" | cut -d: -f2 | sed 's/ //g')
     not_freed_allocations=$(grep "^$test.remaining_allocations:" "$tmp/output" | cut -d: -f2 | sed 's/ //g')
--- a/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
@@ -343,5 +343,72 @@ public func swiftMain() -> Int {
         return buffer.readableBytes
     }
 
+    measureAndPrint(desc: "future_lots_of_callbacks") {
+        struct MyError: Error { }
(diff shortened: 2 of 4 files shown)


Lukasa:
In addition to this, on @weissi's suggestion I added an allocation test for futures. Before this patch, the allocation result on my Mac was `MAX_ALLOCS_ALLOWED_future_lots_of_callbacks=123004`. After this patch, it was `MAX_ALLOCS_ALLOWED_future_lots_of_callbacks=99004`. This is a reduction of 24k allocations over 1k runs, or 24 allocations per run. This implies we have 24 Futures, which matches my manual reading of the code.

The remaining allocation count is 99k (plus a DC offset of 4 allocations for loops, locks, and other things). This means 99 allocations per run. 48 of those allocations remain for the Futures and atomics, leaving 51 unaccounted-for in this test. I may try @weissi's dtrace memory scripts on this binary to see if I can work out where they are, and whether they can be made to go away.

Lukasa:
Ah, and I can account for the DC offset too. An `EmbeddedEventLoop` (+1), containing a `PriorityQueue` containing a `Heap` containing a `ContiguousArray` (+1). Two `Lock`s (+1 each), one for each call to `wait`. That's 4.

Lukasa:
Further digging reveals that almost all of the other allocations are the result of the cross-module generics cost: allocating boxes for type metadata for all the generic methods on EventLoopFuture. I'm going to quickly investigate what happens if I make all the tested methods on EventLoopFuture inlineable.

Lukasa:
For reasons that I don't fully understand at this time, adding `@_inlineable` and `@_versioned` to most of `EventLoopFuture` actually makes this allocate *more*, not less. I haven't worked out why yet, but I've officially spent more time on trying to figure it out than I can justify. The inlineability seems to be working correctly: I'm seeing specialisations for all of the cases in this test. But somehow, this just leads to more allocation. Weird.

tomerd:

Lukasa:
Ok, this is ready for review @normanmaurer @weissi.

weissi:
ok, CI reported `99001` allocations. So in this particular test I'd recommend setting the maximum to `99002` as it should be really stable because it doesn't use multi-threaded event loops

normanmaurer:
Thanks 👍

Lukasa:
Agreed. @tomerd, can we set another CI environment variable? We need `MAX_ALLOCS_ALLOWED_future_lots_of_callbacks=99100` set, please. :grin:

tomerd:
@Lukasa now you do it via the docker compose file :)

Lukasa:
Oh cool, that works.

PR #374, opened 2018-04-30
fix some Swift 4.2 warnings

weissi:
Motivation:

Warnings are bad.

Modifications:

Changed things that still work in 4.0/4.1 but are warnings in 4.2 to
improved versions.

Result:

less warnings on the upcoming 4.2

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -658,10 +658,9 @@ extension StaticString: Collection {
     public var endIndex: Index { return self.utf8CodeUnitCount }
     public func index(after i: Index) -> Index { return i + 1 }
 
-    public subscript(position: Int) -> StaticString.Element {
-        get {
-            return self[position]
-        }
+    public subscript(position: Int) -> UInt8 {
+        precondition(position < self.utf8CodeUnitCount, "index \(position) out of bounds")
+        return self.utf8Start.advanced(by: position).pointee
     }
 }
 
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -24,6 +24,11 @@ public protocol AppendableCollection: Collection {
 /// will automatically expand if more elements than `initialRingCapacity` are stored, it's advantageous to prevent
 /// expansions from happening frequently. Expansions will always force an allocation and a copy to happen.
 public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
+    #if swift(>=4.2)
+    public typealias RangeType = Range
+    #else
+    public typealias RangeType = CountableRange
+    #endif
     private var buffer: ContiguousArray<E?>
 
     /// The index into the buffer of the first item
@@ -168,7 +173,7 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
     }
 
     /// Return all valid indices of the ring.
-    public var indices: CountableRange<Int> {
+    public var indices: RangeType<Int> {
         return 0..<self.count
     }
 
--- a/Sources/NIO/MarkedCircularBuffer.swift
+++ b/Sources/NIO/MarkedCircularBuffer.swift
@@ -18,6 +18,12 @@
 /// writes and mark how far through the buffer the user has flushed, and therefore how far through the buffer is
 /// safe to write.
 public struct MarkedCircularBuffer<E>: CustomStringConvertible, AppendableCollection {
+    #if swift(>=4.2)
+    public typealias RangeType = Range
+    #else
+    public typealias RangeType = CountableRange
+    #endif
+
     private var buffer: CircularBuffer<E>
     private var markedIndex: Int = -1 /* negative: nothing marked */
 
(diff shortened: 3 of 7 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -658,10 +658,9 @@ extension StaticString: Collection {
     public var endIndex: Index { return self.utf8CodeUnitCount }
     public func index(after i: Index) -> Index { return i + 1 }
 
-    public subscript(position: Int) -> StaticString.Element {
-        get {
-            return self[position]
-        }
+    public subscript(position: Int) -> UInt8 {
+        precondition(position < self.utf8CodeUnitCount, "index \(position) out of bounds")
+        return self.utf8Start.advanced(by: position).pointee
```
Lukasa: Did the subscript we were using here go away?

weissi: @Lukasa it did just recurse forever ;). Adding a test though

Lukasa: Lol ok, let me know when that test drops and I’ll re-review.

weissi: @Lukasa updated

Sources/NIO/CircularBuffer.swift
```diff
@@ -24,6 +24,11 @@ public protocol AppendableCollection: Collection {
 /// will automatically expand if more elements than `initialRingCapacity` are stored, it's advantageous to prevent
 /// expansions from happening frequently. Expansions will always force an allocation and a copy to happen.
 public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
+    #if swift(>=4.2)
```
BasThomas: Should this be added in multiple files? What about just one file with the `#if` and `#else`? That also makes it easier to find and clean up in the future.

Sources/NIOPriorityQueue/Heap.swift
```diff
@@ -183,6 +183,7 @@ internal struct Heap<T: Comparable> {
 
 @available(*, deprecated, message: "The NIOPriorityQueue module is deprecated and will be removed in the next major release.")
 extension Heap: CustomDebugStringConvertible {
+    @available(*, deprecated, message: "The NIOPriorityQueue module is deprecated and will be removed in the next major release.")
```
BasThomas: I assume these were warnings in 4.2? What is the alternative to the `NIOPriorityQueue`, is that something that can be added?

Sources/NIO/MarkedCircularBuffer.swift
```diff
@@ -18,6 +18,12 @@
 /// writes and mark how far through the buffer the user has flushed, and therefore how far through the buffer is
 /// safe to write.
 public struct MarkedCircularBuffer<E>: CustomStringConvertible, AppendableCollection {
+    #if swift(>=4.2)
+    public typealias RangeType = Range
+    #else
+    public typealias RangeType = CountableRange
```
helje5: I think the `Type` thing is a Swift 2 leftover (which I personally kinda like, but is inconsistent) and not used at all in Swift 3+ anymore, right?. I would just do:

    #if swift(>=4.2)
    #else
      public typealias Range = CountableRange
    #endif

And use `Swift.Range` in the situations in which it might conflict.

Sources/NIOPriorityQueue/Heap.swift
normanmaurer: @BasThomas we deprecated the old one and moved it to NIO for performance reasons.

tomerd:

Sources/NIOPriorityQueue/Heap.swift
BasThomas: Yes, but I was wondering what "the old ones" are. If they are mentioned, it would make a migration easier.

In an `@available`, you can then use the `renamed` parameter to provide users a fix-it within Xcode.

Lukasa: There is no migration: they have been made non-public. Their publicity was accidental only.

Sources/NIO/MarkedCircularBuffer.swift
weissi: @helje5 but in swift < 4.2 there were _distinct_ `Range` and `CountableRange` types, now they're the same (`typealias CountableRange = Range`) and the typealias is deprecated.

But to your point, I could do

```
#if !swift(>=4.2)
public typealias Range = Range
#else
public typealias Range = CountableRange
#endif
```

and then in the methods return `-> Self.Range` ...

Sources/NIO/CircularBuffer.swift
weissi: @BasThomas I don't like it either but I wanted to have the scope of this typealias to be as small as possible. The best would've been to just use `Swift.Range` but in Swift < 4.2 `Range` and `CountableRange` are distinct types therefore a change would've been SemVer major!

Sources/NIO/MarkedCircularBuffer.swift
weissi: @Lukasa what do you think? Leave as is (as `RangeType`) or make it `Range`?

Lukasa: I'm inclined to call it `RangeType` mostly to avoid the risk of the compiler losing track of what name is meant in a given scope. It'll lead to hard-to-debug errors we don't really need.

weissi: cool, that was kind of what I thought too. I hope to remove that again with 2.0.0 when we can target 4.2+

PR #375, opened 2018-05-01
expose sanity script through docker compose

tomerd:
Motivation:

be able to run the sanity check via docker compose

Modifications:

add sanity task to docker compose file

Result:

you can now run sanity check via docker


normanmaurer:
LGTM but who is x ?

Lukasa:
Yup, may need the commit metadata edited here. :D

PR #383, opened 2018-05-02
Add support for compiling and running NIO on Raspberry Pi 32-bit

helje5:
Port SwiftNIO to 32-bit platforms

### Motivation:

Pretty much every developer in the world want to run MicroExpress applications on Raspberry Pi K8s clusters. The unfortunate precondition to that is getting NIO compiled for RaspberryPi. This PR makes it work.

> Thanks go to @chnmrc for providing a Swift 4.1 Raspi build, also available via a DockerHub: [helje5/rpi-swift](https://hub.docker.com/r/helje5/rpi-swift/).

### Modifications:

Two main changes:

1. Timestamps used to be stored as Int, this needs to be an explicit Int64 to provide the necessary value range for ns values
2. Some hardcoded pointers like 0xdeafbeef got mapped to Int values, but Int32 can't represent 0xdeadbeef. Needs to be UInt32 to capture the full deadness.

### Result:

NIO compiles on Raspi 3, 32-bit Ubuntu. I tried the Echo Server, which seems to work, and also the HTTP sample Server, which also seems to work.

P.S.: This should also allow you to run NIO on AppleWatch, which is 32-bit too.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -89,7 +89,7 @@ public struct ByteBufferAllocator {
         #if !arch(arm) // only complain on 64-bit, this is unfortunate reality on 32-bit
             assert(MemoryLayout<ByteBuffer>.size <= 3 * MemoryLayout<Int>.size,
                    "ByteBuffer has size \(MemoryLayout<ByteBuffer>.size) which is larger than the built-in storage of the existential containers.")
-	#endif
+        #endif
         self.malloc = hookedMalloc
         self.realloc = hookedRealloc
         self.free = hookedFree
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -86,8 +86,10 @@ public struct ByteBufferAllocator {
                   hookedRealloc: @escaping @convention(c) (UnsafeMutableRawPointer?, Int) -> UnsafeMutableRawPointer?,
                   hookedFree: @escaping @convention(c) (UnsafeMutableRawPointer?) -> Void,
                   hookedMemcpy: @escaping @convention(c) (UnsafeMutableRawPointer, UnsafeRawPointer, Int) -> Void) {
-        assert(MemoryLayout<ByteBuffer>.size <= 3 * MemoryLayout<Int>.size,
-               "ByteBuffer has size \(MemoryLayout<ByteBuffer>.size) which is larger than the built-in storage of the existential containers.")
+        #if !arch(arm) // only complain on 64-bit, this is unfortunate reality on 32-bit
+            assert(MemoryLayout<ByteBuffer>.size <= 3 * MemoryLayout<Int>.size,
+                   "ByteBuffer has size \(MemoryLayout<ByteBuffer>.size) which is larger than the built-in storage of the existential containers.")
+	#endif
         self.malloc = hookedMalloc
         self.realloc = hookedRealloc
         self.free = hookedFree
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -379,12 +379,7 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
 
     public func decoderRemoved(ctx: ChannelHandlerContext) {
         // Remove the stored reference to ChannelHandlerContext
-        #if arch(arm) // 32-bit
-            let veryDeadBeef : UInt = 0xdeadbeef
-            parser.data = UnsafeMutableRawPointer(bitPattern: veryDeadBeef)
-        #else
-            parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
-        #endif
+        parser.data = UnsafeMutableRawPointer(bitPattern: 0xdeadbeef as UInt)
 
         // Set the callbacks to nil as we dont need these anymore
         settings.on_body = nil
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -380,9 +380,9 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
     public func decoderRemoved(ctx: ChannelHandlerContext) {
         // Remove the stored reference to ChannelHandlerContext
         #if arch(arm) // 32-bit
-	    let veryDeadBeef : UInt = 0xdeadbeef
+            let veryDeadBeef : UInt = 0xdeadbeef
             parser.data = UnsafeMutableRawPointer(bitPattern: veryDeadBeef)
-	#else
+        #else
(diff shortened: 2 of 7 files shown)


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -379,7 +379,12 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
 
     public func decoderRemoved(ctx: ChannelHandlerContext) {
         // Remove the stored reference to ChannelHandlerContext
-        parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
+        #if arch(arm) // 32-bit
+            let veryDeadBeef : UInt = 0xdeadbeef
+            parser.data = UnsafeMutableRawPointer(bitPattern: veryDeadBeef)
+        #else
+            parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
+        #endif
```
Lukasa: @jw Do we need this change? I think we only wanted this pointer value to be easy to spot in debuggers and core dumps: does it need to have the `deadbeef` portion shifted left 4 bits on 64-bit operating systems or is it ok to leave it in the least significant 32 bits?

Lukasa:
I have...mixed feelings about this.

For NIO's major server-side use-cases 32-bit architectures are not compelling to target. While the convenience of running on a Raspberry Pi is nice, it's certainly not a feature I feel compelled to support.

Supporting it would be easier if it we had CI for 32-bit environments, but currently we don't, so the risk of regression here is high. I think we probably need to decide as a team whether we care about supporting 32-bit environments at all: if we do, then we can merge this for 2.0, and if we don't, we can skip it.

helje5:
I think you should treat this similar to Swift itself - no official support at all, but accept community PRs which make it work.

helje5:
BTW: Why do you think this needs a major version bump?

Sources/NIOHTTP1/HTTPDecoder.swift
weissi: we can just use a 32 bit value `0xdeadbeef` is totally fine by me

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -86,8 +86,10 @@ public struct ByteBufferAllocator {
                   hookedRealloc: @escaping @convention(c) (UnsafeMutableRawPointer?, Int) -> UnsafeMutableRawPointer?,
                   hookedFree: @escaping @convention(c) (UnsafeMutableRawPointer?) -> Void,
                   hookedMemcpy: @escaping @convention(c) (UnsafeMutableRawPointer, UnsafeRawPointer, Int) -> Void) {
-        assert(MemoryLayout<ByteBuffer>.size <= 3 * MemoryLayout<Int>.size,
-               "ByteBuffer has size \(MemoryLayout<ByteBuffer>.size) which is larger than the built-in storage of the existential containers.")
+        #if !arch(arm) // only complain on 64-bit, this is unfortunate reality on 32-bit
+            assert(MemoryLayout<ByteBuffer>.size <= 3 * MemoryLayout<Int>.size,
```
weissi: I think we can just kill this assertion. We now have much better tests (allocation counters) which will blow up if this assertion would fail

weissi:
@helje5 / @Lukasa we could just make a hack and only make it `Int64` for 32 bit platforms. I know that's not how it's meant to be used but I'd be okay with that.

Lukasa:
@weissi Any reason not to just make it an Int64 but provide deprecated compat shims to the old type, so that it can still be accessed and set via Int?

Sources/NIO/ByteBuffer-core.swift
helje5: I assume there is no chance to make this work better on 32-bit?

Sources/NIOHTTP1/HTTPDecoder.swift
helje5: That still needs the UInt32 cast though, because parser.data seems to be Int32.

helje5:
TBH this kind of code doesn't feel quite right in the first place when you have your own TimeAmount abstraction:

```
Int(DispatchTime.now().uptimeNanoseconds) + amount.nanoseconds
```

This code probably belongs into a new `Timestamp` type, and/or the `TimeAmount`.

chnmrc:
Mqtt broker on top of this
https://www.netiot.com/netpi/industrial-raspberry-pi-3/

futurejones:
> Supporting it would be easier if it we had CI for 32-bit environments, but currently we don't, so the risk of regression here is high. I think we probably need to decide as a team whether we care about supporting 32-bit environments at all.

As Swift is supporting 32-bit environments I think it is important that swift-nio also supports 32-bit. 

We now have CI for Armv7 32-bit environments.
https://swift.org/blog/swift-community-hosted-ci/

Sources/NIO/ByteBuffer-core.swift
weissi: @helje5 the existential buffer size is 3 words. On 32 bit that means 12 bytes. Given that byte buffer needs one pointer, we’re left with 8 bytes for reader index, writer index, slice begin and end. I don’t think it’s feasible. So unfortunately for 32bit we’ll need to be larger than the builtin storage of an existential. But we still have the byte buffer special case in NIOAny so not all too bad...

Sources/NIOHTTP1/HTTPDecoder.swift
weissi: @helje5 isn’t parser.data a pointer?

helje5:
I will do the #if fix for the 1.x line. Neat shims are a little harder due to the nanoseconds property. In a different PR we should pull out the time calculations out of the main code base into TimeAmount, but let’s just go with the #if arm for this PR

Sources/NIOHTTP1/HTTPDecoder.swift
helje5: Right it is. I guess the problem is the Swift type overloading again. It will treat 0xdeadbeef as an `Int`. At least that's the error w/o the explicit UInt (cannot represent Integer literal as Int or something like that).

helje5:
@weissi @Lukasa PTAL, I added a typealias which should make it SemVer minor. OK with that?

BTW: Small hint: You don't need an actual Raspi to test a Raspi build! If you have Docker4Mac, it has QEmu included and can run simple Raspi images. To build NIO from within a NIO checkout:
```
docker run --rm \
		-v $(pwd):/src \
		-v $(pwd)/.build-linux-rpi-4.1.0:/src/.build \
		helje5/rpi-swift-dev:4.1.0 \
		bash -c "cd /src && swift build -c release"
```
Takes about 2.5 mins on my MacPro.

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Sure, but can't we just compress this into one line? `parser.data = UnsafeMutableRawPointer(bitPattern: UInt(0xdeadbeef)`? That'll work correctly on all platforms.

weissi: @helje5 I'm happy with this except for this

normanmaurer:

PR #385, opened 2018-05-02
Fix possible header / URI corruption when discardReadBytes() is calle…

normanmaurer:
…d in HTTPDecoder

Motivation:

When discardReadBytes() was called and we still did not pass the Head to the user it was quite possible that the headers / URI could be corrupted as the stored readerIndex did not matchup anymore.

Modifications:

- Override most of the functionality of ByteToMessageDecoder in HTTPDecoder to better work with the provided state machine of http_parser
- Remove allocations by removing the pendingInOut Array as its not needed anymore.
- Add guards against re-entrance calls
- Add unit test that shows that everything works as expected now.

Result:

Fixed bug and reduced allocations (8% - 10% perf win).


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -82,6 +84,7 @@ private struct HTTPParserState {
             let (index, length) = consumeSlice()
             self.currentStatus = self.cumulationBuffer!.getString(at: index, length: length)!
         case .body:
+            self.currentNameIndex = nil
```
Lukasa: Can we ever hit this branch without having gone through `.headerValue`? It seems like should be an assertion instead.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
                 fatalError("the impossible happened: handler neither a HTTPRequestDecoder nor a HTTPResponseDecoder which should be impossible")
             }
             return 0
         }
     }
 
-    public func decoderRemoved(ctx: ChannelHandlerContext) {
-        // Remove the stored reference to ChannelHandlerContext
-        parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
+    private var decoding: Bool = false
```
Lukasa: Can we move this up to the other state declarations rather than hiding it just above `channelRead`?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-        settings.on_url = nil
-        settings.on_status = nil
-        settings.on_chunk_header = nil
-        settings.on_chunk_complete = nil
-        settings.on_header_field = nil
-        settings.on_header_value = nil
-        settings.on_message_begin = nil
-    }
+        // Guard against re-entrance calls of channelRead(...)
+        guard !decoding else {
```
Lukasa: Explicit `self` please. :smile:

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+            return
+        }
 
-    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
-        if let slice = state.slice {
-            // If we stored a slice before we need to ensure we move the readerIndex so we don't try to parse the data again. We
-            // also need to update the reader index to whatever it is now.
-            state.slice = (buffer.readerIndex, slice.length)
-            buffer.moveReaderIndex(forwardBy: state.readerIndexAdjustment)
+        // Needed to guard again re-entrace calls.
```
Lukasa: s/entrace/entrant/

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                //       What we would need to do is basically keep track of the start of a message and discard everything that we already forwarded through the pipeline.
+                break
             }
-            return result
+        } catch {
+            self.cumulationBuffer = nil
+            ctx.fireErrorCaught(error)
+            ctx.close(promise: nil)
         }
+    }
```
Lukasa: This method's 80 lines long, can we factor it into some smaller sub-functions?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                assert(result == bufferSlice.readableBytes)
+
+                // Update readerIndex of the cumulationBuffer itself as we will refetch it in the next loop run if needed.
+                self.cumulationBuffer!.moveReaderIndex(forwardBy: result)
+
+                if bufferSlice.readableBytes == self.cumulationBuffer!.readableBytes + result {
+                    // There was no change to the readable bytes of the cumulationBuffer by a re-entrant call of channelRead. Its safe to break the loop.
+                    break
+                }
+            } while true
```
Lukasa: `repeat...while true` is a strange idiom. I know you didn't add it, but can we swap this to just `while true { }`?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+
+                    let httpError = parser.http_errno
+                    if httpError != 0 {
+                        self.state.currentError = HTTPParserError.httpError(fromCHTTPParserErrno: http_errno(rawValue: httpError))!
+                        throw self.state.currentError!
+                    }
+                    return result
+                }
+
+                guard self.cumulationBuffer != nil else {
```
Lukasa: Would it be better just to not `nil` the buffer out on channel closure? It'll simplify the code here, and the buffer will get dropped anyway, one way or another.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                    return
+                }
+
+                /// http_parser_execute(...) should always consume the whole buffer.
+                assert(result == bufferSlice.readableBytes)
+
+                // Update readerIndex of the cumulationBuffer itself as we will refetch it in the next loop run if needed.
+                self.cumulationBuffer!.moveReaderIndex(forwardBy: result)
+
+                if bufferSlice.readableBytes == self.cumulationBuffer!.readableBytes + result {
```
Lukasa: Given that we have an assertion above that `bufferSlice.readableBytes  == result`, and then the next line is `self.cumulationBuffer!.moveReaderIndex(forwardBy: result)`, I think this conditional is just a very complex way of spelling `if self.cumulationBuffer!.readableBytes == 0`.

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Actually, no, let's use `while var bufferSlice = self.cumulationBuffer, bufferSlice.readableBytes > 0`. That will cover your loop exit condition at the bottom and also create the buffer slice we need.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +376,120 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-            state.readerIndexAdjustment = buffer.readableBytes
-            state.slice = (-1, slice.length)
-            return .needMoreData
-        } else {
-            buffer.moveReaderIndex(forwardBy: result)
-            state.readerIndexAdjustment = 0
-            return .continue
-        }
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        return DecodingState.needMoreData
```
Lukasa: Can you add a code comment here indicating that we don't actually expect this function to be called?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -486,8 +533,14 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         // be totally safe to call this with a null data pointer. Just to make sure, though,
         // let's store nil in the base address. Once we've called this, we've seenEOF: don't
         // let us enter this function again.
-        self.state.baseAddress = nil
         self.state.seenEOF = true
+
+        guard !self.decoding else {
+            // We are currently decoding, return early as we will handle it in the decoding loop.
```
Lukasa: This comment is wrong: we don't handle the EOF in the decoding loop at all. We need to explicitly call `c_nio_http_parser_execute` with a length of `0`, which the other code never does. You may want to add a test that triggers this case to validate that it doesn't work, and then we can fix it. :smile:

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: yep sounds correct... same for the slice imho

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: yep... refactored to use your while loop :)

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: I can do this but this would also need changed in `ByteToMessageDecoder` and may increase memory usage if a user still hold a reference to the handler after the channel become inactive / the handler was removed.

WDYT ?

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: done

Package.swift
```diff
@@ -51,6 +51,8 @@ var targets: [PackageDescription.Target] = [
             dependencies: ["NIO", "NIOHTTP1", "CNIOSHA1"]),
     .target(name: "NIOWebSocketServer",
             dependencies: ["NIO", "NIOHTTP1", "NIOWebSocket"]),
+    .target(name: "NIOHTTP1HelloWorldServer",
+            dependencies: ["NIO", "NIOHTTP1", "NIOConcurrencyHelpers"]),
```
Lukasa: Any reason this isn't just part of `NIOHTTP1Server`?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +377,129 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
-        if let slice = state.slice {
-            // If we stored a slice before we need to ensure we move the readerIndex so we don't try to parse the data again. We
-            // also need to update the reader index to whatever it is now.
-            state.slice = (buffer.readerIndex, slice.length)
-            buffer.moveReaderIndex(forwardBy: state.readerIndexAdjustment)
-        }
+                let result = state.baseAddress!.withMemoryRebound(to: Int8.self, capacity: pointer.count, { p in
+                    c_nio_http_parser_execute(&parser, &settings, p.advanced(by: bufferSlice.readerIndex), bufferSlice.readableBytes)
+                })
```
Lukasa: We can use the trailing closure syntax instead of including the closure in the parenthesis.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +377,129 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
 
-            if let error = state.currentError {
-                throw error
+                let httpError = parser.http_errno
+                if httpError != 0 {
+                    self.state.currentError = HTTPParserError.httpError(fromCHTTPParserErrno: http_errno(rawValue: httpError))!
+                    throw self.state.currentError!
+                }
+                return result
             }
```
Lukasa: This block seems wildly too long to me: do we need the pointer to be valid for this long?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +377,129 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-        settings.on_chunk_complete = nil
-        settings.on_header_field = nil
-        settings.on_header_value = nil
-        settings.on_message_begin = nil
-    }
+    // Decode HTTP until there is nothing more to decode.
+    private func decodeHTTP(ctx: ChannelHandlerContext) throws {
+        // We need to refetch the cumulationBuffer on each loop as it may has changed due re-entrance calls of channelRead(...)
+        while let bufferSlice = self.cumulationBuffer, bufferSlice.readableBytes > 0 {
+            let result = try bufferSlice.withVeryUnsafeBytes { (pointer) -> size_t in
```
Lukasa: Why does this use `withVeryUnsafeBytes` and then manipulate the pointer, instead of using `withUnsafeReadableBytes`?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +377,129 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                    self.state.currentError = HTTPParserError.httpError(fromCHTTPParserErrno: http_errno(rawValue: httpError))!
+                    throw self.state.currentError!
+                }
+                return result
             }
 
-            state.baseAddress = nil
+            guard !self.state.seenEOF else {
+                // We need to notify the parser about the EOF as we received it while in http_parser_excecute.
+                self.notifyParserEOF(ctx: ctx)
```
Lukasa: I think this notifies too early. While there are still bytes in the cumulation buffer, we need to parse them. This should be outside the `while` loop.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -358,91 +377,129 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                break
+            }
 
-            let httpError = parser.http_errno
-            if httpError != 0 {
-                self.state.currentError = HTTPParserError.httpError(fromCHTTPParserErrno: http_errno(rawValue: httpError))!
-                throw self.state.currentError!
+            guard self.cumulationBuffer != nil else {
```
Lukasa: Rather than have this guard, just change the movement of the reader index on line 439 to `self.cumulationBuffer?.moveReaderIndex(forwardBy: result)`. That will cover the buffer being `nil`d out.

Package.swift
normanmaurer: it should not be part of the pr at all... included by mistake.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: nope let me move some stuff outside.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: +1

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: because it makes things easier when we calculate the readerIndex based on the baseAddress.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: makes sense

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: good idea

normanmaurer:
@weissi @Lukasa PTAL again

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -185,15 +187,23 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
     fileprivate var state = HTTPParserState()
 
     fileprivate init(type: HTTPMessageT.Type) {
         /* this is a private init, the public versions only allow HTTPClientResponsePart and HTTPServerRequestPart */
         assert(HTTPMessageT.self == HTTPClientResponsePart.self || HTTPMessageT.self == HTTPServerRequestPart.self)
     }
 
+    deinit {
+        // Remove the stored reference to ChannelHandlerContext
+        self.parser.data = UnsafeMutableRawPointer(bitPattern: 0x0000deadbeef0000)
```
weissi: to make @helje5's life easier and not cause needless merge conflicts, could you make this value `0xdeadbeef`? That way it'll work on 32-bit platforms too

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +368,131 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+                    c_nio_http_parser_execute(&self.parser, &self.settings, p.advanced(by: bufferSlice.readerIndex), bufferSlice.readableBytes)
+                }
+            }
+            try self.rethrowParserError()
+
+            /// http_parser_execute(...) should always consume the whole buffer.
+            assert(result == bufferSlice.readableBytes)
+
+            // Update readerIndex of the cumulationBuffer itself as we will refetch it in the next loop run if needed.
+            self.cumulationBuffer?.moveReaderIndex(forwardBy: result)
```
Lukasa: It's *marginally* faster to use `moveReaderIndex(to: writerIndex)` I think.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +368,131 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
         }
 
-        let result = try buffer.withVeryUnsafeBytes { (pointer) -> size_t in
-            state.baseAddress = pointer.baseAddress!.assumingMemoryBound(to: UInt8.self)
+        if self.state.seenEOF {
+            // We need to notify the parser about the EOF as we received it while in http_parser_excecute.
```
Lukasa: Nit: `http_parser_execute`.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: This will mess up with the loop that handles re-entrance.

normanmaurer:
@weissi @Lukasa PTAL again

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +377,174 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+        case .body, .messageBegin:
+            assert(self.state.currentNameIndex == nil)
+            assert(self.state.currentHeaders.isEmpty)
+            assert(self.state.slice == nil)
+
+            // Its safe to just drop the cumulationBuffer as we not have any extra views into it that are represented as readerIndex / length.
+            self.cumulationBuffer = nil
+
+        case .headerField, .headerValue:
+            if let headerStartIdx = self.state.headerStartIndex {
```
Lukasa: The whole body of this case is in the `if` block, so this may as well be `guard`.

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Ah shoot you're right. Nevermind then.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +377,174 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-            let result = state.baseAddress!.withMemoryRebound(to: Int8.self, capacity: pointer.count, { p in
-                c_nio_http_parser_execute(&parser, &settings, p.advanced(by: buffer.readerIndex), buffer.readableBytes)
-            })
+            // Its safe to just drop the cumulationBuffer as we not have any extra views into it that are represented as readerIndex / length.
+            self.cumulationBuffer = nil
 
-            if let error = state.currentError {
-                throw error
+        case .headerField, .headerValue:
+            guard let headerStartIdx = self.state.headerStartIndex else {
```
Lukasa: Is it valid to be in this state *without* a `headerStartIndex`?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +377,174 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-            // *anyway*, so we want to make sure we can see the bad value in debug errors.
-            buffer.moveReaderIndex(to: slice.readerIndex)
-            state.readerIndexAdjustment = buffer.readableBytes
-            state.slice = (-1, slice.length)
-            return .needMoreData
+        }
+    }
+
+    /// Will discard bytes till readerIndex if its needed and then call `fn`.
+    private func mayDiscardDecodedBytes(readerIndex: Int, _ fn: () -> Void) {
```
Lukasa: I think that the "readerIndex" label could be improved: maybe "upTo" so that this now reads `mayDiscardDecodedBytes(upTo:)`.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -357,92 +377,174 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+        self.cumulationBuffer!.moveReaderIndex(to: readerIndex)
+        if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) && self.cumulationBuffer!.discardReadBytes() {
+            // When discardReadBytes() returns true the readerIndex must be 0 again.
+            assert(self.cumulationBuffer!.readerIndex == 0)
+
+            fn()
+        }
+
+        self.cumulationBuffer!.moveReaderIndex(to: self.cumulationBuffer!.writerIndex)
```
Lukasa: Should we assert that at the start of this function the reader index was equal to the writer index?

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: @Lukasa yes as we reset it to nil if we discarded the decoded bytes.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: sure

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Is there any reason not to set it to 0 instead and avoid this branch?

normanmaurer: Yes as otherwise we will try to increase the writerIndex etc later on which is even more expensive.

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: This was done

Lukasa:
Ok, this looks good to me.

PR #387, opened 2018-05-03
don't crash if close is called when close fails

weissi:
Motivation:

@vlm hit an interesting situation which is very likely the sign of
another bug that we have yet to find:

During a close, @vlm hit an error in close which lead to a user callout
which then closed again.

The immediate fix is simple (this PR) is as always not to call out to
the user before reconciling our own state. But hitting this bug also
means that either a `socket.close` or a `selectableEventLoop.deregister`
failed as we only called out in those situations. That definitely hides
a condition that is hard to imagine without another bug that we still
need to find.

Modifications:

in `BaseSocketChannel.close0` follow rule 0 and don't call out before
reconciling state.

Result:

fewer crashes, less potential to hide other bugs.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -664,11 +664,18 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
             return
         }
 
+        // === BEGIN: No user callouts ===
+
+        // this is to register all error callouts as all the callouts must happen after we transition out state
+        var errorCallouts: [(ChannelPipeline) -> Void] = []
+
         self.interestedEvent = .reset
         do {
             try selectableEventLoop.deregister(channel: self)
         } catch let err {
-            pipeline.fireErrorCaught0(error: err)
+            errorCallouts.append { pipeline in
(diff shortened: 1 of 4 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -664,33 +664,51 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
...
         }
 
         // Transition our internal state.
         let callouts = self.lifecycleManager.close(promise: p)
 
+        // === END: No user callouts (now that our state is reconciled, we can call out to user code.) ===
+
+        // this must be the first to call out as it transitions the PendingWritesManager into the closed state
+        // and we assert elsewhere that the PendingWritesManager has the same idea of 'open' as we have in here.
+        self.cancelWritesOnClose(error: error)
```
weissi: just to point it out: this is a slight semantic change: This method will call out to the user. We used to do that later but we can't do that because this reconciles the PWM's `isOpen` and BaseSocketChannel's `isOpen` and they need to always be in sync (there are assertions for that). Mid-term I want to get rid of PWM's isOpen but I have a lot of code to read to be confident that we actually don't need it...

PR #388, opened 2018-05-03
implement lazy selector registration

### LINKED ISSUE #380: implement selector late registration
With `epoll` we cannot register an fd in a selector that didn't have `bind`/`connect` called on it. Therefore we should really implement late registration where we actually only register when `connect`/`bind` is called.
That way the behaviour on Darwin & Linux is the same.

weissi:
Motivation:

Previously we would eagerly register any fd with the selector
(epoll/kqueue) which works well with kqueue. With epoll however, you'll
get `EPOLLHUP` immediately if you supplied a socket that hasn't had
connect/bind called on it.
We got away with this because we took quite a bit of care to always make
sure we call connect/bind pretty much immediately after we registered
it.  And crucially before we ever asked the selector to tell us about
new events.

Modifications:

made selector registration lazy (when we try activating the channel by
calling connect/bind)

Result:

the user can now register and connect whenever they feel like it, fixes #380

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -18,14 +18,16 @@ private struct SocketChannelLifecycleManager {
     // MARK: Types
     private enum State {
         case fresh
-        case registered
+        case preRegistered // register() has been run but the selector doesn't know about it yet
+        case fullyRegistered // fully registered, ie. the selector knows about it
         case activated
         case closed
     }
 
     private enum Event {
         case activate
-        case register
+        case beginRegistration
+        case finishRegistration
         case close
     }
 
@@ -66,8 +68,13 @@ private struct SocketChannelLifecycleManager {
     }
 
     @inline(__always) // we need to return a closure here and to not suffer from a potential allocation for that this must be inlined
-    internal mutating func register(promise: EventLoopPromise<Void>?) -> ((ChannelPipeline) -> Void) {
-        return self.moveState(event: .register, promise: promise)
+    internal mutating func beginRegistration(promise: EventLoopPromise<Void>?) -> ((ChannelPipeline) -> Void) {
+        return self.moveState(event: .beginRegistration, promise: promise)
+    }
+
+    @inline(__always) // we need to return a closure here and to not suffer from a potential allocation for that this must be inlined
+    internal mutating func finishRegistration(promise: EventLoopPromise<Void>?) -> ((ChannelPipeline) -> Void) {
+        return self.moveState(event: .finishRegistration, promise: promise)
     }
 
     @inline(__always) // we need to return a closure here and to not suffer from a potential allocation for that this must be inlined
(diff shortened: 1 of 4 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -18,14 +18,16 @@ private struct SocketChannelLifecycleManager {
     // MARK: Types
     private enum State {
         case fresh
-        case registered
+        case lazilyRegistered // register() has been run but the selector doesn't know about it yet
+        case fullyRegistered // fully registered, ie. the selector knows about it
```
Lukasa: Nit: I don't love these names. Can we call the new state `preRegistered`, maybe?

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -18,14 +18,16 @@ private struct SocketChannelLifecycleManager {
...
         case activated
         case closed
     }
 
     private enum Event {
         case activate
-        case register
+        case registerLazily
+        case registerFully
```
Lukasa: Similarly, these names aren't necessarily right. Maybe `beginRegistration` and `completeRegistration`?

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -731,27 +760,34 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
...
     public final func registerAlreadyConfigured0(promise: EventLoopPromise<Void>?) {
         assert(self.eventLoop.inEventLoop)
         assert(self.isOpen)
         assert(!self.lifecycleManager.isActive)
         register0(promise: nil)
-        becomeActive0(promise: promise)
+        if self.lifecycleManager.isRegisteredLazily {
+            try! becomeFullyRegistered0()
+            if self.lifecycleManager.isRegisteredFully {
+                becomeActive0(promise: promise)
```
Lukasa: Do these callouts not get triggered by `register0`?

weissi: @Lukasa right now `register0` only does the pre-registration. I didn't change its name as it's on `ChannelCore`. But we could obviously rename it to `preRegister0`

Sources/NIO/EventLoop.swift
```diff
@@ -337,6 +337,11 @@ internal final class SelectableEventLoop: EventLoop {
         _addresses.deallocate()
     }
 
+    /// Is this `SelectableEventLoop` still open (ie. not shutting down or shut down)
+    public var isOpen: Bool {
```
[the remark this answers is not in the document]
weissi: @normanmaurer it's fine as `SelectableEventLoop` is an `internal` type (not public) but regardless I changed this to `internal` and added an assertion

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2317,6 +2317,43 @@ public class ChannelTests: XCTestCase {
 
     }
 
+    func testLazyRegistrationWorksForServerSockets() throws {
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
```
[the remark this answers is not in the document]
weissi: done

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2317,6 +2317,43 @@ public class ChannelTests: XCTestCase {
...
+        XCTAssertNoThrow(try server.eventLoop.submit {
+            XCTAssertFalse(server.isActive)
+            }.wait())
+        XCTAssertEqual(0, server.localAddress!.port!)
+        XCTAssertNoThrow(try server.bind(to: SocketAddress(ipAddress: "0.0.0.0", port: 0)).wait())
+        XCTAssertNotEqual(0, server.localAddress!.port!)
+    }
+
+    func testLazyRegistrationWorksForClientSockets() throws {
+        let group = MultiThreadedEventLoopGroup(numThreads: 1)
```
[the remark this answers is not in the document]
weissi: done

normanmaurer:
Thanks @weissi ... merged :)

PR #395, opened 2018-05-08
Make hopTo(eventLoop:) public.

Lukasa:
Motivation:

When this method was originally introduced it was also intended as a
helper for external users, but we never made it public.

Modifications:

Make hopTo(eventLoop:) public.

Result:

Users will have an easier time managing their threads.


helje5:
should that be `hop(to eventLoop:)`?

Lukasa:
I'm inclined to leave the name as it is.

PR #396, opened 2018-05-08
add performance tester

weissi:
Motivation:

NIO wants to be a highly performant library and therefore we need
performance tests.

Modifications:

Add a couple of performance tests.

Result:

Easier to talk about perf improvements/regressions if we use similar
benchmarks.


Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,504 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOHTTP1
+import NIOFoundationCompat
```
Lukasa: This dependency should be marked in the `Package.swift` as well.

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,504 @@
...
+
+public func measure(_ fn: () throws -> Int) rethrows -> [TimeInterval] {
+    func measureOne(_ fn: () throws -> Int) rethrows -> TimeInterval {
+        let start = Date()
+        _ = try fn()
+        let end = Date()
+        return end.timeIntervalSince(start)
+    }
+    
+    _ = try measureOne(fn) /* pre-head and throw away */
```
Lukasa: Nit: heat, not head.

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,504 @@
...
+            channel.pipeline.add(handler: SimpleHTTPServer())
+        }
+    }.bind(host: "127.0.0.1", port: 0).wait()
+
+defer {
+    try! serverChannel.close().wait()
+}
+
+var head = HTTPRequestHead(version: HTTPVersion(major: 1, minor: 1), method: .GET, uri: "/perf-test-1")
+head.headers.add(name: "Host", value: "apple.com")
```
Lukasa: As a pretty minor nit, maybe we should use `localhost` instead of `apple.com`.

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,504 @@
...
+        precondition([0x41] == arr!, "\(arr!)")
+        let str = buffer.readString(length: 1)
+        precondition("A" == str, "\(str!)")
+    }
+    for _ in 0 ..< 1024*1024 {
+        doWrites(buffer: &buffer)
+        doReads(buffer: &buffer)
+    }
+    return buffer.readableBytes
+}
```
Lukasa: This test is identical to the allocation one: is there a way we can share the code, rather than duplicate it?

weissi: @Lukasa it's not exactly identical (IIRC 1000 runs vs. 1024*1024 runs). Sharing code there seems pretty hard unfortunately :|

weissi:
tests failed to fetch zlib

```
17:01:26 error: terminated(1): /root/.swift/usr/bin/swift-build-tool -f /code/.build/debug.yaml test
17:01:26 Fetching https://github.com/apple/swift-nio-zlib-support.git
17:01:26 Cloning https://github.com/apple/swift-nio-zlib-support.git
17:01:26 Resolving https://github.com/apple/swift-nio-zlib-support.git at 1.0.0
17:01:26 Build step 'Execute shell' marked build as failure
17:01:26 $ ssh-agent -k
```

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,505 @@
...
+                                             X-Some-Header-8: foo\r\n\r\n
+                                             """)
+        }
+
+        func write(ctx: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
+            var buf = self.unwrapOutboundIn(data)
+            if self.expectedResponseBuffer == nil {
+                self.expectedResponseBuffer = buf
+            }
+            precondition(buf == self.expectedResponseBuffer, "got \(buf.readString(length: buf.readableBytes)!)")
```
normanmaurer: Should this be an `assert`?

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,505 @@
...
+        channel.pipeline.add(handler: measuringHandler, first: true)
+    }.wait()
+
+    measuringHandler.kickOff(channel: channel)
+
+    while !done {
+        eventLoop.run()
+    }
+    _ = try channel.finish()
+    precondition(requestsDone == 10_000)
```
normanmaurer: Should this be an `assert`?

Sources/NIOPerformanceTester/main.swift
```diff
@@ -0,0 +1,505 @@
...
+        precondition(0x41 == sliceVal, "\(sliceVal!)")
+        buffer.withUnsafeReadableBytes { ptr in
+            precondition(ptr[0] == 0x41)
+        }
+
+        /* those down here should be one allocation each */
+        let arr = buffer.readBytes(length: 1)
+        precondition([0x41] == arr!, "\(arr!)")
+        let str = buffer.readString(length: 1)
+        precondition("A" == str, "\(str!)")
```
normanmaurer: Should this be an `assert`?

Sources/NIOPerformanceTester/main.swift
weissi: @normanmaurer no because we want to run it in release mode (as the tester is basically only ever run in release mode)

Sources/NIOPerformanceTester/main.swift
weissi: same

Sources/NIOPerformanceTester/main.swift
weissi: same

PR #397, opened 2018-05-08
fix minor unsafe pointer type assumption issue

weissi:
Motivation:

In a test we did `assumingMemoryBound(to: in6_addr.self)` for something
that was actually bound to `UInt8`. So we should use
`bindMemory(to: in6_addr.self, count: 1)` instead

Modifications:

replace one `assumingMemoryBound` by `bindMemory`

Result:

more memory correctness is good even for tests


normanmaurer:
late LGTM

PR #399, opened 2018-05-09
Channel quiescing support

weissi:
Motivation:

In certain cases it's useful to quiesce a channel instead of just
closing them immediately for example when receiving a signal.
This lays the groundwork by introducing the
`ChannelShouldQuiesceUserEvent` user event that when received can be
interpreted by a ChannelHandler in a protocol & application specific
way. Some protocols support tear down and that would be a good place to
initiate the tear down.

Modifications:

- introduce `ChannelShouldQuiesceUserEvent`
- handle `ChannelShouldQuiesceUserEvent` in the `AcceptHandler` with
  closing the server socket
- handle `ChannelShouldQuiesceUserEvent` in the
  `HTTPServerPipelineHandler` by only handling a already in-flight
  request and then no longer accepting input

Result:

- handlers can now support quiescing


Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -227,11 +290,30 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler {
         }
 
         var startReadingAgain = false
-        if case .end = self.unwrapOutboundIn(data) {
+
+        switch self.unwrapOutboundIn(data) {
+        case .head(var head) where self.lifecycleState != .acceptingEvents:
```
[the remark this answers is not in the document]
weissi: @Lukasa I think `!= .acceptingEvents` which is the same as saying  `.quiescing || .noLongerAcceptingEvents` is correct. The reason is that we should add a `connection: close` header whenever we're quiescing. The `.quescing` case however means that we're continuing to accept request parts (like `.body` and `.end`). `.noLongerAcceptingEvents` means that we're quiescing and we have already seen the request `.end`.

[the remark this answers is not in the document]
weissi: I renamed the cases to `.quiescingWaitingForRequestEnd` and `.quiescingLastRequestEndReceived` which should make this clearer

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -76,26 +76,36 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler {
...
 
         /// The server has responded early, before the request has completed. We need
         /// to wait for the request to complete, but won't block anything.
         case requestEndPending
 
+        mutating func requestHeadReceived() {
+            switch self {
+            case .idle:
+                self = .requestAndResponseEndPending
+            default:
```
Lukasa: Might be better to explicitly enumerate all the states here?

BasThomas: Or use an `if case .idle = self`, as it seems just one case is evaluated.

weissi: @BasThomas we want to crash in the other cases; @Lukasa ok

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -456,4 +476,238 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+                       [.channelRead(HTTPServerRequestPart.head(self.requestHead)),
+                        .channelRead(HTTPServerRequestPart.end(nil))])
+
+        XCTAssertFalse(self.channel.isActive)
+        self.channel = nil
+    }
+
+
+
+
```
Lukasa: Bunch of whitespace down here.

Sources/NIO/Channel.swift
```diff
@@ -362,3 +362,13 @@ public enum ChannelEvent: Equatable {
     /// Output portion of the `Channel` was closed.
     case outputClosed
 }
+
+/// A `Channel` user event that is sent when the `Channel` has been asked to quiesce.
+///
+/// The action(s) that should be taken after receiving this event are both application and protocol dependent. If the
+/// protocol supports a notion of requests and responses, it might make sense to stop accepting new requests but finish
+/// processing the request currently in flight.
+public struct ChannelShouldQuiesceUserEvent {
```
normanmaurer: nit: consider using an enum ?

Sources/NIO/CircularBuffer.swift
```diff
@@ -204,6 +204,13 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
         return nextIndex
     }
 
+    /// Removes all members from the circular buffer whist keeping the capacity.
+    public mutating func removeAll() {
```
normanmaurer: In swift in other places we have `removeAll(keepingCapacity:)`. Maybe we should use the same method naming and behaviour ?

https://developer.apple.com/documentation/swift/array/1641473-removeall

Sources/NIO/Channel.swift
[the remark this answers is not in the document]
weissi: I followed `TLSUserEvent` but there's also `ChannelEvent` so there's prior art for both. I don't mind which one. @Lukasa ?

weissi: @normanmaurer we need to be able to instanciate that one. So I could do an enum but what would the cases be? I could only think of one case "please quiesce the channel" and I thought that's easier expressed by just using a type. But happy to change. We can't use `ChannelEvent` because it's an `enum` and extending it would be a major version break

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -191,6 +191,29 @@ extension HTTPRequestHead {
     }
 }
 
+extension HTTPResponseHead {
+    /// Whether this HTTP response is a keep-alive request: that is, whether the
+    /// connection should remain open after the request is complete.
+    public var isKeepAlive: Bool {
```
normanmaurer: I think we should implement it kind of the same way as here:

https://github.com/apple/swift-nio/pull/402/files#diff-9361bcb5608a311475f3a33eeadc3a3eR300

weissi: @normanmaurer ok, will do that tomorrow. Also the 'wrong' impl first?

Sources/NIO/CircularBuffer.swift
weissi: will do tomo

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: @weissi let me just merge my pr and then you can reuse the `headers.isKeepAlive`

Sources/NIO/Channel.swift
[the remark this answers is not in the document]
Lukasa: I think `Event` is better, I think the `TLSUserEvent` name wasn't great.

Tests/NIOTests/IntegerTypesTest.swift
```diff
@@ -0,0 +1,84 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
```
weissi: note to @Lukasa / @normanmaurer : for a second i thought `nextPowerOf2` had a bug but couldn't find the tests so added some

Sources/NIO/Channel.swift
```diff
@@ -362,3 +362,13 @@ public enum ChannelEvent: Equatable {
...
 }
+
+/// A `Channel` user event that is sent when the `Channel` has been asked to quiesce.
+///
+/// The action(s) that should be taken after receiving this event are both application and protocol dependent. If the
+/// protocol supports a notion of requests and responses, it might make sense to stop accepting new requests but finish
+/// processing the request currently in flight.
+public struct ChannelShouldQuiesceEvent {
+    public init() {
+    }
```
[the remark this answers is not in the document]
Lukasa: Ah sorry, you're right.

Lukasa:
Gotta fix those tests up, but I'm happy.

PR #400, opened 2018-05-10
Explicitly define constants through Posix and import errno.h

alblue:
Explicitly define constants through Posix and import errno.h

### Motivation:

We implicitly depend on errno's `ENOENT` in Linux.swift, but don't explicitly import errno.h.

Similarly, in some cases the compiler can be confused as to where `SHUT_*` come from, so make them an exported value from `Posix` instead.

### Result:

Compiles successfully on Linux and Darwin.


normanmaurer:

Lukasa:

alblue:
Thanks for the excellently quick turnaround!

PR #402, opened 2018-05-14
Do not allocate String to check if HTTP message is keep alive

normanmaurer:
Motivation:

We can check if a HTTP message is using keep alive without any allocation.

Modifications:

Directly act on the internal storage of HTTPHeaders to find out if keep alive is used or not.

Result:

Less allocations

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -17,7 +17,6 @@ import NIO
 let crlf: StaticString = "\r\n"
 let headerSeparator: StaticString = ": "
 
-private let connectionUtf8 = "connection".utf8
 
 // Keep track of keep alive state.
 internal enum KeepAliveState {
@@ -298,15 +297,16 @@ private extension UInt8 {
         case .keepAlive:
             return true
         case .unknown:
-            guard let connection = self["connection"].first?.lowercased() else {
-                // HTTP 1.1 use keep-alive by default if not otherwise told.
-                return version.major == 1 && version.minor == 1
+            for header in self.headers {
+                if self.buffer.equalCaseInsensitiveASCII(view: "connection".utf8, at: header.name) {
+                    if self.buffer.equalCaseInsensitiveASCII(view: "close".utf8, at: header.value) {
+                        return false
+                    }
+                    return self.buffer.equalCaseInsensitiveASCII(view: "keep-alive".utf8, at: header.value)
+                }
             }
-            
-            if connection == "close" {
-                return false
-            }
-            return connection == "keep-alive"
+            // HTTP 1.1 use keep-alive by default if not otherwise told.
+            return version.major == 1 && version.minor >= 1
         }
     }
 }
@@ -415,7 +415,7 @@ public struct HTTPHeaders: CustomStringConvertible {
     }
     
     private func isConnectionHeader(_ header: HTTPHeaderIndex) -> Bool {
-         return self.buffer.equalCaseInsensitiveASCII(view: connectionUtf8, at: header)
+         return self.buffer.equalCaseInsensitiveASCII(view: "connection".utf8, at: header)
     }
     
     /// Add a header name/value pair to the block.


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -298,15 +300,18 @@ private extension UInt8 {
...
             return true
         case .unknown:
-            guard let connection = self["connection"].first?.lowercased() else {
-                // HTTP 1.1 use keep-alive by default if not otherwise told.
-                return version.major == 1 && version.minor == 1
-            }
-            
-            if connection == "close" {
-                return false
+            if !self.headers.isEmpty {
```
weissi: this seems superfluous as you iterate `self.headers` just after that. If this is empty, the loop won't do anything either.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -298,15 +300,18 @@ private extension UInt8 {
...
+                        if self.buffer.equalCaseInsensitiveASCII(view: closeUtf8, at: header.value) {
+                            return false
+                        }
+                        return self.buffer.equalCaseInsensitiveASCII(view: keepAliveUtf8, at: header.value)
+                    }
+                }
             }
-            return connection == "keep-alive"
+            // HTTP 1.1 use keep-alive by default if not otherwise told.
+            return version.major == 1 && version.minor == 1
```
weissi: this should be `version.minor >= 1` as RFC says to treat 1.x where x>=1 like 1.1

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -18,6 +18,8 @@ let crlf: StaticString = "\r\n"
 let headerSeparator: StaticString = ": "
 
 private let connectionUtf8 = "connection".utf8
+private let closeUtf8 = "close".utf8
```
weissi: I think these global variables do nothing (in the good case) and cause a slowdown if the compiler isn't able to optimise it. Just put them where you use them

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -298,15 +300,18 @@ private extension UInt8 {
...
-                // HTTP 1.1 use keep-alive by default if not otherwise told.
-                return version.major == 1 && version.minor == 1
-            }
-            
-            if connection == "close" {
-                return false
+            if !self.headers.isEmpty {
+                for header in self.headers {
+                    if self.buffer.equalCaseInsensitiveASCII(view: connectionUtf8, at: header.name) {
+                        if self.buffer.equalCaseInsensitiveASCII(view: closeUtf8, at: header.value) {
```
weissi: technically this isn't a correct implementation, right? could be `connection: keep-alive, foobar` no?

I know, it's not worse than the previous implementation either but the fast path is anyway that there is no `connection` header

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -18,6 +18,8 @@ let crlf: StaticString = "\r\n"
 let headerSeparator: StaticString = ": "
```
weissi: please change the allocation counter test limits and reduce according to the number of allocs saved

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: ok :)

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: @weissi hmm I would have thought its actually better as `.utf8` should only be called once ? Maybe I should make these static ?

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: @weissi I just preserved the old behaviour. @Lukasa thoughts ?

Sources/NIOHTTP1/HTTPTypes.swift
weissi: @normanmaurer `.utf8` is a function that needs to be inlined. If you want to cache the bytes straight, you'll need to do `Array("close".utf8)`

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -298,15 +297,16 @@ private extension UInt8 {
...
-            guard let connection = self["connection"].first?.lowercased() else {
-                // HTTP 1.1 use keep-alive by default if not otherwise told.
-                return version.major == 1 && version.minor == 1
-            }
-            
-            if connection == "close" {
-                return false
+            for header in self.headers {
+                if self.buffer.equalCaseInsensitiveASCII(view: "connection".utf8, at: header.name) {
+                    if self.buffer.equalCaseInsensitiveASCII(view: "close".utf8, at: header.value) {
```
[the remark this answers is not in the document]
normanmaurer: @Lukasa @weissi I can fix this as part of this PR or do a followup ... Whatever you prefer. let me know.

Sources/NIOHTTP1/HTTPTypes.swift
normanmaurer: @weissi let me see if it changes anything (as it will only if we actual have a connection header).

normanmaurer: @weissi @Lukasa ok no changes as we never have such a header.

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
weissi: @Lukasa yes, commented on that one too but I'd be happy with a follow-up.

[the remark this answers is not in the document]
normanmaurer: will do a followup

PR #404, opened 2018-05-16
[2.0] fix integer types for port numbers/nanoseconds

weissi:
Motivation:

In at least two places we picked the integer types in an unfortunate
way:

- `TimeAmount` stores its nano-seconds amount in an `Int` which is fine
   for 64-bit platforms but not for 32-bit (2^32 ns is just over 4 s).
   Fortunately we could define `TimeAmount.Value` which is `Int` on 64-bit
   platforms (to not break the API) but `Int64` on 32-bit platforms.
   Ideally it would just be `Int64` everywhere
- We always use `Int` for port numbers, except the `port` value of
  `SocketAddress` uses `UInt16` for no good reason

Modifications:

- changed `SocketAddress#port` to be `Int`
- changed `TimeAmount` to always work with `Int64`

Result:

- more consistency


Sources/NIO/SocketAddresses.swift
```diff
@@ -112,12 +112,12 @@ public enum SocketAddress: CustomStringConvertible {
     }
 
     /// Get the port associated with the address, if defined.
-    public var port: UInt16? {
+    public var port: Int? {
```
normanmaurer: @weissi isn't that a public API change and so breaking ?

weissi: @normanmaurer yes, that's why it's marked as "⚠️ needs-major-version-bump" and "[2.0]" :).

Just wanted to add this PR (to merge when we decide to release 2.0) just so we won't forget this change.

helje5:
I think you should just do this, but I was thinking about the:

> Ideally it would just be Int64 everywhere

I think from a technical perspective the current version is actually right, just ugly: It should be `Int`, unless `Int` is smaller than 64-bit. Imagine next year when the new 256-core 128bit ARM MacPro comes out. `Int64` would still work, but using `Int` might be faster and integrate better w/ other code.

Just something to keep in mind.

weissi:
@helje5 you're right that `Int` will integrate better with other code. But from a speed perspective they're exactly the same. Both will lower to LLVM's int64 type. The difference between `Int` and `Int64` exists only at compile time (on 64-bit platforms), conversions between the two should be free (no instructions issued), etc.

weissi:
@helje5 🤣, we'll then release NIO 3 :)

weissi:
done in some other PR

PR #408, opened 2018-05-16
Make removeHandlers public.

Lukasa:
Motivation:

If ChannelPipeline.removeHandlers is internal it is extremely difficult
to implement a correct ChannelCore.

Modifications:

Make ChannelPipeline.removeHandlers public.

Result:

Easier to implement ChannelCore.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -499,7 +499,10 @@ public final class ChannelPipeline: ChannelInvoker {
         return name
     }
 
-    /// Remove all the `ChannelHandler`s from the `ChannelPipeline` and destroy these. This method must only be called from within the `EventLoop`.
+    /// Remove all the `ChannelHandler`s from the `ChannelPipeline` and destroy these.
+    ///
+    /// This method must only be called from within the `EventLoop`. It should only be called from a `ChannelCore`
+    /// implementation. Once called, the `ChannelPipeline` is no longer active and cannot be used again.
     func removeHandlers() {
         assert(eventLoop.inEventLoop)
 
--- a/Sources/NIO/Channel.swift
+++ b/Sources/NIO/Channel.swift
@@ -273,6 +273,19 @@ public extension ChannelCore {
     public func unwrapData<T>(_ data: NIOAny, as: T.Type = T.self) -> T {
         return data.forceAs()
     }
+
+    /// Removes the `ChannelHandler`s from the `ChannelPipeline` belonging to `channel`, and
+    /// closes that `ChannelPipeline`.
+    ///
+    /// This method is intended for use when writing custom `ChannelCore` implementations.
+    /// This can be called from `close0` to tear down the `ChannelPipeline` when closure is
+    /// complete.
+    ///
+    /// - parameters:
+    ///     - channel: The `Channel` whose `ChannelPipeline` will be closed.
+    public func removeHandlers(channel: Channel) {
+        channel.pipeline.removeHandlers()
+    }
 }
 
 /// An error that can occur on `Channel` operations.


Sources/NIO/ChannelPipeline.swift
```diff
@@ -499,8 +499,11 @@ public final class ChannelPipeline: ChannelInvoker {
         return name
     }
 
-    /// Remove all the `ChannelHandler`s from the `ChannelPipeline` and destroy these. This method must only be called from within the `EventLoop`.
-    func removeHandlers() {
+    /// Remove all the `ChannelHandler`s from the `ChannelPipeline` and destroy these.
+    ///
+    /// This method must only be called from within the `EventLoop`. It should only be called from a `ChannelCore`
+    /// implementation. Once called, the `ChannelPipeline` is no longer active and cannot be used again.
+    public func removeHandlers() {
```
weissi: I think we should hide this method better, maybe on the ChannelCore and export it through:

    self._unsafe.removeHandlers(self.pipeline)

Lukasa: This actually can't be on the `ChannelCore` because the `ChannelCore` doesn't own the `pipeline`, the `Channel` does. Not really sure about exactly what to do there: suggestions?

normanmaurer: hmm good question. If we make it part of the `ChannelPipeline` we should at least underscore it imho.

Lukasa: @weissi Mostly because `close0` is a method on `ChannelCore`, not `Channel`.

As a broader statement I think this might be reflective of a layering issue here. We have defined `ChannelCore` and `Channel` as separate protocols without a requirement that they be implemented by the same entity. `Channel` has a `close` method but it's explicitly defined as firing down the `ChannelPipeline`, which will eventually end up invoking `ChannelCore.close0`. The unfortunately irony here is that `ChannelCore.close0` cannot actually close the pipeline, because the `ChannelCore` does not know where the pipeline is unless the same object implements both `ChannelCore` and `Channel`.

In practice all currently written channel implementations of which I am aware do implement both `Channel` and `ChannelCore` directly. If we think that's a sensible design pattern then we should enforce that requirement going forward. Otherwise, we should decide how this layering is supposed to work with regard to closing pipelines.

One proposal is to say that, while `Channel` and `ChannelCore` may be implemented by different objects, they naturally must have some kind of API between them, because `closeFuture` (another object required for correctly signalling closure) is *also* on `Channel` and not `ChannelCore`. That means it's simply not possible to write a fully-functional channel without communicating between the two entities. In that circumstance I think the best place to put this extension is actually on `Channel`, and we should define this:

```swift
public extension Channel {
    public func _closedFromChannelCore(error: Error?) {
        self.pipeline.removeHandlers()
        if let error = error {
            self.closeFuture.fail(error: error)
        } else {
            self.closeFuture.succeed(result: ())
        }
    }
}
```

normanmaurer: @Lukasa this sounds sensible to me for now.

weissi: not a huge fan of having this `public` on the `Channel`.

weissi: still prefer doing the layering violation and doing this:

```
public extension ChannelCore {
    public func removeHandlers(channel: Channel) {
        channel.pipeline.removeHandlers()
    }
}
```

Lukasa: Yeah, I think that's the right thing to do right now. @weissi and I discussed this and we think that actually, `ChannelCore` and `Channel` have an implicit expectation that they are implemented on the same object: the fact that they're separate protocols is only to prevent users handling the `Channel` existential from seeing the methods of `ChannelCore` all the time.

Longer term we may want a better solution, but for now I think it's fine to just do the layering violation. I'll leave a comment to that effect.

PR #411, opened 2018-05-17
introduce ByteBufferView

weissi:
RFC: @normanmaurer / @Lukasa / @moiseev / @airspeedswift 

Motivation:

Sometimes it would be handy to have ByteBuffers conform to a
`Collection` where `Element == UInt8`. ByteBuffer however has multiple
modes how to operate so it would be wrong to have say the readable bytes
be the ByteBuffer's collection.
Therefore it would be nice to have a view into a `ByteBuffer` which is a
`Collection where Element == UInt8`. This introduces these.

Modifications:

- add `ByteBufferView`
- add `ByteBuffer.readableBytesView`
- add `ByteBuffer.viewBytes(range: ...)`

Result:

- easier to work with `ByteBuffer`s in the Swift ecosystem.

### DIFF
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -1,0 +1,79 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A read-only view into a portion of a `ByteBuffer`.
+///
+/// A `ByteBufferView` is useful whenever a `Collection where Element == UInt8` representing a portion of a
+/// `ByteBuffer` is needed.
+public struct ByteBufferView: ContiguousCollection, RandomAccessCollection {
+    public typealias Element = UInt8
+    public typealias Index = Int
+    public typealias SubSequence = ByteBufferView
+
+    private let buffer: ByteBuffer
+    private let range: Range<Index>
+
+    internal init(buffer: ByteBuffer, range: Range<Index>) {
+        precondition(range.lowerBound >= 0 && range.upperBound < buffer.capacity)
+        self.buffer = buffer
+        self.range = range
+    }
+
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try self.buffer.withVeryUnsafeBytes { ptr in
+            try body(UnsafeRawBufferPointer.init(start: ptr.baseAddress!.advanced(by: self.range.lowerBound),
+                                                 count: self.range.count))
+        }
+    }
+
+    public var startIndex: Index {
+        return self.range.lowerBound
+    }
+
+    public var endIndex: Index {
+        return self.range.upperBound
+    }
+
+    public func index(after i: Index) -> Index {
+        return i + 1
+    }
+
+    public subscript(position: Index) -> UInt8 {
+        guard position >= self.range.lowerBound && position < self.range.upperBound else {
+            preconditionFailure("index \(position) out of range")
+        }
+        return self.buffer.getInteger(at: position)!
+    }
+
+    public subscript(range: Range<Index>) -> ByteBufferView {
+        return ByteBufferView(buffer: self.buffer, range: range)
+    }
+}
+
+public extension ByteBuffer {
+    /// A view into the readable bytes of the `ByteBuffer`.
+    public var readableBytesView: ByteBufferView {
+        return ByteBufferView(buffer: self, range: self.readerIndex ..< self.readerIndex + self.readableBytes)
+    }
+
+    /// Returns a view into some portion of a `ByteBuffer`.
+    ///
+    /// - parameters:
+    ///   - index: The index the view should start at
+    ///   - length: The length of the view (in bytes)
+    /// - returns A view into a portion of a `ByteBuffer`.
+    public func viewBytes(at index: Int, length: Int) -> ByteBufferView {
+        return ByteBufferView(buffer: self, range: index ..< index+length)
+    }
+}
--- a/Tests/NIOTests/ByteBufferTest+XCTest.swift
+++ b/Tests/NIOTests/ByteBufferTest+XCTest.swift
@@ -116,6 +116,11 @@ extension ByteBufferTest {
                 ("testDiscardReadBytesOnConsumedBuffer", testDiscardReadBytesOnConsumedBuffer),
                 ("testDumpBytesFormat", testDumpBytesFormat),
                 ("testStaticStringCategorySubscript", testStaticStringCategorySubscript),
+                ("testReadableBytesView", testReadableBytesView),
+                ("testReadableBytesViewNoReadableBytes", testReadableBytesViewNoReadableBytes),
+                ("testBytesView", testBytesView),
+                ("testViewsStartIndexIsStable", testViewsStartIndexIsStable),
+                ("testSlicesOfByteBufferViewsAreByteBufferViews", testSlicesOfByteBufferViewsAreByteBufferViews),
            ]
    }
 }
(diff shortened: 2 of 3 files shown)


Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct ByteBufferView: ContiguousCollection {
```
Lukasa: Needs a doc comment.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,51 @@
...
+    public typealias Element = UInt8
+    public typealias Index = Int
+
+    private let buffer: ByteBuffer
+
+    internal init(buffer: ByteBuffer, range: Range<Int>) {
```
Lukasa: Is the range actually a good type here? I'm inclined to say that start index and length is more likely to be helpful here.

Sources/NIO/ByteBuffer-views.swift
Lukasa: This should probably implement `RandomAccessCollection` as well, to give it a bunch more utility functionality. I suspect, however, that implementing `MutableCollection` is unlikely to be a good idea: I suspect it will trigger CoWs if you actually try to mutate it.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,51 @@
...
+        return self.buffer.getInteger(at: position)!
+    }
+}
+
+public extension ByteBuffer {
+    public var readableBytesView: ByteBufferView {
+        return ByteBufferView(buffer: self, range: self.readerIndex ..< self.readerIndex + self.readableBytes)
+    }
+
+    public func viewBytes(range: Range<Int>) -> ByteBufferView {
```
Lukasa: Same note about `Range` here.

Sources/NIO/ByteBuffer-views.swift
weissi: @Lukasa but range is just a start index and an end index?

Sources/NIO/ByteBuffer-views.swift
weissi: (and would be implied by `ContiguousCollection` I guess because I mean if you're in memory contiguously then obvs random access is fast

Sources/NIO/ByteBuffer-views.swift
Lukasa: Sure, but it’s not in line with any of the rest of `ByteBuffer`s methods. We don’t use Range anywhere else. 

More generally, it doesn’t seem to be a common Swift pattern for ranges to be arguments to anything but subscripts. Even `stride` doesn’t take a range.

weissi: hmm, but for a view a Range makes a lot of sense, no? You surely want to write `viewBytes(range: 0..<10)` if you want 10 bytes, no? I mean we could make it a subscript if you want `subscript(viewBytes viewBytes: Range<Int>) -> ByteBufferView`

Lukasa: If it was a subscript, sure. But otherwise it forces the math you’ve done in your test: `index..<(index + count)`.

tomerd:

Sources/NIO/ByteBuffer-views.swift
normanmaurer: I thin I would use `viewBytes(at: Int, length: Int)`

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    public var startIndex: Index { return 0 }
+    public var endIndex: Index { return self.buffer.readableBytes }
+    public func index(after i: Index) -> Index { return i + 1 }
+
+    public subscript(position: Int) -> UInt8 {
+        guard position >= 0 && position < self.buffer.readableBytes else {
+            preconditionFailure("index \(position) out of range")
+        }
+        return self.buffer.getInteger(at: position)!
```
normanmaurer: why not just remove the guard above and just check that it not return nil ?

moiseev:
Since I did not see the `subscript(_:Range<Int>)` definition on the `ByteBufferView`, it must be getting a default `Slice` one, which, for a view-type, makes little sense. And it would be fairly trivial to implement:

```swift
public subscript(range: Range<Index>) -> ByteBufferView {
  return ByteBufferView(buffer: self.buffer, range: range)
}
```

weissi:
thanks @moiseev . Added!

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    internal init(buffer: ByteBuffer, range: Range<Int>) {
+        precondition(range.lowerBound >= 0 && range.upperBound < buffer.capacity)
+        self.buffer = buffer
+        self.range = range
+    }
+
+    public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
+        return try self.buffer.withVeryUnsafeBytes { ptr in
+            try body(UnsafeRawBufferPointer.init(start: ptr.baseAddress.map { $0 + self.range.lowerBound},
```
Lukasa: Should we just force-unwrap this, rather than use the map?

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,79 @@
...
+///
+/// A `ByteBufferView` is useful whenever a `Collection where Element == UInt8` representing a portion of a
+/// `ByteBuffer` is needed.
+public struct ByteBufferView: ContiguousCollection, RandomAccessCollection {
+    public typealias Element = UInt8
+    public typealias Index = Int
+    public typealias SubSequence = ByteBufferView
+
+    private let buffer: ByteBuffer
+    private let range: Range<Int>
```
Lukasa: For minor clarity this should probably be `Range<Index>`.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    internal init(buffer: ByteBuffer, range: Range<Int>) {
```
Lukasa: And here.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    public var endIndex: Index {
+        return self.range.upperBound
+    }
+
+    public func index(after i: Index) -> Index {
+        return i + 1
+    }
+
+    public subscript(position: Int) -> UInt8 {
```
Lukasa: This should probably be `Index` as well.

weissi:
@normanmaurer / @Lukasa updated

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public subscript(range: Range<Index>) -> ByteBufferView {
+        return ByteBufferView(buffer: self.buffer, range: range)
+    }
+}
+
+public extension ByteBuffer {
+    /// A view into the readable bytes of the `ByteBuffer`.
+    public var readableBytesView: ByteBufferView {
```
normanmaurer: @weissi should we rename this to: `viewReadableBytes` just to match `viewBytes` a bit better ?

normanmaurer: Also @Lukasa WDYT ? 

Beside this I am happy with the pr. Great work

weissi: @normanmaurer just for context: I made this `readableBytesView` because it's a property and they aren't usually 'actions', they are things. So the property doesn't allow you 'to view the bytes' but rather it _is_ the 'view of bytes'.
If we make it `viewReadableBytes` then I think it should be a `func`.

normanmaurer: hmm ok. I just found the naming difference a bit odd. That said I am not strong on this and I think as you mentioned it should be a property. So feel free to ignore me

weissi: yes, @Lukasa for input

Lukasa: I'm fine with the names as-is.

weissi:
@Lukasa  yes

Lukasa:
Ah, no, that's not right: it's a syscall slowdown.

PR #413, opened 2018-05-17
don't just drop Channels on the floor if registration fails

weissi:
Motivation:

If registration of a Channel fails we need to close that Channel.
Previously we'd just drop it on the floor which triggers an assertion
that an open channel got leaked.

Modifications:

In cases where the Channel registration fails, we need to close that
channel.

Result:

No crashes if registration fails.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -182,6 +182,7 @@ public final class ServerBootstrap {
                                            group: childEventLoopGroup,
                                            protocolFamily: address.protocolFamily)
         }
+
         return bind0(makeServerChannel: makeChannel) { (eventGroup, serverChannel) in
             serverChannel.registerAndDoSynchronously { serverChannel in
                 serverChannel.bind(to: address)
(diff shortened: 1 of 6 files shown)


Tests/NIOTests/ChannelTests.swift
```diff
@@ -2360,6 +2360,81 @@ public class ChannelTests: XCTestCase {
...
+    }
+}
+
+fileprivate final class FailRegistrationHandler: ChannelOutboundHandler {
+    enum RegistrationFailedError: Error { case error }
+
+    typealias OutboundIn = Never
+
+    func register(ctx: ChannelHandlerContext, promise: EventLoopPromise<Void>?) {
+        promise?.fail(error: RegistrationFailedError.error)
```
normanmaurer: @weissi maybe use `!` as it should never be nil.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2360,6 +2360,81 @@ public class ChannelTests: XCTestCase {
...
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        do {
+            let serverChannel = try ServerBootstrap(group: group)
+                .serverChannelInitializer { channel in
+                    channel.pipeline.add(handler: FailRegistrationHandler())
+                }
+                .bind(host: "localhost", port: 0).wait()
+            XCTFail("shouldn't be reached")
+            XCTAssertNoThrow(try serverChannel.close().wait())
```
normanmaurer: Remove this one as it should never be reached anyway ?

tomerd:

Tests/NIOTests/ChannelTests.swift
```diff
@@ -2360,6 +2360,81 @@ public class ChannelTests: XCTestCase {
...
+            }
+            .bind(host: "localhost", port: 0).wait()
+        defer {
+            XCTAssertNoThrow(try serverChannel.close().wait())
+        }
+        let clientChannel = try ClientBootstrap(group: group)
+            .connect(to: serverChannel.localAddress!)
+            .wait()
+        defer {
+            XCTAssertNoThrow(try clientChannel.close().wait())
```
Lukasa: There's a race in this `close`, racing with the close in the accepted server socket. Better to just wait for the `closeFuture` on this channel, which has the nice side effect of confirming that the channel actually did die.

Sources/NIO/Bootstrap.swift
```diff
@@ -185,6 +185,9 @@ public final class ServerBootstrap {
         return bind0(makeServerChannel: makeChannel) { (eventGroup, serverChannel) in
```
Lukasa: Should `bind0` do this error handling? There are a few other paths through this bootstrap that didn't get changed here but that still do registrations.

Tests/NIOTests/ChannelTests.swift
weissi: I added it because if it fails, not closing it leads to a crash (assert) and then hides the actual bug. I’d leave it in but am also happy to take out.
You know, xctest doesn’t stop on failed assert...

normanmaurer: ah right... Ignore me then (or maybe add a comment) :)

Sources/NIO/Bootstrap.swift
weissi: done now, thanks

weissi:

weissi:
@normanmaurer / @Lukasa updated this PR

Lukasa:
This has regressed allocations on 4.1, it seems.

weissi:
@lukasa sadly that is expected as I added extra future related allocations. But should only be per connection not per request. Will change the docket setup in this PR tomorrow :(

PR #423, opened 2018-05-21
explain why typealias is complicated

weissi:
CC @BasThomas , forgot to address your comment before merge.

Motivation:

We should explain why this typealias

    public typealias RangeType<Bound> = CountableRange<Bound> where Bound: Strideable, Bound.Stride: SignedInteger

isn't just

    public typealias RangeType = CountableRange

and that's because SR-6963.

Modifications:

added a comment explaining why

Result:

code easier to read


BasThomas:
👍

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #424, opened 2018-05-22
fix EventLoops that Bootstrap channel initialiser's call out on

weissi:
Motivation:

Previously we were not running the (child/server)channelInitializers on the
event loop associated to the `Channel` we're initialising. Almost all
operations in there are pipeline modifications which are thread safe so
it presumably wasn't a massive correctness issue. However it's very
expensive to hop threads that often and it is also very unexpected. This
addresses this issue.

Modifications:

made all (child/server)channelInitializers run on the event loop that is
associated to their `Channel`

Result:

- more correctness
- less unexpected behaviour

effects on performance seem to also be positive:

before:

```
measuring: http1_10k_reqs_1_conn: 0.8006269931793213, 0.8106319904327393, 0.7860230207443237, 0.7819960117340088, 0.785086989402771, 0.792726993560791, 0.782039999961853, 0.7802549600601196, 0.7886660099029541, 0.7769789695739746, 
measuring: http1_10k_reqs_100_conns: 0.857327938079834, 0.8519150018692017, 0.8499130010604858, 0.8529999256134033, 0.8657670021057129, 0.8585480451583862, 0.8542360067367554, 0.8532159328460693, 0.8540490865707397, 0.8516910076141357, 
```

so the minimums are `0.7769789695739746` and `0.8499130010604858`

after:

```
measuring: http1_10k_reqs_1_conn: 0.7713319063186646, 0.7732300758361816, 0.7659409046173096, 0.7707960605621338, 0.7677010297775269, 0.8026899099349976, 0.7947590351104736, 0.8216999769210815, 0.7768599987030029, 0.771873950958252, 
measuring: http1_10k_reqs_100_conns: 0.8300169706344604, 0.835252046585083, 0.833683967590332, 0.831184983253479, 0.8326460123062134, 0.8323450088500977, 0.8354840278625488, 0.8349909782409668, 0.8335800170898438, 0.8432250022888184, 
```

minimums: `0.7659409046173096` and `0.8300169706344604`

which suggests that new connections should be ever so slightly faster which is expected.


Sources/NIO/Bootstrap.swift
```diff
@@ -205,18 +205,22 @@ public final class ServerBootstrap {
...
+        return eventLoop.submit {
+            return serverChannelInit(serverChannel).then {
+                serverChannel.pipeline.add(handler: AcceptHandler(childChannelInitializer: childChannelInit,
+                                                                  childChannelOptions: childChannelOptions))
+            }.then {
+                serverChannelOptions.applyAll(channel: serverChannel)
+            }.then {
+                register(eventLoop, serverChannel)
+            }.map {
+                serverChannel as Channel
```
normanmaurer: @weissi why we need this ?

weissi: @normanmaurer because otherwise the compiler complains can't do `EventLoopFuture<ServerSocketChannel>` as `EventLoopFuture<Channel>`

Issue #422, opened 2018-05-21
HTTPDecoder: inline data for upgrades not currently supported


weissi:
### Expected behavior

- extra inline data for HTTP upgrades will be delivered through the pipeline after the upgrade has succeeded

### Actual behavior

- assertion hit
- even when removing the assertion: data lost

### Steps to reproduce

apply this patch and run test

```
diff --git a/Tests/NIOHTTP1Tests/HTTPUpgradeTests.swift b/Tests/NIOHTTP1Tests/HTTPUpgradeTests.swift
index bf3ece0..3397eea 100644
--- a/Tests/NIOHTTP1Tests/HTTPUpgradeTests.swift
+++ b/Tests/NIOHTTP1Tests/HTTPUpgradeTests.swift
@@ -814,4 +814,108 @@ class HTTPUpgradeTestCase: XCTestCase {
         try channel.pipeline.assertDoesNotContain(handlerType: HTTPRequestDecoder.self)
         try channel.pipeline.assertDoesNotContain(handlerType: HTTPResponseEncoder.self)
     }
+    
+    func testUpgradeWithUpgradePayloadInlineWithRequestWorks() throws {
+        var upgradeRequest: HTTPRequestHead? = nil
+        var upgradeHandlerCbFired = false
+        var upgraderCbFired = false
+        
+        class CheckWeReadInlineAndExtraData: ChannelDuplexHandler {
+            typealias InboundIn = ByteBuffer
+            typealias OutboundIn = Never
+            typealias OutboundOut = Never
+            
+            enum State {
+                case fresh
+                case added
+                case inlineDataRead
+                case extraDataRead
+                case closed
+            }
+            
+            private let allDonePromise: EventLoopPromise<Void>
+            private var state = State.fresh
+            
+            init(allDonePromise: EventLoopPromise<Void>) {
+                self.allDonePromise = allDonePromise
+            }
+            
+            func handlerAdded(ctx: ChannelHandlerContext) {
+                XCTAssertEqual(.fresh, self.state)
+                self.state = .added
+            }
+            
+            func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+                let buf = self.unwrapInboundIn(data)
+                XCTAssertEqual(1, buf.readableBytes)
+                let stringRead = buf.getString(at: 0, length: buf.readableBytes)
+                switch self.state {
+                case .added:
+                    XCTAssertEqual("A", stringRead)
+                    self.state = .inlineDataRead
[...]

PR #427, opened 2018-05-22
Ensure HTTPDecoder will not run into re-entrance issues and correctly…

normanmaurer:
… forward bytes left after removal.

Motivation:

We need to ensure we correctly guard against re-entrancy for all cases. Also we did not correctly ensure we forward pending data after removal which could lead to missing data after upgrades.

Modifications:

- pause the parser when we need to callout to the pipeline and so ensure we never fire events throught the pipeline while in callbacks of http_parser
- correctly forward any pending bytes if an upgrade happened when the decoder is removed
- Ensure HTTPUpgradeHandler only remove decoder after the full request is received
- Add testcase provided by @weissi

Result:

Correctly handle upgrades and pending data. Fixes https://github.com/apple/swift-nio/issues/422.

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -34,6 +34,9 @@ private struct HTTPParserState {
     var currentError: HTTPParserError?
     var seenEOF = false
     var headerStartIndex: Int?
+    
+    // Holds the data we need to forward via ctx.fireChannelRead(...) after invoking the parser.
+    var pendingInOut: NIOAny? = nil
 
     enum DataAwaitingState {
         case messageBegin
@@ -57,6 +60,7 @@ private struct HTTPParserState {
         self.currentStatus = nil
         self.slice = nil
         self.headerStartIndex = nil
+        self.pendingInOut = nil
     }
 
     var cumulationBuffer: ByteBuffer?
@@ -196,7 +200,6 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
 
     private var parser = http_parser()
     private var settings = http_parser_settings()
-    private var decoding: Bool = false
 
     fileprivate var state = HTTPParserState()
 
@@ -234,6 +237,25 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         return response
     }
 
+    private func bytesToForwardOnRemoval(ctx: ChannelHandlerContext) -> ByteBuffer? {
+        guard self.parser.upgrade == 1 && ctx.channel.isActive else {
+            return nil
+        }
+        // We take a slice of the cumulationBuffer so the next handler in the pipeline will just see the readable portion of the buffer.
+        // While this is not strictly needed it may make it easier to consume.
+        if let buffer = self.cumulationBuffer?.slice(), buffer.readableBytes > 0 {
+            return buffer
+        }
+        return nil
+    }
+
+    public func handlerRemoved(ctx: ChannelHandlerContext) {
+        if let buffer = self.bytesToForwardOnRemoval(ctx: ctx) {
+            ctx.fireChannelRead(NIOAny(buffer))
+        }
+        self.cumulationBuffer = nil
+    }
+    
     public func decoderAdded(ctx: ChannelHandlerContext) {
         if HTTPMessageT.self == HTTPServerRequestPart.self {
             c_nio_http_parser_init(&self.parser, HTTP_REQUEST)
@@ -258,6 +280,11 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
             let ctx = evacuateChannelHandlerContext(parser)
             let handler = ctx.handler as! AnyHTTPDecoder
 
+            // Ensure we pause the parser after this callback is complete so we can safely callout
+            // to the pipeline.
+            c_nio_http_parser_pause(parser, 1)
+            assert(handler.state.pendingInOut == nil)
+
             handler.state.complete(state: handler.state.dataAwaitingState)
             handler.state.dataAwaitingState = .body
 
@@ -269,7 +296,7 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
                     return -1
                 }
 
-                ctx.fireChannelRead(handler.wrapInboundOut(HTTPServerRequestPart.head(head)))
+                handler.state.pendingInOut = handler.wrapInboundOut(HTTPServerRequestPart.head(head))
                 return 0
             case let handler as HTTPResponseDecoder:
                 let head = handler.newResponseHead(parser)
@@ -278,7 +305,7 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
                     return -1
                 }
 
-                ctx.fireChannelRead(handler.wrapInboundOut(HTTPClientResponsePart.head(head)))
+                handler.state.pendingInOut = handler.wrapInboundOut(HTTPClientResponsePart.head(head))
 
                 // http_parser doesn't correctly handle responses to HEAD requests. We have to do something
                 // annoyingly opaque here, and in those cases return 1 instead of 0. This forces http_parser
@@ -322,14 +349,19 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
             let handler = ctx.handler as! AnyHTTPDecoder
             assert(handler.state.dataAwaitingState == .body)
 
+            // Ensure we pause the parser after this callback is complete so we can safely callout
+            // to the pipeline.
+            c_nio_http_parser_pause(parser, 1)
+            assert(handler.state.pendingInOut == nil)
+
             // Calculate the index of the data in the cumulationBuffer so we can slice out the ByteBuffer without doing any memory copy
             let index = handler.state.calculateIndex(data: data!, length: len)
             let slice = handler.state.cumulationBuffer!.getSlice(at: index, length: len)!
             switch handler {
             case let handler as HTTPRequestDecoder:
-                ctx.fireChannelRead(handler.wrapInboundOut(HTTPServerRequestPart.body(slice)))
+                handler.state.pendingInOut = handler.wrapInboundOut(HTTPServerRequestPart.body(slice))
             case let handler as HTTPResponseDecoder:
-                ctx.fireChannelRead(handler.wrapInboundOut(HTTPClientResponsePart.body(slice)))
+                handler.state.pendingInOut = handler.wrapInboundOut(HTTPClientResponsePart.body(slice))
             default:
                 fatalError("the impossible happened: handler neither a HTTPRequestDecoder nor a HTTPResponseDecoder which should be impossible")
             }
@@ -387,6 +419,12 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         self.settings.on_message_complete = { parser in
             let ctx = evacuateChannelHandlerContext(parser)
             let handler = ctx.handler as! AnyHTTPDecoder
+            
+            // Ensure we pause the parser after this callback is complete so we can safely callout
+            // to the pipeline.
+            c_nio_http_parser_pause(parser, 1)
+            assert(handler.state.pendingInOut == nil)
+
             handler.state.complete(state: handler.state.dataAwaitingState)
             handler.state.dataAwaitingState = .messageBegin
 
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -234,6 +237,15 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         return response
     }
 
+    public func handlerRemoved(ctx: ChannelHandlerContext) {
+        if let buffer = self.cumulationBuffer?.slice(), buffer.readableBytes > 0, self.parser.upgrade == 1, ctx.channel.isActive {
+            self.cumulationBuffer = nil
+            ctx.fireChannelRead(NIOAny(buffer))
```
weissi: is this really safe? It really feels like there's a scenario where the next channel hander expects HTTP messages and this gets invoked during a channel close and then the next handler crashes

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -387,6 +409,12 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         self.settings.on_message_complete = { parser in
             let ctx = evacuateChannelHandlerContext(parser)
             let handler = ctx.handler as! AnyHTTPDecoder
+            
+            // Ensure we pause the parser after this callback is complete so we can safely callout
+            // to the pipeline.
+            c_nio_http_parser_pause(parser, 1)
```
weissi: don't we need similar stuff for `on_body` or so?

weissi: ignore me, we've got that

normanmaurer:
@weissi will run perf Tests after dinner

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -234,6 +237,15 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         return response
     }
 
+    public func handlerRemoved(ctx: ChannelHandlerContext) {
+        if let buffer = self.cumulationBuffer?.slice(), buffer.readableBytes > 0, self.parser.upgrade == 1, ctx.channel.isActive {
```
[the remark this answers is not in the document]
Lukasa: Yeah, we prefer `&&` over `,` except when we need to separate conditional bindings.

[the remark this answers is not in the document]
Lukasa: Also this is a really complex conditional, can we write a function to give this conditional a name?

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Yeah, don't do this if `ctx.channel.isActive` is `false`. That way if `handlerRemoved` is being invoked while the channel is being torn down you won't forward the data on.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -60,6 +60,7 @@ public protocol HTTPProtocolUpgrader {
 /// requests that we choose to punt on it entirely and not allow it. As it happens this is mostly fine:
 /// the odds of someone needing to upgrade midway through the lifetime of a connection are very low.
 public class HTTPServerUpgradeHandler: ChannelInboundHandler {
+
```
Lukasa: Why are we adding this blank line?

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -191,48 +203,45 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
...
-
-                // We unbuffer any buffered data here and, if we sent any,
-                // we also fire readComplete.
-                let bufferedMessages = self.receivedMessages
-                self.receivedMessages = []
-                bufferedMessages.forEach { ctx.fireChannelRead($0) }
-                if bufferedMessages.count > 0 {
-                    ctx.fireChannelReadComplete()
+           
+            return {
```
Lukasa: There's no need to heap allocate this, closing over the `ChannelHandlerContext`. Instead, factor this out to a separate function and just use a state variable.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -131,45 +131,57 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
...
-        if !handleUpgrade(ctx: ctx, request: request, requestedProtocols: requestedProtocols) {
-            notUpgrading(ctx: ctx, data: data)
+        if let upgrade = self.upgrade {
+            switch requestPart {
+            case .head(_):
+                ctx.fireErrorCaught(HTTPUpgradeErrors.invalidHTTPOrdering)
+                notUpgrading(ctx: ctx, data: data)
+                return
+            case .body(_):
+                // ignore
```
Lukasa: Can you mark this with a TODO that indicates that in a future NIO version we want to add an API where we deliver this to the upgrader in some form?

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -131,45 +131,57 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
...
+                break
+            case .end(_):
+                self.seenFirstRequest = true
+
+                // The request is complete now trigger the upgrade.
+                upgrade()
+            }
+        } else {
+            // We should only ever see a request header: by the time the body comes in we should
+            // be out of the pipeline. Anything else is an error.
```
Lukasa: This comment needs rewording as it's no longer accurate. Maybe:

> We should decide if we're going to upgrade based on the first request header: if we aren't upgrading, by the time the body comes in we should be out of the pipeline. That means that if we don't think we're upgrading, the only thing we should see is a request head. Anything else in an error.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
normanmaurer: Will revert, left-over from some debugging Code that was there

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: I legit didn't even see that. That's part of why I wanted this conditional factored out. ;)

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
Lukasa: Eh, I guess that's fine? It feels super weird to me, but ok. We can't use `@convention(thin)` because as @normanmaurer notes this actually closes over a *bunch* of state.

We could encapsulate this in a struct instead, to avoid the heap allocation.

normanmaurer: I am happy with the struct as well I just thought this one is a bit nicer and it should not matter. That said I am fine either way, just tell me what to do :)

normanmaurer:
Alright did run our perf-suite...

master:

```
measuring: no-net_http1_10k_reqs_1_conn: 0.193889021873474, 0.19300103187561, 0.189878940582275, 0.202571034431458, 0.195896029472351, 0.197288036346436, 0.202276945114136, 0.193497061729431, 0.199617028236389, 0.190865039825439,
measuring: http1_10k_reqs_1_conn: 0.559623003005981, 0.557006001472473, 0.561303019523621, 0.561969041824341, 0.565063953399658, 0.564036011695862, 0.570590972900391, 0.555812001228333, 0.560003042221069, 0.556936979293823,
measuring: http1_10k_reqs_100_conns: 0.602265000343323, 0.603261947631836, 0.60293710231781, 0.606209993362427, 0.620349049568176, 0.612009048461914, 0.66516900062561, 0.610424995422363, 0.606754899024963, 0.618997097015381,
```

With this pr:

```
measuring: http1_10k_reqs_1_conn: 0.599761009216309, 0.598889946937561, 0.595099091529846, 0.598055958747864, 0.602496981620789, 0.640310049057007, 0.588855981826782, 0.594573974609375, 0.610931038856506, 0.599555015563965,
measuring: http1_10k_reqs_100_conns: 0.64155101776123, 0.639289975166321, 0.645186901092529, 0.649052023887634, 0.64255690574646, 0.644281029701233, 0.637521982192993, 0.642242074012756, 0.65774393081665, 0.642369031906128,
```

normanmaurer:
So it seems like this definitely harm the perf a little bit, which I think is not really unexpected with the extra branching etc.

normanmaurer:
@weissi @Lukasa ^^

weissi:
@normanmaurer ouch, raw performance, no networking (`no-net_http1_10k_reqs_1_conn`) down 16% 😢.

normanmaurer:
@weissi yeah... that said I am not sure there is much we can do here (except use a pending list again)... I am open to suggestions tho :)

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -613,15 +641,25 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
+        }
+    }
+    
+    private func currentParserError() -> HTTPParserError? {
         let httpError = self.parser.http_errno
-        if httpError != 0 {
-            self.state.currentError = HTTPParserError.httpError(fromCHTTPParserErrno: http_errno(rawValue: httpError))!
-            ctx.fireErrorCaught(self.state.currentError!)
+        // Also take into account that we may have called c_nio_http_parser_pause(...)
+        guard httpError != 0 && httpError != HPE_PAUSED.rawValue else {
```
normanmaurer: @weissi I guess it not really matters much for perf stuff but we may want to consider switching these as most of the time it will be HPE_PAUSED in reality so the first test is a waste mostly.

weissi:
@normanmaurer is it faster with the callout list?

normanmaurer:
@weissi I will check in a few need to fix some other stuff first :(

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
Lukasa: For now we can allocate the closure, we can always remove it later.

normanmaurer:
After https://github.com/apple/swift-nio/pull/427/commits/61b3149f7ccbfa62fbc2c4fff61342df0872c1a8 : 

```
measuring: no-net_http1_10k_reqs_1_conn: 0.2206209897995, 0.221046090126038, 0.211659908294678, 0.212454080581665, 0.210615038871765, 0.208593010902405, 0.209282994270325, 0.208821058273315, 0.209628939628601, 0.209208965301514,
measuring: http1_10k_reqs_1_conn: 0.580152988433838, 0.57462203502655, 0.577759981155396, 0.582140922546387, 0.578655004501343, 0.572811961174011, 0.572220087051392, 0.578151941299438, 0.583968043327332, 0.611997961997986,
measuring: http1_10k_reqs_100_conns: 0.62391996383667, 0.626338005065918, 0.628537058830261, 0.629424929618835, 0.61939001083374, 0.638810038566589, 0.632416963577271, 0.63428795337677, 0.626121044158936, 0.628298044204712,
```

So not too bad imho @weissi

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: reordered this one

weissi:
@normanmaurer much better, 8%. I'm probably happy with this

normanmaurer:
@weissi you want me to try a list as well still ?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -234,6 +237,23 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
         return response
     }
 
+    private func bytesToForwardOnRemoval(ctx: ChannelHandlerContext) -> ByteBuffer? {
+        guard self.parser.upgrade == 1 && ctx.channel.isActive else {
+            return nil
+        }
+        if let buffer = self.cumulationBuffer?.slice(), buffer.readableBytes > 0 {
```
Lukasa: Out of interest, why are we slicing the cumulation buffer here?

weissi:
@normanmaurer up to you but I think it could be a follow-up. Because this fixes a correctness issue so I think we should get it in asap

normanmaurer:
@weissi sounds good to me... waiting for @Lukasa then :) Also please don't forget to approve ;)

Sources/NIOHTTP1/HTTPDecoder.swift
normanmaurer: @Lukasa I just thought it may be a bit nicer to consume. That said we not need to slice it. Happy to remove if you think we should

Lukasa: Unexplained slicing without a code comment is the kind of thing that we'll be real nervous to change in 6 months time. :D I think either remove it or put a comment explaining why it's there.

normanmaurer: added a comment ;)

normanmaurer:
I squashed, rebased and pushed again

PR #430, opened 2018-05-23
Only forward bytes on removal when we know we can handle it

normanmaurer:
Motivation:

398b950eeedb8215bfdafdc07ebf92a7e0722980 introduced a change which forwarded bytes that were buffered in HTTPDecoder after it was removed when an upgrade was detected. This may be risky if we dont know that the pipeline can handle it.

Modifications:

- Add new initializer that allow us to configure if we should forward bytes or not.
- Automatically use this if we know that we added an upgrader
- Add unit tests.

Result:

Less risky implementation.


weissi:
@normanmaurer doesn't compile, probably forgot to record something

normanmaurer:
@weissi lol yeah... stupid me :( PTAL again

normanmaurer:
@Lukasa if you prefer an enum please let me know and also suggest a name ;)

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -145,7 +145,16 @@ private protocol AnyHTTPDecoder: class {
...
-        self.init(type: HTTPServerRequestPart.self)
+        self.init(forwardAfterUpgrade: false)
+    }
+
+    /// Creates a new instance of `HttpRequestDecoder`.
+    ///
+    /// - parameters:
+    ///     - forwardAfterUpgrade: if `true` any remaining bytes that are currently buffered by the `HTTPRequestDecoder` will be forwarded
+    ///                            to the next handler in the pipeline once this decoder is removed from the pipeline and an upgrade was detected.
+    public convenience init(forwardAfterUpgrade: Bool) {
```
Lukasa: Let's have an `enum`.

```swift
enum RemoveAfterUpgradeStrategy {
    case forwardBytes
    case dropBytes
}
```

normanmaurer:
@Lukasa PTAL again... using an enum now.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -187,6 +195,14 @@ public final class HTTPResponseDecoder: HTTPDecoder<HTTPClientResponsePart>, Cha
     }
 }
 
+/// Strategy to use when a decoder is removed from the pipeline and an HTTP upgrade was detected before.
```
Lukasa: Nit: let's write this as "Strategy to use when a `HTTPDecoder` is removed from a pipeline after a HTTP upgrade was detected".

normanmaurer: addressed... will merge once green if I not hear back :)

Issue #426, opened 2018-05-22
Add method to `EventLoop` that is to `EventLoop.submit` as `Future.then` is to `Future.map`.


Lukasa:
Currently `EventLoop.submit` is defined like this: `func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>`.

The problem here is that `T` may be `EventLoopFuture<U>`. In this instance, you end up with `EventLoopFuture<EventLoopFuture<U>>` as the return type of `submit`, and double futures are annoying to work with.

It would be very helpful to have a method that can deal with this. To add this in a backward-compatible way, we could add the method to the `EventLoop` protocol, and also provide an extension with a default implementation that just does `.then { $0 }` on the returned double-future from the default method.

Lukasa:
Whoever does this should also clean up the workarounds from #424.

karim-elngr:
Any name suggestions for the new function?

weissi:
@karim-elngr really sorry for the _huge_ delay, this just slipped through. We just talked about it and `flatSubmit` (similar to `flatMap` and `map`) sounds like the best name to us.

weissi:
fixed in #1174

PR #435, opened 2018-05-24
Add eventloop submitasync

karim-elngr:
Implement #426

### Motivation:

Dealing with double futures is annoying when working with `EventLoop.submit(task)` when the task itself returns an `EventLoopFuture`, so we provide a new function to deal with tasks that return an `EventLoopFuture<T>`

### Modifications:

1- Add a new function  `func submitAsync<T>(_ asyncTask: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T>`

2- Clean up the workarounds from #424

### Result:

less `.then { $0 }` in code base

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -89,6 +89,13 @@ public protocol EventLoop: EventLoopGroup {
     ///     - task: The closure that will be submitted to the `EventLoop` for execution.
     /// - returns: `EventLoopFuture` that is notified once the task was executed.
     func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
+    
+    /// Submit a given async task to start execution. Once the execution is complete the given async task has been started and the returned `EventLoopFuture` will be notified with the result of execution once the task is finished.
+    ///
+    /// - parameters:
+    ///     - asyncTask: The closure that will be submitted to the `EventLoop` to start execution.
+    /// - returns: `EventLoopFuture` that is notified once the task was executed.
+    func submitAsync<T>(_ asyncTask: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T>
 
     /// Schedule a `task` that is executed by this `SelectableEventLoop` after the given amount of time.
     func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T>
@@ -190,6 +197,10 @@ extension EventLoop {
         }
 
         return promise.futureResult
+    }
+    
+    public func submitAsync<T>(_ asyncTask: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T> {
+        return submit(asyncTask).then { $0 }
     }
 
     /// Creates and returns a new `EventLoopPromise` that will be notified using this `EventLoop` as execution `Thread`.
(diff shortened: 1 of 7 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -89,6 +89,13 @@ public protocol EventLoop: EventLoopGroup {
     ///     - task: The closure that will be submitted to the `EventLoop` for execution.
     /// - returns: `EventLoopFuture` that is notified once the task was executed.
     func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
+    
+    /// Submit a given async task to start execution. Once the execution is complete the given async task has been started and the returned `EventLoopFuture` will be notified with the result of execution once the task is finished.
+    ///
+    /// - parameters:
+    ///     - asyncTask: The closure that will be submitted to the `EventLoop` to start execution.
+    /// - returns: `EventLoopFuture` that is notified once the task was executed.
+    func submitAsync<T>(_ asyncTask: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T>
```
normanmaurer: @karim-elngr couldn't this be just done as any extension ? I think there is no need to add it to the protocol itself.

normanmaurer:

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -396,4 +396,42 @@ public class EventLoopTest : XCTestCase {
...
+        
+        do {
+            _ = try future.wait()
+            XCTFail("should've thrown an error")
+        } catch _ as E {
+            /* good */
+        } catch let e {
+            XCTFail("error of wrong type \(e)")
+        }
+    }
```
weissi: can we have a test that shows that the future fails when the eventloop has been shut down?

weissi:
@Lukasa  I'm not in love with `submitFuture` but probably good enough.

@karim-elngr thanks, that looks good. if you grep for `wait().wait()` in the tests, you should find a lot of opportunities to introduce this in tests too :)

Tests/NIOTests/EventLoopTest.swift
karim-elngr: Yes sure

normanmaurer:
@Lukasa @weissi I really dislike `submitFuture`... Let me try to come up with a better name till tomorrow

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -396,4 +396,42 @@ public class EventLoopTest : XCTestCase {
...
+        }
+        
+        let future = submissionEventLoop.submitFuture(asyncTask)
+        
+        do {
+            _ = try future.wait()
+            XCTFail("should've thrown an error")
+        } catch _ as E {
+            /* good */
+        } catch let e {
```
BasThomas: Nit: you can omit the `let e` and access the error as `error` in the block by default.

BasThomas: It would become

```swift
} catch {
  XCTFail("error of wrong type \(error)")
}
```

Lukasa:
@normanmaurer Did you come up with a better name?

weissi:
@normanmaurer ping, we need the better name to unblock @karim-elngr

weissi:
@karim-elngr really sorry for the _huge_ delay, this just slipped through. We just talked about it and `flatSubmit` (similar to `flatMap` and `map`) sounds like the best name to us. Does `flatSubmit` sound good to you?

weissi:
closing in favour of #1174

PR #437, opened 2018-05-24
Don't provide promises when we don't care about results.

Lukasa:
Motivation:

When we don't care about the result of an operation, we shouldn't provide
a promise.

While I'm here I also removed an unnecessary close() call in the
AcceptHandler: the close() would be called again almost immediately.

Modifications:

Removed a close() call.
Set `promise: nil` on a pair of close() calls.

Result:

Fewer promise allocations.


weissi:
@Lukasa just as a question: This should allocate fewer promises, right? Did you check the allocation limits? Sorry should've asked that earlier but maybe we can do that post 1.7.2

Lukasa:
I think this promise removal was only on error paths, so it shouldn’t have affected our tests.

Issue #432, opened 2018-05-23
numOfThreads/numberOfThreads inconsistency


gmosx:
### Expected behavior

```
let group = MultiThreadedEventLoopGroup(threadCount: System.coreCount)
let threadPool = BlockingIOThreadPool(threadCount: 6)
```

### Actual behavior

```
let group = MultiThreadedEventLoopGroup(numThreads: System.coreCount)
let threadPool = BlockingIOThreadPool(numberOfThreads: 6)
```

Please notice the inconsistency between `numThreads` and `numberThreads`. Also, it would be better to use the `count` name used in the standard library.

normanmaurer:
@Lukasa @weissi WDYT... I am also ok with `count` but I think `numberThreads` or `numThreads` are a bit more descriptive... At least we should make it consistent :)

weissi:
I think that's a fair enough issue and I think we should fix `numThreads` at least. Whether the parameters should be `threadCount:` or `numberOfThreads:` or something can be bikeshed I don't really mind

gmosx:
I don't mind `numberOfThreads` but I think `numThreads` should be renamed.

PR #443, opened 2018-05-24
Rename numThreads to numberOfThreads parameter.

normanmaurer:
Motivation:

We should be consistent with naming and also choose descriptive names.

Modifications:

- Deprecate old init method that uses numThreads
- Add new init with numberOfThreads param name
- Everywhere use the new init

Result:

More consistent and descriptive naming. Fixes https://github.com/apple/swift-nio/issues/432.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -693,13 +693,22 @@ final public class MultiThreadedEventLoopGroup: EventLoopGroup {
         return lock.withLock { _loop }
     }
 
+    /// Creates a `MultiThreadedEventLoopGroup` instance which uses `numberOfThreads`.
+    ///
+    /// - arguments:
+    ///     - numberOfThreads: The number of `Threads` to use.
+    public convenience init(numberOfThreads: Int) {
+        let initializers: [ThreadInitializer] = Array(repeating: { _ in }, count: numberOfThreads)
+        self.init(threadInitializers: initializers)
+    }
+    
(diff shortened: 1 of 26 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -693,13 +693,22 @@ final public class MultiThreadedEventLoopGroup: EventLoopGroup {
         return lock.withLock { _loop }
     }
 
+    /// Creates a `MultiThreadedEventLoopGroup` instance which uses `numberOfThreads`.
+    ///
+    /// - arguments:
+    ///     - numThreads: The number of `Threads` to use.
```
BasThomas: This should be `numberOfThreads`, not `numThreads`.

Sources/NIO/EventLoop.swift
```diff
@@ -693,13 +693,22 @@ final public class MultiThreadedEventLoopGroup: EventLoopGroup {
...
+    public convenience init(numberOfThreads: Int) {
+        let initializers: [ThreadInitializer] = Array(repeating: { _ in }, count: numberOfThreads)
+        self.init(threadInitializers: initializers)
+    }
+    
     /// Creates a `MultiThreadedEventLoopGroup` instance which uses `numThreads`.
     ///
     /// - arguments:
     ///     - numThreads: The number of `Threads` to use.
+    @available(*, deprecated, renamed: "numberOfThreads")
```
Lukasa: Strictly this was renamed `init(numberOfThreads:)`.

PR #444, opened 2018-05-25
timerfd_settime(...) will only accept nsec up to 999999999.

normanmaurer:
Motivation:

timerfd_settime(...) will error out if nsec > then 999999999 is used.

Modifications:

Limit nsec to 999999999, even on kqueue to make it simple.

### DIFF
--- a/Sources/NIO/Selector.swift
+++ b/Sources/NIO/Selector.swift
@@ -25,7 +25,9 @@ private extension timespec {
         let nsecPerSec: TimeAmount.Value = 1_000_000_000
         let ns = amount.nanoseconds
         let sec = ns / nsecPerSec
-        self = timespec(tv_sec: Int(sec), tv_nsec: Int(ns - sec * nsecPerSec))
+        // Limit to 999999999 so it works for epoll and kqueue.
+        // See http://www.man7.org/linux/man-pages/man2/timerfd_create.2.html
+        self = timespec(tv_sec: Int(sec), tv_nsec: min(999999999, Int(ns - sec * nsecPerSec)))
     }
 }
 


Sources/NIO/Selector.swift
```diff
@@ -25,7 +25,9 @@ private extension timespec {
         let nsecPerSec: TimeAmount.Value = 1_000_000_000
         let ns = amount.nanoseconds
         let sec = ns / nsecPerSec
-        self = timespec(tv_sec: Int(sec), tv_nsec: Int(ns - sec * nsecPerSec))
+        // Limit to 999999999 so it works for epoll and kqueue.
+        // See http://www.man7.org/linux/man-pages/man2/timerfd_create.2.html
+        self = timespec(tv_sec: Int(sec), tv_nsec: min(999999999, Int(ns - sec * nsecPerSec)))
```
Lukasa: Should we silently do this, or should this fail loudly?

normanmaurer: I think doing this silently is "just fine" as its only for the nanoseconds and its ok to wakeup even a bit before IMHO.

Lukasa: @weissi thoughts?

normanmaurer: I mean even if you use a timed wakeup it may wakeup "spurious" for whatever reasons even if there is no adjusting needed.

weissi: Wait, how can we get more than 999_999_999?

normanmaurer: @weissi hmm let me think about if this is even possible in respect how we calculate this. I only noticed we did not have an guards and so added these.

weissi: @normanmaurer but it's an integer division by `1_000_000_000` so the remainder can't be larger than `999_999_999` right?

normanmaurer: right... closing it

PR #445, opened 2018-05-25
make _UInt24 and _UInt56 printable

weissi:
Motivation:

_UInt24 and _UInt56 didn't have a CustomStringConvertible conformance
which this patch is adding.

Modifications:

add CustomStringConvertible instances.

Result:

easier to debug stuff

### DIFF
--- a/Sources/NIO/IntegerTypes.swift
+++ b/Sources/NIO/IntegerTypes.swift
@@ -72,6 +72,12 @@ extension _UInt24: Equatable {
     }
 }
 
+extension _UInt24: CustomStringConvertible {
+    var description: String {
+        return Int(self).description
+    }
+}
+
 // MARK: _UInt56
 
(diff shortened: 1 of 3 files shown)


Sources/NIO/IntegerTypes.swift
```diff
@@ -135,3 +141,9 @@ extension _UInt56: Equatable {
         return lhs.b1234 == rhs.b1234 && lhs.b56 == rhs.b56 && lhs.b7 == rhs.b7
     }
 }
+
+extension _UInt56: CustomStringConvertible {
+    var description: String {
+        return Int(self).description
```
weissi: oops, this needs to be Int64 for 32bit platforms... will fix

weissi: @normanmaurer / @Lukasa done now

PR #464, opened 2018-06-02
document EmbeddedChannel

weissi:
Motivation:

We had no EmbeddedChannel documentation.

Modifications:

Add EmbeddedChannel documentation.

Result:

Less lacking documentation.

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -251,6 +251,37 @@ class EmbeddedChannelCore: ChannelCore {
     }
 }
 
+/// `EmbeddedChannel` is a `Channel` implementation that does neither any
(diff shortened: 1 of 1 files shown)


Sources/NIO/Embedded.swift
```diff
@@ -251,6 +251,37 @@ class EmbeddedChannelCore: ChannelCore {
     }
 }
 
+/// `EmbeddedChannel` is a `Channel` implementation that does neither do any
+/// actual IO nor has a proper eventing mechanism. The prime use-case for
```
Lukasa: Nit: s/does neither do any/does neither any/

weissi: thanks, done, sorry for the delay

PR #467, opened 2018-06-04
Support faster copying of potentially discontiguous data.

Lukasa:
Motivation:

Data and DispatchData have the potential of being discontiguous:
hiding the fact that they're made up of multiple contiguous chunks of
memory where each of those chunks is separated from the rest.

Currently these are copied in less-efficient mechanisms. Data is copied
via ContiguousCollection: while some Datas are, if the Data is
non-contiguous then ContiguousCollection will actually copy each byte
twice: once to make them contiguous, and then again to move them into the
ByteBuffer.

DispatchData, on the other hand, is treated like any other Sequence of
bytes, which potentially leads to extremely poor copying times, even
though a DispatchData could usually be copied in a small number of memcpy
calls.

This patch improves performance when writing these data types by using their
enumerateBytes methods to obtain a view of what chunks of their memory are
contiguous.

Modifications:

Defined a new CompositeCollection protocol.
Conformed Data and DispatchData to that protocol.
Implemented new ByteBuffer.set(compositeBytes:at:) and ByteBuffer.write(compositeBytes:at:) methods.

Result:

Improved performance when writing Data and DispatchData types.

### DIFF
--- a/Sources/NIO/ByteBuffer-aux.swift
+++ b/Sources/NIO/ByteBuffer-aux.swift
@@ -255,6 +255,20 @@ extension ByteBuffer {
         return written
     }
 
+    /// Write `compositeBytes`, a `CompositeCollection` of `UInt8` into this `ByteBuffer`. Moves the writer index forward by the number of bytes written.
+    /// This method is likely more efficient than the one operating on plain `Collection` as it will use `memcpy` to copy as many bytes as possible in one go.
+    ///
+    /// - parameters:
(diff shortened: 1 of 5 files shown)


weissi:
@Lukasa should we rewrite this with `DataProtocol`?

Lukasa:
Yup, probably.

weissi:
clearing the 2.0.0 milestone but a new version of this can land anytime, not major breaking

PR #471, opened 2018-06-05
Update http_parser

Lukasa:
Motivation:

Our copy of http_parser has become almost a year old now, and there are
both some bugfixes that improve correctness and some performance
improvements in http_parser that we should pick up.

Modifications:

- Updated http_parser
- Added tests to confirm we don't suffer from
    https://github.com/nodejs/http-parser/pull/432
- Added tests that we didn't get broken by SOURCE verb addition.

Result:

Shiny new http_parser!

### DIFF
--- a/Sources/CNIOHTTPParser/c_nio_http_parser.c
+++ b/Sources/CNIOHTTPParser/c_nio_http_parser.c
@@ -25,7 +25,6 @@
 #include <assert.h>
 #include <stddef.h>
 #include <ctype.h>
-#include <stdlib.h>
 #include <string.h>
 #include <limits.h>
 
@@ -53,6 +52,7 @@
 
 #define SET_ERRNO(e)                                                 \
 do {                                                                 \
+  parser->nread = nread;                                             \
   parser->http_errno = (e);                                          \
 } while(0)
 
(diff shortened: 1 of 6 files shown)


Lukasa:
The `http_parser` folks claimed a 15% performance improvement in their benchmark. Naturally we see less of that because our bench has way more code in it, but here's some perf. All of this is without pipelining.

Before:

```
wrk -c 400 -d 1m -t 2 --latency http://localhost:8888
Running 1m test @ http://localhost:8888
  2 threads and 400 connections
    Latency    19.19ms   10.43ms 559.46ms   96.01%
    Req/Sec    10.52k     0.86k   12.67k    76.92%
  Latency Distribution
     50%   18.00ms
     75%   20.77ms
     90%   24.12ms
     99%   37.87ms
  1256058 requests in 1.00m, 61.09MB read
  Socket errors: connect 0, read 276, write 2, timeout 0
Requests/sec:  20928.74
Transfer/sec:      1.02MB
```

After:

```
wrk -c 400 -d 1m -t 2 --latency http://localhost:8888
Running 1m test @ http://localhost:8888
  2 threads and 400 connections
    Latency    18.33ms    9.07ms 505.86ms   95.24%
    Req/Sec    10.99k   718.31    13.53k    75.67%
  Latency Distribution
     50%   17.35ms
     75%   19.91ms
     90%   22.91ms
     99%   33.98ms
  1312015 requests in 1.00m, 63.81MB read
  Socket errors: connect 0, read 269, write 3, timeout 0
Requests/sec:  21861.03
Transfer/sec:      1.06MB
```

Small improvement there, about 4.5%.

Lukasa:
All the latencies have improved too, just seems like a net win for us.

normanmaurer:
@Lukasa thanks!

Issue #474, opened 2018-06-08
Length-delimited frame decoder.


Lukasa:
It would be useful to have a general `ChannelInboundHandler` that can decode a length-delimited framing protocol. This is a protocol that has messages prefixed with an encoding of their length.

Such a protocol would look like this:

```
<fixed width integer n in network byte order><n bytes>
```

The simplest version of this would be one that uses a fixed-width integer.

For now we won't bother with varint support.

normanmaurer:
See https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/FixedLengthFrameDecoder.java

MrMage:
FYI, length-prefixed messages are also used in the gRPC protocol (see https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md). So with minor modifications this might also be helpful for a gRPC implementation on top of NIO-HTTP2 (see https://github.com/MrMage/grpc-swift/blob/4dd2609483a59c64ea67e622587ec6a9dff09f97/Tests/SwiftGRPCNIOTests/EchoServerTests.swift#L61 for my experiments with that).

jml:
Should this issue still be open, given that swift-nio-extras has a fixed-length frame decoder?

Lukasa:
I think so: this issue caused two PRs, including #484. We still want something like #484, so until that lands in swift-nio-extras we probably want to keep this open.

PR #479, opened 2018-06-08
Create FixedLengthFrameDecoder

FranzBusch:
Created the LengthDelimitedInboundHandler

### Motivation:

https://github.com/apple/swift-nio/issues/474


normanmaurer:
@swift-nio-bot test this please.

normanmaurer:
@Lukasa @weissi do we move these handlers to an extra module (as we will have soon a few of these) ? That said I think we can merge this one first and then move ourselves to keep things simple.

normanmaurer:

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// Create `FixedLengthFrameDecoder` with a given frame length.
+    ///
+    /// - parameters:
+    ///    - frameLength: The length of a frame.
+    public init(frameLength: Int) {
+        self.frameLength = frameLength
+    }
+
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        guard let slice = buffer.readSlice(length: frameLength) else {
```
weissi: we should do something here when the handler is removed and `buffer` is non-empty. Options that I can currently think of:

- fire the remaining bytes through the pipeline by implementing `handlerRemoved` as `ctx.fireChannelRead(self.wrapInboundOut(self.buffer)); ctx.fireChannelReadComplete()`
- fire an error that we have some left-over

CC @normanmaurer / @Lukasa

Tests/NIOTests/FixedLengthFrameDecoderTest+XCTest.swift
```diff
@@ -0,0 +1,34 @@
...
+/// NOTE: This file was generated by generate_linux_tests.rb
+///
+/// Do NOT edit this file directly as it will be regenerated automatically when needed.
+///
+
+extension FixedLengthFrameDecoderTest {
+
+   static var allTests : [(String, (FixedLengthFrameDecoderTest) -> () throws -> Void)] {
+      return [
+                ("testDecodeIfLessBytesAreSend", testDecodeIfLessBytesAreSend),
```
weissi: super minor but: this should be `...FewerBytesAreSent` not `...LessBytesAreSend`

Tests/NIOTests/FixedLengthFrameDecoderTest+XCTest.swift
```diff
@@ -0,0 +1,34 @@
...
+                ("testDecodeIfMoreBytesAreSend", testDecodeIfMoreBytesAreSend),
```
weissi: this should also be `Sent` not `Send`

Tests/NIOTests/FixedLengthFrameDecoderTest.swift
```diff
@@ -0,0 +1,55 @@
...
+import XCTest
+import NIO
+
+class FixedLengthFrameDecoderTest: XCTestCase {
+
+    public func testDecodeIfLessBytesAreSend() throws {
+        let channel = EmbeddedChannel()
+
+        let frameLength = 8
+        _ = try channel.pipeline.add(handler: FixedLengthFrameDecoder(frameLength: frameLength)).wait()
```
weissi: remove the `_ =` that's not needed as the type should be `Void` so you don't need to ignore it.

Tests/NIOTests/FixedLengthFrameDecoderTest.swift
```diff
@@ -0,0 +1,55 @@
...
+
+        var outputBuffer: ByteBuffer? = channel.readInbound()
+        XCTAssertEqual("xxxxxxxx", outputBuffer?.readString(length: frameLength))
+    }
+
+    public func testDecodeIfMoreBytesAreSend() throws {
+        let channel = EmbeddedChannel()
+
+        let frameLength = 8
+        _ = try channel.pipeline.add(handler: FixedLengthFrameDecoder(frameLength: frameLength)).wait()
```
weissi: same here, remove `_ = `

Tests/NIOTests/FixedLengthFrameDecoderTest.swift
```diff
@@ -0,0 +1,55 @@
...
+        XCTAssertTrue(try channel.writeInbound(buffer))
+
+        var outputBuffer: ByteBuffer? = channel.readInbound()
+        XCTAssertEqual("xxxxxxxx", outputBuffer?.readString(length: frameLength))
+
+        outputBuffer = channel.readInbound()
+        XCTAssertEqual("aaaaaaaa", outputBuffer?.readString(length: frameLength))
+
+        outputBuffer = channel.readInbound()
+        XCTAssertNil(outputBuffer?.readString(length: frameLength))
```
weissi: we should call `XCTAssertFalse(channel.finish())` here, same above

weissi:
@normanmaurer / @Lukasa yeah, we should make our mind up where we put these handlers. Extra module or maybe in the [`swift-nio-extras`](https://github.com/apple/swift-nio-extras) repo?

Sources/NIO/FixedLengthFrameDecoder.swift
Lukasa: Why not split the difference and fire an error that contains the left-over bytes?

weissi: @Lukasa yes, that's also what I think we should do. Maybe we should have a global 'left over stuff' error type with an associated ByteBuffer?

Lukasa: Sounds good to me.

ianpartridge:
This looks like a useful addition but I'm a bit confused - you reference #474 but that describes decoding a length-delimited framing protocol, not a fixed length frame decoder - they are different things. Probably the title and description of this PR needs updating?

Lukasa:
Yeah, the same issue was interpreted in two different ways. Conveniently, both are useful!

FranzBusch:
@weissi @Lukasa I updated the PR with the latest code review comments and implemented the `handlerRemoved` method which throws a new error `removedWithLeftOverBytes` if anything is left in the `cumulationBuffer`

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+
+///
+/// A decoder that splits the received `ByteBuffer` by a fixed number
+/// of bytes. For example, if you received the following four fragmented packets:
+///
+/// +---+----+------+----+
```
weissi: super minor (feel free to ignore): add 4 extra spaces before that so it gets code formatting

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,65 @@
...
+/// | A | BC | DEFG | HI |
+/// +---+----+------+----+
+/// 
+/// A `FixedLengthFrameDecoder` will decode them into the
+/// following three packets with the fixed length:
+///
+/// +-----+-----+-----+
```
weissi: super minor (feel free to ignore): add 4 extra spaces before that so it gets code formatting

Sources/NIO/ChannelPipeline.swift
```diff
@@ -989,11 +989,13 @@ private extension CloseMode {
 }
 
 /// `Error` that is used by the `ChannelPipeline` to inform the user of an error.
-public enum ChannelPipelineError: Error {
+public enum ChannelPipelineError: Error, Equatable {
```
weissi: unfortunately because of SemVer we cannot extend this `enum` with extra cases before 2.0.0 is released. Therefore please add a separate type for this error. It could literally just be

```public struct LeftOverBytesError: Error {
    public let leftOverBytes: ByteBuffer
}```

CC @normanmaurer / @Lukasa for input about the name

**Comment by FranzBusch** [comment]
@weissi Okay, moved the error to a separate struct. Was that what you had in mind?

File: Sources/NIO/FixedLengthFrameDecoder.swift [hunk 10]
```diff
@@ -0,0 +1,65 @@
...
+/// 
+/// A `FixedLengthFrameDecoder` will decode them into the
+/// following three packets with the fixed length:
+///
+///     +-----+-----+-----+
+///     | ABC | DEF | GHI |
+///     +-----+-----+-----+
+///
+public final class FixedLengthFrameDecoder: ByteToMessageDecoder {
+
```
Lukasa: Nit: remove this whitespace line.

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,65 @@
...
+        guard let slice = buffer.readSlice(length: frameLength) else {
+            return .needMoreData
+        }
+
+        ctx.fireChannelRead(self.wrapInboundOut(slice))
+        return .continue
+    }
+
+    public func handlerRemoved(ctx: ChannelHandlerContext) {
+        guard let buffer = cumulationBuffer, buffer.readableBytes > 0 else {
```
Lukasa: Phrasing this as `guard let` reads weirdly to me, because that's not the expected flow. It might be better to rephrase it as `if let`.

Lukasa:
I believe @weissi has further code review here.

weissi:
@Lukasa I think I'm good for now actually. Only question is if this should go into `NIO` itself or [`swift-nio-extras`](https://github.com/apple/swift-nio-extras). It feels more like a swift-nio-extras thing to me but your call.

Lukasa:
Yeah, let's move this over to `swift-nio-extras`. @FranzBusch do you mind?

PR #480, opened 2018-06-09
Replaced @_inlineable and @_versioned with @usableFromInline

jagtapl:
Motivation:

swfit build generated the deprecated warnings for @_inlineable and @_versioned

Modifications:

Replaced @_inlineable and @_versioned with @usableFromInline

Result:

swift builds with no warnings


Lukasa:
Unfortunately we cannot do this until we drop Swift 4.0 support, which we are not doing for the foreseeable future. Given that this patch will become out of date very quickly, I think we’ll close this. Thanks for the offer!

(As a note, this warning should go away soon.)

PR #484, opened 2018-06-09
Decode a length-delimited framing protocol.

blossomwoo:
Length-delimited frame decoder. (#474)

### Motivation:

It would be useful to have a general ChannelInboundHandler that can decode a length-delimited framing protocol. This is a protocol that has messages prefixed with an encoding of their length.

Such a protocol would look like this:

`<fixed width integer n in network byte order><n bytes>
`

The simplest version of this would be one that uses a fixed-width integer.

For now we won't bother with varint support.

### Modifications:

Added a length-delimited frame decoder.

### Result:

People can now decode length-delimited framing protocols.


Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,47 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public final class FixedLengthFrameDecoder<T: FixedWidthInteger>: ByteToMessageDecoder {
```
normanmaurer: I think we should rename this to `LengthFieldBasedFrameDecoder` as its not really fixes but depends on the length that is used to prefix the actual data.

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,47 @@
...
+                return .needMoreData
+            }
+            state = .data(Int(integer))
+            return .continue
+        case .data(let dataLength):
+            guard let bytes = buffer.readSlice(length: dataLength) else {
+                return .needMoreData
+            }
+            ctx.fireChannelRead(self.wrapInboundOut(bytes))
+            state = .len
```
normanmaurer: I think we should update the state before we call `ctx.fireChannelRead` to better protected of re-entrance.

normanmaurer: also please use `self.state` (that's what we usually do).

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    public init() { }
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) -> DecodingState {
+        switch state {
+        case .len:
+            guard let integer = buffer.readInteger(as: T.self) else {
+                return .needMoreData
+            }
+            state = .data(Int(integer))
```
normanmaurer: please use self.state

Sources/NIO/FixedLengthFrameDecoder.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    enum State {
+        case len
+        case data(Int)
+    }
+    
+    public init() { }
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) -> DecodingState {
+        switch state {
```
normanmaurer: self.state

Tests/NIOTests/CodecTest.swift
```diff
@@ -377,4 +377,76 @@ public class MessageToByteEncoderTest: XCTestCase {
         XCTAssertFalse(try channel.finish())
 
     }
+    
+    struct fixedLengthFrameDecoderHelper<T: FixedWidthInteger> {
```
normanmaurer: I think I would move the tests to an extra file as the codec itself is also in an extra file.

Tests/NIOTests/CodecTest.swift
```diff
@@ -377,4 +377,76 @@ public class MessageToByteEncoderTest: XCTestCase {
...
+        static fileprivate func fixedLengthFrameHelper(_ channel: EmbeddedChannel) throws {
+            _ = try channel.pipeline.add(handler: FixedLengthFrameDecoder<T>()).wait()
```
normanmaurer: Use `XCTAssertNotThrows(...)`

Tests/NIOTests/CodecTest.swift
```diff
@@ -377,4 +377,76 @@ public class MessageToByteEncoderTest: XCTestCase {
...
+            
+            for testString in testStrings {
+                channel.pipeline.fireChannelRead(NIOAny(buffer))
+                
+                guard var result: ByteBuffer = channel.readInbound() else {
+                    XCTFail("Failed to read buffer for \(T.self)")
+                    return
+                }
+                let testResult = result.readString(length: (result.readableBytes))
+                XCTAssert(testResult == testString, "String doesn't match.")
```
normanmaurer: Use `XCTAssertEqual(...)`

Tests/NIOTests/CodecTest.swift
```diff
@@ -377,4 +377,76 @@ public class MessageToByteEncoderTest: XCTestCase {
...
+            
+            let testStrings = ["try", "swift", "san", "jose"]
+            var buffer = channel.allocator.buffer(capacity:1024)
+            for testString in testStrings {
+                buffer.write(integer: T(testString.count))
+                buffer.write(string: testString)
+            }
+            
+            for testString in testStrings {
+                channel.pipeline.fireChannelRead(NIOAny(buffer))
```
normanmaurer: use `channel.writeInbound(...)` and assert the return value to replace `fireChannelRead` calls.

Tests/NIOTests/CodecTest.swift
```diff
@@ -377,4 +377,76 @@ public class MessageToByteEncoderTest: XCTestCase {
...
+                
+                guard var result: ByteBuffer = channel.readInbound() else {
+                    XCTFail("Failed to read buffer for \(T.self)")
+                    return
+                }
+                let testResult = result.readString(length: (result.readableBytes))
+                XCTAssert(testResult == testString, "String doesn't match.")
+            }
+        }
```
normanmaurer: call `channel.finish()` and assert return value.

Sources/NIO/FixedLengthFrameDecoder.swift
normanmaurer: Can you please add some docs, something like this:

https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/LengthFieldBasedFrameDecoder.java#L184

Just a bit simplified as this implementation always assume the length is on offset 0.

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+/// ````
+public final class LengthFieldBasedFrameDecoder<T: FixedWidthInteger>: ByteToMessageDecoder {
+    public typealias InboundIn = ByteBuffer
+    public typealias InboundOut = ByteBuffer
+    
+    public var cumulationBuffer: ByteBuffer?
+    
+    private var state: State = .len
+    
+    enum State {
```
weissi: can we make this `private enum State`?

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+        case len
```
weissi: I'd argue for making those cases more descriptive. Like `case waitingForLength` and `case waitingForPayload(Int)`. @normanmaurer / @Lukasa  what do you think?

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+        case data(Int)
+    }
+    
+    public init() { }
```
weissi: this definitely needs an upper bound parameter. Otherwise an attacker could trivially kill this server by sending a massive length.

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) -> DecodingState {
+        switch self.state {
+        case .len:
+            guard let integer = buffer.readInteger(as: T.self) else {
+                return .needMoreData
+            }
+            self.state = .data(Int(integer))
```
weissi: this needs to verify a few things:

- `> 0`
- `<= self.upperBound`

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,61 @@
...
+        case .data(let dataLength):
+            guard let bytes = buffer.readSlice(length: dataLength) else {
+                return .needMoreData
+            }
+            self.state = .len
+            ctx.fireChannelRead(self.wrapInboundOut(bytes))
+            return .continue
+        }
+    }
+}
```
weissi: we should do something when the handler is removed and `self.cumulationBuffer` is non-empty. @Lukasa / @normanmaurer I reckon we should just fire an error through the pipeline is the handler is removed and we're left with some partial message, no?

Tests/NIOTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,91 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+//
+import XCTest
+
+struct fixedLengthFrameDecoderHelper<T: FixedWidthInteger> {
```
weissi: - the type name should be upper case
- this helper type should go to the end of the file. In Xcode it's really helpful is the test class itself is at the top as it contains the 'run' icon for the whole test suite right next to the class. If it's not at the top it's always kind of hard to find the test class as you need to scroll and look for it...

Tests/NIOTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,91 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+//
```
weissi: please remove the `//` line

Tests/NIOTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,91 @@
...
+            let testResult = result.readString(length: (result.readableBytes))
+            XCTAssertEqual(testResult, testString)
+        }
+        XCTAssertFalse(try channel.finish())
+    }
+}
+
+public class LengthFieldBasedFrameDecoderTest: XCTestCase {
+    func testFixedLengthFrameDecoderInt8() throws {
+        let channel = EmbeddedChannel()
```
weissi: the `EmbeddedChannel` could just be created in `fixedLengthFrameHelper` helper function, no?

Tests/NIOTests/LengthFieldBasedFrameDecoderTest.swift
```diff
@@ -0,0 +1,91 @@
...
+                return
+            }
+            let testResult = result.readString(length: (result.readableBytes))
+            XCTAssertEqual(testResult, testString)
+        }
+        XCTAssertFalse(try channel.finish())
+    }
+}
+
+public class LengthFieldBasedFrameDecoderTest: XCTestCase {
```
weissi: we need tests for the edge cases, for example:

- negative lengths
- absurdly large lengths
- length and then before receiving the appropriate number of bytes we get an EOF
- EOF whilst waiting for length

blossomwoo: @weissi I'm having trouble with the last 2 tests. Could you take a look at `testEOFBeforeReceivingAppropriateNumberOfBytes` & `testEOFWhileWaitingForLength` and provide some pointers? Thanks!

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A `ChannelInboundHandler` used to decode length-delimited frames.
+///
+/// This is a protocol that has messages prefixed with an encoding of their length.
+/// Such a protocol would look like this:
+/// ````
```
weissi: - isn't this one backtick too many?
- super minor: could we have an empty line (just `///`) between the code blocks?

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,85 @@
...
+    private var state: State = .waitingForLength
+    
+    /// Errors thrown by the NIO LengthFieldBasedFrameDecoder module.
+    public enum NIOLengthFieldBasedFrameDecoderError: Error {
+        /// The frame being sent is negative length or larger than the configured maximum
+        /// acceptable frame size
+        case invalidFrameLength
+        
+        /// when the handler is removed and buffer is non-empty
+        case bytesLeftOver
```
weissi: this should have the bytes that were left over as associated data

Sources/NIO/LengthFieldBasedFrameDecoder.swift
```diff
@@ -0,0 +1,85 @@
...
+        self.upperBound = upperBound
+    }
+    
+    public func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        switch self.state {
+        case .waitingForLength:
+            guard let integer = buffer.readInteger(as: T.self) else {
+                return .needMoreData
+            }
+            guard integer > 0 && integer <= self.upperBound else {
```
weissi: this should be `>= 0`, no? I think `0` is fine no?

Tests/NIOTests/LengthFieldBasedFrameDecoderTest.swift
weissi: hi @blossomwoo sure, that's not obvious at all actually. Let me run you through what I think we need to do to get this in asap. We're really close!

If anything doesn't make sense, please just at-mention me and ask for help. And sorry this is quite a bit of stuff actually.

So first we should make some very minor changes to the actual code:

1) we should create a new `private func handleLeftOverData()` which contains the logic to fire the error if there's left-over stuff
2) we should call `self.handleLeftOverData()` from both `func handlerRemoved()` and `func channelInactive()`

The reason we should do it in both `channelInactive` (called when the channel gets inactive, in TCP that means the socket is closed) and `handlerRemoved` (called when the handler gets removed from the pipeline) is: In probably most cases, you'll see `channelInactive` followed by `handlerRemoved` because we remove all handlers because we tear the pipeline down. However, a handler might also be removed in normal operation in which case we would also like to fire the error. The easiest way of achieving this is to call the left-over logic in both. Note: we need to set the left-over bytes to nothing after firing the error.

regarding the tests, what probably works best is to create a new channel handler called something like

    private final class LeftOverCheckHandler: ChannelInboundHandler {
        typealias InboundIn = ByteBuffer
        typealias InboundOut = ByteBuffer

        enum LeftOverCheckHandlerError: Error {
            case wrongErrorReceived(Error)
        }

        private var haveSeenLeftOvers = false
        private let leftOversReceivedPromise: EventLoopPromise<ByteBuffer>

        init(leftOversReceivedPromise: EventLoopPromise<ByteBuffer >) {
            self.leftOversReceivedPromise = leftOversReceivedPromise
        }

        func errorCaught(ctx: ChannelHandlerContext, error: Error) {
            XCTAssertFalse(self.haveSeenLeftOvers)
            self.haveSeenLeftOvers = true

            switch error {
                case let error as NIOLengthFieldBasedFrameDecoderError:
                    switch error {
                        case .bytesLeftOver(let bytes):
                            self.leftOversReceivedPromise.succeed(value: bytes)
                        case .invalidFrameLength:
                            self.leftOversReceivedPromise.fail(error: LeftOverCheckHandlerError.wrongErrorReceived(error))
                    }
                  default:
                    self.leftOversReceivedPromise.fail(error: LeftOverCheckHandlerError.wrongErrorReceived(error))
                }
         }
     }

and you can just add this channel handler to the channel (using the following code)

```
let leftOversReceivedPromise: EventLoopPromise<ByteBuffer> = embedded.eventLoop.newPromise()
XCTAssertNoThrow(try embedded.pipeline.add(handler: LeftOverCheckHandler(leftOversReceivedPromise: leftOversReceivedPromise)).wait())
```

and now you can do a bunch of test cases:

1) just fire `channelInactive` through the pipeline manually: `embedded.pipeline.fireChannelInactive()` then verify that the left over bytes got received correctly by the `LeftOverCheckHandler`:

    XCTAssertEqual(whateverLeftOverBytesYouExpectAsByteBuffer,
                                 leftOversReceivedPromise.futureResult.wait()!)

2) call `embedded.close().wait()` and then spin the event loop (`embedded.eventloop.run()`) and then verify the same thing as in (1)
3) left the embedded channel up and running but remove the handler (`embedded.pipeline.remove(handler: theHandlerInstanceOfYourFixedLengthDecoderHandler)`) and then verify the same thing as in (1) and (2)

(disclaimer: I tested/compiled none of the code in here so there might be some minor mistakes, bugs, typos, etc)

Sources/NIO/Linux.swift
```diff
@@ -103,9 +103,9 @@ internal enum Linux {
     static let SOCK_NONBLOCK = Int32(bitPattern: Glibc.SOCK_NONBLOCK.rawValue)
```
weissi: please also remove all the changes in this file, I think Xcode's 'rename' function got properly confused here...

Sources/NIO/Selector.swift
```diff
@@ -297,13 +297,13 @@ final class Selector<R: Registration> {
```
weissi: please also remove all the changes in this file, I think Xcode's 'rename' function got properly confused here...

Lukasa:
I think this is now covered by apple/swift-nio-extras#17.

PR #486, opened 2018-06-12
Make tests work on 32-bit

helje5:
Ported test code to 32-bit Linux (Raspi)

### Motivation:

Nothing will work without tests!

### Modifications:

NIO has edge conditions wrt Int32.max. Mostly workaround this by using lower values, or adjusting types.
Also replaced some 0xdeadbeef w/ 0x1337beef in the tests to make stuff compile.

### Result:

The test suite runs through w/o hard crashes. One test doesn't run at all (testAllocationOfReallyBigByteBuffer), another one fails regularly.
```
Test Suite 'All tests' failed at 2018-06-12 12:55:45.411
Executed 711 tests, with 2 failures (0 unexpected) in 608.633 (608.633) seconds
```
But at least the suite runs through now!


Sources/NIO/ByteBuffer-int.swift
```diff
@@ -113,13 +113,19 @@ extension UInt32 {
 
         var n = self
 
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+            let max = UInt32(Int32.max)
+        #else
+            let max = UInt32.max
+        #endif
```
Lukasa: This change violates the doc comment up above, so we should probably change this.

Incidentally, I assume the reason for this change is that `malloc` accepts `Int` on 32-bit platforms, so returning UInt32.max is not a super great plan, so it may be worth changing the doc comment to say something like:

"Returns the next power of two unless that would overflow, in which case UInt32.max (on 64-bit systems) or Int32.max (on 32-bit systems) is returned. The returned value is always safe to be cast to `Int` and passed to `malloc` on all platforms."

Sources/NIO/Socket.swift
```diff
@@ -20,7 +20,14 @@ public typealias IOVector = iovec
 
     /// The maximum number of bytes to write per `writev` call.
     static var writevLimitBytes: Int {
-        return Int(Int32.max)
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+            // Note(hh): This is not a _proper_ fix, but necessary because
+            //           other places extend on that. Should be fine in
+            //           practice on 32-bit platforms.
+            return Int(Int32.max / 4)
```
Lukasa: Hrm, I don't love this. AFAICT this change works around the fact that we have a test that validates what happens if we attempt to write more than `writevLimitBytes`, which does some unguarded addition on this value.

For the test in `ChannelTests.swift`, it may be best to change it on 32-bit systems to create the buffer at `writevLimitBytes` and adjust the expected return values. That's the largest possible allocated buffer on 32-bit NIO anyway.

Sources/NIOWebSocket/WebSocketFrameEncoder.swift
```diff
@@ -16,7 +16,14 @@ import NIO
 
 private let maxOneByteSize = 125
 private let maxTwoByteSize = Int(UInt16.max)
-private let maxNIOFrameSize = Int(UInt32.max)
+#if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+    // Note(hh): This is not a _proper_ fix, but necessary because
+    //           other places extend on that. Should be fine in
+    //           practice on 32-bit platforms.
+    private let maxNIOFrameSize = Int(Int32.max / 4)
```
Lukasa: Any reason not to just set this to `Int(Int32.max )`? So far as I know we don't ever do math on this, so setting it to `Int32.max` gets the behaviour as close as reasonably possible to the 64-bit behaviour.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1073,22 +1073,37 @@ class ByteBufferTest: XCTestCase {
     }
 
     func testAllocationOfReallyBigByteBuffer() throws {
+        #if !arch(arm) // !32-bit, Raspi/AppleWatch/etc
```
Lukasa: Can we invert this conditional? It'd be nice to see the intended effect of the other branch closer to the top of the test.

Lukasa: I assume this conditional is intended to be temporary? If it's not, can we remove the tests for `arch(arm)` elsewhere in this test function, as they're unreachable.

Tests/NIOTests/ChannelTests.swift
```diff
@@ -169,9 +169,15 @@ public class ChannelTests: XCTestCase {
         for _ in 0..<bufferSize {
             buffer.write(staticString: "a")
         }
+
```
Lukasa: Mind removing this whitespace line?

Tests/NIOTests/ChannelTests.swift
```diff
@@ -237,8 +243,8 @@ public class ChannelTests: XCTestCase {
             }
             /* assert that the canary values are still okay, we should definitely have never written those */
             XCTAssertEqual(managed.last!.toOpaque(), Unmanaged.passUnretained(o).toOpaque())
-            XCTAssertEqual(0xdeadbeef, Int(bitPattern: iovecs.last!.iov_base))
-            XCTAssertEqual(0xdeadbeef, iovecs.last!.iov_len)
+            XCTAssertEqual(0x1337beef, Int(bitPattern: iovecs.last!.iov_base))
+            XCTAssertEqual(0x1337beef, iovecs.last!.iov_len)
```
Lukasa: Out of interest why does the old value prevent compilation?

Sources/NIO/Socket.swift
helje5: No, there are more NIO issues wrt that. I.e. ByteBuffer uses UInt32 and nextPoW2 to round up the size, which subsequently fails when casting backt to malloc, which wants an Int.
It simply is no good to live on that edge on 32-bit platforms. IMO it is not worthwhile to adjust all of NIO just for the 32-bit edge case. And you don't do similar edge-checks for 64-bit either (but stick to Int32.max ;-) )

Sources/NIOWebSocket/WebSocketFrameEncoder.swift
helje5: Same as above. Probably breaks when it travels to the ByteBuffer, which will use UInt32 and then fail.

Tests/NIOTests/ByteBufferTest.swift
helje5: Yes. This fails in a specific ARM memcpy, I hope that we get a fix for that eventually. I raised this in the ARM-community slack. The checks within make the test working until that memcpy error. (maybe that memcpy has additional size or alignment restrictions, no idea yet).

Tests/NIOTests/ChannelTests.swift
helje5: Int(0xdeadbeef) (which the compiler generates for literals) overflows `Int32` (which then asserts). The other option which I think I actually used in the original patch is `0xdeadbeef as UInt` or sth like that. I don't mind either.

helje5:
Just noticed that this issue is appropriately numbered: #486

Sources/NIO/Socket.swift
Lukasa: > ByteBuffer uses UInt32 and nextPoW2 to round up the size, which subsequently fails when casting backt to malloc

Doesn't this patch change that?

Tests/NIOTests/ChannelTests.swift
Lukasa: Might be better to use that other version, just for consistency with ourselves (we use `deadbeef` as our marker value elsewhere in the code)

Sources/NIO/Socket.swift
helje5: I tracked down this specific thing, yes (but stuff kept failing). What I was trying to say is that it probably makes no sense to support that max-addressable-memory edge-case in NIO (generally). If you *really* want to, I suggest starting out by writing tests that check against Int64.max on 64-bit too. It is not worth it, IMO.

helje5:
OK, I incorporated the easy changes requested, but I don't have the time to harden NIO for the max-mem edge case. So your choice whether you want no tests or all tests with no memory-edge-case like on 64-bit. Or maybe someone else steps in.

helje5:
Do not merge yet, I may have to revert another deadbeef to 1337beef, tests coredump again.

helje5:
OK, done. Runs through again.

weissi:
@helje5 I find it slightly amusing that your patch which makes the 32 bits tests run is #486 :D

Sources/NIO/ByteBuffer-int.swift
weissi: @helje5 / @Lukasa wait. malloc takes `size_t` which is unsigned, so it should be `CSize` == `UInt` imho. So I don't quite understand this change

helje5:
You are late, scroll [upwards](https://github.com/apple/swift-nio/pull/486#issuecomment-396645354):

> Just noticed that this PR is appropriately numbered: #486 

My original PR #383 also just missed the proper primary-key, I need to be more prudent and consider the details before submitting PRs. Sorry about that!

Lukasa:
Ok, I think I'm happy. @weissi want to review?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -332,7 +332,7 @@ public struct ByteBuffer {
                 self._slice = _ByteBufferSlice(_slice.lowerBound ..< self._storage.capacity)
             }
         }
-        assert(self._slice.lowerBound + index + capacity <= self._slice.upperBound)
+        assert(self._slice.lowerBound + index + capacity <= self._slice.upperBound) // TODO(hh): fails on 32-bit
```
normanmaurer: Remove the TODO ?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -231,7 +231,7 @@ public struct ByteBuffer {
         }
 
         private static func allocateAndPrepareRawMemory(bytes: Capacity, allocator: Allocator) -> UnsafeMutableRawPointer {
-            let bytes = Int(bytes)
+            let bytes = Int(bytes) /* Note(hh): this can fail on 32-bit (Int32 vs UInt32 Capacity)) */
```
normanmaurer: remove the Note ?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1073,22 +1073,37 @@ class ByteBufferTest: XCTestCase {
...
-        XCTAssertEqual(AllocationExpectationState.mallocDone, testAllocationOfReallyBigByteBuffer_state)
-        XCTAssertGreaterThanOrEqual(buf.capacity, Int(Int32.max))
-
-        buf.set(bytes: [1], at: 0)
-        /* now make it expand (will trigger realloc) */
-        buf.set(bytes: [1], at: buf.capacity)
-
-        XCTAssertEqual(AllocationExpectationState.reallocDone, testAllocationOfReallyBigByteBuffer_state)
-        XCTAssertEqual(buf.capacity, Int(UInt32.max))
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
```
weissi: instead of indenting everything, can we just do

```
#if arch(arm)
// this test fails on 32 bit platforms
return
#endif
```

Tests/NIOTests/ChannelTests.swift
```diff
@@ -169,9 +169,15 @@ public class ChannelTests: XCTestCase {
         for _ in 0..<bufferSize {
             buffer.write(staticString: "a")
         }
+        
 
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+            let lotsOfData = Int(Int32.max / 8)
+        #else
+            let lotsOfData = Int(INT32_MAX) // TBD(hh): should that be Int32.max?
```
weissi: please use `Int32.max`

Tests/NIOTests/ChannelTests.swift
```diff
@@ -169,9 +169,15 @@ public class ChannelTests: XCTestCase {
         for _ in 0..<bufferSize {
             buffer.write(staticString: "a")
         }
+        
 
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+            let lotsOfData = Int(Int32.max / 8)
```
weissi: why do we use less data here? Is that documented to be the maximum?

Tests/NIOTests/ChannelTests.swift
weissi: I'd propose `0xdedbeef` or `0xdadbeef` or smth

Sources/NIO/ByteBuffer-core.swift
weissi: I'm also not sure why this is an `Int`. `malloc` takes `size_t` which is a `UInt`

Sources/NIO/ByteBuffer-core.swift
weissi: @helje5 why does this fail on 32-bit ? If that fails, the code above doesn't work

Sources/NIO/ByteBuffer-core.swift
weissi: ping @helje5

Sources/NIO/ByteBuffer-core.swift
helje5: Probably was a temporary note for myself and I fixed it on the calling side, need to check.

Sources/NIO/ByteBuffer-core.swift
helje5: Just checked: `size_t` is `unsigned int` on the 32-bit RasPi Ubuntu (I kinda assumed that it might be `signed int` because quite often you can't actually alloc more than 2GB on 32-bit platforms).

It probably did fail when one of the max-out tests passed in 0xDEADBEEF, which will fail the `Int(bytes)` conversion.

Why you are using `Int` I don't know, I didn't introduce this. But your code says:
```
let sysMalloc: @convention(c) (Int) -> UnsafeMutableRawPointer? = malloc
```
I don't know, maybe you should just use `size_t`?

Tests/NIOTests/ByteBufferTest.swift
helje5: I don't care if you prefer that, but visually an indent makes sense, IMO. Sometimes people do a 4-indent for regular Swift code and indent the `#` stuff by 2, which I also kinda like.

Tests/NIOTests/ChannelTests.swift
helje5: Why do you use `Int32.max` instead of `Int64.max` on 64-bit? Same reason ;-)

Sources/NIO/ByteBuffer-core.swift
helje5: Moved this one to: #499.

Tests/NIOTests/ByteBufferTest.swift
helje5: Oh, you are saying I should return early, not remove the actual indent. But this will produce a Swift compiler warning wrt unreachable code.

helje5:
Not sure why `Merging is blocked` now. I updated the `master` and pulled it in ...

Tests/NIOTests/ByteBufferTest.swift
weissi: ahh, ok, then leave this as is.

helje5:
I think this should be good now? I don't know how/why GitHub is confused.

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -105,21 +105,27 @@ extension FixedWidthInteger {
...
+    /// Returns the next power of two unless that would overflow, in which case UInt32.max (on 64-bit systems) or Int32.max (on 32-bit systems) is returned. The returned value is always safe to be cast to Int and passed to malloc on all platforms.
     public func nextPowerOf2ClampedToMax() -> UInt32 {
         guard self > 0 else {
             return 1
         }
 
         var n = self
 
+        #if arch(arm) // 32-bit, Raspi/AppleWatch/etc
+            let max = UInt32(Int32.max)
```
weissi: this is maybe nit-picking but we'd like it to be `Int.max` (yes, same as `Int32.max` on 32-bit platforms) because really we'd be okay with it being larger but Swift uses the `Int` type so maybe it's clearer to users? I'm okay with leaving that too though. @Lukasa

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -105,21 +105,27 @@ extension FixedWidthInteger {
 }
 
 extension UInt32 {
-    /// Returns the next power of two unless that would overflow in which case UInt32.max is returned.
+    /// Returns the next power of two unless that would overflow, in which case UInt32.max (on 64-bit systems) or Int32.max (on 32-bit systems) is returned. The returned value is always safe to be cast to Int and passed to malloc on all platforms.
```
weissi: mind breaking that line in two so it's doesn't go over 120 characters ideally?

Sources/NIO/Socket.swift
weissi: @helje5 I don't understand why `Int32.max` doesn't work here. It fits into an `Int` on both 32 and 64 bit platforms.

Sources/NIOWebSocket/WebSocketFrameEncoder.swift
weissi: we should fully understand this. I think `Int32.max` should just work

Tests/NIOTests/ChannelTests.swift
weissi: but we're only allocating 2MB here, right? We'll write the same buffer over and over again, this should work. The sizes were chosen so it will exhaust both the number of bytes limit (IIRC 2GB) as well as the number of chunks (1024) you're allowed to give to `writev`. If we think this isn't worthwhile, then we can lower the number for all platforms. CC @Lukasa

Sources/NIO/ByteBuffer-int.swift
```diff
@@ -105,21 +105,27 @@ extension FixedWidthInteger {
...
+        #endif
+
         n -= 1
         n |= n >> 1
         n |= n >> 2
         n |= n >> 4
         n |= n >> 8
         n |= n >> 16
-        if n != .max {
+        if n != max {
```
weissi: do we actually have a test that on 32-bit systems tests that this returns the right value?

Sources/NIO/ByteBuffer-int.swift
Lukasa: Fine with changing this to `Int.max`

Tests/NIOTests/ChannelTests.swift
Lukasa: @weissi is correct, the purpose of this is to validate that we correctly handle having more data to pass to `writev` than `writev` will accept. As @weissi notes, we only actually allocate one buffer, so there is no risk.

helje5: Oh, I see. You just repeatedly write the 2MB buffer. Maybe the overflow happens because of `<=`? I guess I need to check this one on a Raspi again.

Sources/NIO/ByteBuffer-int.swift
helje5: I'm not entirely sure about this. You want `UInt32(Int.max)`? That would fail on 64-bit platforms? This really only is for the `UInt32(Int32.max)` case, and it makes sense to be explicit here, no?

Sources/NIO/ByteBuffer-int.swift
helje5: There is no test for this on 64-bit either. /shrug

Sources/NIO/ByteBuffer-int.swift
helje5: This was probably to support the `sysMalloc(Int)` issue. Looks like you changed that upstream to `sysMalloc(size_t)`, so it may not be necessary anymore.

Sources/NIO/Socket.swift
helje5:
```
            // Note(hh): This is not a _proper_ fix, but necessary because
            //           other places extend on that. Should be fine in
            //           practice on 32-bit platforms.
```
Well, it is in the comment above it (and in my reply above yours). It is not a proper fix, but it is a lot of work to track down all edge cases (for little gain).

P.S.: *Maybe* a lot of that went away when you changed sysMalloc(Int) to sysMalloc(size_t).

Sources/NIO/ByteBuffer-int.swift
weissi: that's bad, we should definitely have one

Sources/NIO/ByteBuffer-int.swift
weissi: yes `UInt32(Int.max)` would be good. I want it to fail if it's compiled on 64-bits. This thing should only be there for 32-bit platforms

Tests/NIOTests/ChannelTests.swift
weissi: it'd be awesome if you could validate why it fails. There must be something else wrong.

Sources/NIO/Socket.swift
weissi: @helje5 could we re-check this?

weissi:
@helje5 let's get this merged. If you address the remaining comments then we can get this through really quickly

weissi:

helje5:
@weissi I'm a little busy with other stuff. What exactly is still missing?

Tests/NIOTests/ChannelTests.swift
helje5: Well, see my () comment:
> internally it still has to spool up Int32.max of buffers, which won't work out

Sources/NIO/Socket.swift
weissi: @helje5 if you could just re-check that this is still the case. And then just remove the two `note(hh)` comments with a good explanation why we need (if we need) special casing.

weissi: if that's done we can merge I think.

weissi:

PR #488, opened 2018-06-19
Implement a repeated scheduling mechanism

FranzBusch:
Create a way to schedule repeated tasks

Motivation:
I think it would be nice to have a method to repeatedly schedule a task, this implementation just provides the basics to do so, but is definitely not the final solution. In my opinion, we should provide a method to schedule a task with at least these parameters:

- delay
- interval
- numberOfTimes

The other part I was thinking about is that with my current implementation the scheduling is done by an `EventLoop` but you could as well provide a method on `RepeatedTask` like `func begin(on: EventLoop)` which would allow it to be scheduled on multiple `EventLoop`s at a time, which I don't know if it's thread safe?

I just wanted to get early input if my approach is going in the right direction and if you have any additional requirements for such an implementation.


Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,43 @@ public struct Scheduled<T> {
...
 }
 
+public class RepeatedTask {
+    private let interval: TimeAmount
+    private var scheduled: Scheduled<Void>? {
+        didSet {
+            reschedule()
+        }
+    }
+    private let task: () throws -> Void
```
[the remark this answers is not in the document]
FranzBusch: I was thinking about this as well and I think there should be 2 different options one where a task is scheduled in an interval and one where there is a fixed delay after each task. I think both cases could make sense depending on what the user actually wants to achieve

weissi:
@Lukasa yes, they do continue accepting work (that they'll never run). The issue is that outside of the event loop thread you can't query the lifecycle state. We could and should probably fix this though. We could have another variable (like `scheduledTasks`) that mirrors (parts of) the `lifecycleState` but is readable from any thread when holding `tasksLock`.

Lukasa:
So the main problem with them accepting work is that it means that they'll get stuck in a reference cycle and never be freed. Any scheduled task holds a future that fires on that event loop when the task is completed. This future keeps the loop alive. This is already pretty bad, but a repeated task essentially guarantees an event loop leak because it repeatedly reschedules work, ensuring there is always an event loop reference alive.

FranzBusch:
So this is blocked until we adopt the EventLoop to not be retained by scheduledTasks right?

helje5:
Maybe I don't get the issue, but if I schedule a repeating timer, I most definitely would want the event loop to stay alive and not go away (and presumably my timer with it). No?

FranzBusch:
I think it is quite reasonable to only have the repeating task scheduled as long as the event loop is actually running since the task should not be responsible for the lifetime of the eventloop right?

Lukasa:
> Maybe I don't get the issue, but if I schedule a repeating timer, I most definitely would want the event loop to stay alive and not go away (and presumably my timer with it). No?

The problem is not that the event loop will go away, it's that the event loop won't tell the repeating task that it will never execute again. That is, if you do this:

```swift
loop.scheduleRepeatedTask(interval: .seconds(10)) { print("Still alive") }
sleep(1000)
loop.syncShutdownGracefully()
```

Then after 1000 seconds the repeating task will stop executing (as the loop is closed), but the repeating task will never be told. Worse, the loop will hold a reference to the scheduled task item, which holds a reference to the loop, causing the loop to be leaked.

This is not the end of the world in production applications because generally speaking they do not stop and start loops regularly, but it's very annoying in tests, as any use of repeating tasks will cause each test to leak the loops unless the user very aggressively cleans up their repeating tasks.

Put another way, in an ideal world the task would be told when the event loop has gone away, and any attempt to schedule new work would fail. Neither of those things happens in NIO today: when we shut the loop down we leave any pending tasks pending, and those pending tasks almost always keep the event loop alive.

I don't know that this blocks the proposal, as the problem already exists today with scheduled tasks, but repeating tasks take this possibility and make it an inevitability, which is not necessarily an improvement.

FranzBusch:
So would a fix be as simple as removing all scheduled task while shutting down?

Lukasa:
@FranzBusch That's not quite sufficient, because those shutdown tasks are likely to be necessary for some parts of cleanup. We probably need to design our way out of this, likely by using more than one task queue.

FranzBusch:
I just updated the PR it now includes methods to schedule a task with a fixedRate or fixedDelay. Furthermore, I removed the retain cycle created by the repeated task. Documentation, tests, and two methods which take a closure with a `Void` return parameter as still left to do.
@weissi  and @Lukasa can you have a quick look if the implementation seems good so far?

weissi:
@lukasa & @FranzBusch do we actually need `fixedRate`? If we have an overloaded event loop and this keeps firing it will lead to the EL becomig more overloaded.
Also I’m sure sure I like the names (`fixedRate` vs `fixedDelay`) too much. I think we could use long and descriptive names. But ideally we would only have what we call fixedDelay today.

FranzBusch:
@weissi For my personal project, I only used fixedRate as it was used for a polling mechanism and should fire every second even though the last invocation may still be processing. I get your point that we may overload the EL if the triggered task is being queued again and again, but I see use cases where fixed rate makes sense. 
Regarding the naming, I had a look at the netty implementation and went according to that, though I am open for better suggestions :)

weissi:
@FranzBusch makes sense. If you really want that, why don't you just use `eventloop.execute { myActualWork }; return eventloop.newSucceededFuture(value: ())` or similar to perform the other stuff? Then it will still fire every second even those `myActualWork` might still be going on?

FranzBusch:
@weissi Maybe, I don't get your example correctly but how would a call to `execute` repeatedly schedule a task and how would I cancel such a task sometimes in the future?

weissi:
@FranzBusch Sorry, I wasn't quite so clear: My assumption is that `.fixedDelay` and `.fixedRate` are the same if the work item that is scheduled doesn't take any time to execute (ie. succeeds the promise immediately). If that assumption is correct, you can just use `.fixedDelay` and _make_ the work item take no time but enqueueing the real work item using `el.execute`. Does that make sense?

Sources/NIO/EventLoop.swift
```diff
@@ -222,6 +291,66 @@ extension EventLoop {
...
+
+    /// Schedule a repeated task to be executed by the `EventLoop` with fixed delay between the end and start of tasks.
+    ///
+    /// - parameters:
+    ///     - initialDelay: The delay after which the task is initally executed.
+    ///     - delay: The fixed delay between the end of one task and the start of the next.
+    ///     - task: The closure that will be executed.
+    /// - return: `RepeatedTask`
+    @discardableResult
+    public func scheduleTaskWithFixedDelay(initialDelay: TimeAmount, delay: TimeAmount, _ task: @escaping () throws -> EventLoopFuture<Void>) -> RepeatedTask {
```
weissi: we need a test for this

Sources/NIO/EventLoop.swift
```diff
@@ -222,6 +291,66 @@ extension EventLoop {
...
+
+    /// Schedule a repeated task to be executed by the `EventLoop` with fixed delay between the end and start of tasks.
+    ///
+    /// - parameters:
+    ///     - initialDelay: The delay after which the task is initally executed.
+    ///     - delay: The fixed delay between the end of one task and the start of the next.
+    ///     - task: The closure that will be executed.
+    /// - return: `RepeatedTask`
+    @discardableResult
+    public func scheduleTaskWithFixedDelay(initialDelay: TimeAmount, delay: TimeAmount, _ task: @escaping () throws -> Void) -> RepeatedTask {
```
weissi: we need a test for this

Sources/NIO/EventLoop.swift
```diff
@@ -222,6 +291,66 @@ extension EventLoop {
...
+
+    /// Schedule a repeated task to be executed by the `EventLoop` in a fixed rate between the start of tasks.
+    ///
+    /// - parameters:
+    ///     - initialDelay: The delay after which the task is initally executed.
+    ///     - period: The period between the start of tasks.
+    ///     - task: The closure that will be executed.
+    /// - return: `RepeatedTask`
+    @discardableResult
+    public func scheduleTaskAtFixedRate(initialDelay: TimeAmount, period: TimeAmount, _ task: @escaping () throws -> EventLoopFuture<Void>) -> RepeatedTask {
```
weissi: - we need a test for this
- this shouldn't be `throws` (as it's returning an `EventLoopFuture`)

Sources/NIO/EventLoop.swift
```diff
@@ -222,6 +291,66 @@ extension EventLoop {
...
+    ///
+    /// - parameters:
+    ///     - initialDelay: The delay after which the task is initally executed.
+    ///     - period: The period between the start of tasks.
+    ///     - task: The closure that will be executed.
+    /// - return: `RepeatedTask`
+    @discardableResult
+    public func scheduleTaskAtFixedRate(initialDelay: TimeAmount, period: TimeAmount, _ task: @escaping () throws -> Void) -> RepeatedTask {
+        let futureTask: () throws -> EventLoopFuture<Void> = {
+            try task()
```
weissi: - this should catch the error and make it a failed future

FranzBusch:
@weissi Thanks, for clearing that up! This makes sense for work which succeeds the promise immediately but in both of the implemented methods that cannot be assumed. Furthermore, `execute` is just an immediate scheduling mechanism and my intention with the implementation was to provide a convenient way to schedule any work repeating work with a delay and a rate/delay mode so that users don't have to implement that behaviour all on their own. Of course, you guys still decide if one of these methods actually endangers overloading the EL and thus should not be provided.
Or was your intention to actually change the implementation side of the repeated scheduling?

FranzBusch:
Just saw your Code example and I get what you mean. Yes your solution should be equal to mine :)

FranzBusch:
@weissi Just updated the PR and removed the fixed rate scheduling and added some tests. Is there something else I should test for?

About the fixed rate scheduling why does netty have the possibility and swift-nio shouldn't? Can't the ELs in netty get overloaded as well?

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
...
+        self.delay = interval
+        self.task = task
+    }
+
+    /// Try to cancel the execution of the repeated task.
+    ///
+    /// Whether the execution of the task is immediately canceled depends on whether the execution of a task has already begun.
+    ///  This means immediate cancellation is not guaranteed.
+    public func cancel() {
+        scheduled?.cancel()
```
weissi: please prefix those (and the following) by `self.` (won't repeat this comment for all the other methods :) )

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -91,6 +91,64 @@ public class EventLoopTest : XCTestCase {
         XCTAssertFalse(ran.load())
     }
 
+    public func testScheduleRepeatedTask() throws {
+        let nanos = DispatchTime.now().uptimeNanoseconds
+        let initialDelay: TimeAmount = .milliseconds(50)
+        let delay: TimeAmount = .milliseconds(100)
```
weissi: can we maybe make this `5` and `10` milliseconds? Otherwise this test alone will run almost 1s which is quite slow

FranzBusch:
@weissi Done :)

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
     }
 }
 
+public class RepeatedTask {
```
weissi: this needs to be `final`

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
...
+    private let delay: TimeAmount
+    private var scheduled: Scheduled<EventLoopFuture<Void>>? {
```
weissi: oh sorry, didn't notice that both of those are `var`. Did you run the tests with [TSan (thread sanitizer)](https://www.shinobicontrols.com/wp-content/uploads/2016/08/Enable_Sanitizer.png) enabled? I'm sure TSan will find the threading error here.

The assumption (and that should actually be documented) is that the user can `cancel` a `RepeatedTask` from any thread but this isn't the case here.

Sources/NIO/EventLoop.swift
weissi: and have doc comments

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
...
+            }
+            if case EventLoopError.shutdown = error {
+                return
+            }
+            self.reschedule0()
+        }
+    }
+
+    private func reschedule0() {
+        guard let task = task else {
```
weissi: `self.` missing

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
...
+            return
+        }
+        self.scheduled = self.scheduled?.futureResult.eventLoop.scheduleTask(in: delay, task)
```
weissi: `self.` missing

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,61 @@ public struct Scheduled<T> {
...
+            }
+        }
+        self.scheduled?.futureResult.whenFailure { error in
+            if case EventLoopError.cancelled = error {
+                return
+            }
+            if case EventLoopError.shutdown = error {
+                return
+            }
+            self.reschedule0()
```
weissi: what other errors could we hit here?

FranzBusch: There could be an EventLoopError.unsupportedOperation, EventLoopError.shutdownFailed or any of the users thrown errors right? Should I handle the two EventLoopError cases?

FranzBusch:
@weissi I ran the tests with TSan and noticed multiple data races. One was just an error in the tests and I used an Atomic counter. The other was because of a cancel on a different thread as you described. To fix this I introduced a lock in the `RepeatedTask`. Could you please take a look if the synchronization looks good?

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -91,6 +91,89 @@ public class EventLoopTest : XCTestCase {
...
+        let initialDelay: TimeAmount = .milliseconds(5)
+        let delay: TimeAmount = .milliseconds(10)
+        var eventLoopGroup: EventLoopGroup? = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        weak var weakEventLoop = eventLoopGroup?.next()
+
+        eventLoopGroup?.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in }
+        XCTAssertNoThrow(try eventLoopGroup?.syncShutdownGracefully())
+        XCTAssertNotNil(weakEventLoop)
+        eventLoopGroup = nil
+        Thread.sleep(forTimeInterval: 0.01)
```
FranzBusch: I had to introduce a sleep here otherwise this test was really flaky and sometimes the `weakEventLoop` wasn't nil. Do you have any idea why that could be the case?

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,67 @@ public struct Scheduled<T> {
...
 }
 
+/// Returned once a task was scheduled to be repeatedly executed on the `EventLoop`.
+///
+/// A `RepeatedTask` allows the user to `cancel()` the repeated scheduling of further tasks.
+public final class RepeatedTask {
+    private let delay: TimeAmount
+    private var scheduled: Scheduled<EventLoopFuture<Void>>?
+    private var task: (() throws -> EventLoopFuture<Void>)?
+    private let lock = Lock()
```
Lukasa: I don't think we want to use a lock here: instead, we should use the event loop that this `RepeatedTask` belongs to to manage synchronization.

The TL;DR for this is that locking is hard, and right now this isn't handling it very well. In particular, `begin` does not handle the synchronisation correctly. The other advantage is that with using the parent event loop we can easily check whether we are already holding the "lock" and so provide a fast-path that does not involve cross-loop dispatch.

FranzBusch: I was thinking about the `begin` method as well but since this is only used internally and the user can only call cancel after `reschedule` was triggered once it should be fine. 

Would synchronization over the EL work with `executing` begin, reschedule and cancel on the EL? @Lukasa

Lukasa: Yup, should work fine. In fact, only `cancel` should require the explicit synchronisation: both `begin` and `cancel` should by definition always be called on the loop, and so can just degenerately `assert` that that's what's actually happening.

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,87 @@ public struct Scheduled<T> {
...
+        if self.eventLoop.inEventLoop {
+            self.begin0(in: delay)
+        } else {
+            self.eventLoop.execute {
+                self.begin0(in: delay)
+            }
+        }
+    }
+
+    private func begin0(in delay: TimeAmount) {
```
Lukasa: This function should assert it's in the event loop.

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,87 @@ public struct Scheduled<T> {
...
+            self.cancel0()
+        } else {
+            self.eventLoop.execute {
+                self.cancel0()
+            }
+        }
+    }
+
+    private func cancel0() {
+        assert(eventLoop.inEventLoop)
```
Lukasa: Nit: `self`

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,87 @@ public struct Scheduled<T> {
...
+        self.scheduled?.cancel()
+        self.scheduled = nil
+        self.task = nil
+    }
+
+    private func reschedule() {
+        assert(eventLoop.inEventLoop)
```
Lukasa: Nit: `self`

Lukasa: Nit: We could `guard let` here to clean up the unwrapping code a bit.

Sources/NIO/EventLoop.swift
Lukasa: I don't think it can. If you `schedule` a function that returns a `Future`, the user thrown errors from that function cannot appear in the outer future.

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,87 @@ public struct Scheduled<T> {
...
+            }
+            if case EventLoopError.shutdown = error {
+                return
+            }
+            self.reschedule0()
+        }
+    }
+
+    private func reschedule0() {
+        assert(eventLoop.inEventLoop)
```
Lukasa: Nit: `self`

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,87 @@ public struct Scheduled<T> {
...
+        guard let task = self.task else {
```
Lukasa: May as well also unwrap `self.scheduled` here too.

Sources/NIO/EventLoop.swift
Lukasa: I think it may be fair to say that if the scheduled `futureResult` fails, there is no reason to reschedule, right?

FranzBusch: Yes with the current status the only way the `futureResult` fails is either during shutdown with `.shutdown` or if the task is cancelled with `cancelled`. No other error is possible at the moment. So I can remove this `whenFailure` case.

weissi: if you want to be sure it's one of those errors you could just assert that it's one of them:

    assert(([EventLoopError.cancelled, EventLoopError.shutdown] as [Error]).contains(error), "unexpected error: \(error)")

Sources/NIO/EventLoop.swift
```diff
@@ -396,7 +515,10 @@ internal final class SelectableEventLoop: EventLoop {
             promise.fail(error: error)
         },`in`)
 
-        let scheduled = Scheduled(promise: promise, cancellationTask: {
+        let scheduled = Scheduled(promise: promise, cancellationTask: { [weak self] in
+            guard let `self` = self else {
```
weissi: we don't normally use weak references because their impact on performance. Why is this one necessary?

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,83 @@ public struct Scheduled<T> {
...
+            }
+        }
+    }
+
+    private func reschedule0() {
+        assert(self.eventLoop.inEventLoop)
+        guard let task = self.task, let scheduled = self.scheduled else {
+            return
+        }
+        self.scheduled = scheduled.futureResult.eventLoop.scheduleTask(in: self.delay, task)
```
Lukasa: Ah, we actually don't need the conditional binding of `scheduled` anymore, because we have an event loop on this object we can use.

Sources/NIO/EventLoop.swift
```diff
@@ -93,7 +170,6 @@ public protocol EventLoop: EventLoopGroup {
     /// Schedule a `task` that is executed by this `SelectableEventLoop` after the given amount of time.
     func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T>
 }
-
```
Lukasa: Nit: replace this whitespace line.

FranzBusch: Can I leave this since it follows the overall formatting where no empty line is before the closing brackets?

Lukasa: This line is after the closing brace.

Sources/NIO/EventLoop.swift
weissi: I'm deleting this comment as it's done.

Sources/NIO/EventLoop.swift
```diff
@@ -50,6 +50,83 @@ public struct Scheduled<T> {
...
 }
 
+/// Returned once a task was scheduled to be repeatedly executed on the `EventLoop`.
+///
+/// A `RepeatedTask` allows the user to `cancel()` the repeated scheduling of further tasks.
+public final class RepeatedTask {
+    private let delay: TimeAmount
+    private let eventLoop: EventLoop
+    private var scheduled: Scheduled<EventLoopFuture<Void>>?
+    private var task: (() throws -> EventLoopFuture<Void>)?
```
Lukasa: I think we removed the need for this to be `throws`, didn't we? It doesn't actually manifest in the word `try` anywhere in the code for this object, but let's constrain the type further and not allow it to throw.

weissi:

Lukasa:
Looks like `testScheduleRepeatedTask` is a bit racy and sometimes fails.

weissi:
@FranzBusch ok, we need to address the test flakiness first. I'd recommend to test it with a `for _ in 0..<10000 { /* real test code here */ }` and see if you can repro what our CI hit.

FranzBusch:
I tried running the tests in a loop but couldn't reproduce the flakiness. I just increased the timeouts a bit maybe they were a bit too optimistic.

Lukasa:
You may want to go to more than a second. The timeout should only be hit in failing cases anyway.

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -91,6 +91,88 @@ public class EventLoopTest : XCTestCase {
...
+        let delay: TimeAmount = .milliseconds(10)
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
+        }
+
+        let expect = expectation(description: "Is cancelling RepatedTask")
+        var repeatedTask: RepeatedTask?
+        repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
+            DispatchQueue.main.async {
```
Lukasa: Wait, what is going on here? You can't dispatch to the main dispatch queue.

Lukasa: I'm also not really sure what the utility of doing that is. Just create a separate serial dispatch queue to use. :smile:

Lukasa:

weissi:
@Lukasa given that @FranzBusch has filled the PR description out with the commit template, we could just use the 'squash & merge' button and it'll have the same effect, right?

Lukasa:
@weissi That was my original plan, but actually the PR description contains only 1/3 of the commit template, which is why I didn't land on that plan.

Lukasa:

weissi:
@Lukasa oh sorry, didn't notice that.

Issue #497, opened 2018-07-02
Iterating over event loops of an EventLoopGroup


pushkarnk:
Can we have an iterator over the event loops of an  event loop group? 

This came up during the implementation of a perf improvement suggestion I received on `Kitura-NIO`. We'd want to initialise a `ThreadSpecificVariable` on every event loop of the event group, update it periodically (using a repeating `Scheduled` task) and read from it when needed. The `on every event loop of the event loop group` part would be better implemented if we had an iterator over the event loops. We do have `EventLoopGroup.next()` but some extra logic needs to be added around it to simulate an iterator.

Right now, I can't think of any side-effects of having an iterator over event loops. But I'll be glad to understand them, if there are any!

Lukasa:
This is a nice idea. I’d be fine with this.

normanmaurer:
@pushkarnk @Lukasa sounds like a good idea... For what it worth we also have the same in Netty (at least something similar).

pushkarnk:
Thanks @Lukasa and @normanmaurer !

I tried to implement the iterator on `EventLoopGroup` but we ran into the `Protocol EventLoopGroup can only be used as a generic constraint`error. Working around this will be a big change, I reckon.
Also, with only `EventLoopGroup.next()` which actually provides an infinite stream of `EventLoop`s, it doesn't seem possible to write a stopping condition. `EventLoop` isn't `Equatable` too. 

Implementing an iterator only over `MultiThreadedEventLoopGroup` seems quite simple, but if we have newer implementations of `EventLoopGroup` we'll have to redundantly conform them to Sequence. Here it is: https://github.com/pushkarnk/swift-nio/commit/6ebd7f0e218349b1f05dcffd17ef96c27da6a91b

I felt a little uncomfortable implementing an iterator on a type that has a provision to supply an `infinite stream` of contained objects :) Eager to know what others think!

Lukasa:
`EventLoop` is a class-bound protocol, so the end condition can be defined in terms of object identity I think.

Rather than conform `EventLoopGroup` to `Sequence`, why not instead create an object that can wrap an `EventLoopGroup` and conform it that way? That avoids the redundant conformances but still lets you avoid some of the overhead of making the protocol itself provide the iterator.

pushkarnk:
Thanks for the suggestions @Lukasa ! I've created https://github.com/apple/swift-nio/pull/500

Lukasa:
Fixed in 1.9.0.

PR #500, opened 2018-07-03
Provide an iterator over the event loops of an EventLoopGroup

pushkarnk:
Provide an iterator over the event loops of an EventLoopGroup

### Motivation:

We might want to do certain initialisations on a per-event loop basis. Example: initialising `ThreadSpecificVariable`s. It would be handy to have an iterator over the event loops in an event loop group. See #497 

### Modifications:

Added a new public API class `EventLoopGroupIterator` implementing sequence. The challenges in making `EventLoopGroup` itself a sequence have been mentioned here: https://github.com/apple/swift-nio/issues/497#issuecomment-402107940

### Result:
 Something like this is possible:
```
let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)`
for eventLoop in EventLoopGroupIterator(for: eventLoopGroup) {
    //do something
}
```

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -650,6 +650,42 @@ extension EventLoopGroup {
     }
 }
 
+/// An iterator over the `EventLoop`s of an `EventLoopGroup`.
+///
+/// Conforming to `Sequence`, the `EventLoopIterator` enables iterating over the `EventLoop`s of an `EventLoopGroup` using the `for-in` style.
+/// A common use-case of this is in performing per-`EventLoop` initializations before `BootStrap`ing.
+/// Example:
+///     ```swift
+///         let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+///         for eventLoop in EventLoopIterator(group: group) {
+///             //do something on eventLoop
+///         }
+public struct EventLoopIterator: Sequence, IteratorProtocol {
+
+    private var eventLoopsIterator: IndexingIterator<[EventLoop]>
+
+    /// Create an `EventLoopIterator` for the `EventLoopGroup` `group`.
+    ///
+    /// - parameters:
+    ///     - group: The `EventLoopGroup` whose `EventLoop`s are to be iterated.
+    public init(group: EventLoopGroup) {
+        var eventLoops: [EventLoop] = []
+        let firstEventLoop = group.next()
+        eventLoops.append(firstEventLoop)
+        while true {
+            let nextEventLoop = group.next()
+            if nextEventLoop === firstEventLoop {
+                break
+            }
+            eventLoops.append(nextEventLoop)
+        }
+        self.eventLoopsIterator = eventLoops.makeIterator()
+    }
+
+    mutating public func next() -> EventLoop? {
+        return self.eventLoopsIterator.next()
+    }
+}
 /// Called per `Thread` that is created for an EventLoop to do custom initialization of the `Thread` before the actual `EventLoop` is run on it.
 typealias ThreadInitializer = (Thread) -> Void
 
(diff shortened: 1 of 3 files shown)


FranzBusch:
Your current implementation heavily relies on the underlying implementation of `MultiThreadedEventLoopGroup` and it cannot be guaranteed that other implementations of `EventLoopGroup` actually implement `next()` in a similar way. I talked to @weissi already today about this implementation. What do you think about something along these lines?

```swift
public struct EventLoopIterator: Sequence, IteratorProtocol {
    public typealias Element = EventLoop
    private var eventLoops: IndexingIterator<[EventLoop]>
    
    internal init(_ eventLoops: [EventLoop]) {
        self.eventLoops = eventLoops.makeIterator()
    }
    
    public mutating func next() -> EventLoop? {
        return self.eventLoops.next()
    }
}

public protocol EventLoopGroup: class {
    // ....
    func makeIterator() -> EventLoopIterator
}

extension EventLoop {
    public func makeIterator() -> EventLoopIterator {
        return EventLoopIterator([self])
    }
}

extension MultiThreadedEventLoopGroup {
    public func makeIterator() -> EventLoopIterator {
        return EventLoopIterator(self.eventLoops)
    }
}
```

Lukasa:
@FranzBusch What makes you think this isn't safe? `EventLoopGroup` is a class-bound protocol, as is `EventLoop`, so it seems reasonable to me that this iterator should be safe across all possible groups.

FranzBusch:
@Lukasa on the one side I was getting at your last point, but further more I was going that with this implementation one is reliant that the next function from `EventLoopGroup` is implemented in such a way that between two ELs every other is returned. So what if there is an eventloopgroup that returns a random Loop as the next one. I hope that makes sense what I’m getting at.

pushkarnk:
@FranzBusch Thanks! Yeah, you make an interesting point. The only thing that `EventLoopGroup.next()` promises is that it returns `the next EventLoop to use`. So we may have an implementation that always returns the same event loop, or as you say, a random event loop. Makes sense to move the iterator implementation into the `EventLoopGroup` implementation. I'd like to know @Lukasa 's opinion on this understanding.

Lukasa:
It's certainly possible to implement `next()` that way.

Note that moving the implementation to `EventLoopGroup` will force this to be semver major.

FranzBusch:
I think we really shouldn't implement the Iterator with the assumptions of the current `next()` implementation from `MultiThreadedEventLoopGroup`. Sadly this will require a semver major, but I think we have a more flexible implementation for future `EventLoopGroup`s.

pushkarnk:
`Note that moving the implementation to EventLoopGroup will force this to be semver major.`

@Lukasa Does that simply mean we'd need to wait for `swift-nio 2`, whenever that's decided, to release this? I'd be okay with that.

Lukasa:
Yup, that's what it means :)

Lukasa:
Incidentally, we could still add the implementation to `MultiThreadedEventLoopGroup` and make that public, so you can still get the needed effect in your code today. We just can't make it a protocol requirement.

weissi:
@Lukasa @FranzBusch What we could do is to make the protocol requirement such that it returns an optional `EventLoopIterator` (and make it non-optional in NIO 2). For `MultiThreadedEventLoopGroup` we can still implement it non-optional and we'll provide implementations for all currently known `EventLoopGroup`s.

FranzBusch:
@weissi what benefit would adding a optional over a non-optional protocol requirement make? Doesn‘t adding a new requirement to any public protocol require a semvar major?

weissi:
@FranzBusch not if we can ship a default implementation for the new requirement. That's why I propose to make it optional (default implementation just has `func makeIterator() -> EventLoopIterator? { return nil }`) which means that all existing EventLoops conform automatically.

Lukasa:
I'm ok with that.

weissi:
@normanmaurer ^^

FranzBusch:
@weissi Thanks for clearing that up. Just thought that adding any new public method to a protocol would require a major but makes sense that it doesn't.

weissi:
@FranzBusch & @pushkarnk do you want to collaborate on a PR that implements that?

pushkarnk:
@weissi sure, I'll be glad to give it a shot!

pushkarnk:
@FranzBusch I guess, the implementation will mostly be based on your code from [this](https://github.com/apple/swift-nio/pull/500#issuecomment-402234856) comment. It'd be great if you do the PR. We can close this one then.

FranzBusch:
@pushkarnk Sure, I will open a PR sometime during this weekend.

pushkarnk:
Closing this PR. We have #509 opened.

PR #501, opened 2018-07-04
ByteBuffer: deal correctly with underreporting ContiguousCollections

weissi:
Motivation:

ContiguousCollections shouldn't give access to more than `count` bytes
via `withUnsafeBytes` but if they do we should still handle this
correctly and only copy `count` bytes.

Modifications:

- copy an upper limit of `count` bytes
- fail hard if `withUnsafeBytes` offers fewer bytes as we can't do
  anything sensible then.

Result:

even illegal `ContiguousCollection` implementations don't cause issues.


Lukasa:

PR #503, opened 2018-07-04
ByteBufferAllocator: switch to size_t where appropriate

weissi:
Motivation:

In a few situations we used `Int` in places where malloc/realloc
actually use `size_t`. On 64-bit platforms this doesn't actually matter
but on 32-bit it does!

Modifications:

- replace `Int` with `size_t` where appropriate

Result:

- should be able to use more than 2GB on 32-bit platforms


weissi:
tbh: I have no idea why this didn't cause compile errors before: `Int` and `size_t` are quite different even on 64-bit platforms

PR #509, opened 2018-07-09
Iterating over the EventLoops from EventLoopGroups

FranzBusch:
Provide a way to iterate over the event loops of an `EventLoopGroup`

### Motivation:

For some use cases, it might be necessary to access the underlying event loops. See https://github.com/apple/swift-nio/issues/497.

### Modifications:

Added a property requirement to the `EventLoopGroup` protocol.

### Result:

The event loops are now accessible from the outside.

### Alternative Implementation:
Another approach is to create a separate struct for the iterator and pass in the event loop array. This hides the implementation from the ELG.

```swift
public struct EventLoopIterator: Sequence, IteratorProtocol {
    public typealias Element = EventLoop
    private var eventLoops: IndexingIterator<[EventLoop]>
    
    internal init(_ eventLoops: [EventLoop]) {
        self.eventLoops = eventLoops.makeIterator()
    }
    
    public mutating func next() -> EventLoop? {
        return self.eventLoops.next()
    }
}

public protocol EventLoopGroup: class {
    // ....
    func makeIterator() -> EventLoopIterator
}

extension EventLoop {
    public func makeIterator() -> EventLoopIterator {
        return EventLoopIterator([self])
    }
}

extension MultiThreadedEventLoopGroup {
    public func makeIterator() -> EventLoopIterator {
        return EventLoopIterator(self.eventLoops)
    }
}
```

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -127,6 +127,32 @@ public final class RepeatedTask {
         self.reschedule()
     }
 }
+
+/// An iterator over the `EventLoop`s forming an `EventLoopGroup`.
+///
+/// Usually returned by an `EventLoopGroup`'s `makeIterator()` method.
+///
+///     let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+///     group.makeIterator()?.forEach { loop in
+///         // Do something with each loop
+///     }
+///
+public struct EventLoopIterator: Sequence, IteratorProtocol {
+    public typealias Element = EventLoop
+    private var eventLoops: IndexingIterator<[EventLoop]>
+
+    internal init(_ eventLoops: [EventLoop]) {
+        self.eventLoops = eventLoops.makeIterator()
+    }
+
+    /// Advances to the next `EventLoop` and returns it, or `nil` if no next element exists.
+    ///
+    /// - returns: The next `EventLoop` if a next element exists; otherwise, `nil`.
+    public mutating func next() -> EventLoop? {
+        return self.eventLoops.next()
+    }
+}
+
 /// An EventLoop processes IO / tasks in an endless loop for `Channel`s until it's closed.
 ///
 /// Usually multiple `Channel`s share the same `EventLoop` for processing IO / tasks and so share the same processing `Thread`.
@@ -332,6 +358,14 @@ extension EventLoop {
         let repeated = RepeatedTask(interval: delay, eventLoop: self, task: task)
         repeated.begin(in: initialDelay)
         return repeated
+    }
+
+    /// Returns an `EventLoopIterator` over this `EventLoop`.
+    ///
+    /// - note: The return value of `makeIterator` is currently optional as requiring it would be SemVer major. From NIO 2.0.0 on it will return a non-optional iterator.
+    /// - returns: `EventLoopIterator`
+    public func makeIterator() -> EventLoopIterator? {
+        return EventLoopIterator([self])
     }
 }
 
@@ -733,6 +767,12 @@ public protocol EventLoopGroup: class {
     /// The virtue of this function is to shut the event loop down. To work around that we call back on a DispatchQueue
     /// instead.
     func shutdownGracefully(queue: DispatchQueue, _ callback: @escaping (Error?) -> Void)
+
+    /// Returns an `EventLoopIterator` over the `EventLoop`s in this `EventLoopGroup`.
+    ///
+    /// - note: The return value of `makeIterator` is currently optional as requiring it would be SemVer major. From NIO 2.0.0 on it will return a non-optional iterator.
+    /// - returns: `EventLoopIterator`
+    func makeIterator() -> EventLoopIterator?
 }
 
 extension EventLoopGroup {
@@ -758,6 +798,10 @@ extension EventLoopGroup {
                 throw error
             }
         }
+    }
+
+    public func makeIterator() -> EventLoopIterator? {
+        return nil
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -622,6 +626,9 @@ public protocol EventLoopGroup: class {
     /// The virtue of this function is to shut the event loop down. To work around that we call back on a DispatchQueue
     /// instead.
     func shutdownGracefully(queue: DispatchQueue, _ callback: @escaping (Error?) -> Void)
+
+    /// `EventLoop`s of the group
```
Lukasa: Mind documenting this a bit more extensively? What it's here for, what the rules are for it, why it's optional, that sort of thing?

FranzBusch:
@Lukasa I should have said this in my initial PR comment. I know tests and docs are missing but I wanted to discuss the implementation first before finishing up these things. @weissi and I talked already but wanted to get input from others as well. @pushkarnk

Sources/NIO/EventLoop.swift
```diff
@@ -659,7 +670,11 @@ final public class MultiThreadedEventLoopGroup: EventLoopGroup {
     private static let threadSpecificEventLoop = ThreadSpecificVariable<SelectableEventLoop>()
 
     private let index = Atomic<Int>(value: 0)
-    private let eventLoops: [SelectableEventLoop]
+    private let _eventLoops: [SelectableEventLoop]
```
pushkarnk: There seems to be a convention to use names starting with `_`.  See the requirements mentioned in [this comment](https://github.com/apple/swift-nio/pull/500#discussion_r199924775). This property doesn't seem to match that requirement.

Lukasa:
Ok, I'll let @weissi to take a look.

Sources/NIO/EventLoop.swift
```diff
@@ -622,6 +626,9 @@ public protocol EventLoopGroup: class {
     /// The virtue of this function is to shut the event loop down. To work around that we call back on a DispatchQueue
     /// instead.
     func shutdownGracefully(queue: DispatchQueue, _ callback: @escaping (Error?) -> Void)
+
+    /// The internal `EventLoop`s of the group. This array can be used to iterate over every `EventLoop`.
```
weissi: we should do two things here:

- document that it's optional because not all event loop groups support it
- file a new issue (tagged as 2.0) to make this mandatory & link the issue here

also I'd like to discuss if an array is really the right API here. Mandating an array might force an allocation if some event loop group doesn't store it's event loops in an array already. Maybe we should use an iterator. Unsure. Maybe we can leave it like this because for 2.0 we'll change this API anyway. @Lukasa / @normanmaurer opinions if array or iterator is better here? I'm leaning towards iterator

Lukasa: When you say "use an iterator", what specific return type did you have in mind?

FranzBusch: If we talk about using an iterator like my alternative solution than an allocation is happening as well since it's passed by value right? Could you explain how one can avoid an allocation with an iterator?

weissi: @Lukasa new type `struct EventLoopGroupIterator`

weissi: @FranzBusch well the `EventLoopGroupIterator` could be constructed with different kinds of things. Say `[EventLoop]` if you already have one, or `IndexingIterator<Array<Element>>` or something. The idea is that we can hide what we're storing under the hood and just implement the `Iterator` protocol.

Sources/NIO/EventLoop.swift
```diff
@@ -127,6 +127,32 @@ public final class RepeatedTask {
         self.reschedule()
     }
 }
+
+/// An iterator over an array of 'EventLoop's.
```
weissi: that we're iterating over an _array_ is an implementation detail. Maybe write something like

> An iterator over the `EventLoop`s forming an `EventLoopGroup`.

or something.

Sources/NIO/EventLoop.swift
```diff
@@ -733,6 +766,11 @@ public protocol EventLoopGroup: class {
     /// The virtue of this function is to shut the event loop down. To work around that we call back on a DispatchQueue
     /// instead.
     func shutdownGracefully(queue: DispatchQueue, _ callback: @escaping (Error?) -> Void)
+
+    /// Returns an `EventLoopIterator` over the `EventLoop`s in this `EventLoopGroup`.
+    ///
+    /// - returns: `EventLoopIterator`
```
weissi: we should add something like

    /// - note: The return value of `makeIterator` is currently optional as requiring it would be SemVer major. From NIO 2.0.0 on it will return a non-optional iterator.

and please file an issue (or separate PR (that will only be merged when 2.0.0 gets released)) to do so.

Sources/NIO/EventLoop.swift
```diff
@@ -127,6 +127,32 @@ public final class RepeatedTask {
         self.reschedule()
```
weissi: please implement (and test) that for `EmbeddedEventLoop` too.

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -173,6 +173,36 @@ public class EventLoopTest : XCTestCase {
...
+        }
+
+        var counter = 0
+        iterator?.forEach { loop in
+            XCTAssertTrue(loop === eventLoop)
+            counter += 1
+        }
+
+        XCTAssertEqual(counter, 1)
+    }
```
Lukasa: Can you write two more tests for me? One that each `EventLoop` in the `MultiThreadedEventLoopGroup` itself is an iterator of one, and then one more where you write a dummy conformance to `EventLoopGroup` and prove that we get `nil`?

FranzBusch: Done, I added the first test to an existing one and created a new one for the nil check

Lukasa:

Lukasa:
@FranzBusch Looks like you need to run the linux test generation script (`ruby scripts/generate_linux_tests.rb`).

Lukasa:
Fine if we can get the test generation fixed.

weissi:

weissi:

PR #513, opened 2018-07-13
fix two tests threading issues

weissi:
CC @FranzBusch please take a look too, can't for some reason set you as a reviewer.

Motivation:

testScheduleRepeatedTask and
testScheduleRepeatedTaskCancelFromDifferentThread has threading issues
that I unfortunately missed in review.

Modifications:

fix the threading issues by using a `DispatchGroup`

Result:

fewer theading issues in the tests

### DIFF
--- a/Tests/NIOTests/EventLoopTest.swift
+++ b/Tests/NIOTests/EventLoopTest.swift
@@ -103,15 +103,27 @@ public class EventLoopTest : XCTestCase {
 
         let expect = expectation(description: "Is cancelling RepatedTask")
         let counter = Atomic<Int>(value: 0)
-        var repeatedTask: RepeatedTask?
-        repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
+        let lock = Lock()
+        var repeatedTask: RepeatedTask? = nil // protected by `lock`
+        let task = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
             if counter.load() == 0 {
                 XCTAssertTrue(DispatchTime.now().uptimeNanoseconds - nanos >= initialDelay.nanoseconds)
             } else if counter.load() == count {
                 expect.fulfill()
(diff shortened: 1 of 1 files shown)


Tests/NIOTests/EventLoopTest.swift
```diff
@@ -130,13 +134,17 @@ public class EventLoopTest : XCTestCase {
         }
 
         let expect = expectation(description: "Is cancelling RepatedTask")
-        var repeatedTask: RepeatedTask?
-        repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
+        let group = DispatchGroup()
+        group.enter()
+        let repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
             DispatchQueue(label: "nio.repeatedtask.test").async {
```
Lukasa: This test can avoid using this `DispatchQueue`, because the `.notify` will use a different thread anyway.

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -103,16 +103,20 @@ public class EventLoopTest : XCTestCase {
...
             if counter.load() == 0 {
                 XCTAssertTrue(DispatchTime.now().uptimeNanoseconds - nanos >= initialDelay.nanoseconds)
             } else if counter.load() == count {
                 expect.fulfill()
-                repeatedTask?.cancel()
+                group.leave()
             }
             counter.store(counter.load() + 1)
         }
+        group.notify(queue: DispatchQueue.global()) {
```
Lukasa: This patch isn't acceptable, because it reductively makes this test the same as the one below. Instead, you should use `DispatchQueue.sync` to synchronise access to the repeated task variable. It's vital that this attempt to cancel from on the event loop.

weissi: haha, thanks, didn't actually read the test name, will fix

weissi:

weissi:
@Lukasa adddressed

Issue #514, opened 2018-07-13
pass the `RepeatedTask` to its closure


weissi:
when a `RepeatedTask` invokes its closure the user might want to cancel the `RepeatedTask`. Doing that is pretty error-prone without being handed the `RepeatedTask` itself. We should fix this before 1.9.0 gets released.

CC @FranzBusch

weissi:
cc @Lukasa

FranzBusch:
I will implement this weekend.

Lukasa:
Resolved by #515.

PR #515, opened 2018-07-14
Pass RepeatedTask to its closure #514

FranzBusch:
Pass the reference to `RepeatedTask` to its closure to allow a thread-safe cancelling. https://github.com/apple/swift-nio/issues/514

### Motivation:

With the current implementation, one has to create a reference and use a lock to thread-safely cancel the `RepeatedTask`. This makes the use more difficult and not so straight-forward.

### Modifications:

Added a weak `RepeatedTask?` parameter to the `scheduleRepeatedTask` closure.

### Result:

It is now possible to cancel the `RepeatedTask` from within the task closure.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -57,9 +57,9 @@ public final class RepeatedTask {
     private let delay: TimeAmount
     private let eventLoop: EventLoop
     private var scheduled: Scheduled<EventLoopFuture<Void>>?
-    private var task: (() -> EventLoopFuture<Void>)?
+    private var task: ((RepeatedTask) -> EventLoopFuture<Void>)?
 
-    internal init(interval: TimeAmount, eventLoop: EventLoop, task: @escaping () -> EventLoopFuture<Void>) {
+    internal init(interval: TimeAmount, eventLoop: EventLoop, task: @escaping (RepeatedTask) -> EventLoopFuture<Void>) {
         self.delay = interval
         self.eventLoop = eventLoop
         self.task = task
@@ -80,7 +80,9 @@ public final class RepeatedTask {
         guard let task = self.task else {
             return
         }
-        self.scheduled = eventLoop.scheduleTask(in: delay, task)
+        self.scheduled = eventLoop.scheduleTask(in: delay) {
+            task(self)
+        }
         self.reschedule()
     }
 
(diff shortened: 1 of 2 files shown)


FranzBusch:
@Lukasa Can you check if this usage of the weak reference is fine? I used since I couldn't figure out any other way not to create a retain cycle and breaking the test `testScheduleRepeatedTaskToNotRetainEventLoop`.

Sources/NIO/EventLoop.swift
```diff
@@ -59,10 +59,12 @@ public final class RepeatedTask {
     private var scheduled: Scheduled<EventLoopFuture<Void>>?
     private var task: (() -> EventLoopFuture<Void>)?
 
-    internal init(interval: TimeAmount, eventLoop: EventLoop, task: @escaping () -> EventLoopFuture<Void>) {
+    internal init(interval: TimeAmount, eventLoop: EventLoop, task: @escaping (RepeatedTask?) -> EventLoopFuture<Void>) {
         self.delay = interval
         self.eventLoop = eventLoop
-        self.task = task
+        self.task = { [weak self] in
```
Lukasa: You can avoid the weak reference by recreating the task each time. That is, you call `scheduleTask(in: delay) { task(self) }`.

That does provide a reference cycle, but it's one that already exists, as `reschedule` already gives the `Future` a reference to `self`. Then there's no need to receive the `RepeatedTask` as an optional.

Sources/NIO/EventLoop.swift
```diff
@@ -116,14 +120,20 @@ public final class RepeatedTask {
                 self.reschedule0()
             }
         }
+
+        scheduled.futureResult.whenFailure { _ in
```
Lukasa: Can you ignore the result but specify its type? That helps us catch unexpected API changes. E.g., write `{ (_: Error) in`

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -144,7 +131,7 @@ public class EventLoopTest : XCTestCase {
         let expect = expectation(description: "Is cancelling RepatedTask")
         let group = DispatchGroup()
         group.enter()
-        let repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { () -> Void in
+        let repeatedTask = eventLoopGroup.next().scheduleRepeatedTask(initialDelay: initialDelay, delay: delay) { _ -> Void in
```
Lukasa: Can you ignore this value but specify its type, please? Same with the two changes below.

Lukasa:

Issue #519, opened 2018-07-16
Failure in WebSocket protocol upgrade


nethraravindran:
`Upgrade` header for a web socket upgrade request can either have `Websocket` or `websocket` as its value. 

Upgrade request with ```Websocket``` protocol fails because  the supported protocol is ```websocket```.

However the protocol names should be case-insensitive for the `HTTPServerUpgradeHandler` to handle the upgrade request

weissi:
fixed in #520

PR #520, opened 2018-07-16
Use case-insensitive protocol names for indexing the upgraders dictionary

nethraravindran:
**Motivation** 

`handleUpgrade(ctx: ChannelHandlerContext, request: HTTPRequestHead, requestedProtocols: [String])` in `HTTPServerUpgradeHandler` upgrades a request only if the `Upgrade` header of the request is `websocket` protocol. For `Websocket` protocol, it fails. For indexing the upgraders, case-insensitive protocol names should be used.

**Modifications:**

```
 for proto in requestedProtocols {
            guard let upgrader = upgraders[proto.lowercased()] else {
                continue
            }
```
Here, all the protocols in the `requestedProtocols` are lowercased before indexing the upgraders dictionary.

### DIFF
--- a/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
@@ -94,7 +94,7 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
     public init(upgraders: [HTTPProtocolUpgrader], httpEncoder: HTTPResponseEncoder?, httpDecoder: HTTPRequestDecoder?, upgradeCompletionHandler: @escaping (ChannelHandlerContext) -> Void) {
         var upgraderMap = [String: HTTPProtocolUpgrader]()
         for upgrader in upgraders {
-            upgraderMap[upgrader.supportedProtocol] = upgrader
+            upgraderMap[upgrader.supportedProtocol.lowercased()] = upgrader
         }
         self.upgraders = upgraderMap
         self.upgradeCompletionHandler = upgradeCompletionHandler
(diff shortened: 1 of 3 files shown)


nethraravindran:
Issue: https://github.com/apple/swift-nio/issues/519

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -187,7 +187,7 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
         let allHeaderNames = Set(request.headers.map { $0.name.lowercased() })
 
         for proto in requestedProtocols {
-            guard let upgrader = upgraders[proto] else {
+            guard let upgrader = upgraders[proto.lowercased()] else {
```
Lukasa: This should be accompanied by an equivalent lowercasing when we create the `upgraders` dictionary.

nethraravindran: @Lukasa Thanks! Will add the tests :)

Lukasa:

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
Lukasa: @ianpartridge Not really.

The spec is silent on whether or not case is meaningful in the upgrade header values, but I'd want to take my cue from the Connection header values, which end up being treated case insensitively for maximum compatibility. I'd suspect the same is true here, and I sincerely doubt the IETF would ever produce a specification which allowed two protocols to register an upgrade token distinguished only by case.

PR #521, opened 2018-07-18
Optional promotion bug

John-Connolly:
Incorrect comparison due to optional promotion

### Motivation:
Fix comparison

### Modifications: 
Changed comparison to use _none

### Result:
Correct comparison


Lukasa:

PR #525, opened 2018-07-23
Made some HTTPHeaders API public, to use in NIOHPACK

AlanQuatermain:
There's a public static creator which takes a ByteBuffer and array of `HTTPHeader` now, and the `buffer` and `headers` properties (and thus the `HTTPHeader` and `HTTPHeaderIndex` types) are public to support that.

### Motivation:

HTTP/2 adds an additional item to a header key/value pair: its indexability. By default, headers can be inserted into the HPACK header table index; sometimes this will be denied (common case is 'content-length' header, which is rarely going to be the same, and will just cause the table to purge more often than is really necessary). Additionally, a header may be marked as not only non-indexable but also non-rewritable by proxies. `HTTPHeaders` provides nowhere to store this, so the `HPACKHeaders` implementation referenced in https://github.com/apple/swift-nio-http2/pull/10 was created to add in that capability.

Now, given that we really want the HTTP version to be an implementation detail, we want to keep the HPACK details hidden, and would thus be using `HTTPHeaders` in client APIs. NIOHTTP2 already has a HTTP1-to-HTTP2 channel handler that translates between the two. Thus it behooves us to have the means to copy the actual key/value pairs between the two types without making a round-trip from UTF-8 bytes to UTF-16 `String`s. These changes allow NIOHPACK or NIOHTTP2 to implement that round-trip internally.

### Modifications:

- `HTTPHeader` and `HTTPHeaderIndex` types are now public.
- `HTTPHeaders.buffer` and `HTTPHeaders.headers` properties are now public.
- A new static method, `HTTPHeaders.createHeaderBlock(buffer:headers:)` was created to serve as a public means to create a `HTTPHeaders` from a `ByteBuffer` from outside NIOHTTP1. @Lukasa suggested this should be a static method rather than an initializer.

### Result:

Nothing in NIOHTTP1 changes in operation. All public types have documentation comments noting that they are only public for very specific reasons, and are not intended for general use. Once this is committed, NIOHPACK & NIOHTTP2 will be able to implement fast round-trips between `HTTPHeaders` and `HPACKHeaders`.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -26,6 +26,7 @@ needs to be listed here.
 - Helge Heß <helge@alwaysrightinstitute.com>
 - Ian Partridge <i.partridge@uk.ibm.com>
 - Jason Toffaletti <toffaletti@gmail.com>
+- Jim Dovey <jimdovey@mac.com>
 - Johannes Weiß <johannesweiss@apple.com>
 - John Connolly <connoljo2@gmail.com>
 - Karim ElNaggar <karimfarid.naggar@gmail.com>
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -368,13 +368,19 @@ public struct HTTPResponseHead: Equatable {
 }
 
 /// The Index for a header name or value that points into the underlying `ByteBuffer`.
-struct HTTPHeaderIndex {
+///
+/// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
+///         NIOHTTP2 and NIOHPACK. It is not intended for general use.
+public struct HTTPHeaderIndex {
     let start: Int
     let length: Int
 }
 
 /// Struct which holds name, value pairs.
-struct HTTPHeader {
+///
+/// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
+///         NIOHTTP2 and NIOHPACK. It is not intended for general use.
+public struct HTTPHeader {
     let name: HTTPHeaderIndex
     let value: HTTPHeaderIndex
 }
@@ -493,6 +499,41 @@ public struct HTTPHeaders: CustomStringConvertible {
             headersArray.append((self.string(idx: h.name), self.string(idx: h.value)))
         }
         return headersArray.description
+    }
+    
+    /// Creates a header block from a pre-filled contiguous string buffer containing a
+    /// UTF-8 encoded HTTP header block, along with a list of the locations of each
+    /// name/value pair within the block.
+    ///
+    /// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
+    ///         NIOHTTP2 and NIOHPACK. It is not intended for general use.
+    ///
+    /// - Parameters:
+    ///   - buffer: A buffer containing UTF-8 encoded HTTP headers.
+    ///   - headers: The locations within `buffer` of the name and value of each header.
+    /// - Returns: A new `HTTPHeaders` using the provided buffer as storage.
+    public static func createHeaderBlock(buffer: ByteBuffer, headers: [HTTPHeader]) -> HTTPHeaders {
+        return HTTPHeaders(buffer: buffer, headers: headers, keepAliveState: KeepAliveState.unknown)
+    }
+    
+    
+    /// Provides access to raw UTF-8 storage of the headers in this header block, along with
+    /// a list of the header strings' indices.
+    ///
+    /// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
+    /// NIOHTTP2 and NIOHPACK. It is not intended for general use.
+    ///
+    /// - parameters:
+    ///   - block:      A block that will be provided UTF-8 header block information.
+    ///   - buf:        A raw `ByteBuffer` containing potentially-contiguous sequences of UTF-8 encoded
+    ///                 characters.
+    ///   - locations:  An array of `HTTPHeader`s, each of which contains information on the location in
+    ///                 the buffer of both a header's name and value.
+    ///   - contiguous: A `Bool` indicating whether the headers are stored contiguously, with no padding
+    ///                 or orphaned data within the block. If this is `true`, then the buffer represents
+    ///                 a HTTP/1 header block appropriately encoded for the wire.
+    public func withUnsafeBufferAndIndices<R>(_ block: (_ buf: ByteBuffer, _ locations: [HTTPHeader], _ contiguous: Bool) throws -> R) rethrows -> R {
+        return try block(self.buffer, self.headers, self.continuous)
     }
 
     /// Constructor used by our decoder to construct headers without the need of converting bytes to string.
--- a/.mailmap
+++ b/.mailmap
@@ -3,6 +3,7 @@ Tomer Doron <tomerd@apple.com>
 Max Moiseev <moiseev@apple.com>
 Johannes Weiß <johannesweiss@apple.com>
 Adam Nemecek <adamnemecek@gmail.com>
+Jim Dovey <jimdovey@mac.com> <jdovey@linkedin.com>
 <bas@basbroek.nl> <BasThomas@users.noreply.github.com>
 <daniel_dunbar@apple.com> <daniel@zuster.org>
 <johannesweiss@apple.com> <github@tux4u.de>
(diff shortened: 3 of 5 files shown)


AlanQuatermain:
I like the idea of a `with*` function. I wasn't entirely certain how to couch things, so I went with the smallest actual API change.

weissi:
thanks @AlanQuatermain! I'm not too fussed about exposing the `HTTPHeaderIndex` and `HTTPHeader` types but exposing `buffer` and `headers` directly (as properties) I find a little more troublesome. The reason is: If they're properties they should really be constant time and for those two that means our implementation then needs to always back them by a `ByteBuffer`. For a `with*` function that's not the case so I think I'd prefer that too.

AlanQuatermain:
Cool, I'll update this to use the `with*` approach.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -388,6 +398,12 @@ public struct HTTPHeaders: CustomStringConvertible {
         }
         return headersArray.description
     }
+    
+    /// This is public to aid in the creation of supplemental HTTP libraries, e.g.
+    /// NIOHTTP2 and NIOHPACK. It is not intended for general use.
```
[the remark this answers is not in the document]
Lukasa: I'd still like an explanation in this function. :smile:

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -388,6 +394,31 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    /// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
+    /// NIOHTTP2 and NIOHPACK. It is not intended for general use.
+    ///
+    /// - parameters:
+    ///   - block:      A block that will be provided UTF-8 header block information.
+    ///   - buf:        A raw `ByteBuffer` containing potentially-contiguous sequences of UTF-8 encoded
+    ///                 characters.
+    ///   - locations:  An array of `HTTPHeader`s, each of which contains information on the location in
+    ///                 the buffer of both a header's name and value.
+    ///   - contiguous: A `Bool` indicating whether the headers are stored contiguously, with no gaps between
```
Lukasa: Does this render properly? My assumption is that it does not, that these parameter meanings need to be documented in the text for `block`.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -388,6 +394,31 @@ public struct HTTPHeaders: CustomStringConvertible {
...
+    ///                 adjacent name/value pairs.
```
Lukasa: Do you mind elaborating on this last parameter? Specifically, there *are* gaps, but those gaps correspond to a HTTP/1.1-formatted header block (that is, the gaps are either `: ` or `\r\n` in the buffer).

Sources/NIOHTTP1/HTTPTypes.swift
AlanQuatermain: Yep, it does. I had to do some digging to find out the proper layout, but it transpires you just have to give the block parameters names and include then with the method's parameters. They render correctly in a separate table in the `block` parameter's documentation.

Sources/NIOHTTP1/HTTPTypes.swift
AlanQuatermain: Argh, I'd forgotten in contained those. I was thinking of the `continuous` property that is marked `false` when something is removed from the buffer, i.e. when the buffer also contains strings that aren't intended to be included in the header list. In other words, whether the buffer can be directly copied.

In retrospect, it doesn't really match my intended use case (i.e. whether `HPACKHeaders` can just grab the buffer directly without caring about copying extra data—since the HPACK structure isn't supporting the same add/remove functionality and would like a smaller storage load). However, it still seems useful since it can tell receivers if the buffer is already a valid encoded HTTP/1 header block, ready for the wire as-is.

Sources/NIOHTTP1/HTTPTypes.swift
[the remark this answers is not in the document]
Lukasa: I think this comment got lost in the file.

AlanQuatermain:
Le oops. Duly excised 😊

Lukasa:

Lukasa:
Looks like we're missing the Linux tests. @AlanQuatermain, you may find it useful to place the following file at `.git/hooks/pre-commit` in the repository:

```bash
#!/bin/sh

FIRST_OUT="$(git status --porcelain)"
ruby scripts/generate_linux_tests.rb > /dev/null
NEW_OUT="$(git status --porcelain)"

if [[ "$FIRST_OUT" != "$NEW_OUT" ]]; then
    printf "\033[0;31mMissing linux test changes!\033[0m\n"
    printf "The appropriate changes have been left in the repo.\n"
    printf "Please fix up and then merge.\n"
    exit 1
fi
```

AlanQuatermain:

CONTRIBUTORS.txt
```diff
@@ -14,18 +14,24 @@ needs to be listed here.
...
+- Esteban Torres <me@estebantorr.es>
+- Eugen <eugenf78@gmail.com>
 - Frank Kair <frankkair@gmail.com>
+- Franz Busch <privat@franz-busch.de>
 - Gopal Sharma <gopal@the-sharmas.org>
 - Helge Heß <helge@alwaysrightinstitute.com>
 - Ian Partridge <i.partridge@uk.ibm.com>
 - Jason Toffaletti <toffaletti@gmail.com>
+- Jim Dovey <jimdovey@mac.com>
+- Johannes Weiss <johannesweiss@apple.com>
```
Lukasa: Let's remove the contributors script, as it has inadvertently added Johannes twice. We should re-run it in a separate PR, and then add it to our sanity check build system to start rejecting builds where the user is not in the contributor mesh.

Lukasa: Ok, see #550 for this. You should leave your own name's entry here in the right place, as that will prevent us hitting a problem if we end up merging #550 before this patch.

Lukasa:
Having @AlanQuatermain should mean we don't have to prompt the bot to build anymore, it should auto-build all of @AlanQuatermain's PRs on this repo.

Lukasa:
Oh wait, he isn't on the whitelist here.

Lukasa:

CONTRIBUTORS.txt
AlanQuatermain: Hehe, I was going to suggest the `.mailmap` thing—I saw the same in the http2 repo, and that the `.mailmap` file there had two entries for @weissi that were the wrong way round, it seems. I copied his approach first, but then that didn't work so I copied yours 😉

AlanQuatermain: I've simply reverted the Contributors file changes; if the script will be run & inlined with the build or merge process then that should be handled by something other than my commit, I think. I expect Github provides something like a post-pull-request-approval hook that could be used to automate the process?

AlanQuatermain: Oh, wait, I mis-read. So I should update the contributors file but manually delete @weissi's ASCII clone—no problem.

Lukasa: Sorry, no, you were right the first time: let's remove the entire CONTRIBUTORS file change except your entry.

weissi: sorry guys 😬. I recently had too much trouble with 'Weiß' so switched to 'Weiss' everywhere 🙃.

Lukasa:
@AlanQuatermain Looks like we have some merge conflicts: can you fix those up and re-push? Then we should be able to land this.

AlanQuatermain:
Face → palm. So much annoying commit replication there. Sigh. If only there were a way to rebase these things cleanly, without generating weird commit histories. I would squash all this if I though that would actually not cause even more problems when I try to push it to my own repo; the pull-request model kinda breaks when you need to have two repositories (local and Github) that need to rebase when there's a merge conflict.

AlanQuatermain:
Boom. Damned thing. Made a new branch based on apple/master, merge-squashed my changes onto that, hard-reset master back to 57c6724, then force-pushed.

Single. Commit.

**\*mic drop\***

AlanQuatermain:
For the record, there was a chain of about 50 commits, many of which were identical due to multiple merges/rebases, before the merge-squash created one very small commit containing only the *actual* changes for this PR. Sigh.

PR #526, opened 2018-07-24
Avoid hot-looping when autoRead is disabled.

Lukasa:
Motivation:

On macOS it is possible to hot-loop on kevent when a channel has autoRead
disabled but pending data in the receive buffer. This is because
EVFILT_EXCEPT will fire when there is pending data in the receive buffer.

For now we work around this by setting NOTE_LOWAT to Int.max, which will
ensure that we don't actually get EVFILT_EXCEPT fired for pending data.

Modifications:

- When registering EVFILT_EXCEPT, set NOTE_LOWAT to Int.max.
- Fix a test that definitely didn't test what it wanted to.

Result:

No hot-loops with pending data.


Lukasa:

PR #529, opened 2018-07-28
Support user defined writability in Channels

kdubb:
Mimics the user defined writability feature present in Netty’s `ChannelOutboundBuffer`.  

### Motivation:

Currently porting the traffic shaping features from Netty to Swift NIO.  Proper implementation of a  `ChannelTrafficShapingHandler` requires the ability to disable writability based on external factors.

### Modifications:

It adds the feature to `PendingWritesManager` for both stream & datagram sockets. Instead of exposing a reference to a `PendingWritesManager` interface/protocol in `Channel.unsafe` (as Netty does with `ChannelOutboundBuffer`) the `Channel`'s options feature (`getOption`/`setOption`) is used to allow setting the user defined writability of the `Channel` (which is forwarded to the `PendingWritesManager` internally).

### Result:

External handlers can control the writability of a channel without affecting the status calculated internally by the `Channel`s or `PendingWritesManager`s

### DIFF
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -186,6 +186,36 @@ public enum AllowRemoteHalfClosureOption: ChannelOption {
     case const(())
 }
 
+/// `UserDefinedWritabilityOption` allows users to control the `isWritable` status of the channel aside from its
+/// normal calculation.  This is useful for handlers that need to have a say in when the channel is writable. For
+/// example, the traffic shaping handler needs to be able to disable writability at certain times and this option
+/// provides that capability.
+///
+/// There are 63 available flags and users must negotatiate an index on their own. The option carries the index
+/// of a specific flag to set and a value for the flag.
+public enum UserDefinedWritabilityOption: ChannelOption {
+    public typealias AssociatedValueType = Int
+    public typealias OptionType = Bool
+
+    case const(AssociatedValueType)
+
+    /// Create a new `UserDefinedWritabilityOption`.
+    ///
+    /// - parameters:
+    ///       - index: Index of the writability flag to set.
+    public init(index: Int) {
+        assert(index >= 1 && index < 64, "User defined writability indices must be one of 1..63")
+        self = .const(index)
+    }
+
+    public var value: Int {
+        switch self {
+        case .const(let index):
+            return index
+        }
+    }
+}
+
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
 public struct ChannelOptions {
     /// - seealso: `SocketOption`.
@@ -217,4 +247,7 @@ public struct ChannelOptions {
 
     /// - seealso: `AllowRemoteHalfClosureOption`.
     public static let allowRemoteHalfClosure = AllowRemoteHalfClosureOption.const(())
+
+    /// - seealso: `UserDefinedWritabilityOption`.
+    public static let userDefinedWritability = { (index: Int) -> UserDefinedWritabilityOption in .const(index) }
 }
--- a/Sources/NIO/PendingWritesManager.swift
+++ b/Sources/NIO/PendingWritesManager.swift
@@ -273,7 +273,7 @@ final class PendingStreamWritesManager: PendingWritesManager {
     private var storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>
 
     internal var waterMark: WriteBufferWaterMark = WriteBufferWaterMark(low: 32 * 1024, high: 64 * 1024)
-    internal let channelWritabilityFlag: Atomic<Bool> = Atomic(value: true)
+    internal let unwritableFlags: Atomic<UInt64> = Atomic(value: 0)
 
     internal var writeSpinCount: UInt = 16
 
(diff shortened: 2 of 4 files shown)


Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+        var oldValue: UInt64
+        var newValue: UInt64
+        repeat {
+            oldValue = unwritableFlags.load()
+            newValue = oldValue | mask
+        }
+        while !unwritableFlags.compareAndExchange(expected: oldValue, desired: newValue)
+
+        return oldValue == 0 && newValue != 0
+    }
```
Lukasa: So we've duplicated the basic body of this method 4 times, differing only on the bitwise ops and the mask.

We can and should begin by dropping this to two implementations that both accept a writability mask, one for setting and one for unsetting the mask. Each of these 4 functions can call that base implementation.

I don't think the advantage of going to one implementation helps us, though, as we'd need to pass closures in, which will be a bit slower than we really want at the moment.

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+
+    internal func clearUserDefinedWritability(index: Int) -> Bool {
+        let mask = writabilityMask(index)
+        var oldValue: UInt64
+        var newValue: UInt64
+        repeat {
+            oldValue = unwritableFlags.load()
+            newValue = oldValue | mask
+        }
+        while !unwritableFlags.compareAndExchange(expected: oldValue, desired: newValue)
```
Lukasa: This loop isn't actually necessary, and neither is the CaS. (To be fair, the CaS wasn't necessary in the original implementation either, so you were misled by what was there.)

While the writability flag is an atomic, the reason it is so is because the writability flag can be *read* from outside the event loop. It is never set from outside the event loop. Because it's only set from on the event loop, all write operations on this atomic are synchronized. That means it's just fine to replace the CaS with load/store, which has the advantage of performing vastly better, and also eliding these loops and simplifying the code.

Sources/NIO/ChannelOption.swift
```diff
@@ -186,6 +186,36 @@ public enum AllowRemoteHalfClosureOption: ChannelOption {
     case const(())
 }
 
+/// `UserDefinedWritabilityOption` allows users to control the `isWritable` status of the channel aside from its
+/// normal calculation.  This is useful for handlers that need to have a say in when the channel is writable. For
+/// example, the traffic shaping handler needs to be able to disable writability at certain times and this option
+/// provides that capability.
+///
+/// There are 63 available flags and users must negotatiate an index on their own. The option carries the index
```
Lukasa: I don't really love having users negotiate an index on their own. I'd vastly prefer users being able to ask for an index. I appreciate that that will require a bigger change, and we don't necessarily need to do it right away, but I don't love that at all.

weissi: I'm in favour of asking rather than the user negotiating as well, think we should do this straight away.

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+            oldValue = unwritableFlags.load()
+            newValue = oldValue & UInt64(bitPattern: -2);
+        }
+        while !unwritableFlags.compareAndExchange(expected: oldValue, desired: newValue)
+        return true
+    }
+
+    internal func setUnwritable() -> Bool {
+        // Check if flag already cleared
+        if unwritableFlags.load() & 1 == 1 {
```
weissi: needs a `self.`

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+            return false
+        }
+        var oldValue: UInt64
+        var newValue: UInt64
+        repeat {
+            oldValue = unwritableFlags.load()
```
weissi: `self.`

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+            newValue = oldValue | 1
+        }
+        while !unwritableFlags.compareAndExchange(expected: oldValue, desired: newValue)
```
weissi: the `while` should be on the same line as the closing `}`

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+    internal func getUserDefinedWritability(index: Int) -> Bool {
+        let mask = writabilityMask(index)
+        return (unwritableFlags.load() & mask) == mask
+    }
+
+    internal func setUserDefinedWritability(index: Int, writable: Bool) -> Bool {
+        if (writable) {
+            return setUserDefinedWritability(index: index)
+        }
+        else {
```
weissi: `else` one line up

Sources/NIO/PendingWritesManager.swift
```diff
@@ -447,13 +447,92 @@ internal protocol PendingWritesManager {
...
+
+    internal func setUserDefinedWritability(index: Int) -> Bool {
+        let mask = ~writabilityMask(index)
+        var oldValue: UInt64
+        var newValue: UInt64
+        repeat {
+            oldValue = unwritableFlags.load()
+            newValue = oldValue & mask
+        }
+        while !unwritableFlags.compareAndExchange(expected: oldValue, desired: newValue)
```
weissi: `while` one line up

weissi:
ping @kdubb

kdubb:
@weissi Got sidetracked with other things. I'll get these completed and pushed in the coming week.

weissi:
@kdubb thank you & no worries, whenever you find time!

weissi:
This is quite out of date, closing for now, please feel free to reopen at any time

PR #531, opened 2018-07-29
Added write function to NonBlockingFileIO

### LINKED ISSUE #406: NonBlockingFileIO should be extended with write methods
Could we get better file streaming support please?

CC: @Lukasa

maxdesiatov:
### Motivation:

`NonBlockingFileIO` currently provides multiple read functions, but no functions to write to a `FileHandle` in an asynchronous non-blocking way.

### Modifications:

`NonBlockingFileIO` struct now has simple basic `write` function that supports writing from a `ByteBuffer` to a `FileHandle`. Added `testWriting()` to `NonBlockingFileIOTest`.

### Result:

`NonBlockingFileIO` now has a basic building block for asynchronous non-blocking writes.

Resolves #406.

### DIFF
--- a/Sources/NIO/NonBlockingFileIO.swift
+++ b/Sources/NIO/NonBlockingFileIO.swift
@@ -214,6 +214,46 @@ public struct NonBlockingFileIO {
         }
     }
 
+    /// Write `buffer` to `fileHandle` in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///   - fileHandle: The `FileHandle` to write to.
+    ///   - buffer: The `ByteBuffer` to write.
+    ///   - eventLoop: The `EventLoop` to create the returned `EventLoopFuture` from.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the write was successful or fails on error.
+    public func write(fileHandle: FileHandle,
+                      buffer: ByteBuffer,
+                      eventLoop: EventLoop) -> EventLoopFuture<()> {
+        var byteCount = buffer.readableBytes
+
+        guard byteCount > 0 else {
+            return eventLoop.newSucceededFuture(result: ())
+        }
+
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            var buf = buffer
+            while byteCount > 0 {
+                let n = try buf.readWithUnsafeReadableBytes { ptr in
+                    precondition(ptr.count == byteCount)
+                    let res = try fileHandle.withUnsafeFileDescriptor { descriptor in
+                        try Posix.write(descriptor: descriptor,
+                                        pointer: ptr.baseAddress!,
+                                        size: byteCount)
+                    }
+                    switch res {
+                    case .processed(let n):
+                        assert(n >= 0, "write claims to have written a negative number of bytes \(n)")
+                        return n
+                    case .wouldBlock:
+                        throw Error.descriptorSetToNonBlocking
+                    }
+                }
+
+                byteCount -= n
+            }
+        }
+    }
+
     /// Open the file at `path` on a private thread pool which is separate from any `EventLoop` thread.
     ///
     /// This function will return (a future) of the `FileHandle` associated with the file opened and a `FileRegion`
--- a/Tests/NIOTests/NonBlockingFileIOTest+XCTest.swift
+++ b/Tests/NIOTests/NonBlockingFileIOTest+XCTest.swift
@@ -45,6 +45,7 @@ extension NonBlockingFileIOTest {
                 ("testFileRegionReadFromPipeFails", testFileRegionReadFromPipeFails),
                 ("testReadFromNonBlockingPipeFails", testReadFromNonBlockingPipeFails),
                 ("testSeekPointerIsSetToFront", testSeekPointerIsSetToFront),
+                ("testWriting", testWriting),
                 ("testFileOpenWorks", testFileOpenWorks),
                 ("testFileOpenWorksWithEmptyFile", testFileOpenWorksWithEmptyFile),
                 ("testFileOpenFails", testFileOpenFails),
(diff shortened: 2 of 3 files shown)


maxdesiatov:
This PR resolves #406. It adds only one basic function, additional wrappers providing `FileRegion` API can be provided in separate PRs if needed.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,54 @@ public struct NonBlockingFileIO {
...
+                    switch res {
+                    case .processed(let n):
+                        assert(n >= 0, "write claims to have read a negative number of bytes \(n)")
+                        return n
+                    case .wouldBlock:
+                        throw Error.descriptorSetToNonBlocking
+                    }
+                }
+
+                if n == 0 {
```
Lukasa: This end condition doesn't seem necessary: we're looping to handle this case. I think we can avoid the conditional branch here and unconditionally add `n`. We can `precondition` if you want the check, but we should never really enter this branch.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,54 @@ public struct NonBlockingFileIO {
...
+            while bytesWritten < byteCount {
+                let n = try buf.readWithUnsafeMutableReadableBytes { ptr in
+                    let res = try fileHandle.withUnsafeFileDescriptor { descriptor in
+                        try Posix.write(descriptor: descriptor,
+                                        pointer: ptr.baseAddress!,
+                                        size: byteCount - bytesWritten)
+                    }
+                    switch res {
+                    case .processed(let n):
+                        assert(n >= 0, "write claims to have read a negative number of bytes \(n)")
```
Lukasa: Nit: "write claims to have written a negative number".

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,54 @@ public struct NonBlockingFileIO {
...
+                      buffer: ByteBuffer,
+                      byteCount: Int,
+                      eventLoop: EventLoop) -> EventLoopFuture<()> {
+        guard byteCount > 0 else {
+            return eventLoop.newSucceededFuture(result: ())
+        }
+
+        var buf = buffer
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            var bytesWritten = 0
```
Lukasa: We need to be a bit defensive here, as the user may as us to write 100 bytes from a byte buffer that only has 50 readable bytes, in which case we'll enter a nasty loop forever. We should create a local `byteCount = min(buffer.readableBytes, byteCount)`.

Lukasa: In fact once we do that we don't need a `bytesWritten` local variable, because we can just subtract from `byteCount`.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,54 @@ public struct NonBlockingFileIO {
...
+            while bytesWritten < byteCount {
+                let n = try buf.readWithUnsafeMutableReadableBytes { ptr in
+                    let res = try fileHandle.withUnsafeFileDescriptor { descriptor in
+                        try Posix.write(descriptor: descriptor,
+                                        pointer: ptr.baseAddress!,
+                                        size: byteCount - bytesWritten)
```
Lukasa: We need some caution here too, as the `ptr` may not have `bytesCount - bytesWritten` bytes in it. If we add the check above to use `min(buffer.readableBytes, byteCount)` then we'll be safe, but we should probably still put a `precondition` that `ptr.count == byteCount` in this body just to be sure.

Lukasa:

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,51 @@ public struct NonBlockingFileIO {
         }
     }
 
+    /// Write `byteCount` bytes from `buffer` to `fileHandle in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// The returned `ByteBuffer` will not have less than `byteCount` bytes unless we hit end-of-file in which
+    /// case the `ByteBuffer` will contain the bytes available to read.
```
Lukasa: Please remove this paragraph.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,48 @@ public struct NonBlockingFileIO {
         }
     }
 
+    /// Write `byteCount` bytes from `buffer` to `fileHandle in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
```
Lukasa: Ah, sorry, I missed the unbalanced backtick around `fileHandle`.

Lukasa:

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -214,6 +214,48 @@ public struct NonBlockingFileIO {
...
+    ///
+    /// - parameters:
+    ///   - fileHandle: The `FileHandle` to write to.
+    ///   - buffer: The `ByteBuffer` to write.
+    ///   - byteCount: The number of bytes to write from `buffer` to `fileHandle`.
+    ///   - eventLoop: The `EventLoop` to create the returned `EventLoopFuture` from.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the write was successful or fails on error.
+    public func write(fileHandle: FileHandle,
+                      buffer: ByteBuffer,
+                      byteCount: Int,
```
weissi: sorry, I'm pretty late to the party: Do we need `byteCount`? If the user wants to write less, they can just slice the ByteBuffer, right? @FranzBusch / @Lukasa

Lukasa: I'm fine with that.

weissi:

PR #533, opened 2018-07-30
[EventLoopFuture] Ensure client #file/#line are forwarded in `mapIfError`.

ddunbar:
[EventLoopFuture] Ensure client #file/#line are forwarded in `mapIfError`.

### Motivation:

This ensures that the client's context information will be propagated into
the promise which is implicitly created by thenIfError(). Without this, if
that promise leaked the client would only see information about where it was
allocated inside of NIO, which is not helpful.

### Modifications:

Propagate #file/#line.

### Result:

Client's will see a #file/#line in their code, rather than something like:

```
Fatal error: leaking promise created at (file: ".../.build/checkouts/swift-nio.git-bla/Sources/NIO/EventLoopFuture.swift", line: 583)
```


ddunbar:
Note, I didn't do a full audit of EventLoopFuture.swift, but I did a partial one. The extensions at the bottom of the file are generally lacking propagation of #file/#line, and could use updating, but I didn't try tackling that here.

Lukasa:

PR #539, opened 2018-07-31
Deprecate and replace changeCapacity.

### LINKED ISSUE #537: Deprecate and replace ByteBuffer.changeCapacity
## The Problem

Unfortunately, `ByteBuffer.changeCapacity` is not a very good function. It has the following issues:

1. It's not clear whether users can safely use this function for shrinking the buffer as well as raising it. They can, but only if they don't try to shrink beyond the writer index, which the documentation does not mention.
2. The naming is inconsistent with Swift convention, which would generally call this function `reserveCapacity`.
3. This function is slower than it needs to be in the vast majority of cases:
    1. If the capacity we are being asked to have is the same as what we currently have, but the buffer is a slice, this will force a copy. There's no reason for that: while it's likely that a write will follow, it's not certain, and our CoW logic will catch that if it happens.
    2. This function will never call `realloc`: it will instead always call `malloc` + `memcpy`. That can be *extremely* slow in comparison to `realloc`: in cases where the `ByteBuffer` is uniquely referenced and is more than one page in size, `realloc` is an option and almost free in comparison to the `malloc`/`memcpy` invocation.

## Proposed Solution

1. Deprecate `ByteBuffer.changeCapacity`.
2. Provide two new functions: `ByteBuffer.reserveCapacity`, which will only ever enlarge the storage, and `ByteBuffer.shrink`, which will only ever shrink the storage. Both of these functions will be no-ops in cases where the capacity is already met, regardless of the number of references to the backing storage.
3. Implement the new functions using the following pattern:

    ```swift
        self._copyStorageAndRebase(capacity: newCapacity)
    } else {
        self._ensureAvailableCapacity(newCapacity, at: 0)
    }
    ```
4. Reimplement the now-deprecated `changeCapacity` in terms of the two

This set of changes should make the API substantially clearer, and also in many cases make it vastly faster when ensuring the capacity of a uniquely owned buffer, which can be handled simply via realloc instead of via the vastly slower `malloc`/`memcpy` dance.

Lukasa:
Motivation:

As described in #537, ByteBuffer.changeCapacity is not a very good
function. We want to replace it with a better one, more similar to
reserveCapacity on the standard library data structures, and ideally
substantially faster.

Modifications:

- Deprecated ByteBuffer.changeCapacity.
- Added ByteBuffer.reserveCapacity.
- Changed the implementation of changeCapacity to call reserveCapacity
  when we're increasing the capacity.

Result:

Better APIs, better performance in the legacy ones, everything is
better.

Resolves #537.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -437,6 +437,7 @@ public struct ByteBuffer {
     ///
     /// - parameters:
     ///     - to: The desired minimum capacity.
+    @available(*, deprecated, message: "changeCapacity has been replaced by reserveCapacity")
     public mutating func changeCapacity(to newCapacity: Int) {
         precondition(newCapacity >= self.writerIndex,
                      "new capacity \(newCapacity) less than the writer index (\(self.writerIndex))")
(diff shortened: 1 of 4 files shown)


Lukasa:
There were, not any more.

weissi:
thanks, still happy then 👍

PR #541, opened 2018-07-31
remove more memory binding assumptions

weissi:
Motivation:

It's poor style to assume a certain piece of memory is bound to a
certain type across a whole program. That's because any piece of code
run in between might have rebound the memory and also it's hard to prove
that the binding assumption is actually correct. Therefore we should
restrain from using them.

Modifications:

remove more assumptions that memory is bound to a certain type.

Result:

better code

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -27,6 +27,16 @@ let sysFree: @convention(c) (UnsafeMutableRawPointer?) -> Void = free
             self.copyBytes(from: src)
         }
     }
+    public extension UnsafeRawBufferPointer {
+        public func bindMemory<T>(to type: T.Type) -> UnsafeBufferPointer<T> {
+            guard let base = self.baseAddress else {
+                return UnsafeBufferPointer<T>(start: nil, count: 0)
+            }
+            let capacity = count / MemoryLayout<T>.stride
+            let ptr = base.bindMemory(to: T.self, capacity: capacity)
+            return UnsafeBufferPointer<T>(start: ptr, count: capacity)
+        }
+    }
 #endif
 
 extension _ByteBufferSlice: Equatable {
@@ -284,8 +294,8 @@ public struct ByteBuffer {
 
         public func dumpBytes(slice: Slice, offset: Int, length: Int) -> String {
             var desc = "["
-            for i in Int(slice.lowerBound) + offset ..< Int(slice.lowerBound) + offset + length {
-                let byte = self.bytes.advanced(by: i).assumingMemoryBound(to: UInt8.self).pointee
+            let bytes = UnsafeRawBufferPointer(start: self.bytes, count: Int(self.capacity))
+            for byte in bytes[Int(slice.lowerBound) + offset ..< Int(slice.lowerBound) + offset + length] {
                 let hexByte = String(byte, radix: 16)
                 desc += " \(hexByte.count == 1 ? "0" : "")\(hexByte)"
             }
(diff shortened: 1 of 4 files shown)


weissi:
ok, finally we have good diagnostics: It's a connect timeout:

```
20:01:27 - Test Case '-[NIOTests.BootstrapTest testBootstrapsCallInitializersOnCorrectEventLoop]' started.
20:01:27 - /Tests/NIOTests/BootstrapTest.swift:53: error: -[NIOTests.BootstrapTest testBootstrapsCallInitializersOnCorrectEventLoop] : failed - unexpected error connectTimeout(NIO.TimeAmount(nanoseconds: 10000000000)) thrown
20:01:27 - Fatal error: leaking promise created at (file: "/Tests/NIOTests/BootstrapTest.swift", line: 28): file /Tests/NIOTests/BootstrapTest.swift, line 28
20:01:27 - Exited with signal code 4
```

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -392,7 +402,8 @@ public struct ByteBuffer {
                "called the slower set<S: Sequence> function even though \(S.self) is a ContiguousCollection")
         func ensureCapacityAndReturnStorageBase(capacity: Int) -> UnsafeMutablePointer<UInt8> {
             self._ensureAvailableCapacity(_Capacity(capacity), at: index)
-            return self._storage.bytes.advanced(by: Int(self._slice.lowerBound + index)).assumingMemoryBound(to: UInt8.self)
+            return self._storage.bytes.advanced(by: Int(self._slice.lowerBound + index)).bindMemory(to: UInt8.self,
```
Lukasa: Do we need to bind the memory here at all?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -477,15 +477,23 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
     private func decodeHTTP(ctx: ChannelHandlerContext) throws {
         // We need to refetch the cumulationBuffer on each loop as it may has changed due re-entrance calls of channelRead(...)
         while let bufferSlice = self.cumulationBuffer, bufferSlice.readableBytes > 0 {
+            // we need to get `readerIndex` and `readableBytes` now because `withVeryUnsafeBytes` owns the
+            // `ByteBuffer` exlusively.
```
Lukasa: Nit: "exclusively".

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -477,15 +477,23 @@ public class HTTPDecoder<HTTPMessageT>: ByteToMessageDecoder, AnyHTTPDecoder {
...
-                self.state.baseAddress = pointer.baseAddress!.assumingMemoryBound(to: UInt8.self)
+                self.state.baseAddress = pointer.baseAddress!
                 defer {
                     self.state.baseAddress = nil
                 }
-                return self.state.baseAddress!.withMemoryRebound(to: Int8.self, capacity: pointer.count) { p in
-                    c_nio_http_parser_execute(&self.parser, &self.settings, p.advanced(by: bufferSlice.readerIndex), bufferSlice.readableBytes)
-                }
+                let httpParserPointer = UnsafeRawBufferPointer(start: pointer.baseAddress!.advanced(by: readerIndex),
+                                                               count: readableBytes).bindMemory(to: Int8.self)
```
Lukasa: I don't think this is safe.

By default the ByteBuffer address is bound to `UInt8` (we do it in `allocateAndPrepareRawMemory`). This bind, according to SE-107, *rebinds* the ByteBuffer memory to `Int8`, invalidating the assumption that this is bound to `UInt8`. This is fine, but if we think we're going to bind memory to `UInt8` in the initialiser we should just use `withMemoryRebound(to:)` here to put the old binding back afterwards.

PR #545, opened 2018-08-01
more debugging for flaky tests

weissi:
Motivation:

We suspect a DNS resolver issue in one of our test harnesses to cause
most of the flaky tests. This patch adds more debug information about
the resolver to flaky tests.

Modifications:

print the resolver output when flaky tests fail

Result:

one step closer to no flaky tests


weissi:

PR #549, opened 2018-08-02
add missing @discardableResults in readWith* methods

weissi:
Motivation:

ByteBuffer has a number of readWith* methods, some of them return the
`T` which was returned from the closure. This shouldn't be discardable.
Others however just return thr number of bytes read which should be
discardable.

Modifications:

made the readWith* functions that return the number of bytes read
@discardableResults

Result:

nicer user code


Lukasa:
Uh...that test appears to have failed for no reason.

Lukasa:

PR #550, opened 2018-08-06
Update CONTRIBUTORS and improve workflow.

Lukasa:
Motivation:

The CONTRIBUTORS file is getting out-of-date quickly, and we forget
to update it. We should make updating it part of the contribution flow
such that it remains current.

I noticed this because it got picked up in #525.

Modifications:

- Added a check to the sanity script to validate the CONTRIBUTORS
  file is up-to-date.
- Updated the mailmap file because @weissi doesn't know how to spell
  his own name.
- Minor clean up in the sanity script to handle the fact it checks
  three programming languages, so the output is a bit better.

Result:

Better sanity checking.


Lukasa:
Hrm, I think our Jenkins isn't checking out the full history. Gotta think whether we can/should change that.

Lukasa:
I think we need @tomerd to weigh in here, he owns the Jenkins config and I can't edit it.

CONTRIBUTORS.txt
```diff
@@ -34,9 +38,14 @@ needs to be listed here.
 - Lev Walkin <vlm@lionet.info>
 - Ludovic Dewailly <me@ldewailly.com>
 - Matt Eaton <agnosticdev@gmail.com>
+- Max Desiatov <max.desiatov@gmail.com>
```
maxdesiatov: can you please change this to max@desiatov.com? Thanks!

Lukasa: Done.

Lukasa:
I think we need to change the way the initial `git clone` is done. Can you find the setting where that's invoked and let me know what the current invocation is? Because I think the clone is probably shallow, and we don't want it to be.

weissi:
@Lukasa this has some minor whitespace issues:

```
08:50:04 -- Wilson Ding <hello@wilsonding.com>
08:50:04 -- fadi-botros <botros_fadi@yahoo.com>
08:50:04 -- sun_ <sunlubo.sun@gmail.com>
08:50:04 +
08:50:04  
08:50:04  **Updating this list**
```

(note the `+` at the end)

I'm taking this out of the 2.0.0 milestone because this can literally land anytime, the contributors are already up to date etc.

PR #552, opened 2018-08-06
tests hygiene: a lot fewer naked `try`s

weissi:
Motivation:

Naked `try`s in tests are a problem because if they're hit you'll get a
bad error message and likely no file/line information. That has made
many flaky tests harder to debug than necessary.

Modifications:

- fix a lot of naked `try`s
- symlink `TestUtils` from `NIOTests` into `NIOHTTP1Tests`

Result:

- better code
- easier to debug test failures
- setting better example


Tests/NIOHTTP1Tests/TestUtils.swift
```diff
@@ -0,0 +1 @@
+../NIOTests/TestUtils.swift
```
Lukasa: Rather than create this symlink, should we just move this into a target and depend on it from the two test targets?

weissi: @Lukasa SwiftPM doesn't unfortunately support that AFAIK. We can't make it a test target as we can't depend on that and we can't make it a normal target as it depends on `XCTest`. Therefore this is the only way I'm aware of...

Issue #548, opened 2018-08-02
EventLoopTest.testScheduleRepeatedTask flaky


weissi:
```
Test Case 'EventLoopTest.testScheduleRepeatedTask' started at 2018-08-01 19:58:11.911
/tmp/.nio-release-tools_rLD6qG/swift-nio/Tests/NIOTests/EventLoopTest.swift:117: error: EventLoopTest.testScheduleRepeatedTask : XCTAssertEqual failed: ("5") is not equal to ("6") - 
Test Case 'EventLoopTest.testScheduleRepeatedTask' failed (0.059 seconds)
```

CC @FranzBusch

weissi:
hit that again

```
16:24:57 Test Case 'EventLoopTest.testScheduleRepeatedTask' started at 2018-08-06 15:24:57.147
16:24:57 /code/Tests/NIOTests/EventLoopTest.swift:117: error: EventLoopTest.testScheduleRepeatedTask : XCTAssertEqual failed: ("5") is not equal to ("6") - 
16:24:57 Test Case 'EventLoopTest.testScheduleRepeatedTask' failed (0.081 seconds)
```

in here: https://ci.swiftnio.io/job/swift-nio-master-410-prb/414/console

weissi:
ok, I identified the race condition for this, working on a fix.

weissi:
ok, I identified the race condition for this, working on a fix.

PR #553, opened 2018-08-06
better wait on EventLoop detection

weissi:
Motivation:

It's important that in NIO programs you don't call
EventLoopFuture.wait() on an EventLoop. We previously had some detection
built in but we can do better.

Modifications:

Detect wait() called on any of MultiThreadedEventLoopGroup's EventLoops.

Result:

Quicker and more deterministic detection if wait() abuse.


Lukasa:
Is it worth us defining protocols we can let `EventLoop` and `EventLoopGroup`s opt in to that would allow us to make this more general? `wait` is not a high-performance call anyway, so the overhead of the indirection isn't necessarily a big deal, and it will let us extend this checking to things like `NIOTSEventLoopGroup`.

weissi:
@Lukasa sounds good to me! The only thing we've got is the `EventLoop` `self` though, right? How would we get from there to the group to ask it if we're in any of those?

Lukasa:
Yeah, we'd need two protocols. One would need to be a protocol that extends `EventLoop` with a `parentGroup` property, then we could have the one on the group that tells you whether you're currently running in an event loop context.

weissi:
this hit #548 too (fixed in #555)

weissi:

weissi:
@Lukasa do you want that those new protocols before this work or should we merge like this and later improve how it's done so NIOTS can benefit too?

Lukasa:
Let's merge this now, and we can enhance this with new protocols later.

PR #558, opened 2018-08-07
Make withUnsafeMutable{Readable,Writable}Bytes inlinable.

Lukasa:
Motivation:

I spotted this while flamegraphing: apparently we missed the
declaration for inlineability here.

Modifications:

Made these two functions inlineable.

Result:

Better optimizations.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -494,7 +494,8 @@ public struct ByteBuffer {
         }
     }
 
-    private mutating func copyStorageAndRebaseIfNeeded() {
+    @_versioned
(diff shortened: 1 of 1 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -494,7 +494,8 @@ public struct ByteBuffer {
         }
     }
 
-    private mutating func copyStorageAndRebaseIfNeeded() {
+    @_versioned
+    /* private but versioned */ mutating func copyStorageAndRebaseIfNeeded() {
```
weissi: we made all the others underscored, so this should be too.

weissi: can then also remove the comment

PR #560, opened 2018-08-08
make ByteBuffer._Storage.fullSlice a computed property

weissi:
Motivation:

ByteBuffer._Storage.fullSlice can always be trivially computed, so it
should be.

Modifications:

make ByteBuffer._Storage.fullSlice computed

Result:

saved almost 56 bits per ByteBuffer 😂

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -231,18 +231,20 @@ public struct ByteBuffer {
     @_versioned final class _Storage {
         private(set) var capacity: _Capacity
         @_versioned private(set) var bytes: UnsafeMutableRawPointer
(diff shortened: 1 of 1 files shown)


Lukasa:
@weissi Looks like an integration test failed here, want to take a quick look?

weissi:
@Lukasa thanks filed as #563 will investigate shortly.

weissi:

weissi:
hit a slow node:

```
14:47:25 Fatal error: Posix wrapper adds more than 20% overhead (with wrapper: 0.751054048538208, without: 0.594343900680542: file /tmp/.swift-nio-syscall-wrappers-sh-test_71I49G/syscallwrapper/Sources/syscallwrapper/SystemCallWrapperHelpers.swift, line 124
```

PR #561, opened 2018-08-09
Remove unnecessary import in c-atomics.h

Lukasa:
We don't need this import, so let's not bring it along.


Lukasa:
Backporting to 1.9.

PR #567, opened 2018-08-10
use cross-testTarget dependencies for TestUtils

weissi:
Motivation:

SwiftPM seems to support cross-testTarget dependencies when running
`swift test`.

Modifications:

use cross-testTarget dependencies instead of symlink

Result:

better


weissi:
oh no, the generated Xcode project doesn't work then, let's close that again.

PR #568, opened 2018-08-10
Add Cocoapods podspec generation.

Lukasa:
Motivation:

A number of users have asked for the ability to install SwiftNIO
via Cocoapods. This is an entirely reasonable request, so let's
do it.

Modifications:

- Added a script that can generate, and optionally upload,
  podspecs.
- Made a slight modification to the dependency script to provide
  target-specific dependencies.

Result:

Users can use cocoapods if they prefer.

### DIFF
--- a/scripts/build_podspecs.sh
+++ b/scripts/build_podspecs.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+set -eu
+
+function usage() {
+  echo "$0 [-u] VERSION"
+  echo
+  echo "OPTIONS:"
+  echo "  -u: Additionally, upload the podspecs as they are generated"
+}
+
+OPTIND=1
+upload=false
+while getopts ":u" opt; do
+  case $opt in
+    u)
+      upload=true
+      ;;
+    \?)
+      usage
+      exit 1
+      ;;
+  esac
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+version=$1
+newline=$'\n'
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpfile=$(mktemp -d /tmp/.build_podspecsXXXXXX)
+echo "Building podspecs in $tmpfile"
+
+# We have a name transformation here. We want the pod names to be
+# SwiftX or CX, but our target names aren't. We add SwiftX to the front
+# of all NIO targets, and then remove it in the source files declaration
+# if needed.
+targets=( $("${here}/list_topsorted_dependencies.sh" -l -r | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/') )
+
+for target in "${targets[@]}"; do
+  echo "Building podspec for $target"
+
+  dependencies=()
+
+  while read -r raw_dependency; do
+    dependencies+=( "${newline}  s.dependency '$raw_dependency', s.version.to_s" )
+  done < <("${here}/list_topsorted_dependencies.sh" -d "${target#Swift}" | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
+
+  libraries=""
+
+  if [[ "$target" == "CNIOZlib" ]]; then
+    libraries="s.libraries = 'z'"
+  fi
+
+  cat > "${tmpfile}/${target}.podspec" <<- EOF
+Pod::Spec.new do |s|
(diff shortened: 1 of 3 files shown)


scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+
+function usage() {
+  echo "$0 [-u] VERSION"
+  echo
+  echo "OPTIONS:"
+  echo "  -u: Additionally, upload the podspecs as they are generated"
+}
+
+upload=false
+while getopts ":u" opt; do
```
weissi: - I think you don't want the colon in there, right? The colon (usually after an option name) means that the option gets an argument. Or am I missing something.
- Also I'd recommend to do `OPTIND=1` before the `while`. Just in case someone was nasty enough to `export OPTIND=7` or so before running the script

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+tmpfile=$(mktemp -d /tmp/.build_podspecsXXXXXX)
+echo "Building podspecs in $tmpfile"
+
+# We have a name transformation here. We want the pod names to be
+# SwiftX or CX, but our target names aren't. We add SwiftX to the front
+# of all NIO targets, and then remove it in the source files declaration
+# if needed.
+TARGETS=$(${here}/list_topsorted_dependencies.sh -l -r | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
```
weissi: that needs to be quoted, maybe run `shellcheck`

weissi: and I think if you do

```
targets=( $( ... ) )
```

then targets becomes an array

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+
+for target in $TARGETS; do
```
weissi: once targets is an array you can loop like `for target in "${targets[@]}"; do`

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+  esac
+done
+shift "$((OPTIND-1))"
+
+if [[ $# -eq 0 ]]; then
+  echo "Must provide target version"
+  exit 1
+fi
+
+VERSION=$1
```
weissi: we should use lower case, upper case are reserved by shell for stuff with special meaning

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+# of all NIO targets, and then remove it in the source files declaration
+# if needed.
+TARGETS=$(${here}/list_topsorted_dependencies.sh -l -r | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
+
+for target in $TARGETS; do
+  echo "Building podspec for $target"
+
+  # We need to suppress the return code of grep here.
+  RAW_DEPENDENCIES=$(${here}/list_topsorted_dependencies.sh -d "${target#Swift}" | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
+  DEPENDENCIES=""
```
weissi: this should be an array

    dependencies=()

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+
+  for raw_dependency in $RAW_DEPENDENCIES; do
+    DEPENDENCIES="${DEPENDENCIES}${NEWLINE}  s.dependency '$raw_dependency', s.version.to_s"
```
weissi: this should be array append

    dependencies+=( "${dependencies}${newline}  s.dependency '$raw_dependency', s.version.to_s" )

scripts/build_podspecs.sh
```diff
@@ -0,0 +1,103 @@
...
+# SwiftX or CX, but our target names aren't. We add SwiftX to the front
+# of all NIO targets, and then remove it in the source files declaration
+# if needed.
+TARGETS=$(${here}/list_topsorted_dependencies.sh -l -r | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
+
+for target in $TARGETS; do
+  echo "Building podspec for $target"
+
+  # We need to suppress the return code of grep here.
+  RAW_DEPENDENCIES=$(${here}/list_topsorted_dependencies.sh -d "${target#Swift}" | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
```
weissi: generally I think looping over output is better done like this:

    while read -r line; do
        # process each line
    done < <(command that output lines)

scripts/list_topsorted_dependencies.sh
```diff
@@ -53,6 +58,12 @@ if $do_reversed; then
     transform=tac_compat
 fi
 
+if [[ ! -z "$package_dependency" ]]; then
+  swift package dump-package | jq -r ".targets |
+                                      map(select(.name == \"$package_dependency\" and .type == \"regular\") | .dependencies | map(.name)) | .[] | .[]"
```
weissi: can we break that `jq` in multiple lines (maybe after each `|`)?

Lukasa:
Feedback addressed.

fcorneli:
Could this be updated for version 1.9.5? Only 1.9.2 is available for now.

weissi:
@fcorneli thanks for the report, updating them right now! I'll also add that to our release checklist so we shouldn't forget in the future...

weissi:
ok, 1.9.5. uploaded

Issue #563, opened 2018-08-09
integration test test_21_connection_reset_tcp.sh flaky


weissi:
happened [in CI](https://ci.swiftnio.io/job/swift-nio-master-403-prb/445/console) on

```
11:54:28 + docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.1404.403.yaml -p swift-nio-master-403-prb run --rm test
```

so that's Swift 4.0.3 on Ubuntu 14.04

```
12:02:32 Running test 'test_21_connection_reset_tcp.sh'... FAILURE (4)
[...]
12:02:32 ++ /code/.build/x86_64-unknown-linux/debug/NIOHTTP1Server --disable-half-closure 127.0.0.1 0 /tmp/.swift-nio-http1-server-sh-tests_EGvGeF/test.tmp_sGPeP8/htdocs
12:02:32 COMMAND    PID USER   FD   TYPE    DEVICE SIZE/OFF      NODE NAME
12:02:32 NIOHTTP1S 5157 root    0r   CHR       1,3      0t0 113560129 /dev/null
12:02:32 NIOHTTP1S 5157 root    1w   REG    202,80     2162   6160564 /tmp/.swift-nio-http1-server-sh-tests_EGvGeF/test.out_jM5VRc
12:02:32 NIOHTTP1S 5157 root    2w   REG    202,80     2162   6160564 /tmp/.swift-nio-http1-server-sh-tests_EGvGeF/test.out_jM5VRc
12:02:32 NIOHTTP1S 5157 root    3w  FIFO       0,8      0t0 113559855 pipe
12:02:32 NIOHTTP1S 5157 root    4w  FIFO       0,8      0t0 113559856 pipe
12:02:32 NIOHTTP1S 5157 root    5u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root    6u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root    7u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root    8u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root    9u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root   10u  0000       0,9        0      5987 anon_inode
12:02:32 NIOHTTP1S 5157 root   11u  IPv4 113687421      0t0       TCP localhost:35615 (LISTEN)
[...]

PR #573, opened 2018-08-14
output a hint if we get a reset connection w/o error set

weissi:
Motivation:

We have one case in the code where we expect a socket error to be set
when we hit a connection that got the `.reset` event. On Linux that's
expected and therefore we just map this to `ECONNRESET`.
On other platforms however that is unexpected. Now I just found an
[Envoy commit](https://github.com/envoyproxy/envoy/commit/62f8e51bdd2d72cbac739087dabd3f51bdfc5cbd#diff-3315c7b2    91f78a482ebd4cbfc0545362R345)
which suggests this might not work on FreeBSD. This should prompt our
users to let us know if we ever see this error on non-Linux systems.

Modifications:

Add a message that makes it clear this is unexpected and link our
ticket.

Result:

we should learn if this actually happens on macOS.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -921,7 +921,12 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
                 } else {
                     // we don't have a socket error, this must be connection reset without an error then
                     // this path should only be executed on Linux (EPOLLHUP, no EPOLLERR)
-                    error = IOError(errnoCode: ECONNRESET, reason: "connection reset (no error set)")
+                    #if os(Linux)
+                    let message: String = "connection reset (no error set)"
+                    #else
+                    let message: String = "BUG IN SwiftNIO (possibly #572), please report! Connection reset (no error set)."
+                    #endif
+                    error = IOError(errnoCode: ECONNRESET, reason: message)
                 }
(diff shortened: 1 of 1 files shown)


weissi:
hit #563 retring

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -921,7 +921,12 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
                 } else {
                     // we don't have a socket error, this must be connection reset without an error then
                     // this path should only be executed on Linux (EPOLLHUP, no EPOLLERR)
-                    error = IOError(errnoCode: ECONNRESET, reason: "connection reset (no error set)")
+                    #if os(Linux)
+                    let message: String = "connection reset (no error set)"
+                    #else
+                    let message: String = "BUG IN SwiftNIO (possibly #572), please report! Connection reset (no error set)."
```
Lukasa: Should this `precondition` instead?

weissi: @Lukasa I was on the fence but it doesn't actually cause any bad behaviour at all. It's "just" potentially not quite the error message in that case.

Lukasa: Heh, I suspect that even if this path does execute on Darwin no-one will *ever* report a bug with it. But it doesn't hurt.

PR #580, opened 2018-08-16
fix limits on 32-bit systems to values that work

weissi:
Motivation:

We should have different constants for 32-bit systems where necessary
for example all the maximums should be Int32.max and not UInt32.max
because UInt32.max doesn't fit into an Int on 32-bit platforms. But
where unnecessary 32-bit and 64-bit should share maximum values.

Modifications:

- removed some constraints that should be unnecessary
- skip tests that can't work on 32-bit systems

Result:

better and more consistent 32-bit runs


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -105,10 +105,6 @@ public struct ByteBufferAllocator {
                   hookedRealloc: @escaping @convention(c) (UnsafeMutableRawPointer?, size_t) -> UnsafeMutableRawPointer?,
                   hookedFree: @escaping @convention(c) (UnsafeMutableRawPointer?) -> Void,
                   hookedMemcpy: @escaping @convention(c) (UnsafeMutableRawPointer, UnsafeRawPointer, size_t) -> Void) {
-        #if !arch(arm) // only complain on 64-bit, this is unfortunate reality on 32-bit
```
weissi: note to @Lukasa: This isn't necessary and covered much better by our allocation testing tests.

PR #585, opened 2018-08-19
Enhanced the output of the NIOEchoClient Example

agnosticdev:
Motivation:

I was testing the NIOEchoClient example on Linux and noticed that I received no character output.
I thought this would be handy and added this code to do so.

Modifications:

Small modification to the channelRead method in NIOEchoClient to read the bytes and transform to a string.

Result:

The display of the characters on your echo client before the channel closes.

### DIFF
--- a/Sources/NIOEchoClient/main.swift
+++ b/Sources/NIOEchoClient/main.swift
@@ -23,19 +23,14 @@ private final class EchoHandler: ChannelInboundHandler {
 
 
     public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
-
-        var byteBuffer = self.unwrapInboundIn(data) as ByteBuffer
+        var byteBuffer = self.unwrapInboundIn(data)
         numBytes -= byteBuffer.readableBytes
 
         assert(numBytes >= 0)
 
         if numBytes == 0 {
-            if let charArray = byteBuffer.readBytes(length: byteBuffer.readableBytes) {
-                var str = ""
-                for byte in charArray {
-                    str += String(UnicodeScalar(byte))
-                }
-                print("Received: '\(str)' back from the server, closing channel.")
+            if let string = byteBuffer.readString(length: byteBuffer.readableBytes) {
+                print("Received: '\(string)' back from the server, closing channel.")
             }
             ctx.close(promise: nil)
         }
--- a/Sources/NIOEchoClient/main.swift
+++ b/Sources/NIOEchoClient/main.swift
@@ -12,7 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 import NIO
-import Foundation
 
 print("Please enter line to send to the server")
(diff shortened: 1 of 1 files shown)


Sources/NIOEchoClient/main.swift
```diff
@@ -21,13 +21,22 @@ private final class EchoHandler: ChannelInboundHandler {
...
 
         assert(numBytes >= 0)
 
         if numBytes == 0 {
-            print("Received the line back from the server, closing channel")
+            if let charArray = byteBuffer.readBytes(length: byteBuffer.readableBytes) {
+                var str = ""
+                for byte in charArray {
+                    str += String(UnicodeScalar(byte))
+                }
```
Lukasa: Ah, with this tighter loop it's a lot easier to see what we're getting into here. This can all be replaced with `if let string = byteBuffer.readString(length: byteBuffer.readableBytes)`. We'll also want an error message for the non-bound branch.

Sources/NIOEchoClient/main.swift
```diff
@@ -21,13 +21,22 @@ private final class EchoHandler: ChannelInboundHandler {
     public typealias OutboundOut = ByteBuffer
     private var numBytes = 0
 
+
     public func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
-        numBytes -= self.unwrapInboundIn(data).readableBytes
+
+        var byteBuffer = self.unwrapInboundIn(data) as ByteBuffer
```
Lukasa: We don't need the `as` or the extra whitespace line.

Sources/NIOEchoClient/main.swift
agnosticdev: Thank you!  Using `if let string = byteBuffer.readString(length: byteBuffer.readableBytes)` is a lot more concise than using the loop I have above.

Sources/NIOEchoClient/main.swift
```diff
@@ -21,13 +21,17 @@ private final class EchoHandler: ChannelInboundHandler {
     public typealias OutboundOut = ByteBuffer
     private var numBytes = 0
 
+
```
Lukasa: I don't think we need this extra whitespace line either.

Sources/NIOEchoClient/main.swift
```diff
@@ -21,13 +21,17 @@ private final class EchoHandler: ChannelInboundHandler {
...
+        var byteBuffer = self.unwrapInboundIn(data)
+        numBytes -= byteBuffer.readableBytes
 
         assert(numBytes >= 0)
 
         if numBytes == 0 {
-            print("Received the line back from the server, closing channel")
+            if let string = byteBuffer.readString(length: byteBuffer.readableBytes) {
+                print("Received: '\(string)' back from the server, closing channel.")
+            }
```
Lukasa: I think we need a print clause in the implicit else branch here too.

PR #589, opened 2018-08-21
Add SocketOptionChannel for wider socket options.

Lukasa:
Hey @weissi @normanmaurer, I'd like to solicit extra feedback from the two of you here. There are a lot of possible designs we could use for this and this is only one of them, so please feel free to provide substantial design review as well as regular old code review.

Motivation:

A small number of socket options have values that do not fit into a C
int type. Our current ChannelOption based approach for setting these
simply does not work, and cannot be extended to support the truly arbitrary
types that the setsockopt/getsockopt functions allow here.

This makes it impossible to use some socket options, which is hardly a
good place to be.

There were a number of ways we could have addressed this: we could have
special-cased all socket options with non-integer types in ChannelOption,
but I believe that would be too manual, and risk limiting users that need
to set other socket options. We could also have added a ChannelOption
that simply allows users to pass a buffer to write into or read from,
but that is a very un-Swift-like API that feels pretty gross to hold.

Ultimately, the nicest seemed to be a new protocol users could check
for, and that would provide APIs that let users hold the correct concrete
type. As with setsockopt/getsockopt, while this API is typed it is
not type-safe: ultimately, the struct we have here is treated just as a
buffer by setsockopt/getsockopt. We do not attempt to prevent users from
shooting themselves in the foot here.

This PR does not include an example use case in any server, as I will
provide such an example in a subsequent multicast PR.

Modifications:

- Added a SocketOptionChannel protocol.
- Conformed BaseSocketChannel to SocketOptionChannel.
- Wrote some tests for this.

Result:

Users can set and get sockopts with data that is larger than a C int.


Sources/NIO/SocketOptionChannel.swift
```diff
@@ -0,0 +1,103 @@
...
+    func setSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) -> EventLoopFuture<Void>
+
+    /// Obtain the value of the socket option for the given level and name.
+    ///
+    /// - parameters:
+    ///     - level: The socket option level, e.g. `SOL_SOCKET` or `IPPROTO_IP`.
+    ///     - name: The name of the socket option, e.g. `SO_REUSEADDR`.
+    /// - returns: An `EventLoopFuture` containing the value of the socket option, or
+    ///     any error that occurred while retrieving the socket option.
+    func getSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName) -> EventLoopFuture<Value>
```
weissi: - should we make this have an 'unsafe' or similar word in there?
- other idea would be to make `Value: CSockOptValue` (or similar) and make that a marker protocol which can be assigned to all known types that `set/getsockopt` allow. so `extension timeval: CSockOptValue {} extension CInt: CSockOptValue {}` etc.

weissi: I just want to prevent users from passing arbitrary stuff (like say an array of bytes) to that `Value` and never ever getting a compiler error. It's really a universal blackhole.

tomerd:

tomerd:

tomerd:

tomerd:

Lukasa:
Ok, this is a new design in place now. Right now the only example of the "safe" API is `SO_LINGER`, because the other good choices all only work on multicast sockets, which this PR blocks.

Ready for re-review.

Sources/NIO/SocketOptionChannel.swift
```diff
@@ -0,0 +1,147 @@
...
+        return self.unsafeGetSocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_LINGER)
+    }
+}
+
+
+extension BaseSocketChannel: SocketOptionChannel {
+    public func unsafeSetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) -> EventLoopFuture<Void> {
+        if eventLoop.inEventLoop {
+            let promise: EventLoopPromise<Void> = eventLoop.newPromise()
+            executeAndComplete(promise) { try setSocketOption0(level: level, name: name, value: value) }
```
weissi: @normanmaurer/@lukasa what do you think about not having a new line after `{` ? I think we generally don't do it in NIO but I also don't mind it too much (but I have certainly made external contributors add extra newlines).

Tests/NIOTests/SocketOptionChannelTest.swift
```diff
@@ -0,0 +1,129 @@
...
+        guard let socketOptionChannel = self.clientChannel as? SocketOptionChannel else {
+            XCTFail("Unable to cast \(String(describing: self.clientChannel)) to SocketOptionChannel", file: file, line: line)
+            throw CastError()
+        }
+        return socketOptionChannel
+    }
+
+    override func setUp() {
+        self.group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        self.serverChannel = try? assertNoThrowWithValue(ServerBootstrap(group: group).bind(host: "127.0.0.1", port: 0).wait())
```
weissi: isn't `setUp` `throws`? If yes, then I think it'd be better to do `try` instead of `try?`

Lukasa: Nop, `setUp` is not throws.

Sources/NIO/SocketOptionChannel.swift
Lukasa: Eh, I just added the newlines in.

Tests/NIOTests/SocketOptionChannelTest.swift
weissi: cool, please disregard then :)

weissi:
hit #563 :|

weissi:

weissi:
@normanmaurer / @Lukasa there's one comment from Norman left (but he approved it), how should we proceed?

normanmaurer:
@weissi I am fine with merging... Just wondered what would be a "swift-y" name :)

tomerd:
@weissi i have been seeing the tests failing on what looks to be flakiness a few times friday. probably worth looking into?

weissi:
@tomerd indeed. So I've spent quite some time to investigate already and so far I have been unable to reproduce locally. It's pretty mysterious at this points, see the comments in #563. I might raise the wait before querying the number of open file descriptors and see if it's just slowness of the CI maybe. But the odd thing is that is _only ever_ happened on 4.0.3 and not on 4.1.

Lukasa:

Lukasa:
Ok, after further discussion about naming I've rewritten this protocol a bit. It's now called `SocketOptionProvider`, and is entirely unrelated to the `Channel` protocol and stands alone.

This does raise a question about whether we need to enhance the bootstraps to use this protocol though. Thoughts?

PR #590, opened 2018-08-23
warn that ChannelHandlers should be instantiated within initializer c…

weissi:
…losures

Motivation:

For (stateful) ChannelHandlers it's very important that the
ChannelHandlers get instantiated _within_ the bootstrap's initializer
closures (and not before that).

Modifications:

Add warnings about the fact that they should be instantiated within the
closures.

Result:

- better docs
- more correct user code

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -27,6 +27,8 @@
 ///         .childChannelInitializer { channel in
 ///             // Ensure we don't read faster then we can write by adding the BackPressureHandler into the pipeline.
 ///             channel.pipeline.add(handler: BackPressureHandler()).then { () in
+///                 // make sure to instantiate your `ChannelHandlers` inside of
+///                 // the closure as it will be invoked once per connection.
 ///                 channel.pipeline.add(handler: MyChannelHandler())
 ///             }
 ///         }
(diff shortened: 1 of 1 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -352,6 +361,17 @@ public final class ClientBootstrap {
...
+    /// - warning: The `handler` closure may be invoked _multiple times_ so always instantiate
+    ///            `ChannelHandler`s within `handler`. The reason `handler` may be invoked multiple
+    ///            times is that to successfully set up a connection multiple connections might be
+    ///            setup in the process. Assuming a hostname that resolves to both IPv4 and IPv6
+    ///            addresses, NIO will follow [_Happy Eyeballs_](https://en.wikipedia.org/wiki/Happy_Eyeballs)
+    ///            and race both an IPv4 and an IPv6 connection. It is possible that both
+    ///            connections get fully established before the IPv4 connection will be closed
+    ///            again because the IPv6 connection 'won the race'. Therefore the
+    ///            `channelInitializer` might be called multiple times and it's important not to
+    ///            share stateful `ChannelHandler`s in more than one `Channel`.
```
normanmaurer: @weissi I think this is not a 100 % correct advice... Sometimes you want to share the handler even if it has state (for example if you would count the connections). I think we should just say that if its stateful and shared it must be thread-safe.

weissi: @normanmaurer I know, found it hard to get the message across that "always instantiate inside, unless you know what you're doing :P"

weissi: ok, have an idea, let's phrase like this:

> Therefore stateful `ChannelHandler`s should generally be instantiated within the closure. There are use-cases where a stateful `ChannelHandler` needs to be shared between multiple `Channel`s in which case the user is responsible for correctly synchronising the access to the state.

weissi:
@normanmaurer updated

PR #592, opened 2018-08-23
Fix readme erroneous statement regarding channel pipeline

arukshani:
Fix readme erroneous statement regarding channel pipeline not being thread safe

### Motivation:

Readme says channel pipeline is not tread safe which is incorrect

### Modifications:

Correct ChannelPipeline description so that it says thread safe.

### Result:

Statement regarding channel pipeline will be correct


Lukasa:

PR #595, opened 2018-08-24
pipeline handler: fix error state tracking

weissi:
Motivation:

@pushkarnk hit an interesting edge case that we previously mishandled
and crashed. All these conditions need to be true

1. Whilst have an ongoing request (ie. outstanding response), ...
2. ... a `HTTPParserError` arrives (which we correctly queue)
3. Later the outstanding response is completed and then ...
4. ... the `HTTPServerProtocolErrorHandler` sends a .badRequest response

We previously crashed. The reason we crashed is that the
`HTTPServerPipelineHandler` obviously tracks state and then asserts that
no response is sent for a wrong request. It does have an affordance to
allow a .badRequest response for a request it couldn't parse. However
this state tracking wasn't done if the error itself was enqueued for
later delivery.

Thanks very much @pushkarnk for the report!

Modifications:

instead of delivering the error directly use the `deliverOneError`
function which transitions the state correctly.

Result:

fewer crashes & hopefully happy Pushkar

### DIFF
--- a/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest+XCTest.swift
+++ b/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest+XCTest.swift
@@ -43,6 +43,8 @@ extension HTTPServerPipelineHandlerTest {
                 ("testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndBeforeResponseEnd", testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndBeforeResponseEnd),
                 ("testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndAfterResponseEnd", testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndAfterResponseEnd),
                 ("testQuiescingAfterHavingReceivedOneRequestButBeforeResponseWasSentWithMoreRequestsInTheBuffer", testQuiescingAfterHavingReceivedOneRequestButBeforeResponseWasSentWithMoreRequestsInTheBuffer),
+                ("testParserErrorOnly", testParserErrorOnly),
+                ("testLegitRequestFollowedByParserErrorArrivingWhilstResponseOutstanding", testLegitRequestFollowedByParserErrorArrivingWhilstResponseOutstanding),
            ]
    }
 }
--- a/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
+++ b/Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
@@ -20,7 +20,7 @@
 
 import XCTest
 import NIO
-import NIOHTTP1
+@testable import NIOHTTP1
 
 private final class ReadRecorder: ChannelInboundHandler {
     typealias InboundIn = HTTPServerRequestPart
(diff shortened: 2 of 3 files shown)


Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -706,4 +706,94 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+            func errorCaught(ctx: ChannelHandlerContext, error: Error) {
+                XCTAssertEqual(HTTPParserError.closedConnection, error as? HTTPParserError)
+                XCTAssertEqual(.errorExpected, self.state)
+                self.state = .done
+            }
+
+            func channelRead(ctx: ChannelHandlerContext, data: NIOAny) {
+                switch self.unwrapInboundIn(data) {
+                case .head:
+                    ctx.eventLoop.execute {
```
normanmaurer: Can you add a comment to explain why this is dispatched on the loop here?

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -706,4 +706,94 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+                                                                           status: .ok))),
+                                          promise: nil)
+                    }
+                    XCTAssertEqual(.reqHeadExpected, self.state)
+                    self.state = .reqEndExpected
+                case .body:
+                    XCTFail("no body expected")
+                case .end:
+                    ctx.eventLoop.execute {
+                        ctx.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: nil)
```
normanmaurer: Can you add a comment to explain why this is dispatched on the loop here?

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
weissi: addressed

tomerd:

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest+XCTest.swift
```diff
@@ -43,6 +43,8 @@ extension HTTPServerPipelineHandlerTest {
                 ("testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndBeforeResponseEnd", testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndBeforeResponseEnd),
                 ("testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndAfterResponseEnd", testQuiescingAfterRequestAndResponseHeadsButBeforeAnyEndsThenRequestEndAfterResponseEnd),
                 ("testQuiescingAfterHavingReceivedOneRequestButBeforeResponseWasSentWithMoreRequestsInTheBuffer", testQuiescingAfterHavingReceivedOneRequestButBeforeResponseWasSentWithMoreRequestsInTheBuffer),
+                ("testParserErrorOnly", testParserErrorOnly),
+                ("testLegitRequestFollowedByParserErrorArrivingWhilstReponseOutstanding", testLegitRequestFollowedByParserErrorArrivingWhilstReponseOutstanding),
```
ianpartridge: `Reponse` -> `Response`.

weissi: fixing

PR #601, opened 2018-08-30
Include ipaddress in SocketAddress.description.

normanmaurer:
Motivation:

WE should include the ipaddress in the description of SocketAddress if any way used as there may be no host at all.

Modifications:

- Always include ipaddress if possible.
- Add unit tests.

Result:

More useful description of the SocketAddress

### DIFF
--- a/Sources/NIO/SocketAddresses.swift
+++ b/Sources/NIO/SocketAddresses.swift
@@ -75,28 +75,38 @@ public enum SocketAddress: CustomStringConvertible {
 
     /// A human-readable description of this `SocketAddress`. Mostly useful for logging.
     public var description: String {
+        let addressString: String
         let port: String
         let host: String?
         let type: String
         switch self {
         case .v4(let addr):
-            host = addr.host
+            host = addr.host.isEmpty ? nil : addr.host
             type = "IPv4"
+            var mutAddr = addr.address.sin_addr
+            addressString = descriptionForAddress(family: AF_INET, bytes: &mutAddr, length: Int(INET_ADDRSTRLEN))
+
             port = "\(self.port!)"
         case .v6(let addr):
-            host = addr.host
+            host = addr.host.isEmpty ? nil : addr.host
             type = "IPv6"
+            var mutAddr = addr.address.sin6_addr
+            addressString = descriptionForAddress(family: AF_INET6, bytes: &mutAddr, length: Int(INET6_ADDRSTRLEN))
+    
             port = "\(self.port!)"
         case .unixDomainSocket(let addr):
             var address = addr.address
             host = nil
             type = "UDS"
+            addressString = ""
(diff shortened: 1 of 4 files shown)


Sources/NIO/SocketAddresses.swift
```diff
@@ -75,28 +75,48 @@ public enum SocketAddress: CustomStringConvertible {
 
     /// A human-readable description of this `SocketAddress`. Mostly useful for logging.
     public var description: String {
+        let addressString: String;
```
weissi: extra semicolon

Sources/NIO/SocketAddresses.swift
```diff
@@ -75,28 +75,48 @@ public enum SocketAddress: CustomStringConvertible {
...
         let host: String?
         let type: String
         switch self {
         case .v4(let addr):
-            host = addr.host
+            host = addr.host.isEmpty ? nil : addr.host
             type = "IPv4"
+            var mutAddr = addr.address.sin_addr;
+            var addressBuffer = [CChar](repeating:0, count: Int(INET_ADDRSTRLEN))
+            if let value = inet_ntop(AF_INET, &mutAddr, &addressBuffer, socklen_t(INET_ADDRSTRLEN)) {
```
weissi: we already have this function

```
private func descriptionForAddress(family: CInt, bytes: UnsafeRawPointer, length byteCount: Int) -> String {
    var addressBytes: [Int8] = Array(repeating: 0, count: byteCount)
    return addressBytes.withUnsafeMutableBufferPointer { (addressBytesPtr: inout UnsafeMutableBufferPointer<Int8>) -> String in
        try! Posix.inet_ntop(addressFamily: family,
                             addressBytes: bytes,
                             addressDescription: addressBytesPtr.baseAddress!,
                             addressDescriptionLength: socklen_t(byteCount))
        return addressBytesPtr.baseAddress!.withMemoryRebound(to: UInt8.self, capacity: byteCount) { addressBytesPtr -> String in
            String(cString: addressBytesPtr)
        }
    }
}
```

maybe make `internal` and use here?

Sources/NIO/SocketAddresses.swift
```diff
@@ -75,28 +75,48 @@ public enum SocketAddress: CustomStringConvertible {
...
+            }
             port = "\(self.port!)"
         case .v6(let addr):
-            host = addr.host
+            host = addr.host.isEmpty ? nil : addr.host
             type = "IPv6"
+            var mutAddr = addr.address.sin6_addr;
+
+            var addressBuffer = [CChar](repeating:0, count: Int(INET6_ADDRSTRLEN))
+            if let value = inet_ntop(AF_INET, &mutAddr, &addressBuffer, socklen_t(INET6_ADDRSTRLEN)) {
```
weissi: same here

normanmaurer:
@weissi addressed PTAL again

PR #604, opened 2018-09-01
add Ubuntu 18.04 and Swift 4.2 support

weissi:
Motivation:

We should support the latest and greatest.

Modifications:

- add Ubuntu 18.04 support (packages exist for Swift 4.2 only)
- add Swift 4.2 support (for Ubuntu 14.04, 16.04 and 18.04)

Result:

- supporting the latest stuff

### DIFF
--- a/IntegrationTests/tests_01_http/defines.sh
+++ b/IntegrationTests/tests_01_http/defines.sh
@@ -167,3 +167,18 @@ function do_curl() {
             ;;
     esac
 }
+
+_new_nc=false
+if nc -h 2>&1 | grep -- -N | grep -q EOF; then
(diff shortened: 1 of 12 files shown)


tomerd:
on he surface his looks fine, did you have a chance to test the dockerfile changes for other versions / permutations?

additionally, how do you see us using this? should we add a pull request tester, or should we move some of these to nightlies?

weissi:
@tomerd I did (manually) run most of the configurations to test it and they seem to all work. I think we should definitely add a PR tester for at least 18.04 + 4.2 because I think in a couple of weeks time that'll probably be the most common configuration people will run.

PR #610, opened 2018-09-10
Actually report errors when hitting blacklisted errnos.

Lukasa:
Motivation:

It's not all that useful to print the value of the pointer from sterror.
Instead we should aim to print the string. While we're here, let's avoid
repeating too much code.

Modifications:

- Centralised the handling of blacklisted errnos.
- Correctly printed the error string when a blacklisted errno is hit.
- Updated the output of the assertion to include the function failing.

Result:

Better diagnostics when blacklisted errnos are hit.


weissi:
haha

Issue #614, opened 2018-09-12
IdleStateHandler should forward forward channelActive events


ldewailly:
### Expected behavior
If I add a `ChannelInboundHandler` after a `IdleStateHandler`, I'd expect to receive the `channelActive` event in my handler

### Actual behavior
`channelActive(ctx)` is never invoked on the channel inbound handler.

### Steps to reproduce

Create a pipeline as illustrated below and establish a connection. The second handler doesn't receive the `channelActive` event

### If possible, minimal yet complete reproducer code (or URL to code)

```swift
// start the server
let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
final class ServerHandler: ChannelInboundHandler {
    typealias InboundIn = ByteBuffer
}
defer { try? group.syncShutdownGracefully() }
let serverChannel = try ServerBootstrap(group: group)
    .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
    .childChannelInitializer { channel in
        channel.pipeline.add(handler: ServerHandler())
    }.bind(host: "127.0.0.1", port: 0).wait()

// connect the client
class MyChannelHandler: ChannelInboundHandler {
    typealias InboundIn = NIOAny
    func channelActive(ctx: ChannelHandlerContext) {
        print("-> channelActive")
    }
}
_ = try ClientBootstrap(group: group)
    .channelInitializer { channel in
        channel.pipeline.add(handler: IdleStateHandler()).then {
            channel.pipeline.add(handler: MyChannelHandler())
        }
    }.connect(to: serverChannel.localAddress!).wait()
```

I would expect `-> channelActive` to be printed to the console, but it's not

### SwiftNIO version/commit hash

59ad7133484eec63ff5d75d407b7b84b87d7589b

### Swift & OS version (output of `swift --version && uname -a`)

n/a

normanmaurer:
@ldewailly PTAL https://github.com/apple/swift-nio/pull/615

PR #615, opened 2018-09-12
Correct propagate channelActive events to the next handler when using…

normanmaurer:
… IdleStateHandler.

Motivation:

We missed to propagate the channelActive event to the next handler in IdleStateHandler, which means handlers placed after it in the pipeline will never receive such an event.

Modifications:

- Correctly call ctx.fireChannelActive()
- Add unit test to assert we forward all inbound events.

Result:

Fixes https://github.com/apple/swift-nio/issues/614.

### DIFF
--- a/Tests/NIOTests/IdleStateHandlerTest.swift
+++ b/Tests/NIOTests/IdleStateHandlerTest.swift
@@ -98,7 +98,7 @@ class IdleStateHandlerTest : XCTestCase {
     }
     
     func testPropagateInboundEvents() {
-        class EventHandler : ChannelInboundHandler {
+        class EventHandler: ChannelInboundHandler {
             typealias InboundIn = Any
             
             var active = false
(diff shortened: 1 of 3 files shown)


Tests/NIOTests/IdleStateHandlerTest.swift
```diff
@@ -96,4 +96,86 @@ class IdleStateHandlerTest : XCTestCase {
...
+        channel.pipeline.fireChannelRegistered()
+        channel.pipeline.fireChannelActive()
+        channel.pipeline.fireChannelRead(NIOAny(""))
+        channel.pipeline.fireChannelReadComplete()
+        channel.pipeline.fireErrorCaught(ChannelError.alreadyClosed)
+        channel.pipeline.fireUserInboundEventTriggered("")
+
+        channel.pipeline.fireChannelWritabilityChanged()
+        channel.pipeline.fireChannelUnregistered()
+        channel.pipeline.fireChannelInactive()
```
Lukasa: Nit: it's probably worth putting these two into the correct order.

Tests/NIOTests/IdleStateHandlerTest.swift
```diff
@@ -96,4 +96,86 @@ class IdleStateHandlerTest : XCTestCase {
         }
         XCTAssertNoThrow(try clientChannel.closeFuture.wait())
     }
+    
+    func testPropagateInboundEvents() {
+        class EventHandler : ChannelInboundHandler {
```
Lukasa: Nit: no space before the colon.

ldewailly:
Likewise. Belated lgtm. Thanks for prompty fixing @normanmaurer

PR #616, opened 2018-09-18
CI: use Swift 4.2-RELEASE instead of CONVERGENCE

weissi:
Motivation:

Now that 4.2 has been released, there's no reason to use the CONVERGENCE
version anymore.

Modifications:

use the RELEASE instead of the CONVERGENCE version of Swift 4.2

Result:

use the latest and greatest


Lukasa:

PR #621, opened 2018-09-20
make the allocation limits for Swift 4.2 the main ones

weissi:
Motivation:

We always care most about the latest released Swift version. Therefore I
added allocation limits for Swift 4.2 and indicated that the old ones
can be removed as soon as they change.

Modifications:

- added allocation limits for Swift 4.2
- added a comment to the old ones that they don't need to be changed
  anymore now

Result:

- best support for best version
- document that Swift 4.2 is awesome, lots fewer allocations!

### DIFF
--- a/docker/docker-compose.1804.42.yaml
+++ b/docker/docker-compose.1804.42.yaml
@@ -15,9 +15,21 @@ services:
 
   integration-tests:
     image: swift-nio:18.04-4.2
+    environment:
+      - MAX_ALLOCS_ALLOWED_1000_reqs_1_conn=36750
+      - MAX_ALLOCS_ALLOWED_1_reqs_1000_conn=698050
+      - MAX_ALLOCS_ALLOWED_ping_pong_1000_reqs_1_conn=4600
+      - MAX_ALLOCS_ALLOWED_bytebuffer_lots_of_rw=2150
+      - MAX_ALLOCS_ALLOWED_future_lots_of_callbacks=99100
 
   test:
     image: swift-nio:18.04-4.2
+    environment:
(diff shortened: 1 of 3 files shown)


docker/docker-compose.1804.42.yaml
```diff
@@ -15,9 +15,21 @@ services:
 
   integration-tests:
     image: swift-nio:18.04-4.2
+    environment:
+      - MAX_ALLOCS_ALLOWED_1000_reqs_1_conn=36750
```
weissi: CC @eeckstein, @gottesmm, @airspeedswift, @milseman (random list of people CC'd that I know helped reducing these limits) the NIO team say thanks for this: Check this out:

| test | allocations in 4.1 | allocations in 4.2 |
| ----| ----| ---- |
| 1000 HTTP/1.1 requests over 1 connection | 47 allocations per request/response pair | 36 allocs for req/res pair |
| 1 HTTP/1.1 request per connection with 1000 connections | 719 allocs per req/res pair | 698 allocs |

Also CC @tanner0101 / @LoganWright and @ianpartridge / @pushkarnk who'll be happy to know we're down 20% of the allocations (from 47 to 36) to for a HTTP/1.1 request response pair, that's pretty major for 0 NIO code change. And wait for Swift 5.0 💛

normanmaurer: wow... thats really a lot!

weissi: I know right!

weissi: 🍻 for the Swift team!!

Lukasa: 🎉🎉🎉

PR #622, opened 2018-09-23
Remove useless lock.

c98:
Remove the useless lock op.

It looks like an obvious issue, here the context `DispatchGroup` / `enter` / `leave` / `wait`, it has been ensured that the `loop`'s getter/setter op can NOT be called in multi threads. So the `lock` op seems useless, i have just removed it ;)


Lukasa:

normanmaurer:
@weissi can we close this one ?

weissi:
Yes

PR #625, opened 2018-09-28
Rest between runs of tests that join multicast groups.

### LINKED ISSUE #624: testPassingInvalidSizeToSetComplexSocketOptionFails flaky
weird 'out of memory' failure in virtualised macOS

```
Test Case '-[NIOTests.SocketOptionProviderTest testPassingInvalidSizeToSetComplexSocketOptionFails]' started.
/private/tmp/task-sandbox/-7497412305172451960.V4H4jt/0~vGyWzLVgFkF3b1-I_8usZW9BljRgcBsESxOp0H8x74_B_5DQUoNVy0bnPnok2DJumZvVn60yZSXgJUzcjTpT9w==/SRCROOT/Tests/NIOTests/SocketOptionProviderTest.swift:64: error: -[NIOTests.SocketOptionProviderTest testPassingInvalidSizeToSetComplexSocketOptionFails] : failed - unexpected error setsockopt(socket:level:optionName:optionValue:optionLen:) failed: Cannot allocate memory (errno: 12) thrown
Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

CC'ing @Lukasa

Lukasa:
Motivation:

On some machines we've seen tests that rapidly join and leave multicast groups
hitting spurious test failures. These test failures manifest as hitting ENOMEM on
setsockopt(IP_ADD_MEMBERSHIP). The reason appears to be that the IGMP join/leave
messages that the kernel needs to emit for these joins can be queued up waiting for
a timer to fire, and if we move sufficiently quickly we can push too many
messages into that queue.

This can be avoided by avoiding the need to emit the IGMP join/leave messages
at all. We can do that by forcing the multicast joins onto the loopback interface,
as there's no value in the kernel emitting join/leaves on the loopback. Given that
for these tests the join is necessary only to get the kernel to have IGMP state,
there's no reason not to do this.

Modifications:

Forced all multicast joins onto the loopback address.

Result:

- No weird ENOMEM crashes in test runs.
- fixes #624

### DIFF
--- a/Tests/NIOTests/SocketOptionProviderTest.swift
+++ b/Tests/NIOTests/SocketOptionProviderTest.swift
@@ -58,15 +58,25 @@ final class SocketOptionProviderTest: XCTestCase {
         self.serverChannel = try? assertNoThrowWithValue(ServerBootstrap(group: group).bind(host: "127.0.0.1", port: 0).wait())
         self.clientChannel = try? assertNoThrowWithValue(ClientBootstrap(group: group).connect(to: serverChannel.localAddress!).wait())
 
+        // We need to join these multicast groups on the loopback interface to work around issues with rapidly joining and leaving
+        // many multicast groups. On some OSes, if we do that on a public interface, we can build up a kernel backlog of IGMP
+        // joins/leaves that may eventually lead to an ENOMEM and a spurious test failure. As joining/leaving groups on loopback
+        // interfaces does not require IGMP joins/leaves, forcing these joins onto the loopback interface saves us from this
+        // risk.
+        let v4LoopbackAddress = try! SocketAddress(ipAddress: "127.0.0.1", port: 0)
+        let v6LoopbackAddress = try! SocketAddress(ipAddress: "::1", port: 0)
+        let v4LoopbackInterface = try! System.enumerateInterfaces().filter { $0.address == v4LoopbackAddress }.first!
+
         self.ipv4DatagramChannel = try? assertNoThrowWithValue(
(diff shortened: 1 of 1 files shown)


Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -58,15 +58,25 @@ final class SocketOptionProviderTest: XCTestCase {
         self.serverChannel = try? assertNoThrowWithValue(ServerBootstrap(group: group).bind(host: "127.0.0.1", port: 0).wait())
         self.clientChannel = try? assertNoThrowWithValue(ClientBootstrap(group: group).connect(to: serverChannel.localAddress!).wait())
 
+        // We need to join these multicast groups on the loopback interface to work around issues with rapidly joining and leaving
+        // many multicast groups. On some OSes, if we do that on a public interface, we can build up a kernel backlog of IGMP
+        // joins/leaves that may eventually lead to an ENOMEM and a spurious test failure. As joining/leaving groups on loopback
+        // interfaces does not require IGMP joins/leaves, forcing these joins onto the loopback interface saves us from this
+        // risk.
+        let v4LoopbackAddress = try! SocketAddress(ipAddress: "127.0.0.1", port: 0)
+        let v6LoopbackAddress = try! SocketAddress(ipAddress: "::1", port: 0)
```
weissi: won't this blow up in non ipv6 systems?

Lukasa: Nope. `inet_pton` works just fine on non-IPv6 systems, which is all this requires.

normanmaurer:
LGTM

PR #634, opened 2018-10-18
fix potential deadlock in BlockingIOThreadPool

weissi:
Motivation:

BlockingIOThreadPool used to call out with a lock held and on top of
that on the wrong thread...

Modifications:

Make BlockingIOThreadPool call out on the supplied `queue`

Result:

fewer deadlock and surprises


weissi:
I need to fix the `unowned` thing here, think that's no longer true. Will need to capture `self` in the cancellation blocks

PR #635, opened 2018-10-18
verify that IOThreadPools get shut down

weissi:
# DO NOT MERGE BEFORE 1.10.0 GOT RELEASED

This should go into 1.11.0

Motivation:

Not shutting down IOThreadPools leads to thread leaks so we should
verify they're actually shut down.

Modifications:

add a check that we're shutting down

Result:

fewer thread leaks


Sources/NIO/BlockingIOThreadPool.swift
```diff
@@ -167,6 +167,15 @@ public final class BlockingIOThreadPool {
             }
         }
     }
+    
+    deinit {
+        switch self.state {
+        case .stopped, .shuttingDown:
+            ()
+        default:
+            assertionFailure("wrong state \(self.state)")
```
Lukasa: What's the reasoning for this being an assertion instead of a precondition?

weissi: @Lukasa same as for `BaseSocketChannel`, also it's not really a correctness issue if it's leaked I thought...

Lukasa: Fair enough.

Lukasa:
Looks like an integration test failure.

weissi:
@Lukasa intermittent failure:

```
Fatal error: Posix wrapper adds more than 20% overhead (with wrapper: 0.806146025657654, without: 0.64440906047821
```

deffo unrelated

weissi:

PR #637, opened 2018-10-21
CONNECT method should not have request body

norio-nomura:
`CONNECT` method should not have request body.

### Motivation:

[rfc7231](https://tools.ietf.org/html/rfc7231#section-4.3.6) said:
> A payload within a `CONNECT` request message has no defined semantics;
   sending a payload body on a `CONNECT` request might cause some existing
   implementations to reject the request.

### Modifications:

Change `HTTPMethod.CONNECT.hasRequestBody` to returns `.no`.

### Result:

`HTTPRequestEncoder` does not generate chunked body for `CONNECT` method.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -957,9 +957,9 @@ public enum HTTPMethod: Equatable {
     /// Whether requests with this verb may have a request body.
     public var hasRequestBody: HasBody {
         switch self {
-        case .HEAD, .DELETE, .TRACE:
+        case .HEAD, .DELETE, .CONNECT, .TRACE:
             return .no
-        case .POST, .PUT, .CONNECT, .PATCH:
+        case .POST, .PUT, .PATCH:
             return .yes
         case .GET, .OPTIONS:
             fallthrough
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks for this patch!

Sadly I wonder if perhaps we may need to fix this implementation a bit more. When RFC 7231 says “a payload has no defined semantics”, that is not the same as saying it may not be there. GET, for example, uses the same wording. 

So I think this enum is wrong for a number of methods, and many of them should go into `.unlikely` instead of `.no`.

Lukasa:
It would also be really good to add a regression test for the CONNECT change to validate that it continues to work in future.

norio-nomura:
> So I think this enum is wrong for a number of methods, and many of them should go into .unlikely instead of .no.

Changed too `.no`.

> It would also be really good to add a regression test for the CONNECT change to validate that it continues to work in future.

`HTTPRequestEncoderTests.testCONNECT()` is not enough?
If `hasRequestBody` returns `.yes`, request contains `transfer-encoding: chunked\r\n\r\n0` without writing `HTTPClientRequestPart.body` to `HTTPRequestEncoder`.

Lukasa:

weissi:

norio-nomura:
Thanks! 🙏

PR #639, opened 2018-10-30
String decoding: go through URBP directly instead of Slice

weissi:
Motivation:

String decoding can work faster when going through URBP directly instead
of a Slice of a URBP. The reason is that if the String has the utf8
represenation stored directly it'll just be a memcpy (+ validation).

Modifications:

go through URBP when decoding a String using String(decoding:as:)

Result:

faster if the String utf8 representation happens to be available


normanmaurer:
@weissi nice one!

PR #644, opened 2018-11-01
EventLoop.assertInEventLoop and preconditionInEventLoop

weissi:
Motivation:

Dispatch can only precondition that code is run from a certain queue, it
can't return if code is running on a certain queue.
To support that this introduces new EventLoop.assertInEventLoop and
EventLoop.preconditionInEventLoop methods instead of
assert(eventLoop.inEventLoop).

Modifications:

add EventLoop.assertInEventLoop and EventLoop.preconditionInEventLoop

Result:

better support for Network.framework in NIOTS

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -76,7 +76,7 @@ public final class RepeatedTask {
     }
 
     private func begin0(in delay: TimeAmount) {
-        assert(self.eventLoop.inEventLoop)
+        self.eventLoop.assertInEventLoop()
         guard let task = self.task else {
             return
         }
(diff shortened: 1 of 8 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -384,6 +388,22 @@ extension EventLoop {
...
     }
+
+    /// Checks that this call is run from the EventLoop. If this is called from within the EventLoop this function will
+    /// have no effect, if called from outside the EventLoop it will crash the process with a trap if run in debug mode.
+    /// In release mode this function never has any effect.
+    ///
+    /// - note: This is not a customization point so calls to this function can be fully optimized out in release mode.
+    @_inlineable
+    public func assertInEventLoop(file: StaticString = #file, line: UInt = #line) {
+        releaseOnly {
```
Lukasa: This condition is backwards.

Sources/NIO/Utilities.swift
```diff
@@ -23,6 +23,18 @@ internal func debugOnly(_ body: () -> Void) {
     assert({ body(); return true }())
 }
 
+@inline(__always)
+private func isRelease() -> Bool {
+    return _isDebugAssertConfiguration()
```
Lukasa: Actually isn't *this* backwards? The naming here is a bit confusing.

Lukasa: Regardless, the condition we want is `debugOnly`, not `releaseOnly`, so I don't think we need this change.

weissi: thanks @Lukasa , updated

PR #652, opened 2018-11-09
remove ugly and unnecessary ignore result (_ = ...)

weissi:
Motivation:

_ = expression() is ugly and in many cases unimportant. In fact we train
ourselves to overread it which makes special cases where you'd actually
expect a result to be used just look normal.

Modifications:

remove _ = from a lot of places. In many cases we already had a better
(and sometimes cheaper way) to not return a value but in some cases
(mostly `remove` functions) I added `@discardableResult`

Result:

NIO source code looks nicer

### DIFF
--- a/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
@@ -215,7 +215,7 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
                 // internal handler, then call the user code, and then finally when the user code is done we do
                 // our final cleanup steps, namely we replay the received data we buffered in the meantime and
                 // then remove ourselves from the pipeline.
-                _ = self.removeExtraHandlers(ctx: ctx).then {
+                self.removeExtraHandlers(ctx: ctx).then {
                     self.sendUpgradeResponse(ctx: ctx, upgradeRequest: request, responseHeaders: responseHeaders)
                 }.then {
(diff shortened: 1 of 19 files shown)


Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -236,8 +236,8 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
                     if bufferedMessages.count > 0 {
                         ctx.fireChannelReadComplete()
                     }
-                }.then {
-                    ctx.pipeline.remove(ctx: ctx)
+                }.whenComplete {
```
Lukasa: This is not quite semantically equivalent, but I think it's more accurate anyway.

Lukasa:
Re-tagged this as a server-minor as adding `@discardableResult` changes the API in a backward-compatible way.

weissi:
@Lukasa how does it change the API? Anything that was valid before is still valid, no?

Lukasa:
Ah, forgot that discarding a non-discardable result only produced a warning, not an error. You're right, this is fine: warnings don't affect semver.

weissi:
@Lukasa also, it can also _remove_ a warning from existing code not add one

Lukasa:
@weissi That's why it would have been a minor, not a major. A minor means that code written against the new code may not compile against the old, but code written against the old will compile against the new. If `@discardableResult` made *errors* go away, it'd be a minor.

PR #654, opened 2018-11-14
More `swift-nio-http2` compatibility.

AlanQuatermain:
Make the member variables of `HTTPHeader` and `HTTPHeaderIndex` visible to swift-nio-http2.

### Motivation:

When gleefully attempting to use the methods I added to `HTTPHeaders` to get a direct copy/map of the header buffers into the H2 `HPACKHeaders` structure, I discovered I had neglected to make the actual member variables available, and thus could do approximately nothing with the new API I'd created. Sigh.

### Modifications:

Attached a `public` keyword to the members of the `HTTPHeader` and `HTTPHeaderIndex` types. The note on the type itself still applies here: this is intended to help interoperability with other SwiftNIO projects, it's not really something for end users to use.

### Result:

I'll feel marginally less foolish for missing it in the first place, and I'll actually be able to do what I intended when converting from `HTTPHeaders` to `HPACKHeaders` and back again.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -372,8 +372,8 @@ public struct HTTPResponseHead: Equatable {
 /// - note: This is public to aid in the creation of supplemental HTTP libraries, e.g.
 ///         NIOHTTP2 and NIOHPACK. It is not intended for general use.
 public struct HTTPHeaderIndex {
-    let start: Int
-    let length: Int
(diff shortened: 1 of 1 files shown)


AlanQuatermain:
It's worth noting that, if this structure changes with the move to Swift 5 in SwiftNIO 2.0, that this particular change may or may not become unnecessary. I'm fine with that, if the new API affords me what I need to switch these types around properly.

Lukasa:
🙈 Whoops!

PR #660, opened 2018-11-22
Added public init for HTTPServerProtocolErrorHandler

kjessup:
Added an empty public init to HTTPServerProtocolErrorHandler.

### Motivation:

HTTPServerProtocolErrorHandler is a class declared as public, yet it had no initializers, therefore it could not be instantiated by end-user code.

### Modifications:

An empty init function was added `public init() {}`

### Result:

The functionality of the object is unchanged, but end user code can now instantiate HTTPServerProtocolErrorHandler and add it to pipelines.


Lukasa:

PR #671, opened 2018-11-27
optional type parameter for newPromise()

weissi:
Motivation:

When creating new promises I always find it very frustrating to type the
`: EventLoopPromise<Type>` type annotation but it's necessary for the
compiler to know type the promise will be fulfilled with.

Modifications:

allow an optional `for: SomeType.self` parameter for `newPromise` as

    let p = eventLoop.newPromise(for: Int.self)

is much easier to type than

    let p: EventLoopPromise<Int> = eventLoop.newPromise()

Result:

easier to write code

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -318,7 +318,7 @@ extension EventLoop {
     }
 
     /// Creates and returns a new `EventLoopPromise` that will be notified using this `EventLoop` as execution `Thread`.
-    public func newPromise<T>(file: StaticString = #file, line: UInt = #line) -> EventLoopPromise<T> {
+    public func newPromise<T>(for type: T.Type = T.self, file: StaticString = #file, line: UInt = #line) -> EventLoopPromise<T> {
         return EventLoopPromise<T>(eventLoop: self, file: file, line: line)
     }
 
@@ -757,7 +757,7 @@ internal final class SelectableEventLoop: EventLoop {
         if self.inEventLoop {
             return closeGently0()
         } else {
-            let p: EventLoopPromise<Void> = self.newPromise()
+            let p = self.newPromise(for: Void.self)
             self.execute {
                 closeGently0().cascade(promise: p)
             }
--- a/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/template/Sources/SwiftBootstrap/SwiftMain.swift
@@ -347,7 +347,7 @@ public func swiftMain() -> Int {
         struct MyError: Error { }
         @inline(never)
         func doThenAndFriends(loop: EventLoop) {
-            let p: EventLoopPromise<Int> = loop.newPromise()
+            let p = loop.newPromise(for: Int.self)
             let f = p.futureResult.then { (r: Int) -> EventLoopFuture<Int> in 
                 // This call allocates a new Future, and
                 // so does then(), so this is two Futures.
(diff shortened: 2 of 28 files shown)


helje5:
I know that some people in the server space do this, but I always found those `type.self` things particularly ugly and AFAIK it is not done anywhere else in Swift or Foundation itself. I think the expected API is more like:

    let p = EventLoopPromise<Int>(in: eventLoop)

Which, funny enough, is what `newPromise` is doing. (The `newPromise` thing looks like one of the Java'ish factory patterns you find a lot in NIO.)

Lukasa:
> I know that some people in the server space do this, but I always found those type.self things particularly ugly and AFAIK it is not done anywhere else in Swift or Foundation itself.

Sure it is. Codable uses it heavily, both in the API ([`Decoder.container(keyedBy:)`](https://developer.apple.com/documentation/swift/decoder/2892621-container)) and in the implementation (check out all these [`decode`](https://github.com/apple/swift/blob/806cf57687bb27f15d7ce33426c72d063375a32b/stdlib/public/SDK/Foundation/JSONEncoder.swift#L1359) functions in `JSONEncoder.swift`). It's used in `String` ([`String.decoding(codeUnits:as:)`](https://github.com/apple/swift/blob/806cf57687bb27f15d7ce33426c72d063375a32b/stdlib/public/core/String.swift#L400)), it's used in the unsafe pointer types (see [`UnsafeRawBufferPointer.load(fromByteOffset:as:)`](https://github.com/apple/swift/blob/806cf57687bb27f15d7ce33426c72d063375a32b/stdlib/public/core/UnsafeRawBufferPointer.swift.gyb#L327) and [`UnsafeRawBufferPointer.storeBytes(of:toByteOffset:as:)`](https://github.com/apple/swift/blob/806cf57687bb27f15d7ce33426c72d063375a32b/stdlib/public/core/UnsafeRawBufferPointer.swift.gyb#L361) as examples).

And in fact the last time we did something like this (in 94c21b604375baac0d4f9c7eafde4631a1f1370f)  it was at the suggestion of some members of the standard library team.

TL;DR: the prior art for `.Type` here is pretty good.

The concern about the factory method is reasonable, but as this is intended as a semver minor we can't fix it here. I'm +1 on the idea of changing to a construct for SwiftNIO 2.0 though.

Sources/NIO/EventLoop.swift
```diff
@@ -318,7 +318,7 @@ extension EventLoop {
     }
 
     /// Creates and returns a new `EventLoopPromise` that will be notified using this `EventLoop` as execution `Thread`.
-    public func newPromise<T>(file: StaticString = #file, line: UInt = #line) -> EventLoopPromise<T> {
+    public func newPromise<T>(type: T.Type = T.self, file: StaticString = #file, line: UInt = #line) -> EventLoopPromise<T> {
```
Lukasa: I don't like the name `type` here. It sits weirdly with me: it's not the type of the promise, it's the type of what it contains. Not sure what the better name choice is though: `resultType`, maybe? Or `futureResult`? Or `producing`?

weissi: @Lukasa `makePromise<T>(for type: T.Type = T.self)` ? that reads okay: 'make promise for integer'.

weissi: well, in 1.0 obviously `newPromise`

Lukasa: Works for me.

weissi:
@Lukasa / @normanmaurer I'm open to opening up `EventLoopPromise`'s `init` and making it public but I'm against removing the factory from the event loop. I think we can have both.

Lukasa:
I'm happy enough with that as a compromise position.

Sources/NIO/EventLoop.swift
weissi: @Lukasa updated

ianpartridge: Would `makePromise<T>(of type: T.Type = T.self)` be better?

ktoso: Seems I'm also late to the party here, and wanted to suggest `of` as well when I was reading the PR. "a promise of string (in the future)" somehow reads better than "a promise for string (in the future)". Your call tho

weissi: we can still change it, it's not released. Let's page stdlib folk for API guidance, CC @airspeedswift / @lorentey

weissi: ok, we now landed on `newPromise(of: T.Type = T.self)` because 'making a promise of a type' is more grammatical than 'making a promise for a type' -->  #672

PR #675, opened 2018-12-04
ByteToMessageDecoder 2

weissi:
Motivation:

`ByteToMessageDecoder` is ~~almost~~(?) impossible to use correctly with reentrancy. We can't access the bytes when re-entered (exclusivity) so we stash them away, ie. old, already processed bytes are still reachable when being re-entered, ...
Also, it's a bad API as the user is expected to only overwrite some of the protocol requirements but not others. Oh, and the default implementations (`decodeLast`) were also broken.

Modifications:

- rewrite the whole thing, re-entrancy-safe
- `ByteToMessageDecoder` is no longer a `ChannelInboundHandler`, it needs to be wrapped in a `ByteToMessageHandler`
- way more tests

Result:

sanity

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -45,9 +45,8 @@ public enum DecodingState {
 ///    Please note, however, that the numerical value of the `readerIndex` itself is not preserved, and may not be the same from one call to the next. Please do not rely on this numerical value: if you need
 ///    to recall where a byte is relative to the `readerIndex`, use an offset rather than an absolute value.
 /// 2. Mutating the bytes in the buffer will cause undefined behaviour and likely crash your program
-public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == ByteBuffer {
-    /// The cumulationBuffer which will be used to buffer any data.
-    var cumulationBuffer: ByteBuffer? { get set }
+public protocol ByteToMessageDecoder {
+    associatedtype InboundOut
 
     /// Decode from a `ByteBuffer`. This method will be called till either the input
     /// `ByteBuffer` has nothing to read left or `DecodingState.needMoreData` is returned.
@@ -57,7 +56,7 @@ public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == B
     ///     - buffer: The `ByteBuffer` from which we decode.
     /// - returns: `DecodingState.continue` if we should continue calling this method or `DecodingState.needMoreData` if it should be called
     //             again once more data is present in the `ByteBuffer`.
-    func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState
+    mutating func decode(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState
 
     /// This method is called once, when the `ChannelHandlerContext` goes inactive (i.e. when `channelInactive` is fired)
     ///
@@ -66,19 +65,19 @@ public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == B
     ///     - buffer: The `ByteBuffer` from which we decode.
     /// - returns: `DecodingState.continue` if we should continue calling this method or `DecodingState.needMoreData` if it should be called
     //             again when more data is present in the `ByteBuffer`.
-    func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws  -> DecodingState
+    mutating func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws  -> DecodingState
 
     /// Called once this `ByteToMessageDecoder` is removed from the `ChannelPipeline`.
     ///
     /// - parameters:
     ///     - ctx: The `ChannelHandlerContext` which this `ByteToMessageDecoder` belongs to.
-    func decoderRemoved(ctx: ChannelHandlerContext)
+    mutating func decoderRemoved(ctx: ChannelHandlerContext)
 
     /// Called when this `ByteToMessageDecoder` is added to the `ChannelPipeline`.
     ///
     /// - parameters:
     ///     - ctx: The `ChannelHandlerContext` which this `ByteToMessageDecoder` belongs to.
-    func decoderAdded(ctx: ChannelHandlerContext)
+    mutating func decoderAdded(ctx: ChannelHandlerContext)
 
     /// Determine if the read bytes in the given `ByteBuffer` should be reclaimed and their associated memory freed.
     /// Be aware that reclaiming memory may involve memory copies and so is not free.
@@ -86,7 +85,36 @@ public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == B
     /// - parameters:
     ///     - buffer: The `ByteBuffer` to check
     /// - return: `true` if memory should be reclaimed, `false` otherwise.
-    func shouldReclaimBytes(buffer: ByteBuffer) -> Bool
+    mutating func shouldReclaimBytes(buffer: ByteBuffer) -> Bool
+}
+
+extension ByteToMessageDecoder {
+    public mutating func decoderRemoved(ctx: ChannelHandlerContext) {
+    }
+
+    public mutating func decoderAdded(ctx: ChannelHandlerContext) {
+    }
+
+    /// Default implementation to detect once bytes should be reclaimed.
+    public func shouldReclaimBytes(buffer: ByteBuffer) -> Bool {
+        // We want to reclaim in the following cases:
+        //
+        // 1. If there is more than 2kB of memory to reclaim
+        // 2. If the buffer is more than 50% reclaimable memory and is at least
+        //    1kB in size.
+        if buffer.readerIndex > 2048 {
+            return true
+        }
+        return buffer.capacity > 1024 && (buffer.capacity - buffer.readerIndex) >= buffer.readerIndex
+    }
+
+    public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        return .needMoreData
+    }
+
+    public func wrapInboundOut(_ value: InboundOut) -> NIOAny {
+        return NIOAny(value)
+    }
 }
 
 private extension ChannelHandlerContext {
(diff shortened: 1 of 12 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -458,7 +458,7 @@ public final class ChannelPipeline: ChannelInvoker {
     /// - parameters:
     ///     - handlerType: The type of the handler to search for.
     /// - returns: the `EventLoopFuture` which will be notified once the the operation completes.
-    public func context<T>(handlerType: T.Type) -> EventLoopFuture<ChannelHandlerContext> {
+    public func context<T: ChannelHandler>(handlerType: T.Type) -> EventLoopFuture<ChannelHandlerContext> {
```
Lukasa: Lol whoops.

Sources/NIO/Codec.swift
```diff
@@ -89,6 +88,32 @@ public protocol ByteToMessageDecoder: ChannelInboundHandler where InboundIn == B
...
+    }
+
+    public func decoderAdded(ctx: ChannelHandlerContext) {
+    }
+
+    public func shouldReclaimBytes(buffer: ByteBuffer) -> Bool {
+        return true
+    }
+
+    public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
```
Lukasa: I remain unsure about whether we should have a default implementation of this.

Lukasa: As a separate note, I will say that if we *do* want to have this we can get the different behaviours here by doing this:

```swift
extension ByteToMessageDecoder {
    public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
        return .needMoreData
    }
}

extension ByteToMessageDecoder where InboundOut == ByteBuffer {
    public func decodeLast(ctx: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
        ctx.fireChannelRead(NIOAny(buffer))
        return .needMoreData
    }
}
```

Sources/NIO/Codec.swift
```diff
@@ -101,7 +126,31 @@ private extension ChannelHandlerContext {
     }
 }
 
-extension ByteToMessageDecoder {
+public class ByteToMessageHandler<Decoder: ByteToMessageDecoder>: ChannelInboundHandler {
+    public typealias InboundIn = ByteBuffer
+    public typealias InboundOut = Decoder.InboundOut
+
+    internal private(set) var decoder: Decoder
```
Lukasa: Is there any reason for the decoder to be `internal`?

Sources/NIO/Codec.swift
```diff
@@ -101,7 +126,31 @@ private extension ChannelHandlerContext {
...
+
+    internal var cumulationBuffer: ByteBuffer? = nil
```
Lukasa: Why is this `internal`?

Tests/NIOWebSocketTests/EndToEndTests.swift
```diff
@@ -397,7 +397,7 @@ class EndToEndTests: XCTestCase {
                          expectedResponseLine: "HTTP/1.1 101 Switching Protocols",
                          expectedResponseHeaders: ["Upgrade: websocket", "Sec-WebSocket-Accept: OfS0wDaT5NoxF2gqm7Zj2YtetzM=", "Connection: upgrade"])
 
-        let decoder = (try server.pipeline.context(handlerType: WebSocketFrameDecoder.self).wait()).handler as! WebSocketFrameDecoder
+        let decoder = ((try server.pipeline.context(handlerType: ByteToMessageHandler<WebSocketFrameDecoder>.self).wait()).handler as! ByteToMessageHandler<WebSocketFrameDecoder>).decoder
```
Lukasa: Hrm. This is a bad test, we should fix it. It should test this by attempting to *violate* the max frame size rule and seeing that that fails.

Lukasa: Appreciate that we probably shouldn't fix it as part of this PR though.

Sources/NIO/ChannelPipeline.swift
weissi: yep :)

Sources/NIO/Codec.swift
weissi: we shouldn't I think. This is more mean to be the smallest diff to passing the tests. I think in decodeLast we also should make the user do something with the left-overs. Needs some design

Lukasa: 👍

weissi: yeah we could but the whole design that we throw it under the bus unless it's ByteBuffer isn't good I think. We should make the user do smth sensible

Sources/NIO/Codec.swift
weissi: tests... Also to be thought through

Sources/NIO/Codec.swift
weissi: tests again

tomerd:

weissi:

Sources/NIO/Codec.swift
Lukasa: Deferring this for follow-up work.

Lukasa: Deferring this for follow-up work.

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,266 @@ private extension ChannelHandlerContext {
...
+    mutating func finishProcessing(remainder buffer: ByteBuffer) -> Void {
+        assert(self.state == .processingInProgress)
+        self.state = .ready
+        if buffer.readableBytes > 0 {
+            self.buffers.prepend(buffer)
         } else {
-            self.cumulationBuffer = self.unwrapInboundIn(data)
+            var buffer = buffer
+            buffer.clear()
+            mergeBuffers(self.buffers, into: &buffer)
```
Lukasa: Should there ever be any buffers in `self.buffers` at the time of this call?

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,266 @@ private extension ChannelHandlerContext {
...
+            return .cannotProcessReentrantly
+        case .nothingAvailable:
+            return .didProcess(.needMoreData)
+        case .available(var buffer):
+            var decoder: Decoder? = nil
+            swap(&decoder, &self.decoder)
+            defer {
+                swap(&decoder, &self.decoder)
+                if buffer.readableBytes > 0 {
+                    if self.decoder!.shouldReclaimBytes(buffer: buffer) {
```
Lukasa: Need to comment this force-unwrap to justify it.

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,266 @@ private extension ChannelHandlerContext {
...
+                        buffer.discardReadBytes()
+                    }
+                }
+                self.buffer.finishProcessing(remainder: buffer)
+            }
```
Lukasa: This `defer` block is very, very large. Are we sure it's clearer than just saving off the return value from `body`?

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,266 @@ private extension ChannelHandlerContext {
...
+                self.state = .done
+            case .cannotProcessReentrantly:
+                self.state = .leftoversNeedProcessing
+            }
+        }
     }
 
-    /// Do nothing by default.
-    public func decoderRemoved(ctx: ChannelHandlerContext) {
+    private func decodeLoop(ctx: ChannelHandlerContext, extraData: ByteBuffer? = nil, decodeMode: DecodeMode) throws -> B2MDBuffer.BufferProcessingResult {
```
Lukasa: I don't think this `extraData` parameter makes anything particularly clearer. Any reason not to just `append` it to `self.buffer` in `channelRead`?

Sources/NIO/Codec.swift
weissi: @Lukasa yes, if we re-entrantly got some more. So for example:

    startProcessing
        decode
            write
                pipeline.fireChannelRead (because draining the socket buffer on error
    finishProcessing

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,266 @@ private extension ChannelHandlerContext {
...
+            }
+            return firstBuffer
         }
+    }
 
-        // Check if we should reclaim some bytes and if so do it.
-        if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) {
-            self.cumulationBuffer!.discardReadBytes()
+    private func mergeBuffers(_ buffers: CircularBuffer<ByteBuffer>, into target: inout ByteBuffer) {
+        guard self.buffers.count > 0 else {
```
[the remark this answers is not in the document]
weissi: oops, good catch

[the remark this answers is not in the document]
weissi: and great idea, done

Sources/NIO/Codec.swift
weissi: yes, thanks done

Sources/NIO/Codec.swift
weissi: @Lukasa `body` throws so taking this out of the `defer` would it super easy to introduce bugs...

Sources/NIO/Codec.swift
weissi: that's a good idea, thanks!

weissi:
@Lukasa thanks, updated

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,265 @@ private extension ChannelHandlerContext {
...
+    /// Start processing some bytes if possible, if we receive a returned buffer (through `.available(ByteBuffer)`)
+    /// we _must_ indicate the processing has finished by calling `finishProcessing`.
+    mutating func startProcessing() -> BufferAvailability {
+        switch self.state {
+        case .processingInProgress:
+            return .bufferAlreadyBeingProcessed
+        case .ready where self.buffers.count > 0:
+            var buffer = self.buffers.removeFirst()
+            buffer.writeBuffers(self.buffers)
+            self.buffers.removeAll(keepingCapacity: true)
```
normanmaurer: should we only keep the capacity up to a max value to minimise memory overhead ?

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,265 @@ private extension ChannelHandlerContext {
...
+        assert(self.state == .processingInProgress)
+        self.state = .ready
+        if buffer.readableBytes > 0 {
+            self.buffers.prepend(buffer)
         } else {
-            self.cumulationBuffer = self.unwrapInboundIn(data)
+            var buffer = buffer
+            buffer.clear()
+            buffer.writeBuffers(self.buffers)
+            self.buffers.removeAll(keepingCapacity: true)
```
normanmaurer: same comment as above.

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,265 @@ private extension ChannelHandlerContext {
...
         }
-
-        // Check if we should reclaim some bytes and if so do it.
-        if self.shouldReclaimBytes(buffer: self.cumulationBuffer!) {
-            self.cumulationBuffer!.discardReadBytes()
+        var allReadableBytes: Int = 0
+        for buffer in buffers {
+            allReadableBytes += buffer.readableBytes
+        }
+        self.reserveCapacity(self.writerIndex + allReadableBytes)
```
normanmaurer: nit: you could use `var allReadableBytes = self.writerIndex` before to eliminate on `+` operation ;)

Sources/NIO/Codec.swift
```diff
@@ -101,115 +129,265 @@ private extension ChannelHandlerContext {
...
-        if self.cumulationBuffer != nil {
-            ctx.withThrowingToFireErrorAndClose {
-                try self.decodeLoop(ctx: ctx, decodeFunc: self.decodeLast)
+private extension B2MDBuffer {
+    func _testOnlyOneBuffer() -> ByteBuffer? {
+        if buffers.count == 0 {
+            return nil
+        } else if buffers.count == 1 {
+            return self.buffers.first
+        } else {
```
normanmaurer: nit: consider using a switch statement.

PR #683, opened 2018-12-07
fix CircularBuffer.replaceSubrange expands

weissi:
Motivation:

CircularBuffer.replaceSubrange was crazy eager to double the capacity
which is bad.

Modifications:

Made CircularBuffer.replaceSubrange only double if necessary.

Result:

less memory usage


Lukasa:
EBROKENCI

weissi:

PR #690, opened 2018-12-10
Reunify errors into a single enum.

### LINKED ISSUE #620: Bring MulticastError and ChannelLifecycleError into ChannelError.
As `ChannelError` is an `enum` we haven't been able to expand it, but a number of related errors have popped up over time. In 2.0 we should consolidate them. The known errors are:

- `ChannelLifecycleError`
- `MulticastError`

Lukasa:
Motivation:

If we're going to have errors in enumerations, we should aim to have as
few as possible. We wanted to throw away these two, so let's do it.

Modifications:

- Removed `ChannelLifecycleError`, moved case to `ChannelError`.
- Removed `MulticastError`, moved cases to `ChannelError`.
- Removed manual `Equatable` conformance, used synthesised one.
- Pulled `connectFailed` out of `ChannelError`, as it made it impossible to
    write a correct implementation of `Equatable`.
- Conformed `NIOConnectionError` to `Error` instead.

Result:

Better error handling.

Resolves #620

### DIFF
--- a/docs/public-api-changes-NIO1-to-NIO2.md
+++ b/docs/public-api-changes-NIO1-to-NIO2.md
@@ -16,3 +16,6 @@
 - changed `HTTPVersion`'s `major` and `minor` properties to `Int` (from `UInt16`)
 - renamed the generic parameter name to `Bytes` where we're talking about a
   generic collection of bytes
+- Moved error `ChannelLifecycleError.inappropriateOperationForState` to `ChannelError.inappropriateOperationForState`.
+- Moved all errors in `MulticastError` enum into `ChannelError`.
+- Removed `ChannelError.connectFailed`. All errors that triggered this now throw `NIOConnectError` directly.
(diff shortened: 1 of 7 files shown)


Lukasa:
Updated, thanks. :smile:

docs/public-api-changes-NIO1-to-NIO2.md
```diff
@@ -10,3 +10,6 @@
 - `EventLoop.makePromise`/`makeSucceededFuture`/`makeFailedFuture` instead of `new*`
 - `SocketAddress.makeAddressResolvingHost(:port:)` instead of
   `SocketAddress.newAddressResolving(host:port:)`
+  - Moved error `ChannelLifecycleError.inappropriateOperationForState` to `ChannelError.inappropriateOperationForState`.
```
normanmaurer: @Lukasa I think all of them needs to have two whitespaces removed so these are not sub-lists.

Lukasa: That'll teach me to write my markdown in Xcode.

Lukasa: @normanmaurer Resolved, please re-review.

PR #691, opened 2018-12-10
changed port/HTTP version parts to Ints

weissi:
Motivation:

- `Int` is the currency type for integral things and we already sort of
  changed the the type for ports but not everywhere.
- generic parameters that aren't just 'any type' shouldn't be just `T`
  or `S` or something

Modifications:

- make more use of `Int`
- rename generic parameters to something more descriptive like `Bytes`

Result:

easier to read/write NIO code

### DIFF
--- a/Sources/NIO/SocketAddresses.swift
+++ b/Sources/NIO/SocketAddresses.swift
@@ -126,12 +126,14 @@ public enum SocketAddress: CustomStringConvertible {
     }
 
     /// Get the port associated with the address, if defined.
-    public var port: UInt16? {
+    public var port: Int? {
         switch self {
         case .v4(let addr):
-            return UInt16(bigEndian: addr.address.sin_port)
+            // looks odd but we need to first convert the endianness as `in_port_t` and then make the result an `Int`.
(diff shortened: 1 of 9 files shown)


Sources/NIO/SocketAddresses.swift
```diff
@@ -126,12 +126,12 @@ public enum SocketAddress: CustomStringConvertible {
     }
 
     /// Get the port associated with the address, if defined.
-    public var port: UInt16? {
+    public var port: Int? {
         switch self {
         case .v4(let addr):
-            return UInt16(bigEndian: addr.address.sin_port)
+            return Int(UInt16(bigEndian: addr.address.sin_port))
```
Lukasa: Why is this round-tripping through `UInt16`?

weissi: @Lukasa step 1) change the endianness (only available within the same type), step 2) change the type

Lukasa: Can you add a comment to that effect please? People will definitely forget.

weissi: yes, done

PR #694, opened 2018-12-11
Make WebSocketOpcode a struct.

### LINKED ISSUE #617: WebSocketOpcode should be improved for 2.0
`WebSocketOpcode` is defined like this:

```swift
public enum WebSocketOpcode {
    case continuation
    case text
    case binary
    case unknownNonControl(UInt8)
    case connectionClose
    case ping
    case pong
    case unknownControl(UInt8)

    /// Create an opcode from the encoded representation.
    ///
    /// - parameters
    ///     - encoded: The encoded representation of the opcode as an 8-bit integer.
    ///          Must be no more than 4 bits large.
    public init?(encodedWebSocketOpcode encoded: UInt8) {
        guard encoded < 0x10 else {
            return nil
        }
       [...]
    }
   [...]
}
```

which looks like there's no way to create an illegal `WebSocketOpcode` but in reality it's super easy: `WebSocketOpcode.unknownNonControl(0xff)` is illegal, so is `WebSocketOpcode.unknownNonControl(0)` which looks more legal but should be `WebSocketOpcode.continuation`.

We should make the illegal states unrepresentable. Instead of `UInt8` we could use

```swift
public struct UnknownOpcode {
    public var private(set) opcode: UInt8

    public init?(integerLiteral value: UInt8) {
        guard value < 0x10 && value != 0 && value != 1 && ... else {
            return nil
        }
        self.opcode = value
    }
}
```

and rephrase the `enum` as

```swift
public enum WebSocketOpcode {
    case continuation
    case text
    case binary
    case unknownNonControl(UnknownOpcode)
    case connectionClose
    case ping
    case pong
    case unknownControl(UnknownOpcode)
}
```

An alternative would be to give up exhaustive switching and just use a `public struct WebSocketOpcode` with a bunch of `static let` and `static func` to create the well-known cases. Maybe better?

that way we can trust that any `WebSocketOpcode` is actually valid and we don't need to duplicate that check in many places.

Lukasa:
Motivation:

WebSocketOpcode was made an enumeration to make it possible to cleanly
switch over the values in switch statements. However, this unfortunately
made it possible to construct contradictory values of WebSocketOpcode,
such as .unknownNonControl(0x1) (which should be spelled .text), or
.unknownControl(0xFF) (which is simply invalid).

This patch removes the ability to construct invalid values of
WebSocketOpcode by turning it into a struct and using static lets for known
values. While we're here it cleans some other stuff up.

Modifications:

- Made WebSocketOpcode a struct
- Added several static lets for easy access and to reduce code churn
- Used synthesised Equatable conformance
- Added synthesised Hashable conformance
- Added CustomStringConvertible conformance to better represent known values
- Added CaseIterable conformance to provide entire range of valid values.

Result:

WebSocketOpcode will be a better-behaved type with more guarantees and
fewer places to trap and explode.

Resolves #617.

### DIFF
--- a/Sources/NIOWebSocket/WebSocketOpcode.swift
+++ b/Sources/NIOWebSocket/WebSocketOpcode.swift
@@ -13,15 +13,15 @@
 //===----------------------------------------------------------------------===//
 
 /// An operation code for a websocket frame.
-public enum WebSocketOpcode {
-    case continuation
-    case text
-    case binary
-    case unknownNonControl(UInt8)
-    case connectionClose
-    case ping
-    case pong
-    case unknownControl(UInt8)
+public struct WebSocketOpcode {
+    fileprivate let networkRepresentation: UInt8
+
+    public static let continuation = WebSocketOpcode(rawValue: 0x0)
+    public static let text = WebSocketOpcode(rawValue: 0x1)
+    public static let binary = WebSocketOpcode(rawValue: 0x2)
+    public static let connectionClose = WebSocketOpcode(rawValue: 0x8)
+    public static let ping = WebSocketOpcode(rawValue: 0x9)
+    public static let pong = WebSocketOpcode(rawValue: 0xA)
 
     /// Create an opcode from the encoded representation.
     ///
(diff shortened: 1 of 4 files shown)


Sources/NIOWebSocket/WebSocketOpcode.swift
```diff
@@ -33,56 +33,50 @@ public enum WebSocketOpcode {
...
-             (.unknownNonControl(let i), .unknownNonControl(let j)):
-            return i == j
-        default:
-            return false
+extension WebSocketOpcode: Equatable { }
+
+extension WebSocketOpcode: Hashable { }
+
+extension WebSocketOpcode: CaseIterable {
+    public static var allCases = (0..<0x10).map { WebSocketOpcode.init(rawValue: $0) }
```
weissi: nit: we don't need the `.init` here

Lukasa: Fixed.

PR #697, opened 2018-12-11
Enable more specialisations for Atomics

ogres:
Enabled UInt and Int64 specialisations for Atomics as they were disabled due to Swift 4.0 not allowing more than three specialisations.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/atomics.swift
+++ b/Sources/NIOConcurrencyHelpers/atomics.swift
@@ -34,16 +34,11 @@ import CNIOAtomics
 /// communicate or cooperate across multiple threads. In the case of
 /// SwiftNIO this usually involves communicating across multiple event loops.
 public struct UnsafeEmbeddedAtomic<T: AtomicPrimitive> {
-    private let value: OpaquePointer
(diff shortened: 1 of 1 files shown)


Lukasa:
Rather than add the specialisations I think we should remove the ones we have and use `@inlinable` instead. @weissi?

ogres:
In that case, we would also need to change the access level of `UnsafeEmbeddedAtomic.value` and `Atomic.embedded` to at least internal and mark them as `@usableFromInline`. Currently they are private and cannot be referenced from `@inlinable` function

weissi:
Agreed with Cory, we should remove all specialisations and make more stuff inlinable. Specialisations are only useful of you want to have an ABI which we don’t.

normanmaurer:
+1 to what @Lukasa and @weissi said.

Lukasa:
Closing in favour of #698.

ogres:
Unfortunately this branch got orphaned and could not rebase all the changes into one commit, I've opened another PR for this reason - https://github.com/apple/swift-nio/pull/698

Sorry for inconvenience!

PR #698, opened 2018-12-12
@inlinable atomics

ogres:
Followup from https://github.com/apple/swift-nio/pull/697

Removed specialisations from Atomics functions and added @inlinable attribute instead.
Access level of properties used from @inlinable functions are now internal and they are marked as @usableFromInline.


Lukasa:

PR #704, opened 2018-12-14
update tools-version to 5.0 & fix latest warnings

weissi:
Motivation:

The latest Swift 5.0s bring us new goodies (hence tools version) and
also more warnings, hence the fixes for them.

Modifications:

- set tools-version to 5.0
- fix warnings

Result:

happier developers

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -238,10 +238,10 @@ public struct ByteBuffer {
     public typealias _Index = UInt32
     public typealias _Capacity = UInt32
 
-    @usableFromInline private(set) var _storage: _Storage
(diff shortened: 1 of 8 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -238,10 +238,10 @@ public struct ByteBuffer {
     public typealias _Index = UInt32
     public typealias _Capacity = UInt32
 
-    @usableFromInline private(set) var _storage: _Storage
-    @usableFromInline private(set) var _readerIndex: _Index = 0
-    @usableFromInline private(set) var _writerIndex: _Index = 0
-    @usableFromInline private(set) var _slice: Slice
```
Lukasa: Why did all of these go away?

PR #707, opened 2018-12-14
use String.UTF8View.withContiguousStorageIfAvailable

weissi:
Motivation:

The new shiny Swift 5 gives us fast access to the UTF8 bytes (if
available, which mostly means not bridged from Cocoa), we should use
that.

Modifications:

use UTF8View.withContiguousStorageIfAvailable

Result:

faster String access

### DIFF
--- a/Sources/NIO/ByteBuffer-aux.swift
+++ b/Sources/NIO/ByteBuffer-aux.swift
@@ -105,7 +105,21 @@ extension ByteBuffer {
     /// - returns: The number of bytes written.
     @discardableResult
     public mutating func set(string: String, at index: Int) -> Int {
-        return self.set(bytes: string.utf8, at: index)
+        if let written = string.utf8.withContiguousStorageIfAvailable({ utf8Bytes in
+            self.set(bytes: utf8Bytes, at: index)
+        }) {
+            // best case, directly available
+            return written
+        } else {
+            // second best case, let's try to force the string to be native
+            if let written = (string + "").utf8.withContiguousStorageIfAvailable({ utf8Bytes in
+                self.set(bytes: utf8Bytes, at: index)
+            }) {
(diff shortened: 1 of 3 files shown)


weissi:
(needs major bump because it requires Swift 5.0 from 13th Dec or better)

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -107,8 +107,22 @@ extension ByteBuffer {
...
-        return self.set(bytes: string.utf8, at: index)
+    public mutating func set(string: String, at index: Int) -> Int {
+        if let written = string.utf8.withContiguousStorageIfAvailable({ utf8Bytes in
+            self.set(bytes: utf8Bytes, at: index)
+        }) {
+            // best case, directly available
+            return written
+        } else {
+            // second best case, let's try to force the string to be native
+            if let written = (string + "").utf8.withContiguousStorageIfAvailable({ utf8Bytes in
```
normanmaurer: so the `+ ""` is forcing it ?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1601,6 +1601,17 @@ class ByteBufferTest: XCTestCase {
 
         XCTAssertEqual(0, buf.readableBytes)
     }
+
+    func testPossiblyLazilyBridgedString() {
```
normanmaurer: Does the test verify all three different code-paths that we have now ?

weissi: @normanmaurer on correctly functioning toolchains we can't hit all three. But on Mac we hit the first 2 and on Linux we'll also hit the last one (at the moment but that'll be fixed).

PR #708, opened 2018-12-14
[probably close in favour of #709] prefix ContiguousCollection with NIO

weissi:
Motivation:

The standard library will likely get a ContiguousCollection protocol
(https://forums.swift.org/t/contiguous-collection-protocols/17875) so we
should prefix ours.
Also we made a public Data extension which isn't necessary.

Modifications:

- rename ContiguousCollection to NIOContiguousCollection (to be removed
  when the stdlib gets its ContiguousCollection)
- remove the public extension on Data

Result:

- more compatibility
- https://bugs.swift.org/browse/SR-9505

### DIFF
--- a/Sources/NIO/NIOContiguousCollection.swift
+++ b/Sources/NIO/NIOContiguousCollection.swift
@@ -13,9 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 /// A `Collection` that is contiguously laid out in memory and can therefore be duplicated using `memcpy`.
-public protocol ContiguousCollection: Collection {
+public protocol NIOContiguousCollection: Collection {
     @inlinable
-    func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
+    func withUnsafeBytesNIO<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
 }
 
 extension StaticString: Collection {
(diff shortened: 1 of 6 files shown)


itaiferber:
Thanks for the quick fix here!
(Also, I believe the stdlib is no longer getting `ContiguousCollection` — instead, `Sequence` got `withContiguousStorageIfAvailable`)

Sources/NIO/NIOContiguousCollection.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// A `Collection` that is contiguously laid out in memory and can therefore be duplicated using `memcpy`.
-public protocol ContiguousCollection: Collection {
+public protocol NIOContiguousCollection: Collection {
     @inlinable
     func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
```
Lukasa: We should namespace this function too.

weissi: yes, thanks now done

weissi: @Lukasa I think we should just close this and take #709 . @itaiferber's suggestion is much better, we don't need `ContiguousCollection` anymore at all...

itaiferber: Also, in case it helps, Foundation now offers the `ContiguousBytes` protocol if that also suits your needs.
(I don't have a particular stake in this; either way, we just need to ensure that SR-9505 is resolved.)

weissi:
Closing for #709 thanks @itaiferber

PR #709, opened 2018-12-14
remove ContiguousCollection

weissi:
Motivation:

ContiguousCollection is no longer necessary because Sequence gained
withContiguousStorageIfAvailable.

Modifications:

Removed ContiguousCollection and switch to withContiguousStorageIfAvailable.

Result:

less code


Sources/NIO/ContiguousCollection.swift
```diff
@@ -12,12 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A `Collection` that is contiguously laid out in memory and can therefore be duplicated using `memcpy`.
-public protocol ContiguousCollection: Collection {
-    @inlinable
-    func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
-}
-
 extension StaticString: Collection {
```
Lukasa: Do we need the StaticString conformance here?

weissi: good catch!

PR #710, opened 2018-12-14
re-enable allocation test for Swift 5 (fixes #685)

### LINKED ISSUE #685: reenable allocation test for Swift 5
we're hitting test_01_allocation_counts_for_http1, therefore I disabled the alloc counter test temporarily :(

weissi:
Motivation:

We had to disable the allocation test because of a SwiftPM bug which is
now fixed.

Modifications:

re-enable the allocation tests

Result:

more tests == better


Lukasa:
Looks like we have an allocation issue. Mind tracking down whether the increase in allocations is expected or not?

weissi:
ok, much better now. Due to [SR-9518](https://bugs.swift.org/browse/SR-9518) fixed (thanks @eeckstein) there are now no more memory leaks. However we still seem to be 2 allocations worse per HTTP1 req/res pair but AFAIK only on Linux... /me investigating

PR #718, opened 2018-12-24
[NIOHTTP1Server] Use fixed default string to write in response and do not hardcode content length

### LINKED ISSUE #714: NIOHTTP1Server/main.swift: "content-length" hardwiring on default content path is ...weird?
(This may be a bit disjointed as I really don't know what I'm doing yet)

In class HTTPHandler, the channelRead() method falls through to effectively returning a default response of "Hello World!" if the request.uri isn't recognized as having a particular handler.

This returns the content as encoded by handlerAdded().

However, channelRead() emits a "content-length" header of 12;  the # of characters in "Hello World!".   If that value is changed to a lower number, the response is truncated.  If it is changed to a larger number the browser will wait on the server to produce more bytes in the body (I think that is what is going on).

At the least, the content generated in handlerAdded() should likely be abstracted so that the "content-length" header's value in channelRead() can be derived from it directly.

----

This seems [to this relative newbie] that the default content generation path (for lack of a better term) that leads to handlerAdded()'s "Hello World!" as being the content is structured in a confusing fashion.   

It looks like the "content-length" header generation should really move to the .end case of the default content path (as is done in, say, handleInfo())?

ilyapuchka:
Motivation:

Resolve #714, Improve example code

Modifications:

Move hardcoded "Hello world" string to a constant property and removed hardcoding of the `content-length` header value

Result:

In the example code, `content-length` will be written correctly when default response changes


Lukasa:

ilyapuchka:

weissi:

Lukasa:
Heh, this test is failing because the previous output was wrong! We were setting `Content-Length` to `12`, but the returned bytes were `"Hello World\r\n"`. Naturally, that is 13 bytes, not 12. The new answer is correct, but the test is looking for the old one. You'll need to update this file here:

https://github.com/apple/swift-nio/blob/0221ab8efc5f44855e0533e1aae35f584d8fbe2d/IntegrationTests/tests_01_http/test_23_repeated_reqs_with_half_closure.sh#L21

ilyapuchka:
@Lukasa indeed, took me few moments to realise that before I saw your comment. Thanks! I've updated it.

Lukasa:
Huh, looks like GitHub thought these two issues were linked.

Lukasa:

ilyapuchka:
It seems because I mistakenly referred this PR instead of an issue in the other PR =/

PR #719, opened 2018-12-25
[Gardening] Use multiline string literals instead of concatenation

broadwaylamb:
Use multiline string literals instead of concatenation

### Motivation:

The rest of the code in NIO uses multiline literals.

### Modifications:

Converted string concatenations to multiline literals.

### Result:

NFC, prettier code.


weissi:
nice one, thanks!

PR #728, opened 2019-01-02
override content encoding header value in compressor

ilyapuchka:
Resolves #718

### Motivation:

Content encoding header may contain invalid value if handlers before compressor have already set it.

### Modifications:

Instead of `headers.add` which appends header value use `headers.replaceOrAdd` which replaces existing value

### Result:

If `Content-Encoding` header was set to some `value`, compressor will reset it to the value of compression algorithm it have picked.

### DIFF
--- a/Sources/NIOHTTP1/HTTPResponseCompressor.swift
+++ b/Sources/NIOHTTP1/HTTPResponseCompressor.swift
@@ -119,8 +119,9 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler {
                 ctx.write(wrapOutboundOut(.head(responseHead)), promise: promise)
                 return
             }
-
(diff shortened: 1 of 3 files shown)


Lukasa:

Lukasa:
Fantastic, thanks @ilyapuchka. We need to back port this to the nio-1.12 branch: do you want to create that back port?

ilyapuchka:
@Lukasa do you mean just a separate PR branched from 1.12?

Lukasa:
Yeah, a PR branched from nio-1.12 that targets nio-1.12. You can even make this easy by cherry picking the merged commit on master, 
https://github.com/apple/swift-nio/commit/b878f06b2079ada6d087ed5acbd25533f60332a3

PR #731, opened 2019-01-03
Fix encoding header

### LINKED ISSUE #717: using HTTPResponseCompressor results in invalid encoding headers
### Expected behavior
HTTPResponseCompressor sets encoding headers to the expected value, i.e. if a response created by the previous handler in the pipeline already set `Content-Encoding` header, the compressor should respect this.

### Actual behavior
When the previous handler in the pipeline sets `Content-Encoding` header to i.e. `gzip` value, then HTTPResponseCompressor, if it detects that response should be compressed with `gzip` (based on `Accept-Encoding` header), adds the same value in the header, which results in its value being `gzip, gzip`, which is then not properly handled by the client (decompression fails). Instead, the value should be just `gzip`.

### Steps to reproduce

1. In `NIOHTTP1Server/main.swift` add `HTTPResponseCompressor` handler after `HTTPHandler`
2. Make `HTTPHandler` set `Content-Encoding` and `Accept-Encoding` headers to `gzip`.

or

1. In [SwiftNIOMock](https://github.com/ilyapuchka/SwiftNIOMock) comment out [SwiftNIOMock/Server.swift:267](https://github.com/ilyapuchka/SwiftNIOMock/blob/58bed932e93da7fad7a1aaf644587f60574d48a9/SwiftNIOMock/Sources/SwiftNIOMock/Server.swift#L267)

2. Run tests, the assert in `testCanRedirectRequestAndInterceptResponse` will fail because no data is received (because it is failing to be decoded)

Problematic code:

https://github.com/apple/swift-nio/blob/305ee9818b8d6de85433bd4da7c70aa860cd0189/Sources/NIOHTTP1/HTTPResponseCompressor.swift#L123

1.12.0

```
Target: x86_64-apple-darwin17.7.0
Darwin b-0800.local 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64
```

ilyapuchka:
This is a back port of #728 into 1.12 branch

Resolves #717

### DIFF
--- a/Sources/NIOHTTP1/HTTPResponseCompressor.swift
+++ b/Sources/NIOHTTP1/HTTPResponseCompressor.swift
@@ -119,8 +119,9 @@ public final class HTTPResponseCompressor: ChannelDuplexHandler {
                 ctx.write(wrapOutboundOut(.head(responseHead)), promise: promise)
                 return
             }
-
-            responseHead.headers.add(name: "Content-Encoding", value: algorithm!.rawValue)
+            // Previous handlers in the pipeline might have already set this header even though
+            // they should not as it is compressor responsibility to decide what encoding to use
+            responseHead.headers.replaceOrAdd(name: "Content-Encoding", value: algorithm!.rawValue)
             initializeEncoder(encoding: algorithm!)
             pendingResponse.bufferResponseHead(responseHead)
             chainPromise(promise)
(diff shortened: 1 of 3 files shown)


Lukasa:

Lukasa:

PR #732, opened 2019-01-03
better DispatchData support

weissi:
better DispatchData support
Motivation:

DispatchData's APIs are less than optimal and the current one we used
did allocate at least on Linux to copy bytes out.

Modifications:

introduce ByteBuffer.write/read/set/getDispatchData

Result:

when Dispatch gets fixed, lower allocations

### DIFF
--- a/Sources/NIO/ByteBuffer-aux.swift
+++ b/Sources/NIO/ByteBuffer-aux.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
+
 extension ByteBuffer {
 
     // MARK: Bytes ([UInt8]) APIs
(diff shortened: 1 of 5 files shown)


Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -169,6 +171,79 @@ extension ByteBuffer {
...
+    ///           `readerIndex` and (excluding) `writerIndex` are always initialized by contract and therefore must be
+    ///           safe to read.
+    /// - parameters:
+    ///     - index: The starting index into `ByteBuffer` containing the string of interest.
+    ///     - length: The number of bytes.
+    /// - returns: A `DispatchData` value deserialized from this `ByteBuffer` or `nil` if the requested bytes aren't contained in this `ByteBuffer`.
+    public func getDispatchData(at index: Int, length: Int) -> DispatchData? {
+        precondition(index >= 0, "index must not be negative")
+        precondition(length >= 0, "length must not be negative")
+        return withVeryUnsafeBytes { pointer in
```
Lukasa: Nit: explicit `self`.

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -169,6 +171,79 @@ extension ByteBuffer {
...
     }
 
+    // MARK: DispatchData APIs
+    /// Write `dispatchData` into this `ByteBuffer`, moving the writer index forward appropriately.
+    ///
+    /// - parameters:
+    ///     - dispatchData: The `DispatchData` instance to write to the `ByteBuffer`.
+    /// - returns: The number of bytes written.
+    @discardableResult
+    public mutating func write(dispatchData: DispatchData) -> Int {
```
Lukasa: Any reason this isn't `@inlinable`?

weissi: @Lukasa don't think there would be any benefit there. No closures/generics so a DispatchData is just a pointer (wrapped in a struct) and we return just an Int.

Lukasa: In which case, why is `set` inlinable? I don’t really care either way, but I do think we should be consistent.

weissi: @Lukasa oh, sorry, I'll remove that. Too much c&p

PR #743, opened 2019-01-10
Support using recvmmsg to perform gathering reads.

Lukasa:
Motivation:

Some DatagramChannels are extremely high-throughput. This can occur
when UDP is being used as an encapsulation for a transport protocol
(e.g. QUIC), or when a highly utilized service is being offered on a
single port (e.g. DNS).

In these cases, the syscall overhead of datagram reads is quite high
(one recvfrom call per datagram). It would be helpful to be able to
trade off some memory usage in order to reduce the cost of the syscalls
in this situation.

This patch adds support for transparently using the recvmmsg syscall
to achieve this kind of tradeoff.

I don't love the API here, and I think it may need more documentation: let me know if you agree.

Modifications:

- Added support to the DatagramChannel for issuing vector reads.
- Added new ChannelOption: DatagramVectorReadMessageCountOption
- Added UDP echo server example to exercise functionality.
- Tests

Result:

Able to achieve higher read throughput on supported systems.


Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -0,0 +1,203 @@
...
+            if self != nil {
+                self!.destroy()
+            }
+            self = nil
+        }
+    }
+}
+
+
+fileprivate extension UnsafeMutableBufferPointer {
```
weissi: could we remove the `fileprivate` here and propagate to the individual funcs?

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -0,0 +1,203 @@
...
+    ///
+    /// - parameters:
+    ///     - socket: The underlying socket from which to read.
+    ///     - buffer: The single large buffer into which reads will be written.
+    func readFromSocket(socket: Socket, buffer: inout ByteBuffer) throws -> ReadResult {
+        assert(buffer.readerIndex == 0, "Buffer was not cleared between calls to readFromSocket!")
+
+        let messageSize = buffer.capacity / self.messageCount
+
+        let result = try buffer.withVeryUnsafeMutableBytes { bufferPointer -> IOResult<Int> in
```
weissi: why are we taking the whole storage and not `writerIndex ..< end` ?

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -0,0 +1,203 @@
...
+        case .wouldBlock(let messagesProcessed):
+            assert(messagesProcessed == 0)
+            return .none
+        case .processed(let messagesProcessed):
+            return self.buildMessages(messageCount: messagesProcessed, sliceSize: messageSize, buffer: &buffer)
+        }
+    }
+
+    /// Destroys this vector read manager, rendering it impossible to re-use. Use of the vector read manager after this is called is not possible.
+    mutating func destroy() {
```
weissi: I don't really like that `init` and `destroy` need to be paired exactly 1:1 without any way of checking that. Could we move the allocations to a function called `allocate` and the destruction into `deallocate`?

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -0,0 +1,203 @@
...
+    mutating func updateMessageCount(_ newCount: Int) {
+        if newCount > 0 {
+            if self != nil {
+                self!.messageCount = newCount
+            } else {
+                self = DatagramVectorReadManager(messageCount: newCount)
+            }
+        } else {
+            if self != nil {
+                self!.destroy()
```
weissi: here it's non-obvious to me why every `.init` is paired with exactly one `destroy`

Lukasa:
Ok, feedback addressed.

Sources/NIO/ChannelOption.swift
```diff
@@ -240,7 +273,7 @@ extension ChannelOptions {
 
                 applier(channel)(key, value).map {
                     applyNext()
-                }.cascadeFailure(to: applyPromise)
+                    }.cascadeFailure(to: applyPromise)
```
weissi: nit: might want to fix the indent, thanking Xcode whilst doing that ;)

Lukasa: Done. :grin:

weissi:
@Lukasa computer is unhappy with license headers:
```
15:24:22 => Checking license headers... missing headers in file './Sources/NIOUDPEchoServer/main.swift'!
15:24:24 --- /dev/fd/63	2019-03-04 15:24:24.585989389 +0000
15:24:24 +++ /tmp/.swift-nio-sanity_aRbLaS	2019-03-04 15:24:22.393003546 +0000
15:24:24 @@ -1,4 +1,3 @@
15:24:24 -
15:24:24  //===----------------------------------------------------------------------===//
15:24:24  //
15:24:24  // This source file is part of the SwiftNIO open source project
15:24:24 @@ -11,3 +10,4 @@
15:24:24  //
15:24:24  // SPDX-License-Identifier: Apache-2.0
15:24:24  //
15:24:24 +//===----------------------------------------------------------------------===//
```

PR #745, opened 2019-01-11
Fix broken test for byte delivery.

Lukasa:
Motivation:

When we write unit tests, we generally like them to actually test the
thing they claim to. This test never ran, and the assertions in place
to verify that it ran also never ran, so we didn't notice the test
not running.

Modifications:

- Moved the assertion that validates the test ran to a place where
  it will definitely execute.
- Ensured the EmbeddedChannel gets marked as active, so that we
  actually deliver the bytes from the HTTPDecoder.

Result:

This test will run, and actually pass.


Lukasa:
This should also be back ported to NIO 1.12.

PR #748, opened 2019-01-15
Swap EventLoopPromise<Bool> to EventLoopPromise<Void>

### LINKED ISSUE #747: ChannelPipeline remove0(ctx:promise:) takes boolean promise, but only ever fulfills with true
## Short Description
This function (https://github.com/apple/swift-nio/blob/nio-1.12/Sources/NIO/ChannelPipeline.swift#L507) takes a `EventLoopPromise<Bool>`, but that promise is only ever fulfilled with `true` which kinda takes away the point of it being a boolean promise.

## Proposed Solution
Replace the `EventLoopPromise<Bool>` with an `EventLoopPromise<Void>` since there lies no further info in the value anyways.

MrLotU:
Fixes #747 

Put in some extra comments/questions in the code

### DIFF
--- a/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
@@ -189,7 +189,7 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
                     self.sendUpgradeResponse(ctx: ctx, upgradeRequest: request, responseHeaders: responseHeaders)
                 }.then {
                     self.removeHandler(ctx: ctx, handler: self.httpEncoder)
-                }.map { (_: Bool) in
+                }.map {
                     self.upgradeCompletionHandler(ctx)
                 }.then {
                     upgrader.upgrade(ctx: ctx, upgradeRequest: request)
@@ -235,11 +235,11 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
     }
 
     /// Removes the given channel handler from the channel pipeline.
-    private func removeHandler(ctx: ChannelHandlerContext, handler: ChannelHandler?) -> EventLoopFuture<Bool> {
+    private func removeHandler(ctx: ChannelHandlerContext, handler: ChannelHandler?) -> EventLoopFuture<Void> {
         if let handler = handler {
             return ctx.pipeline.remove(handler: handler)
         } else {
-            return ctx.eventLoop.makeSucceededFuture(result: true)
+            return ctx.eventLoop.makeSucceededFuture(result: ())
         }
     }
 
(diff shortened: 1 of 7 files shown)


Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -249,7 +249,6 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler {
             return ctx.eventLoop.makeSucceededFuture(result: ())
         }
 
-        return EventLoopFuture<Void>.andAll(self.extraHTTPHandlers.map { ctx.pipeline.remove(handler: $0).map { (_: Bool) in () }},
-                                            eventLoop: ctx.eventLoop)
+        return EventLoopFuture<Void>.andAll(self.extraHTTPHandlers.map { ctx.pipeline.remove(handler: $0).map { }}, eventLoop: ctx.eventLoop)
```
MrLotU: I made this 1 line, but not sure what the cutoff length should be

Tests/NIOTests/AcceptBackoffHandlerTest.swift
```diff
@@ -96,7 +96,9 @@ public class AcceptBackoffHandlerTest: XCTestCase {
             return readCountHandler.readCount
         }.wait())
 
-        XCTAssertTrue(try serverChannel.pipeline.remove(name: acceptHandlerName).wait())
+        try serverChannel.pipeline.remove(name: acceptHandlerName).mapIfError { _ in
```
MrLotU: Wasn't fully sure how to convert this, but figured this'd be the best way to go

Tests/NIOTests/CodecTest.swift
```diff
@@ -392,8 +392,6 @@ public class ByteToMessageDecoderTest: XCTestCase {
 
         channel.pipeline.context(handlerType: ByteToMessageHandler<PairOfBytesDecoder>.self).then { ctx in
             return channel.pipeline.remove(ctx: ctx)
-        }.map {
-            XCTAssertTrue($0)
         }.whenFailure { error in
```
MrLotU: I removed the map to `Bool` here since: first off, it won't happen ever, and 2nd off, if the handler is not removed, it'll throw an error, so the `XCTFail` will trigger

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
weissi: that `.map { }` can go. Hmm, I'd break the line after the opening brace for the first map

Tests/NIOTests/AcceptBackoffHandlerTest.swift
weissi: looks good

Tests/NIOTests/AcceptBackoffHandlerTest.swift
```diff
@@ -96,7 +96,9 @@ public class AcceptBackoffHandlerTest: XCTestCase {
...
+            XCTFail("Handler was not removed")
+        }.wait()
```
Lukasa: This conversion can just be `XCTAssertNoThrow(try serverChannel.pipeline.remove(name: acceptHandlerName).wait())`.

Lukasa:
Can you also update the public-api-changes-NIO1-to-NIO2.md document?

Tests/NIOTests/AcceptBackoffHandlerTest.swift
MrLotU: Fixed :)

Lukasa:

weissi:
this has already landed with #767 , sorry @MrLotU

PR #751, opened 2019-01-16
Conform NIOAny to CustomStringConvertible

JovanMilenkovic:
Motivation:

NIOAny is used extensively across SwiftNIO but can't be reasonably debugged. Users may want to simply call String(describing: CustomStringConvertible) to inspect what is the data they are working with at the moment instead of seeing an output that has no debug value.

Modifications:

Conform NIOAny to CustomStringConvertible, along with other types it currently uses that lacked string representation.

Result:

Users can perform simple debugging by printing their NIOAny object.


Sources/NIO/FileHandle.swift
```diff
@@ -90,3 +90,9 @@ extension FileHandle {
         self.init(descriptor: fd)
     }
 }
+
+extension FileHandle: CustomStringConvertible {
+    public var description: String {
+        return "FileHandle { descriptor: \(descriptor), isOpen: \(isOpen) }"
```
weissi: mind adding explicit `self.` in front of descriptor and isOpen?

JovanMilenkovic:
> Thanks for this patch, it seems very useful! Do you mind addressing a few of the notes I've left in the diff?

Sure, how do I go about following the commit template for this new commit?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -49,6 +49,13 @@ extension _ByteBufferSlice {
     }
 }
 
+extension _ByteBufferSlice: CustomStringConvertible {
+    @usableFromInline
+    var description: String {
+        return "\(lowerBound)..<\(upperBound)"
```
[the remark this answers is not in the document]
JovanMilenkovic: I followed the ByteBuffer: CustomStringConvertible documentation here, I assumed it includes "slice: 256..<768" because _ByteBufferSlice is to be considered an implementation detail. Should I modify both?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -664,7 +671,7 @@ extension ByteBuffer: CustomStringConvertible {
         capacity: \(self.capacity), \
         slice: \(self._slice), \
         storage: \(self._storage.bytes) (\(self._storage.capacity) bytes)\
-        }
+         }
```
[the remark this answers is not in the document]
JovanMilenkovic: Sure, but isn't this a bug? Perhaps I should have put the space at the end of the previous line?

Sources/NIO/ByteBuffer-core.swift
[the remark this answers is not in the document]
Lukasa: `slice` is a property name, whereas this is a type name. I don’t think it matters too much to add it, especially as this should now be meaningful standalone.

Sources/NIO/ByteBuffer-core.swift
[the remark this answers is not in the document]
Lukasa: Yup, the space should go at the end of the previous line as shown elsewhere.

weissi:
@Lukasa are you happy with this?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -49,6 +49,13 @@ extension _ByteBufferSlice {
     }
 }
 
+extension _ByteBufferSlice: CustomStringConvertible {
+    @usableFromInline
+    var description: String {
+        return "_ByteBufferSlice { \(self.lowerBound)..<\(self.upperBound) }"
```
Lukasa: Should this print the parent buffer?

weissi: probably yes, @JovanMilenkovic happy adding that? Otherwise we can also do this as a followup

JovanMilenkovic: I’ll see if I could check it out later tonight, but if I remember correctly `ByteBuffer` has it’s own description that also contains the slice. Wouldn’t that double print the whole buffer and cause a confusing print or am I missing something?

JovanMilenkovic: Also potentially infinite if it returns the full description itself...

weissi: @jovanmilenkovic you’re right. I thought it was ByteBufferView but this is the slice so your code is fine as is. Sorry

weissi:

Sources/NIO/ByteBuffer-core.swift
Lukasa: Oof, so did I. The price you pay for reviewing on a phone.

Lukasa:

Lukasa:
@JovanMilenkovic Looks like you need to run `scripts/generate_linux_tests.rb`.

weissi:

weissi:

weissi:
@JovanMilenkovic doesn't compile

```
19:26:28 /code/Sources/NIO/FileHandle.swift:94:11: error: use of undeclared type 'FileHandle'
19:26:28 extension FileHandle: CustomStringConvertible {
19:26:28           ^~~~~~~~~~
19:26:31 Build step 'Execute shell' marked build as failure
```

that has been renamed a while ago to `NIOFileHandle`, sorry. I added the missing license header in one of the files.

weissi:

weissi:
@JovanMilenkovic I made a few fixes but the code still doesn't compile:

```
19:30:00 /code/Tests/NIOTests/NIOAnyDebugTest.swift:28:36: error: incorrect argument label in call (have 'descriptor:', expected 'fileDescriptor:')
19:30:00         let fileHandle = FileHandle(descriptor: 1)
19:30:00                                    ^~~~~~~~~~~
19:30:00                                     fileDescriptor
19:30:00 /code/Tests/NIOTests/NIOAnyDebugTest.swift:61:9: error: value of type 'ByteBuffer' has no member 'write'
19:30:00         buffer.write(string: string)
19:30:00         ^~~~~~ ~~~~~
```

mind fixing them? Sorry

JovanMilenkovic:
@weissi pushed a fix, it should be okay now. Sorry wasn't even checking, seems like this branch went pretty stale.

Lukasa:

Lukasa:

PR #753, opened 2019-01-16
Docs: move defer closer to the group instantiation

kevints:
This makes the boot/shutdown pair visually closer, reinforcing that you shouldn't forget to shutdown the EventLoopGroup.


Lukasa:

Issue #713, opened 2018-12-17
handleJustWrite ignores statusCode


bbum:
In NIOHTTP1Server/main.swift, the handleJustWrite(...) method of the HTTPHandler class has a statusCode argument (that defaults to .ok).

The argument is ignored in the call to httpResponseHead(), which makes all responses from handleJustWrite() locked to .ok.

This includes responses to /dynamic/ requests to nonsense sub-requests-- say, /dynamic/mumblejumble-- which passes a status of .notFound (and a body of "not found").

bbum:
--- a/Sources/NIOHTTP1Server/main.swift
+++ b/Sources/NIOHTTP1Server/main.swift
@@ -172,7 +172,7 @@ private final class HTTPHandler: ChannelInboundHandler {
         case .head(let request):
             self.keepAlive = request.isKeepAlive
             self.state.requestReceived()
-            ctx.writeAndFlush(self.wrapOutboundOut(.head(httpResponseHead(request: request, status: .ok))), promise: nil)
+            ctx.writeAndFlush(self.wrapOutboundOut(.head(httpResponseHead(request: request, status: statusCode))), promise: nil)
         case .body(buffer: _):
             ()
         case .end:

normanmaurer:
@bbum good catch... want to provide a PR with a fix ?

PR #755, opened 2019-01-17
Fix NIOHTTP1Server handleJustWrite always using response status "ok"

JovanMilenkovic:
Motivation:

Addressing the https://github.com/apple/swift-nio/issues/713

Modifications:

Forward the statusCode from NIOHTTP1Server main handleJustWrite to ctx.writeAndFlush instead of always using "ok"

Result:

ChannelHandlerContext will write the correct status code even when it's not "ok"


Lukasa:

Issue #756, opened 2019-01-18
make ELF.cascade accept an optional promise


weissi:
code like this is annoying

```
        let f = ctx.connect(to: address).then {
            ctx.channel.setOption(option: NIOTSChannelOptions.waitForActivity, value: false)
        }
        if let promise = promise {
            f.cascade(promise: promise)
        }
```

and should be

```
        let f = ctx.connect(to: address).then {
            ctx.channel.setOption(option: NIOTSChannelOptions.waitForActivity, value: false)
        }.cascade(promise)
```

fix is easy: `func cascade(_ promise: EventLoopPromise<T>?)`

PR #757, opened 2019-01-19
make ELF.cascade accept an optional promise

DavidSkrundz:
Make EventLoopFuture.cascade and EventLoopFuture.cascadeFailure accept an optional promise

### Motivation:

Addressing #756

Code like this

```Swift
    let f = ctx.connect(to: address).then {
        ctx.channel.setOption(option: NIOTSChannelOptions.waitForActivity, value: false)
    }
    if let promise = promise {
        f.cascade(promise: promise)
        // or f.cascadeFailure(promise)
    }
```

can be replaced with this

```Swift
    let f = ctx.connect(to: address).then {
        ctx.channel.setOption(option: NIOTSChannelOptions.waitForActivity, value: false)
    }.cascade(promise)
    // or }.cascadeFailure(promise)
```

### Modifications:

Change EventLoopFuture.cascade to `func cascade(promise: EventLoopPromise<T>?)`
Change EventLoopFuture.cascadeFailure to `func cascadeFailure<U>(promise: EventLoopPromise<U>?)`

### Result:

`cascade` and `cascadeFailure` can be called without needing to check if the promise is nil

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -802,9 +802,7 @@ class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore {
         registerPromise.futureResult.whenFailure { (_: Error) in
             self.close(promise: nil)
         }
-        if let promise = promise {
-            registerPromise.futureResult.cascadeFailure(promise: promise)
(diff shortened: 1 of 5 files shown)


weissi:

PR #761, opened 2019-01-19
speed up getBytes

weissi:
Motivation:

Due to https://bugs.swift.org/browse/SR-9604, ByteBuffer.getBytes
seriously regressed (factor 10x) since NIO 1.11 . This works around it
until the issue is fixed.

Modifications:

bind the memory to UInt8s to get an UnsafeBufferPointer<UInt8> rather
than an UnsafeRawBufferPointer.

Result:

- getBytes 10x faster
- make @John-Connolly happy, thanks for reporting!


Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -39,7 +39,8 @@ extension ByteBuffer {
         }
 
         return self.withVeryUnsafeBytes { ptr in
-            Array<UInt8>(ptr[index..<(index+length)])
+            // this is not technically 100% but we need to work around https://bugs.swift.org/browse/SR-9604
```
Lukasa: Not technically 100% what? I assume the next word is "safe".

weissi: thanks! updated

PR #762, opened 2019-01-19
[1.12 backport] speed up getBytes

weissi:
Motivation:

Due to https://bugs.swift.org/browse/SR-9604, ByteBuffer.getBytes
seriously regressed (factor 10x) since NIO 1.11 . This works around it
until the issue is fixed.

Modifications:

bind the memory to UInt8s to get an UnsafeBufferPointer<UInt8> rather
than an UnsafeRawBufferPointer.

Result:

- getBytes 10x faster
- make @John-Connolly happy, thanks for reporting!

(cherry picked from commit 264fbe0f315bd50e4a5b81b92ffc5cc93dc558f3)

### DIFF
--- a/Sources/NIO/ByteBuffer-aux.swift
+++ b/Sources/NIO/ByteBuffer-aux.swift
@@ -37,7 +37,10 @@ extension ByteBuffer {
         }
 
         return self.withVeryUnsafeBytes { ptr in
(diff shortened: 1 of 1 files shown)


weissi:
ping @Lukasa , same code as the other one for `master` but for `nio-1.12`

Issue #243, opened 2018-03-27
EmbeddedChannel.readOutbound should be able to deal with types other than IOData


weissi:
### Expected behavior

You'd expect that with an `EmbeddedChannel` you can read any type from the outbound buffer. Unfortunately the signature is `public func readOutbound() -> IOData?`. This should be changed to `public func readOutbound() -> NIOAny?` or `public func readOutbound() -> Any?`.

The reason that this is important is that for testing you don't want to serialise everything before testing what went through the `ChannelPipeline`. Sure you can workaround with an `OutboundChannelHandler` that collects everything but that's rather annoying and isn't necessary.

Annoyingly `EmbeddedChannel` is `public` that means this will need to wait until we make a 2.0 release.

### Actual behavior

can only read `IOData`.

### Steps to reproduce

1. Make an `EmbeddedChannel`
2. Write some arbitrary type that isn't `IOData` to it (for example `String`)
3. try to read it out

### SwiftNIO version/commit hash

1.x

### Swift & OS version (output of `swift --version && uname -a`)

any

IOOI-SqAR:
I think this should have Milestone 2.0.0 too

weissi:
fixed now with #763

PR #763, opened 2019-01-20
Remove EmbeddedChannel restriction to IOData

JovanMilenkovic:
Motivation:

You should be able to read any type from the outbound buffer. Related issue: https://github.com/apple/swift-nio/issues/243

Modifications:

Change Embedded channel outboundBuffer and pendingOutboundBuffer to store NIOAny instead of IOData and refactor test code accordingly.

Result:

Users can test what went through the pipeline without serializing everything and using "if case let" syntax everywhere.


Sources/NIO/Embedded.swift
```diff
@@ -325,7 +320,7 @@ public class EmbeddedChannel: Channel {
     // Embedded channels never have parents.
     public let parent: Channel? = nil
 
-    public func readOutbound() -> IOData? {
+    public func readOutbound() -> NIOAny? {
```
Lukasa: Returning `NIOAny` from here is really painful, because without importing NIO as `@testable` you cannot unwrap this to the appropriate type (the various `tryAs` functions are internal).

This should use the same pattern as `readInbound`, which is that it should be generic and return a `T`.

Sources/NIO/Embedded.swift
```diff
@@ -359,7 +354,7 @@ public class EmbeddedChannel: Channel {
         }
     }
 
-    private func readFromBuffer(buffer: inout [IOData]) -> IOData? {
+    private func readFromBuffer(buffer: inout [NIOAny]) -> NIOAny? {
```
Lukasa: If you make the change suggested above, this function can just go away and we can use the generic variant instead.

Sources/NIO/Embedded.swift
Lukasa: That said, it may be helpful to adopt a change we've made elsewhere, which is that this and `readInbound` could take an optional type parameter. That is:

```swift
public func readOutbound<T>(as type: T.Type) -> T?
    return self.readFromBuffer(buffer: &self.channelCore.outboundBuffer)
}
```

The same change could be made to `readInbound` as well, which will help clear up the testing changes.

Tests/NIOHTTP1Tests/HTTPServerProtocolErrorHandlerTest.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import XCTest
-import NIO
+@testable import NIO
```
Lukasa: With my changes above, we should be able to remove this testable import.

Tests/NIOWebSocketTests/WebSocketFrameDecoderTest.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import XCTest
-import NIO
+@testable import NIO
```
Lukasa: We should be able to remove this testable import.

Tests/NIOWebSocketTests/WebSocketFrameEncoderTest.swift
```diff
@@ -13,13 +13,13 @@
 //===----------------------------------------------------------------------===//
 
 import XCTest
-import NIO
+@testable import NIO
```
Lukasa: We should be able to remove this testable import.

Sources/NIO/Embedded.swift
JovanMilenkovic: Great solution tnx.
Did you mean `public func readOutbound<T>(as type: T.Type = T.self) -> T?`?

JovanMilenkovic: Also in the meantime I ran into an issue. You would not be able to ever use
`XCTAssertNotNil(channel.readOutbound())`
since that would try to get `Any` which would crash when trying to read `.other` while you actually have say a `ByteBuffer` stored. How should this be addressed(if it should be)?

weissi: @JovanMilenkovic yes, `public func readOutbound<T>(as type: T.Type = T.self) -> T?` is perfect.

So maybe we could make the `tryAsAny` case a bit smarter and have it no mandate the `.other` case? I think if we request `Any` but find `.byteBuffer` we should still handle this properly. But I'm happy if we don't do this in this PR but in a separate one instead.

JovanMilenkovic: I’m not sure at what point is semantics broken. If i `tryAsOther` and request a `ByteBuffer` should it work? The  case and `tryAsOther`

JovanMilenkovic: I’m not sure at what point is semantics broken. If i `tryAsOther` and request a `ByteBuffer` should it work? 
The storage cases and `tryAsOther` are at the same access control level. Any input?

Lukasa: > If yes, how does it help with clearing testing changes since there are no changes to `readInbound` methods?

You're right, we'll still need to change those. However, they'll be closer to the style of `readInbound` than this proposed change.

> Also would it be preferred to use the method parameter unless the type is already defines somewhere above?

Yes.

> How should this be addressed(if it should be)?

You write this:

```swift
XCTAssertNotNil(channel.readOutbound(as: IOData.self))
```

Or whatever is the correct type. You are never allowed to not say what type you want. This is a feature, not a bug, IMO.

> The storage cases and `tryAsOther` are at the same access control level. Any input?

I think ignore the `tryAs` functions: they are not for NIO users, they are specifically used internally to NIO. We should not expect users to call them.

JovanMilenkovic: The thing that looks like a bug is that you say that you want `Any`(or the compiler infers it) and you get an assert if storage is not on `.other` and the message is quite cryptic with you tried to cast to `Any` but failed. That said, I’ve tried both and I think it should be explicit as you said, it doesn’t seem like it’s worth complicating `NIOAny`

Tests/NIOHTTP1Tests/HTTPServerProtocolErrorHandlerTest.swift
Lukasa: Can we remove this?

Tests/NIOWebSocketTests/WebSocketFrameDecoderTest.swift
Lukasa: Any this?

Tests/NIOWebSocketTests/WebSocketFrameEncoderTest.swift
Lukasa: And this?

Lukasa:
It looks like they got over-fixed: some of those test suites did actually need the `@testable` import.

Lukasa:
(I should note that none of the ones that had `@testable` *added* needed them, but @JovanMilenkovic removed some from other files that did in fact need them)

weissi:

weissi:
@Lukasa I just fixed the extra changes that lead to the breakage, fingers crossed that it will pass the tests now :)

weissi:

JovanMilenkovic:
Whoops sorry I was sure I ran the tests locally before pushing

weissi:
@JovanMilenkovic no problem. This time however it hit a real failure:
```
13:55:29 Test Case 'EmbeddedChannelTest.testSendingAnythingOnEmbeddedChannel' started at 2019-01-29 13:55:29.329
13:55:29 Assertion failed: leaked open FileHandle(descriptor: 1). Call `close()` to close or `takeDescriptorOwnership()` to take ownership and close by some other means.: file /code/Sources/NIO/FileHandle.swift, line 37
```

weissi:
@JovanMilenkovic I think the `1` you use is a fake file descriptor and not referring to `stdout`, right? If that's the case you can just call `takeDescriptorOwnership()` and discard the result. If that's a real file descriptor, please close it

JovanMilenkovic:
I’ll try to find some time tonight to fix it but if you have a pending release or something like that feel free to alter the PR.

weissi:
@JovanMilenkovic thank you and no rush

weissi:

PR #767, opened 2019-01-21
better ChannelHandler removal API

weissi:
Motivation:

If ChannelHandler removal worked correctly, it was often either by
accident or by intricate knowledge about the implementation of the
ChannelHandler that is to be removed. Especially when it comes to
re-entrancy it mostly didn't work correctly.

Modifications:

- introduce a `RemovableChannelHandler` API
- raise allocation limit per HTTP connection by 1
  (https://bugs.swift.org/browse/SR-9905)

Result:

Make things work by contruction rather than accident

### DIFF
--- a/Sources/NIO/ChannelHandler.swift
+++ b/Sources/NIO/ChannelHandler.swift
@@ -309,3 +309,37 @@ extension _ChannelInboundHandler {
     }
 }
 
+/// A `RemovableChannelHandler` is a `ChannelHandler` that can be dynamically removed from a `ChannelPipeline` whilst
+/// the `Channel` is operating normally.
+/// A `RemovableChannelHandler` is required to remove itself from the `ChannelPipline` (using
+/// `ChannelHandlerContext.removeHandler`) as soon as possible.
+///
+/// - note: When a `Channel` gets torn down, every `ChannelHandler` in the `Channel`'s `ChannelPipeline` will be
+///         removed from the `ChannelPipeline`. Those removals however happen synchronously and are not going through
+///         the methods of this protocol.
+public protocol RemovableChannelHandler: ChannelHandler {
+    /// Ask the receiving `RemovableChannelHandler` to remove itself from the `ChannelPipeline` as soon as possible.
+    /// The receiving `RemovableChannelHandler` may elect to remove itself sometime after this method call, rather than
+    /// immediately, but if it does so it must take the necessary precautions to handle events arriving between the
+    /// invocation of this method and the call to `ChannelHandlerContext.removeHandler` that triggers the actual
+    /// removal.
+    ///
+    /// - note: Like the other `ChannelHandler` methods, this method should not be invoked by the user directly. To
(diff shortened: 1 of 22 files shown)


Sources/NIO/ChannelHandler.swift
```diff
@@ -309,3 +309,12 @@ extension _ChannelInboundHandler {
     }
 }
 
+public protocol RemovableChannelHandler: ChannelHandler {
+    func removeHandler(ctx: ChannelHandlerContext, promise: EventLoopPromise<Void>?)
```
Lukasa: As discussed earlier today, we should investigate opaquifing this.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1472,6 +1476,13 @@ public final class ChannelHandlerContext: ChannelInvoker {
     }
 }
 
+extension ChannelHandlerContext {
+    public func removeHandler(promise: EventLoopPromise<Void>?) {
```
Lukasa: Can we get a doc comment here?

Sources/NIO/Codec.swift
```diff
@@ -429,3 +436,13 @@ extension MessageToByteEncoder {
...
     }
 }
+
+extension ByteToMessageHandler: RemovableChannelHandler {
+    public func removeHandler(ctx: ChannelHandlerContext, promise: EventLoopPromise<Void>?) {
+        ctx.eventLoop.execute {
+            self.processLeftovers(ctx: ctx)
+            assert(self.state != .leftoversNeedProcessing)
+            ctx.removeHandler(promise: promise)
+        }
```
Lukasa: This implementation is a bit awkward: specifically, it doesn't stop the decode loop if one is running. This means it's entirely possible for the `B2MD` to process more data. It would probably be nicer if this interrupted the decode loop.

Sources/NIO/Codec.swift
```diff
@@ -351,6 +359,9 @@ extension ByteToMessageHandler: ChannelInboundHandler {
 
 
     public func handlerRemoved(ctx: ChannelHandlerContext) {
+        assert(self.removalState == .notBeingRemoved || self.removalState == .removalCompleted,
+               "illegal removal state in handlerRemoved: \(self.removalState)")
+        self.removalState = .removalCompleted
         self.processLeftovers(ctx: ctx)
```
Lukasa: Do we still want this call? I think it's only useful if the user removes the handler from the body of `decode` using the `ChannelHandlerContext`, and even there it may be a little spurious as the leftovers aren't reachable (the call re-enters immediately). The only other way we can get here without going through the closure in `removeHandler` is if the pipeline is being torn down, and that teardown means delivering the leftovers is both a) of limited use, and b) semantically invalid (the user never asked to remove B2MD from the pipeline, so the arrival of a ByteBuffer will just cause a crash).

weissi: oops, no

Lukasa:
Minor build failure, but otherwise looks good

PR #768, opened 2019-01-21
provide NIO1 API shim for ELF.then*

weissi:
Motivation:

The switch from `ELF.then*` to `ELF.flatMap*` is pretty big so we should
provide NIO1 API shims.

Modifications:

add API shims for ELF.then*

Result:

easier for people to migrate from NIO1 to NIO2.


weissi:
good catch @Lukasa , added

Lukasa:
Huh, CI failure.

weissi:
@Lukasa damn, was using the deprecated `T` (deliberately) but warnings to errors :P

PR #769, opened 2019-01-21
Replace ByteBufferTest use of deprecated "withUnsafeBytes" with new API

JovanMilenkovic:
Motivation:

Remove the only warning in the project.

Modifications:

Use new withUnsafeBytes API.

Result:

The project will have no warnings.

### DIFF
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -763,9 +763,7 @@ class ByteBufferTest: XCTestCase {
         buf.write(string: str)
         var written1: Int = -1
         var written2: Int = -1
-        let hwDataCount = hwData.count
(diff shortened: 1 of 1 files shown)


Lukasa:
Leaving this for @weissi because I haven't seen the deprecation warning that is apparently being emitted here.

JovanMilenkovic:
Ah right, I knew there had to be a reason.

weissi:
@millenomi should we file a bug that Linux diverges from Darwin here?

millenomi:
No need; I’m fast-tracking https://github.com/apple/swift-corelibs-foundation/pull/1813 which is the implementation you’re looking for.

millenomi:
Also: https://github.com/apple/swift-corelibs-foundation/pull/1853

millenomi:
Upstream is merged.

weissi:
thank you! Merged with #786

Issue #603, opened 2018-08-31
consider changing timeouts to deadlines


weissi:
I think deadlines generally work better than timeouts. Why? Imagine for example a caller wants to have something complete in no more than 5 seconds. So for example

    doSomething(timeout: .seconds(5))

If `doSomething` kicks off the operation straight away, it's all good. However `doSomething` might need to use two other sub-systems to complete the operation. Example:

    func doSomething(timeout: Timeout) -> EventLoopFuture<Void> {
        return step1(timeout: timeout).then {
            step2(timeout: ???)
        }
    }

So what do we fill into the question marks? Sure, we could write

    func doSomething(timeout: Timeout) -> EventLoopFuture<Void> {
        let start = Date()
        return step1(timeout: timeout).then {
            let step1Stop = Date()
            let timeTaken = step1Stop - start
            let timeRemaining = timeout - timeTaken
            if timeRemaining > 0 {
                step2(timeout: timeRemaining)
            } else {
                return eventloop.newFailedFuture(error: TimeoutError())
            }
        }
    }

but that's incredibly ugly and error prone. With deadlines however the call would look like

    doSomething(timeout: .now() + .seconds(5))

and `doSomething` would look much nicer:

    func doSomething(deadline: Deadline) -> EventLoopFuture<Void> {
        return step1(deadline: deadline).then {
            step2(deadline: deadline)
        }
    }

and the `stepX` functions would just need to check if they're still within the deadline.

Basically, deadlines just compose much better than timeouts. So I think for NIO 2.0 we should make this change. @normanmaurer / @Lukasa for input.

fumoboy007:
FWIW, Go also uses deadlines. [`context.WithTimeout`](https://golang.org/pkg/context/#WithTimeout) simply wraps `context.WithDeadline`.

fumoboy007:
Also, some protocols such as [TChannel](https://github.com/uber/tchannel/blob/master/docs/protocol.md) use deadlines and the deadline of the inbound request is expected to be reused for outbound requests.

AlanQuatermain:
I think this sounds like a good idea. Generally speaking the caller of an API is expecting it to function in this way, and I've often found myself (as an API implementor) converting the timeout to a hard date as the first thing I do within a function, ready to pass the resulting deadline to other pieces as necessary (and I've hit the corresponding bugs when I failed to do so).

[...]

PR #770, opened 2019-01-22
Add deadlines as an alternative to timeouts

DavidSkrundz:
`scheduleTask` can be called with a `NIODeadline` or `TimeAmount`

```Swift
func scheduleTask<T>(at: NIODeadline, _ task: @escaping () throws -> T) -> Scheduled<T>
```

### Motivation:

Address #603

### Modifications:

- Add Time struct
- Add `at: NIODeadline` alternatives to `in: TimeAmount` functions
- Reduce the use of DispatchTime in favor of Time
- Replace timeout calls with deadline calls where the intention was a deadline

### Result:

Tasks can be scheduled after an amount of time has passed, or at a certain time.


Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,74 @@ extension TimeAmount: Comparable {
...
+        return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
+    }
+
+    public static func * (lhs: TimeAmount.Value, rhs: TimeAmount) -> TimeAmount {
+        return TimeAmount(lhs * rhs.nanoseconds)
+    }
+
+    public static func * (lhs: TimeAmount, rhs: TimeAmount.Value) -> TimeAmount {
+        return TimeAmount(lhs.nanoseconds * rhs)
+    }
```
Lukasa: The bottom two implementations are written against `Int64`. Should we instead write them to be more generic, e.g. over `BinaryInteger`? I'm not sure that we should, but I *definitely* think we should also write an implementation that uses `Int`.

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,74 @@ extension TimeAmount: Comparable {
...
+}
+
+/// Represents a point in time.
+///
+/// - note: `Time` should not be used to represent an interval
```
Lukasa: Can you flesh this doc comment out a bit? Explanations of how it works would be good.

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,74 @@ extension TimeAmount: Comparable {
...
+public struct Time {
```
Lukasa: We can't call this `Time`, it's against our new enhancement policy. At the very least we have to call it `NIOTime`.

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,74 @@ extension TimeAmount: Comparable {
...
+
+    public static func exactly(_ nanoseconds: Value) -> Time {
+        return Time(nanoseconds)
+    }
+}
+
+extension Time: Comparable {
+    public static func < (lhs: Time, rhs: Time) -> Bool {
+        return lhs.nanoseconds < rhs.nanoseconds
+    }
```
Lukasa: May as well implement `>` as well, given that it's so straightforward.

Sources/NIO/EventLoop.swift
Lukasa: This struct is missing some conformances. I think we need:

- `Equatable` and `Hashable` (the synthesised conformances will be fine)
- `CustomStringConvertible`

Sources/NIO/EventLoop.swift
DavidSkrundz: I added an `Int` version

Lukasa:

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,106 @@ extension TimeAmount: Comparable {
...
+extension TimeAmount {
+    public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
+        return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
+    }
+
+    public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
+        return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
+    }
+
+    public static func * (lhs: TimeAmount.Value, rhs: TimeAmount) -> TimeAmount {
```
weissi: can we remove all those `*`s? `TimeAmount` (1 ns * 1 ns != 1 ns, it's 1 ns^2)

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,106 @@ extension TimeAmount: Comparable {
...
+/// }
+/// doSomething(deadline: .now() + .seconds(5))
+/// ```
+///
+/// - note: `NIODeadline` should not be used to represent a time interval
+public struct NIODeadline: Equatable, Hashable {
+    public typealias Value = UInt64
+
+    /// The nanoseconds representation of the `NIODeadline`.
+    public let nanoseconds: Value
```
weissi: this should be called something else I think. Maybe `uptimeNanoseconds` too? Otherwise people think that they can save this number and re-create a new `NIODeadline` from it elsewhere. But that's only true on the same computer and the same boot.

Sources/NIO/EventLoop.swift
```diff
@@ -291,6 +295,106 @@ extension TimeAmount: Comparable {
...
+
+    private init(_ nanoseconds: Value) {
+        self.nanoseconds = nanoseconds
+    }
+
+    public static func now() -> NIODeadline {
+        return NIODeadline(DispatchTime.now().uptimeNanoseconds)
+    }
+
+    public static func exactly(_ nanoseconds: Value) -> NIODeadline {
```
weissi: _exactly_ suggests that the other methods aren't quite as exact. Why not `static func uptimeNanoseconds(...)`?

Sources/NIO/EventLoop.swift
```diff
@@ -555,7 +659,7 @@ internal final class SelectableEventLoop: EventLoop {
         return thread.isCurrent
     }
 
-    public func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T> {
+    public func scheduleTask<T>(at: NIODeadline, _ task: @escaping () throws -> T) -> Scheduled<T> {
```
weissi: I'm happy enough with `at:` but why not just `deadline: NIODeadline`? Just a thought. @Lukasa / @normanmaurer  any input?

Sources/NIO/Selector.swift
```diff
@@ -248,7 +246,7 @@ final class Selector<R: Registration> {
     private typealias EventType = Epoll.epoll_event
     private let eventfd: Int32
     private let timerfd: Int32
-    private var earliestTimer: UInt64 = UInt64.max
+    private var earliestTimer: NIODeadline = .exactly(.max)
```
weissi: we should have a constant for this called `.distantFuture` or `.forever` or something

Sources/NIO/EventLoop.swift
Lukasa: This isn't actually wrong, it's `TimeAmount.Value` * `TimeAmount`, which is to say `UInt64` * `TimeAmount`, which does actually have units of `ns`.

Sources/NIO/EventLoop.swift
Lukasa: Agreed.

Sources/NIO/EventLoop.swift
Lukasa: `deadline:` works for me.

Sources/NIO/Selector.swift
Lukasa: Yup, `.distantFuture` and `.distantPast` are the relevant prior art.

Sources/NIO/EventLoop.swift
Lukasa: Possibly? I think @weissi has stronger opinions than I do on this topic.

weissi: oh sorry, I totally missed that `lhs` and `rhs` have different types. I think `BinaryInteger` and `TimeAmount` sounds better. Not entirely sure if we really need multiplication here but I'm fine having it too :)

weissi:

weissi:

Lukasa:
Looks like a minor issue with one of the calls falling out of step with the `master` branch, mind updating?

weissi:
@Mordil  that't the error

```
09:13:45 [10/37] Compiling Swift Module 'NIO' (54 sources)
09:13:48 /code/Sources/NIO/EventLoop.swift:680:25: error: extraneous argument label 'error:' in call
09:13:48             promise.fail(error: error)
09:13:48                         ^~~~~~~~
```

weissi:

PR #773, opened 2019-01-23
Add `whenAllComplete` EventLoopFuture method

Mordil:
Add `whenAllComplete` method to group a homogenous array of future results into a single ELF that returns a `[Result<Value, Error>]`

### Motivation:
Methods like `reduce`, `and`, or `fold` all "fail fast" by failing the top-level ELF if any of the sub-ELFs hit a failure.

However, it is common to have a group of sub-tasks that may have results to work with at the end, but each task's success or failure status is mutually exclusive from each other.

An example might be a bulk update of records in a database - where I just want to know when all of them completed. If one failed, I still want all the others to continue.

For example:

```swift
let taskResults: EventLoopFuture<[Result<Int, Error>]> = .whenAllComplete([
    obj1.updateDatabase(),
    obj2.updateDatabase(),
    // ...
], eventLoop: eventLoop)
```

This will allow me to handle each individual failure, or ignoring the results altogether.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -1030,10 +1030,10 @@ extension EventLoopFuture {
                     guard remainingCount == 0 else { return }
 
                     // verify that all operations have been completed
-                    assert(results.contains(where: {
+                    assert(!results.contains(where: {
                         guard case let .failure(error) = $0 else { return false }
                         return error is OperationPlaceholderError
-                    }) == false)
+                    }))
 
                     promise.succeed(result: results)
                 }
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -1035,7 +1035,7 @@ extension EventLoopFuture {
                         return error is OperationPlaceholderError
                     }))
 
-                    promise.succeed(result: results)
+                    promise.succeed(results)
                 }
         }
     }
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -996,9 +996,9 @@ extension EventLoopFuture {
     /// use one of the `reduce` methods instead.
     /// - Parameter futures: An array of homogenous `EventLoopFuture`s to gather results from.
     /// - Returns: A new `EventLoopFuture` with all the results of the provided futures.
-    public static func whenAllComplete<InputValue>(_ futures: [EventLoopFuture<InputValue>],
-                                                   eventLoop: EventLoop) -> EventLoopFuture<[Result<InputValue, Error>]> {
-        let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)
+    public static func whenAllComplete(_ futures: [EventLoopFuture<Value>],
+                                       eventLoop: EventLoop) -> EventLoopFuture<[Result<Value, Error>]> {
+        let promise = eventLoop.makePromise(of: [Result<Value, Error>].self)
 
         if eventLoop.inEventLoop {
             _whenAllComplete0(promise, futures, eventLoop: eventLoop)
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -106,6 +106,9 @@ private struct CallbackList: ExpressibleByArrayLiteral {
 
 }
 
+/// Internal error for operations that return results that were not replaced
+private struct OperationPlaceholderError: Error { }
+
 /// A promise to provide a result later.
 ///
 /// This is the provider API for `EventLoopFuture<Value>`. If you want to return an
@@ -995,28 +998,46 @@ extension EventLoopFuture {
     /// - Returns: A new `EventLoopFuture` with all the results of the provided futures.
     public static func whenAllComplete<InputValue>(_ futures: [EventLoopFuture<InputValue>],
                                                    eventLoop: EventLoop) -> EventLoopFuture<[Result<InputValue, Error>]> {
-        return eventLoop.makeSucceededFuture(result: [])
-            .flatMap { _ in
-                let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)
+        let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)
 
-                var remainingCount = futures.count
-                var results: [Result<InputValue, Error>?] = .init(repeating: nil, count: futures.count)
-
-                for (index, future) in futures.enumerated() {
-                    future
-                        .hopTo(eventLoop: eventLoop)
-                        ._whenComplete {
-                            results[index] = future.value!
-                            remainingCount -= 1
-
-                            guard remainingCount == 0 else { return CallbackList() }
-
-                            return promise._setValue(value: .success(results.compactMap({ $0 })))
-                        }
-                }
-
-                return promise.futureResult
+        if eventLoop.inEventLoop {
+            _whenAllComplete0(promise, futures, eventLoop: eventLoop)
+        } else {
+            eventLoop.execute {
+                _whenAllComplete0(promise, futures, eventLoop: eventLoop)
             }
+        }
+
+        return promise.futureResult
+    }
+
+    private static func _whenAllComplete0<InputValue>(_ promise: EventLoopPromise<[Result<InputValue, Error>]>,
+                                                      _ futures: [EventLoopFuture<InputValue>],
+                                                      eventLoop: EventLoop) {
+        eventLoop.assertInEventLoop()
+
+        var remainingCount = futures.count
+        var results: [Result<InputValue, Error>] = .init(repeating: .failure(OperationPlaceholderError()), count: futures.count)
+
+        // loop through the futures to chain callbacks to execute on the initiating event loop and grab their index
+        // in the "futures" to store their result in the same location
+        for (index, future) in futures.enumerated() {
+            future.hopTo(eventLoop: eventLoop)
+                .whenComplete { result in
+                    results[index] = result
+                    remainingCount -= 1
+
+                    guard remainingCount == 0 else { return }
+
+                    // verify that all operations have been completed
+                    assert(results.contains(where: {
+                        guard case let .failure(error) = $0 else { return false }
+                        return error is OperationPlaceholderError
+                    }) == false)
+
+                    promise.succeed(result: results)
+                }
+        }
     }
 }
 
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -983,6 +983,43 @@ extension EventLoopFuture {
     }
 }
 
+extension EventLoopFuture {
+    /// Returns a new `EventLoopFuture` that succeeds when all of the provided `EventLoopFuture`s complete.
+    /// The new `EventLoopFuture` will contain an array of results, maintaining ordering for each of the `EventLoopFuture`s.
+    ///
+    /// The returned `EventLoopFuture` always succeeds, regardless of any failures from the waiting futures.
+    ///
+    /// If it is desired to flatten them into a single `EventLoopFuture` that fails on the first `EventLoopFuture` failure,
+    /// use one of the `reduce` methods instead.
+    /// - Parameter futures: An array of homogenous `EventLoopFuture`s to gather results from.
+    /// - Returns: A new `EventLoopFuture` with all the results of the provided futures.
+    public static func whenAllComplete<InputValue>(_ futures: [EventLoopFuture<InputValue>],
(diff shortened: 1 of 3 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -957,6 +957,43 @@ extension EventLoopFuture {
...
+
+                            guard remainingCount == 0 else { return CallbackList() }
+
+                            return promise._setValue(value: .success(results.compactMap({ $0 })))
+                        }
+                }
+
+                return promise.futureResult
+            }
+    }
```
Lukasa: Rather than use the do-nothing future object so you can get access to `flatMap`, I propose you use `eventLoop.execute` instead. Specifically, let's pull out the body of the `flatMap` call into a separate function (call it `_whenAllComplete0`) and then you can do this:

```swift
let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)

if eventLoop.inEventLoop {
    _whenAllComplete0(promise, futures, eventLoop: eventLoop)
} else {
    eventLoop.execute {
        _whenAllComplete0(promise, futures, eventLoop: eventLoop)
    }
}

return promise.futureResult
```

That way we don't need the extra do-nothing future, and we keep the ability to avoid unnecessary thread-hops.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -957,6 +957,43 @@ extension EventLoopFuture {
...
+        return eventLoop.makeSucceededFuture(result: [])
+            .flatMap { _ in
+                let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)
+
+                var remainingCount = futures.count
+                var results: [Result<InputValue, Error>?] = .init(repeating: nil, count: futures.count)
+
+                for (index, future) in futures.enumerated() {
+                    future
+                        .hopTo(eventLoop: eventLoop)
```
Lukasa: This indentation doesn't match NIO style: please rewrite to `future.hopTo(eventLoop: eventLoop)`. The subsequent "new line chaining" is just fine though, you can keep that.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -957,6 +957,43 @@ extension EventLoopFuture {
...
+                        ._whenComplete {
+                            results[index] = future.value!
+                            remainingCount -= 1
+
+                            guard remainingCount == 0 else { return CallbackList() }
+
+                            return promise._setValue(value: .success(results.compactMap({ $0 })))
+                        }
```
Lukasa: This block uses a load of `EventLoopFuture` SPI: let's not. Per #184 I'd like to reduce the amount of internal SPI used in `EventLoopFuture` rather than increase it, and as we can write all of this without SPI, I think we should.

That means replacing `._whenComplete` with `whenComplete`, removing `future.value!` (as it'll no longer be necessary), and replacing `promise._setValue` with `promise.succeed`.

Tests/NIOTests/EventLoopFutureTest.swift
```diff
@@ -828,4 +828,38 @@ class EventLoopFutureTest : XCTestCase {
...
+            group.next().makeFailedFuture(error: EventLoopFutureTestError.example),
+            group.next().makeSucceededFuture(result: 5)
+        ], eventLoop: group.next()).wait()
+
+        XCTAssertEqual(try results[0].get(), 3)
+        XCTAssertThrowsError(try results[1].get())
+        XCTAssertEqual(try results[2].get(), 10)
+        XCTAssertThrowsError(try results[3].get())
+        XCTAssertEqual(try results[4].get(), 5)
+    }
```
Lukasa: Can you add a test here that validates that the returned Future doesn't succeed until all the Futures passed to it have completed?

Sources/NIO/EventLoopFuture.swift
```diff
@@ -957,6 +957,43 @@ extension EventLoopFuture {
...
+    /// - Parameter futures: An array of homogenous `EventLoopFuture`s to gather results from.
+    /// - Returns: A new `EventLoopFuture` with all the results of the provided futures.
+    public static func whenAllComplete<InputValue>(_ futures: [EventLoopFuture<InputValue>],
+                                                   eventLoop: EventLoop) -> EventLoopFuture<[Result<InputValue, Error>]> {
+        return eventLoop.makeSucceededFuture(result: [])
+            .flatMap { _ in
+                let promise = eventLoop.makePromise(of: [Result<InputValue, Error>].self)
+
+                var remainingCount = futures.count
+                var results: [Result<InputValue, Error>?] = .init(repeating: nil, count: futures.count)
```
Lukasa: Oh, yes, I meant to come back to this.

The `compactMap { $0 }` you need to perform at the bottom here is a bit awkward: it forces a new array allocation, and we happen to know analytically that there should never be nils in this array. So I think I'd propose rewriting this to remove the optionality, and instead initialising the array with a bunch of `Result`s in the `Error` state with an appropriate `Error`. That way the optionals are never needed.

Lukasa:
No. `andAll` as you propose it would be a fast-failure API. This is explicitly a *slow* failure API: it waits for all Futures to complete before returning. That's also why it always succeeds.

Lukasa:
To be clear, I'm fine having both APIs, but you can't build this one in terms of the one you propose.

Sources/NIO/EventLoopFuture.swift
Mordil: Oh, that's much better! This will probably come to me more naturally as I spend more time working in the mindset of making sure things are in the event loop, rather than working just off of attached callbacks on ELFS :)

Sources/NIO/EventLoopFuture.swift
Mordil: Thanks for the context - I was going off of what I saw in the file already. I'll make the updates to not use SPI.

Tests/NIOTests/EventLoopFutureTest.swift
Mordil: Will do!

Sources/NIO/EventLoopFuture.swift
Mordil: Quick question - what is the meaning behind the `0` suffix? I understand having a `_` prefix to label as SPI and remove ambiguities, but what additional meaning does `0` provide?

Lukasa: We use it as an internal style marker to indicate that the function is *not* thread safe. Almost all functions with `0` in their name begin with `eventLoop.assertInEventLoop()` in their body. I recommend doing the same, incidentally.

Sources/NIO/EventLoopFuture.swift
Mordil: Is there an existing data structure for EventLoopFuture errors, or should I create a new one?

Mordil:
@weissi @Lukasa I was planning to submit a PR after this one that changes `andAll` for that use case.

Users will then have a `whenAllSucceed` and a `whenAllComplete` method to choose how they want to treat the collection of `futures` being passed: As a mutually exclusive task list, or a related task list.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -957,6 +960,61 @@ extension EventLoopFuture {
...
+                    results[index] = result
+                    remainingCount -= 1
+
+                    guard remainingCount == 0 else { return }
+
+                    // verify that all operations have been completed
+                    assert(results.contains(where: {
+                        guard case let .failure(error) = $0 else { return false }
+                        return error is OperationPlaceholderError
+                    }) == false)
```
Lukasa: Minor style nit, but I prefer to `assert(!` instead of `assert( == false)`. Explicit equality operations against booleans just read weirdly to me.

Tests/NIOTests/EventLoopFutureTest.swift
```diff
@@ -828,4 +828,65 @@ class EventLoopFutureTest : XCTestCase {
...
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        var tally = 0
+        
+        var futures: [EventLoopFuture<Int>] = []
+        for index in 1...5 {
+            futures.append(
+                group.next().makeSucceededFuture(result: index).map {
+                    sleep(UInt32(1 / index))
```
Lukasa: There's no need to `sleep` here: all that does is slow down the test unnecessarily.

The better approach is to do something like this:

```swift
let promises = (0..<5).map { _ in group.next().makePromise(of: Int.self) }
let futures = promises.map { $0.futureResult }

var succeeded = false
var completedPromises = false

let mainFuture = EventLoopFuture.whenAllComplete(futures, eventLoop: group.next())
mainFuture.whenSuccess { result in
    XCTAssertTrue(completedPromises)
    XCTAssertFalse(succeeded)
    succeeded = true
}

XCTAssertFalse(succeeded)

// Complete first four promises.
for (index, promise) in promises.dropLast().enumerated() {
    promise.succeed(result: idx)
}

// Still hasn't succeeded.
XCTAssertFalse(succeeded)

// Complete last promise.
completedPromises = true
promises.last!.succeed(result: 4)

let result = try assertNoThrowWithValue(mainFuture.wait())
XCTAssertEqual(result, [0, 1, 2, 3, 4])
```

This test is much more deterministic, observes all the intermediate states, and doesn't need to sleep unnecessarily.

Sources/NIO/EventLoopFuture.swift
Mordil: Haha, I'm the other way, I find that `!results.contains(...)` to read weirdly - but I don't have to read this code as often as you :P

weissi:

weissi:
@Mordil oops this conflicts now, mind rebasing and fixing the merge conflicts? Then I think we’re pretty much good to merge

weissi:
And sorry, I’m probably the one causing the merge conflicts...

Mordil:
Should be all good - pending further changes to master :)

weissi:

weissi:
@mordil still some compile errors in the tests unfortunately 

```
NIOTLSTests' (4 sources)
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:870:26: error: 'makeFailedFuture(error:)' is deprecated: renamed to 'makeFailedFuture(_:)'
23:15:34             group.next().makeFailedFuture(error: EventLoopFutureTestError.example),
23:15:34                          ^
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:870:26: note: use 'makeFailedFuture(_:)' instead
23:15:34             group.next().makeFailedFuture(error: EventLoopFutureTestError.example),
23:15:34                          ^~~~~~~~~~~~~~~~ ~~~~~~~
23:15:34                          makeFailedFuture 
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:871:26: error: 'makeSucceededFuture(result:)' is deprecated: renamed to 'makeSucceededFuture(_:)'
23:15:34             group.next().makeSucceededFuture(result: true)
23:15:34                          ^
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:871:26: note: use 'makeSucceededFuture(_:)' instead
23:15:34             group.next().makeSucceededFuture(result: true)
23:15:34                          ^~~~~~~~~~~~~~~~~~~ ~~~~~~~~
23:15:34                          makeSucceededFuture 
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:883:26: error: 'makeSucceededFuture(result:)' is deprecated: renamed to 'makeSucceededFuture(_:)'
23:15:34             group.next().makeSucceededFuture(result: 3),
23:15:34                          ^
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:883:26: note: use 'makeSucceededFuture(_:)' instead
23:15:34             group.next().makeSucceededFuture(result: 3),
23:15:34                          ^~~~~~~~~~~~~~~~~~~ ~~~~~~~~
23:15:34                          makeSucceededFuture 
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:884:26: error: 'makeFailedFuture(error:)' is deprecated: renamed to 'makeFailedFuture(_:)'
23:15:34             group.next().makeFailedFuture(error: EventLoopFutureTestError.example),
23:15:34                          ^
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:884:26: note: use 'makeFailedFuture(_:)' instead
23:15:34             group.next().makeFailedFuture(error: EventLoopFutureTestError.example),
23:15:34                          ^~~~~~~~~~~~~~~~ ~~~~~~~
23:15:34                          makeFailedFuture 
23:15:34 /code/Tests/NIOTests/EventLoopFutureTest.swift:885:26: error: 'makeSucceededFuture(result:)' is deprecated: renamed to 
```

weissi:
@mordil sorry, indeed! We compile with warnings to errors, hence they showed up as errors. Thanks for fixing!!

weissi:

PR #776, opened 2019-01-24
remove ELP.succeed/fail's labels

weissi:
Motivation:

EventLoopPromise.succeed/fail has extraneous labels that don't add
anything but noise.

Modifications:

remove extraneous labels

Result:

less noise in the code

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -41,7 +41,7 @@ public struct Scheduled<T> {
     /// Whether this is successful depends on whether the execution of the task already begun.
     ///  This means that cancellation is not guaranteed.
     public func cancel() {
-        promise.fail(error: EventLoopError.cancelled)
+        promise.fail(EventLoopError.cancelled)
     }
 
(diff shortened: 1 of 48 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -395,6 +395,20 @@ extension EventLoop {
...
+    public func makeFailedFuture<T>(error: Error) -> EventLoopFuture<T> {
+        return self.makeFailedFuture(error)
+    }
+
+    @available(*, deprecated, renamed: "makeSucceededFuture(_:)")
+    public func makeSucceededFuture<Success>(result: Success) -> EventLoopFuture<Success> {
+        return self.makeSucceededFuture(result)
+    }
+
+}
```
Lukasa: Why aren't these in the shims module?

weissi: because the API they shim isn't NIO 1 API but API introduced in NIO 2... I filed #779 to remove them before 2.0.0

PR #777, opened 2019-01-24
rename ELF.mapIfError to ELF.recover

weissi:
Motivation:

The name `mapIfError` implied that you can transform the error value but
really that's never what it did. Instead, it allowed you to recover
from a failure.

Modifications:

- rename `mapIfError` to `recover`

Result:

cleaner method names


ktoso:
Sure about the `fromFailure` part? `recover` already implies there's something to recover from -- a failure. `recover { failure in ... }`

weissi:
thanks @ktoso  & @Lukasa , I like `recover` and prior art is 👌

ktoso:
Yay :)

PR #778, opened 2019-01-24
provide ELF.flatMapResult

weissi:
Motivation:

ELFs and the stdlib's Result should be as compatible as possible so a
flatMapResult function which lifts and combines a result into an
EventLoopFuture sounds like a good idea.

Modifications:

add `flatMapResult` which does what it says.

Result:

more interop between `Result` and `EventLoopFuture`.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -136,7 +136,7 @@ private struct CallbackList: ExpressibleByArrayLiteral {
 /// * If you just want to get a value back after running something on another thread,
 ///     use `EventLoopFuture<ResultType>.async()`
 /// * If you already have a value and need an `EventLoopFuture<>` object to plug into
-///     some other API, create an already-resolved object with `eventLoop.newSucceededFuture(result)`
+///     some other API, create an already-resolved object with `eventLoop.makeSucceededFuture(result)`
 ///     or `eventLoop.newFailedFuture(error:)`.
 ///
 /// - note: `EventLoopPromise` has reference semantics.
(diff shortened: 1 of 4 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -955,6 +955,32 @@ extension EventLoopFuture {
...
+    /// performs a simple data transformation that can potentially error.
+    ///
+    ///
+    /// - parameters:
+    ///     - body: Function that will receive the value of this `EventLoopFuture` and return
+    ///         a new value or error lifted into a new `EventLoopFuture`.
+    /// - returns: A future that will receive the eventual value.
+    public func flatMapResult<NewValue, SomeError: Error>(file: StaticString = #file,
+                                                          line: UInt = #line,
+                                                          _ body: @escaping (Value) -> Result<NewValue, SomeError>) -> EventLoopFuture<NewValue> {
```
Lukasa: You aren't using the `file` and `line` parameters here.

weissi: thanks! fixed

weissi:
sorry for forgetting the test, added 2 now

Sources/NIO/EventLoopFuture.swift
```diff
@@ -955,6 +955,32 @@ extension EventLoopFuture {
...
+        return self.flatMap { value in
```
Lukasa: `flatMap` also takes `file` and `line` arguments.

weissi: oh yes, thank you!!

PR #780, opened 2019-01-24
CI: use DEVELOPMENT-SNAPSHOT-2019-01-23-a

### LINKED ISSUE #726: track down the two extra allocations in Swift 5
in #725 I had to raise the allocation limits for Swift 5 because there's a regression in Swift 5. We need to track down the problem in Swift and then hopefully lower them again.

### LINKED ISSUE #727: track down extra allocation in bytebuffer_lots_of_rw
when the `bytebuffer_lots_of_rw` allows 2 allocations (creating a String and creating an Array). For some reason we see a regression in Swift 5 up to 3 allocations. That's odd because the String should actually fit into a small string so we'd expect one _fewer_ allocation not one more.

weissi:
Motivation:

Newer Swift versions are good and we should test them, this one threw a few curveballs at us:
- started using `posix_memalign` which needed fixing in our alloc tests (fixed in #782)
- the optimiser decided not to inline a function that needs to be inlined

Modifications
- `@inline(__always)` to help the optimiser
- update Swift version

Result:
- fixes #727
- fixes #726

### DIFF
--- a/Sources/NIOHTTP1/HTTPEncoder.swift
+++ b/Sources/NIOHTTP1/HTTPEncoder.swift
@@ -74,6 +74,9 @@ private func writeTrailers(wrapOutboundOut: (IOData) -> NIOAny, ctx: ChannelHand
     }
 }
 
+// starting about swift-5.0-DEVELOPMENT-SNAPSHOT-2019-01-20-a, this doesn't get automatically inlined, which costs
+// 2 extra allocations so we need to help the optimiser out.
+@inline(__always)
 private func writeHead(wrapOutboundOut: (IOData) -> NIOAny, writeStartLine: (inout ByteBuffer) -> Void, ctx: ChannelHandlerContext, headers: HTTPHeaders, promise: EventLoopPromise<Void>?) {
 
     var buffer = ctx.channel.allocator.buffer(capacity: 256)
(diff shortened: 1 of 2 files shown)


Lukasa:
Woop CI failure.

weissi:
Huh, negative amount of remaining allocations, will investigate 🤣

weissi:
yaay, @Lukasa the inliner threshold must have changed, but now it finally passes 🙌

Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -74,6 +74,9 @@ private func writeTrailers(wrapOutboundOut: (IOData) -> NIOAny, ctx: ChannelHand
     }
 }
 
+// starting about swift-5.0-DEVELOPMENT-SNAPSHOT-2019-01-20-a, this doesn't get automatically inlined, which costs
+// 2 extra allocations so we need to help the optimiser out.
+@inline(__always)
```
weissi: @eeckstein between swift-5.0-DEVELOPMENT-SNAPSHOT-2019-01-13-a and swift-5.0-DEVELOPMENT-SNAPSHOT-2019-01-20-a this suddenly didn't get inlined anymore without any code changes.

Is that expected? It was stable from at least Swift 4.0.1 until now and then suddenly last week it broke...

Issue #676, opened 2018-12-04
All Sequences should be Collections


Lukasa:
This issue was originally targeted to `HTTPHeaders`, but further searching through the code reveals other data structures (e.g. `Heap<T>` and `PriorityQueue<T>` as a quick first note) that depend on `Sequence`. `Sequence` is bad: all of these should be conformed to better protocols. At the very least `Collection`, but ideally some better leaf protocol.

PR #784, opened 2019-01-28
Drop HTTPHeaders Sequence conformance in favor of RandomAccessCollection

JovanMilenkovic:
Motivation:

Use better protocol instead of Sequence since it doesn't do much. Inspired by https://github.com/apple/swift-nio/issues/676

Modifications:

Conform HTTPHeaders to RandomAccessCollection

Result:

HTTPHeaders will have RandomAccessCollection functionalities.


JovanMilenkovic:
Sure, thanks for review. By "correctly establishing" the index, do you mean this commit should have traversed the heap to find it in order to be correct, instead of just providing a look into the underlying storage? Also, what do you think about the HTTPHeaders iterator, does it do anything meaningful? I'll push the removal as soon as I get some spare time.

Lukasa:
> By "correctly establishing" the index, do you mean this commit should have traversed the heap to find it in order to be correct, instead of just providing a look into the underlying storage?

Yes.

A (potentially unstated) requirement of conformance to `Collection` is that you should get the same iteration order from your iterator as you do by calling `index(after:)` in a loop and using the index `subscript`. This is not met by simply passing the indices through from the underlying storage.

Unfortunately, actually implementing this is *rough*. We spitballed a few ideas, but all were pretty severely limited. `Collection` requires the complexity of the index `subscript` to be `O(1)` if at all possible, and to achieve that you end up with the cost of `index(after:)` being `O(m log n)`, which is pretty damn gross. Additionally, iteration via this method would trigger repeated `CoW` operations, which is pretty terrible as well.

Ultimately, for the use-case we have those two are better off being `Sequence`.

> Also, what do you think about the HTTPHeaders iterator, does it do anything meaningful?

What do you mean by "do anything meaningful"?

JovanMilenkovic:
I meant that it still works if you delete it and use the default one. Maybe the documentation is considered useful but there’s no easy way of reaching iterator documentation.

Lukasa:
> I meant that it still works if you delete it and use the default one. Maybe the documentation is considered useful but there’s no easy way of reaching iterator documentation.

Sure, that's because manually providing an `Iterator` is not necessary when you conform to `Collection`. If you don't provide `makeIterator`, the `Collection` will give you a default implementation that provides you an `IndexingIterator` which uses `index(after:)` to provide the iteration.

A correct implementation of `Iterator` should always give identical results to that from iterating over `index(after:)`. Whether it's worth keeping a custom iterator is entirely down to whether the custom implementation provides a better performance story. In this case, we can probably get rid of it: the performance story is probably a bit better when using `IndexingIterator` as partial-iteration can avoid reifying the entire data structure.

JovanMilenkovic:
@Lukasa pushed an update that changes HTTPHeaders only. Thanks for all the in-depth info. The current patch doesn't, but maybe it should provide a convenient `Int` based subscript. Are there realistic use cases for that?

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -738,27 +738,35 @@ internal extension ByteBuffer {
...
-        fileprivate init(headerParts: Array<(String, String)>.Iterator) {
-            self.headerParts = headerParts
-        }
 
-        public mutating func next() -> Element? {
-            return headerParts.next()
+extension HTTPHeaders: RandomAccessCollection {
+    public typealias Element = (name: String, value: String)
+    public struct Index: Comparable {
+        let base: Int
```
Lukasa: Let's make this `fileprivate let` to avoid the rest of the project being able to see it.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -738,27 +738,35 @@ internal extension ByteBuffer {
...
+    public var endIndex: Index {
+        return Index(base: self.headers.endIndex)
+    }
+    
+    public func index(before i: Index) -> Index {
+        return Index(base: i.base - 1)
+    }
+    
+    public func index(after i: Index) -> Index {
+        return Index(base: i.base + 1)
```
Lukasa: So there's nothing wrong with these functions, but I'd like to make them a bit more future-proof.

Right now what we're doing is delegating to the index of our backing storage, which is `Array<HTTPHeader>`. I think we may be best served by explicitly stating this, and then moving to use the correct functions for manipulating those indices.

To do this, we can redefine `Index` such that `base` is of type `Array<HTTPHeaders>.Index`. This is a no-op change, as that's just a typealias for `Int`, but it more correctly indicates the intended design (and makes us resilient against changing the backing store).

Once we've done that, we can change `index(before:)` and `index(after:)` to just call those functions on `Array`, rather than hard-code knowledge about how `Array`'s indexing works.

Does that make sense?

JovanMilenkovic: Yes, that's a nice one. Just pushed that update.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -738,27 +738,35 @@ internal extension ByteBuffer {
...
-
-        fileprivate init(headerParts: Array<(String, String)>.Iterator) {
-            self.headerParts = headerParts
-        }
 
-        public mutating func next() -> Element? {
-            return headerParts.next()
+extension HTTPHeaders: RandomAccessCollection {
+    public typealias Element = (name: String, value: String)
+    public struct Index: Comparable {
```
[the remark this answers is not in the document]
JovanMilenkovic: Should be all good now

JovanMilenkovic:
I've edited the commit message and updated this PR description to reflect what this now actually does(only changes HTTPHeaders).

Lukasa:

PR #794, opened 2019-01-31
Make use of dropFirst(_:) in example code

slashmo:
Motivation:

While getting into NIO through the examples I stumbled
across some chained `dropFirst()` calls, which can be
simplified by using `dropFirst(_:)` instead.

Modifications:

I did a quick project-wide search of chained `dropFirst()`
usage and modified every line to use `dropFirst(_:)` instead.

Result:

This will only affect the readability of the code,
not it's functionality.


Lukasa:
Thanks a bunch!

Lukasa:

PR #804, opened 2019-02-02
Add `whenAllSucceed` method to ELF

Mordil:
- Add: `whenAllSucceed(_:on:)` method
- Rename: `whenAllComplete(_:eventLoop:)` to `whenAllComplete(_:on:)`

This introduces the "fail fast" companion method to `whenAllComplete` that collects all fulfilled values to return to the user.


Lukasa:

Sources/NIO/EventLoopFuture.swift
```diff
@@ -997,7 +1011,7 @@ extension EventLoopFuture {
     /// - Parameter futures: An array of homogenous `EventLoopFuture`s to gather results from.
     /// - Returns: A new `EventLoopFuture` with all the results of the provided futures.
     public static func whenAllComplete(_ futures: [EventLoopFuture<Value>],
-                                       eventLoop: EventLoop) -> EventLoopFuture<[Result<Value, Error>]> {
+                                       on eventLoop: EventLoop) -> EventLoopFuture<[Result<Value, Error>]> {
```
Lukasa: Sorry, just noticed this: this label change needs a note in the migration notes in `docs` and also a helper in `_NIO1APIShims`?

Mordil: `whenAllComplete` doesn't exist in NIO 1 - this was introduced directly to NIO 2 because of its use of `Result<Value, Error>`

See #773

Lukasa: Ah, you're quite right.

Lukasa:

PR #807, opened 2019-02-06
Avoid race condition in Posix.accept() on macOS

gwynne:
Fix a race condition when remote closes connection after `accept()` but before `fcntl()` on macOS.

### Motivation:

A race condition occurs when a socket is closed from the remote end after `accept()` returns it, but before `fcntl(F_SETFL, F_SETNOSIGPIPE)` is invoked. This causes the `fcntl()` call to fail with `EINVAL` ("The socket has been shut down."). This currently triggers an assertion failure. As the race depends on the remote and can be easily demonstrated (using `nmap` is a good way to set it off), the assertion is semantically invalid (`EINVAL` is not programmer error). Additionally, errors reported by `fcntl()` in general are being ignored when not asserted.

### Modifications:

The race is avoided by treating `EINVAL` as a normal error while retaining the existing behavior for all other error cases. Both uses of `fcntl()` to disable `SIGPIPE` now properly report errors.

### Result:

`Posix.accept()` and `Posix.socket()` may now potentially throw additional errors. It is no longer possible to crash a debug server by running `nmap` against it.

### DIFF
--- a/Sources/NIO/System.swift
+++ b/Sources/NIO/System.swift
@@ -255,6 +255,7 @@ internal enum Posix {
     }
 
     @inline(never)
+    @discardableResult
     // TODO: Allow varargs
     public static func fcntl(descriptor: CInt, command: CInt, value: CInt) throws -> CInt {
         return try wrapSyscall {
@@ -272,9 +273,12 @@ internal enum Posix {
                 _ = unsafeBitCast(Glibc.signal(SIGPIPE, SIG_IGN) as sighandler_t?, to: Int.self)
             #else
                 if fd != -1 {
-                    let ret = try? Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
-                    assert(ret == .some(0),
-                           "unexpectedly, fcntl(\(fd), F_SETFL, F_SETNOSIGPIPE) returned \(ret.debugDescription)")
+                    do {
+                        try Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
+                    } catch {
+                        _ = sysClose(fd) // don't care about failure here
+                        throw error
+                    }
                 }
             #endif
             return fd
@@ -311,9 +315,12 @@ internal enum Posix {
 
             #if !os(Linux)
                 if fd != -1 {
-                    let ret = try? Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
-                    assert(ret == .some(0),
-                           "unexpectedly, fcntl(\(fd), F_SETFL, F_SETNOSIGPIPE) returned \(ret.debugDescription)")
+                    do {
+                        try Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
+                    } catch {
+                        _ = sysClose(fd) // don't care about failure here
+                        throw error
+                    }
                 }
             #endif
             return fd
--- a/Tests/NIOTests/SocketChannelTest+XCTest.swift
+++ b/Tests/NIOTests/SocketChannelTest+XCTest.swift
@@ -45,6 +45,7 @@ extension SocketChannelTest {
                 ("testPendingConnectNotificationOrder", testPendingConnectNotificationOrder),
                 ("testLocalAndRemoteAddressNotNilInChannelInactiveAndHandlerRemoved", testLocalAndRemoteAddressNotNilInChannelInactiveAndHandlerRemoved),
                 ("testSocketFlagNONBLOCKWorks", testSocketFlagNONBLOCKWorks),
+                ("testInstantTCPConnectionResetThrowsError", testInstantTCPConnectionResetThrowsError),
            ]
    }
 }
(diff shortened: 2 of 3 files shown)


gwynne:
I'm crossing my fingers this is a sufficiently minimal change to be able to go into a patch version; it was already possible (if rare) for `EINVAL` to be thrown from this code, so the API surface is unchanged.

Sources/NIO/System.swift
```diff
@@ -310,10 +310,22 @@ internal enum Posix {
             let fd = sysAccept(descriptor, addr, len)
 
             #if !os(Linux)
-                if fd != -1 {
-                    let ret = try? Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
-                    assert(ret == .some(0),
```
Lukasa: Is there any reason this assertion went away? Given that the error was throwing, I'd assume that the assertion could stay there.

Lukasa:
> 1. As of the time of this writing, I have yet to figure out a way to deliberately trigger this particular race from an XCTest.

Yeah, this one is rough. Your best option is likely to be a probabilistic test: try making 100 connections and close them as rapidly as possible, skipping the NIO machinery (that is, make the underlying syscalls directly rather than using channels). Ultimately the timing window here *is* very small, and I don't think we plan to add a shim point where you can force it to be hit, so a probabilistic test is likely the best we can do.

> whether _all_ errors should be thrown unconditionally in Release.

Yeah, good question. EINVAL is a tricky one here because it's the most likely error to be thrown by an order of magnitude in cases where the FD does not accept this option. I'm a bit tempted to say we shouldn't be silently swallowing these errors: @normanmaurer, what do you think? Why did we silence these to begin with?

Lukasa:

Sources/NIO/System.swift
gwynne: The assertion is redundant once `try?` was replaced with `try` - `fcntl()` with `F_SETFL` will only ever return `0` (success) or `-1` (error). In the old logic, errors were being eaten anyway (with the side effect that you could only see _what_ error occurred by inspecting _errno_ in the debugger) so it was just "assert that no error occurred". With the new logic, any error results in an `IOError` being thrown, so the only remaining return value possible is `0`, and the assertion is moved into the `catch` block so we can treat `EINTR` as non-crashing.

I suppose there's an argument that you could still assert that the non-error return value really was zero, but at that point you're checking for either the postcondition of `wrapSyscall {}` (which should be unit-tested directly, not asserted by callers) or the documented behavior of `fcntl(F_SETFL)`, which is an endless rabbit hole anyway because if the syscalls start returning nonsense results you probably have a larger issue than a failed assertion 🙂 (Not to  mention this particular code only runs on macOS anyway and we really ought to know the behavior there!)

Sources/NIO/System.swift
```diff
@@ -310,10 +310,22 @@ internal enum Posix {
...
+                } catch let err as IOError {
+                    // Per setsockopt(SO_NOSIGPIPE), fcntl(F_SETNOSIGPIPE) can return EINVAL if
+                    // the remote closed the connection after accept() returned. This is not
+                    // programmer error and thus should not be covered by an assertion.
+                    assert(err.errnoCode == EINVAL,
+                           "unexpectedly, fcntl(\(fd), F_SETFL, F_SETNOSIGPIPE) returned \(err)")
+                    
+                    // Throw on EINVAL but silently ignore other errors when assertions are disabled.
+                    if err.errnoCode == EINVAL {
+                        throw err
```
normanmaurer: After talking a bit with @Lukasa we both agreed that we should just always throw and not "discard" errors in some cases. I really dont remember why we even ignored these at the first place.

gwynne: Done in latest commits.

Sources/NIO/System.swift
```diff
@@ -311,9 +311,8 @@ internal enum Posix {
 
             #if !os(Linux)
                 if fd != -1 {
-                    let ret = try? Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
-                    assert(ret == .some(0),
-                           "unexpectedly, fcntl(\(fd), F_SETFL, F_SETNOSIGPIPE) returned \(ret.debugDescription)")
+                    // Result is uninteresting, fcntl() only returns nonzero on error.
+                    _ = try Posix.fcntl(descriptor: fd, command: F_SETNOSIGPIPE, value: 1)
```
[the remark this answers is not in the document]
gwynne: Done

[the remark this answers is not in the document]
gwynne: Fixed and applied to `Posix.socket()` as well

gwynne:
Whoopsy, forgot about the template. Will do :)

weissi:
nice one, thanks @gwynne

gwynne:
@weissi Can't take all the credit on this one, @adtrevor first found and repro'd the issue and `tholo` and @Lukasa helped a lot in figuring out a test ❤️

weissi:
awesome effort, thanks @adtrevor , @gwynne & @Lukasa :)

adtrevor:
Thank you everyone! 😃

PR #816, opened 2019-02-12
CircularBuffer: speedup removeAll

weissi:
Motivation:

Previously CircularBuffer.removeAll would always allocate (a new backing
store), even when `keepingCapacity: true`.

Modifications:

Don't make CircularBuffer allocate unnecessarily

Result:

faster CircularBuffer.removeAll

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -155,9 +155,17 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
 
     /// Removes all members from the circular buffer whist keeping the capacity.
     public mutating func removeAll(keepingCapacity: Bool = false) {
+        if keepingCapacity {
+            for index in 0..<self.count {
+                self.buffer[self.bufferIndex(ofIndex: index)] = nil
+            }
+        } else {
+            self.buffer.removeAll(keepingCapacity: false)
+            self.buffer.append(nil)
+        }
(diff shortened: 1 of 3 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -155,9 +155,17 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
 
     /// Removes all members from the circular buffer whist keeping the capacity.
     public mutating func removeAll(keepingCapacity: Bool = false) {
+        if keepingCapacity {
+            for index in 0..<self.count {
+                self.buffer[self.bufferIndex(ofIndex: index)] = nil
+            }
```
Lukasa: Is there any reason we aren't just using `self.buffer.removeAll(keepingCapacity: true)`?

weissi: @Lukasa yes, the ring buffer uses `self.buffer` as a pretty much constantly sized backing array. So an empty ring buffer with ring capacity 4 would look like this:

    [nil, nil, nil, nil]

if I were to call `self.buffer.removeAll(keepingCapacity: true)`, `self.buffer` were

    []

with some capacity reserved at the end of `self.buffer` for future expansions.

Basically it's `count` vs `capacity` in the backing store. Does that make sense?

Lukasa: Yup, makes sense thanks.

PR #820, opened 2019-02-12
rename Channel._unsafe to Channel._channelCore

weissi:
Motivation:

_unsafe is non-descriptive

Modifications:

- rename Channel._unsafe to Channel._channelCore

Result:

clearer API


Lukasa:
_ChannelCore *is* public API: implementing it is necessary to implement a new Channel.

weissi:
@Lukasa yeah, it's implementor's public API but not for users. But I think you're right. Maybe we should leave the type name alone (ie. `ChannelCore`) and just rename the property from `_unsafe` to `_channelCore`. What do you think?

Lukasa:
I think that's a good plan. 👍

weissi:
@Lukasa updated, thanks

PR #823, opened 2019-02-13
require ChannelOptions to be Equatable

### LINKED ISSUE #598: make ChannelOptions Equatable
`ChannelOptions` should be equatable, we should do this for 1.10.0. This patch should do it

```swift
diff --git a/Sources/NIO/Bootstrap.swift b/Sources/NIO/Bootstrap.swift
index 43eefac..4adef67 100644
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -119,21 +119,45 @@ public final class ServerBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
+    @available(*, deprecated, message: "non-Equatable ChannelOptions are deprecated")
     public func serverChannelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
         serverChannelOptions.put(key: option, value: value)
         return self
     }
 
+    /// Specifies a `ChannelOption` to be applied to the `ServerSocketChannel`.
+    ///
+    /// - note: To specify options for the accepted `SocketChannel`s, look at `ServerBootstrap.childChannelOption`.
+    ///
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    public func serverChannelOption<T: ChannelOption & Equatable>(_ option: T, value: T.OptionType) -> Self {
+        serverChannelOptions.put(key: option, value: value)
+        return self
+    }
+
     /// Specifies a `ChannelOption` to be applied to the accepted `SocketChannel`s.
     ///
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
+    @available(*, deprecated, message: "non-Equatable ChannelOptions are deprecated")
     public func childChannelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
         childChannelOptions.put(key: option, value: value)
         return self
     }
 
+    /// Specifies a `ChannelOption` to be applied to the accepted `SocketChannel`s.
+    ///
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    public func childChannelOption<T: ChannelOption & Equatable>(_ option: T, value: T.OptionType) -> Self {
+        childChannelOptions.put(key: option, value: value)
+        return self
+    }
+
     /// Bind the `ServerSocketChannel` to `host` and `port`.
     ///
     /// - parameters:
@@ -385,11 +409,22 @@ public final class ClientBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
+    @available(*, deprecated, message: "non-Equatable ChannelOptions are deprecated")
     public func channelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
         channelOptions.put(key: option, value: value)
         return self
     }
 
+    /// Specifies a `ChannelOption` to be applied to the `SocketChannel`.
+    ///
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    public func channelOption<T: ChannelOption & Equatable>(_ option: T, value: T.OptionType) -> Self {
+        channelOptions.put(key: option, value: value)
+        return self
+    }
+
     /// Specifies a timeout to apply to a connection attempt.
     //
     /// - parameters:
@@ -581,11 +616,22 @@ public final class DatagramBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
+    @available(*, deprecated, message: "non-Equatable ChannelOptions are deprecated")
     public func channelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
         channelOptions.put(key: option, value: value)
         return self
     }
 
+    /// Specifies a `ChannelOption` to be applied to the `DatagramChannel`.
+    ///
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    public func channelOption<T: ChannelOption & Equatable>(_ option: T, value: T.OptionType) -> Self {
+        channelOptions.put(key: option, value: value)
+        return self
+    }
+
     /// Use the existing bound socket file descriptor.
     ///
     /// - parameters:
diff --git a/Sources/NIO/ChannelOption.swift b/Sources/NIO/ChannelOption.swift
index 1b6bdc5..64b98b3 100644
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -48,7 +48,7 @@ public typealias SocketOptionName = Int32
 /// `SocketOption` allows to specify configuration settings that are directly applied to the underlying socket file descriptor.
 ///
 /// Valid options are typically found in the various man pages like `man 4 tcp`.
-public enum SocketOption: ChannelOption {
+public enum SocketOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = (SocketOptionLevel, SocketOptionName)
 
     public typealias OptionType = (SocketOptionValue)
@@ -70,58 +70,107 @@ public enum SocketOption: ChannelOption {
             return (level, name)
         }
     }
+
+    public static func == (lhs: SocketOption, rhs: SocketOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(let lLevel, let lName), .const(let rLevel, let rName)):
+            return lLevel == rLevel && lName == rName
+        }
+    }
 }
 
 /// `AllocatorOption` allows to specify the `ByteBufferAllocator` to use.
-public enum AllocatorOption: ChannelOption {
+public enum AllocatorOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = ByteBufferAllocator
 
     case const(())
+
+    public static func == (lhs: AllocatorOption, rhs: AllocatorOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `RecvAllocatorOption` allows to specify the `RecvByteBufferAllocator` to use.
-public enum RecvAllocatorOption: ChannelOption {
+public enum RecvAllocatorOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = RecvByteBufferAllocator
 
     case const(())
+
+    public static func == (lhs: RecvAllocatorOption, rhs: RecvAllocatorOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `AutoReadOption` allows to configure if a `Channel` should automatically call `Channel.read` again once all data was read from the transport or
 /// if the user is responsible to call `Channel.read` manually.
-public enum AutoReadOption: ChannelOption {
+public enum AutoReadOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = Bool
 
     case const(())
+
+    public static func == (lhs: AutoReadOption, rhs: AutoReadOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `WriteSpinOption` allows users to configure the number of repetitions of a only partially successful write call before considering the `Channel` not writable.
 /// Setting this option to `0` means that we only issue one write call and if that call does not write all the bytes,
 /// we consider the `Channel` not writable.
-public enum WriteSpinOption: ChannelOption {
+public enum WriteSpinOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = UInt
 
     case const(())
+
+    public static func == (lhs: WriteSpinOption, rhs: WriteSpinOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `MaxMessagesPerReadOption` allows to configure the maximum number of read calls to the underlying transport are performed before wait again until
 /// there is more to read and be notified.
-public enum MaxMessagesPerReadOption: ChannelOption {
+public enum MaxMessagesPerReadOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = UInt
 
     case const(())
+
+    public static func == (lhs: MaxMessagesPerReadOption, rhs: MaxMessagesPerReadOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `BacklogOption` allows to configure the `backlog` value as specified in `man 2 listen`. This is only useful for `ServerSocketChannel`s.
-public enum BacklogOption: ChannelOption {
+public enum BacklogOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = Int32
 
     case const(())
+
+    public static func == (lhs: BacklogOption, rhs: BacklogOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// The watermark used to detect when `Channel.isWritable` returns `true` or `false`.
@@ -158,20 +207,34 @@ public struct WriteBufferWaterMark {
 /// `Channel.isWritable` will return `false`. Once we were able to write some data out of the outbound buffer and the amount of bytes queued
 /// falls below `WriteBufferWaterMark.low` the `Channel` will become writable again. Once this happens `Channel.writable` will return
 /// `true` again. These writability changes are also propagated through the `ChannelPipeline` and so can be intercepted via `ChannelInboundHandler.channelWritabilityChanged`.
-public enum WriteBufferWaterMarkOption: ChannelOption {
+public enum WriteBufferWaterMarkOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = WriteBufferWaterMark
 
     case const(())
+
+    public static func == (lhs: WriteBufferWaterMarkOption, rhs: WriteBufferWaterMarkOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `ConnectTimeoutOption` allows to configure the `TimeAmount` after which a connect will fail if it was not established in the meantime. May be
 /// `nil`, in which case the connection attempt will never time out.
-public enum ConnectTimeoutOption: ChannelOption {
+public enum ConnectTimeoutOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = TimeAmount?
 
     case const(())
+
+    public static func == (lhs: ConnectTimeoutOption, rhs: ConnectTimeoutOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// `AllowRemoteHalfClosureOption` allows users to configure whether the `Channel` will close itself when its remote
@@ -179,11 +242,18 @@ public enum ConnectTimeoutOption: ChannelOption {
 /// will be closed automatically if the remote peer shuts down its send stream. If set to true, the `Channel` will
 /// not be closed: instead, a `ChannelEvent.inboundClosed` user event will be sent on the `ChannelPipeline`,
 /// and no more data will be received.
-public enum AllowRemoteHalfClosureOption: ChannelOption {
+public enum AllowRemoteHalfClosureOption: ChannelOption, Equatable {
     public typealias AssociatedValueType = ()
     public typealias OptionType = Bool
 
     case const(())
+
+    public static func == (lhs: AllowRemoteHalfClosureOption, rhs: AllowRemoteHalfClosureOption) -> Bool {
+        switch (lhs, rhs) {
+        case (.const(()), .const(())):
+            return true
+        }
+    }
 }
 
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
@@ -2697,12 +2705,3 @@ fileprivate class VerifyConnectionFailureHandler: ChannelInboundHandler {
         ctx.fireChannelUnregistered()
     }
 }
-
-extension SocketOption: Equatable {
-    public static func == (lhs: SocketOption, rhs: SocketOption) -> Bool {
-        switch (lhs, rhs) {
-        case (.const(let lLevel, let lName), .const(let rLevel, let rName)):
-            return lLevel == rLevel && lName == rName
-        }
-    }
-}
```

weissi:
Motivation:

ChannelOptions should've always been Equatable and so far we've hacked
around them not being Equatable when we wanted to compare them.

Modifications:

- make all ChannelOptions Equtable

Result:

- ChannelOption comparison actually works
- fixes #598

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -59,8 +59,10 @@ public final class ServerBootstrap {
     private let childGroup: EventLoopGroup
     private var serverChannelInit: ((Channel) -> EventLoopFuture<Void>)?
     private var childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
-    private var serverChannelOptions = ChannelOptionStorage()
-    private var childChannelOptions = ChannelOptionStorage()
+    @usableFromInline
+    internal var _serverChannelOptions = ChannelOptionStorage()
+    @usableFromInline
+    internal var _childChannelOptions = ChannelOptionStorage()
 
     /// Create a `ServerBootstrap` for the `EventLoopGroup` `group`.
     ///
@@ -119,8 +121,9 @@ public final class ServerBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
-    public func serverChannelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
-        serverChannelOptions.put(key: option, value: value)
+    @inlinable
+    public func serverChannelOption<Option: ChannelOption>(_ option: Option, value: Option.OptionType) -> Self {
+        self._serverChannelOptions.put(key: option, value: value)
         return self
     }
 
@@ -129,8 +132,9 @@ public final class ServerBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
-    public func childChannelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
-        childChannelOptions.put(key: option, value: value)
+    @inlinable
+    public func childChannelOption<Option: ChannelOption>(_ option: Option, value: Option.OptionType) -> Self {
+        self._childChannelOptions.put(key: option, value: value)
         return self
     }
 
@@ -201,10 +205,10 @@ public final class ServerBootstrap {
     private func bind0(makeServerChannel: (_ eventLoop: SelectableEventLoop, _ childGroup: EventLoopGroup) throws -> ServerSocketChannel, _ register: @escaping (EventLoop, ServerSocketChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
         let eventLoop = self.group.next()
         let childEventLoopGroup = self.childGroup
-        let serverChannelOptions = self.serverChannelOptions
+        let serverChannelOptions = self._serverChannelOptions
         let serverChannelInit = self.serverChannelInit ?? { _ in eventLoop.makeSucceededFuture(()) }
         let childChannelInit = self.childChannelInit
-        let childChannelOptions = self.childChannelOptions
+        let childChannelOptions = self._childChannelOptions
 
         let serverChannel: ServerSocketChannel
         do {
@@ -346,7 +350,8 @@ public final class ClientBootstrap {
 
     private let group: EventLoopGroup
     private var channelInitializer: ((Channel) -> EventLoopFuture<Void>)?
-    private var channelOptions = ChannelOptionStorage()
+    @usableFromInline
+    internal var _channelOptions = ChannelOptionStorage()
     private var connectTimeout: TimeAmount = TimeAmount.seconds(10)
     private var resolver: Resolver?
 
@@ -385,8 +390,9 @@ public final class ClientBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
-    public func channelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
-        channelOptions.put(key: option, value: value)
+    @inlinable
+    public func channelOption<Option: ChannelOption>(_ option: Option, value: Option.OptionType) -> Self {
+        self._channelOptions.put(key: option, value: value)
         return self
     }
 
@@ -477,7 +483,7 @@ public final class ClientBootstrap {
         }
 
         return channelInitializer(channel).flatMap {
-            self.channelOptions.applyAll(channel: channel)
+            self._channelOptions.applyAll(channel: channel)
         }.flatMap {
             let promise = eventLoop.makePromise(of: Void.self)
             channel.registerAlreadyConfigured0(promise: promise)
@@ -494,7 +500,7 @@ public final class ClientBootstrap {
                          protocolFamily: Int32,
                          _ body: @escaping (Channel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
         let channelInitializer = self.channelInitializer ?? { _ in eventLoop.makeSucceededFuture(()) }
-        let channelOptions = self.channelOptions
+        let channelOptions = self._channelOptions
 
         let promise = eventLoop.makePromise(of: Channel.self)
         let channel: SocketChannel
@@ -556,7 +562,8 @@ public final class DatagramBootstrap {
 
     private let group: EventLoopGroup
     private var channelInitializer: ((Channel) -> EventLoopFuture<Void>)?
-    private var channelOptions = ChannelOptionStorage()
+    @usableFromInline
+    internal var _channelOptions = ChannelOptionStorage()
 
     /// Create a `DatagramBootstrap` on the `EventLoopGroup` `group`.
     ///
@@ -581,8 +588,9 @@ public final class DatagramBootstrap {
     /// - parameters:
     ///     - option: The option to be applied.
     ///     - value: The value for the option.
-    public func channelOption<T: ChannelOption>(_ option: T, value: T.OptionType) -> Self {
-        channelOptions.put(key: option, value: value)
+    @inlinable
+    public func channelOption<Option: ChannelOption>(_ option: Option, value: Option.OptionType) -> Self {
+        self._channelOptions.put(key: option, value: value)
         return self
     }
 
@@ -651,7 +659,7 @@ public final class DatagramBootstrap {
     private func bind0(makeChannel: (_ eventLoop: SelectableEventLoop) throws -> DatagramChannel, _ registerAndBind: @escaping (EventLoop, DatagramChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
         let eventLoop = self.group.next()
         let channelInitializer = self.channelInitializer ?? { _ in eventLoop.makeSucceededFuture(()) }
-        let channelOptions = self.channelOptions
+        let channelOptions = self._channelOptions
 
         let channel: DatagramChannel
         do {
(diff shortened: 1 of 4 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -675,27 +675,8 @@ public final class DatagramBootstrap {
 /* for tests */ internal struct ChannelOptionStorage {
```
Lukasa: Here's a question: should this object be public? It's annoying as hell to have to reimplement this in other NIO channel implementations, especially as it's quite fussy.

Lukasa: I'm happy for us to just file an issue to make it public and document it and clean it up, rather than do it in this PR, of course.

Sources/NIO/ChannelOption.swift
```diff
@@ -78,6 +85,13 @@ public enum AllocatorOption: ChannelOption {
     public typealias OptionType = ByteBufferAllocator
 
     case const(())
+
+    public static func == (lhs: AllocatorOption, rhs: AllocatorOption) -> Bool {
```
Lukasa: Can these trivial ones not be synthesised?

weissi: @Lukasa unfortunately it doesn't do it, even when explicitly adding `, Equatable`. maybe the associated types?

Sources/NIO/Bootstrap.swift
weissi: great idea, adding it as `ChannelOptions.Storage`, nicely nested too

weissi: @Lukasa oh sorry, I put it in this PR, okay with that or split?

Sources/NIO/ChannelOption.swift
Lukasa: Hrm. @lorentey do you have any idea why this wouldn't synthesise?

Sources/NIO/Bootstrap.swift
Lukasa: Sure, it's fine here.

Sources/NIO/ChannelOption.swift
weissi: it's https://bugs.swift.org/browse/SR-9925

lorentey: Yes, sadly tuple types (including `Void`) aren't `Equatable`/`Hashable` yet. 

One workaround is to replace `()` with a nominal unit type. A dummy empty struct or single-element enum can be trivially `Equatable`/`Hashable`, and it would allow synthesis for `AllocatorOption`, too.

Sources/NIO/Bootstrap.swift
weissi: I've split it in the end.

Sources/NIO/ChannelOption.swift
weissi: @lorentey thanks! No idea why we actually has `case const(())`, maybe that was some workaround for some pre-Swift 4 bug? Who knows, your suggestion works now which is all that matters. Thank you!

PR #824, opened 2019-02-13
Accept dictionary literals for HTTPHeaders

johnno1962:
Just an idea - be able to provide dictionary literals where an HTTPHeader struct is expected. I could add the conformance in an extension if you prefer.

### Motivation:

A convenience/shorthand  for developers.

### Modifications:

Add conformance to ExpressibleByDictionaryLiteral along with implementation.

### Result:

The following is would then be legal in Vapor:
```
        return HTTPResponse(status: .ok, headers: ["Content-Type": "text/html"],
                            body: "<table border=1><tr><td><i>Hello</i>, plugin!")
```

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -430,7 +430,7 @@ private extension UInt8 {
 /// field when needed. It also supports recomposing headers to a maximally joined
 /// or split representation, such that header fields that are able to be repeated
 /// can be represented appropriately.
-public struct HTTPHeaders: CustomStringConvertible {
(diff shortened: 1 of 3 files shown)


johnno1962:
Maybe you can close this as it is already in https://github.com/vapor/http/blob/master/Sources/HTTPKit/Message/HTTPHeaders.swift#L73 so it's not such a daft idea. Be good if it made it up stream though.

johnno1962:
I’ve pushed an amend with a test. The test is… untested.

weissi:

johnno1962:
Thanks @weissi!

Issue #682, opened 2018-12-07
Provide opaque indices for CircularBuffer.


Lukasa:
When we made `CircularBuffer` a `Collection`, we gave it an `Int` `Index` and kept it zero-indexed. In this sense it behaved like `Array`.

After some conversations with @lorentey I think I've been convinced that we should use an opaque index type. The reasoning for it is this:

1. In general, the opaque index will improve the performance of indexing operations, as it will contain the index into the backing buffer. This will avoid the addition and modulo operation we need to perform today. The performance diff here is minor though, and isn't the main reason to do it.
2. Treating the index as an `Integer` means we invalidate the `Index` in situations where we do not need to. For example, if you have the index of a value in the circular buffer today, any operation on it other than `subscript { set }` will invalidate that index: it will now be pointing at a different element. This is a frustrating behaviour: it would be nicer to be able to document a set of operations on `CircularBuffer` that do *not* invalidate indices.
3. It's out-of-step with best-practices around `Collection` indexing. Acknowledging `Array` to be the odd-one-out here for both historical and cultural reasons, in general `Collection` types in Swift *should* define their `Index` as an opaque type, to avoid users performing unexpected operations on them.

My proposal is that we should move to the following `Index` definition:

```swift
struct Index {
    var baseIndex: Int  // Array.Index
    var age: UInt32
}
```

The use of a modification counter allows us to invalidate indices when an operation is performed that would break them, but keep them the same when they do not. It also allows indices to be safe to use across CoW copies of the data structure, again assuming that nothing has happened that invalidates the indices.

The operations that would bump `age` are:

1. Reallocating the underlying storage.
2. Invoking `replaceSubrange` (including its variants such as removing elements, or inserting new ones anywhere in the array).

Things that would *not* bump `age`:

1. `prepend()` or `append()` where the underlying storage is not reallocated.
2. `removeLast(_)`
3. Any other `Collection` method we can see that shouldn't do this
4. `subscript { set }`

Of course, integer indexing is useful, so we should give people a useful spelling. We can do this by implementing this subscript:

```swift
public subscript(offset: Int) -> E {
    get {
        let index = self.index(self.startIndex, offsetBy: offset)
        return self[index]
    }
    set {
        let index = self.index(self.startIndex, offsetBy: offset)
        self[index] = newValue
    }
}
```

This will allow users to continue to perform integer-based indexing with the same semantics as today, while improving the general conformance to the Swift collection protocols.

Thoughts?

[...]

PR #826, opened 2019-02-13
opaque indices for (Marked)CircularBuffer

weissi:
Motivation:

Opaque indices are a good idea for collections, (Marked)CircularBuffer
should use those too.

Modifications:

Cheapest possible implementation for opaque indices, could be much
improved. This is the cheapest possible implementation for opaque
indices and this could be much improved later, see #682 but the
improvements won't be SemVer major :).

Result:

first stab at #682

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -23,9 +23,9 @@ protocol AppendableCollection: Collection {
 /// An automatically expanding ring buffer implementation backed by a `ContiguousArray`. Even though this implementation
 /// will automatically expand if more elements than `initialCapacity` are stored, it's advantageous to prevent
 /// expansions from happening frequently. Expansions will always force an allocation and a copy to happen.
-public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
+public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollection {
     public typealias RangeType<Bound> = Range<Bound> where Bound: Strideable, Bound.Stride: SignedInteger
-    private var buffer: ContiguousArray<E?>
+    private var buffer: ContiguousArray<Element?>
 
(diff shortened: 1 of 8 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -33,6 +33,30 @@ public struct CircularBuffer<E>: CustomStringConvertible, AppendableCollection {
...
+        @inlinable
+        public func distance(to other: Index) -> Int {
+            return self._value.distance(to: other._value)
+        }
+
+        @inlinable
+        public func advanced(by n: Int) -> Index {
+            return Index(_value: self._value + n)
+        }
+    }
```
Lukasa: This is a note for the future, we may find it useful to implement the `Comparable` functions ourselves rather than use the synthesised implementations on `Strideable`, just to avoid stressing the inliner too badly.

Sources/NIO/MarkedCircularBuffer.swift
```diff
@@ -63,7 +67,7 @@ public struct MarkedCircularBuffer<E>: CustomStringConvertible, AppendableCollec
     }
 
     /// Retrieves the element at the given index from the buffer, without removing it.
-    public subscript(index: Int) -> E {
+    public subscript(index: CircularBuffer<E>.Index) -> E {
```
Lukasa: I think we should use a typealias for this type, given that we have to reproduce the generic type name everywhere.

weissi: good point, thanks! done

Sources/NIO/CircularBuffer.swift
weissi: done too :)

weissi:
@Lukasa I think this could now go in if you're still happy. The shims have been done

PR #827, opened 2019-02-14
make ChannelOptions.Storage public

weissi:
Motivation:

To implement your own `Channel`s, you very likely need
`ChannelOptions.Storage`, so let's make it public.

Modifications:

- make ChannelOptions.Storage public

Result:

easier custom Channels


weissi:
@Lukasa fixed the merge commits here

PR #831, opened 2019-02-20
improve B2MD EOF handling

weissi:
Motivation:

Previously B2MDs didn't really have defined semantics regarding EOFs and
we didn't tell them if there was an EOF. Also `decodeLast` was optional
and all that was bad.

Modifications:

- require `decodeLast`
- add a `seenEOF: Bool` parameter to `decodeLast` which tells the
  decoder if an EOF has been seen

Result:

- clearer semantics
- more information

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -63,9 +63,10 @@ public protocol ByteToMessageDecoder {
     /// - parameters:
     ///     - ctx: The `ChannelHandlerContext` which this `ByteToMessageDecoder` belongs to.
     ///     - buffer: The `ByteBuffer` from which we decode.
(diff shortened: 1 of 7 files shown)


Sources/NIO/Codec.swift
```diff
@@ -266,7 +273,8 @@ public class ByteToMessageHandler<Decoder: ByteToMessageDecoder> {
     internal private(set) var decoder: Decoder? // only `nil` if we're already decoding (ie. we're re-entered)
     private var state: State = .active
     private var removalState: RemovalState = .notBeingRemoved
-    private var buffer: B2MDBuffer = B2MDBuffer()
+    private var buffer: B2MDBuffer!
```
Lukasa: Please add a comment that explains why this is an IUO.

weissi: thanks, done

Issue #829, opened 2019-02-19
Make localAddress/remoteAddress mockable in EmbeddedChannel


mkieselmann:
### Problem
Testing channel handlers that rely on `Channel.localAddress` or `Channel.remoteAddress` is currently quite inconvenient as one cannot set test values for `localeAddress` or `remoteAddress` on `EmbeddedChannel`.

### Expected behavior

1) The two properties could be variables instead of constants.
```
public class EmbeddedChannel: Channel {
    // ...
    public var localAddress: SocketAddress? = nil
    public var remoteAddress: SocketAddress? = nil
    // ...
}
```

2) `EmbeddedChannel` could be `open` for subclassing. 
```
open class EmbeddedChannel: Channel {
    // ...
    open let localAddress: SocketAddress? = nil
    open let remoteAddress: SocketAddress? = nil
    // ...
}
```

### Actual behavior
 It seems that the only solution is to replicate most of the code of `EmbeddedChannel` and implement `ChannelHandler` protocol.

weissi:
@mkieselmann thanks for your report! One other option we would have is just to add a setter on `EmbeddedChannel`, right? That way you could just set them which would also be enough, right?

We could also just store the values from `connect` and `bind` but that's a bit odd too...

@Lukasa  what do you think?

mkieselmann:
@weissi thanks for the quick response. Sure, having a setter would be fine. Not sure how it would work with `connect` or `bind` since we would need some kind of bootstrapping in the test setup. Currently I just have an `EmbeddedChannel` with one or more `ChannelHandlers` added to the pipeline.

Lukasa:
@weissi If we were going to have a setter, why not just make them `var` properties?

weissi:
> @weissi If we were going to have a setter, why not just make them `var` properties?

Yes, that works and it a really easy one to PR :)

weissi:
> Not sure how it would work with `connect` or `bind` since we would need some kind of bootstrapping in the test setup.

You could just

XCTAssertNoThrow(channel.connect(...).wait())
XCTAssertNoThrow(channel.bind(...).wait())

so you don't need to use a bootstrap if you're not actually doing a real connection. But I think the setters (through `var localAddress` / var `remoteAddress`) seem better anyway.

weissi:
we can obviously also do both, set the addresses from `connect`/`bind` as well as allowing the properties to be set directly...

mkieselmann:
> You could just
>
> ```swift
[...]

PR #833, opened 2019-02-22
make localAddress and remoteAddress testable in EmbeddedChannel

mkieselmann:
### Motivation:

Testing channel handlers that rely on `Channel.localAddress` or `Channel.remoteAddress` is currently quite inconvenient as one cannot set test values for `localeAddress` or `remoteAddress` on `EmbeddedChannel`.

### Modifications:

Change `localAddress` and `remoteAddress` to variables instead of constants and set the address values in `bind()` and `connect()` as proposed in #829.

### Result:

Arbitrary values for `localAddress` or `remoteAddress` can be set on `EmbeddedChannel`.

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -318,8 +318,8 @@ public class EmbeddedChannel: Channel {
     public var allocator: ByteBufferAllocator = ByteBufferAllocator()
     public var eventLoop: EventLoop = EmbeddedEventLoop()
 
-    public let localAddress: SocketAddress? = nil
-    public let remoteAddress: SocketAddress? = nil
+    public var localAddress: SocketAddress? = nil
+    public var remoteAddress: SocketAddress? = nil
 
     // Embedded channels never have parents.
     public let parent: Channel? = nil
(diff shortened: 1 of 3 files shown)


Lukasa:

mkieselmann:
> Cool, this LGTM. Do you mind using the commit message template as shown in the `dev` directory?

Sure, done!

Sources/NIO/Embedded.swift
```diff
@@ -396,4 +396,18 @@ public class EmbeddedChannel: Channel {
...
+
+    public func bind(to address: SocketAddress, promise: EventLoopPromise<Void>?) {
+        promise?.futureResult.whenSuccess { _ in
+            self.localAddress = address
+        }
+        pipeline.bind(to: address, promise: promise)
+    }
+
+    public func connect(to address: SocketAddress, promise: EventLoopPromise<Void>?) {
+        promise?.futureResult.whenSuccess { _ in
```
Lukasa: Nit: you shouldn't need an `_ in` here, as the return value of the `promise` is `Void`, which you're allowed to implicitly discard.

Sources/NIO/Embedded.swift
```diff
@@ -396,4 +396,18 @@ public class EmbeddedChannel: Channel {
         }
         fatalError("option \(option) not supported")
     }
+
+    public func bind(to address: SocketAddress, promise: EventLoopPromise<Void>?) {
+        promise?.futureResult.whenSuccess { _ in
```
Lukasa: Nit: you shouldn't need a `_ in` here, as the return value of the `promise` is `Void`, which you're allowed to implicitly discard.

Sources/NIO/Embedded.swift
mkieselmann: Good catch 👍

Sources/NIO/Embedded.swift
mkieselmann: Updated...

Lukasa:

Lukasa:
This LGTM. Thanks so much!

PR #834, opened 2019-02-22
introduce WriteObservingByteToMessageDecoder

weissi:
Motivation:

Some B2MDs need to observe writes.

Modifications:

Make those B2MD's B2MHs ChannelOutboundHandlers and let the B2MDs
observe the writes.

Result:

HTTPDecoder can be expressed as B2MD.

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -89,6 +89,22 @@ public protocol ByteToMessageDecoder {
     mutating func shouldReclaimBytes(buffer: ByteBuffer) -> Bool
 }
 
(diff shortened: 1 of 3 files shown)


Sources/NIO/Codec.swift
```diff
@@ -337,6 +358,12 @@ extension ByteToMessageHandler {
         }
     }
 
+    private func tryDecodeWrites() {
+        if self.queuedWrites.count > 0 {
+            (self as! CanDequeueWrites).dequeueWrites()
```
Lukasa: Why do we need this force cast?

weissi: @Lukasa it's a precondition. We can only enter this is we are a `CanDequeueWrites`, I'll add a comment though

weissi:
@Lukasa updated

PR #835, opened 2019-02-24
Add CircularBuffer Index implementation

### LINKED ISSUE #836: MarkedCircularBuffer uses CircularBuffer indices wrongly
thanks to @JovanMilenkovic for reporting:

`MarkedCircularBuffer` relies on indices into `CircularBuffer` remaining valid after mutating operations. That's not legal and needs to change. `MarkedCircularBuffer` should just store an offset from the start/end.

JovanMilenkovic:
Motivation:

Provide a way for CircularBuffer to directly access underlying buffer with an index instead of needing to calculate depending on current head index and underlying buffer count. Related issue: https://github.com/apple/swift-nio/issues/682

Modifications:

Added needed properties to circular buffers Index and adjust CircularBuffer code accordingly.
Removed MarkedCircularBuffer assumption that an index into it's underlying buffer will not be invalidated.

Result:

- CircularBuffer will have a complete Index implementation
- fixes #836

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -28,61 +28,80 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
     private var buffer: ContiguousArray<Element?>
 
     /// The index into the buffer of the first item
-    private(set) /* private but tests */ internal var headIdx = 0
+    private(set) /* private but tests */ internal var headIdx: Index
 
     /// The index into the buffer of the next free slot
-    private(set) /* private but tests */ internal var tailIdx = 0
+    private(set) /* private but tests */ internal var tailIdx: Index
 
     public struct Index: Strideable, Comparable {
-        @usableFromInline
-        var _value: Int
-
-        public typealias Stride = Int
+        @usableFromInline var backingIndex: ContiguousArray<Element?>.Index
+        @usableFromInline let backingCount: Int
+        @usableFromInline var isIndexGEQHeadIndex: Bool
+        @inlinable var mask: Int { return self.backingCount - 1 }
 
         @inlinable
-        internal init(_value: Int) {
-            self._value = _value
+        internal init(backingIndex: ContiguousArray<Element?>.Index, backingIndexOfHead: ContiguousArray<Element?>.Index, backingCount: Int) {
+            self.isIndexGEQHeadIndex = backingIndex >= backingIndexOfHead
+            self.backingCount = backingCount
+            self.backingIndex = backingIndex
         }
-
+        
+        private init(backingIndex: ContiguousArray<Element?>.Index, greater: Bool, backingCount: Int) {
+            self.backingIndex = backingIndex
+            self.isIndexGEQHeadIndex = greater
+            self.backingCount = backingCount
+        }
+        
         @inlinable
         public func distance(to other: Index) -> Int {
-            return self._value.distance(to: other._value)
+            if self.isIndexGEQHeadIndex && other.isIndexGEQHeadIndex {
+                return self.backingIndex.distance(to: other.backingIndex)
+            } else if self.isIndexGEQHeadIndex && !other.isIndexGEQHeadIndex {
+                return self.backingIndex.distance(to: backingCount) + other.backingIndex
+            } else if !self.isIndexGEQHeadIndex && other.isIndexGEQHeadIndex {
+                return self.backingIndex.distance(to: 0) + backingCount.distance(to: other.backingIndex)
+            } else {
+                return self.backingIndex.distance(to: other.backingIndex)
+            }
         }
 
-        @inlinable
         public func advanced(by n: Int) -> Index {
-            return Index(_value: self._value + n)
+            var index = Index(backingIndex: 0, greater: self.isIndexGEQHeadIndex, backingCount: self.backingCount)
+            if n >= 0 {
+                if backingIndex + n >= self.backingCount {
+                    index.isIndexGEQHeadIndex = !self.isIndexGEQHeadIndex
+                }
+            } else {
+                if backingIndex + n < 0 {
+                    index.isIndexGEQHeadIndex = !self.isIndexGEQHeadIndex
+                }
+            }
+            index.backingIndex = (self.backingIndex + n) & (self.mask)
+            return index
         }
 
-        @inlinable
         public static func < (lhs: Index, rhs: Index) -> Bool {
-            return lhs._value < rhs._value
-        }
-
-        @inlinable
-        public static func <= (lhs: Index, rhs: Index) -> Bool {
-            return lhs._value <= rhs._value
-        }
-
-        @inlinable
-        public static func == (lhs: Index, rhs: Index) -> Bool {
-            return lhs._value == rhs._value
+            if lhs.isIndexGEQHeadIndex && rhs.isIndexGEQHeadIndex {
+                return lhs.backingIndex < rhs.backingIndex
+            } else if lhs.isIndexGEQHeadIndex && !rhs.isIndexGEQHeadIndex {
+                return true
+            } else if !lhs.isIndexGEQHeadIndex && rhs.isIndexGEQHeadIndex {
+                return false
+            } else {
+                return lhs.backingIndex < rhs.backingIndex
+            }
         }
     }
 }
 
 extension CircularBuffer {
-    /// Bitmask used for calculating the tailIdx / headIdx based on the fact that the underlying storage
-    /// has always a size of power of two.
-    private var mask: Int {
-        return self.buffer.count - 1
-    }
-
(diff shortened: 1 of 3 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -126,29 +136,42 @@ extension CircularBuffer {
...
+        newBacking.append(contentsOf: repeatElement(nil, count: repeatitionCount))
+        self.headIdx = Index(value: 0, headValue: 0, underlyingCount: newBacking.count)
+        self.tailIdx = Index(value: newBacking.count - repeatitionCount, headValue: self.headIdx._value, underlyingCount: newBacking.count)
         self.buffer = newBacking
     }
+    
+    /// Return element `offset` from first element.
+    ///
+    /// *O(1)*
+    public subscript(offset: Int) -> Element {
```
weissi: would you mind making this label named, ie `buffer[offset: 7]` which would mean that it would be `public subscript(offset offset: Int) -> Element`

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,69 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
 
     public struct Index: Strideable, Comparable {
-        @usableFromInline
-        var _value: Int
-
-        public typealias Stride = Int
+        @usableFromInline var _value: ContiguousArray<Element?>.Index
+        @usableFromInline var _headValue: ContiguousArray<Element?>.Index
+        @usableFromInline let underlyingCount: Int
+        @usableFromInline var mask: Int { return self.underlyingCount - 1 }
```
weissi: `mask` should be `@inlinable`

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,69 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
     /// The index into the buffer of the next free slot
-    private(set) /* private but tests */ internal var tailIdx = 0
+    private(set) /* private but tests */ internal var tailIdx: Index
 
     public struct Index: Strideable, Comparable {
-        @usableFromInline
-        var _value: Int
-
-        public typealias Stride = Int
+        @usableFromInline var _value: ContiguousArray<Element?>.Index
```
weissi: I think I would prefer naming this `bufferIndex` and `bufferIndexOfHead` or something, think that describes it  better. What do you think?

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,69 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
+        @usableFromInline var _headValue: ContiguousArray<Element?>.Index
+        @usableFromInline let underlyingCount: Int
```
weissi: and I guess `underlyingCount` could be `bufferCount`. Or maybe we could use the word `backing`, ie. `backingIndex`, `backingIndexOfHead` and `backingCount`? I would also like `underlyingBuffer` but then all becomes a bit long

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,69 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
-        @inlinable
-        public static func <= (lhs: Index, rhs: Index) -> Bool {
-            return lhs._value <= rhs._value
-        }
-
-        @inlinable
-        public static func == (lhs: Index, rhs: Index) -> Bool {
-            return lhs._value == rhs._value
+            let head = lhs._headValue
+            if lhs._value >= head && rhs._value >= head {
```
weissi: I think it might be worth adding a few test cases to hit all those cases

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,69 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
-    private(set) /* private but tests */ internal var tailIdx = 0
+    private(set) /* private but tests */ internal var tailIdx: Index
 
     public struct Index: Strideable, Comparable {
-        @usableFromInline
-        var _value: Int
-
-        public typealias Stride = Int
+        @usableFromInline var _value: ContiguousArray<Element?>.Index
+        @usableFromInline var _headValue: ContiguousArray<Element?>.Index
```
weissi: what I'm just wondering: Do we actually need to index of the head at all? Couldn't we just store a bit whether our index is >= head index or not?

weissi:

Sources/NIO/CircularBuffer.swift
JovanMilenkovic: Nice catch, it may work.

Sources/NIO/CircularBuffer.swift
JovanMilenkovic: A bit off topic but I'm unable to run the tests in XCode with code coverage enabled so I can't see if all code paths are covered already by other tests. Is this a general issue or is it just me?

Sources/NIO/CircularBuffer.swift
JovanMilenkovic: I think we should go with `backing`, seems it's the best one.

weissi:

Sources/NIO/CircularBuffer.swift
weissi: @JovanMilenkovic that's unfortunately a bug in the Xcode beta...

weissi:

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,80 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
     private var buffer: ContiguousArray<Element?>
 
     /// The index into the buffer of the first item
-    private(set) /* private but tests */ internal var headIdx = 0
+    private(set) /* private but tests */ internal var headIdx: Index
```
weissi: I think it would be enough to just store the backing index as `Int` here. We can always fabricate the `startIndex` as actual `Index` when we need to as `Index(backingIndex self.headIdx, count: self.backing.count, isGEQ: true)` right?

Sources/NIO/CircularBuffer.swift
```diff
@@ -28,61 +28,80 @@ public struct CircularBuffer<Element>: CustomStringConvertible, AppendableCollec
...
 
     /// The index into the buffer of the next free slot
-    private(set) /* private but tests */ internal var tailIdx = 0
+    private(set) /* private but tests */ internal var tailIdx: Index
```
weissi: same here: is there a reason we store a whole `Index` instead of just the backing index as `Int`?

weissi:

Sources/NIO/CircularBuffer.swift
JovanMilenkovic: Well mostly because those are 2 main indexes of the `CircularBuffer` so it made sense they are of `Index` type since all operations are now performed over that type.
I would feel like seeing something that someone didn't intend working that way in this code if those were `Int` and constructed on access:
` /// Return all valid indices of the ring.
    public var indices: RangeType<Index> {
        return self.startIndex ..< self.endIndex
    }
    /// Returns the index of the first element of the ring.
    public var startIndex: Index {
        return self.headIdx
    }

    /// Returns the ring's "past the end" position -- that is, the position one greater than the last valid subscript argument.
    public var endIndex: Index {
        return self.tailIdx
    }`
Additionally, masking would need to be introduced into `CircularBuffer` for `count`, and it's contained inside `Index` at the moment.
The rest of the code relies on `count`, `startIndex` and `endIndex` and maybe less important for this question but many operations move head and tail by 1 or -1. All of this seems very unpredictable performance wise compared to storing them, but I'm

weissi: @JovanMilenkovic the downside of storing those as `Index` is that it makes `CircularBuffer` a lot wider. If we stored them as `Int`, that's two words ( 1 for startIndex, 1 for endIndex). Storing them as `Index` would be 6 words ((2 words + 1 byte) for startIndex + padding + (2 words + 1 byte) for endIndex + padding). So I'd advise to only store what we need to store and calculate the rest

weissi:
@JovanMilenkovic Thanks very much for your work here! I'm building a follow-up PR which will contain exactly your commit and the test fixes etc. I would've merged this one but unfortunately it doesn't pass the tests just yet so I'm just building a followup PR which will contain your commit & another from me.

Is that okay with you?

JovanMilenkovic:
Sure, I’m sorry I didn’t have the time to follow up, and wouldn’t until saturday. That’s cool, thanks.

weissi:
don't apologise @JovanMilenkovic , that's great work! I'm closing this PR in favour of #906 which contains the commit of this PR as is (and another one to fix the tests and further improve the indices).

Issue #600, opened 2018-08-29
don't make HTTPDecoder close the connection immediately on parser error


weissi:
### Expected behavior

like most NIO handlers, a random decoder in the pipeline shouldn't just close the connection. The usual mechanism is to fire an error and let later handlers in the pipeline deal with them. The `HTTPDecoder` however just [closes the connection](https://github.com/apple/swift-nio/blob/df764fedf0c7938629581ad1f61b2b4474fc66b4/Sources/NIOHTTP1/HTTPDecoder.swift#L606) on parser error.

### Actual behavior

In pipelined scenarios (where the parser error arrives whilst a previously pipelined request is still being processed) that leads to the scenario that the connection gets shut early.

### Steps to reproduce

1. enqueue two (or more) HTTP requests, have the first one processing for a while and let the second one be a parser error

### Swift & OS version (output of `swift --version && uname -a`)

all NIOs

### Other info

- this would be SemVer major breaking so a change would need to be an option in an upcoming minor release
- we had a similar thing in the [websockets decoder](https://github.com/apple/swift-nio/blob/df764fedf0c7938629581ad1f61b2b4474fc66b4/Sources/NIOWebSocket/WebSocketFrameDecoder.swift#L253-L260). In the websockets decoder we introduced a `automaticErrorHandling: Bool  = true` (which means to just shut the connection). So we could have a similar thing for the HTTPDecoder.

weissi:
this is implemented in #837 and #814 .

weissi:
all done now

PR #837, opened 2019-02-25
B2MD: Don't deliver data after error

weissi:
Motivation:

B2MD should never call any of the decode functions after the user has
thrown an error.

Modifications:

introduce an error state and make sure we don't call out if in error
state

Result:

B2MD more correct

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -22,6 +22,19 @@ public enum DecodingState {
     case needMoreData
 }
 
+/// Common errors thrown by `ByteToMessageDecoder`s.
+public enum ByteToMessageDecoderError: Error {
+    /// More data has been received by a `ByteToMessageHandler` despite the fact that an error has previously been
+    /// emitted. The associated `Error` is the error previously emitted and the `ByteBuffer` is the extra data that has
+    /// been received. The common cause for this error to be emitted is the user not having torn down the `Channel`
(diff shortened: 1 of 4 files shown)


Sources/NIO/Codec.swift
```diff
@@ -22,6 +22,19 @@ public enum DecodingState {
     case needMoreData
 }
 
+/// Common errors throw by `ByteToMessageDecoder`s.
```
Lukasa: Nit: thrown.

Sources/NIO/Codec.swift
```diff
@@ -322,22 +364,26 @@ extension ByteToMessageHandler {
...
 
     private func processLeftovers(ctx: ChannelHandlerContext) {
-        guard self.state == .active else {
+        guard self.state.isActive else {
             // we are processing or have already processed the leftovers
             return
         }
 
-        ctx.withThrowingToFireErrorAndClose {
+        do {
```
Lukasa: This change is potentially going to cause some breakage for folks, as there is now nothing that closes a HTTP connection on error by default. Same for any other B2MD.

weissi: yes, that was deliberate. I added it now to the public API changes. Good enough?

weissi: #600 also asks for that

PR #844, opened 2019-02-26
Allow delaying HTTP upgrade decisions.

### LINKED ISSUE #699: Allow HTTPProtocolUpgrader.buildUpgradeResponse to return a Future
`buildUpgradeResponse` may want to perform some amount of blocking work before it returns. For example, it is possible that users may want to require that web socket requests be authenticated, and to look credentials up in a database.

This requires us to enhance HTTPProtocolUpgrader to allow returning a future.

A side effect of doing this is that it makes the state flow substantially more complex. For this reason, we are probably going to have to improve the `HTTPServerUpgradeHandler` substantially to manage the state interactions. We also need to decide whether we race the upgraders, or whether we still ask for upgrade sequentially.

/cc @tanner0101

Lukasa:
Motivation:

While the HTTPServerUpgrader and WebSocketUpgrader allowed users to take
their time when reconfiguring the pipeline after they decided to upgrade,
users had to synchronously decide if they wanted to upgrade. This is a bit
inconvenient.

A particular limitation here is that some routes may want to upgrade
only if the upgrading user is authenticated. Checking authentication
credentials is almost always an I/O operation, and so cannot safely be done
on the event loop. Our original design made this impossible.

Modifications:

- Changed the shouldUpgrade callback to return a Future.
- Passed the shouldUpgrade callback the Channel that is upgrading, in
    no small part so that it has an EventLoop it can create Futures
    on.
- Rewrote the upgrader to handle this new state.

Result:

Users can delay decisions about when to upgrade.
Resolves #699.

### DIFF
--- a/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/HTTPUpgradeHandler.swift
@@ -41,8 +41,8 @@ public protocol HTTPServerProtocolUpgrader {
 
     /// Builds the upgrade response headers. Should return any headers that need to be supplied to the client
     /// in the 101 Switching Protocols response. If upgrade cannot proceed for any reason, this function should
-    /// throw.
-    func buildUpgradeResponse(upgradeRequest: HTTPRequestHead, initialResponseHeaders: HTTPHeaders) throws -> HTTPHeaders
+    /// fail the future.
+    func buildUpgradeResponse(channel: Channel, upgradeRequest: HTTPRequestHead, initialResponseHeaders: HTTPHeaders) -> EventLoopFuture<HTTPHeaders>
 
     /// Called when the upgrade response has been flushed. At this time it is safe to mutate the channel pipeline
     /// to add whatever channel handlers are required. Until the returned `EventLoopFuture` succeeds, all received
@@ -74,8 +74,8 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableChannelHa
     private var seenFirstRequest = false
 
     /// The closure that should be invoked when the end of the upgrade request is received if any.
-    private var upgrade: (() -> Void)?
-    private var receivedMessages: [NIOAny] = []
+    private var upgradeState: UpgradeState = .idle
+    private var receivedMessages: CircularBuffer<NIOAny> = CircularBuffer()
 
     /// Create a `HTTPServerUpgradeHandler`.
     ///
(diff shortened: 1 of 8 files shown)


Lukasa:
Sadly the breaking API change here cannot be shimmed.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -108,74 +108,99 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableChannelHa
...
+            preconditionFailure("We think we saw .end before and began upgrading, but somehow we have not set seenFirstRequest")
+        }
+    }
+
+    private func firstRequestHeadReceived(context: ChannelHandlerContext, requestPart: HTTPServerRequestPart) {
+        // We should decide if we're going to upgrade based on the first request header: if we aren't upgrading,
+        // by the time the body comes in we should be out of the pipeline. That means that if we don't think we're
+        // upgrading, the only thing we should see is a request head. Anything else in an error.
+        guard case .head(let request) = requestPart else {
+            context.fireErrorCaught(HTTPServerUpgradeErrors.invalidHTTPOrdering)
```
weissi: is this not a precondition?

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
```diff
@@ -74,7 +74,7 @@ public class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableChannelHa
     private var seenFirstRequest = false
 
     /// The closure that should be invoked when the end of the upgrade request is received if any.
-    private var upgrade: (() -> Void)?
+    private var upgradeState: UpgradeState = .idle
     private var receivedMessages: [NIOAny] = []
```
weissi: this should be a `CircularBuffer<NIOAny>` because we're looping over `removeFirst()` which is quite expensive as array

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
Lukasa: I'm open to it being one. This isn't a change in behaviour from the old model: we also treated it as an error there. I think I was reluctant to precondition on *inbound* data, but I'm open to deciding that there are enough handlers between here and the network that it's safe to do so.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
Lukasa: Good catch, will amend.

Sources/NIOHTTP1/HTTPUpgradeHandler.swift
weissi: Ok, let’s leave fof now

PR #846, opened 2019-02-26
ELF: make everything inlinable

weissi:
Motivation:

ELFs have loads of generic types and that basically screams allocations
and swift_getGenericMetdata everywhere because nothing was inlinable.

Modifications:

make everything inlinable

Result:

better performance working with ELFs


weissi:
ping @Lukasa

PR #862, opened 2019-03-03
Added vendored copy of zlib

satishbabariya:
Added vendored copy of zlib.

### Motivation:

Yesterday i encountered this,
```sh
satish@ubuntu:~/swift-nio$ swift build
<module-includes>:1:10: note: in file included from <module-includes>:1:
#include "/home/satish/swift-nio/Sources/CNIOZlib/include/CNIOZlib.h"
         ^
/home/satish/swift-nio/Sources/CNIOZlib/include/CNIOZlib.h:17:10: error: 'zlib.h' file not found
#include <zlib.h>
         ^
```

So, for easy to build and deploy swift-nio on all platforms for all users. provide vendored copy of zlib with swift-nio.
also nodejs ships vendored copy of zlib with it.

Exposing CZlib will make easy to integrate zlib for swift-nio adoptors.
For example if swift-nio adoptors want to implement compression or gzip
they can use CZlib from swift-nio directly.

### Modifications:

Added vendored copy of zlib.
Introducing CZlib in products.

### Result:

swift-nio will stop linking against the system copy of zlib and use vendored copy of zlib.
CZlib will exposed in swift-nio Package file


Lukasa:
I'm afraid I don't think this is a good idea, so I'm -1 on this.

> So, for easy to build and deploy swift-nio on all platforms for all users. provide vendored copy of zlib with swift-nio.
> also nodejs ships vendored copy of zlib with it.

There is essentially no platform that does not have an easily-accessible copy of zlib either in the base distribution or a single package manager operation away. It's also easily identified from the compiler error if such a thing is not installed

> Exposing CZlib will make easy to integrate zlib for swift-nio adoptors.
> For example if swift-nio adoptors want to implement compression or gzip
they can use CZlib from swift-nio directly.

This is also a thing I don't think we should do, even if we wanted to drop the dependency on the system zlib. I'd strongly have proposed using a separate SwiftPM package that is appropriately versioned and then add an appropriate dependency on it, rather than have swift-nio be a backdoor mechanism to add zlib bindings.

satishbabariya:
I'm closing on this.

PR #869, opened 2019-03-04
ByteBuffer get* now respects reader/writer indexes

### LINKED ISSUE #669: ByteBuffer methods to respect reader/writer indexes
`ByteBuffer` has some `get*` methods which do not respect the reader/writer indexes, and instead only respect the capacity of the buffer. This may result in accessing uninitialized memory and lead to unexpected results. I would propose that by default we have methods respect the reader/writer indexes and only if necessary expand the `ByteBuffer` API to include methods which ignore the reader/writer index (e.g. new methods, optional parameter, etc...).

Netty issue: https://github.com/netty/netty/issues/8601

documented here:
- [Important Usage Notes](https://apple.github.io/swift-nio/docs/current/NIO/Structs/ByteBuffer.html#important-usage-notes)
- for example in [`ByteBuffer.getInteger`](https://apple.github.io/swift-nio/docs/current/NIO/Structs/ByteBuffer.html#/s:3NIO10ByteBufferV10getIntegerxSgSi2at_AA10EndiannessO10endiannessxm2asts010FixedWidthE0RzlF)'s warning

weissi:
Motivation:

Previously, ByteBuffers get* methods would get bytes from anywhere, even
if the bytes were not readable. That can lead to security issues in code
that doesn't expect this. NIO shouldn't contain such unsafe methods
without the word 'unsafe'.

Modifications:

Made all `get*` methods respect reader/writerIndex.

Result:

- safer user code
- fixes #669

### DIFF
--- a/Sources/NIO/DatagramVectorReadManager.swift
+++ b/Sources/NIO/DatagramVectorReadManager.swift
@@ -113,7 +113,10 @@ struct DatagramVectorReadManager {
             assert(messagesProcessed == 0)
             return .none
         case .processed(let messagesProcessed):
-            return self.buildMessages(messageCount: messagesProcessed, sliceSize: messageSize, buffer: &buffer)
+            buffer.moveWriterIndex(to: messageSize * messagesProcessed)
+            return self.buildMessages(messageCount: messagesProcessed,
+                                      sliceSize: messageSize,
+                                      buffer: &buffer)
         }
     }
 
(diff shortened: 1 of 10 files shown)


Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -113,7 +113,10 @@ struct DatagramVectorReadManager {
             assert(messagesProcessed == 0)
             return .none
         case .processed(let messagesProcessed):
-            return self.buildMessages(messageCount: messagesProcessed, sliceSize: messageSize, buffer: &buffer)
+            buffer.moveWriterIndex(to: messageSize * messagesProcessed)
```
weissi: @Lukasa requesting re-review because of this change.

weissi: basically need you to confirm that this is correct (test crashed on Linux otherwise)

Lukasa: LGTM.

Issue #527, opened 2018-07-27
NIOWebSocket could let users handle protocol errors


pushkarnk:
### Expected behavior
When the WebSocketFrameDecoder flags a WebSocket error, it must be possible for the user to handle it.

### Actual behavior
The WebSocketFrameDecoder does some error handling in [handleError](https://github.com/apple/swift-nio/blob/master/Sources/NIOWebSocket/WebSocketFrameDecoder.swift#L304). This doesn't permit any kind of error handling on the user's side, in the `errorCaught` callback.

The `handleError` function, which today looks like:
```
private func handleError(_ error: Error, ctx: ChannelHandlerContext) {
        guard let error = error as? NIOWebSocketError else {
            fatalError("Can only handle NIOWebSocketErrors")
        }
        self.shouldKeepParsing = false

        var data = ctx.channel.allocator.buffer(capacity: 2)
        data.write(webSocketErrorCode: WebSocketErrorCode(error))
        let frame = WebSocketFrame(fin: true,
                                   opcode: .connectionClose,
                                   data: data)
        ctx.writeAndFlush(self.wrapOutboundOut(frame)).whenComplete {
            ctx.close(promise: nil)
        }
        ctx.fireErrorCaught(error)
    }
```
... could be reduced to

```
private func handleError(_ error: Error, ctx: ChannelHandlerContext) {
        guard let error = error as? NIOWebSocketError else {
            fatalError("Can only handle NIOWebSocketErrors")
        }
        self.shouldKeepParsing = false

        ctx.fireErrorCaught(error)
    }
```

pushkarnk:
cc @Lukasa

Lukasa:
See #528.

PR #885, opened 2019-03-08
remove WebSocketFrameDecoder inline error handling

### LINKED ISSUE #534: Remove WebSocketFrameDecoder inline error handling.
Follows on from the work done in #528 for #527: we have moved the the default error handling out of `WebSocketFrameDecoder`, but had to leave the code there for backward compatibility reasons. We should remove that code now. The code is marked by a TODO.

weissi:
Motivation:

Follows on from the work done in #528 for #527: we have moved the the
default error handling out of WebSocketFrameDecoder, but had to leave
the code there for backward compatibility reasons. We can remove that
code now.

Modifications:

Removed automatic error handling code in WebSocketFrameDecoder.

Result:

- fixes #534

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,7 +17,7 @@ import PackageDescription
 
 var targets: [PackageDescription.Target] = [
     .target(name: "_NIO1APIShims",
-            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "NIOFoundationCompat"]),
+            dependencies: ["NIO", "NIOHTTP1", "NIOTLS", "NIOFoundationCompat", "NIOWebSocket"]),
     .target(name: "NIO",
             dependencies: ["CNIOLinux",
                            "CNIODarwin",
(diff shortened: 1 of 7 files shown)


Sources/NIOWebSocket/WebSocketProtocolErrorHandler.swift
```diff
@@ -35,10 +35,10 @@ public final class WebSocketProtocolErrorHandler: ChannelInboundHandler {
...
                 context.close(promise: nil)
             }
-        }
 
-        // Regardless of whether this is an error we want to handle or not, we always
-        // forward the error on to let others see it.
-        context.fireErrorCaught(error)
+            // Regardless of whether this is an error we want to handle or not, we always
+            // forward the error on to let others see it.
+            context.fireErrorCaught(error)
```
normanmaurer: This change does not look correct. We only forward now when its a `NIOWebSocketError`. I think we should always forward (as also stated in the comment) ;)

normanmaurer: Please also add a test for this

weissi: @normanmaurer good catch!

weissi: @normanmaurer done

PR #886, opened 2019-03-08
remove overloading from scheduleRepeatedTask

### LINKED ISSUE #882: go hunt overloaded functions that make NIO hard to use
functions like `scheduleRepeatedTask` are basically impossible to use because they're overloaded on the closure's return type. We ought to rename them and I'm sure there's similar other ones out there.

Proposal:

 - for sync stuff: `scheduleRepeatedSyncTask` or maybe just `scheduleRepeatedTask` (closure returns `Void`)
 - for async stuff: `scheduleRepeatedAsyncTask` (closure returns `ELF<Void>`)

weissi:
Motivation:

scheduleRepeatedTask had two overloads: for synchronous and asynchronous
tasks. They however had the same name so they were pretty much
impossible to use.

Modifications:

rename the asynchronous version of scheduleRepeatedTask to
scheduleRepeatedAsyncTask

Result:

- NIO easier to use
- fixes #882

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -462,7 +462,8 @@ extension EventLoop {
         // Do nothing
     }
 
-    /// Schedule a repeated task to be executed by the `EventLoop` with a fixed delay between the end and start of each task.
+    /// Schedule a repeated task to be executed by the `EventLoop` with a fixed delay between the end and start of each
+    /// task.
     ///
     /// - parameters:
     ///     - initialDelay: The delay after which the first task is executed.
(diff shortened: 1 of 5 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -492,7 +499,10 @@ extension EventLoop {
     ///     - task: The closure that will be executed.
     /// - return: `RepeatedTask`
     @discardableResult
-    public func scheduleRepeatedTask(initialDelay: TimeAmount, delay: TimeAmount, notifying promise: EventLoopPromise<Void>? = nil, _ task: @escaping (RepeatedTask) -> EventLoopFuture<Void>) -> RepeatedTask {
+    public func scheduleRepeatedAsyncTask(initialDelay: TimeAmount,
```
normanmaurer: do we also need to update the tests so it not use the old deprecated name ?

weissi: @normanmaurer wow, good catch. There was no test (contributor contribution) for that method. But all synchronous ones use the async one to execute. Added a good test.

PR #650, opened 2018-11-04
Multicast Updates

khoogheem:
Support for Interface to report if Multicast is supported.  Channel can provide Error if Interface doesn't support Multicast.

### Motivation:

Allows for checking if the interface you are setting up for Multicast even supports Multicast in the first place.  In multiple projects we have done with Multicast it helps as we still run into some network managers that shutdown multicast support.

### Modifications:

Interface exposes a new variable to determine if it supports Multicast.
performGroupOperation0 will check if interface is not nil and supports multicast, otherwise will throw error that the interface doesn't support Multicast 

### Result:

This should hopefully better inform if the Multicast attempt is going to even work.

PR #887, opened 2019-03-08
NIONetworkInterface: Return if multicast is supported

weissi:
this is a fixup and replacement of #650. We need to get this in because NIO converges.

Support for Interface to report if Multicast is supported. Channel can provide Error if Interface doesn't support Multicast.

Motivation:

Allows for checking if the interface you are setting up for Multicast even supports Multicast in the first place. In multiple projects we have done with Multicast it helps as we still run into some network managers that shutdown multicast support.

Modifications:

Interface exposes a new variable to determine if it supports Multicast.
performGroupOperation0 will check if interface is not nil and supports multicast, otherwise will throw error that the interface doesn't support Multicast

Result:

This should hopefully better inform if the Multicast attempt is going to even work.

### DIFF
--- a/Sources/NIO/Channel.swift
+++ b/Sources/NIO/Channel.swift
@@ -337,6 +337,9 @@ public enum ChannelError: Error {
     /// address.
     case illegalMulticastAddress(SocketAddress)
 
+    /// Multicast is not supported on Interface
+    case multicastNotSupported(NIONetworkInterface)
+
     /// An operation that was inappropriate given the current `Channel` state was attempted.
     case inappropriateOperationForState
 
--- a/Tests/NIOTests/MulticastTest.swift
+++ b/Tests/NIOTests/MulticastTest.swift
@@ -40,7 +40,7 @@ final class MulticastTest: XCTestCase {
     }
 
     override func tearDown() {
-        try? self.group.syncShutdownGracefully()
+        XCTAssertNoThrow(try self.group.syncShutdownGracefully())
     }
 
     struct NoSuchInterfaceError: Error { }
@@ -165,14 +165,40 @@ final class MulticastTest: XCTestCase {
 
     func testCanJoinBasicMulticastGroupIPv4() throws {
         let multicastInterface = try assertNoThrowWithValue(self.interfaceForAddress(address: "127.0.0.1"))
+        guard multicastInterface.multicastSupported else {
+            // alas, we don't support multicast, let's skip but test the right error is thrown
+
+            XCTAssertThrowsError(try self.bindMulticastChannel(host: "0.0.0.0",
+                                                               port: 0,
+                                                               multicastAddress: "224.0.2.66",
+                                                               interface: multicastInterface).wait()) { error in
+                if case .some(.multicastNotSupported(let actualInterface)) = error as? ChannelError {
+                    XCTAssertEqual(multicastInterface, actualInterface)
+                } else {
+                    XCTFail("unexpected error: \(error)")
+                }
+            }
+            return
+        }
 
         // We avoid the risk of interference due to our all-addresses bind by only joining this multicast
         // group on the loopback.
-        let listenerChannel = try assertNoThrowWithValue(self.bindMulticastChannel(host: "0.0.0.0",
+        let listenerChannel: Channel
+        do {
+            listenerChannel = try assertNoThrowWithValue(self.bindMulticastChannel(host: "0.0.0.0",
                                                                                    port: 0,
                                                                                    multicastAddress: "224.0.2.66",
                                                                                    interface: multicastInterface).wait())
-        
+            // no error, that's great
+        } catch {
+            if case .some(.multicastNotSupported(_)) = error as? ChannelError {
+                XCTFail("network interface (\(multicastInterface)) claims we support multicast but: \(error)")
+            } else {
+                XCTFail("unexpected error: \(error)")
+            }
+            return
+        }
+
         defer {
             XCTAssertNoThrow(try listenerChannel.close().wait())
         }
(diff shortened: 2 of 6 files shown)


khoogheem:
That works.. I was just adding in on the MulticastTest to first check if the Interface supports with a guard so the test just skips if Multicast is not supported.

Tests/NIOTests/MulticastTest.swift
```diff
@@ -165,14 +165,29 @@ final class MulticastTest: XCTestCase {
 
     func testCanJoinBasicMulticastGroupIPv4() throws {
         let multicastInterface = try assertNoThrowWithValue(self.interfaceForAddress(address: "127.0.0.1"))
+        guard multicastInterface.multicastSupported else {
+            // alas, we don't support multicast, let's skip
+            return
```
normanmaurer: I guess there is nothing like `assumeTrue(...)` ?

https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeTrue(boolean)

Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -67,22 +67,28 @@ final class SocketOptionProviderTest: XCTestCase {
...
+                DatagramBootstrap(group: group).bind(host: "127.0.0.1", port: 0).flatMap { channel in
+                    return (channel as! MulticastChannel).joinGroup(try! SocketAddress(ipAddress: "224.0.2.66", port: 0), interface: v4LoopbackInterface).map { channel }
+                    }.wait()
+            )
+        }
+
+        // Only run The setup if the Loopback supports multicast
+        if v6LoopbackAddress.isMulticast {
+            // The IPv6 setup is allowed to fail, some hosts don't have IPv6.
+            let v6LoopbackInterface = try! System.enumerateInterfaces().filter { $0.address == v6LoopbackAddress }.first
```
normanmaurer: should we better use `assertNoThrow` ?

Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -67,22 +67,28 @@ final class SocketOptionProviderTest: XCTestCase {
...
-       let v6LoopbackInterface = try! System.enumerateInterfaces().filter { $0.address == v6LoopbackAddress }.first
-        self.ipv6DatagramChannel = try? DatagramBootstrap(group: group).bind(host: "::1", port: 0).flatMap { channel in
-            return (channel as! MulticastChannel).joinGroup(try! SocketAddress(ipAddress: "ff12::beeb", port: 0), interface: v6LoopbackInterface).map { channel }
-        }.wait()
+        // Only run The setup if the Loopback supports multicast
+        if v4LoopbackAddress.isMulticast {
+            self.ipv4DatagramChannel = try? assertNoThrowWithValue(
+                DatagramBootstrap(group: group).bind(host: "127.0.0.1", port: 0).flatMap { channel in
+                    return (channel as! MulticastChannel).joinGroup(try! SocketAddress(ipAddress: "224.0.2.66", port: 0), interface: v4LoopbackInterface).map { channel }
+                    }.wait()
```
normanmaurer: formatting looks a bit strange.. remove 4 spaces ?

Tests/NIOTests/MulticastTest.swift
weissi: @normanmaurer no :(

Tests/NIOTests/SocketOptionProviderTest.swift
weissi: fixed

Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -159,15 +167,18 @@ final class SocketOptionProviderTest: XCTestCase {
     }
 
     func testSoIpMulticastIf() throws {
-        let channel = self.ipv4DatagramChannel!
+        guard let channel = self.ipv4DatagramChannel else {
+            // no multicst support
```
normanmaurer: nit: typo.. `multicast`

Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -63,26 +63,34 @@ final class SocketOptionProviderTest: XCTestCase {
...
+                DatagramBootstrap(group: group).bind(host: "127.0.0.1", port: 0).flatMap { channel in
+                    return (channel as! MulticastChannel).joinGroup(try! SocketAddress(ipAddress: "224.0.2.66", port: 0), interface: v4LoopbackInterface).map { channel }
+                }.wait()
+            )
+        }
 
-        self.ipv4DatagramChannel = try? assertNoThrowWithValue(
-            DatagramBootstrap(group: group).bind(host: "127.0.0.1", port: 0).flatMap { channel in
-                return (channel as! MulticastChannel).joinGroup(try! SocketAddress(ipAddress: "224.0.2.66", port: 0), interface: v4LoopbackInterface).map { channel }
+        // Only run The setup if the Loopback supports multicast
```
normanmaurer: nit: the...

Tests/NIOTests/SocketOptionProviderTest.swift
```diff
@@ -63,26 +63,34 @@ final class SocketOptionProviderTest: XCTestCase {
...
-        let v4LoopbackAddress = try! SocketAddress(ipAddress: "127.0.0.1", port: 0)
-        let v6LoopbackAddress = try! SocketAddress(ipAddress: "::1", port: 0)
-        let v4LoopbackInterface = try! System.enumerateInterfaces().filter { $0.address == v4LoopbackAddress }.first!
+        let v4LoopbackAddress = try! assertNoThrowWithValue(SocketAddress(ipAddress: "127.0.0.1", port: 0))
+        let v6LoopbackAddress = try! assertNoThrowWithValue(SocketAddress(ipAddress: "::1", port: 0))
+        let v4LoopbackInterface = try! assertNoThrowWithValue(System.enumerateInterfaces().filter {
+            $0.address == v4LoopbackAddress
+        }.first)!
+
+        // Only run The setup if the Loopback supports multicast
```
normanmaurer: nit: `the ...`

Sources/NIO/Channel.swift
```diff
@@ -337,6 +337,9 @@ public enum ChannelError: Error {
     /// address.
     case illegalMulticastAddress(SocketAddress)
 
+    /// Multicast is not supported on Interface
+    case multicastNotSupported(NIONetworkInterface)
```
normanmaurer: can we also add a test that triggers this and assert that the passed in interface is what we expect ?

weissi: @normanmaurer we'd need an interface that doesn't support multicast for that, right?

normanmaurer: yeah

PR #890, opened 2019-03-08
MarkedCircularBuffer: don't used indices incorrectly

### LINKED ISSUE #836: MarkedCircularBuffer uses CircularBuffer indices wrongly

weissi:
Motivation:

MarkedCircularBuffer was assuming that indices remain stable across
mutations, that's wrong.

Modifications:

Don't store the marked index but rather the offset from the start
instead.

Result:

- MarkedCircularBuffer uses indices correctly.
- fixes #836 (thanks for @JovanMilenkovic for finding this issue)
- will enable #835 to work properly

### DIFF
--- a/Sources/NIO/MarkedCircularBuffer.swift
+++ b/Sources/NIO/MarkedCircularBuffer.swift
@@ -22,7 +22,7 @@ public struct MarkedCircularBuffer<Element>: CustomStringConvertible, Appendable
     public typealias Index = CircularBuffer<Element>.Index
 
     private var buffer: CircularBuffer<Element>
-    private var markedIndex: Index? = nil /* nil: nothing marked */
+    private var markedIndexOffset: Int? = nil /* nil: nothing marked */
 
     /// Create a new instance.
     ///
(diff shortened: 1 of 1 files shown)


normanmaurer:
@weissi shouldn't we also include a unit test ?

weissi:
> @weissi shouldn't we also include a unit test ?

@normanmaurer there are plenty of them. We can't write a unit test yet that the use was incorrect before #835 lands in its final implementation

JovanMilenkovic:
Give me a minute

weissi:
@JovanMilenkovic there's no rush on your PR, it's not SemVer major breaking so you've got time :)

JovanMilenkovic:
Pushed an update

JovanMilenkovic:
Oh cool, I thought it's blocking you.

PR #892, opened 2019-03-09
remove deprecated methods

### LINKED ISSUE #779: remove deprecated functions introduced in #776
remove deprecated functions introduced in #776 . #776 introduces deprecated functions because we removed a label into a function that was only introduced in NIO 2. The deprecation should make it much easier for users to track NIO's master. But before tagging 2.0, we really should remove them again.

The functions are in `EventLoopFuture.swift` and _not_ in the `_NIO1APIShims` module.

weissi:
Motivation:

There were some deprecated methods left on EventLoop.

Modifications:

remove them

Result:

- fixes #779

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -532,20 +532,6 @@ extension EventLoop {
     }
 }
 
-// to be removed before 2.0
-extension EventLoop {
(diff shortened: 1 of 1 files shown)


weissi:
ping @Lukasa , we shouldn't forget this one before tagging 2.0.0 :D

PR #896, opened 2019-03-11
ByteBuffer: improve and simplify get* implementations

### LINKED ISSUE #883: check ByteBuffer docs for wrong info about get*
I'm sure the ByteBuffer docs contain wrong info around the previously unsafe `get*` functions. Let's find them.

### LINKED ISSUE #884: improve sub-optimal get* implementations in ByteBuffer
ByteBuffer does repeated range checks and re-implements the logic for the get* functions repeatedly, that's bad and needs fixing, but not super high priority.

weissi:
Motivation:

For historic reasons, the get* implementations checked the same indices
repeatedly and the checks were also repetitive.

Modifications:

unify and de-deplicate get* range checks

Result:

- fixes #884
- fixes #883

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -570,9 +570,7 @@ public struct ByteBuffer {
     /// - returns: A `ByteBuffer` containing the selected bytes as readable bytes or `nil` if the selected bytes were
     ///            not readable in the initial `ByteBuffer`.
     public func getSlice(at index: Int, length: Int) -> ByteBuffer? {
-        precondition(index >= 0, "index must not be negative")
-        precondition(length >= 0, "length must not be negative")
-        guard index >= self.readerIndex && index <= self.writerIndex - length else {
+        guard index >= 0 && length >= 0 && index >= self.readerIndex && index <= self.writerIndex - length else {
             return nil
         }
         let index = _toIndex(index)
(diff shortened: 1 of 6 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -776,3 +774,14 @@ extension ByteBuffer: Equatable {
         }
     }
 }
+
+extension ByteBuffer {
+    @inlinable
+    func rangeWithinReadableBytesIfReadable(index: Int, length: Int) -> Range<Int>? {
```
Lukasa: I think this could just be `rangeWithinReadableBytes`.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -80,14 +80,11 @@ extension ByteBuffer {
...
-        precondition(length >= 0, "length must not be negative")
-        let index = index0 - self.readerIndex
-        guard index >= 0 && index <= self.readableBytes - length else {
+    public func viewBytes(at index: Int, length: Int) -> ByteBufferView? {
+        guard index >= self.readerIndex && index <= self.writerIndex - length else {
             return nil
         }
 
-        return ByteBufferView(buffer: self, range: index0 ..< index0+length)
+        return ByteBufferView(buffer: self, range: index ..< (index + length))
```
Lukasa: Can we not use the range-generation function here?

weissi: @Lukasa `ByteBufferView` uses (and has to use as `Collection`) absolute ByteBuffer indices so here, we're not interested in range relative to readable bytes but range relative to the whole ByteBuffer

Sources/NIO/ByteBuffer-core.swift
weissi: @Lukasa sounds like a much better name

Sources/NIO/ByteBuffer-views.swift
Lukasa: Ah, yeah, I see that.

This represents a difficulty with the requirement that `ByteBufferView`s only index readable bytes: specifically, what is readable can *change* after the `ByteBufferView` is created, without invalidating the `ByteBufferView` itself. I don't think this particularly matters much, as this will only happen if you're going outside the standard NIO usage patterns, but it's worth being aware of.

weissi: @Lukasa no, `ByteBufferView` holds a (private) `ByteBuffer` which is a value, so the readable bytes can't change.

PR #910, opened 2019-03-19
less confusion for Jazzy

weissi:
[looks like a SemVer major change but isn't (or it really shouldn't be)]

Motivation:

Jazzy adds everything that has a `public` modifier to the documentation,
even if a type is `internal` or `private`. So conforming and internal
type to some protocol and making all those functions `public` has the
effect of essentially documenting stuff that you can't call (as it's not
actually public).

Modifications:

- remove `public` from functions in extensions on types that aren't
  public
- change one `final public` to `public final` because Jazzy doesn't
  normalise that

Result:

better docs

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -25,10 +25,10 @@ private final class EmbeddedScheduledTask {
 }
 
 extension EmbeddedScheduledTask: Comparable {
-    public static func < (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
+    static func < (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
         return lhs.readyTime < rhs.readyTime
     }
-    public static func == (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
+    static func == (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
         return lhs === rhs
     }
 }
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -988,7 +988,7 @@ typealias ThreadInitializer = (NIOThread) -> Void
 ///            tests. In those cases it's important to shut the `MultiThreadedEventLoopGroup` down at the end of the
 ///            test. A good place to start a `MultiThreadedEventLoopGroup` is the `setUp` method of your `XCTestCase`
 ///            subclass, a good place to shut it down is the `tearDown` method.
-final public class MultiThreadedEventLoopGroup: EventLoopGroup {
+public final class MultiThreadedEventLoopGroup: EventLoopGroup {
 
     private static let threadSpecificEventLoop = ThreadSpecificVariable<SelectableEventLoop>()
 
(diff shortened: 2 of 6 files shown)


Sources/NIO/Embedded.swift
```diff
@@ -25,10 +25,10 @@ private final class EmbeddedScheduledTask {
 }
 
 extension EmbeddedScheduledTask: Comparable {
-    public static func < (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
+    static func < (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
         return lhs.readyTime < rhs.readyTime
     }
-    public static func == (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
+    static func == (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
```
weissi: `private final class EmbeddedScheduledTask`

Sources/NIO/EventLoop.swift
```diff
@@ -1148,11 +1148,11 @@ extension ScheduledTask: CustomStringConvertible {
 }
 
 extension ScheduledTask: Comparable {
-    public static func < (lhs: ScheduledTask, rhs: ScheduledTask) -> Bool {
+    static func < (lhs: ScheduledTask, rhs: ScheduledTask) -> Bool {
```
weissi: `private final class ScheduledTask`

Sources/NIO/Heap.swift
```diff
@@ -183,7 +183,7 @@ internal struct Heap<T: Comparable> {
 }
 
 extension Heap: CustomDebugStringConvertible {
-    public var debugDescription: String {
+    var debugDescription: String {
```
weissi: `internal struct Heap<T: Comparable> `

Sources/NIO/SocketChannel.swift
```diff
@@ -829,7 +829,7 @@ extension DatagramChannel: MulticastChannel {
         }
     }
 
-    public func joinGroup(_ group: SocketAddress, interface: NIONetworkInterface?, promise: EventLoopPromise<Void>?) {
+    func joinGroup(_ group: SocketAddress, interface: NIONetworkInterface?, promise: EventLoopPromise<Void>?) {
```
weissi: `final class DatagramChannel: BaseSocketChannel<Socket> `

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -235,7 +235,7 @@ extension SocketOptionProvider {
 
 
 extension BaseSocketChannel: SocketOptionProvider {
-    public func unsafeSetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) -> EventLoopFuture<Void> {
+    func unsafeSetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) -> EventLoopFuture<Void> {
```
weissi: `class BaseSocketChannel<T: BaseSocket>: SelectableChannel, ChannelCore`

Sources/NIO/PriorityQueue.swift
```diff
@@ -59,7 +59,7 @@ extension PriorityQueue: Equatable {
 }
 
 extension PriorityQueue: Sequence {
-    public struct Iterator: IteratorProtocol {
+    struct Iterator: IteratorProtocol {
```
weissi: `internal struct PriorityQueue<Element: Comparable>`

PR #913, opened 2019-03-20
Scheduled: Don't hold onto cancellationTask

weissi:
Motivation:

For whatever reason, Scheduled stored the cancellationTask in two
places, one of them it never even used :).

Modifications:

only have the promise have a reference to the cancellationTask

Result:

- don't hold onto things longer than necessary
- shrink Scheduled by one word :)

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -21,7 +21,6 @@ import Dispatch
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
     private let promise: EventLoopPromise<T>
-    private let cancellationTask: () -> Void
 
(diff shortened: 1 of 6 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -21,7 +21,6 @@ import Dispatch
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
     private let promise: EventLoopPromise<T>
-    private let cancellationTask: () -> Void
```
ddunbar: Hopefully you filed a Swift bug about why the compiler didn't catch this for you :)

weissi:
I didn’t actually. But in this case private + not computed + only written, never read could actually be a good diagnostic

PR #925, opened 2019-03-25
DEVELOPMENT-SNAPSHOT-2019-03-24-a

weissi:


Lukasa:
This should maybe just become the real Swift 5 now, right?

weissi:
> This should maybe just become the real Swift 5 now, right?

Yes! Just wanted to have a test run with the latest, just in case there's any firefighting needed. But seems all good so I'll close this and make a new one with Swift 5.

PR #926, opened 2019-03-26
use Swift 5.0-RELEASE

weissi:
Motivation:

Now that Swift 5 is released, we should use the release version.

Modifications:

switch to Swift 5.0-RELEASE.

Result:

newer, shinier


normanmaurer:
late LGTM

Issue #465, opened 2018-06-03
Please provide a WebSocket client example


gmosx:
### Expected behavior

It would be great to have an example for a WebSocket client (or even an HTTP client
 for that matter)

MrLotU:
Would be greatly appreciated 😄

tigerpixel:
While investigating this, I noticed that there were also no UDP or HTTP1 client examples. In addition, the readme for other client examples were missing. Those have now been added.

tigerpixel:
I have started coding a WebSocket client example.

In order to do so, I have written a 'HTTPClientUpgradeHandler' and 'WebClientSocketUpgrader' pair to match the 'HTTPServerUpgradeHandler' and 'WebSocketUpgrader' which already exist to handle the server upgrade. At the moment, I have mostly done this by copying the original pair and then changing/deleting code as required. So there is a decent amount of duplication. I am at the stage where it generally works for the happy path. It is very rough and ready and without any tests. I can post some code if it is felt that would help to understand my description but I am a while away from any proper PR.

Mostly though, before pressing on too much, I was hoping to get feedback this technique. Specifically, I would like to ask if this seems like a generally correct approach? Or might there be a simpler solution that I may have missed? 

Also, should I open a separate issue for the client upgrader? #306 seems as related to a client upgrader as this ticket.

thanks

Lukasa:
I don’t think we need a separate issue opened. You may want to check Vapor’s codebase, which I know has a client upgrader you could draw experience from.

weissi:
fixed in #1064

PR #933, opened 2019-03-28
UDP Echo Client Example.

tigerpixel:
UDP echo client example.

### Motivation:

There is a UDP echo server example but it does not have a matching client example.
The TCP echo server has both client and server examples. It would be preferable for the UDP example to match.
The UDP client connection method also differs from both the UDP server and the TCP client, so it is good to provide an example of this difference.

### Modifications:

Added a UDP client example and a matching readme file.
The main swift file is as similar to the TCP echo example as possible.

### Result:

There is now a clear example of how to create a simple UDP client-side connection.


tigerpixel:
I originally looked at ticket #465 (WebSocket Client Example). However, in doing so, I noticed the UDP client example was also missing. There was no ticket for it but I thought it was probably a better place to start. As there wasn’t any discussion on the topic to guide me, I tried to stick closely to the original echo client example and only make changes that are required. I am submitting the code as an early example to facilitate discussion but realize that there may still be some work to carry out.

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -0,0 +1,147 @@
...
+    
+    init(remoteAddressInitializer: @escaping () throws -> SocketAddress) {
+        self.remoteAddressInitializer = remoteAddressInitializer
+    }
+
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let envelope = self.unwrapInboundIn(data)
+        var byteBuffer = envelope.data
+
+        numBytes -= byteBuffer.readableBytes
```
weissi: we use explicit `self.` so if you wouldn't mind changing this to `self.numBytes -= byteBuffer.readableBytes`

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -0,0 +1,147 @@
...
+
+        assert(numBytes >= 0)
```
weissi: I'm wondering if we should assert here or not, because that makes this crash if the server is sending us too much data...

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -0,0 +1,147 @@
...
+        
+        if numBytes == 0 {
```
weissi: I think we should make this `if self.numBytes <= 0 {` (also not the explicit `self.`).

Sources/NIOUDPEchoClient/main.swift
tigerpixel: Sure, this is exactly the same as the TCP client. I think it is a good idea to keep them as similar as possible, so I will raise another PR to make the change there too.

Sources/NIOUDPEchoClient/main.swift
tigerpixel: sure

Sources/NIOUDPEchoClient/main.swift
tigerpixel: I agree. I don't think it adds anything. I just added it to match the TCP example. Will remove.

tigerpixel:
I added a new PR #937. This will change the TCP client to match any stylistic changes, so I am free to make any we fell necessary without compromising comparability to the TCP client.

weissi:

Sources/NIOUDPEchoClient/README.md
```diff
@@ -0,0 +1,11 @@
+# NIOEchoClient
```
weissi: actually, shouldn't this mention `NIOUDPEchoClient` (I think the `UDP` is missing)?

Sources/NIOUDPEchoClient/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This sample application provides a simple UDP echo client that will send a single line to a UDP echo server and wait for a response. Invoke it using one of the following syntaxes:
+
+```bash
+swift run NIOEchoClient # Connects to a server on ::1, server port 9999 and listening port 8888.
+swift run NIOEchoClient 9899 9888 # Connects to a server on ::1, server port 9899 and listening port 9888
+swift run NIOEchoClient /path/to/unix/server/socket /path/to/unix/listening/socket # Connects to a server using the first UNIX socket, listening on the second socket.
```
weissi: let's remove the UNIX Domain Socket line from the UDP thing

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -0,0 +1,145 @@
+ //===----------------------------------------------------------------------===//
+ //
+ // This source file is part of the SwiftNIO open source project
+ //
+ // Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: this should be `2017-2019` or just `2019`

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -0,0 +1,145 @@
+ //===----------------------------------------------------------------------===//
```
weissi: @tigerpixel ha, the license headers aren't missing but the whole file got an extra space (` `) at the beginning of every line ;)

weissi:

Sources/NIOUDPEchoClient/main.swift
weissi: I fixed it

Sources/NIOUDPEchoClient/main.swift
weissi: fixed that

Sources/NIOUDPEchoClient/README.md
weissi: fixed that

Sources/NIOUDPEchoClient/README.md
weissi: fixed that

weissi:

PR #937, opened 2019-03-29
Update style in TCP Client and HTTP1Server examples.

tigerpixel:
Update the style in the TCP Client example and HTTP1Server example to match the UDP client example.

### Motivation:

There is a UDP echo client which is very similar to the NIOEchoClient.
The styles of these two examples should match closely so ‘diff’ comparing will show the pertinent API differences.

### Modifications:

Add 'self.' to member variables in the TCP echo client.
Add a 'self.' to a member variable in the HTTP1 server.
Remove an assert from the echo client.

### Result:

The code in the TCP and UDP examples now matches and fits the SwiftNIO style.


tigerpixel:
I have a separate pull request out (#933). This PR adds a UDP client example. During the code review stage, a few style changes were requested, which would make it differ from the TCP example. It seemed like a good idea to change the TCP example too. That other PR has not finished code review, so it is probably not a good idea to merge this until that is the case. Further changes may be needed and could easily go into this PR.

tigerpixel:
Only because you hadn't approved the same changes in the other file. They are approved now, so I have removed the 'do not merge' tag. I am happy to merge when you are.

weissi:

PR #939, opened 2019-04-01
B2MD verifier

weissi:
Motivation:

When writing B2MDs, there are a couple of scenarios that always need to
be tested: firehose feeding, drip feeding, many messages, ...

It's tedious writing those tests over and over again for every B2MD.

Modifications:

Add a simple B2MD verifier that users can use in unit tests.

Result:

Hopefully fewer bugs in B2MDs.


Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
...
+    ///     var exampleInput1 = channel.allocator.buffer(capacity: 16)
+    ///     exampleInput1.writeString("example-in1")
+    ///     var exampleInput2 = channel.allocator.buffer(capacity: 16)
+    ///     exampleInput2.writeString("example-in2")
+    ///     let expectedInOuts = [(exampleInput1, ExampleDecoderOutput("1")),
+    ///                           (exampleInput2, ExampleDecoderOutput("2"))
+    ///                          ]
+    ///     XCTAssertNoThrow(try ByteToMessageHandler<ExampleDecoder>.verifyDecoder(inputOutputPairs: expectedInOuts,
+    ///                                                                             decoderFactory: { ExampleDecoder() }))
+    public static func verifyDecoder<Decoder: ByteToMessageDecoder>(inputOutputPairs: [(ByteBuffer, [Decoder.InboundOut])],
```
normanmaurer: it's a bit odd to have this in the "normal" code-base... Is there a way to have "test-only" code, so users would explicit import it in test-code only. Maybe the only way would be a new "product"...

weissi: > it's a bit odd to have this in the "normal" code-base... Is there a way to have "test-only" code

No, unfortunately not.

>  Maybe the only way would be a new "product"...

We could do that. @Lukasa ?

Lukasa: I think a new product would probably be sensible.

Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
         }
     }
 }
+
+enum ByteToMessageDecoderVerificationError<OutputType>: Error {
+    case wrongDecodingResult(actual: OutputType, expected: OutputType)
+    case unexpectedDecodingResult(OutputType)
```
Lukasa: This may want a rename to better distinguish `wrongDecodingResult` from `unexpectedDecodingResult`.

Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
...
+
+        let decoder: Decoder = decoderFactory()
+        let channel = EmbeddedChannel(handler: ByteToMessageHandler<Decoder>(decoder))
+
+        try verifySimple(channel: channel)
+        try verifyDripFeed(channel: channel)
+        try verifyManyAtOnce(channel: channel)
+
+
+        if case .leftOvers(inbound: let ib, outbound: let ob, pendingOutbound: let pob) =  try channel.finish() {
```
Lukasa: Nit: too much space on the right side of `=`.

Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
...
+                }
+            }
+        }
+
+        func verifyManyAtOnce(channel: EmbeddedChannel) throws {
+            var overallBuffer = channel.allocator.buffer(capacity: 1024)
+            var overallExpecteds: [Out] = []
+
+            for _ in 0..<10 {
+                for (var input, expectedOutputs) in inputOutputPairs.shuffled() {
```
Lukasa: It’s not necessarily the case that you can assume that you can shuffle the inputs and get the same result: for example, the HTTPDecoder cannot tolerate that.

Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
...
+            for (input, expectedOutputs) in inputOutputPairs.shuffled() {
+                try channel.writeInbound(input)
+                for expectedOutput in expectedOutputs {
+                    let actualOutput = try channel.readInbound(as: Decoder.InboundOut.self)
+                    guard actualOutput == expectedOutput else {
+                        throw ByteToMessageDecoderVerificationError.wrongDecodingResult(actual: actualOutput,
+                                                                                        expected: expectedOutput)
+                    }
+                }
+                let actualExtraOutput = try channel.readInbound(as: Decoder.InboundOut.self)
```
Lukasa: Should this be `while let` to obtain the full overproduction?

Sources/NIO/Codec.swift
weissi: @Lukasa In general, you're right. B2MDs that retain state across one message can't be used like this. I think this is still useful as is to verify B2MDs that don't do that. We could offer some form of configuration but when I thought about it, it quickly became too complicated. For HTTPDecoder this means that this B2MD verifier will only work for keep-alive messages where I can shuffle.

Another issue is that some B2MDs (like HTTPDecoder) generate different outputs depending on whether the input came in a large chunk or was drip-fed. HTTP bodies for example are just handed through by HTTPDecoder in whatever shape they arrived. Hence, I'm using 1 byte bodies in the tests here.

I should probably make this more clear in the documentation but this is meant as a test for quite straightforward B2MDs, the ones that don't retain any state across messages and I think that's quite useful. This isn't supposed to replace the regular unit tests but it's supposed to give you a baseline. And most B2MDs (including HTTPDecoder) can be used in this way.

Sources/NIO/Codec.swift
```diff
@@ -674,3 +674,118 @@ extension MessageToByteHandler {
...
+    ///                           (exampleInput2, ExampleDecoderOutput("2"))
+    ///                          ]
+    ///     XCTAssertNoThrow(try ByteToMessageHandler<ExampleDecoder>.verifyDecoder(inputOutputPairs: expectedInOuts,
+    ///                                                                             decoderFactory: { ExampleDecoder() }))
+    public static func verifyDecoder<Decoder: ByteToMessageDecoder>(inputOutputPairs: [(ByteBuffer, [Decoder.InboundOut])],
+                                                                    decoderFactory: @escaping () -> Decoder) throws where Decoder.InboundOut: Equatable {
+        typealias Out = Decoder.InboundOut
+
+        func verifySimple(channel: EmbeddedChannel) throws {
+            for (input, expectedOutputs) in inputOutputPairs.shuffled() {
```
ianpartridge: Wouldn't this make tests non-deterministic?

weissi: @ianpartridge Yes, it does. The expectation is that B2MDs shouldn't mind that and if they do that's a bug. I could add (in the error) the specific input that was sent maybe?

It's a similar issue like when relying on the order of the key/value pairs in a dictionary I guess.

ianpartridge: Yes I'm just thinking about how users could debug their B2MD when faced with an order-dependent test failure.

weissi: @ianpartridge well, it's one B2MD that they'd need to look at and it needs to be state that persists across two messages (of which there should be very little). However, the whole thing could be solved by just giving the user the data in a failed attempt. Then they can make that  particular input a regression test by itself.

ianpartridge: Yep, sounds good to me.

Lukasa: Can we instead accept a hash to affect the shuffling, and then print that hash in the event of test failure?

Sources/NIO/Codec.swift
Lukasa: Ok, I can get behind accepting this as a limitation for now.

Sources/NIO/Codec.swift
weissi: @Lukasa Sure, that works too.

weissi: @Lukasa / @ianpartridge I added the delivery of the inputs, does that work?

Sources/NIO/Codec.swift
weissi: done

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -0,0 +1,211 @@
...
+        return try ByteToMessageDecoderVerifier.verifyDecoder(inputOutputPairs: ioPairs, decoderFactory: decoderFactory)
+    }
+
+    /// Verifies a `ByteToMessageDecoder` by performing a number of tests.
+    ///
+    /// This method is mostly useful in unit tests for `ByteToMessageDecoder`s. It feeds the inputs from
+    /// `inputOutputPairs` into the decoder in various ways and expects the decoder to produce the outputs from
+    /// `inputOutputPairs`.
+    ///
+    /// The verification is performing various tests, for example:
```
Lukasa: Nit: "The verification performs various tests, for example:"

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -0,0 +1,211 @@
...
+    ///     var exampleInput2 = channel.allocator.buffer(capacity: 16)
+    ///     exampleInput2.writeString("example-in2")
+    ///     let expectedInOuts = [(exampleInput1, ExampleDecoderOutput("1")),
+    ///                           (exampleInput2, ExampleDecoderOutput("2"))
+    ///                          ]
+    ///     XCTAssertNoThrow(try ByteToMessageDecoderVerifier.verifyDecoder(inputOutputPairs: expectedInOuts,
+    ///                                                                     decoderFactory: { ExampleDecoder() }))
+    public static func verifyDecoder<Decoder: ByteToMessageDecoder>(inputOutputPairs: [(ByteBuffer, [Decoder.InboundOut])],
+                                                                    decoderFactory: @escaping () -> Decoder) throws where Decoder.InboundOut: Equatable {
+        typealias Out = Decoder.InboundOut
```
Lukasa: Nit: this `typealias` is used inconsistently in this function: we may as well use it everywhere, no?

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -0,0 +1,211 @@
...
+        func verifyDripFeed(channel: RecordingChannel) throws {
+            for _ in 0..<10 {
+                for (input, expectedOutputs) in inputOutputPairs.shuffled() {
+                    for c in input.readableBytesView {
+                        var buffer = channel.allocator.buffer(capacity: 12)
+                        buffer.writeString("BEFORE")
+                        buffer.writeInteger(c)
+                        buffer.writeString("AFTER")
+                        buffer.moveReaderIndex(forwardBy: 6)
+                        buffer.moveWriterIndex(to: buffer.readerIndex + 1)
```
Lukasa: Might it be clearer to move the `buffer` setup dance outside the loop, and then just have the loop body do `setInteger(_:at:)` instead?

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        let decoder: Decoder = decoderFactory()
+        let channel = RecordingChannel(EmbeddedChannel(handler: ByteToMessageHandler<Decoder>(decoder)))
+
+        try verifySimple(channel: channel)
+        try verifyDripFeed(channel: channel)
+        try verifyManyAtOnce(channel: channel)
+
+        if case .leftOvers(inbound: let ib, outbound: let ob, pendingOutbound: let pob) =  try channel.finish() {
```
Lukasa: Super duper nit: delete one space character before `try`. Sorry, if I wasn't already leaving feedback I'd have let this one slide. 😆

Tests/NIOTestUtilsTests/ByteToMessageDecoderVerifierTest.swift
```diff
@@ -0,0 +1,167 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+@testable import NIO
+import NIOTestUtils
+import XCTest
+
+typealias VerificationError = ByteToMessageDecoderVerifier.VerificationError<String>
+
+class UtilitiesTest: XCTestCase {
```
Lukasa: Let's name this `ByteToMessageDecoderVerifierTest` instead.

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
weissi: done

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
weissi: done

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
weissi: done

Tests/NIOTestUtilsTests/ByteToMessageDecoderVerifierTest.swift
weissi: oopsie, copy & pasta for the win

weissi:

weissi:
```
16:45:27 removal of container a462f06554d90c3e6f6b04d1ddc6a72ee249d60df04e493d84ced4bffed4a67c is already in progress
16:45:27 Build step 'Execute shell' marked build as failure
```

PR #941, opened 2019-04-02
Add a HTTP1 client example with readme.

tigerpixel:
Add an HTTP1 client example. It is only a simple echo which closely resembles the TCP and UDP echo examples.

### Motivation:

To provide a simple example of the HTTP1 client pipeline.
To create a comparable example to the TCP and UDP demos to show what extra steps are needed for an HTTP level request and response.  

### Modifications:

Adds a NIOHTTP1Client example in a main file.
Adds a readme to the NIOHTTP1Client example.

### Result:

There is now a simple example of how to send an HTTP1 client request to the server.


Sources/NIOHTTP1Client/main.swift
```diff
@@ -0,0 +1,134 @@
...
+
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        print("error: ", error)
+
+        // As we are not really interested getting notified on success or failure we just pass nil as promise to
+        // reduce allocations.
+        context.close(promise: nil)
+    }
+
+    public func channelActive(context: ChannelHandlerContext) {
```
Lukasa: Should we guard against being added to an already-active channel here?

Lukasa: So in *general* it is possible that a channel handler is added to a stream *after* that stream has fired `channelActive`. In this specific case it’s not possible in the current implementation, but certain changes (e.g. for higher level protocol negotiation) could make it possible.

The way this is detected is by checking whether the channel is active in `handlerAdded`. Of course, caution needs to be required to avoid double-activating. This may all be more complexity than we need in this example, but it’s a fairly easy mistake to make.

tigerpixel: Thanks. I am actually doing this one as a prequel to the WebSockets client, so that very helpful for that in particular. I am not sure we need it for this example though. It doesn't really have any error checking. My impression (perhaps incorrectly) is that these are just 'as simple as possible' examples to get people started with a pipeline that can expand upon. I am happy to add it if you think it is required, but I would imagine that, in that case, we should probably add it to all client examples.

Lukasa: @weissi thoughts?

weissi: @Lukasa / @tigerpixel I'd say it's fine here, we can always improve

weissi:

weissi:

PR #953, opened 2019-04-06
Reorder ‘channel active’ calls

tigerpixel:
Reorder ‘channel active’ calls to the same order as `_ChannelInboundHandler` and their likely chronological order.

### Motivation:

When first viewing the example classes, coming to the ‘read’ method first, leaves the subject unclear as to what the method is ‘reading’.
It is preferable to view something being sent first, and then to view the reading of the response.
It also matches the call order against the protocol making it a little easier for those unfamiliar with the protocol to see which methods have been implemented.

### Modifications:

Moved channel active calls to be top of the class.
Despite the diff, there are no actual code modifications. Just moving methods.
UDP Client changed to indent using spaces to match the rest of the project. Incidental change. 

### Result:

The examples are slighter clearer to read, particularly for newcomers to swift-no as the calls are in a logical chronological order.

### DIFF
--- a/Sources/NIOChatServer/main.swift
+++ b/Sources/NIOChatServer/main.swift
@@ -54,6 +54,31 @@ final class ChatHandler: ChannelInboundHandler {
     // All access to channels is guarded by channelsSyncQueue.
     private let channelsSyncQueue = DispatchQueue(label: "channelsQueue")
     private var channels: [ObjectIdentifier: Channel] = [:]
+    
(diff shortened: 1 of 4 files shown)


tigerpixel:
Please note:
a) There are no real code changes, just a rather opinionated change to the method order.  
b) This is a very small and rather trivial change.
If the core team does not agree with this change, for any reason please feel free to just reject the request. Don't feel like you need to spend a lot of time justifying your response.

weissi:

PR #952, opened 2019-04-05
swift 5 docker support

tomerd:
Motivation: easily build and test swift-nio v1 using swift 5 docker

Modifications: adjust dockerfile and add docker-compose file for ubuntu 18.04, swift 5

PR #954, opened 2019-04-08
alloc tests: hook posix_memalign & reallocf (#782)

weissi:
Motivation:

Swift started using posix_memalign very recently (Jan 2019) so we need
to hook that too to get correct results.

Modifications:

Hook posix_memalign, reallocf & valloc

Result:

alloc tests should pass again

(cherry picked from commit de8d3e260e54ee27d55d82d48a552236d57a704e)


Lukasa:
We're ignoring the Swift 5 test failure here: @weissi assures me that he expected that failure and that it's not relevant right now.

weissi:
5.0 failure is expected because the docker file only comes in #952 (which can't go in ahead of this one because this backports the allocation counter fixes from `master`)

PR #957, opened 2019-04-08
optional max payload size for ByteToMessageDecoder

tomerd:
Motivation:

ByteToMessageDecoder aggregate data in memory as part of their normal operation. the ability to limit how much they aggregate is critical in many real-life applications

Modifications:

* add optional maximumBufferSize argument to ByteToMessageDecoder initializer
* test for buffer size when maximumBufferSize is set and throw ByteToMessageDecoderError.payloadTooLarge error

Result:

users can limit how much memory ByteToMessageDecoder takes and handle the exception on their end

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -36,11 +36,8 @@ public enum ByteToMessageDecoderError: Error {
 }
 
 // TODO (tomer): Merge into ByteToMessageDecoderError next major version
-/// Data errors thrown by `ByteToMessageDecoder`s.
-public enum ByteToMessageDecoderDataError: Error {
-    /// This error can be thrown by `ByteToMessageDecoder`s if the incoming payload is larger than the max specified.
-    case payloadTooLarge
-}
+/// This error can be thrown by `ByteToMessageDecoder`s if the incoming payload is larger than the max specified.
+public struct ByteToMessageDecoderPayloadTooLargeError: Error {}
 
 /// `ByteToMessageDecoder`s decode bytes in a stream-like fashion from `ByteBuffer` to another message type.
 ///
@@ -366,6 +363,11 @@ public final class ByteToMessageHandler<Decoder: ByteToMessageDecoder> {
     private var seenEOF: Bool = false
     private var selfAsCanDequeueWrites: CanDequeueWrites? = nil
 
+    /// Initialize a `ByteToMessageHandler`.
+    ///
+    /// - parameters:
+    ///     - decoder: The `ByteToMessageDecoder` to decode the bytes into message.
+    ///     - maximumBufferSize: The maximum number of bytes to aggregate in-memory.
     public init(_ decoder: Decoder, maximumBufferSize: Int? = nil) {
         self.decoder = decoder
         self.maximumBufferSize = maximumBufferSize
@@ -457,16 +459,18 @@ extension ByteToMessageHandler {
         var allowEmptyBuffer = decodeMode == .last
         while (self.state.isActive && self.removalState == .notBeingRemoved) || decodeMode == .last {
             let result = try self.withNextBuffer(allowEmptyBuffer: allowEmptyBuffer) { decoder, buffer in
-                if let maximumBufferSize = self.maximumBufferSize, buffer.readableBytes > maximumBufferSize {
-                    throw ByteToMessageDecoderDataError.payloadTooLarge
-                }
+                let decoderResult: DecodingState
                 if decodeMode == .normal {
                     assert(self.state.isActive, "illegal state for normal decode: \(self.state)")
-                    return try decoder.decode(context: context, buffer: &buffer)
+                    decoderResult = try decoder.decode(context: context, buffer: &buffer)
                 } else {
                     allowEmptyBuffer = false
-                    return try decoder.decodeLast(context: context, buffer: &buffer, seenEOF: self.seenEOF)
+                    decoderResult = try decoder.decodeLast(context: context, buffer: &buffer, seenEOF: self.seenEOF)
                 }
+                if decoderResult == .needMoreData, let maximumBufferSize = self.maximumBufferSize, buffer.readableBytes > maximumBufferSize {
+                    throw ByteToMessageDecoderPayloadTooLargeError()
+                }
+                return decoderResult
             }
             switch result {
             case .didProcess(.continue):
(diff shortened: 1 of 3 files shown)


Sources/NIO/Codec.swift
```diff
@@ -33,6 +33,9 @@ public enum ByteToMessageDecoderError: Error {
     /// This error can be thrown by `ByteToMessageDecoder`s if there was unexpectedly some left-over data when the
     /// `ByteToMessageDecoder` was removed from the pipeline or the `Channel` was closed.
     case leftoverDataWhenDone(ByteBuffer)
+
+    /// This error can be thrown by `ByteToMessageDecoder`s if the incoming payload is larger than the max defined.
+    case payloadTooLarge
```
weissi: this is major version breaking, you need to add a new error type for this

Sources/NIO/Codec.swift
```diff
@@ -346,6 +349,7 @@ public final class ByteToMessageHandler<Decoder: ByteToMessageDecoder> {
     }
 
     internal private(set) var decoder: Decoder? // only `nil` if we're already decoding (ie. we're re-entered)
+    private let maximumBufferSize: Int
```
weissi: this should be stored as `Int?`

Sources/NIO/Codec.swift
```diff
@@ -448,6 +453,9 @@ extension ByteToMessageHandler {
         var allowEmptyBuffer = decodeMode == .last
         while (self.state.isActive && self.removalState == .notBeingRemoved) || decodeMode == .last {
             let result = try self.withNextBuffer(allowEmptyBuffer: allowEmptyBuffer) { decoder, buffer in
+                if self.maximumBufferSize > 0, buffer.readableBytes >= self.maximumBufferSize {
```
weissi: shouldn't this be `> self.maximumBufferSize`?

Sources/NIO/Codec.swift
tomerd: no strong opinion, do you prefer to carry the optional and optional check?

Sources/NIO/Codec.swift
tomerd: no strong opinion, do you prefer a separate error to get in earlier?

Sources/NIO/Codec.swift
tomerd: depends on how you think of max? can go either way. lmk if you prefer = to be fine

Sources/NIO/Codec.swift
Lukasa: I think it'd be a good idea: if we don't use a separate error type this will have to sit here for probably at least a year.

Sources/NIO/Codec.swift
Lukasa: I think I'd do the optional: the optional check is not more expensive or syntactically heavy than the current -1 check you're doing now, and more accurately communicates the intent.

Sources/NIO/Codec.swift
Lukasa: My view is the the maximum is an inclusive bound: if told me in a conversation that your maximum buffer size is 1024 bytes, I would expect you to accept a 1024 byte buffer.

Sources/NIO/Codec.swift
```diff
@@ -448,6 +453,9 @@ extension ByteToMessageHandler {
...
+                    throw ByteToMessageDecoderError.payloadTooLarge
+                }
```
weissi: could we move the check to after we try to decode? Just to give the decoder 'one last chance' because we loaded everything in memory already anyway so we might just as well give the decoder a try.

Just be sure you _don't_ use `defer`, this check must not run if we already threw an error

tomerd:
@weissi @Lukasa feedback addressed + rebased. still one open question re the timing of the check. i can go either way, tho slightly prefer to do it early

Sources/NIO/Codec.swift
weissi: @tomerd sure, but we load n+100 into memory _anyway_. So sure, the user intent is to never load more than `n` bytes into memory. But the current implementation will still load `n+100` into memory but it _won't_ give the decoder a last chance. I'm happy with saying that `n` isn't an absolute limit but best effort but we should give the decoder one last chance.

It's totally fine for this to be a best effort limit because we can't really implement anything else anyway.

Sources/NIO/Codec.swift
```diff
@@ -358,8 +366,9 @@ public final class ByteToMessageHandler<Decoder: ByteToMessageDecoder> {
     private var seenEOF: Bool = false
     private var selfAsCanDequeueWrites: CanDequeueWrites? = nil
 
-    public init(_ decoder: Decoder) {
+    public init(_ decoder: Decoder, maximumBufferSize: Int? = nil) {
```
weissi: @tomerd would you mind adding a doc comment here?

Sources/NIO/Codec.swift
tomerd: yes, you are absolutely correct from an “how much data is loaded to memory” pov. i am also thinking about the pov of the user wanting to only ever process n bytes and no more, for example if they have downstream logic that ties the expected max size to some preallocated space or simply tests for honoring max size

we could make this more complicated and introduce a behavior of type but i feel it would be an overshot. another alternative could be to truncate at max size if we buffered over it

other ideas/thoughts?

weissi: @tomerd the user should never expect 'X and no more' because that is very rarely what we can do. If their parsing algorithm is so shit that they can't tolerate a few extra bytes, then I believe they should implement a check themselves using `guard buffer.readableBytes <= whatICanTolerate else { throw .... }`

Sources/NIO/Codec.swift
```diff
@@ -35,6 +35,13 @@ public enum ByteToMessageDecoderError: Error {
     case leftoverDataWhenDone(ByteBuffer)
 }
 
+// TODO (tomer): Merge into ByteToMessageDecoderError next major version
+/// Data errors thrown by `ByteToMessageDecoder`s.
+public enum ByteToMessageDecoderDataError: Error {
+    /// This error can be thrown by `ByteToMessageDecoder`s if the incoming payload is larger than the max specified.
+    case payloadTooLarge
```
weissi: there's no real point in making this an enum because it has just one case and we can never add a case. This should be a struct:

```swift
public struct ByteToMessageDecoderPayloadTooLargeError: Error {}
```

tomerd:
@weissi updated

PR #947, opened 2019-04-05
Backport `whenAll[Succeed]` to NIO 1

MrMage:
See https://github.com/apple/swift-nio/pull/943. `{and,when}AllComplete` were omitted because `Result` is not available in Swift 4, and `whenAll` should be sufficient for Vapor's needs. (I guess we don't want to add more functionality to NIO 1 than necessary.)

Performance results on NIO 1 (nearly identical to NIO 2):

```
measuring: future_whenall_100k_immediately_succeeded_off_loop: 0.13129901885986328, 0.13130605220794678, 0.12960290908813477, 0.12814700603485107, 0.1288670301437378, 0.13253700733184814, 0.1318880319595337, 0.13585007190704346, 0.13087499141693115, 0.13029694557189941, 
measuring: future_whenall_100k_immediately_succeeded_on_loop: 0.13315296173095703, 0.12973999977111816, 0.13161206245422363, 0.13184499740600586, 0.13040506839752197, 0.13026797771453857, 0.12884008884429932, 0.129052996635437, 0.12953698635101318, 0.12920796871185303, 
measuring: future_whenall_100k_deferred_off_loop: 0.8608689308166504, 0.8739509582519531, 0.8166019916534424, 0.8779599666595459, 0.8566949367523193, 0.834915041923523, 0.8747600317001343, 0.8622258901596069, 0.8150960206985474, 0.8113300800323486, 
measuring: future_whenall_100k_deferred_on_loop: 0.20211589336395264, 0.20115602016448975, 0.20297491550445557, 0.2031630277633667, 0.20395588874816895, 0.1977750062942505, 0.20329594612121582, 0.20048999786376953, 0.19947195053100586, 0.20047104358673096, 
```

No baseline because `whenAll` did not previously exist in NIO 1.

PR #963, opened 2019-04-11
Make `EventLoopFuture.hopTo(eventLoop:)` public again.

MrMage:
This was accidentally made `internal` with https://github.com/apple/swift-nio/pull/947.

Maybe there should be a separate test case file (with a non-`@testable` `import`) that verifies that all `public` interfaces are still present?

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -1053,7 +1053,7 @@ extension EventLoopFuture {
     /// - parameters:
     ///     - target: The `EventLoop` that the returned `EventLoopFuture` will run on.
     /// - returns: An `EventLoopFuture` whose callbacks run on `target` instead of the original loop.
-    func hopTo(eventLoop target: EventLoop) -> EventLoopFuture<T> {
+    public func hopTo(eventLoop target: EventLoop) -> EventLoopFuture<T> {
         if target === self.eventLoop {
             // We're already on that event loop, nothing to do here. Save an allocation.
             return self


MrMage:

normanmaurer:
Verify API / ABI breakage would be good for sure... I wonder if there is anything similar to what we have in java to verify this sort of stuff in netty:

https://siom79.github.io/japicmp/MavenPlugin.html

Basically it always confirms the current "branch" to the last stable release and fails the build if there is an API / ABI issue.

Lukasa:
Verifying ABI breakage is not relevant due to the lack of stable ABI on Linux, so we don't need to worry about that.

Verifying API breakage in Swift is an unsolved problem. We'd have to invest time in writing tooling that presumably uses SwiftSyntax or something similar to generate an idea of what the public interface to SwiftNIO even *is*, and then to write diffing tools that understand the semantic nature of the changes. This is a *hard* problem.

Writing tests that don't use `@testable` imports is always better: we should consider whether we can split these future tests up to avoid doing so.

MrMage:
> Verifying API breakage in Swift is an unsolved problem. We'd have to invest time in writing tooling that presumably uses SwiftSyntax or something similar to generate an idea of what the public interface to SwiftNIO even _is_, and then to write diffing tools that understand the semantic nature of the changes. This is a _hard_ problem.

I think Xcode sometimes gets confused when Cmd-clicking a library symbol and only shows that library's exports, rather than the implementation source file. Maybe that could be a place to start. I could imagine that it would be easier to simply wait until that tooling has improved though, rather than invest unreasonable amounts of time now.

weissi:
Ouch! I think the Swift team somewhere have a public API diffing tool. We need that kind of thing.

PR #964, opened 2019-04-11
Make hopTo public again

ffried:
This makes the `hopTo(eventLoop:)` function `public` again.

### Motivation:

In combination of 6b914e891f6c30f332b14dc15bcc78a1985942f5 and 26bee216743165ee6ce32e27a7c6e82a2b0e52e5, the `hopTo(eventLoop:)` function was accidentally made `internal`, because the `public` modified was removed from both, the `extension` and the `func` itself.

### Modifications:

Re-add `public` to the definition of `hopTo(eventLoop:)`.

### Result:

Projects using `hopTo(eventLoop:)` will build fine again. They're broken with the current 1.14.0 release.


ffried:
Ugh, sorry. Missed that there's already #963. Closing this.

PR #967, opened 2019-04-11
Add a Channel Handler that closes the channel

### LINKED ISSUE #841: Provide ChannelHandler that closes Channel on unhandled error.
As we change #600, many users will start leaking Channels because they don't close on error. We should provide a ChannelHandler that closes on unhandled error that is typically added last to the pipeline.

Palleas:
Motivation:

Because of #600, many users will start leaking Channels because they
don't close on error.

Modifications:

* Add ChannelHandlerError
* Add ChannelHandlerErrorTest

Result:

User will have access to an handler to use to automatically close
the channel in case of unhandled errors

Closes #841.


Sources/NIO/ChannelHandlerError.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+final public class ChannelHandlerError: ChannelInboundHandler {
```
Lukasa: This should be called `CloseOnErrorHandler` or something similar.

Sources/NIO/ChannelHandlerError.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
```
Lukasa: This import of `Foundation` is unnecessary and can be removed. Additionally, you’ll need to place a license header in this file, as seen in the others.

Tests/NIOTests/ChannelHandlerErrorTest.swift
```diff
@@ -0,0 +1,46 @@
+import XCTest
```
Lukasa: This file also needs a license header.

Sources/NIO/ChannelHandlerError.swift
Lukasa: Please also update the file names with the new handler name.

Lukasa: Actually @normanmaurer is there prior art in Netty here?

Palleas:
@Lukasa All done, thanks for the feedback!

Sources/NIO/ChannelHandlerError.swift
normanmaurer: We do not include such a handler in Netty, so no :)

Sources/NIO/CloseOnErrorHandler.swift
```diff
@@ -0,0 +1,23 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
normanmaurer: 2019 ?

Sources/NIO/ChannelHandlerError.swift
weissi: we also need to prefix this with `NIO`. I.e. `NIOCloseOnErrorHandler` because that's the guarantee we gave in the [public API declaration](https://github.com/apple/swift-nio/blob/master/docs/public-api.md).

The other question is: is this `swift-nio` or `swift-nio-extras`?

Sources/NIO/CloseOnErrorHandler.swift
Palleas: Oops, I wondered when I copy/pasted 🙈  Fixed!
Do you want me to squash into one commit since only the first matches the template?

Lukasa: Squashing the commits would be great!

Sources/NIO/ChannelHandlerError.swift
Lukasa: Good question. I could go either way there.

Sources/NIO/CloseOnErrorHandler.swift
Palleas: Done!

Sources/NIO/ChannelHandlerError.swift
Palleas: Added the NIO prefix, for the rest just let me know 😀

Sources/NIO/NIOCloseOnErrorHandler.swift
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+final public class NIOCloseOnErrorHandler: ChannelInboundHandler {
```
weissi: - sorry, super nit but Jazzy renders this as is, would you mind making it `public final` (instead of `final public`)
- this needs documentation for all the `public` bits that don't come from at least `ChannelInboundHandler`, ie. the class and the `init`

Palleas: Done! Let me know if you want me to put more details in the docs.

Palleas:
@normanmaurer @Lukasa Fixed! Sorry I should have pinged you!

Sources/NIO/NIOCloseOnErrorHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+/// A `ChannelInboundHandler` that closes the channel when an error is caught
+public final class NIOCloseOnErrorHandler: ChannelInboundHandler {
+
+    public typealias InboundIn = NIOAny
+    
+    /// Initialize a `NIOCloseOnErrorHandler`
+    public init() {}
+    
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        context.close(promise: nil)
```
Lukasa: Here's a question: should this forward the error on? @weissi?

weissi: @Lukasa I think I agree,  yes, don't think this could do anything harmful. In most cases absolutely nothing will happen (because this will likely be the last handler) but in certain cases it might be better to give the user visibility.

Palleas: So, forward then close? I’m guessing the order matters here

Palleas: @Lukasa @weissi tried something in 061f1632, is that the right approach?

weissi: yes, that works!

Palleas: Perfect, squashed into one commit!

Palleas:
@Lukasa @normanmaurer all squashed and ready for another review 🦸🏻‍♂️

Lukasa:

weissi:
@Palleas your pull request uses windows newlines (`\r\n`) instead of UNIX newlines (`\n`) and hence it doesn't pass the license header check tests. Could you run `dos2unix Tests/NIOTests/NIOCloseOnErrorHandlerTest.swift Sources/NIO/NIOCloseOnErrorHandler.swift` and add all the changes to your PR?

You might also want to fix your text editor to produce UNIX newlines

Palleas:
Uh, how did that happen 🤨 I’ll fix that today!

Palleas:
@Lukasa @weissi all done!

Lukasa:

Lukasa:

PR #974, opened 2019-04-18
Add a way to get a handler from a pipeline

### LINKED ISSUE #966: ChannelPipeline should have a way to give you a handler directly instead of via ChannelHandlerContext
Currently (due to #508), to get to the HTTP/2 multiplexer you need to do something like

```swift
bootstrap.connect(...).flatMap { channel in
    return channel.pipeline.context(handlerType: HTTP2StreamMultiplexer.self).map { $0.handler as! HTTP2StreamMultiplexer }.flatMap { multiplexer in
        // here we've got `channel` and `multiplexer`
        // [...]
    }
}
```

but we could offer

```swift
extension ChannelPipeline {
    func handler<T: ChannelHandler>(type: T) -> EventLoopFuture<T> {
        return self.context(handlerType: T.self).map { $0.handler as! T }
    }
}
```

Palleas:
Motivation:

https://github.com/apple/swift-nio/issues/966

Modifications:

- Add a `handler(type:)` method on ChannelPipeline
- Add a test in ChannelPipelineTest

Result:

Provides a nicer way to get a typed handler directly from the pipeline

Closes #966

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the first `ChannelHandler` of the given type.
+    ///
+    /// - parameters:
+    ///     - handlerType: the type of `ChannelHandler` to return.
+    public func handler<Handler: ChannelHandler>(type _: Handler.Type) -> EventLoopFuture<Handler> {
+        return self.context(handlerType: Handler.self).map { context in
+            guard let typedContext = context.handler as? Handler else {
+                preconditionFailure("Expected channel handler of type \(Handler.self), got \(type(of: context.handler)) instead.")
+            }
+            
+            return typedContext
+        }
+    }
+    
     private struct ChannelHandlerDebugInfo {
         let handler: ChannelHandler
         let name: String
--- a/Tests/NIOTests/ChannelPipelineTest.swift
+++ b/Tests/NIOTests/ChannelPipelineTest.swift
@@ -69,6 +69,67 @@ private extension EmbeddedChannel {
 
 class ChannelPipelineTest: XCTestCase {
 
+    class SimpleTypedHandler1: ChannelInboundHandler {
+        typealias InboundIn = NIOAny
+    }
+    class SimpleTypedHandler2: ChannelInboundHandler {
+        typealias InboundIn = NIOAny
+    }
+    class SimpleTypedHandler3: ChannelInboundHandler {
+        typealias InboundIn = NIOAny
+    }
+
+    func testGetHandler() throws {
+        let handler1 = SimpleTypedHandler1()
+        let handler2 = SimpleTypedHandler2()
+        let handler3 = SimpleTypedHandler3()
+        
+        let channel = EmbeddedChannel()
+        try channel.pipeline.addHandlers([
+            handler1,
+            handler2,
+            handler3
+        ]).wait()
+        
+        let result1 = try channel.pipeline.handler(type: SimpleTypedHandler1.self).wait()
+        XCTAssertTrue(result1 === handler1)
+        
+        let result2 = try channel.pipeline.handler(type: SimpleTypedHandler2.self).wait()
+        XCTAssertTrue(result2 === handler2)
+
+        let result3 = try channel.pipeline.handler(type: SimpleTypedHandler3.self).wait()
+        XCTAssertTrue(result3 === handler3)
+    }
+    
+    func testGetFirstHandler() throws {
+        let sameTypeHandler1 = SimpleTypedHandler1()
+        let sameTypeHandler2 = SimpleTypedHandler1()
+        let otherHandler = SimpleTypedHandler2()
+
+        let channel = EmbeddedChannel()
+        try channel.pipeline.addHandlers([
+            sameTypeHandler1,
+            sameTypeHandler2,
+            otherHandler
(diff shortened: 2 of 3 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the `ChannelHandlerContext` that belongs to a `ChannelHandler` of the given type.
+    ///
+    /// - parameters:
+    ///     - handlerType: the type of `ChannelHandler` to returns
+    public func handler<T: ChannelHandler>(type _: T.Type) -> EventLoopFuture<T> {
```
Lukasa: Should we have a version of this for each of our `context` methods? I've always thought the `context` methods were pretty weird, and ideally I'd get rid of them and replace them with these versions. @normanmaurer is there a good reason not to do that?

weissi: @Lukasa I think the `context` ones are there so you can search and later on remove without walking the pipeline. But I agree they're a bit weird, not sure if we should kill them though.

Tests/NIOTests/ChannelPipelineTest.swift
```diff
@@ -69,6 +69,38 @@ private extension EmbeddedChannel {
...
+        ]).wait()
+        
+        let result1 = try channel.pipeline.handler(type: SimpleTypedHandler1.self).wait()
+        XCTAssertTrue(result1 === handler1)
+        
+        let result2 = try channel.pipeline.handler(type: SimpleTypedHandler2.self).wait()
+        XCTAssertTrue(result2 === handler2)
+
+        let result3 = try channel.pipeline.handler(type: SimpleTypedHandler3.self).wait()
+        XCTAssertTrue(result3 === handler3)
```
weissi: nice! We could do some more tests here, like

- two handlers of the same type in the pipeline, verify that we get the first of those
- no handler of the given type there, verify that we get `nil`

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the `ChannelHandlerContext` that belongs to a `ChannelHandler` of the given type.
```
weissi: that documentation isn't quite right. It returns the first `ChannelHandler` of a given type.

Sources/NIO/ChannelPipeline.swift
weissi: I think we should name the generic parameter `Handler` instead of just `T`

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the first `ChannelHandler` that belongs to a ChannelHandler of the given type.
```
weissi:
```suggestion
    /// Returns the first `ChannelHandler` of the given type.
```

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+    ///
+    /// - parameters:
+    ///     - handlerType: the type of `ChannelHandler` to returns
+    public func handler<Handler: ChannelHandler>(type _: Handler.Type) -> EventLoopFuture<Handler> {
+        return context(handlerType: Handler.self).map { context in
+            guard let typedContext = context.handler as? Handler else {
+                precondition(false, "Expected channel context handler to be of type \(Handler.self), got \(type(of: context.handler)) instead.")
```
weissi:
```suggestion
                preconditionFailure("Expected channel handler of type \(Handler.self), got \(type(of: context.handler)) instead.")
```

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the first `ChannelHandler` that belongs to a ChannelHandler of the given type.
+    ///
+    /// - parameters:
+    ///     - handlerType: the type of `ChannelHandler` to returns
```
weissi:
```suggestion
    ///     - handlerType: the type of `ChannelHandler` to return.
```

minor typo here

weissi:

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1555,6 +1555,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return desc.joined(separator: "\n")
     }
     
+    /// Returns the first `ChannelHandler` of the given type.
+    ///
+    /// - parameters:
+    ///     - handlerType: the type of `ChannelHandler` to return.
+    public func handler<Handler: ChannelHandler>(type _: Handler.Type) -> EventLoopFuture<Handler> {
+        return context(handlerType: Handler.self).map { context in
```
Lukasa: Nit: can we use the explicit `self` keyword here?

Palleas: @Lukasa done!

Lukasa:

weissi:
@Mordil You'll like this PR I think ;)

PR #975, opened 2019-04-18
Add ability to open a file for writing

artemredkin:
Motivation:
Users can use NonBlockingFileIO for reading files, but if one wants to
write a file, there is no way to achieve that, apart from opening the file
manually.

Modifications:
* Rename parameter path in FileHandle convinience init and NonBlockingFileIO's openFile to
forReadingAtPath
* Add convinience init to FileHandle that can open file for writing
* Add method to NonBlockingFileIO to open file for writing
* rename argument path to forReadingAtPath in tests and examples

Result:
Users can now open files for writing and reading in the same way.

### DIFF
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -81,21 +81,50 @@ public final class NIOFileHandle: FileDescriptor {
 }
 
 extension NIOFileHandle {
-    /// Open a new `NIOFileHandle`.
-    ///
-    /// - parameters:
-    ///     - path: the path of the file to open for reading. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(path: String) throws {
-        let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
-        self.init(descriptor: fd)
+    /// `Mode` represents file access modes.
+    public enum Mode {
+        /// Opens file for reading only
+        case readOnly
+        /// Opens file for wrinting only
+        case writeOnly
+        /// Opens file for both reading and writing
+        case readWrite
+
+        internal var rawValue: Int32 {
+            switch self {
+            case .readOnly:
+                return O_RDONLY
+            case .writeOnly:
+                return O_WRONLY
+            case .readWrite:
+                return O_RDWR
+            }
+        }
+    }
+
+    public struct Flags {
+        internal var posixMode: UInt16
+        internal var posixFlags: Int32
+
+        public static let `default` = Flags(posixMode: 0, posixFlags: 0)
+
+        /// Allows file creation when opening file for writing. File owner is set to the effective user ID of the process.
+        ///
+        /// - parameters:
+        ///     - posixMode: `file mode` applied when file is created. Default permissions are: read and write for file owner, read for owners group and others.
+        public static func allowFileCreation(posixMode: UInt16 = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) -> Flags {
+            return Flags(posixMode: posixMode, posixFlags: O_CREAT)
+        }
     }
 
     /// Open a new `NIOFileHandle`.
     ///
     /// - parameters:
-    ///     - forWritingAtPath: the path of the file to open for writing. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called. If file does not exist, it will be created with default permissions 0644.
-    public convenience init(forWritingAtPath path: String, mode: UInt16 = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) throws {
-        let fd = try Posix.open(file: path, oFlag: O_WRONLY | O_CREAT | O_CLOEXEC, mode: mode)
+    ///     - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
+    ///     - mode: Access mode. Default mode is `.readOnly`.
+    ///     - flags: Additional POSIX flags.
+    public convenience init(path: String, mode: Mode = .readOnly, flags: Flags = .default) throws {
+        let fd = try Posix.open(file: path, oFlag: mode.rawValue | O_CLOEXEC | flags.posixFlags, mode: flags.posixMode)
         self.init(descriptor: fd)
     }
 }
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -91,7 +91,7 @@ extension NIOFileHandle {
 
         /// Opens file for reading
         public static let read = Mode(rawValue: O_RDONLY)
-        /// Opens file for wrinting
+        /// Opens file for writing
         public static let write = Mode(rawValue: O_WRONLY)
     }
 
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -82,24 +82,17 @@ public final class NIOFileHandle: FileDescriptor {
 
 extension NIOFileHandle {
     /// `Mode` represents file access modes.
-    public enum Mode {
-        /// Opens file for reading only
-        case readOnly
-        /// Opens file for wrinting only
-        case writeOnly
-        /// Opens file for both reading and writing
-        case readWrite
+    public struct Mode: OptionSet {
+        public let rawValue: Int32
 
-        internal var rawValue: Int32 {
-            switch self {
-            case .readOnly:
-                return O_RDONLY
-            case .writeOnly:
-                return O_WRONLY
-            case .readWrite:
-                return O_RDWR
-            }
+        public init(rawValue: Int32) {
+            self.rawValue = rawValue
         }
+
+        /// Opens file for reading
+        public static let read = Mode(rawValue: O_RDONLY)
+        /// Opens file for wrinting
+        public static let write = Mode(rawValue: O_WRONLY)
     }
 
     public struct Flags {
@@ -123,7 +116,7 @@ extension NIOFileHandle {
     ///     - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
     ///     - mode: Access mode. Default mode is `.readOnly`.
     ///     - flags: Additional POSIX flags.
-    public convenience init(path: String, mode: Mode = .readOnly, flags: Flags = .default) throws {
+    public convenience init(path: String, mode: Mode = .read, flags: Flags = .default) throws {
         let fd = try Posix.open(file: path, oFlag: mode.rawValue | O_CLOEXEC | flags.posixFlags, mode: flags.posixMode)
         self.init(descriptor: fd)
     }
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -84,9 +84,18 @@ extension NIOFileHandle {
     /// Open a new `NIOFileHandle`.
     ///
     /// - parameters:
-    ///     - path: the path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(path: String) throws {
+    ///     - forReadingAtPath: the path of the file to open for reading. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
+    public convenience init(forReadingAtPath path: String) throws {
         let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
+        self.init(descriptor: fd)
+    }
+
+    /// Open a new `NIOFileHandle`.
+    ///
+    /// - parameters:
+    ///     - forReadingAtPath: the path of the file to open for writing. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called. If file does not exist, it will be created with permissions 0644.
+    public convenience init(forWritingAtPath path: String) throws {
+        let fd = try Posix.open(file: path, oFlag: O_WRONLY | O_CREAT, mode: S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH)
         self.init(descriptor: fd)
     }
 }
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -89,7 +89,7 @@ extension NIOFileHandle {
             self.rawValue = rawValue
         }
 
-        internal var posixMode: CInt {
+        internal var posixFlags: CInt {
             switch self {
             case [.read, .write]:
                 return O_RDWR
@@ -108,6 +108,7 @@ extension NIOFileHandle {
         public static let write = Mode(rawValue: 1 << 1)
     }
 
+    /// `Flags` allows to specify additional flags to `Mode`, such as permission for file creation.
     public struct Flags {
         internal var posixMode: mode_t
         internal var posixFlags: CInt
@@ -130,7 +131,7 @@ extension NIOFileHandle {
     ///     - mode: Access mode. Default mode is `.readOnly`.
     ///     - flags: Additional POSIX flags.
     public convenience init(path: String, mode: Mode = .read, flags: Flags = .default) throws {
-        let fd = try Posix.open(file: path, oFlag: mode.posixMode | O_CLOEXEC | flags.posixFlags, mode: flags.posixMode)
+        let fd = try Posix.open(file: path, oFlag: mode.posixFlags | O_CLOEXEC | flags.posixFlags, mode: flags.posixMode)
         self.init(descriptor: fd)
     }
 }
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -109,7 +109,7 @@ extension NIOFileHandle {
     }
 
     public struct Flags {
-        internal var posixMode: UInt16
+        internal var posixMode: mode_t
         internal var posixFlags: CInt
 
         public static let `default` = Flags(posixMode: 0, posixFlags: 0)
@@ -118,7 +118,7 @@ extension NIOFileHandle {
         ///
         /// - parameters:
         ///     - posixMode: `file mode` applied when file is created. Default permissions are: read and write for file owner, read for owners group and others.
-        public static func allowFileCreation(posixMode: UInt16 = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) -> Flags {
+        public static func allowFileCreation(posixMode: mode_t = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) -> Flags {
             return Flags(posixMode: posixMode, posixFlags: O_CREAT)
         }
     }
--- a/Sources/NIO/FileHandle.swift
+++ b/Sources/NIO/FileHandle.swift
@@ -84,8 +84,8 @@ extension NIOFileHandle {
     /// Open a new `NIOFileHandle`.
     ///
     /// - parameters:
-    ///     - forReadingAtPath: the path of the file to open for reading. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(forReadingAtPath path: String) throws {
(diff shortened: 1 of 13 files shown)


Sources/NIO/FileHandle.swift
```diff
@@ -84,11 +84,20 @@ extension NIOFileHandle {
     /// Open a new `NIOFileHandle`.
     ///
     /// - parameters:
-    ///     - path: the path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(path: String) throws {
+    ///     - forReadingAtPath: the path of the file to open for reading. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
+    public convenience init(forReadingAtPath path: String) throws {
```
artemredkin: this is to mirror `Foundation's` `FileHandle`, do you think its a good idea?

Lukasa: Well it makes this a semver major change, so at the very least you’ll have to return the pre-existing one. 

I don’t mind this change, but we need the old one. We can deprecate it though.

Sources/NIO/FileHandle.swift
```diff
@@ -84,11 +84,20 @@ extension NIOFileHandle {
...
         let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
         self.init(descriptor: fd)
     }
+
+    /// Open a new `NIOFileHandle`.
+    ///
+    /// - parameters:
+    ///     - forReadingAtPath: the path of the file to open for writing. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called. If file does not exist, it will be created with permissions 0644.
```
Lukasa: This doc string is not right for this function.

Sources/NIO/FileHandle.swift
```diff
@@ -84,11 +84,20 @@ extension NIOFileHandle {
...
+    public convenience init(forWritingAtPath path: String) throws {
+        let fd = try Posix.open(file: path, oFlag: O_WRONLY | O_CREAT, mode: S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH)
```
Lukasa: Where are all these extra mode flags coming from? Why is `O_CLOEXEC` missing? I think the differences between this and the other initialiser need to be explained or removed.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -253,20 +253,45 @@ public struct NonBlockingFileIO {
...
     ///
     /// - note: The reason this returns the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as the querying of its size are blocking.
     ///
     /// - parameters:
-    ///     - path: The path of the file to be opened.
+    ///     - forReadingAtPath: The path of the file to be opened for reading.
     ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
     /// - returns: An `EventLoopFuture` containing the `NIOFileHandle` and the `FileRegion` comprising the whole file.
-    public func openFile(path: String, eventLoop: EventLoop) -> EventLoopFuture<(NIOFileHandle, FileRegion)> {
+    public func openFile(forReadingAtPath path: String, eventLoop: EventLoop) -> EventLoopFuture<(NIOFileHandle, FileRegion)> {
```
Lukasa: Again, this can only be changed if we either make this a semver major or return the old version and deprecate it.

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -253,20 +253,45 @@ public struct NonBlockingFileIO {
...
+    /// comprising of the whole file. The caller must close the returned `NIOFileHandle` when it's no longer needed.
+    /// If file does not exist, it will be created with permissions 0644
+    ///
+    /// - note: The reason this returns the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as the querying of its size are blocking.
+    ///
+    /// - parameters:
+    ///     - forWritingAtPath: The path of the file to be opened for writing.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing the `NIOFileHandle` and the `FileRegion` comprising the whole file.
+    public func openFile(forWritingAtPath path: String, eventLoop: EventLoop) -> EventLoopFuture<(NIOFileHandle, FileRegion)> {
```
Lukasa: Is there any point returning a`FileRegion` here?

Sources/NIO/FileHandle.swift
artemredkin: What do you mean? We need to specify mode for `O_CREAT`, otherwise it creates a file with 0000 permissions

Sources/NIO/FileHandle.swift
artemredkin: Reverted reading path

Sources/NIO/FileHandle.swift
artemredkin: fixed!

Sources/NIO/FileHandle.swift
artemredkin: added O_CLOEXEC

Sources/NIO/NonBlockingFileIO.swift
artemredkin: Reverted reading argument name

Sources/NIO/NonBlockingFileIO.swift
artemredkin: Ah, good point, fixed

Sources/NIO/FileHandle.swift
Lukasa: Hrm. Can we add those mode flags as an argument to this method and give it that default value?

artemredkin: Done

Sources/NIO/FileHandle.swift
```diff
@@ -84,11 +84,20 @@ extension NIOFileHandle {
...
     public convenience init(path: String) throws {
         let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
         self.init(descriptor: fd)
     }
+
+    /// Open a new `NIOFileHandle`.
+    ///
+    /// - parameters:
+    ///     - forWritingAtPath: the path of the file to open for writing. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called. If file does not exist, it will be created with default permissions 0644.
+    public convenience init(forWritingAtPath path: String, mode: UInt16 = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH) throws {
```
weissi: could we have a nicer mode argument here? The POSIX API isn't that nice I think. Also the `forWritingAtPath` looks a bit too much like old day cocoa ;).

Also that kind of mode only matters when _creating_ a new file.

What about an API like this:

```swift

    enum Mode {
        case .readOnly
        case .writeOnly
        case .readWrite
    } // or as OptionSet?
    public struct Flags {
        private posixMode: UInt16

        public static default: Flags = .init(posixMode: 0)

        public func allowFileCreation(posixMode: UInt16) {
            return .init(posixMode: posixFlags)
        }
    }
    public init(path: String, mode: Mode = .read, flags: Flags = .default) { ...}
```

that way first of all we don't need a _new_ API, but we can extend the existing API. Also we don't make the weird POSIX mode the only game in town and also the POSIX mode is only required when it's actually used (on file creation, if that's allowed).

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -277,4 +277,22 @@ public struct NonBlockingFileIO {
...
+    /// comprising of the whole file. The caller must close the returned `NIOFileHandle` when it's no longer needed.
+    /// If file does not exist, it will be created with permissions 0644
+    ///
+    /// - note: The reason this returns the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as the querying of its size are blocking.
+    ///
+    /// - parameters:
+    ///     - forWritingAtPath: The path of the file to be opened for writing.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing the `NIOFileHandle` and the `FileRegion` comprising the whole file.
+    public func openFile(forWritingAtPath path: String, eventLoop: EventLoop) -> EventLoopFuture<NIOFileHandle> {
```
weissi: same comments here.

Sources/NIO/FileHandle.swift
artemredkin: good idea, one question: do you propose to expose POSIX flags as something more readable (`.readOnly == O_RDONLY`) or introduce our own modes that map to sensible set of POSIX flags (`.reading = O_RDONLY | O_CLOEXEC`)?

Sources/NIO/NonBlockingFileIO.swift
artemredkin: What about returning `FileRegion` when opening for writing? If we have single API with an enum/option set, then we have to return `FileRegion`

Sources/NIO/FileHandle.swift
artemredkin: `forWritingAtPath` looks exactly like old day cocoa :D

weissi: @artemredkin I'd expose some options as more readable things. Like `.readOnly` should be `O_RDONLY` but `O_CLOEXEC` we should always set, there's no point in not setting `O_CLOEXEC`.

Sources/NIO/NonBlockingFileIO.swift
weissi: @artemredkin you're right, `FileRegion` probably doesn't make sense for writable stuff.

weissi: then maybe we should have an API that supports read, write, read+write and does _not_ return a `FileRegion`

tomerd:

Sources/NIO/FileHandle.swift
artemredkin: Does this look better?

Sources/NIO/NonBlockingFileIO.swift
artemredkin: Added another `openFile` method that requires `mode` and does not return a `FileRegion`

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,50 @@ public final class NIOFileHandle: FileDescriptor {
 }
 
 extension NIOFileHandle {
+    /// `Mode` represents file access modes.
+    public enum Mode {
```
weissi: just a note, this could be an `OptionSet`?

artemredkin: It could, but I don't think it makes much sense, what would combining `.readOnly` and `.readWrite` mean for example? And combining `.readOnly` and '.writeOnly` gives you `.readWrite`, if I'm not mistaken.

weissi: well but you could just have `.read` and `.write`. Then you would have

- `[.read, .write]` --> `O_RDWR`
- `[.read]` --> `O_RDONLY`
- `[.write]` --> `O_WRONLY`

artemredkin: What could work. Converted to `OptionSet`

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,43 @@ public final class NIOFileHandle: FileDescriptor {
...
+
+        public init(rawValue: Int32) {
+            self.rawValue = rawValue
+        }
+
+        /// Opens file for reading
+        public static let read = Mode(rawValue: O_RDONLY)
+        /// Opens file for writing
+        public static let write = Mode(rawValue: O_WRONLY)
+    }
```
weissi:
```suggestion
    /// `Mode` represents file access modes.
	    public struct Mode: OptionSet {
	        public let rawValue: UInt8
	
	        public init(rawValue: Int32) {
	            self.rawValue = rawValue
	        }
	
	        /// Opens file for reading
	        public static let read = Mode(rawValue: 1 << 0)
	        /// Opens file for writing
	        public static let write = Mode(rawValue: 1 << 1)
	    }
	    
	    internval var posixMode: CInt {
	        if self == [.read, .write] {
	            return O_RDWR
	       } else {
	           if self.contains(.read) {
	               return O_RDONLY
	           }
	           if self.contains(.write) {
	               return O_WRONLY
	           }
	           return 0
	      }
	  }
```

unfortunately, I don't think we can use the underlying POSIX values here. I don't think POSIX guarantees that `O_RDONLY | O_WRONLY == O_RDWR`. I think you will need to change this to something like the above which creates an OptionSet and then maps it to POSIX.

artemredkin: Maybe then we don't need `OptionSet` at all? Especially given that users can add arbitrary values to it. What do we do in that case? Ignore user-provided values? Fail/crash if user specified something unexpected?

weissi: @artemredkin but the `OptionSet` gives us the nice `[.read, .write]` syntax for free, right?

artemredkin: Yes, it will, but how should we treat something like this: ```let mode: Mode = [.read, .init(rawValue: 127), .init(rawValue: UInt8(O_CREAT))]``` for example?

weissi: well, they're illegal. We can ignore or crash. Sure, it's not really awesome but that's how `OptionSet`s work...

weissi: if you have it too much you could also implement it from `ArrayLiteralConvertible`.

artemredkin: Ignore or crash are runtime errors, wouldn't it be better to have compile-time safety here by using enums?

weissi: @artemredkin Sure, it's just how `OptionSet`s work. I'm totally fine if you build a thing that doesn't use `OptionSet`s but I think it should probably be extensible. With an `enum` we can just about support `read` and `write`, by adding a third case `readAndWrite`. But adding anything else will be painful because 1) it breaks public API and 2) the number of cases will explode. For example if we need to support `O_APPEND` then it would be much cooler do support `[.read, .write, .append]` rather than having

```
enum Mode {
    case read
    case write
    case writeAndAppend
    case readAndWrite
    case readAndWriteAndAppend
}
```

just because there are so many things we need to support I'd like to have it extensible. The standard solution in Swift is `OptionSet` but as I said above, totally happy for you to not use `OptionSet` and make `[.read, .write]` work using `ArrayLiteralConvertible`.

artemredkin: `open`'s manual states:
```
The argument flags must include one of the following access modes:
       O_RDONLY, O_WRONLY, or O_RDWR.
```
so those three are required, they are base modes, everything else is an option. So potential `O_APPEND` would be better set by `Flags`, in exact same manner as `O_CREAT`:
```swift
public struct Flags {
    public static let append = Flags(posixMode: O_APPEND, posixFlags: 0)
    public static let async = Flags(posixMode: O_ASYNC, posixFlags: 0)
    public static let tmp = Flags(posixMode: O_TMPFILE,  posixFlags: 0)
}

let file = openFile(path: "/some/file", mode: .writeOnly, flags: [.append, .async, .allowFileCreation()])
```
That would make it consistent with how we set `O_CREAT`. What do you think?

weissi: @artemredkin well, but then there's `O_EVTONLY` which is in the same category as `O_RDONLY`, `O_WRONLY`, and `O_RDWR`

artemredkin: Makes sense. Not sure `ArrayLiteralConvertible` looks nice, left as an `OptionSet` and added `precondition` to prevent passing in unsupported values

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,56 @@ public final class NIOFileHandle: FileDescriptor {
...
+            }
+        }
+
+        /// Opens file for reading
+        public static let read = Mode(rawValue: 1 << 0)
+        /// Opens file for writing
+        public static let write = Mode(rawValue: 1 << 1)
+    }
+
+    public struct Flags {
```
Lukasa: Can we get a doc comment here too?

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,56 @@ public final class NIOFileHandle: FileDescriptor {
 }
 
 extension NIOFileHandle {
+    /// `Mode` represents file access modes.
+    public struct Mode: OptionSet {
+        public let rawValue: UInt8
```
Lukasa: This should probably be `private var` instead of `public let`.

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,56 @@ public final class NIOFileHandle: FileDescriptor {
...
+
+        public init(rawValue: UInt8) {
```
Lukasa: Is there any reason for this initialiser to be public?

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,56 @@ public final class NIOFileHandle: FileDescriptor {
...
     ///
     /// - parameters:
-    ///     - path: the path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(path: String) throws {
-        let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
+    ///     - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
+    ///     - mode: Access mode. Default mode is `.readOnly`.
+    ///     - flags: Additional POSIX flags.
+    public convenience init(path: String, mode: Mode = .read, flags: Flags = .default) throws {
+        let fd = try Posix.open(file: path, oFlag: mode.posixMode | O_CLOEXEC | flags.posixFlags, mode: flags.posixMode)
```
Lukasa: This looks a bit weird, which makes me realise that `posixMode` is overloaded: there is `Mode.posixMode` and `Flags.posixMode`, but they don't go to the same place. Can we re-evaluate those names? Maybe `Mode.posixMode` should actually be `mode.posixFlags` in order to match `Flags`?

Sources/NIO/FileHandle.swift
artemredkin: This is a requirement from `RawRepresentable`, we cannot make it private

Sources/NIO/FileHandle.swift
artemredkin: Yes, it is a requirement of `OptionSet`

Sources/NIO/FileHandle.swift
artemredkin: Done

Sources/NIO/FileHandle.swift
artemredkin: Fixed!

Sources/NIO/FileHandle.swift
Lukasa: Wow I always forget how annoying a protocol `OptionSet` is.

Sources/NIO/FileHandle.swift
Lukasa: Wow I always forget how annoying a protocol `OptionSet` is.

Sources/NIO/FileHandle.swift
```diff
@@ -81,12 +81,57 @@ public final class NIOFileHandle: FileDescriptor {
...
+    }
+
     /// Open a new `NIOFileHandle`.
     ///
     /// - parameters:
-    ///     - path: the path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
-    public convenience init(path: String) throws {
-        let fd = try Posix.open(file: path, oFlag: O_RDONLY | O_CLOEXEC)
+    ///     - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
+    ///     - mode: Access mode. Default mode is `.readOnly`.
```
Lukasa: No it's not, it's `.read` ;)

artemredkin: good catch :) fixed, thanks!

PR #980, opened 2019-04-25
Bump all copyrights to 2019

Palleas:
Motivation:

Generated tests use 2018 at the upper bound date in the copyright
header, which is not accurate for new files.

Modifications:

Bump the date to 2019.

Result:

Properly copyrighted files.

After a quick check with @normanmaurer, we may not want to update the copyright on files that didn't change, but this was the quickest way otherwise I need to tweak the sanity script as well. Please let me know which approach you prefer!

### DIFF
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2019 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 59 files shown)


weissi:
I _think_ that's not what legal is recommending. CC @tomerd / @normanmaurer here. I think you're only supposed to adjust the copyright if you make changes??

ktoso:
I believe this is correct from what I’ve seen: the year should only be updated if a file is touched within a given year was the official guidance...

Palleas:
Fair enough, I can do that. That means changing the script that generates the Linux tests as well though, cause we don’t want to generate all the Linux tests anymore, right?

weissi:
> Fair enough, I can do that. That means changing the script that generates the Linux tests as well though, cause we don’t want to generate all the Linux tests anymore, right?

I think the test gen script can stay maybe? You could argue that the code hasn't changed if the generator didn't change?

Palleas:
@weissi you mean, just closing this PR? 😬

weissi:
> @weissi you mean, just closing this PR? 😬

Well, I meant I think we can leave the generated files alone. If we did miss copyright updates and you would like to update those, please feel free to do so in this PR :)

Palleas:
The problem I had was in a previous PR where I generated a new test file and manually changed the year to 2019. I guess if we argue that it’s coming from the generator script and we don’t need to update it, I’m ok closing it, no worries!

weissi:
@Palleas ah sorry, I see now. The PR _only_ touched generated files.

Issue #665, opened 2018-11-27
futures should have an `always` or something


weissi:
often you need to do some cleanup that needs to always happen, why not

```swift
extension EventLoopFuture {
    func always(_ body: () -> Void) -> EventLoopFuture<T>
}
```

that always runs?

Lukasa:
Why does `always` return a `T`?

weissi:
because it's too early in the morning 😂. fixing it :P

normanmaurer:
@weissi so this is something that will be executed no matter if the future is completed because of a failure or success ?

Lukasa:
@weissi Is this any different to:

```swift
public func whenComplete(_ callback: @escaping () -> Void)
```

Because we already have that.

Lukasa:
Oh, I see, it chains. Sure, why not.

normanmaurer:
If we do this then we should most likely just mark `whenComplete` as deprecated and introduce `always` but mark its return value as `@discardableResult`.

weissi:
@normanmaurer need both chaining and non-chaining variants

normanmaurer:
@weissi why ?

weissi:
@normanmaurer because otherwise you can't write this:

```swift
func runDatabaseQuery(_ query: String, database: DB) -> EventLoopFuture<[Row]> {
    return database.openConnetion { conn in
        conn.runQuery(query).always {
            conn.close()
        }
    }
}
```

You want to always close the DB connection. Currently you need to spell this

```swift
func runDatabaseQuery(_ query: String, database: DB) -> EventLoopFuture<[Row]> {
    return database.openConnetion { conn in
        let f = conn.runQuery(query)
        f.whenComplete {
            conn.close()
        }
        return f
    }
}
```

which is super ugly

normanmaurer:
@weissi I may need more coffee but I still dont get why we could not only have `always` at the end ?

weissi:
@Lukasa we'll still create an extra, unnecessary future. We also have `then` & `whenSuccess`, `thenIfError` & `whenFailure`, but _nothing_ & `whenSuccess`. I just want to fill the _nothing_-hole :)

normanmaurer:
@weissi why ? Couldn't it just return `self` in this case ?

weissi:
@normanmaurer true

Lukasa:
Bear in mind that the semantics of that are *slightly* different than the semantics of returning a new future, though only slightly.

[...]

PR #981, opened 2019-04-27
Add an `always` method on EventLoopFuture

Palleas:
Motivation:

https://github.com/apple/swift-nio/issues/665

Modifications:

* Add `always` in EventLoopFuture.swift
* Add tests

Result:

NIO Users will be able to use `.always` to, well, always run an action
whether the future succeed or not.


Sources/NIO/EventLoopFuture.swift
```diff
@@ -1267,3 +1267,19 @@ func executeAndComplete<Value>(_ promise: EventLoopPromise<Value>?, _ body: () t
         promise?.fail(e)
     }
 }
+
+
+extension EventLoopFuture {
+
```
Lukasa: Minor nit: can we remove this whitespace?

Palleas: @Lukasa Done!

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1267,3 +1267,18 @@ func executeAndComplete<Value>(_ promise: EventLoopPromise<Value>?, _ body: () t
...
+
+extension EventLoopFuture {
+    /// Adds an observer callback to this `EventLoopFuture` that is called when the
+    /// `EventLoopFuture` has any result.
+    ///
+    /// - parameters:
+    ///     - callback: the callback that is called when the `EventLoopFuture` is fulfilled.   
+    /// - returns: the current `EventLoopFuture`
+    public func always(_ callback: @escaping () -> Void) -> EventLoopFuture<Value> {
```
weissi: @Lukasa do you think we should give users the result here? For `whenComplete` we hand them the result so why not here?

glbrntt: FWIW I just came across a case where having the above with the result (or `alwaysSuccess`/`alwaysError`) would have been helpful.

Lukasa: I don't think I mind particularly. Arguably there are a family of functions here (giving them the name `always` for now though the compiler wouldn't like it much) created by twiddling both the return value of the function and the return value of the closure it accepts:

1. `Future<T>.always(_: () -> Void) -> Future<T>`
2. `Future<T>.always(_: () -> Void) -> Void`
3. `Future<T>.always(_: (Result<T, Error>) -> Void) -> Future<T>`
4. `Future<T>.always(_: (Result<T, Error>) -> Void) -> Void`
5. `Future<T>.always<U>(_: (Result<T, Error>) -> U) -> Future<U>`
6. `Future<T>.always<U>(_: (Result<T, Error>) -> Future<U>) -> Future<U>`
7. `Future<T>.always(_: (Result<T, Error>) -> Future<Void>) -> Future<T>`

This proposal covers (1).  (4) is currently spelled `whenComplete`. (3) is what you're proposing instead, and can do everything (1) can do at the cost of being a bit noisier in the code when you want (1). (2) is what `whenComplete` *used* to be in NIO 1, before we had a `Result` type to use.

(5) is arguably a function that would be called `mapBoth`, and (6) would be `flatMapBoth` by analogy. (7) is akin to (4) but allows returning a `Future` while persisting the result of the computation.

I think there could be a case made for providing *all* of these functions: the biggest challenge would be providing descriptive names, as we cannot rely on the compiler to provide overload resolution. (7) is the least obviously useful.

I think we should stick with `always` being the name for this function. We should also provide (3) and call it something like `alwaysWithResult`. Otherwise we'll force a lot of people to write `always { _ in }` and that kinda sucks.

weissi: @Lukasa Mostly agree, just that for `whenComplete` we decided to only offer the 'with result' case (but we didn't name it `whenCompleteWithResult`). Therefore, I'm not sure if for `always` we should do `always` and `alwaysWithResult`... But I agree that the `_ in` kinda sucks and obviously that the compiler won't like overloading at all. I'd probably still go for consistency with `whenComplete`.

Lukasa: No strong opinion here, I'm happy to follow your preference.

glbrntt: @weissi @Lukasa If there's a version providing the result then I'd think `alwaysWithSuccess` and `alwaysWithError` would probably be requested too? Providing a value like this is pretty close to Java's `peek`.

If that's the case then maybe these are distinct from the `always` as it stands at the moment?

```
Future<T>.always(_: () -> Void) -> Future<T>
Future<T>.peek(_: (T) -> Void) -> Future<T>
Future<T>.peekError(_: (Error) -> Void) -> Future<T>
Future<T>.peekResult(_: (Result<T, Error>) -> Void) -> Future<T>
```

weissi: Heh, I kind of like the `peek`s. My only concern with leaving `always` as is would be  `always` _not_ giving you a value but `whenComplete` giving you the result. Just looks a bit inconsistent, that's all :). Especially that in NIO 1 we had `whenComplete` giving you nothing and we switched to `whenComplete` giving you the result for NIO2 :).

glbrntt: True, it's probably best to be consistent with `whenComplete` even though I prefer the NIO 1 version of it! I think I ignore the result more often than not...

I would vote to change it back in some future version of NIO and introduce `whenResult` to be closer in name to the other methods which provide a `Result` to the callback. The above would then be analogues of the `when{Complete,Success,Failure,Result}` which return the future.

weissi: Sounds good. `when{Complete,Success,Failure,Result}` make sense to me.

Palleas:
@Lukasa fixed the failing build, sorry about that. I'm slowly engraving in my brain the reflex of running the sanity check locally.

Sources/NIO/EventLoopFuture.swift
weissi: @Palleas just talked to @Lukasa about this. Let's add the `Result<...>` to the closure for `always` for consistency with `whenComplete`. The cleanup and providing the other missing methods can be done as #986 .

Palleas:
@weissi your wish is my command 🙇

PR #983, opened 2019-04-29
Update naming to allow for the easy addition of a HTTP client upgrader.

tigerpixel:
Update server upgrader naming to allow for the easy addition of an HTTP client upgrader.

Motivation:
To prepare the class naming for the presence of a client upgrader by improving server upgrader naming.

Modifications:
Renamed upgrader and test files to include the word ‘server’.
Ran the Linux test script.

Result:
Clearer naming in relation to server upgrader classes.


tigerpixel:
This has an additional commit that I didn't mean to send. I will sort asap, but can't do it right now.

Sources/NIOHTTP1/HTTPServerUpgradeHandler.swift
```diff
@@ -186,12 +186,12 @@ public final class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableCha
...
         }
 
-        guard let upgrader = upgraders[proto.lowercased()] else {
+        guard let upgrader = self.upgraders[proto.lowercased()] else {
             return self.handleUpgradeForProtocol(context: context, protocolIterator: protocolIterator, request: request, allHeaderNames: allHeaderNames, connectionHeader: connectionHeader)
         }
 
         let requiredHeaders = Set(upgrader.requiredUpgradeHeaders.map { $0.lowercased() })
-        guard requiredHeaders.isSubset(of: allHeaderNames) && requiredHeaders.isSubset(of: connectionHeader) else {
+        guard requiredHeaders.isSubset(of: allHeaderNames) && connectionHeader.isSubset(of: allHeaderNames) else {
```
Lukasa: I don't think the original check was incorrect. An `Upgrade` header is by definition a hop-by-hop header field, and so must be mentioned in the `Connection` header field value. RFC 7230 § 6.7 calls this out explicitly:

> When Upgrade is sent, the sender MUST also send a Connection header field (Section 6.1) that contains an "upgrade" connection option, in order to prevent Upgrade from being accidentally forwarded by intermediaries that might not implement the listed protocols

Somewhat by definition, then, any header required for upgrade must *also* be a hop-by-hop header, otherwise it will be sent unnecessarily when passing through intermediaries that do not support the upgrade. For this reason, we validate that.

In the case of protocols that have implementations that incorrectly fail to do this (most notably WebSocket), this can be handled by implementing an empty required header array and then checking in the upgrade callback.

Sources/NIOWebSocket/WebSocketServerUpgrader.swift
```diff
@@ -51,14 +51,14 @@ fileprivate extension HTTPHeaders {
...
-/// reason, this `WebSocketUpgrader` only knows how to do the required parts of the upgrade and to
+/// reason, this `WebServerSocketUpgrader` only knows how to do the required parts of the upgrade and to
 /// complete the handshake. Users are expected to provide a callback that examines the HTTP headers
 /// (including the path) and determines whether this is a websocket upgrade request that is acceptable
 /// to them.
 ///
 /// This upgrader assumes that the `HTTPServerUpgradeHandler` will appropriately mutate the pipeline to
 /// remove the HTTP `ChannelHandler`s.
-public final class WebSocketUpgrader: HTTPServerProtocolUpgrader {
+public final class WebSocketServerUpgrader: HTTPServerProtocolUpgrader {
```
Lukasa: Changing the name of this class is a semver major change, which means it'll have to wait until SwiftNIO 3.0. That won't come any time soon!

If you want to land this patch sooner, consider adding a deprecated typealias that gives back the old name. That will allow existing code to continue to function. Per our API evolution guidelines, you'll also have to rename this to `NIOWebSocketServerUpgrader`.

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
```diff
@@ -18,7 +18,7 @@ import Dispatch
 @testable import NIOHTTP1
 
 extension ChannelPipeline {
-    func assertDoesNotContainUpgrader() throws {
+    fileprivate func assertDoesNotContainUpgrader() throws {
```
Lukasa: Why did these become file private?

Sources/NIOHTTP1/HTTPServerUpgradeHandler.swift
tigerpixel: I think I slightly misunderstood the spec. Thanks for the description. You are right, the original code is correct. I will set this back.

Sources/NIOWebSocket/WebSocketServerUpgrader.swift
tigerpixel: Okay, will do. Thanks.

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
tigerpixel: They are tied to checking for 'HTTPServerUpgradeHandler'. I am currently writing a client handler, so they become a bit misleading. I could have renamed them, but I thought it was just easier to make them fileprivate as that is where they are used. 

To be honest, it isn't a big deal, I was more concerned with the other changes. I can see that the general pattern seems to be to put extensions for testing in the most relevant file but allow usage from any other test file. If you would rather not affect that, I can make separate client versions that explicitly have the word 'client', for example, and leave these alone.

Lukasa: No, your reasoning makes sense, I just wanted to understand the change. For now making these `fileprivate` shouldn't hurt, we can always deal with the issue later if we need them elsewhere.

Sources/_NIO1APIShims/NIO1APIShims.swift
```diff
@@ -522,6 +522,9 @@ public typealias HTTPUpgradeEvents = HTTPServerUpgradeEvents
 @available(*, deprecated, renamed: "HTTPServerUpgradeErrors")
 public typealias HTTPUpgradeErrors = HTTPServerUpgradeErrors
 
+@available(*, deprecated, renamed: "NIOWebSocketServerUpgrader")
+public typealias WebSocketUpgrader = NIOWebSocketServerUpgrader
```
Lukasa: This can't go in the `_NIO1APIShims` module as it's going away. This typealias will need to go into the `NIOWebSocketServerUpgrader` file.

tigerpixel: Okay, thanks. Have moved it now.

tigerpixel:
I have seen the conflict and will re-run the Linux file this evening. (Or resolve the conflict in another way.)

Lukasa:

tigerpixel:
Thanks.

PR #990, opened 2019-05-02
add tests for opening files RW

weissi:
Motivation:

The new file RW opening facilities didn't have any tests and it didn't
let you specify arbitrary POSIX flags.

Modifications:

- Add tests.
- Let users specify both POSIX flags & mode.

Result:

Better code quality.


Lukasa:
LGTM.

PR #991, opened 2019-05-02
add Swift 5.1 docker file

weissi:
Motivation:

We should test all available Swift versions, including snapshots.

Modifications:

Add docker compose file for Swift 5.1.

Result:

More testing.

### DIFF
--- a/docker/docker-compose.1604.51.yaml
+++ b/docker/docker-compose.1604.51.yaml
@@ -1,0 +1,42 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio:16.04-5.1
+    build:
+      args:
+        ubuntu_version: "16.04"
+        swift_version: "5.1"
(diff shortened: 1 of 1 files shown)


weissi:
> Why is this using Ubuntu 16.04?

Just to cover more of the OS / Swift version matrix. 5.0 tests on 18.04, so I made 5.1 test on 16.04.

docker/docker-compose.1604.51.yaml
```diff
@@ -0,0 +1,42 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio:16.04-5.1
+    build:
+      args:
+        ubuntu_version: "16.04"
```
tomerd: why not 18.04?

Lukasa: Heh, [great minds](https://github.com/apple/swift-nio/pull/991#issuecomment-488660476)!

PR #993, opened 2019-05-03
Improve MarkedCircularBuffer performance.

Lukasa:
Motivation:

While CircularBuffer has an inlinable implementation to improve performance, MarkedCircularBuffer
never got enhanced to have it. This harms performance in many core NIO libraries which heavily use
MCB, forcing heap allocations on almost all MCB operations.

Modifications:

- Make basically everything on MarkedCircularBuffer inlinable.

Result:

Faster code.

### DIFF
--- a/Sources/NIO/MarkedCircularBuffer.swift
+++ b/Sources/NIO/MarkedCircularBuffer.swift
@@ -18,77 +18,87 @@
 /// writes and mark how far through the buffer the user has flushed, and therefore how far through the buffer is
 /// safe to write.
 public struct MarkedCircularBuffer<Element>: CustomStringConvertible {
-    private var buffer: CircularBuffer<Element>
-    private var markedIndexOffset: Int? = nil /* nil: nothing marked */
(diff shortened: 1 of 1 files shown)


Sources/NIO/MarkedCircularBuffer.swift
```diff
@@ -18,29 +18,33 @@
 /// writes and mark how far through the buffer the user has flushed, and therefore how far through the buffer is
 /// safe to write.
 public struct MarkedCircularBuffer<Element>: CustomStringConvertible {
-    private var buffer: CircularBuffer<Element>
-    private var markedIndexOffset: Int? = nil /* nil: nothing marked */
+    @usableFromInline internal var buffer: CircularBuffer<Element>
+    @usableFromInline internal var markedIndexOffset: Int? = nil /* nil: nothing marked */
```
weissi: Mind prefixing with `_`? We and the stdlib use that convention for ‘should be private’

Lukasa: Addressed.

PR #996, opened 2019-05-03
new alloc counter tests

weissi:
Motivation:

Previously, writing an allocation test was quite hard, the allocation tests
were hard to debug and also the allocation testing framework and the
allocations tests were mangled together. That was bad and therefore
we never started writing allocation tests for other NIO repositories such
as `swift-nio-http2`.

Modifications:

- Make it easy to write alloc tests (just drop a `test_...swift` file in a folder).
- separate the alloc counter framework from the tests

Result:

- easier to write allocation tests
- 'productise' the alloc counter framework


Lukasa:
Looks like we have some test failures here. :grin:

weissi:
@Lukasa ha, it expects package id to be the same as package basename :P:

```
18:57:02 Completed resolution in 0.62s
18:57:02 error: package at '/code' is swift-nio but was expecting code
```

but we check the code out into `/code`... Will fix

PR #997, opened 2019-05-06
HTTPDecoder: don't crash if response has no headers

weissi:
Motivation:

With HTTP/1.0 and EOF framing it's possible to have a HTTP response that
doesn't have any headers. HTTPDecoder unfortunately fell flat on its
face and crashed the program if that happened.

Modifications:

- don't expect headers on responses
- add tests for that
- rename the state machine cases to reflect a request and response
  parsing

Result:

fewer crashes.

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -23,8 +23,8 @@ private extension UnsafeMutablePointer where Pointee == http_parser {
 }
 
 private enum HTTPDecodingState {
-    case beforeRequest
-    case beforeURL
+    case beforeMessageBegin
+    case afterMessageBegin
     case url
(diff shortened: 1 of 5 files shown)


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -256,7 +256,10 @@ private class BetterHTTPParser {
             self.finish { delegate, bytes in
                 delegate.didReceiveURL(bytes)
             }
-        case .beforeRequest, .headersComplete, .headerName, .trailerName, .trailerValue, .beforeURL:
+        case .afterMessageBegin:
```
weissi: @Lukasa note, this is the real change: we're happy with `afterMessageBegin` (previously `.beforeURL`) iff we're parsing a response

PR #1004, opened 2019-05-16
Allow adding a sequence of headers to HTTPHeaders

glbrntt:
Motivation:

When combining two collections of headers, adding each pair one-by-one
using add(name:value:) can cause multiple unnecessary reallocations of
the underlying array.

Modifications:

Provide two additional add methods to HTTPHeaders: one to add a sequence
of name/value pairs and another to add the headers from another
HTTPHeaders, both of which use `append(contentsOf:)` on the underlying
array to avoid additional reallocations.

Result:

Fewer reallocations when adding multiple headers.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -278,6 +278,7 @@ extension HTTPHeaders {
 /// or split representation, such that header fields that are able to be repeated
 /// can be represented appropriately.
 public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiteral {
+    @usableFromInline
     internal var headers: [(String, String)]
     internal var keepAliveState: KeepAliveState = .unknown
 
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -320,8 +320,7 @@ public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiter
     /// Add a header name/value pair to the block.
     ///
     /// This method is strictly additive: if there are other values for the given header name
-    /// already in the block, this will add a new entry. `add` performs case-insensitive
-    /// comparisons on the header field name.
+    /// already in the block, this will add a new entry.
     ///
     /// - Parameter name: The header field name. For maximum compatibility this should be an
     ///     ASCII string. For future-proofing with HTTP/2 lowercase header names are strongly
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -335,6 +335,33 @@ public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiter
...
+    /// Add a sequence of header name/value pairs to the block.
+    ///
+    /// This method is strictly additive: if there are other entries with the same header
+    /// name already in the block, this will add new entries. `add` performs case-insensitive
+    /// comparisons on the header field names.
+    ///
+    /// - Parameter contentsOf: The sequence of header name/value pairs. For maximum compatibility
+    ///     the header should be an ASCII string. For future-proofing with HTTP/2 lowercase header
+    ///     names are strongly recommended.
+    public mutating func add<S: Sequence>(contentsOf other: S) where S.Element == (String, String) {
```
Lukasa: This function should be `@inlinable`.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -335,6 +335,33 @@ public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiter
         }
     }
 
+    /// Add a sequence of header name/value pairs to the block.
+    ///
+    /// This method is strictly additive: if there are other entries with the same header
+    /// name already in the block, this will add new entries. `add` performs case-insensitive
```
Lukasa: This comment looks wrong: we don't compare the field names at all, except inasmuch as we're looking for the Connection header.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -335,6 +334,33 @@ public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiter
...
+    /// This method is strictly additive: if there are other entries with the same header
+    /// name already in the block, this will add new entries.
+    ///
+    /// - Parameter contentsOf: The sequence of header name/value pairs. For maximum compatibility
+    ///     the header should be an ASCII string. For future-proofing with HTTP/2 lowercase header
+    ///     names are strongly recommended.
+    @inlinable
+    public mutating func add<S: Sequence>(contentsOf other: S) where S.Element == (String, String) {
+        precondition(!other.contains { !$0.0.utf8.contains(where: { !$0.isASCII }) }, "names must be ASCII")
+        self.headers.append(contentsOf: other)
```
weissi: why doesn't this call `self.add` for every element (whilst reserving the capacity in `self.headers` beforehands)?

glbrntt: Sure, makes sense!

PR #1005, opened 2019-05-17
TestUtils: EventCounterHandler

weissi:
Motivation:

In many unit tests, it's important to count the events. So far we have
many many ad-hoc handlers to do so but having one that counts all events
would be handy.

Modifications:

Add `EventCounterHandler` to count all events travelling the
`ChannelPipeline`.

Result:

Easier unit testing


Sources/NIOTestUtils/EventCounterHandler.swift
```diff
@@ -0,0 +1,363 @@
...
+
+    public struct ValidityError: Error {
+        public var reason: String
+
+        init(_ reason: String) {
+            self.reason = reason
+        }
+    }
+
+    private let _channelRegisteredCalls = Atomic<Int>(value: 0)
```
Lukasa: Jees, I really want some property delegates now to ease all of this boilerplate!

Sources/NIOTestUtils/EventCounterHandler.swift
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /// Returns the number of `channelRead` events seen so far in the `ChannelPipeline`.
+    public var channelReadCalls: Int {
+        return self._channelReadCalls.load()
+    }
+
+    /// Returns the number of `channelReadComplete` events seen so far in the `ChannelPipeline`.
+    public var channelReadCompleteCalls: Int {
+        return self._channelReadCalls.load()
```
Lukasa: This is returning the wrong atomic value. Looks like you may need a better unit test for this one.

Sources/NIOTestUtils/EventCounterHandler.swift
```diff
@@ -0,0 +1,363 @@
...
+            throw ValidityError("register should be called no more than once")
+        }
+        guard self.bindCalls <= 1 else {
+            throw ValidityError("bind should be called no more than once")
+        }
+        guard self.connectCalls <= 1 else {
+            throw ValidityError("connect should be called no more than once")
+        }
+        guard self.closeCalls <= 1 else {
+            throw ValidityError("close should be called no more than once")
```
Lukasa: This assertion is not universally correct: no bugs will occur if `close` is called more than once, and I don't know if I'd class it as the same kind of error as the others.

Sources/NIOTestUtils/EventCounterHandler.swift
```diff
@@ -0,0 +1,363 @@
...
+
+        if closeCalls != 0 {
+            allEvents.append(("close", self.closeCalls))
+        }
+
+        if triggerUserOutboundEventCalls != 0 {
+            allEvents.append(("triggerUserOutboundEvent", self.triggerUserOutboundEventCalls))
+        }
+
+        return Set(allEvents.map { $0.0 })
```
Lukasa: This function is weird. Why does it create an intermediate array? Why does it load the atomic twice? Why does only the second load use the explicit `self`? Why does it perform the second load at all?

Sources/NIOTestUtils/EventCounterHandler.swift
weissi: thank you! I did now check all individual events.

Sources/NIOTestUtils/EventCounterHandler.swift
weissi: @Lukasa thanks, that's right, will remove

Sources/NIOTestUtils/EventCounterHandler.swift
weissi: oops, leftovers, will fix. Thank you!

PR #1010, opened 2019-05-21
replace redundant commands in docker-compose files with generic one

tomerd:
Motivation: keep it DRY

Modifications: move redundant commands from 18-04-5.0 and 16.04-5.1 to the generic task definitions


Lukasa:
LGTM, will let @weissi review as well.

PR #1016, opened 2019-05-23
Add check_no_api_breakages, a script to check public API breakages

weissi:
Motivation:

It's incredibly hard not to accidentally break the public API, so let's
automate that :).

Modifications:

Add the first version of a public API check script.

Result:

Fewer public API breakages.

### DIFF
--- a/scripts/check_no_api_breakages.sh
+++ b/scripts/check_no_api_breakages.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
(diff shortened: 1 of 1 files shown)


scripts/check_no_api_breakages.sh
```diff
@@ -0,0 +1,108 @@
...
+    local repodir=$1
+    local tag=$2
+    local output=$3
+
+    (
+    cd "$repodir"
+    git checkout "$tag" 2> /dev/null
+    swift build
+    while read -r module; do
+        swift api-digester -dump-sdk -module "$module" \
```
ktoso: Wohoo 🥳

PR #1020, opened 2019-05-30
enable TCP_NODELAY by default

weissi:
Motivation:

Networking software like SwiftNIO that always has explicit flushes
usually does not benefit from having TCP_NODELAY switched off. The
benefits of having it turned on are usually quite substantial and yet we
forced our users for the longest time to enable it manually.

Quite a bit of engineering time has been lost finding performance
problems and it turns out switching TCP_NODELAY on solves them
magically.

Netty has made the switch to TCP_NODELAY on by default, SwiftNIO should
follow.

Modifications:

Enable TCP_NODELAY by default.

Result:

If the user forgot to enable TCP_NODELAY, their software should now be
faster.


weissi:
CC @glbrntt

weissi:
> Regresses allocations due to `reserveCapacity`.

unfortunately with and without :|. I need to look into what's going on there exactly

Lukasa:
Oh sure, I saw that you needed to do it. Just means that at the bare minimum we need to increase the threshold because the CI gets mad. :wink:

PR #1022, opened 2019-05-30
alloc counters: disallow too high limits

weissi:
Motivation:

For the allocation counter tests to be as useful as possible, we should
disallow limits that are too high, we should capture the allocation
count very close to where it actually is.

Modifications:

Don't allow any allocation counter test to allocate fewer than
'limit - 1000' allocations.

Result:

Better limits.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_allocation_counts.sh
+++ b/IntegrationTests/tests_04_performance/test_01_allocation_counts.sh
@@ -46,6 +46,8 @@ for test in "${all_tests[@]}"; do
             warn "    export $max_allowed_env_name=$total_allocations"
         fi
     else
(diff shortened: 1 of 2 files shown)


IntegrationTests/tests_04_performance/test_01_resources/test_bytebuffer_lots_of_rw.swift
```diff
@@ -37,21 +37,6 @@ func run(identifier: String) {
         }
         @inline(never)
         func doReads(buffer: inout ByteBuffer) {
-            /* these ones are zero allocations */
```
weissi: this will go away, just to temporarily test this PR actually works :)

PR #1024, opened 2019-06-03
Add an optimization tips document.

glbrntt:
Motivation:

NIO doesn't offer any tips of on how to write performant code. After
discussing with @weissi we decided it would be useful to gather and
document some useful tips and tricks.

Modifications:

Started an optimization tips document.

Result:

Easier for NIO users to write performant code.

### DIFF
--- a/docs/optimization-tips.md
+++ b/docs/optimization-tips.md
@@ -18,7 +18,9 @@ Anything passing through a NIO pipeline that is not special-cased by `NIOAny`
 (i.e. is not one of `ByteBuffer`, `FileRegion`, `IOData` or
 `AddressedEnvelope<ByteBuffer>`) needs to have careful attention paid to its
 size. If the type isn't one of the aforementioned special cases then it must be
-no greater than 24 bytes in size to avoid a heap-allocation.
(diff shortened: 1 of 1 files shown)


docs/optimization-tips.md
```diff
@@ -0,0 +1,30 @@
...
+and is a great place to start.
+
+Wrapping types in `NIOAny`
+--------------------------
+
+Anything passing through a NIO pipeline that is not special-cased by `NIOAny`
+(i.e. is not one of `ByteBuffer`, `FileRegion`, `IOData` or
+`AddressedEnvelope<ByteBuffer>`) needs to have careful attention paid to its
+size. If the type isn't one of the aforementioned special cases then it must be
+no greater than 24 bytes in size to avoid a heap-allocation.
```
Lukasa: We should elaborate here.

"It must be no greater than 24 bytes in size to avoid a heap-allocation each time it is added to a `NIOAny`."

Lukasa: You may also want to point to the use of `MemoryLayout.size` to work out the size for folks who are unsure.

glbrntt: Good idea :)

PR #1029, opened 2019-06-09
Give EventLoopFuture a Combine Publisher.

Lukasa:
Motivation:

Combine provides users with a nunber of powerful tools for asynchronous
programming that allow users to define a reactive style of programming.
Over time SwiftNIO will want to add a number of hooks for working with
Combine to help enable users to integrate SwiftNIO with their
Combine-using applications.

Modifications:

- Gave EventLoopFuture a Combine Publisher that will provide the result
    of the EventLoopFuture

Result:

Beginnings of Combine integration for SwiftNIO


Sources/NIO/EventLoopFuture+Combine.swift
```diff
@@ -0,0 +1,134 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Combine)
```
ianpartridge: This means that NIO will have different public API depending on whether Combine is available or not.

It feels to me like this code belongs in `swift-nio-extras` or a new `swift-nio-combine` project, so that NIO remains cross-platform.

Lukasa: While that’s true, I don’t know how big of an issue that is. These types are not actually usable without `Combine`: even if you called this API on a platform without `Combine`, you wouldn’t be able to do anything with the returned type, and in order to actually compute on the value you’d have to `import Combine`, which would naturally fail.

I’m interested to see if anyone can actually write an example of this that is both a) reasonable, and b) does not compile on both platforms due to the differing NIO public API. My suspicion is that this can’t be done.

Regardless, I’m open to making a change here if the community and other maintainers think it’s necessary.

ianpartridge: Thanks. This is a SemVer minor change for Darwin, but no change at all for Linux, which seems odd. Hypothetically, would you release SwiftNIO 2.3 containing just this PR?

Lukasa: As a SemVer minor. The appropriate semver status for a change is the max of any hypothetical change on any platform.

Sources/NIO/EventLoopFuture+Combine.swift
```diff
@@ -0,0 +1,134 @@
...
+
+// MARK:- Subscription conformance
+@available(OSX 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+extension EventLoopFuture.Publisher.Subscription: Subscription {
+    @inlinable
+    func request(_ demand: Subscribers.Demand) {
+        // This is unlikely, but double check that this is actually for non-zero demand.
+        if case .max(let limit) = demand, limit <= 0 {
+            return
+        }
```
ktoso: Always good to keep such checks 👍 
The RS thing to do is actually to signal an error -- there's no situation where negative or zero demand makes sense, so you likely got a "somehow wrong" downstream so best to let them know:

In RS it's rule 3.9: https://github.com/reactive-streams/reactive-streams-jvm#3.9

While the Subscription is not cancelled, Subscription.request(long n) MUST signal onError with a java.lang.IllegalArgumentException if the argument is <= 0. The cause message SHOULD explain that non-positive request signals are illegal.

Combine did not really publish or document if it follows all those rules, but this is a good one to apply I believe.

Sources/NIO/EventLoopFuture+Combine.swift
```diff
@@ -0,0 +1,134 @@
...
+        self.future.eventLoop.execute {
+            self.subscriber = nil
+        }
+    }
+
+    @inlinable
+    var combineIdentifier: CombineIdentifier {
+        return CombineIdentifier(self)
+    }
+}
```
ktoso: Looking good, single element ones are quite simple to implement... 
Thanks to riding on the event loop for all things impl is quite nice too 👍

Sources/NIO/EventLoopFuture+Combine.swift
Lukasa: Well the rule must be at least slightly different, because Combine provides [`Demand.none`](https://developer.apple.com/documentation/combine/subscribers/demand/3213707-none) as a helper. I presume this is in place to facilitate the fact that demand can be requested in the return value of [`receive(_:)`](https://developer.apple.com/documentation/combine/subscriber/3213653-receive).

The question is whether the rule is that zero demand may be returned from `receive(_:)` but not passed to `request`, or whether zero demand is allowed in both cases. @parkera, do you have an insight here?

parkera: You can return none here.

parkera: All of our operator implementations assert in request if passed demand <= 0.

ktoso: Right, returning `.none` is one thing, actually invoking upstream's `request(0)` is another IMHO, so one should not tolerate zero demand being signalled into an receive. I'd argue even noop or or other implementations should check `demand > 0`. IMHO best and most useful to signal downstream an error if someone does so, rather than precondition but definitely not silently ignore it..

Makes the returning demand semantics a bit weird, but I know they're useful for performance optimization reasons so so be it.

ktoso: I guess you don't want to signal the error downstream since many impls can be `Failure = Never` thus the decision to make it a fault? That's fair enough to be honest, makes sense given that we only want to signal errors when types allow to after all 🤔

parkera: Right - unlike Java we can't always assume error, or even the type of the error.

Lukasa:
Given the concerns about differing APIs on different platforms, I think this has aged out.

pjuzeliunas:
@Lukasa : is there any chance of resuming this work? Having Combine publisher on an `EventLoopFuture` would bring it to the next level.

Sources/NIO/EventLoopFuture+Combine.swift
jpsim: > I'll be pulling this out to a new target in the near future.

Was this ever done? I'm erasing `EventLoopFuture`s to `AnyPublisher`s in a project of mine and could use something like this.

Lukasa: Hey JP, no, this wasn't done. If you wanted to do the work you could port this change to a new `NIOCombineCompat` target.

PR #1033, opened 2019-06-15
Updated the embedded channel tests to write to the input and output multiple times.

tigerpixel:
Updated the EmbeddedChannel tests to check write to the input and output multiple times.

### Motivation:

The current test check that the buffer can be written to and then that it is cleared. 
They do not check that once the buffer is cleared, it can be written to again. 

### Modifications:

Added a test for both the inbound and outbound buffer to check multiple writes to each buffer.

### Result:

We can not be sure that multiple writes to either the inbound or outbound buffers on the embedded channel works.


tigerpixel:
I wrote these as I couldn't get a test to work on a different branch I have locally. I wanted to make sure this functionality worked, it does. They are quite trivial, so if you think they are overkill or not useful, just close the PR, no explanation necessary. Thanks.

tigerpixel:
sure, sorry I just forgot as it was only a quick change.

Lukasa:

PR #1035, opened 2019-06-15
Update the generate Linux test script to use a computed end date. Run Script.

tigerpixel:
Update the generate Linux test script to use a computed end date in the copyright text and run the script.

### Motivation:

The script has been run multiple times this year but the date still shows as 2018.

### Modifications:

A small edit to the ruby generation script for the Linux test files.
A date change to every file generated by this script.

### Result:

Updated copyright on all Linux generated files for the foreseeable future.

### DIFF
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2019 Apple Inc. and the SwiftNIO project authors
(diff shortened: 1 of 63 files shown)


tigerpixel:
This is a very minor script change, so if you would prefer not to make the edit feel free to reject this PR without explanation. I just thought it was work doing as a bit of bookkeeping, no specific reason why I need it.

tigerpixel:
True. That is a better change. I am currently still working on the WebSockets upgrader and client example. So, I think it is better to add an issue for now and I will come back to it later. I just sent this over as it was very quick to do.

Lukasa:
Closing in favour of #1036.

Issue #1030, opened 2019-06-10
ServerSocketChannel behaves non-sensical if fcntl on accepted socket returns EINVAL


weissi:
As of the docs, `fcntl` on a newly accepted socket shouldn't return `EINVAL` but it seems to sporadically do that on macOS. We need to analyse this. Still, it's probably not a sane behaviour that if that happens, we just close the server socket...

### Expected behavior

don't actually know but just silently closing the server socket isn't good

### Actual behavior

debug mode: crash
release mode: silently close server socket

### Steps to reproduce

dunno

### If possible, minimal yet complete reproducer code (or URL to code)

n/a

### SwiftNIO version/commit hash

all (probably)

### Swift & OS version (output of `swift --version && uname -a`)

macOS

PR #1040, opened 2019-06-18
when making a socket non-blocking, get & OR flags

weissi:
Motivation:

Previously, we would just `F_SETFL, O_NONBLOCK` which would erase all
possibly pre-existing flags which isn't great.

Modifications:

First get the existing flags and then do `F_SETFL, existingFlags |
O_NONBLOCK`.

Result:

Possibly helps with #1030, definitely better style.


Lukasa:
It'd be nice to run a test: you can probably do that by `dup`ing an FD. Want to write it?

PR #1041, opened 2019-06-18
don't close server socket if fcntl(O_NONBLOCK) throws EINVAL

### LINKED ISSUE #1030: ServerSocketChannel behaves non-sensical if fcntl on accepted socket returns EINVAL

weissi:
Motivation:

Darwin does sometimes hand us back an (as of the docs) illegal error
code of EINVAL. So far, NIO has done the non-sensical thing of closing
the server socket as a result.

Modifications:

Just close the accepted socket, leave the server socket open, and fire
the error through the pipeline.

Result:

- fixes #1030
- better behaviour

### DIFF
--- a/Sources/NIO/Channel.swift
+++ b/Sources/NIO/Channel.swift
@@ -349,6 +349,13 @@ public enum ChannelError: Error {
 
 extension ChannelError: Equatable { }
 
+/// `NIOFailedToSetSocketNonBlockingError` indicates that NIO was unable to set a socket to non-blocking mode, either
+/// when connecting a socket as a client or when accepting a socket as a server.
+///
+/// This error should never happen because a socket should always be able to be set to non-blocking mode. Unfortunately,
+/// we have seen this happen on Darwin.
+public struct NIOFailedToSetSocketNonBlockingError: Error {}
+
 /// An `Channel` related event that is passed through the `ChannelPipeline` to notify the user.
 public enum ChannelEvent: Equatable {
     /// `ChannelOptions.allowRemoteHalfClosure` is `true` and input portion of the `Channel` was closed.
--- a/Sources/NIO/SocketChannel.swift
+++ b/Sources/NIO/SocketChannel.swift
@@ -406,16 +406,18 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
             guard self.isOpen else {
                 throw ChannelError.eof
             }
-            if let accepted =  try self.socket.accept(setNonBlocking: true) {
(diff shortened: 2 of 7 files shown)


Sources/NIO/SocketChannel.swift
```diff
@@ -436,6 +438,16 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
             // These are errors we may be able to recover from. The user may just want to stop accepting connections for example
             // or provide some other means of back-pressure. This could be achieved by a custom ChannelDuplexHandler.
             return false
+        case EINVAL:
+            if case .function(let function) = err.reason {
+                // see https://github.com/apple/swift-nio/issues/1030
+                // on Darwin, fcntl(fd, F_SETFL, O_NONBLOCK) sometimes returns EINVAL...
+                //
+                // the comparison is so weird because StaticString isn't Equatable and doesn't have an == method...
+                return String(describing: function) != "fcntl(descriptor:command:value:)"
```
Lukasa: Yuck, this seems like a really easy way to break the code later on. Can we push this error handling deeper into the code in any way? Potentially by catching the fcntl EINVAL and either rethrowing or doing something else? I kinda hate this string-based coupling.

Lukasa: In fact, maybe `accept` should swallow this.

weissi: @Lukasa I think we should deliver it

Lukasa: In that case, `accept` should wrap this error into something clearer.

weissi: @Lukasa agreed, implemented that yesterday but github still shows the 'outdated' change :|

Sources/NIO/Channel.swift
```diff
@@ -349,6 +349,8 @@ public enum ChannelError: Error {
 
 extension ChannelError: Equatable { }
 
+public struct NIOFailedToSetSocketNonBlockingError: Error {}
```
Lukasa: I think this needs a doc comment.

PR #1042, opened 2019-06-18
remove platform availability again

weissi:
Motivation:

The platform availability that we recently introduced caused problems
because it would require all downstream projects to define the same or
narrower platform availability to still compile with NIO which would
make this very disrupting and also SemVer major.

Modifications:

Remove the platform availability declarations.

Result:

No SemVer breakages.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -77,9 +77,6 @@ var targets: [PackageDescription.Target] = [
 
 let package = Package(
     name: "swift-nio",
-    platforms: [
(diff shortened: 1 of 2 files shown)


weissi:
```
19:00:46 removal of container cdfa48b92d773b046680d7a412b7467632326d3122e5afab5de367c39a052e28 is already in progress
```

weissi:

PR #1046, opened 2019-06-19
ByteBuffer.get/readData: Heuristic when to copy/not copy

### LINKED ISSUE #935: heuristics for when getData/readData should copy
we need heuristics for when getData/readData should copy: https://forums.swift.org/t/memory-leaking-in-vapor-app/22209

weissi:
Motivation:

Previously, ByteBuffer tried to never copy the bytes when transferring
to a `Data`. For small copies that's definitely not a good choice
because it means an extra allocation.

Modifications:

- introduce `ByteTransferStrategy` so the user can choose
- make a heuristic: copy is less than 256kB, try not to copy above

Result:

- More sensible defaults.
- fixes #935


Lukasa:

Issue #1052, opened 2019-06-27
Mirror ELF.cascade(to: ELP) onto ELPromise.completeWith(ELF)


ktoso:
### Expected behavior

One often wants to complete a Promise with the value of a Future.

This is possible with `ELF.cascade(to: promise)`, but it is less discoverable if someone is used to the other "way" of thinking about it (e.g. like in Scala https://www.scala-lang.org/api/2.12.1/scala/concurrent/Promise.html#completeWith(other:scala.concurrent.Future[T]):Promise.this.type )

The two "ways" are equivalent, and the APIs could mirror eachother if we wanted to.

### Actual behavior

Allow for `promise.completeWith(future)` for easier discoverability.

Document that it mirrors cascade.

PR #1053, opened 2019-06-27
EventLoopPromise.completeWith(Future), mirroring ELF.cascade(to:) #1052

ktoso:
Adds `EventLoopPromise.completeWith(ELF)`

### Motivation:

Easier discoverability of the promise / future combining APIs.

### Result:
Resolves https://github.com/apple/swift-nio/issues/1052

Now possible to write `promise.completeWith(future)`, which may also be easier to discover using autocomplete, when looking at `promise.succeed / fail` etc.

cc @weissi @Lukasa, let me know if any more docs required etc.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -183,6 +183,19 @@ public struct EventLoopPromise<Value> {
     public func fail(_ error: Error) {
         self._resolve(value: .failure(error))
     }
+    
+    /// Complete the promise with the passed in `EventLoopFuture<Value>`.
+    ///
+    /// This method is equivalent to invoking `future.cascade(to: promise)`,
+    /// but sometimes may read better than its cascade counterpart.
+    /// 
+    /// - parameters:
+    ///     - future: The future whose value will be used to succeed or fail this promise.
+    /// - seealso: `EventLoopFuture.cascade(to:)`
+    @inlinable
+    public func completeWith(_ future: EventLoopFuture<Value>) {
+        future.cascade(to: self)
+    }
 
     /// Fire the associated `EventLoopFuture` on the appropriate event loop.
     ///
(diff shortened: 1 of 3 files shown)


Lukasa:

Sources/NIO/EventLoopFuture.swift
```diff
@@ -183,6 +183,24 @@ public struct EventLoopPromise<Value> {
...
+    /// - seealso: `EventLoopFuture.cascade(to:)`
+    @inlinable
+    public func completeWith(_ future: EventLoopFuture<Value>) {
+        future.whenComplete { result in
+            switch result {
+            case let .success(value): self.succeed(value)
+            case let .failure(error): self.fail(error)
+            }
+        }
+    }
```
ktoso: no strong opinion about expressing it in line OR delegating to `future.cascade(to: self)`, let me know if better to delegate.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -183,6 +183,24 @@ public struct EventLoopPromise<Value> {
...
+    ///     - future: The future whose value will be used to succeed or fail this promise.
+    /// - seealso: `EventLoopFuture.cascade(to:)`
+    @inlinable
+    public func completeWith(_ future: EventLoopFuture<Value>) {
+        future.whenComplete { result in
+            switch result {
+            case let .success(value): self.succeed(value)
+            case let .failure(error): self.fail(error)
+            }
+        }
```
Lukasa: Should this just literally call `cascade`?

Sources/NIO/EventLoopFuture.swift
Lukasa: Heh, I just wrote that comment. Let's delegate.

ktoso: ok :)

ktoso:
And done via delegation to cascade now 👍

ktoso:
Ah ofc forgot about that, will do later. On my own repos got that hooked up as got hook so forgot to do so here, sorry :)

ktoso:
Amended the PR https://github.com/apple/swift-nio/pull/1053/commits/ba5b07d1a130b07d293381f3492d949ec1a484e4

weissi:
thanks @ktoso

PR #1059, opened 2019-07-05
Support building NIO with CMake

kevints:
Adds support for building NIO, packaging or installing its demo programs and running its integration tests with CMake.

### Motivation:

Recently @compnerd [announced Swift support for CMake](https://forums.swift.org/t/announcing-swift-support-in-cmake/24792). While CMake is a more complex build tool to use than Swift Package Manager, it has a few advantages that make it well-suited as an alternative for NIO. I figure a working PR is as good a place as any to start the conversation about adding another buildsystem to swift-nio. The places it shines are highlighted in this PR:

* It supports installation of build products and generation of a number of package formats. This is useful to NIO, for example, you might want to package binaries of `NIOPerformanceTester` to deploy on a number of servers that you're evaluating.
* It supports complex build automation with a number of custom steps. This is useful for NIO's integration tests, which devote a lot of space to building custom variants of NIO libraries with special build settings.
* It makes embedding NIO into other CMake projects easier. This could be useful with https://github.com/apple/swift-corelibs-foundation, as that project already uses CMake.

Of course it's also worth pointing out its drawbacks:

* Publication of the swift-nio package is much better supported by Swift Package Manager because of its declarative syntax, Semantic Versioning enforcement, and deep integration with tooling like Xcode.
* CMake doesn't currently support running the unit tests under Tests due to lack of integration with XCTest.
* Since NIO doesn't have a stable ABI, there's no point in using the library packaging support in CMake to package NIO - if you're using it as a library you will still want to depend on it with SwiftPM. The binary packaging support is still useful though, as you can distribute command-line tools like `NIOPerformanceTester`.
* Since the Swift support in CMake is experimental it's subject to frequent breakage for a while.

### Modifications:

Created a new CMake buildsystem (`CMakeLists.txt`). The NIO source code is unmodified with a couple of minor exceptions:

* I created explicit `module.modulemap` files for each C library. Swift Package Manager generates these automatically.
* I modified the HTTP1 integration tests to optionally skip building using SwiftPM when invoked by cmake (so that CMake+Ninja can take care of building the NIO binaries themselves).

### Result:

You can build NIO with CMake's experimental Swift support.


CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/SwiftNIODemo)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/swiftmodules)
+list(APPEND CMAKE_BUILD_RPATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+
+add_compile_options($<$<COMPILE_LANGUAGE:Swift>:-warnings-as-errors>)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+  add_compile_options("$<$<COMPILE_LANGUAGE:Swift>:-target;x86_64-apple-macosx10.14>")
```
compnerd: This really shouldn't be needed.  `-DCMAKE_SYSTEM_NAME=Darwin` and `-DCMAKE_SYSTEM_PROCESSOR=x86_64` should do the right thing.  To pass the target explicitly, use `-DCMAKE_Swift_COMPILER_TARGET=`.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+
+install(TARGETS ${NIO_EXECUTABLE_PRODUCTS} ${NIO_LIBRARY_PRODUCTS}
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/SwiftNIODemo
+)
+
+### MARK - Hacks
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+  # XXX: CMake doesn't set LC_ID on Swift SHARED libraries, so we need to fix install_name here
```
compnerd: I *think* that I fixed the `LC_ID` issue in a newer CMake.

CMakeLists.txt
```diff
@@ -0,0 +1,213 @@
...
+  foreach(nio_library IN LISTS NIO_Swift_LIBRARIES)
+    target_link_options(${nio_library} PRIVATE LINKER:-install_name,@rpath/$<TARGET_LINKER_FILE_NAME:${nio_library}>)
+    target_link_options(${nio_library} INTERFACE LINKER:-rpath,@executable_path/../lib/${PROJECT_NAME})
+  endforeach()
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  foreach(nio_library IN LISTS NIO_Swift_LIBRARIES)
+    # XXX: CMake doesn't set SONAME on Swift SHARED libraries, so we need to fix it here
```
compnerd: This definitely was fixed in a later release.

weissi:
@Lukasa / @yim-lee did we have a "generate CMake from Package.swift" script somewhere that we could use for NIO?

Lukasa:
Crypto has the beginnings of one.

Lukasa:
This PR is substantially out of date and there is currently no-one who needs the functionality, so for now we're going to drop this.

PR #1070, opened 2019-07-16
ELF: On reduce, hop to target EventLoop before doing operation

### LINKED ISSUE #1054: Investigate if ELF.reduce called from outside the EL don’t do way to much hopping
I think if you call WLF.reduce from outside the target EL, it will do a lot of unnecessary EventLoop hopping for no good reason.

I think we could just hop over once and then hit the fast path. Needs investigation though.

weissi:
Motivation:

Folds often involves chaining multiple futures together. It's
beneficial if you're already on the target EventLoop as it might
vastly reduce the amount of EventLoop hops.

Modifications:

Execute onto the target EventLoop before doing the folds.

Result:

Fixes #1054


weissi:

PR #1080, opened 2019-07-23
Make pipeline handler behave better when removed

Lukasa:
Motivation:

The HTTPServerPipelineHandler is removed from pipelines in many cases,
but the most common case is over upgrade. While the handler is
removable, it doesn't make any effort to ensure that it leaves the
pipeline in a sensible state, which is pretty awkward.

In particular, there are 3 things the pipeline handler may be holding
on to that can lead to damage. The first is pipelined requests: if there
are any, they should be delivered, as the user may be deliberately
allowing pipelining.

The second thing is read() calls. The HTTPServerPipelineHandler exerts
backpressure on clients that aggressively pipeline by refusing to read
from the socket. If that happens, and then the handler is removed from
the channel, it will "forget" to restart reading from the socket on the
way out. That leaves the channel quietly in a state where no reads will
occur ever again, which is pretty uncool.

The third thing is quiescing. The HTTPServerPipelineHandler catches
quiescing events and allows them to deliver a response before closing a
connection. If that has happened when the pipeline handler is removed,
it should fall back to the behaviour as though it were not there.

Modifications:

- Added a handlerRemoved implementation to play event state that should
    be replayed.
- Added a channelInactive implementation to drop data.

Result:

More graceful handler removal.

### DIFF
--- a/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
+++ b/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
@@ -164,6 +164,9 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
 
         /// Quiescing and the last request's `.end` has been seen which means we no longer accept any input.
         case quiescingLastRequestEndReceived
+
+        /// Quiescing and we have issued a channel close. Further I/O here is not expected, and won't be managed.
+        case quiescingCompleted
     }
 
     private var lifecycleState: LifecycleState = .acceptingEvents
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -340,6 +340,56 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+                context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
+            case .error(let error):
+                context.fireErrorCaught(error)
+            }
+        }
+
+
+        if case .quiescingWaitingForRequestEnd = self.lifecycleState {
+            context.fireUserInboundEventTriggered(ChannelShouldQuiesceEvent())
+        } else if self.readPending {
```
weissi: quick Q: why aren't we doing the `read` if we're waiting for the `.end`?

Lukasa: If `readPending` isn't true, we haven't swallowed a call to `read`, so we don't need to fire one. Presumably another one is coming.

weissi: @Lukasa I meant if we have `.quiescingWaitingForRequestEnd == self.lifecycleState` AND `self.readPending == true`

Lukasa: Oh yeah, that's an artefact of my previous version. In that version if we were quiescing then I'd close the channel, rather than replay the event. Good catch, will fix.

Lukasa: Ok, amended. Sadly that made the patch a lot larger, as it requires we keep a bit better track of whether we've actually _completed_ the quiescing to avoid double-notifying.

PR #1084, opened 2019-07-29
Tolerate unsolicited HTTP responses.

Lukasa:
Motivation:

Right now NIO clients will crash in the (admittedly unlikely) event that
a HTTP server sends an unsolicited extra response on a connection. This
seems pretty unnecessary, and fairly graceless.

Modifications:

- Replace a forcible pop with an optional one.
- Added NIOHTTPDecoderError type.

Result:

Removed a crash that we definitely don't want.

### DIFF
--- a/Tests/NIOHTTP1Tests/HTTPDecoderTest.swift
+++ b/Tests/NIOHTTP1Tests/HTTPDecoderTest.swift
@@ -745,4 +745,44 @@ class HTTPDecoderTest: XCTestCase {
         XCTAssertNoThrow(XCTAssert(try channel.finish().isClean))
     }
 
+    func testAppropriateErrorWhenReceivingUnsolicitedResponse() throws {
+        let channel = EmbeddedChannel()
+        var buffer = channel.allocator.buffer(capacity: 64)
+        buffer.writeStaticString("HTTP/1.1 200 OK\r\nServer: a-bad-server/1.0.0\r\n\r\n")
+
+        let decoder = ByteToMessageHandler(HTTPResponseDecoder(leftOverBytesStrategy: .dropBytes))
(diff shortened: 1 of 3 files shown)


Tests/NIOHTTP1Tests/HTTPDecoderTest.swift
```diff
@@ -745,4 +745,17 @@ class HTTPDecoderTest: XCTestCase {
         XCTAssertNoThrow(XCTAssert(try channel.finish().isClean))
     }
 
+    func testAppropriateErrorWhenReceivingUnsolicitedResponse() throws {
```
weissi: could we add a test case that verifies that we also cannot silently 'recover' from having a bad server? Ie:

- `-> HTTP/1.1 200 OK\r\nServer: a-bad-server/1.0.0\r\n\r\n`
- `<- GET / HTTP/1.1`
- `-> HTTP/1.1 200 OK\r\nServer: a-bad-server/1.0.0\r\n\r\n`

isn't accepted

Lukasa: I think right now it absolutely will be. Do we need to defend against that behaviour? This seems like the kind of thing we normally leave to having a pipeline error handler.

weissi: @Lukasa I think we should protect. Not that we should close the connection but I believe we should consistently error. A server that starts responding before a request isn't quite right and I think we should continue sending errors. But happy to discuss that.

Lukasa: Ok, test added to verify.

PR #1091, opened 2019-07-31
fix double remove of handlers

weissi:
Motivation:

Previously, it was possible that a handler was removed twice: Once by
Channel teardown and another time by a user-trigger removal that wasn't
instantaneous. This racy situation can happen in the real world. NIO
behaved wrongly in two ways:

1. we would call `handlerRemoved` twiced
2. ByteToMessageHandler would fail an assertion about the current
   removal state

Modifications:

- Only call `handlerRemoved` when the handler actually gets removed.
- Fix the assertion.

Result:

fewer bugs & crashes

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -500,7 +500,7 @@ public final class ChannelPipeline: ChannelInvoker {
         }
 
         guard inThePipeline else {
-            // if both next and prev are nil already, the we were previously removed from the pipeline
+            // if both next and prev are nil already, then we were previously removed from the pipeline
             promise?.succeed(())
             return
         }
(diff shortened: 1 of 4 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -488,13 +488,23 @@ public final class ChannelPipeline: ChannelInvoker {
...
+            inThePipeline = true
             prevCtx.next = nextCtx
         }
         if let nextCtx = nextCtx {
+            inThePipeline = true
             nextCtx.prev = prevCtx
         }
 
+        guard inThePipeline else {
+            // if both next and prev are nil already, the we were previously removed from the pipeline
```
Lukasa: nit: s/the/then/

Lukasa:
```suggestion
            // if both next and prev are nil already, then we were previously removed from the pipeline
```

PR #1093, opened 2019-08-06
HTTPDecoder: don't deliver unsolicited responses

weissi:
Motivation:

Despite the fact that we already stopped delivering unsolicited
responses for the HTTP response decoder, there was a window where we
would deliver a second .head for a response that come without a request.

Modifications:

Don't deliver a second .head even if send together with a first, legit
response.

Result:

Being a HTTP client using NIO is now easier.

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -265,15 +265,8 @@ private class BetterHTTPParser {
         }
         assert(self.firstNonDiscardableOffset == nil)
         self.decodingState = .headersComplete
-        let success = self.delegate.didFinishHead(versionMajor: versionMajor,
-                                                  versionMinor: versionMinor,
-                                                  isUpgrade: isUpgrade,
-                                                  method: method,
-                                                  statusCode: statusCode,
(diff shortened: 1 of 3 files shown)


weissi:

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -296,24 +289,30 @@ private class BetterHTTPParser {
             // does not meet the requirement of RFC 7230. This is an outstanding http_parser issue:
             // https://github.com/nodejs/http-parser/issues/251. As a result, we check for these status
             // codes and override http_parser's handling as well.
-            if self.requestHeads.count > 0 {
-                let method = self.requestHeads.removeFirst().method
-                if method == .HEAD || method == .CONNECT {
-                    return .skipBody
+            if let method = self.requestHeads.popFirst()?.method {
```
Lukasa: This whole thing seems like it would read better as a `guard let`, not an `if let`. That will also let us keep the `method` and `statusCode` if blocks separate as they were before, which I find is substantially less mentally taxing to understand.

weissi:

PR #1096, opened 2019-08-07
Fail promise in MessageToByteHandler write error cases

wlisac:
This fails the `write` promise in `MessageToByteHandler` for all write error cases.

### Motivation:
The promise is leaked if the encoder fails or if the handler is an error state.

### Modifications:
Call `fail` on the `write` promise if the encoder fails or if the handler is an error state.

### Result:
The `write` promise will now fail in all error cases.

### Additional Information
I debated on if the promise should be failed before or after `context.fireErrorCaught`. I found some prior art, here: https://github.com/apple/swift-nio-http2/blob/master/Sources/NIOHTTP2/HTTP2ToHTTP1Codec.swift#L98

I also debated on if the promise should be failed before or after the state in the handler has been changed to `error`. I decided to fail it after changing the state to `error` since I was concerned about the promise failing, calling write again, and triggering another error before the first error state was set.

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -767,6 +767,7 @@ extension MessageToByteHandler {
         case .notInChannelYet:
             preconditionFailure("MessageToByteHandler.write called before it was added to a Channel")
         case .error(let error):
+            promise?.fail(error)
             context.fireErrorCaught(error)
             return
         case .done:
@@ -784,6 +785,7 @@ extension MessageToByteHandler {
             context.write(self.wrapOutboundOut(self.buffer!), promise: promise)
         } catch {
             self.state = .error(error)
+            promise?.fail(error)
             context.fireErrorCaught(error)
         }
     }
--- a/Tests/LinuxMain.swift
+++ b/Tests/LinuxMain.swift
@@ -75,6 +75,7 @@ import XCTest
          testCase(IntegerTypesTest.allTests),
          testCase(MarkedCircularBufferTests.allTests),
          testCase(MessageToByteEncoderTest.allTests),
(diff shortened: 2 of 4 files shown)


Lukasa:

wlisac:
Thanks @Lukasa. I added a test case that exercises both code paths. 

I considered breaking it into two tests, but they looked almost identical. Still happy to break into two distinct tests if you'd prefer.

As an aside, while I was looking at related tests, it looks like there may be an issue with `MessageToByteEncoderTest`. 

Notice that `testEncoder` ignores the passed in handler. If you honor the passed in handler, then there's an assertion failure that gets triggered.
```
XCTAssertEqual(MemoryLayout<Int32>.size, 256)
```

I didn't want to bloat this PR with additional changes, but I wanted to call it out in case something should be changed there :)

Appreciate the review!

– Will

Lukasa:
Yeah, that test is perplexing. Thanks for not placing the changes into this patch, I'd much rather have a fixup for those tests in a separate PR.

Naturally, that assertion will always fail, as the size of an Int32 is pretty clearly 4. That invalid assertion has been in the code a long time so I don't know what it was intended to test, but I think it was actually _supposed_ to assert on the capacity of the provided buffer. This makes sense if you look at the version of the file from before #868:  https://github.com/apple/swift-nio/blob/cde067e603c0569d522f0dc0bd9d8f804ba40b26/Tests/NIOTests/CodecTest.swift#L1155-L1185

In this code, it's clear that the purpose of the twin tests was to validate that you could override the default allocator. That test is no longer useful as you now _cannot_ override the default allocator, so we should probably just collapse that test down to one.

Lukasa:

Issue #1098, opened 2019-08-08
CircularBuffer not equatable


weissi:
CircularBuffer isn’t equatable which is very surprising. It should have a

```swift
extension CircularBuffer: Equatable where Element: Equatable {
...
}
```

This bug should be implementable with only some Swift knowledge, no special SwiftNIO knowledge necessary.

adtrevor:
#1101 Solves this! 🙂

PR #1101, opened 2019-08-09
Add conditional conformances of CircularBuffer to Equatable and Hashable

adtrevor:
CircularBuffer is now `Equatable` when its `Element` is `Equatable`, and `Hashable` when its `Element` is `Hashable`

Solves #1098

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -692,3 +692,17 @@ extension CircularBuffer {
         return self.verifyInvariants() && self.unreachableAreNil()
     }
 }
+
+extension CircularBuffer: Equatable where Element: Equatable {
+    public static func ==(lhs: CircularBuffer, rhs: CircularBuffer) -> Bool {
+        return lhs.count == rhs.count && zip(lhs, rhs).allSatisfy(==)
+    }
+}
+
+extension CircularBuffer: Hashable where Element: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        for element in self {
+            hasher.combine(element)
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -692,3 +692,16 @@ extension CircularBuffer {
         return self.verifyInvariants() && self.unreachableAreNil()
     }
 }
+
+extension CircularBuffer: Equatable where Element: Equatable {
+    public static func ==(lhs: CircularBuffer, rhs: CircularBuffer) -> Bool { lhs.count == rhs.count && zip(lhs, rhs).allSatisfy(==)
```
Lukasa: No prizes for avoiding the extra line here: add a newline after `{` :wink:

adtrevor: Oops, copy-pasted it from GitHub into another branch and the newline was lost 😅

adtrevor: @Lukasa I don't understand what's happening, because when I go directly to my branch here: https://github.com/adtrevor/swift-nio/commit/25553e86b4b44fdc150207da1333841f5ce5108b
Everything is correct, and it's also the case in Xcode ! 😮

Lukasa:
I think we're still missing the tests.

adtrevor:
@Lukasa Everything should now be ok! 🙂

Lukasa:

Sources/NIO/CircularBuffer.swift
```diff
@@ -692,3 +692,17 @@ extension CircularBuffer {
         return self.verifyInvariants() && self.unreachableAreNil()
     }
 }
+
+extension CircularBuffer: Equatable where Element: Equatable {
+    public static func ==(lhs: CircularBuffer, rhs: CircularBuffer) -> Bool {
+        lhs.count == rhs.count && zip(lhs, rhs).allSatisfy(==)
```
Lukasa: We need Swift 5 compatibility, so sadly for you the `return` statement cannot be elided here.

adtrevor: Fixed! 🙂

Lukasa:

PR #1108, opened 2019-08-13
bootstraps: tolerate random EventLoops

weissi:
Motivation:

Although that's a rare usecase, the Bootstraps must support random
EventLoops. That includes the EventLoops they're called on but also the
EventLoops of the futures returned from the channel initializers.
Thanks to @adtrevor for reporting the initial bug which also served as a tree
to be shaken for more  bugs.

Modifications:

Make sure we're always on the correct EventLoop.

Result:

Fewer crashes and bugs.


adtrevor:
Thanks a lot for the quick fix

PR #1116, opened 2019-08-19
Also collect total bytes allocated in alloc tests #1115

### LINKED ISSUE #1115: Alloc counter tests to also gather "total bytes" allocated
### Expected behavior
The number of malloc/free calls is very useful, and good to check regressions.

We would also want to check how much memory is allocated though, more for cases about learning about the footprint of specific use case than setting a hard limit on it though.

### Actual behavior

Actual nr. of allocated bytes is not tracked.

I can provide a PR for this in a bit, as I had it modified for some of my experimentation work recently.

ktoso:
Motivation:

It is useful/interesting to have an overview about total bytes
allocated, not only the number of malloc calls.

This is not used to fail tests though, more as an informational thing.

Modifications:

Introduce and hit new counter with allocated size whenever allocations
happen.

Result:

- It is possible to inspect the total allocated bytes.
- Resolves #1115


ktoso:
Though it can be debated if / how / when / where such counting should take place.
I had a need to write such tests to inspect some things, and either we'll end up having the mini framework copied or we can push these upstream... 

One could argue that those counters are more interesting "development time", though then the question remains where and how to insert them when I want to explore memory footprint etc. 🤔

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
```diff
@@ -24,8 +24,10 @@ function die() {
 
 function make_git_commit_all() {
     git init > /dev/null
-    git config --local user.email does@really-not.matter
-    git config --local user.name 'Does Not Matter'
+    if [[ "$(git config user.email)" == ""]]; then
+        git config --local user.email does@really-not.matter
+        git config --local user.name 'Does Not Matter'
+    fi
```
ktoso: matters if someone (me) has globally configured "always sign all commits"

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
```diff
@@ -22,4 +22,8 @@ void inc_malloc_counter(void);
 void reset_malloc_counter(void);
 long read_malloc_counter(void);
 
+void add_malloc_bytes_counter(int v);
```
Lukasa: This function doesn't exist: `MAKE_COUNTER` doesn't define an `add_counter` function and you haven't extended the macro to use this, so I suspect this would fail to link.

ktoso: Right, sorry; sloppy copy-pasta-ing; In general, we happy to merge such new counter?

Lukasa: @weissi owns this mostly. I don't care, but he may care.

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
```diff
@@ -24,8 +24,10 @@ function die() {
 
 function make_git_commit_all() {
     git init > /dev/null
-    git config --local user.email does@really-not.matter
-    git config --local user.name 'Does Not Matter'
+    if [[ "$(git config user.email)" == ""]]; then
```
weissi: nit, space missing before `]]`

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
weissi: haha sorry :)

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
weissi: I'm happy with that. But the type should match `malloc`'s argument and be `size_t`.

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
ktoso: :)

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
ktoso: Right, thanks :) Sorry for the messy PR, should have vetted my hacky thing before I cherry picked it over

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
ktoso: THanks :)

ktoso:
Comments addresed, sorry that was a bit messy; thanks for accepting, will make my life a bit simpler :)

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/src/atomic-counter.c
```diff
@@ -20,6 +20,9 @@
 */    atomic_fetch_add_explicit(&g_ ## name ## _counter, 1, memory_order_relaxed); /*
 */ } /*
 */ /*
+*/ void add_ ## name ## _counter(int v) { /*
```
weissi: @ktoso this still uses `int` but it needs to use either `int` or `size_t` I think. We could fix  `MAKE_COUNTER` like `#define MAKE_COUNTER(_type, _name) ...`. The other option would probably be to make everything `ssize_t` or something but that's a bit annoying as we'd get some signed/unsigned conversions.

weissi: actually making everything `intptr_t` might be the right call. That matches Swift's `Int` which is the type that we propagate to the user in Swift.

ktoso: Is changing int here to `size_t` good enough?

weissi: @ktoso `size_t` is unsigned

ktoso: Right; summing up though, those to become accepting `size_t` then and we're good...?

weissi: @ktoso if we have one type for all, it has to be `intptr_t`, right? In the overridden `malloc` function we then need to do the (explicit) cast.

Lukasa: Or `ssize_t`?

weissi: wanted to match `Swift.Int` as closely as possible. Pretty much all platforms that matter have pointer size == word size so `intptr_t` is kinda good. Exceptions are `x32` and Watch Series 4 and maybe WoW64?

ktoso: Sounds like `intptr_t` it is then; I'll check that tomorrow and check why CI failed. 
Sorry for being slow here

ktoso:
Sorry for delay, will get back to this today or tomorrow

weissi:
friendly ping @ktoso

ktoso:
Whoa that's been a long "tomorrow" :-/

To conclude https://github.com/apple/swift-nio/pull/1116#discussion_r315679328 , the type should be `intptr_t` then? I'll make the change.

Terribly sorry for leaving it hanging such long time

weissi:
@ktoso don't apologise! I'm happy with `intptr_t` or `uintptr_t`, whatever is easiest for you.

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
```diff
@@ -22,7 +22,7 @@ void inc_malloc_counter(void);
 void reset_malloc_counter(void);
 long read_malloc_counter(void);
 
-void add_malloc_bytes_counter(size_t v);
+void add_malloc_bytes_counter(intptr_t v);
 void reset_malloc_bytes_counter(void);
 long read_malloc_bytes_counter(void);
```
ktoso: I think that's what you meant in https://github.com/apple/swift-nio/pull/1116/files#r315248349 @weissi ?
Please sanity check, I'm not 100% sure that's right.

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
```diff
@@ -22,4 +22,8 @@ void inc_malloc_counter(void);
 void reset_malloc_counter(void);
 long read_malloc_counter(void);
 
+void add_malloc_bytes_counter(intptr_t v);
+void reset_malloc_bytes_counter(void);
+long read_malloc_bytes_counter(void);
```
weissi: there's a `long` here that wants to be `intptr_t` I think.

weissi:
@ktoso the integration tests failed. You should be able to run them locally with

    IntegrationTests/run-tests -d -f allocation

the `-d` enables debug and the `-f` filters out (a regex).

IntegrationTests/allocation-counter-tests-framework/template/AtomicCounter/Sources/AtomicCounter/include/atomic-counter.h
weissi: yes, thanks!

ktoso:
Updating branch.

weissi:
@Lukasa ping

PR #1119, opened 2019-08-20
Fix dead link to process_test_files.rb in NOTICE.txt

ktoso:
Fix dead link in NOTICE.txt

### Motivation:

Previous link is dead, so we might want to point to a specific commit that introduced what our scripts are based on.

### Modifications:

- add new links to same authors script

### Result:

- up to date and more "future-proof" links in NOTICE


ktoso:
Ah, and thanks to @normanmaurer for the detective work 🕵

Lukasa:
Thanks @ktoso!

PR #1124, opened 2019-08-29
Allow promises to be completed with a `Result<Value, Error>`

glbrntt:
Motivation:

When dealing with `Result` types and promises it is inconvenient having to
switch over the result in order to succeed or fail a promise.

Modifications:

- Allow promises to be completed with a `Result<Value, Error>` directly.
- Added tests.

Result:

Promises are easier to fulfill with `Result` types.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -197,6 +197,25 @@ public struct EventLoopPromise<Value> {
         future.cascade(to: self)
     }
 
+    /// Complete the promise with the passed in `Result<Value, Error>`.
+    ///
+    /// This method is equivalent to invoking:
+    /// ```
+    /// switch result {
+    /// case .success(let value):
+    ///     promise.succeed(value)
+    /// case .failure(let error):
+    ///     promise.fail(error)
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - result: The result which will be used to succeed or fail this promise.
+    @inlinable
+    public func completeWith(_ result: Result<Value, Error>) {
+        self._resolve(value: result)
+    }
+
     /// Fire the associated `EventLoopFuture` on the appropriate event loop.
     ///
     /// This method provides the primary difference between the `EventLoopPromise` and most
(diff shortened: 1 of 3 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -197,6 +197,25 @@ public struct EventLoopPromise<Value> {
...
+    ///     promise.succeed(value)
+    /// case .failure(let error):
+    ///     promise.fail(error)
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - result: The result which will be used to succeed or fail this promise.
+    @inlinable
+    public func completeWith(_ result: Result<Value, Error>) {
```
glbrntt: Not super happy about the overloading of `completeWith` but didn't want to add an explicit `result` label.

weissi: let's CC @Lukasa  & @normanmaurer for the naming. Could also always do `public func fulfill(_ result: Result)`?

t089: I think `fulfill(_ result: Result)` is nice, gives you a nice and readable way to pass a promise "as a completion handler" for callback based APIs

```
func doSomething(completion: @escaping (Result<String, Error>) -> ()) {
  ...
  completion(.success("Done"))
}

let p: EventLoopPromise<String> = ...

doSomething(completion: p.fulfill)

p.futureResult.whenComplete { ... }
```

ktoso: So far I've not yet seen how such kind of overload could end up being problematic... as it reads quite well (and going for another name like `fulfill` or other, seems a bit ad hoc if only to avoid an overload) I'm not sure if avoiding the overload is worth it here... Any secret swift knowledge I'm missing about such specific overload is bad?

(more complex ones, esp. with differing return types I'd be on board with avoiding, here it seems to help more than hurt I thought)

glbrntt: I'm actually fine with `completeWith(_ result: Result)` but I know not everyone is a fan of overloads which is why I called it out.

I'm 👎 on `fulfil` for the consistency reason.

weissi: I don't like overloads but in this particular case I'm probably fine with it.

weissi:

weissi:
@Lukasa are you happy with this API?

PR #1129, opened 2019-09-04
Avoid having `cascade` be title of all methods in jazzy docs

ktoso:
Motivation:

Documentation in jazzy looks misleading, listing all methods under "cascade" since no new MARKs are defined; 

Like this:

Screen Shot 2019-09-04 at 18 59 18

Modifications:

- declare more MARKs, it's repetetive but not wrong now

Result:

- Nicer docs

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -430,6 +430,8 @@ extension EventLoopFuture: Equatable {
     }
 }
 
+// MARK: flatMap and map
+
 // 'flatMap' and 'map' implementations. This is really the key of the entire system.
 extension EventLoopFuture {
     /// When the current `EventLoopFuture<Value>` is fulfilled, run the provided callback,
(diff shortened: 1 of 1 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -603,6 +605,33 @@ extension EventLoopFuture {
...
+    /// event loop. `flatMapResult` is intended for use when you have a data-driven function that
+    /// performs a simple data transformation that can potentially error.
+    ///
+    ///
+    /// - parameters:
+    ///     - body: Function that will receive the value of this `EventLoopFuture` and return
+    ///         a new value or error lifted into a new `EventLoopFuture`.
+    /// - returns: A future that will receive the eventual value.
+    @inlinable
+    public func flatMapResult<NewValue, SomeError: Error>(file: StaticString = #file,
```
ktoso: this moves the flatmap from where all `reduce` were defined, to where all the `flatMap` are defined

ktoso:
Results in: 
Screen Shot 2019-09-04 at 19 11 35


Repetetive, but not wrong ¯\\_(ツ)_/¯

weissi:

PR #1131, opened 2019-09-06
Lock: guarantee crash instead of undefined behaviour

weissi:
Motivation:

NIO's Lock currently just deadlocks if a thread which already holds a
lock tries to reacquire it. This however isn't even really defined
behaviour.

Modifications:

Switch us to a guaranteed crash from a unguarnateed hang.

Result:

Easier to debug deadlocks with NIO's lock.


weissi:

weissi:

PR #1133, opened 2019-09-08
ChannelPipeline: fail double removal of handler

weissi:
### Motivation:

If a user attempted to double remove a handler, we would make this the user's handler's problem. That would lead to pretty random situations, sometimes crashed because assertions were violated.

### Modifications:

Fail every removal attempt as soon as the removal has begun.

### Result:

Much more predictable outcome of problems in user code around double removal.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -400,11 +400,15 @@ public final class ChannelPipeline: ChannelInvoker {
             promise?.fail(ChannelError.unremovableHandler)
             return
         }
+        func removeHandler0() {
+            context.startUserTriggeredRemoval(promise: promise)
+        }
+
         if self.eventLoop.inEventLoop {
-            handler.removeHandler(context: context, removalToken: .init(promise: promise))
+            removeHandler0()
         } else {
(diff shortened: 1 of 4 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -400,11 +400,20 @@ public final class ChannelPipeline: ChannelInvoker {
...
         }
-        if self.eventLoop.inEventLoop {
+        func removeHandler0() {
+            guard !context.userTriggeredRemovalStarted else {
+                promise?.fail(NIOAttemptedToRemoveHandlerMultipleTimesError())
+                return
+            }
+            context.userTriggeredRemovalStarted = true
             handler.removeHandler(context: context, removalToken: .init(promise: promise))
+        }
```
Lukasa: Given that his function touches `context` a few times, would it be better to have this be a function on `ChannelHandlerContext` instead? That would also allow `userTriggeredRemovalStarted` to be a `private var` instead of an `internal var`.

weissi: @Lukasa sounds good, like the way I wrote it now?

weissi:

PR #1138, opened 2019-09-12
add PipeChannel

weissi:
Motivation:

There are use-cases for SwiftNIO where there are no actual sockets but
rather two pipe file descriptors - one for input, one for output.
There's no real reason why SwiftNIO shouldn't work for those.

Modifications:

Add a PipeChannel.

Result:

More use-cases for SwiftNIO.


Sources/NIO/BaseStreamSocketChannel.swift
```diff
@@ -0,0 +1,243 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket> {
```
weissi: everything here is moved from `BaseSocketChannel`. Basically the lines that are now left in `BaseSocketChannel` plus `BaseStreamSocketChannel` are pretty much exactly what used to be in `BaseSocketChannel`.

Sources/NIO/SocketChannel.swift
```diff
@@ -32,42 +32,21 @@ private extension ByteBuffer {
 /// A `Channel` for a client socket.
 ///
 /// - note: All operations on `SocketChannel` are thread-safe.
-final class SocketChannel: BaseSocketChannel<Socket> {
-
+final class SocketChannel: BaseStreamSocketChannel<Socket> {
     private var connectTimeout: TimeAmount? = nil
-    private var connectTimeoutScheduled: Scheduled<Void>?
```
weissi: all the removed lines here, you'll find in `BaseStreamSocketChannel` above.

Tests/NIOTests/StreamChannelsTest.swift
```diff
@@ -0,0 +1,221 @@
...
+            XCTAssertNoThrow(try writeFullyDonePromise.futureResult.wait())
+            // To finish up, let's just tear this down.
+            XCTAssertNoThrow(try chan2.close().wait())
+            XCTAssertNoThrow(try chan1.closeFuture.wait())
+        }
+        XCTAssertNoThrow(try forEachCrossConnectedStreamChannelPair(runTest))
+    }
+
+    ///
+    /// TODO: Re-enable test
```
weissi: this will need fixing before we can merge.

weissi:
@Lukasa / @normanmaurer this is getting pretty much ready. We need to settle on the correct behaviour for half-close from both sides and I'll add some more tests but the general shape is getting there. Feel free to have an early peek :)

weissi:
At the moment, we're regressing 2 allocations per connection :|.

Sources/NIO/BaseSocket.swift
```diff
@@ -12,6 +12,28 @@
...
 
+import NIOConcurrencyHelpers
+
+private let globallyIgnoredSIGPIPE: Atomic<Bool> = .init(value: false)
+
+func ignoreSIGPIPE(descriptor fd: CInt) throws {
+    if !globallyIgnoredSIGPIPE.load() {
+        #if os(Linux)
+        /* no F_SETNOSIGPIPE on Linux :( */
+        _ = unsafeBitCast(Glibc.signal(SIGPIPE, SIG_IGN) as sighandler_t?, to: Int.self)
```
Lukasa: What's the point of an `unsafeBitCast` if we're throwing away the result?

Sources/NIO/BaseSocket.swift
```diff
@@ -12,6 +12,28 @@
...
+        globallyIgnoredSIGPIPE.store(true)
+        #else
+        assert(fd >= 0, "illegal file descriptor \(fd)")
```
Lukasa: May as well be a `precondition` tbh.

Sources/NIO/BaseSocket.swift
```diff
@@ -12,6 +12,28 @@
 //
 //===----------------------------------------------------------------------===//
 
+import NIOConcurrencyHelpers
+
+private let globallyIgnoredSIGPIPE: Atomic<Bool> = .init(value: false)
+
+func ignoreSIGPIPE(descriptor fd: CInt) throws {
```
Lukasa: Is there any reason for this function not to be namespaced into `BaseSocket`?

Sources/NIO/BaseSocket.swift
```diff
@@ -319,6 +327,7 @@ class BaseSocket: Selectable {
     init(descriptor: CInt) {
         precondition(descriptor >= 0, "invalid file descriptor")
         self.descriptor = descriptor
+        try! ignoreSIGPIPE(descriptor: descriptor)
```
Lukasa: Isn't the dance with the atomic above unnecessary? Just do it inside an implicit `dispatch_once` through means of a global or static let.

Sources/NIO/BaseStreamSocketChannel.swift
Lukasa: That...seems to be untrue? There aren't many deletions in `BaseSocketChannel`. Do you mean everything here is moved from `StreamSocketChannel`?

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,88 @@ public final class DatagramBootstrap {
...
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    @inlinable
+    public func channelOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> Self {
+        self._channelOptions.append(key: option, value: value)
+        return self
+    }
+
+    /// Create the `PipeChannel` with the provided input and output file descriptors..
```
Lukasa: Duplicate period.

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,88 @@ public final class DatagramBootstrap {
...
+    ///
+    /// - parameters:
+    ///     - inputDescriptor: The _Unix file descriptor_ for the input (ie. the read side).
+    ///     - outputDescriptor: The _Unix file descriptor_ for the input (ie. the write side).
+    /// - returns: an `EventLoopFuture<Channel>` to deliver the `Channel` immediately.
```
Lukasa: The word "immediately" here is not accurate: this future may take a while to be delivered if the initializer delays setup.

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,139 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension NIOFileHandle: Selectable {
+}
```
Lukasa: I assume we don't need this?

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,139 @@
...
+final class PipePair: SocketProtocol {
+    typealias SocketType = NIOFileHandle
+
+    let inputFD: NIOFileHandle
+    let outputFD: NIOFileHandle
+
+    init(inputFD: NIOFileHandle, outputFD: NIOFileHandle) {
+        self.inputFD = inputFD
+        self.outputFD = outputFD
+        try! inputFD.withUnsafeFileDescriptor {
```
Lukasa: I assume the plan isn't to crash if either of these things fail.

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    func sendmmsg(msgs: UnsafeMutableBufferPointer<MMsgHdr>) throws -> IOResult<Int> {
+        throw ChannelError.operationUnsupported
+    }
+
+    func shutdown(how: Shutdown) throws {
+        switch how {
+        case .RD:
+            throw ChannelError.operationUnsupported
```
Lukasa: Is this a reasonable semantic? Why not close the input FD? Pipes are nearly the only place where `shutdown(.RD)` actually has some remotely observable behaviour.

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,139 @@
...
+        case .WR:
+            try self.outputFD.close()
+        case .RDWR:
+            try self.close()
+        }
+    }
+
+    var isOpen: Bool {
+        return self.inputFD.isOpen && self.outputFD.isOpen
```
Lukasa: Is this right? Are we not open if _either_ is open? Certainly `close()` thinks we are.

Sources/NIO/SocketChannel.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-private extension ByteBuffer {
+extension ByteBuffer {
     mutating func withMutableWritePointer(body: (UnsafeMutableRawBufferPointer) throws -> IOResult<Int>) rethrows -> IOResult<Int> {
```
Lukasa: Add `private` here please, unless you are actually planning to make this more public. If you are, please add it to `BaseStreamSocketChannel.swift`.

Sources/NIO/BaseStreamSocketChannel.swift
weissi: From `SocketChannel` actually :)

Sources/NIO/BaseSocket.swift
weissi: done (to `BaseSocketProtocol`)

Sources/NIO/BaseSocket.swift
weissi: none :)

Sources/NIO/BaseSocket.swift
weissi: @Lukasa usually we do 'if it's a bug in NIO that doesn't lead to disaster' == `assert`, no? This doesn't cause any harm if `< 0`, `fcntl` will return `EBADF`.

Sources/NIO/BaseSocket.swift
weissi: no, thanks

Sources/NIO/Bootstrap.swift
weissi: ta, fixed

Sources/NIO/Bootstrap.swift
weissi: yeah, sorry, I copy & pasted that from our existing docs, will fix that.

Sources/NIO/PipePair.swift
weissi: We do. For `PipePair` to be a `SocketProtocol` implementation, it's `SocketType` must conform to `Selectable`. `PipePair'`s `SocketType` however is `NIOFileHandle`.

Sources/NIO/PipePair.swift
weissi: thanks, fixing.

Sources/NIO/PipePair.swift
weissi: fair.

Sources/NIO/PipePair.swift
weissi: oops :), thanks

Sources/NIO/SocketChannel.swift
weissi: @Lukasa we're using this from two places now (`BaseStreamSocketChannel.swift` and `SocketChannel.swift`)

Sources/NIO/BaseSocket.swift
```diff
@@ -204,7 +206,8 @@ extension sockaddr_storage {
 /// Base class for sockets.
 ///
 /// This should not be created directly but one of its sub-classes should be used, like `ServerSocket` or `Socket`.
-class BaseSocket: Selectable {
+class BaseSocket: Selectable & BaseSocketProtocol {
```
Lukasa: I don't believe in `&` for protocol conformances.

Sources/NIO/BaseSocket.swift
```diff
@@ -319,6 +308,7 @@ class BaseSocket: Selectable {
     init(descriptor: CInt) {
         precondition(descriptor >= 0, "invalid file descriptor")
         self.descriptor = descriptor
+        try! self.ignoreSIGPIPE(descriptor: descriptor)
```
Lukasa: Please add a comment explaining why we force-unwrap here.

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,151 @@
...
+                    atLeastOneSuccess = true
+                }
+            } catch let err as IOError where err.errnoCode == EBADF {
+                // this is what `withUnsafeFileDescriptor` throws if the file handle is closed.
+                error = err
+            } catch {
+                throw error
+            }
+        }
+        guard atLeastOneSuccess else {
```
Lukasa: Why do we tolerate one failure?

Sources/NIO/BaseSocket.swift
weissi: @Lukasa wdyt?

Lukasa: Happy to ignore this then.

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,114 @@ public final class DatagramBootstrap {
...
+    @inlinable
+    public func channelOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> Self {
+        self._channelOptions.append(key: option, value: value)
+        return self
+    }
+
+    private func validateFileDescriptorIsNotAFile(_ descriptor: CInt) throws {
+        var s: stat = .init()
+        try withUnsafeMutablePointer(to: &s) { ptr in
+            try Posix.fstat(descriptor: descriptor, outStat: ptr)
```
weissi: @Lukasa this is currently run on the EventLoop. I think I might actually do what you suggested and for now just assert that it's not run on the EventLoop. We can always improve later by allowing that by offloading it.

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,114 @@ public final class DatagramBootstrap {
...
+        }
+        if (s.st_mode & S_IFREG) != 0 {
```
weissi: this should also check for `S_IFDIR`

weissi:
ping @Lukasa

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,116 @@ public final class DatagramBootstrap {
...
+///                       }
+///                       .withPipes(inputDescriptor: STDIN_FILENO, outputDescriptor: STDOUT_FILENO)
+///
+public final class NIOPipeBootstrap {
+    private let group: EventLoopGroup
+    private var channelInitializer: ((Channel) -> EventLoopFuture<Void>)?
+    @usableFromInline
+    internal var _channelOptions = ChannelOptions.Storage()
+
+    /// Create a `PipeBootstrap` on the `EventLoopGroup` `group`.
```
glbrntt:
```suggestion
    /// Create a `NIOPipeBootstrap` on the `EventLoopGroup` `group`.
```

glbrntt: And elsewhere.

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,116 @@ public final class DatagramBootstrap {
...
+
+    /// Initialize the connected `PipeChannel` with `initializer`. The most common task in initializer is to add
+    /// `ChannelHandler`s to the `ChannelPipeline`.
+    ///
+    /// The connected `Channel` will operate on `ByteBuffer` as inbound and outbound messages. Please note that
+    /// `IOData.fileRegion` is _not_ supported for `PipeChannel`s because `sendfile` only works on sockets.
+    ///
+    /// - parameters:
+    ///     - handler: A closure that initializes the provided `Channel`.
+    public func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> Self {
```
glbrntt: Isn't `Self` Swift 5.1?

Sources/NIO/Bootstrap.swift
```diff
@@ -705,3 +705,116 @@ public final class DatagramBootstrap {
...
+        if (s.st_mode & S_IFREG) != 0 || (s.st_mode & S_IFDIR) != 0 {
+            throw ChannelError.operationUnsupported
+        }
+    }
+
+    /// Create the `PipeChannel` with the provided input and output file descriptors.
+    ///
+    /// - parameters:
+    ///     - inputDescriptor: The _Unix file descriptor_ for the input (ie. the read side).
+    ///     - outputDescriptor: The _Unix file descriptor_ for the input (ie. the write side).
```
glbrntt:
```suggestion
    ///     - outputDescriptor: The _Unix file descriptor_ for the output (ie. the write side).
```

Sources/NIO/PipePair.swift
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+    }
+
+    var description: String {
+        return "PipePair { in=\(inputFD), out=\(outputFD) }"
+    }
+
+    func connect(to address: SocketAddress) throws -> Bool {
+        return true
```
glbrntt: Why isn't this a `ChannelError.operationUnsupported`?

weissi: thanks, no reason I think, fixing.

Sources/NIO/Bootstrap.swift
weissi: for return types, it was there for a long time

Sources/NIO/Bootstrap.swift
weissi: thansk!

weissi:

PR #1142, opened 2019-09-20
fix thread-safety of BaseSocketChannel's description

### LINKED ISSUE #1141: Printing a Channel from outside the EventLoop is not thread safe
### Expected behavior
Printing a Channel should be thread safe

### Actual behavior
TSAN complains about concurrent access to `BaseSocket`

```
#0 BaseSocket.description.getter BaseSocket.swift:434 (xxx)
    #1 protocol witness for CustomStringConvertible.description.getter in conformance BaseSocket <compiler-generated> (xxx)
    #2 DefaultStringInterpolation.appendInterpolation<A>(_:) <null>:16073552 (libswiftCore.dylib:x86_64+0x3c989)
    #3 protocol witness for CustomStringConvertible.description.getter in conformance SocketChannel <compiler-generated> (xxx)
    #4 _print_unlocked<A, B>(_:_:) <null>:16073552 (libswiftCore.dylib:x86_64+0x955e2)
   ...
```

### Steps to reproduce

1. Create a `SocketChannel`
2. Print it from outside the EL
3. Close the channel

weissi:
Motivation:

BaseSocketChannel's description method accessed the (mutable) socket
property of the channel which must only be accessed on the EventLoop.
Description however call be called on any thread.

Modifications:

Don't access the socket from description.

Result:

- more thread-safety
- fixes #1141

### DIFF
--- a/Tests/NIOTests/TestUtils.swift
+++ b/Tests/NIOTests/TestUtils.swift
@@ -14,6 +14,7 @@
 
 import XCTest
 @testable import NIO
+import NIOConcurrencyHelpers
 
 func withPipe(_ body: (NIO.NIOFileHandle, NIO.NIOFileHandle) -> [NIO.NIOFileHandle]) throws {
     var fds: [Int32] = [-1, -1]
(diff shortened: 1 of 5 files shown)


Tests/NIOTests/TestUtils.swift
```diff
@@ -294,3 +295,76 @@ func assertSuccess<Value>(_ result: Result<Value, Error>, file: StaticString = #
...
+        lock.withLock {
+            eventLoopBag.append(group)
+        }
+        return group.next()
+    }
+
+    // TCP
+    let tcpChildChannel = freshEventLoop().makePromise(of: Channel.self)
+    let tcpAcceptedChannel = freshEventLoop().makePromise(of: Channel.self)
+    let tcpServerChannel = try assertNoThrowWithValue(ServerBootstrap(group: freshEventLoop())
```
Lukasa: The server channel and the accepted channel here are identical, no?

weissi: No, the server channel is the actual `ServerSocketChannel` (the one we call `accept` on) and the accepted channel is the `SocketChannel` (the one we accepted). Type-wise, the _child_ channel and the _accepted_ channel are the same (`SocketChannel`) so we could totally reduce to only one.

Lukasa: `tcpAcceptedChannel` is fulfilled in the `serverChannelInitializer`, and `tcpServerChannel` is the result of the call to `bind()`. I’m pretty sure that if you did an object identity comparison here they’d be identical.

Put another way, I think `tcpAcceptedChannel` is actually of type `ServerSocketChannel`. I think the promise naming here has led to confusion. You can observe the result of this confusion at the bottom of this file, where your loop code is over `tcpServerChannel`, `tcpChildChannel`, `tcpAcceptedChannel`, and `udpChannel`. This does not pass the `tcpClientChannel` at all, and if it did there would be 5 channels passed even though there are only 4 unique sources of channel. I think you can simply drop the `tcpAcceptedChannel` altogether.

Tests/NIOTests/TestUtils.swift
```diff
@@ -294,3 +295,76 @@ func assertSuccess<Value>(_ result: Result<Value, Error>, file: StaticString = #
...
+        })
+    }
+
+    func freshEventLoop() -> EventLoop {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        lock.withLock {
+            eventLoopBag.append(group)
+        }
+        return group.next()
+    }
```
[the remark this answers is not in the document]
weissi: I'll remove em

Tests/NIOTests/TestUtils.swift
```diff
@@ -294,3 +295,76 @@ func assertSuccess<Value>(_ result: Result<Value, Error>, file: StaticString = #
...
+            defer {
+                tcpChildChannel.succeed(channel)
+            }
+            return channel.eventLoop.makeSucceededFuture(())
+        }
+        .serverChannelInitializer { channel in
+            defer {
+                tcpAcceptedChannel.succeed(channel)
+            }
+            return channel.eventLoop.makeSucceededFuture(())
```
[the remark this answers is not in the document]
weissi: done

Tests/NIOTests/TestUtils.swift
weissi: you're right, thanks!

PR #1151, opened 2019-09-30
assert EventLoopGroup::syncShutdownGracefully is not called on the event loop

tomerd:
motivation: calling syncShutdownGracefully on the event loop can lead to deadlock

changes: assert before calling syncShutdownGracefully that no eventloop context exists

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ Package.pins
 Package.resolved
 .podspecs
 DerivedData
+.swiftpm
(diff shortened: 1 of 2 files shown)


.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ianpartridge: Probably `.swiftpm` needs to go in the gitignore.

tomerd:
added .swiftpm to gitignore, rest is till there

PR #1152, opened 2019-10-01
NIOHTTP1TestServer

mariosangiorgio:
Added `NIOHTTP1TestServer`, which serializes requests making it easier to write unit tests where the clients connect to a "real" server.

### Motivation:

I am working on a client where a logical request consists of multiple HTTP requests, some of which might happen in parallel. The client uses [`AsyncHTTPClient`](https://github.com/swift-server/async-http-client), which doesn't expose the NIO `Channel` it uses. This makes it hard to write unit tests as I normally would.

### Modifications:

I added `NIOHTTP1TestServer` and some tests that cover its behaviour. The test server exposes the same API we would use while interacting with a `Channel`, making the same testing approach viable also for more complex clients.

### Result:

After my change developers writing client based on complex HTTP1 requests will have the facility they need to write unit tests following the same patters they would use writing unit tests for a NIO `Channel`.


Lukasa:

Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -0,0 +1,318 @@
...
+        case waitingForChannel(EventLoopPromise<Void>)
+        case idle
+        case stopped
+    }
+    private var state: State = .idle
+
+    func handleChannels() {
+        self.eventLoop.assertInEventLoop()
+
+        let channel: Channel!
```
Lukasa: This doesn’t need to be an IUO, there must be a `Channel` if we get past this section.

Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -0,0 +1,318 @@
...
+            return
+        case .waitingForChannel:
+            preconditionFailure("illegal state \(self.state)")
+        case .stopped:
+            return
+        }
+
+        assert(self.currentClientChannel == nil)
+        self.currentClientChannel = channel
+        channel.closeFuture.whenSuccess {
```
Lukasa: Do we need to consider what happens here if the connection is closed by the remote peer while we’re waiting to handle a previous connection? In this case I think we’d just lose the connection as this promise will fire immediately.

Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -0,0 +1,318 @@
...
+        .wait()
+    }
+}
+
+// MARK: - Public API for test driver
+extension NIOHTTP1TestServer {
+    struct NonEmptyInboundBufferOnStop: Error {}
+
+    public func stop() throws {
+        assert(!self.eventLoop.inEventLoop)
```
Lukasa: Why does this have to be called off the loop?

Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -0,0 +1,318 @@
...
+                channels.forEach {
+                    $0.close(promise: nil)
+                }
+            case .waitingForChannel(let promise):
+                self.state = .stopped
+                promise.fail(ChannelError.ioOnClosedChannel)
+            case .idle:
+                self.state = .stopped
+            case .stopped:
+                preconditionFailure("double stopped NIOHTTP1TestServer")
```
Lukasa: Should this crash? It seems unnecessarily forceful.

Tests/NIOTestUtilsTests/NIOHTTP1TestServerTest.swift
```diff
@@ -0,0 +1,343 @@
...
+        // Setup the test environment.
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let allocator = ByteBufferAllocator()
+        let testServer = NIOHTTP1TestServer(group: group)
+        defer {
+            XCTAssertNoThrow(try testServer.stop())
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        // -- SNIP --
```
Lukasa: What is this comment for?

weissi: @Lukasa everything below the line serves as the example that's copied into the docs :). Just to have it exeutable.

mariosangiorgio: That was the intent, but on a second thought if we move `sendRequestTo` definition outside this function we won't need to snip anything.

PR #1153, opened 2019-10-01
ByteBuffer: add direct Codable support

weissi:
Motivation:

So far, it has been harder than necessary to use Codable & ByteBuffer.
These new APIs should simplify that and allow future optimisations.

Modifications:

Add new API to use `JSONEncoder` and `JSONDecoder` directly with
`ByteBuffer`.

Result:

Easier Codable + ByteBuffer usage.


weissi:

tomerd:
NIOFoundationCompatTests source missing?

weissi:

Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
```diff
@@ -0,0 +1,59 @@
...
+    ///
+    /// If the `ByteBuffer` does not contain valid JSON, this method throws the
+    /// `DecodingError.dataCorrupted(_:)` error. If a value within the JSON
+    /// fails to decode, this method throws the corresponding error.
+    ///
+    /// - parameters:
+    ///     - type: The type of the value to decode from the supplied JSON object.
+    ///     - buffer: The `ByteBuffer` that contains JSON object to decode.
+    /// - returns: The decoded object.
+    public func decode<T: Decodable>(_ type: T.Type, from buffer: ByteBuffer) throws -> T {
```
glbrntt: Since `buffer` isn't `inout` the bytes in the callers buffer won't be read: we should call this out in the documentation.

Maybe we can add another API to `ByteBuffer` to read a `Decodable` which _does_ consume the bytes: this would  be more inline with the other `read` APIs.

weissi: great idea, done.

Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
```diff
@@ -0,0 +1,136 @@
...
+extension ByteBuffer {
+    /// Attempts to decode the `length` bytes from `index` using the `JSONDecoder` `decoder` as `T`.
+    ///
+    /// - parameters:
+    ///    - type: The type type that is attempted to be decoded.
+    ///    - decoder: The `JSONDecoder` that is used for the decoding.
+    ///    - index: The index of the first byte to decode.
+    ///    - length: The number of bytes to decode.
+    /// - returns: The decoded value is successful or `nil` if there are not enough readable bytes available.
+
```
glbrntt:
```suggestion

```

Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
```diff
@@ -0,0 +1,136 @@
...
+extension ByteBuffer {
+    /// Attempts to decode the `length` bytes from `index` using the `JSONDecoder` `decoder` as `T`.
+    ///
+    /// - parameters:
+    ///    - type: The type type that is attempted to be decoded.
+    ///    - decoder: The `JSONDecoder` that is used for the decoding.
+    ///    - index: The index of the first byte to decode.
+    ///    - length: The number of bytes to decode.
+    /// - returns: The decoded value is successful or `nil` if there are not enough readable bytes available.
```
glbrntt:
```suggestion
    /// - returns: The decoded value if successful or `nil` if there are not enough readable bytes available.
```

Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    @inlinable
+    public func getJSONDecodable<T: Decodable>(_ type: T.Type,
+                                               decoder: JSONDecoder = JSONDecoder(),
+                                               at index: Int, length: Int) throws -> T? {
```
glbrntt: Can we provide a default `type` here and shift it to the last argument? That'd be more inline with the equivalent integer methods.

```swift
public func getJSONDecodable<T: Decodable>(at index: Int, length: Int, using decoder: JSONDecoder(), as type: T.Type = T.self)
```

Tests/NIOFoundationCompatTests/Codable+ByteBufferTest.swift
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    func testReadJSONDecodableFromBufferFailsBecauseShort() {
+        let beginIndex = self.buffer.writerIndex
+        self.buffer.writeString(#"{"string": "hello", "int": 42}"#)
+        let endIndex = self.buffer.writerIndex
+
+        XCTAssertThrowsError(try self.buffer.readJSONDecodable(StringAndInt.self,
+                                                               length: endIndex - beginIndex - 1)) { error in
+
```
glbrntt:
```suggestion

```

Tests/NIOFoundationCompatTests/Codable+ByteBufferTest.swift
```diff
@@ -0,0 +1,165 @@
...
+                                                                XCTAssert(error is DecodingError)
```
glbrntt: Is this meant to be super indented?

Tests/NIOFoundationCompatTests/Codable+ByteBufferTest.swift
```diff
@@ -0,0 +1,165 @@
...
+        for _ in 0..<10 {
+            XCTAssertNoThrow(try self.buffer.writeJSONEncodable(expectedSandI, encoder: JSONEncoder()))
+        }
+        for _ in 0..<11 {
+            XCTAssertNoThrow(try self.buffer.readJSONDecodable(StringAndInt.self, length: writtenBytes ?? -1))
+        }
+        XCTAssertEqual(0, self.buffer.readableBytes)
+    }
+
+    func testGetSetJSONDecodableWOrks() {
```
glbrntt:
```suggestion
    func testGetSetJSONDecodableWorks() {
```

Tests/NIOFoundationCompatTests/Codable+ByteBufferTest+XCTest.swift
```diff
@@ -0,0 +1,42 @@
...
+      return [
+                ("testSimpleDecode", testSimpleDecode),
+                ("testSimpleEncodeIntoBuffer", testSimpleEncodeIntoBuffer),
+                ("testSimpleEncodeToFreshByteBuffer", testSimpleEncodeToFreshByteBuffer),
+                ("testGetJSONDecodableFromBufferWorks", testGetJSONDecodableFromBufferWorks),
+                ("testGetJSONDecodableFromBufferFailsBecauseShort", testGetJSONDecodableFromBufferFailsBecauseShort),
+                ("testReadJSONDecodableFromBufferWorks", testReadJSONDecodableFromBufferWorks),
+                ("testReadJSONDecodableFromBufferFailsBecauseShort", testReadJSONDecodableFromBufferFailsBecauseShort),
+                ("testReadWriteJSONDecodableWorks", testReadWriteJSONDecodableWorks),
+                ("testGetSetJSONDecodableWOrks", testGetSetJSONDecodableWOrks),
```
glbrntt:
```suggestion
                ("testGetSetJSONDecodableWorks", testGetSetJSONDecodableWorks),
```

Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
weissi: a couple of thoughts here:

- `JSONDecoder` _requires_ you to explicitly state the type and I think that's good because for security reasons, type inference might introduce a very subtle bug if it were to infer a different type
- if we accept that the type should be a required argument, then I believe Swift's naming rules want it to be the first parameter. Because this reads well in English:

> Read a JSON decodable _Sandwich_ at index 7 with length 23.

which translates to swift as `readJSONDecodable(Sandwich.self, at: 7, length: 23)`.

The reverse

`readJSONDecodable(at: 7, length: 23, type: Sandwich.self)`

would read as

> read JSON decodable at index 7 with length 23, type Sandwich

glbrntt: > * `JSONDecoder` _requires_ you to explicitly state the type and I think that's good because for security reasons, type inference might introduce a very subtle bug if it were to infer a different type

Ah -- good point!

> * if we accept that the type should be a required argument, then I believe Swift's naming rules want it to be the first parameter. Because this reads well in English:
> 
> > Read a JSON decodable _Sandwich_ at index 7 with length 23.
> 
> which translates to swift as `readJSONDecodable(Sandwich.self, at: 7, length: 23)`.
> 
> The reverse
> 
> `readJSONDecodable(at: 7, length: 23, type: Sandwich.self)`
> 
> would read as
> 
> > read JSON decodable at index 7 with length 23, type Sandwich

Makes sense!

PR #1157, opened 2019-10-05
fix a number of memory leaks in the test suite

weissi:
Motivation:

Memory leaks are bad, even in test suites.

Modifications:

Fix a number of memory leaks.

Result:

Fewer leaks.

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -273,6 +273,9 @@ class EmbeddedChannelCore: ChannelCore {
 /// `EmbeddedChannel` is in unit tests when you want to feed the inbound events
 /// and check the outbound events manually.
 ///
+/// Please remember to call `finish()` when you are no longer using this
+/// `EmbeddedChannel`.
+///
 /// To feed events through an `EmbeddedChannel`'s `ChannelPipeline` use
 /// `EmbeddedChannel.writeInbound` which accepts data of any type. It will then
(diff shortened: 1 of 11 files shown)


Sources/NIO/Embedded.swift
```diff
@@ -391,15 +394,22 @@ public final class EmbeddedChannel: Channel {
...
     ///
+    /// - parameters:
+    ///     - acceptAlreadyClosed: Whether `finish` should throw if the `EmbeddedChannel` has been previously `close`d.
     /// - returns: The `LeftOverState` of the `EmbeddedChannel`. If all the inbound and outbound events have been
     ///            consumed (using `readInbound` / `readOutbound`) and there are no pending outbound events (unflushed
     ///            writes) this will be `.clean`. If there are any unconsumed inbound, outbound, or pending outbound
     ///            events, the `EmbeddedChannel` will returns those as `.leftOvers(inbound:outbound:pendingOutbound:)`.
-    public func finish() throws -> LeftOverState {
-        try close().wait()
+    public func finish(acceptAlreadyClosed: Bool) throws -> LeftOverState {
```
Lukasa: This flag doesn’t actually do anything...

weissi: I think I fixed this already

Lukasa:
Looks like a bunch of tests are now failing.

PR #1159, opened 2019-10-10
Add some websocket encode benchmarks.

Lukasa:
Motivation:

To improve the performance of a given component it is very helpful to
know how it performs! This patch adds benchmarks and allocation counter
tests to NIO for websocket frame encoding.

Modifications:

- Wrote some benchmarks for WebSocket frame encoding.
- Wrote some allocation counter tests for WebSocket frame encoding.

Result:

Better insight

### DIFF
--- a/Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
+++ b/Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
@@ -1,0 +1,119 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import NIOWebSocket
+
+final class WebSocketFrameEncoderBenchmark {
+    private let channel: EmbeddedChannel
+    private let dataSize: Int
+    private let data: ByteBuffer
+    private let runCount: Int
+    private let dataStrategy: DataStrategy
+    private let cowStrategy: CoWStrategy
+    private var frame: WebSocketFrame?
+
+    init(dataSize: Int, runCount: Int, dataStrategy: DataStrategy, cowStrategy: CoWStrategy) {
+        self.channel = EmbeddedChannel()
+        self.dataSize = dataSize
+        self.runCount = runCount
+        self.dataStrategy = dataStrategy
+        self.cowStrategy = cowStrategy
+        self.data = ByteBufferAllocator().buffer(size: dataSize, dataStrategy: dataStrategy)
+    }
+}
+
+
+extension WebSocketFrameEncoderBenchmark {
+    enum DataStrategy {
+        case spaceAtFront
+        case noSpaceAtFront
+    }
+}
+
+
+extension WebSocketFrameEncoderBenchmark {
+    enum CoWStrategy {
+        case always
+        case never
+    }
+}
+
+
+extension WebSocketFrameEncoderBenchmark: Benchmark {
+    func setUp() throws {
+        // We want the pipeline walk to have some cost.
+        for _ in 0..<3 {
+            try! self.channel.pipeline.addHandler(NoOpOutboundHandler()).wait()
+        }
+        try! self.channel.pipeline.addHandler(WebSocketFrameEncoder()).wait()
+        self.frame = WebSocketFrame(opcode: .binary, data: self.data, extensionData: nil)
+    }
+
+    func tearDown() {
+        _ = try! self.channel.finish()
+    }
+
+    func run() throws -> Int {
+        switch self.cowStrategy {
+        case .always:
+            let frame = self.frame!
+            return self.runWithCoWs(frame: frame)
+        case .never:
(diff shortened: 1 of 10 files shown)


Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
```diff
@@ -0,0 +1,119 @@
...
+    enum CoWStrategy {
+        case always
+        case never
+    }
+}
+
+
+extension WebSocketFrameEncoderBenchmark: Benchmark {
+    func setUp() throws {
+        // We want the pipeline walk to have some cost.
```
glbrntt: I assume this is to make it closer to a "real" use case?

Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
```diff
@@ -0,0 +1,119 @@
...
+        case .never:
+            return self.runWithoutCoWs()
+        }
+    }
+
+    private func runWithCoWs(frame: WebSocketFrame) -> Int {
+        for _ in 0..<self.runCount {
+            self.channel.write(frame, promise: nil)
+        }
+        return 1
```
glbrntt: What are we returning here..?

docker/docker-compose.1804.50.yaml
```diff
@@ -26,6 +26,10 @@ services:
       - MAX_ALLOCS_ALLOWED_scheduling_10000_executions=20150
       - MAX_ALLOCS_ALLOWED_creating_10000_headers=10100
       - MAX_ALLOCS_ALLOWED_modifying_1000_circular_buffer_elements=50
+      - MAX_ALLOCS_ALLOWED_encode_1000_ws_frames_holding_buffer=4010
+      - MAX_ALLOCS_ALLOWED_encode_1000_ws_frames_holding_buffer_with_space=4010
+      - MAX_ALLOCS_ALLOWED_encode_1000_ws_frames_new_buffer=6010
+      - MAX_ALLOCS_ALLOWED_encode_1000_ws_frames_new_buffer_with_space=6010
```
glbrntt: How do you come up with these value, do you just run once and figure out whether the number is reasonable/expected i.e. X allocs per whatever-is-being measured?

Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
weissi: exactly

Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
weissi: For benchmarks in general, it's very important that the optimiser doesn't see through it and decides to not to anything at all. There are many options (for example a `blackhole` function) to stop the optimiser seeing through this being "unnecessary" work. In NIO's perf tests we usually use an `Int` return and the only thing you need to do is to return some integer that depends on the calculation you did.

For example if you received a number of bytes, you could just return `buffer.readableBytes`. In this case, Cory returns `1` here to return that through the `run` function. Arguably, `1` doesn't really depend on the work that's being done here, so we could remove this. The reason this will work is that there's no way the compiler sees through `self.channel.write(...)` doing nothing (it does at least write into the `EmbeddedChannel`'s `outboundBuffer` so it does have an effect.)

Long story short (@Lukasa correct me if I'm wrong): In _this_ case, it's unnecessary but generally returning `Int`s is what we do, usually they should reflect some (it doesn't matter which) property of something you calculate in the work you want to benchmark. In this benchmark, there isn't really anything we can do synchronously so Cory uses `1` :)

docker/docker-compose.1804.50.yaml
weissi: need to run it (on Linux) and seed with the initial numbers with what you see. Typically we add between 10 and 50 to the result to cover for any slack. And typically we run something at least 1000 times, so anything below 1000 is definitely slack. But these things are usually super stable, so 10 slack is usually good enough. I don't think we've ever seen a failure because of slack.

weissi: So the idea is: Let's seed it with something we've seen. And from then on we know when we regress (and also if we improve).

Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
Lukasa: Yeah: the question this benchmark is going to try to answer is how useful it is to avoid an extra pipeline walk, so we need to have the pipeline walk have some actual overhead.

Sources/NIOPerformanceTester/WebSocketFrameEncoderBenchmark.swift
Lukasa: Nothing really. Just something to avoid the compiler getting too sassy.

docker/docker-compose.1804.50.yaml
Lukasa: Yeah, the system is to first run them and see what the numbers are. If they're insane (they were the first time), find where the extra allocs have come from and kill them. Then, push to CI, let the CI tell you what the answers it got were, and put those in with a buffer of four or five allocations to allow for different Swift versions to impose new costs at startup.

Lukasa:

Lukasa:
Hmm, tests failing without a clear indication of why. It's definitely in the allocation counter tests though. Going to rerun in case it was weird noise.

Lukasa:
@weissi Can I get you to re-review my change to the integration test bash files?

PR #1161, opened 2019-10-11
Reduce allocations in WebSocketFrameEncoder.

Lukasa:
Motivation:

The WebSocketFrameEncoder naively allocated a new block to write the
frame header into every time it wrote. This is excessive: in many cases
it would be able to re-use the same buffer as last time.

Modifications:

- Attempt to re-use the buffer we used for the last header.

Result:

Fewer allocations in some applications.

### DIFF
--- a/Sources/NIOWebSocket/WebSocketFrameEncoder.swift
+++ b/Sources/NIOWebSocket/WebSocketFrameEncoder.swift
@@ -39,40 +39,49 @@ public final class WebSocketFrameEncoder: ChannelOutboundHandler {
     public typealias OutboundIn = WebSocketFrame
     public typealias OutboundOut = ByteBuffer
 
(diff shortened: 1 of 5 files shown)


Lukasa:

Lukasa:
@tomerd What's up with the performance bot? It's missing a bunch of test cases.

Lukasa:
Marking this do not merge until we know why the performance bot isn't giving us all the test cases.

tomerd:

tomerd:

tomerd:
@Lukasa as discussed `future_whenallsucceed_100k_immediately_succeeded_on_loop` is crashing producing partial results. updated the job to fail when that happens and lets investigate why its crashing

Lukasa:
The issue in the performance bot is fixed by #1164.

Lukasa:

Lukasa:
r4r @weissi

PR #1164, opened 2019-10-14
Save the performance tests from the OOM killer

Lukasa:
Motivation:

A dark force walks abroad our lands. In a place where well-behaved
processes close their blinds, lock their doors, and try not to make
noise late at night, processes that shine just a little too brightly,
consume just a little too much RAM, attract the attention of a dark
force whose name is spoken only in whispers. Some call this force a
myth, but others say that the deniers of this force have whispered dark
incantations, invoking the power word "oom_adj".

It is time we intervened to keep the law-abiding processes in our CI
systems safe.

Modifications:

- Prevent the websocket performance tests comnsuming 4GB of RAM.

Result:

Peace will reign.


Lukasa:
The benchmark uses `EmbeddedChannel`, which keeps hold of writes until they are consumed from the channel. This meant that each frame was resident in memory for the entire duration of the test, which is less than ideal. The new handler drops all writes it receives, so it allows the memory to be freed.

tomerd:

tomerd:
@Lukasa looks good now!

PR #1165, opened 2019-10-15
Bind Before Connect functionality for a SocketChannel

adamhoracek:
Bind Before Connect functionality for a SocketChannel

### Motivation:

The current implementation lacks the ability of binding the TCP socket before connecting to a remote endpoint, thus we cannot chose which port (e.g.) we should initiate the connection from.

### Modifications:

* Added a `BindToSocketAddressBeforeConnectOption` to ChannelOptions with a `SocketAddress` value
* Changed the mechanism of `connectSocket()` in `SocketChannel` so it binds first

### Result:

After these changes we can observe that newly created SocketChannel with the specified option are initiated from a given port.

### DIFF
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -93,6 +93,13 @@ public struct BacklogOption: ChannelOption {
     public init() {}
 }
 
+/// `BindToSocketAddressBeforeConnectOption` specifies local address to be bound before connecting to remote.
+public struct BindToSocketAddressBeforeConnectOption: ChannelOption {
+    public typealias Value = SocketAddress
+
+    public init() {}
+}
+
 /// `DatagramVectorReadMessageCountOption` allows users to configure the number of messages to attempt to read in a single syscall on a
 /// datagram `Channel`.
 ///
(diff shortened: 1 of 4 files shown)


Lukasa:
Thanks for this patch!

I don't think this design is necessarily the right way to go though. Specifically, the `ChannelPipeline` has a `bind` event that we should probably be using here. I think the better mechanism would be to add `bind` to the `ClientBootstrap` object which will return a `Channel` that can then have `connect` called on it. @weissi, how do you feel about this?

Also /cc @normanmaurer to confirm what Netty's design here is.

weissi:
@Lukasa I'd suggest to do what I did in #1192 which allows `bind` to be used in a `channelInitializer` for a `SocketChannel`. Then we can just use

```
ClientBootstrap(...)
    channel.bind(...)
}
.connect(...)
```

WDYT?

Lukasa:
@weissi I think your approach is reasonable, but it interacts poorly with happy eyeballs because the happy eyeballs logic does not expect the socket to already have a local address. That means it'll repeatedly set up and attempt to connect on channels that are bound to the wrong address family, leading to the channel setup being inevitably wasted. If the caller doesn't set SO_REUSEADDR it'll be even worse, because we'll inevitably hit the reuse problem.

We can come to the following decisions:

1. We don't care.
2. We care, and should enhance the happy eyeballs connector to know this, which requires telling the bootstrap.
3. We care, and think binding and happy eyeballs don't play well together, and so have the bootstrap elide HE altogether when connecting.

I think any of the above choices are reasonable. We can also do (2) in stages.

weissi:
@Lukasa that is very true. I have to think about which of 1, 2, or 3 I think is best :)

Lukasa:
Superseded by #1490, I believe.

PR #1166, opened 2019-10-15
DatagramChannel in Connected Mode

adamhoracek:
DatagramChannel in Connected Mode

### Motivation:

Current implementation has no support for datagram sockets in connected mode. In order to fix that, here is an proposition of an implementation.

### Modifications:

* Added a `ConnectToSocketAddressAfterBindOption` to `ChannelOptions`
* Added a “cache” (`self.updateCachedAddressesFromSocket`) to store already connected channels
* Changed the implementation of `bind0()` in `DatagramChannel` so it connects the socket after a successful bind

### Result:

After these changes we are able to connect `DatagramChannel`s to a remote address after they bind to a local address.

### DIFF
--- a/Sources/NIO/Channel.swift
+++ b/Sources/NIO/Channel.swift
@@ -345,6 +345,9 @@ public enum ChannelError: Error {
 
     /// An attempt was made to remove a ChannelHandler that is not removable.
     case unremovableHandler
+
+    /// An attempt to connect datagram socket after it was bound did not succeed.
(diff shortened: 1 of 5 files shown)


Lukasa:
Norman proposed accepting both in connected mode, which is what Netty does.

weissi:
@Lukasa fine by me

weissi:
@adamhoracek friendly ping, are you still interested in pushing this over the line?

PR #1167, opened 2019-10-15
ServerDatagramChannel implementation (also in a Connected Mode)

adamhoracek:
ServerDatagramChannel implementation (also in a Connected Mode)

### Motivation:

An implementation of a similar mechanism to a `ServerSocketChannel` but for datagram sockets has not been implemented yet. 

This implementation creates a new `DatagramChannel` (and a socket) for each first message which comes from the same source address, so kernel can delegate all future messages to the given socket (`DatagramChannel`).

If connected mode is not specified to be used, it only creates a number of `DatagramChannel`s to distribute the load across multiple threads.

### Modifications:

* Added `ServerDatagramBootstrap` which is an updated copy of `ServerBootstrap`
* Added a `ShouldConnectAfterBindOption` to a `ChannelOptions`
* Updated `EventLoop` and `Selector` so it handles the new type of a `Channel` (`ServerDatagramChannel`)
* Updated `DatagramChannel` so it uses two promises for notifying when the channel has been initiated and activated - for queueing and ordering purposes
* Added a `ServerDatagramChannel` which a mixture of an existing `ServerSocketChannel` and `DatagramChannel` + a mechanism of creating new socket and `DatagramChannel` and optionally connecting it

### Result:

After implementing these changes we can observe, that the `ServerDatagramBootstrap` and `ServerDatagramChannel` behave in the similar way as their TCP counterparts.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -705,3 +705,300 @@ public final class DatagramBootstrap {
         }
     }
 }
+
+/// A `ServerDatagramBootstrap` is an easy way to bootstrap a `ServerDatagramChannel` when creating datagram network servers.
+///
+/// Example:
+///
+/// ```swift
+///     let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+///     defer {
+///         try! group.syncShutdownGracefully()
+///     }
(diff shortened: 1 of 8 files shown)


Lukasa:
I believe this is now covered with the connected mode on datagram channels.

PR #1168, opened 2019-10-16
Attempt to write WS header in user buffer.

Lukasa:
Motivation:

In some cases it may be possible to write the (usually small) web socket
frame header to the buffer provided by the user. If we do this we can
avoid an extra pipeline traversal for the small write. This is only
going to be a performance win in cases where we can avoid a
copy-on-write operation on the buffer, but if we can then it's a useful
small win to achieve.

Modifications:

- Refactored the WebSocketFrameEncoder to potentially prepend the frame
    header.
- Added a missing @inlinable attribute.
- Tests.

Result:

Potentially improved performance.


Lukasa:

Lukasa:
Yup, so this regresses allocations in some of the allocation counter tests because it adds some CoWs. Based on the performance results above the cost of the CoW is very similar to the cost of a shortish pipeline walk, so we probably only want to do this if we avoid a CoW. Sadly we don't have tools for that right now.

@weissi do we think it's worth adding an API for asking whether a BB modification might CoW?

weissi:
@Lukasa I think it's worth to add that API. 

```
mutating func canBeWrittenToSwiftly() -> Bool { ... }
```

? :P

Lukasa:
@weissi Do you think it's worth adding that API irregardless of this patch? If so, we should consider it a separate-but-gating patch here.

weissi:
@Lukasa I think that'd be a useful API anyway, I'm happy with an extra PR or in this PR, up to you.

Lukasa:
If we think it's a useful API anyway we should have a separate patch.

Lukasa:

Lukasa:
This is ready for review.

PR #1169, opened 2019-10-17
Add API for whether BB operations may CoW.

Lukasa:
Motivation:

From time to time a NIO application holds a ByteBuffer that it may want
to edit. Whether it wants to do that may potentially be contingent on
whether having to do so will cause an allocation, as it may be
preferable to prefer a different operation.

Modifications:

- Added API for asking BB if it thinks it'll CoW.

Result:

People can ask the question.


Lukasa:
Closing because @douggregor would feel better if we did something else.

PR #1173, opened 2019-10-19
Add API for modifying BB without CoW

Lukasa:
Motivation:

From time to time a NIO application holds a ByteBuffer that it may want
to edit. Whether it wants to do that may potentially be contingent on
whether having to do so will cause an allocation, as it may be
preferable to prefer a different operation.

Modifications:

- Added API for modifying a ByteBuffer if it will not CoW.

Result:

People can write allocation-sensitive code on ByteBuffer.


fadi-botros:
Doesn't this have multithreading concerns ???

```swift
let byteBuffer: ByteBuffer = getByteBufferFromWhateverSource()

func fromThread1() {
     byteBuffer. modifyIfUniquelyOwned {     // 1
         // Do some writing to the buffer          // 2
     }
}

func fromThread2() {
     // Do some writing to the buffer              // 3
}
```

What if it happens in this order?
First: 1, which already checks for uniqueness
Second: 2, the first thread is doing the write, and still didn't finish it
Third: 3, the second thread doing its write, while still 2 didn't finish, this could do serious problems, at least data races (and also more serious issues could happen, like dangling pointers due to expansion of a byte buffer by one of the threads while the other is still holding the old pointer).

Lukasa:
> Doesn't this have multithreading concerns ???

No. This is the standard construction for performing modifications of copy-on-write data structures.

If your example runs in multiple threads, it is buggy. This is not often said enough, but copy-on-write data types _are not thread safe_: if you have multiple threads modifying the same data at the same time without synchronisation, you have a bug, regardless of whether that data is a value or a reference. This can be observed with all standard library data types: if you replaced `ByteBuffer` with `Array` in your above example and used `append`, the same issue exists.

To work correctly with value types from multiple threads you must _copy_ them into each thread's memory. At that stage the refcount will correctly go above 1 and the buffer will correctly be called shared.

> EDIT: Also what about weak references?? What if one object holds the `ByteBuffer` as a weak reference?

It cannot: only classes can be held `weak`, structs cannot.

> EDIT: Also I don't feel it is obvious to use `nil` as a signal for failing operation due to non-uniqueness of the `ByteBuffer`, what if a developer use `nil` as a return value for his passed closure itself??

Then the return type of the buffer will be a double optional. We could definitely use a new type, but I don't think the ergonomics greatly improve here.

PR #1174, opened 2019-10-21
Add flatSubmit method to EventLoop (#426)

2bjake:
Add flatSubmit method to EventLoop (#426)

### Motivation:

Calling EventLoop's submit method with a task that returns an EventLoopFuture results in a double future. flatSubmit provides an alternative which does not return a double future.

### Modifications:

Added a flatSubmit() default implementation that returns the result of calling flatMap { $0 } on the EventLoopFuture returned from task.

### Result:

Additive change only. flatSubmit method on EventLoop will be available for use.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -232,6 +232,13 @@ public protocol EventLoop: EventLoopGroup {
     /// - returns: `EventLoopFuture` that is notified once the task was executed.
     func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
 
+    /// Submit a given task to be executed by the `EventLoop`. Once the execution is complete the returned `EventLoopFuture` is notified.
+    ///
+    /// - parameters:
+    ///     - task: The closure that will be submitted to the `EventLoop` for execution.
+    /// - returns: `EventLoopFuture` that is notified once the task was executed.
+    func flatSubmit<T>(_ task: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T>
+
     /// Schedule a `task` that is executed by this `SelectableEventLoop` at the given time.
     @discardableResult
     func scheduleTask<T>(deadline: NIODeadline, _ task: @escaping () throws -> T) -> Scheduled<T>
@@ -463,6 +470,18 @@ extension EventLoop {
         }
 
         return promise.futureResult
+    }
+
+    /// Submit `task` to be run on this `EventLoop`.
+    ///
+    /// The returned `EventLoopFuture` will be completed when `task` has finished running. It will be identical to
+    ///  the `EventLoopFuture` returned by `task`.
+    ///
+    /// - parameters:
+    ///     - task: The synchronous task to run. As everything that runs on the `EventLoop`, it must not block.
+    /// - returns: An `EventLoopFuture` identical to the `EventLooopFuture` returned from `task`.
+    public func flatSubmit<T>(_ task: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T> {
+        return submit(task).flatMap { $0 }
     }
 
     /// Creates and returns a new `EventLoopPromise` that will be notified using this `EventLoop` as execution `NIOThread`.
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -232,13 +232,6 @@ public protocol EventLoop: EventLoopGroup {
     /// - returns: `EventLoopFuture` that is notified once the task was executed.
     func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
 
-    /// Submit a given task to be executed by the `EventLoop`. Once the execution is complete the returned `EventLoopFuture` is notified.
-    ///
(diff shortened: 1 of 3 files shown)


Lukasa:

Sources/NIO/EventLoop.swift
```diff
@@ -232,6 +232,13 @@ public protocol EventLoop: EventLoopGroup {
     /// - returns: `EventLoopFuture` that is notified once the task was executed.
     func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
 
+    /// Submit a given task to be executed by the `EventLoop`. Once the execution is complete the returned `EventLoopFuture` is notified.
+    ///
+    /// - parameters:
+    ///     - task: The closure that will be submitted to the `EventLoop` for execution.
+    /// - returns: `EventLoopFuture` that is notified once the task was executed.
+    func flatSubmit<T>(_ task: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T>
```
weissi: We don't need this as a customisation point on the protocol, you can just remove this line and provide an implementation in the extension (as you already do).

Sources/NIO/EventLoop.swift
```diff
@@ -465,6 +472,18 @@ extension EventLoop {
...
 
+    /// Submit `task` to be run on this `EventLoop`.
+    ///
+    /// The returned `EventLoopFuture` will be completed when `task` has finished running. It will be identical to
+    ///  the `EventLoopFuture` returned by `task`.
+    ///
+    /// - parameters:
+    ///     - task: The synchronous task to run. As everything that runs on the `EventLoop`, it must not block.
+    /// - returns: An `EventLoopFuture` identical to the `EventLooopFuture` returned from `task`.
+    public func flatSubmit<T>(_ task: @escaping () throws -> EventLoopFuture<T>) -> EventLoopFuture<T> {
```
weissi: the task closure returns a future so it shouldn't be throwing.

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -613,6 +613,40 @@ public final class EventLoopTest : XCTestCase {
         XCTAssertFalse(channel.isActive)
     }
 
+    public func testFlatSubmitCloseFutureNotifiedBeforeUnblock() throws {
+        class AssertHandler: ChannelInboundHandler {
```
weissi: this test case tests the new `flatSubmit` functionality but also a quite a bit of unrelated other unrelated stuff. I think we could reduce this test case to only the things that are actually needed such as:

- a test case that `flatSubmit`s a successful computation to `EmbeddedEventLoop` 
- a test case that `flatSubmit`s a failing computation to `EmbeddedEventLoop`
- a test case that submits a computation to a `MultiThreadedEventLoopGroup` that is already shut down.

as an example for the first case:

```
func testSuccessfulFlatSubmit() {
    let el = EmbeddedEventLoop()
    XCTAssertNoThrow(XCTAssertEqual(1, try el.flatSubmit {
        el.makeSucceededFuture(1)
    }.wait()))
}
```

Sources/NIO/EventLoop.swift
2bjake: Sounds good. Removed.

Sources/NIO/EventLoop.swift
2bjake: Nice. Removed throws.

Tests/NIOTests/EventLoopTest.swift
2bjake: Agreed. Thanks for the example. I was unfamiliar with the EmbeddedEventLoop, so that was helpful.

weissi:

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -928,4 +928,48 @@ public final class EventLoopTest : XCTestCase {
...
+            XCTAssertEqual(.failed, error as? TestError)
+        }
+    }
+
+    func testFlatSubmitOnShutdownLoopGroup() {
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        let eventLoop = eventLoopGroup.next()
+        XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
+
+        let _ = eventLoop.flatSubmit { () -> EventLoopFuture<Void> in
```
weissi: shame that our `submit` API doesn't return a failed future if the ELG is shut down but that's definitely not this PR's responsibility.

2bjake: I had to experiment to determine what the expected behavior was. On my machine, the tests would hang if I called ```wait``` after submitting a task to a shutdown group.

However, I noticed that the tests seem to be failing intermittently on the build server (one failed twice, one failed only once). So it seems that the submitted tasks are somehow running after the group is shutdown? I haven't seen that behavior on my machine.

weissi: --> 1177

2bjake: Ok, sounds good.

weissi:

Lukasa:

Issue #1105, opened 2019-08-12
Enhance writeWithUnsafeMutableBytes


Lukasa:
In general, we don't want people using the unsafe APIs on `ByteBuffer`, but in the cases where it really is the best solution we should ensure that it works as well as it reasonably can.

Most of this API is pretty ok, but `writeWithUnsafeMutableBytes` is a very tricky function to use properly. This is because, unlike all of the other `write` functions, it is not able to tolerate arbitrarily large writes. The reason here is that by the time the body gets the pointer to the writable bytes it is not possible to resize the underlying buffer.

The safe way to use this function is to preflight by telling the `ByteBuffer` how many bytes you need. You can do this via `reserveCapacity`, but calculating the correct number to pass to that function is annoying.

We should enhance `writeWithUnsafeMutableBytes` to take an `Int` that indicates how many bytes the caller would like to be free, and then we can do this calculation ourselves.

2bjake:
I thought I'd take a stab at this one, but I wanted to make sure I understood the intention correctly. Does this look right?

```swift
    public mutating func writeWithUnsafeMutableBytes(bodyByteCount: Int? = nil, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
        if let bodyByteCount = bodyByteCount, bodyByteCount > capacity - writerIndex {
            reserveCapacity(writerIndex + bodyByteCount)
        }
        // write bytes (existing function body)
    }
```

Lukasa:
Broadly right, but we want to do a bit more. Specifically we want the user to ask for a minimum number of uninitialised bytes, not a total capacity, then calculate what the required capacity is.

2bjake:
The user should be able to specify the minimum number of uninitialized bytes (writableBytes?) before the write occurs? If that's the case, I think this code does that (though the name ```bodyByteCount``` should be changed to something like ```withReservedWritableBytes```).

Unless I'm mistaken (which is quite possible!) adding the user specified writableBytes count to the ```writerIndex``` will calculate a new minimumCapacity that can be passed to ```reserveCapacity```.

Lukasa:
Yeah, writable bytes, and good catch: I missed that.  That’s what I get for reviewing while walking down the street!

Lukasa:
So yes, this is broadly correct.

weissi:
fixed with #1175

PR #1175, opened 2019-10-22
Add option to reserve writable capacity to writeWithUnsafeMutableBytes (#1105)

2bjake:
Add a version of writeWithUnsafeMutableBytes which allows users to specifie a minimum write capacity (#1105)

### Motivation:

```writeWithUnsafeMutableBytes(_:)``` cannot tolerate arbitrarily large writes because it is not possible to resize the buffer by the time body receives the buffer pointer. This change adds a new method ```writeWithUnsafeMutableBytes(minimumWritableBytes:_:)``` which allows the user to reserve writable capacity before the pointer is passed to body.

### Modifications:

Add a new method ```writeWithUnsafeMutableBytes(minimumWritableBytes:_:)``` which allows the user to reserve writable capacity before the pointer is passed to body. Deprecate the existing ```writeWithUnsafeMutableBytes(_:)```.

### Result:

Use of ```writeWithUnsafeMutableBytes(_:)``` results in a deprecation warning. New method ```writeWithUnsafeMutableBytes(minimumWritableBytes:_:)``` is available to users.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -504,9 +504,9 @@ public struct ByteBuffer {
     @inlinable
     public mutating func writeWithUnsafeMutableBytes(minimumWritableBytes: Int, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
         if minimumWritableBytes > 0 {
-            self.reserveCapacity(writerIndex + minimumWritableBytes)
+            self.reserveCapacity(self.writerIndex + minimumWritableBytes)
         }
-        let bytesWritten = try withUnsafeMutableWritableBytes(body)
+        let bytesWritten = try self.withUnsafeMutableWritableBytes(body)
         self._moveWriterIndex(to: self._writerIndex + _toIndex(bytesWritten))
         return bytesWritten
     }
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -492,24 +492,30 @@ public struct ByteBuffer {
         return try body(.init(rebasing: self._slicedStorageBuffer.dropFirst(self.writerIndex)))
     }
 
-    /// This vends a pointer of the `ByteBuffer` at the `writeIndex`. If `minimumWritableBytes` is specified, then the `ByteBuffer` will reserve the necessary
-    /// capacity before vending the pointer to `body`.
+    /// This vends a pointer of the `ByteBuffer` at the `writerIndex` after ensuring that the buffer has at least `minimumWritableBytes` of writable bytes available.
     ///
     /// - warning: Do not escape the pointer from the closure for later use.
     ///
     /// - parameters:
-    ///     - minimumWritableBytes: The number of writable bytes to reserve capacity for before vending the `ByteBuffer` pointer to `body`. (optional parameter)
+    ///     - minimumWritableBytes: The number of writable bytes to reserve capacity for before vending the `ByteBuffer` pointer to `body`.
     ///     - body: The closure that will accept the yielded bytes and return the number of bytes written.
     /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
-    public mutating func writeWithUnsafeMutableBytes(minimumWritableBytes: Int? = nil, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
-        if let minimumWritableBytes = minimumWritableBytes {
-            reserveCapacity(writerIndex + minimumWritableBytes)
+    public mutating func writeWithUnsafeMutableBytes(minimumWritableBytes: Int, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
+        if minimumWritableBytes > 0 {
+            self.reserveCapacity(writerIndex + minimumWritableBytes)
         }
         let bytesWritten = try withUnsafeMutableWritableBytes(body)
         self._moveWriterIndex(to: self._writerIndex + _toIndex(bytesWritten))
         return bytesWritten
+    }
+
+    @available(*, deprecated, message: "please use writeWithUnsafeMutableBytes(minimumWritableBytes:_:) instead to ensure sufficient write capacity.")
+    @discardableResult
+    @inlinable
+    public mutating func writeWithUnsafeMutableBytes(_ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
+        return try self.writeWithUnsafeMutableBytes(minimumWritableBytes: 0, body)
     }
 
     /// This vends a pointer to the storage of the `ByteBuffer`. It's marked as _very unsafe_ because it might contain
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -492,9 +492,21 @@ public struct ByteBuffer {
         return try body(.init(rebasing: self._slicedStorageBuffer.dropFirst(self.writerIndex)))
     }
 
+    /// This vends a pointer of the `ByteBuffer` at the `writeIndex`. If `minimumWritableBytes` is specified, then the `ByteBuffer` will reserve the necessary
+    /// capacity before vending the pointer to `body`.
+    ///
+    /// - warning: Do not escape the pointer from the closure for later use.
+    ///
+    /// - parameters:
+    ///     - minimumWritableBytes: The number of writable bytes to reserve capacity for before vending the `ByteBuffer` pointer to `body`. (optional parameter)
+    ///     - body: The closure that will accept the yielded bytes and return the number of bytes written.
+    /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Sources/NIO/ByteBuffer-core.swift [hunk 1]
```diff
@@ -492,9 +492,21 @@ public struct ByteBuffer {
         return try body(.init(rebasing: self._slicedStorageBuffer.dropFirst(self.writerIndex)))
     }
 
+    /// This vends a pointer of the `ByteBuffer` at the `writeIndex`. If `minimumWritableBytes` is specified, then the `ByteBuffer` will reserve the necessary
```
Lukasa: Nit: `writerIndex`, not `writeIndex`.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -492,9 +492,21 @@ public struct ByteBuffer {
...
+    /// - warning: Do not escape the pointer from the closure for later use.
+    ///
+    /// - parameters:
+    ///     - minimumWritableBytes: The number of writable bytes to reserve capacity for before vending the `ByteBuffer` pointer to `body`. (optional parameter)
+    ///     - body: The closure that will accept the yielded bytes and return the number of bytes written.
+    /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
-    public mutating func writeWithUnsafeMutableBytes(_ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
+    public mutating func writeWithUnsafeMutableBytes(minimumWritableBytes: Int? = nil, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
```
Lukasa: Unfortunately this is a backward-incompatible API change: we cannot add new parameters to existing methods. It'd be better to add a new method where `minimumWritableBytes` is not optional, and that duplicates the body. We can then deprecate the old one (as I think it's just generally not useful), and have it call through to the new one.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -492,9 +492,21 @@ public struct ByteBuffer {
...
+        if let minimumWritableBytes = minimumWritableBytes {
+            reserveCapacity(writerIndex + minimumWritableBytes)
```
Lukasa: Nit: we use explicit `self` in the NIO project, at least where we remember to.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -233,6 +233,40 @@ class ByteBufferTest: XCTestCase {
...
+    func testWithMutableWritePointerWithMinimumSpecifiedAdjustsCapacity() {
+        XCTAssertEqual(0, buf.writerIndex)
+        XCTAssertEqual(512, buf.capacity)
+
+        var bytesWritten = buf.writeWithUnsafeMutableBytes{ (_: UnsafeMutableRawBufferPointer) in return 256 }
+        XCTAssertEqual(256, bytesWritten)
+        XCTAssertEqual(256, buf.writerIndex)
+        XCTAssertEqual(512, buf.capacity)
+
+        bytesWritten += buf.writeWithUnsafeMutableBytes(minimumWritableBytes: 1024) { (_: UnsafeMutableRawBufferPointer) in return 1024 }
```
Lukasa: Can these callbacks validate the pointer has the correct minimum size?

Lukasa:

Sources/NIO/ByteBuffer-core.swift
2bjake: Ah, nice catch. Fixed.

Sources/NIO/ByteBuffer-core.swift
2bjake: Ok, sounds good.

Sources/NIO/ByteBuffer-core.swift
2bjake: Ok, I'll make the change. I'm curious to hear the reasoning behind that style if you're willing to share. :)

Tests/NIOTests/ByteBufferTest.swift
2bjake: Good idea! Added.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -492,14 +492,32 @@ public struct ByteBuffer {
...
+    /// - parameters:
+    ///     - minimumWritableBytes: The number of writable bytes to reserve capacity for before vending the `ByteBuffer` pointer to `body`.
+    ///     - body: The closure that will accept the yielded bytes and return the number of bytes written.
+    /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
-    public mutating func writeWithUnsafeMutableBytes(_ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
+    public mutating func writeWithUnsafeMutableBytes(minimumWritableBytes: Int, _ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
+        if minimumWritableBytes > 0 {
+            self.reserveCapacity(writerIndex + minimumWritableBytes)
```
weissi: please use `self.writerIndex`

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -492,14 +492,32 @@ public struct ByteBuffer {
...
+        }
         let bytesWritten = try withUnsafeMutableWritableBytes(body)
         self._moveWriterIndex(to: self._writerIndex + _toIndex(bytesWritten))
         return bytesWritten
     }
 
+    @available(*, deprecated, message: "please use writeWithUnsafeMutableBytes(minimumWritableBytes:_:) instead to ensure sufficient write capacity.")
+    @discardableResult
+    @inlinable
+    public mutating func writeWithUnsafeMutableBytes(_ body: (UnsafeMutableRawBufferPointer) throws -> Int) rethrows -> Int {
```
weissi: Not sure if we should really deprecate this function? It doesn't cause any harm. CC @Lukasa / @glbrntt

glbrntt: I don't think we should: I think there are cases where you can elide the `self.reserveCapacity` call because enough space has been allocated ahead of time and the number of bytes written by `writeWithUnsafeMutableBytes` does not account for all of the bytes (i.e. allocate X bytes, write 4 bytes as an Int32, write X-4 bytes with `writeWithUnsafeMutableBytes`).

2bjake: Yeah, I tend I agree. @Lukasa asked me to deprecate it but as I was moving the codebase off of the now deprecated API, I was having to specify the ```minimumWritableBytes``` in places where the buffer had just been instantiated with the right size. So I was having to repeat myself unnecessarily.

Sources/NIO/ByteBuffer-core.swift
Lukasa: The reasoning is fairly straightforward: it avoids ambiguity. There is no question about where `reserveCapacity` comes from or what its side effects may be if you explicitly see that `self` is being passed to it.

Sources/NIO/ByteBuffer-core.swift
Lukasa: I think we should deprecate it because it is easier to misuse than anything else. It simply doesn't hurt to ask for the number of bytes you want, even if that's repeating yourself. It also avoids bugs by having NIO double-check your work. Specifically, `reserveCapacity` is an extremely cheap function call in the case where it does nothing, so I don't think the cost of making it hurts at all.

2bjake: I'm willing to do it either way, so just let me know what you guys decide.

weissi: I'm still in favour of not deprecating the old version because I don't see the real point. But I'm not strong on it. So could me as

> weakly if favour of keeping the old version

@Lukasa are you strongly or weakly in favour of removing the old version
@glbrntt are you strongly or weakly in favour of keeping the old version (or did you change your mind)
@2bjake do you have any further thoughts or would you like to just defer to us?

Lukasa: Strongly in favour of removing. I don't think the old version serves any value with the new one in existence. If we implemented this new version first we'd never write the old one, so we should move towards the better API.

glbrntt: I'm weakly in favour of keeping the older version.

weissi: Assuming @2bjake is not 'strongly in favour of keeping', let's kill it then (I'm counting weakly about as much as 0.3 or less)

glbrntt: Sounds good.

Sources/NIO/ByteBuffer-core.swift
2bjake: added

Sources/NIO/ByteBuffer-core.swift
2bjake: Yeah, I'm happy to defer to y'all.

weissi:

weissi:

Issue #1148, opened 2019-09-26
namespace ChannelOptions types


weissi:
all of our ChannelOptions enums like `WriteBufferWaterMarkOption` are global, public types. They should be namespaced into `ChannelOptions.Types.WriteBufferWaterMarkOption` or similar.

This can be done before NIO 3 by replacing the current ones with deprecated `public typealiases`.

PR #1176, opened 2019-10-22
namespace ChannelOptions types (#1148)

2bjake:
### Motivation:

ChannelOptions types like WriteBufferWaterMarkOption are global, public types. These types should be namespaced.

### Modifications:

Moved global, public ChannelOptions types into ChannelOptions.Types enum. Created global, public typealiases with deprecation attributes which point to the namespaced types. Switched usage of these types in the codebase to use the namespaced versions.

### Result:

Usage of the global, public versions of these types will cause deprecation warnings.


Lukasa:

PR #1179, opened 2019-10-23
make use of EventLoop.flatSubmit

weissi:
Motivation:

Since #1174 we've finally got EventLoop.flatSubmit and we should make
use of it to replace the silly `.wait().wait()` calls as well as the
similarly silly `.flatMap { $0 }`.

Modifications:

Use `flatSubmit` where appropriate.

Result:

Nicer codebase.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -347,7 +347,7 @@ extension TimeAmount {
 ///
 /// ```
 /// func doSomething(deadline: NIODeadline) -> EventLoopFuture<Void> {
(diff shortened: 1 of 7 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -347,7 +347,7 @@ extension TimeAmount {
 ///
 /// ```
 /// func doSomething(deadline: NIODeadline) -> EventLoopFuture<Void> {
-///     return step1(deadline: deadline).then {
+///     return step1(deadline: deadline).flatMap {
```
glbrntt: Nice find!

PR #1192, opened 2019-10-25
SocketChannel: allow bind before register

weissi:
Motivation:

In certain cases, on a client channel, the user might want to `bind` in
the `channelInitializer` which happens before the channel is registered
or active.

Modifications:

Allow `bind` in the `channelInitializer`.

Result:

Can `bind` before `connect`.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -565,10 +565,6 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             promise?.fail(ChannelError.ioOnClosedChannel)
             return
         }
-        guard self.lifecycleManager.isPreRegistered else {
-            promise?.fail(ChannelError.inappropriateOperationForState)
-            return
-        }
 
         executeAndComplete(promise) {
             try socket.bind(to: address)
--- a/Tests/NIOTests/BootstrapTest+XCTest.swift
+++ b/Tests/NIOTests/BootstrapTest+XCTest.swift
@@ -31,6 +31,7 @@ extension BootstrapTest {
                 ("testUDPBootstrapToleratesFuturesFromDifferentEventLoopsReturnedInInitializers", testUDPBootstrapToleratesFuturesFromDifferentEventLoopsReturnedInInitializers),
                 ("testPreConnectedClientSocketToleratesFuturesFromDifferentEventLoopsReturnedInInitializers", testPreConnectedClientSocketToleratesFuturesFromDifferentEventLoopsReturnedInInitializers),
                 ("testPreConnectedServerSocketToleratesFuturesFromDifferentEventLoopsReturnedInInitializers", testPreConnectedServerSocketToleratesFuturesFromDifferentEventLoopsReturnedInInitializers),
+                ("testCanBindOnClientConnection", testCanBindOnClientConnection),
            ]
    }
 }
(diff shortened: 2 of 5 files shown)


weissi:
ping @Lukasa

Lukasa:
I believe we have an outstanding question on this [here](https://github.com/apple/swift-nio/pull/1165#issuecomment-546307316).

weissi:
@Lukasa apologies. So I agree we should think about the question over there and I think we should at least file an issue to resolve/think about it. On the other hand, this PR here adds functionality and doesn't make any working program worse. Therefore I'd argue that is an improvement. Sure, the new functionality is a bit brittle to use because you need to know how to not use happy eyeballs by resolving the address manually before.
Regarding your question, I'd say we should take option 1 (don't care) for now and in an issue that is to be filed we should decide if we want 2 or 3. WDYT?

Lukasa:
I'm fine with choosing to not care. I'm not necessarily thrilled about it, but we can live with that outcome. We should file an issue to add support that is compatible with HE.

Lukasa:
I think the comment here needs addressing.

Lukasa:
I think the comment here needs addressing.

weissi:
@Lukasa so you'd be happy to leave the HE wrinkle in here for now (creating an issue ofc) but you'd like me to add support for ServerSocketChannel and DatagramChannel?

Lukasa:
No, my thing is that I think it _does_ work now with the change you've made, but the tests don't explore the behaviour. I think they should.

normanmaurer:
@weissi why not expose a `bind(...)` method on the bootstrap itself ? This way the user can bind and the call `connect(...)` on the channel itself. ? Thats what we do in netty.

Lukasa:
We don't want them to call `connect` on the channel itself because then they lose Happy Eyeballs. A proper Happy-Eyeballs aware implementation _does_ have both a `bind` and `connect` method on the client bootstrap, but only so that HE knows what the local interface will be to discard a bunch of connection targets.

normanmaurer:
@Lukasa hmm... but isn't it quite error prone to call `bind(...)` in the initialiser as well as the user may end up trying to bind to the same port multiple times (as the initialiser may be run multiple times) ?

Lukasa:
Yes, but that's no different to calling `bind` then `connect` except that they do wasted work. In both cases only one connection attempt can be made.

The _right_ thing to do is to tell the resolver to restrain itself to only connections that can be made from a given local interface. `getaddrinfo` is capable of using that information to scope its results, as are many others. But the whole nature of this patch is that it doesn't do that.

weissi:
@Lukasa what about we do `bind` on the bootstrap and disable HE if that has been done?

Lukasa:
I'm fine with that. It's not quite as good, but it doesn't require an amendment to the `Resolver` interface.

weissi:
@Lukasa exactly. It's a pretty rare use-case so I think that's okay. We can add an issue for NIO3. @normanmaurer WDYT?

Lukasa:
I don't think it has to be NIO3: the in-planning "new" resolver interface can take a `hints` structure that requests that all provided addresses be reachable from a given interface. Then we can go back to firing `bind` through the pipeline for HE.

weissi:
@Lukasa very good point!

weissi:
@Lukasa is there a bug for the new resolver interface already? So we can add thoughts/requirements?

Lukasa:
#1487

Lukasa:
Closing in favour of #1490.

PR #1193, opened 2019-10-25
Make future builders and scheduled tasks inlinable

Lukasa:
Motivation:

While we successfully made all operations _on_ futures inlinable a while
ago, we managed to miss the functions that construct futures. These
remain public generic non-inlinable functions, and are called quite
heavily in NIO programs. Making them inlinable would unlock a bunch of
free performance.

Modifications:

- Make future and promise constructors inlinable.
- Make Scheduled inlinable.

Result:

A whole bunch of code will be faster.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -20,10 +20,11 @@ import Dispatch
 /// A `Scheduled` allows the user to either `cancel()` the execution of the scheduled task (if possible) or obtain a reference to the `EventLoopFuture` that
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
-    private let promise: EventLoopPromise<T>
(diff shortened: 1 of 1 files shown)


Lukasa:

Sources/NIO/EventLoop.swift
```diff
@@ -20,8 +20,9 @@ import Dispatch
 /// A `Scheduled` allows the user to either `cancel()` the execution of the scheduled task (if possible) or obtain a reference to the `EventLoopFuture` that
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
-    private let promise: EventLoopPromise<T>
+    /* private but usableFromInline */ @usableFromInline let promise: EventLoopPromise<T>
```
weissi: please underscore because really it's private :)

Lukasa: Done.

Issue #1195, opened 2019-10-27
ApplicationProtocolNegotiationHandler should accept a closure that delivers the Channel


Lukasa:
Currently `ApplicationProtocolNegotiationHandler` accepts a closure that takes only one argument, the ALPN result. That's fine, but it causes a needless allocation on every connection setup attempt because the channel is inevitably going to be mutated, and as a result it must be closed over.

That's a bit sad, and we could easily remove the allocation in performance-sensitive use-cases by having the closure take two arguments, the result and the channel. Adding this should be easy enough: we can create a new initializer that takes the new closure. We can then reimplement the old initializer in terms of the new one by simply wrapping it.

We may want to be a bit careful that we haven't led to ambiguity, but our tests should catch that, though we want to double check the work.

PR #1196, opened 2019-10-28
Add ApplicationProtocolNegotiationHandler init that takes a closure that receives the Channel (#1195)

2bjake:
Motivation:

Currently ApplicationProtocolNegotiationHandler accepts a closure that takes only one argument, the ALPN result. This forces the user to capture the Channel in the closure so that it can be mutated.

Modifications:

Add a new init which takes a closure that has both the result and the Channel as parameters. Modify the original init to call the new init (wrapping the passed in closure.)

Result:

New APNL init available that takes a 2 parameter closure. Original APNL init will now be a convenience init.


weissi:

PR #1197, opened 2019-10-29
test that we tolerate out-of-band data

weissi:
Motivation:

We don't support OOB data but we should make sure that nothing breaks if
OOB arrives.

Modifications:

Add a test that makes sure we tolerate OOB data.

Result:

More test coverage.


Sources/NIO/System.swift
```diff
@@ -364,10 +364,13 @@ internal enum Posix {
     }
 
     @inline(never)
-    public static func sendto(descriptor: CInt, pointer: UnsafeRawPointer, size: size_t,
+    public static func sendto(descriptor: CInt,
+                              pointer: UnsafeRawPointer,
+                              size: size_t,
+                              flags: CInt,
```
glbrntt: ~Isn’t this an API breaking change? Or am I missing something?~

How come we have public methods on an internal enum?

glbrntt:
New test failed on 5.1... 👎

weissi:
Aged out, closing.

PR #1198, opened 2019-10-29
add socketpair to the POSIX wrappers

weissi:
Motivation:

For some reason, we never added `socketpair` to the POSIX wrappers but
it should be there.

Modifications:

Add socketpair to the POSIX wrappers and make use of it.

Result:

Nicer code base.


Lukasa:

PR #1200, opened 2019-10-30
Make SocketAddress.port mutable

### LINKED ISSUE #1189: NIO makes it _incredibly_ hard to change a port number on a given socket address
if we want to set the port number on a given `SocketAddress` we need to use code like this which is really dodgy and outright wrong if we don't know it's an IPv4 address:

```
let destAddress = en0Interface.broadcastAddress?.withSockAddr { (ptr, size) -> sockaddr_in in
    // this is terrible code making fragile assumptions just to change a port number :(
    precondition(size == MemoryLayout<sockaddr_in>.size)
    return ptr.withMemoryRebound(to: sockaddr_in.self, capacity: 1) { ptr in
        var newAddress = sockaddr_in()
        newAddress = ptr.pointee
        newAddress.sin_port = in_port_t(destPort).bigEndian
        return newAddress
    }
}
```

we need to fix this.

siemensikkema:
Make SocketAddress.port mutable

### Motivation:

Previously the port was read only. Fixes #1189.
 
### Modifications:

In case of IPv4 and IPv6 the port will be set to the new value or defaulted to 0 when setting it with the value `nil`. The rationale for this is that both `nil` and `0` can be interpreted as "no preference". When the underlying socket is a unix socket the incoming value is ignored.

### Result:

SocketAddress.port can now be written to as well as be read.

### DIFF
--- a/Sources/NIO/SocketAddresses.swift
+++ b/Sources/NIO/SocketAddresses.swift
@@ -146,17 +146,36 @@ public enum SocketAddress: CustomStringConvertible {
         }
     }
 
-    /// Get the port associated with the address, if defined.
+    /// Get and set the port associated with the address, if defined.
+    /// When setting to `nil` the port will default to `0` for compatible sockets. The rationale for this is that both `nil` and `0` can
+    /// be interpreted as "no preference".
     public var port: Int? {
-        switch self {
-        case .v4(let addr):
-            // looks odd but we need to first convert the endianness as `in_port_t` and then make the result an `Int`.
-            return Int(in_port_t(bigEndian: addr.address.sin_port))
-        case .v6(let addr):
-            // looks odd but we need to first convert the endianness as `in_port_t` and then make the result an `Int`.
-            return Int(in_port_t(bigEndian: addr.address.sin6_port))
-        case .unixDomainSocket:
-            return nil
+        get {
+            switch self {
+            case .v4(let addr):
+                // looks odd but we need to first convert the endianness as `in_port_t` and then make the result an `Int`.
+                return Int(in_port_t(bigEndian: addr.address.sin_port))
+            case .v6(let addr):
+                // looks odd but we need to first convert the endianness as `in_port_t` and then make the result an `Int`.
+                return Int(in_port_t(bigEndian: addr.address.sin6_port))
+            case .unixDomainSocket:
+                return nil
+            }
+        }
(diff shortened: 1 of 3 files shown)


Lukasa:

Davidde94:
Run `generate_linux_tests.rb` to let non-Apple platforms run the new test `testPortIsMutable` :)

Sources/NIO/SocketAddresses.swift
```diff
@@ -146,17 +146,36 @@ public enum SocketAddress: CustomStringConvertible {
...
+            case .v4(let addr):
+                var mutAddr = addr.address
+                mutAddr.sin_port = in_port_t(newValue ?? 0).bigEndian
+                self = .v4(.init(address: mutAddr, host: addr.host))
+            case .v6(let addr):
+                var mutAddr = addr.address
+                mutAddr.sin6_port = in_port_t(newValue ?? 0).bigEndian
+                self = .v6(.init(address: mutAddr, host: addr.host))
+            case .unixDomainSocket:
+                // ignore
```
glbrntt: We should be checking whether `newValue` is `nil` or not here as per @Lukasa's comment:  https://github.com/apple/swift-nio/issues/1189#issuecomment-546255468

It would also be worth documenting these semantics!

siemensikkema: @glbrntt Should that be a precondition? Fatal error?

glbrntt: `precondition` should be okay here

Sources/NIO/SocketAddresses.swift
```diff
@@ -146,17 +146,36 @@ public enum SocketAddress: CustomStringConvertible {
         }
     }
 
-    /// Get the port associated with the address, if defined.
+    /// Get and set the port associated with the address, if defined.
+    /// When setting to `nil` the port will default to `0` for compatible sockets. The rationale for this is that both `nil` and `0` can
+    /// be interpreted as "no preference".
+    /// Setting a non-nil value to a unix port is invalid and will result in a fatal error.
```
glbrntt:
```suggestion
    /// Setting a non-nil value for a unix domain socket is invalid and will result in a fatal error.
```

Lukasa:

Lukasa:

Issue #1057, opened 2019-07-02
ByteBuffer.clear(minimumCapacity:) API


weissi:
We should have an API in `ByteBuffer` that works like `clear` but requests at least `minimumCapacity` bytes. The use for that is if the current buffer is uniquely referenced, it'll just be cleared (followed by `reserveCapacity`) but if the current buffer is not uniquely referenced, then we'd allocate a new one with `minimumCapacity`.

weissi:
This could be a good first issue. The only NIO knowledge that's required would be basic `ByteBuffer` usage. If this issue sounds attractive, please comment here and we can give much more information.

The basic idea is that often it's beneficial to re-use an existing `ByteBuffer` if that doesn't cause a copy-on-write copy (regardless of the size of the existing `ByteBuffer`). If it would cause a copy-on-write copy however it's often beneficial to allocate a new `ByteBuffer` of _the right size_ rather than the size of the one that happens to be around.

akshay1992kalbhor:
Hey Johannes, has this been assigned to anyone? If not can I take a shot at this? It will be my first contribution to this repo!

weissi:
hi @akshay1992kalbhor , thanks for reaching out! No this hasn't been assigned to anybody yet. Please feel free to assign it to yourself. And please reach out if there's anything we can help with :)

akshay1992kalbhor:
Thanks. I think I have a fairly good understanding of how ByteBuffer works and the first two comments will hopefully allow me to start working on this one. Is there anything else I should know before starting?

akshay1992kalbhor:
Also, I don't have permission to assign the issue to myself. Can someone help me with that?

weissi:
Thank you and done!

weissi:
Don’t think there’s anything else to know if the issue makes sense to you

weissi:
@akshay1992kalbhor are you still working on this?

kam800:
I will work on that.

Keno42:
I'm working on it, too.

Lukasa:
Resolved by #1204, with the work done collaboratively by @keno42, @kam800, and @cweinberger. Great work all!

PR #1204, opened 2019-10-30
Add an option to ByteBuffer.clear() to specify minimumCapacity

cweinberger:
Add an option to `ByteBuffer.clear()` to specify `minimumCapacity`.

### Motivation:

The basic idea is that often it's beneficial to re-use an existing `ByteBuffer` if that doesn't cause a copy-on-write copy (regardless of the size of the existing `ByteBuffer`). If it would cause a copy-on-write copy however it's often beneficial to allocate a new `ByteBuffer` of the right size rather than the size of the one that happens to be around. See #1057 

### Modifications:

* changed visibility of `_Storage()allocateStorage` from `private` to `internal`
* added an implementation of `ByteBuffer().clear()` that takes a parameter to specify desired capacity`capacity minimumNeededCapacity: _Capacity`
* added tests:
  * `ByteBufferTests().testClearDoesReallocateStorageCorrectlyIfTheresOnlyOneBuffer()`
  * `ByteBufferTests().testClearDoesAllocateStorageCorrectlyIfTheresTwoBuffersSharingStorage()`

### Result:

* NIO users can provide a minimum capacity when clearing a `ByteBuffer`.
* Existing implementations are not affected.


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -235,7 +235,7 @@ public struct ByteBuffer {
             return self.allocateStorage(capacity: self.capacity)
         }
 
-        private func allocateStorage(capacity: _Capacity) -> _Storage {
+        public func allocateStorage(capacity: _Capacity) -> _Storage {
```
Lukasa: This should be `internal`: it doesn’t need higher visibility than the containing object.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -659,9 +659,15 @@ public struct ByteBuffer {
     ///
     /// - note: This method will allocate if the underlying storage is referenced by another `ByteBuffer`. Even if an
     ///         allocation is necessary this will be cheaper as the copy of the storage is elided.
-    public mutating func clear() {
```
Lukasa: Can we have this function back? I’m afraid it’s a semver major to remove it. We can just implement it in terms of the new function.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -659,9 +659,15 @@ public struct ByteBuffer {
...
+    ///
+    /// - parameters:
+    ///     - minimumNeededCapacity: The minimum capacity that will be (re)allocated for this buffer
+    public mutating func clear(capacity minimumNeededCapacity: _Capacity? = nil) {
```
Lukasa: Combined with the comment above, this capacity may as well be non-optional and remove the default value.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -659,9 +659,15 @@ public struct ByteBuffer {
...
+        let capacity = minimumNeededCapacity ?? self._storage.capacity
```
Lukasa: I’m not sure we should shrink the buffer here: @weissi?

Sources/NIO/ByteBuffer-core.swift
cweinberger: Sure! Updated!

Sources/NIO/ByteBuffer-core.swift
cweinberger: Sure! Updated!

Sources/NIO/ByteBuffer-core.swift
cweinberger: Sure! Updated!

cweinberger:
@Lukasa I have incorporated your feedback and pushed my changes.

Lukasa:

cweinberger:
> Can you please run `ruby scripts/generate_linux_tests.rb`?
Sure, done!

Sources/NIO/ByteBuffer-core.swift
cweinberger: With @Keno42 commit this has been changed to never shrink the capacity.

Lukasa:

Lukasa:
Unfortunately this patch has a small conflict in the auto generated test file. If you rebase this onto the current master branch you should be able to regenerate the test file to resolve the conflict.

Lukasa:

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -667,6 +667,27 @@ public struct ByteBuffer {
...
+    /// Set both reader index and writer index to `0`. This will reset the state of this `ByteBuffer` to the state
+    /// of a freshly allocated one, if possible without allocations. This is the cheapest way to recycle a `ByteBuffer`
+    /// for a new use-case.
+    ///
+    /// - note: This method will allocate if the underlying storage is referenced by another `ByteBuffer`. Even if an
+    ///         allocation is necessary this will be cheaper as the copy of the storage is elided.
+    ///
+    /// - parameters:
+    ///     - minimumNeededCapacity: The minimum capacity that will be (re)allocated for this buffer
+    public mutating func clear(capacity minimumNeededCapacity: _Capacity) {
```
weissi: I think this parameter should be named `minimumCapacity` because we are likely to give you a larger buffer, especially if the existing buffer is uniquely referenced.

Lukasa: +1

cweinberger: @weissi @Lukasa I chose `minimumNeededCapacity` intentionally to align with the existing API from `public func reallocStorage(capacity minimumNeededCapacity: _Capacity)`

cweinberger: Changes have been implemented!

glbrntt: Also +1 for `minimumCapacity`

Lukasa:

Lukasa:
@weissi want to re-review?

PR #1206, opened 2019-10-30
Fix `NIODeadline.Value` deprecation message typo

Davidde94:
Fixes a typo `doesn't server any purpose` -> `doesn't serve any purpose`

### Motivation:

Typos are bad.

### Modifications:

Simple correction to the deprecation message for `typealias Value = UInt64`

### Result:

No downstream impact and no patch release required. Can be rolled out with the next release.


Lukasa:

PR #1208, opened 2019-10-30
Make ByteBufferView mutable

### LINKED ISSUE #1160: Extend ByteBufferView to be a MutableCollection and a RangeReplaceableCollection
When we added `ByteBufferView` (#411), @weissi and I thought that we couldn't make it mutable because it would inevitably CoW. This is because the way you build `ByteBufferView` objects is as constructed objects from a `ByteBuffer` (i.e. via `ByteBuffer.viewBytes(at:length:)` or `ByteBuffer.readableBytesView`), and as a result the original buffer still exists and holds a reference to the backing storage.

What we missed is that this can be enabled by adding a `ByteBufferView.init()` function that takes a `ByteBuffer` as its argument. In this mode the `ByteBufferView` would take ownership of the `ByteBuffer` entirely, including the writable bytes, and could therefore appropriately conform to the mutable protocols. This also creates proper bidirectional conversion between `ByteBuffer` and `ByteBufferView`, which is handy when interoperating with algorithms written generically over the Swift Collection types.

To make this work, we'd need to do a few things:

1. Add the appropriate initializer.
2. Extend `ByteBufferView` to have an idea that it may hold a slice larger than its range.
3. Implement `MutableCollection` and `RangeReplaceableCollection` for `ByteBufferView`.
4. Write tests verifying that, when using this API on the happy path, we don't incur CoW operations.
5. (optional, extension) Extend `NIOFoundationCompat` so that `ByteBufferView` conforms to `MutableDataProtocol` as well.

glbrntt:
Motivation:

ByteBufferView isn't a mutable collection, but it probably should be.

Modifications:

- Add another `ByteBufferView` initializer which takes ownership of a buffer
- Add another `ByteBuffer` initializer which takes ownership of a view
- Conform `ByteBufferView` to `MutableCollection`, `RangeReplaceableCollection`
  and `MutableDataProtocol`
- Deprecate `ByteBuffer.readableBytesView` in favour of `ByteBuffer.view`
- Add an allocation counting test

Result:

- `ByteBufferView` is now mutable.
- fixes #1160


Sources/NIO/ByteBuffer-views.swift
```diff
@@ -16,48 +16,161 @@
...
+    }
+
+    /// Copy bytes from `newElements` to the given `subrange`.
+    ///
+    /// - Precondition: `newElements.count` must be less than or equal to `subrange.count`.
+    /// - Parameter subrange: The subrange of the view into why the elements should be copied.
+    /// - Parameter newElements: The elements to copy into the subrange.
+    @inlinable
+    /* private but inlinable */ mutating func _copyBytes<C: Collection>(to subrange: Range<Index>, from newElements: C) where ByteBufferView.Element == C.Element {
+        precondition(newElements.count <= subrange.count, "subrange must not be longer than newElements")
```
weissi: isn't this whole function just

    self._buffer.setBytes(newElements, at: subrange.startIndex)

?

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -16,48 +16,161 @@
...
+
+                    let numberOfBytesToMove = endOfRangeToReplace.distance(to: oldWriterIndex)
+                    memmove(dstPtr, srcPtr, numberOfBytesToMove)
+                }
+            }
+        } else {
+            // We have more elements to write in the subrange than the subrange currently has space
+            // for: widen the capacity of the underlying buffer.
+            let additionalByteCount = newElements.count - subrange.count
+            self._buffer.reserveCapacity(self._buffer.readableBytes + additionalByteCount)
```
weissi: this isn't quite right I think, this should be `self._buffer.writerIndex + additionalByteCount`.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -69,8 +182,14 @@ public struct ByteBufferView: RandomAccessCollection {
 }
 
 extension ByteBuffer {
-    /// A view into the readable bytes of the `ByteBuffer`.
+    /// A view into the bytes of the `ByteBuffer`.
+    @available(*, deprecated, renamed: "view")
```
weissi: I don't think we should rename `readableBytesView`. What's wrong with just leaving this as is?

Sources/NIO/SocketChannel.swift
```diff
@@ -491,7 +491,7 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
 
     private func vectorReadFromSocket() throws -> ReadResult {
         #if os(Linux) || os(FreeBSD) || os(Android)
-        var buffer = self.recvAllocator.buffer(allocator: self.allocator)
+        var _buffer = self.recvAllocator.buffer(allocator: self.allocator)
```
weissi: this change doesn't seem related.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -16,48 +16,161 @@
 ///
 /// A `ByteBufferView` is useful whenever a `Collection where Element == UInt8` representing a portion of a
 /// `ByteBuffer` is needed.
-public struct ByteBufferView: RandomAccessCollection {
+public struct ByteBufferView: RandomAccessCollection, MutableCollection, RangeReplaceableCollection {
```
weissi: I think it would be better to put the `MutableCollection, RangeReplaceableCollection` conformances into extensions instead of directly onto the type?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1524,30 +1524,144 @@ class ByteBufferTest: XCTestCase {
...
+
+        var anotherView = anotherBuf.view
+        anotherView.replaceSubrange(0..<3, with: [])
+        XCTAssertTrue(anotherView.isEmpty)
+
+        self.buf = ByteBuffer(takingOwnershipOf: anotherView)
+        XCTAssertEqual(self.buf.readerIndex, 0)
+        XCTAssertEqual(self.buf.writerIndex, 0)
+        XCTAssertEqual([], self.buf.readBytes(length: self.buf.readableBytes))
+    }
```
weissi: we need some tests that prove that you can hold a mutable `ByteBufferView` in a way that it doesn't actually trigger a CoW.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -16,48 +16,161 @@
...
+
+        if newElements.count <= subrange.count {
+            // Copy the bytes over into the subrange.
+            self._copyBytes(to: subrange, from: newElements)
+
+            // `newElements` is shorter than subrange, so we now have unwanted bytes between the
+            // copied bytes and the end of the subrange. Let's move all of the bytes after the
+            // subrange taking care to update the buffer's writer index as well.
+            if newElements.count < subrange.count {
+                let oldWriterIndex = self._buffer.writerIndex
```
weissi: instead of doing all of this inline, could we create a `ByteBuffer` function

    public func moveBytes(at atIndex: Int, to toIndex: Int, length: Int) {
        ...
    }

or something similar

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -16,48 +16,161 @@
...
+                }
+            }
+        } else {
+            // We have more elements to write in the subrange than the subrange currently has space
+            // for: widen the capacity of the underlying buffer.
+            let additionalByteCount = newElements.count - subrange.count
+            self._buffer.reserveCapacity(self._buffer.readableBytes + additionalByteCount)
+
+            // We have capacity but we still need to shift everything after the subrange forwards.
+            let writerIndex = self._buffer.writerIndex
```
weissi: this code should also be expressed with the new `moveBytes`

Sources/NIO/ByteBuffer-views.swift
glbrntt: Ah, yup, it is!

Tests/NIOTests/ByteBufferTest.swift
glbrntt: There's an allocation test at the top of this PR, unfortunately Swift can't currently doesn't avoid the CoW but as of https://github.com/apple/swift/commit/6acf8db2468ee1aaed346f1fc38a5179841722b4 this has been fixed (confirmed by re-running the alloc test)

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -740,6 +740,37 @@ extension ByteBuffer {
...
+        if isKnownUniquelyReferenced(&self._storage) {
+            self._ensureAvailableCapacity(_toCapacity(newWriterIndex), at: 0)
+        } else {
+            self._copyStorageAndRebase(capacity: _toCapacity(newWriterIndex))
+        }
+
+        self.withVeryUnsafeMutableBytes { ptr in
+            let srcPtr = ptr.baseAddress!.advanced(by: fromIndex)
+            let dstPtr = ptr.baseAddress!.advanced(by: toIndex)
+            memmove(dstPtr, srcPtr, length)
```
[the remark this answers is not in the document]
glbrntt: Okay, I think I follow, so this function would become something like:

```swift
func copyBytes(at fromIndex: Int, to toIndex: Int, length: Int) {
        self._copyStorageAndRebase(...)
    }

    self._ensureAvailableCapacity(...)
    self.withVeryUnsafeBytes { ptr in
        let srcPtr = UnsafeRawBufferPointer(start: ptr.baseAddress!.advanced(by: fromIndex), count: length)
        self._setBytesAssumingUniqueBufferAccess(srcPtr, at: _toIndex(toIndex))    
    }
}
```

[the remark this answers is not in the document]
weissi: @glbrntt yes, that looks great! Plus probably throwing as we discussed earlier :)

glbrntt:

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -800,6 +804,53 @@ extension ByteBuffer {
...
+    ///
+    /// - Note: Overlapping ranges, for example `copyBytes(at: 1, to: 2, length: 5)` are allowed.
+    /// - Precondition: The range represented by `fromIndex` and `length` must be readable bytes,
+    ///     that is: `fromIndex >= readerIndex` and `fromIndex + length <= writerIndex`.
+    /// - Parameter fromIndex: The index of the first byte to copy.
+    /// - Parameter toIndex: The index into to which the first byte will be copied.
+    /// - Parameter length: The number of bytes which should be copied.
+    @discardableResult
+    @inlinable
+    public mutating func copyBytes(at fromIndex: Int, to toIndex: Int, length: Int) throws -> Int {
```
Lukasa: Throwing errors is a slightly unusual property for `ByteBuffer`: we normally use preconditions or optional returns. I think the only reason this has to throw is because it's public API, otherwise the error cases in which we throw could technically be other things, right?

Are we happy adding this public API to `ByteBuffer` @weissi?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -800,6 +804,53 @@ extension ByteBuffer {
...
+        guard length > 0 else {
+            if length == 0 {
+                return 0
+            } else {
+                throw CopyBytesError.negativeLength
+            }
+        }
```
Lukasa: This would better be represented as a `switch`.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -800,6 +804,53 @@ extension ByteBuffer {
...
+        guard self.readerIndex <= fromIndex && fromIndex + length <= self.writerIndex else {
+            throw CopyBytesError.unreadableSourceBytes
+        }
+
+        let newEndIndex = max(self._writerIndex, _toIndex(toIndex + length))
```
Lukasa: We may as well do this math inside the `if` block and save ourselves some computation.

Sources/NIO/ByteBuffer-core.swift
Lukasa: I actually wonder if this API should be `internal` for now and we can simply replace all the `throws` with `precondition`. The only callers of this API use `try!` anyway.

glbrntt: Discussed both options with @weissi before – IIRC the preference for `public` over `internal` wasn't strong. I'm happy to do either.

Lukasa: Let's leave this as-is.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -800,6 +804,53 @@ extension ByteBuffer {
     }
 }
 
+extension ByteBuffer {
+    /// Errors thrown when calling `copyBytes`.
+    public enum CopyBytesError: Error, Equatable {
```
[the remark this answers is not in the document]
glbrntt: Updated to be extensible.

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -800,6 +804,69 @@ extension ByteBuffer {
...
+            let srcPtr = UnsafeRawBufferPointer(start: ptr.baseAddress!.advanced(by: fromIndex), count: length)
+            self._setBytesAssumingUniqueBufferAccess(srcPtr, at: _toIndex(toIndex))
+        }
+
+        return length
+    }
+
+    /// Errors thrown when calling `copyBytes`.
+    public struct CopyBytesError: Error {
+        internal enum BaseError: Error, Hashable {
```
Lukasa: This internal enum doesn't need to conform to `Error`. It may be a good idea to prefix this with an underscore as well to discourage its use internally.

glbrntt: Made it `private` instead.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -21,43 +21,61 @@ public struct ByteBufferView: RandomAccessCollection {
...
         precondition(range.lowerBound >= 0 && range.upperBound <= buffer.capacity)
-        self.buffer = buffer
-        self.range = range
+        self._buffer = buffer
+        self._range = range
+    }
+
+    /// Creates a `ByteBufferView` from the readable bytes of the given `buffer`.
+    public init(_ buffer: ByteBuffer) {
+        self = ByteBufferView(buffer: buffer, range: buffer.readerIndex ..< buffer.readerIndex + buffer.readableBytes)
```
Lukasa: Nit: `buffer.readerIndex + buffer.readableBytes` is `buffer.writerIndex`, so we can just use that directly.

PR #1209, opened 2019-10-30
Add API ByteBuffer.clear(minimumCapacity:) (#1057)

### LINKED ISSUE #1057: ByteBuffer.clear(minimumCapacity:) API

Keno42:
### Motivation:

This commit closes #1057.
The demand is a new API to ensure that the buffer has space for at least minimumCapacity after clearing the buffer.
In case the storage is not uniquely referenced, and the minimumCapacity is much smaller than the current capacity, this API can avoid allocating excessive memory.

### Modifications:

Add a new API ByteBuffer.clear(minimumCapacity:)

### Result:

More memory performant buffer clearing.

### DIFF
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -1227,6 +1227,76 @@ class ByteBufferTest: XCTestCase {
         }
         XCTAssertEqual(bufPtrValPre, bufPtrValPost)
     }
+    
+    func testClearWithBiggerMinimumCapacityDupesStorageIfTheresTwoBuffersSharingStorage() throws {
+        let alloc = ByteBufferAllocator()
+        let buf1 = alloc.buffer(capacity: 16)
+        var buf2 = buf1
+
+        var buf1PtrVal: UInt = 1
+        var buf2PtrVal: UInt = 2
+        
+        buf1.assign(to: &buf1PtrVal)
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertEqual(buf1PtrVal, buf2PtrVal)
+
+        buf2.clear(minimumCapacity: 32)
+
+        buf1.assign(to: &buf1PtrVal)
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertNotEqual(buf1PtrVal, buf2PtrVal)
+    }
+    
+    func testClearWithSmallerMinimumCapacityDupesStorageIfTheresTwoBuffersSharingStorage() throws {
+        let alloc = ByteBufferAllocator()
+        let buf1 = alloc.buffer(capacity: 16)
+        var buf2 = buf1
+
+        var buf1PtrVal: UInt = 1
+        var buf2PtrVal: UInt = 2
+        
+        buf1.assign(to: &buf1PtrVal)
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertEqual(buf1PtrVal, buf2PtrVal)
+
+        buf2.clear(minimumCapacity: 8)
+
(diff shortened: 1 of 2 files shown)


Lukasa:
Hm, this appears to be a re-file of #1204. As I understand it you and @cweinberger worked on this together: do you two want to work out which of these should be preferred?

Keno42:
Oh, I was working on this with @kam800 and was not aware of #1204 .
I will talk with @cweinberger .

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -2134,3 +2204,11 @@ private func testReserveCapacityLarger_reallocHook(_ ptr: UnsafeMutableRawPointe
 private func testReserveCapacityLarger_memcpyHook(_ dst: UnsafeMutableRawPointer, _ src: UnsafeRawPointer, _ count: Int) -> Void {
     // No copying
 }
+
+extension ByteBuffer {
+    func assign(to pointer: inout UInt) {
```
weissi: I think we should give this function a better name and also return a `UInt` instead of taking it `inout`.

So I think we could use

```
func storagePointerIntegerValue() -> UInt {
   ...
}
```

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1227,6 +1227,76 @@ class ByteBufferTest: XCTestCase {
...
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertEqual(buf1PtrVal, buf2PtrVal)
+
+        buf2.clear(minimumCapacity: 32)
+
+        buf1.assign(to: &buf1PtrVal)
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertNotEqual(buf1PtrVal, buf2PtrVal)
```
weissi: could we add two asserts here:

```
XCTAssertLessThan(buf1.capacity, 32)
XCTAssertGreaterThanOrEqual(buf2.capacity, 32)
```

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1227,6 +1227,76 @@ class ByteBufferTest: XCTestCase {
...
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertEqual(buf1PtrVal, buf2PtrVal)
+
+        buf2.clear(minimumCapacity: 8)
+
+        buf1.assign(to: &buf1PtrVal)
+        buf2.assign(to: &buf2PtrVal)
+        
+        XCTAssertNotEqual(buf1PtrVal, buf2PtrVal)
```
weissi: Could we also asserts size asserts here?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1227,6 +1227,76 @@ class ByteBufferTest: XCTestCase {
...
+        var buf = alloc.buffer(capacity: 16)
+
+        var bufPtrValPre: UInt = 1
+        var bufPtrValPost: UInt = 2
+
+        buf.assign(to: &bufPtrValPre)
+        buf.clear(minimumCapacity: 32)
+        buf.assign(to: &bufPtrValPost)
+        
+        XCTAssertNotEqual(bufPtrValPre, bufPtrValPost)
```
weissi: same here

PR #1211, opened 2019-10-30
Cow-box WebSocketFrame.

marlimox:
### Motivation:

WebSocketFrameDecoder makes an allocation due to WebSocketFrame not fitting in an
existential. We should reduce this overhead.

### Modifications:

Moved data and extensionData into a cow-box and reorder fields to more efficiently
pack them.

### Result:

WebSocketFrame is now 14 bytes, down from 55.

### DIFF
--- a/Sources/NIOWebSocket/WebSocketFrame.swift
+++ b/Sources/NIOWebSocket/WebSocketFrame.swift
@@ -106,6 +106,19 @@ extension WebSocketMaskingKey: Collection {
 
 /// A structured representation of a single WebSocket frame.
 public struct WebSocketFrame {
+    private var _storage: WebSocketFrame._Storage
+
+    /// The masking key, if any.
+    ///
+    /// A masking key is used to prevent specific byte sequences from appearing in the network
+    /// stream. This is primarily used by entities like browsers, but should be used any time it
+    /// is possible for a malicious user to control the data that appears in a websocket stream.
+    ///
+    /// If this value is `nil`, and this frame was *received from* the network, the data in `data`
+    /// is not masked. If this value is `nil`, and this frame is being *sent to* the network, the
+    /// data in `data` will not be masked.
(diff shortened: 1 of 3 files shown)


Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -170,30 +183,39 @@ public struct WebSocketFrame {
...
+        }
+    }
 
     /// The extension data, if any.
     ///
     /// On frames received from the network, this data is not necessarily unmasked. This is to provide as much
     /// information as possible. If unmasked data is desired, either use the computed `unmaskedExtensionData` property to
     /// obtain it, or transform this data directly by calling `extensionData.unmask(maskKey)`.
-    public var extensionData: ByteBuffer? = nil
+    public var extensionData: ByteBuffer? = nil {
```
Lukasa:
```suggestion
    public var extensionData: ByteBuffer? {
```

Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -270,3 +292,22 @@ public struct WebSocketFrame {
...
+    fileprivate class _Storage {
+        var data: ByteBuffer
+        var extensionData: ByteBuffer?
+
+        fileprivate init(data: ByteBuffer, extensionData: ByteBuffer?) {
+            self.data = data
+            self.extensionData = extensionData
+        }
+
+        fileprivate init(copying original: Self) {
```
Lukasa:
```suggestion
        fileprivate init(copying original: WebSocketFrame._Storage) {
```

Lukasa:
Hey @marlimox, great work! We need to drop the threshold on the allocation counter test now, as it wants us to lower the watermark for allocations so we don’t regress it.

marlimox:
Ah yep.

marlimox:
Thank you for the help! 😄

PR #1214, opened 2019-11-01
Allow single test to be specified as an argument to run-nio-alloc-counter-tests.sh.

marlimox:
### Motivation:

`run-nio-alloc-counter-tests.sh` currently runs all tests. When writing or
debugging a test, you commonly want to run a single test. At the moment the
common practice is to edit the script to hardcode a test to run, which is
annoying and error-prone.

### Modifications:

Add an optional argument to the script to specify the test to run.

### Result:

`$ ./run-nio-alloc-counter-tests.sh`

runs all tests.

`$ ./run-nio-alloc-counter-tests.sh test_decode_1000_ws_frames.swift`

runs only test_decode_1000_ws_frames.swift.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
+++ b/IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
@@ -18,25 +18,28 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 tmp_dir="/tmp"
 
-function die() {
-    echo >&2 "ERROR: $*"
-    exit 1
-}
-
 while getopts "t:" opt; do
     case "$opt" in
         t)
             tmp_dir="$OPTARG"
             ;;
-        \?)
-            die "unknown option $opt"
+        *)
+            exit 1
             ;;
     esac
 done
+
+shift $((OPTIND-1))
+
+tests_to_run=("$here"/test_*.swift)
+
+if [ $# -gt 0 ]; then
+    tests_to_run=$@
+fi
 
 "$here/../../allocation-counter-tests-framework/run-allocation-counter.sh" \
     -p "$here/../../.." \
     -m NIO -m NIOHTTP1 -m NIOWebSocket \
     -s "$here/shared.swift" \
     -t "$tmp_dir" \
-    "$here"/test_*.swift
+    "${tests_to_run[@]}"
--- a/IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
+++ b/IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
@@ -33,7 +33,7 @@ shift $((OPTIND-1))
 
 tests_to_run=("$here"/test_*.swift)
 
-if [ $# -gt 0 ]; then
+if [[ $# -gt 0 ]]; then
     tests_to_run=$@
 fi
 


IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
```diff
@@ -18,25 +18,24 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 tmp_dir="/tmp"
 
-function die() {
```
weissi: could we leave this in so we get better error messages?

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
```diff
@@ -18,25 +18,24 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-    echo >&2 "ERROR: $*"
-    exit 1
-}
-
 while getopts "t:" opt; do
     case "$opt" in
         t)
             tmp_dir="$OPTARG"
             ;;
-        \?)
```
weissi: why this change?

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
```diff
@@ -18,25 +18,24 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-            die "unknown option $opt"
+        *)
+            exit 1
             ;;
     esac
 done
 
+shift $((OPTIND-1))
+
+test_pattern="${1:-test_*.swift}"
```
weissi: I think a better way to do this (which also doesn't break on newlines) is to start with

    tests_to_run=( "$here"/test_*.swift )

and if we only run one test, then we can override this with all the arguments that we were given:

    tests_to_run=$@

then, below we can use ...

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
```diff
@@ -18,25 +18,24 @@ here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+
 "$here/../../allocation-counter-tests-framework/run-allocation-counter.sh" \
     -p "$here/../../.." \
     -m NIO -m NIOHTTP1 -m NIOWebSocket \
     -s "$here/shared.swift" \
     -t "$tmp_dir" \
-    "$here"/test_*.swift
+    "$here"/$test_pattern
```
weissi:
```
"${tests_to_run[@]}"
```

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
marlimox: Yeah for sure. The reason I removed it is that it didn't seem like it was actually providing better error messages. For example, if you specify an unknown flag, you get the following output:

```
$ ./run-nio-alloc-counter-tests.sh -d
./run-nio-alloc-counter-tests.sh: illegal option -- d
ERROR: unknown option ?
```

The `?` in this case is `$opt`, which is always "?" because it doesn't know what the option is. All of the useful information is contained in the first line, which is the error message from `getopts`.

Removing the only call to `die` (and then the `die` function itself because it's no longer used) gives the following output, which seemed a bit cleaner and less confusing:

```
$ ./run-nio-alloc-counter-tests.sh -d
./run-nio-alloc-counter-tests.sh: illegal option -- d
```

Happy to revert back to the previous code if you prefer though!

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
marlimox: My thinking was most of the examples of `getopts` I've seen use `*` to handle the unknown case, as it's idiomatic with the default branch in a `case` block in general.

With `\?`, you have to remember that `getopts` will set the flag variable to `?` if it doesn't recognise a flag, so it's the unknown case you're handling with this branch. You also have to remember that you don't need a separate default branch for the `case`, because the variable will always be `?` if a flag is unknown.

Compared to a list of flags and then a default `*` case, that seemed less clear and harder to remember at first glance.

I don't feel strongly about it though, happy to revert it.

IntegrationTests/tests_04_performance/test_01_resources/run-nio-alloc-counter-tests.sh
marlimox: Yeah great suggestion. I only handled specifying a single test argument, as that was the use case I kept running into. But there's no reason not to allow multiple tests to be specified, and this is a good implementation for that.

I'll update!

marlimox: Ah I think I might have misunderstood the intention here. Your suggestion will still only run one test right? It will just treat all of the remaining arguments as a single test argument?

weissi:

PR #1219, opened 2019-11-04
Add extra WebSocket decoder benchmarks

emarashliev:
### Motivation:

We want to have some extra WebSocket decoder benchmarks, for a WebSocketFrame with a maskingKey, a WebSocketFrame with a 16-bit length and a WebSocketFrame with a 64-bit length.

### Modifications:

- Changed `WebSocketFrameDecoderBenchmark` to have ability for dynamic WebSocketFrame length through the initializer parameter `dataSize`.
- Added ability for `WebSocketFrameDecoderBenchmark` to benchmark a WebSocketFrame with a maskingKey by passing `withMaskKey` parameter to the initializer.
- Added new benchmarks in the `main.swift` for the WebSocketFrame decoding with 16-bit length, for the WebSocketFrame decoding with 64-bit length and for the maskingKey.

### Result:

 Better WebSocket decoder benchmarks.


Sources/NIOPerformanceTester/main.swift
```diff
@@ -739,4 +739,10 @@ try measureAndPrint(desc: "websocket_encode_50b_no_space_at_front_10k_frames",
 try measureAndPrint(desc: "websocket_encode_1kb_no_space_at_front_1k_frames",
                     benchmark: WebSocketFrameEncoderBenchmark(dataSize: 1024, runCount: 1_000, dataStrategy: .noSpaceAtFront, cowStrategy: .never))
 
-try measureAndPrint(desc: "websocket_decode", benchmark: WebSocketFrameDecoderBenchmark(dataSize: 1024, runCount: 100_000))
```
glbrntt: @Lukasa @weissi do we want to keep the test for `1024`?

glbrntt:

Sources/NIOPerformanceTester/main.swift
```diff
@@ -739,4 +739,10 @@ try measureAndPrint(desc: "websocket_encode_50b_no_space_at_front_10k_frames",
...
+try measureAndPrint(desc: "websocket_decode_16b", benchmark: WebSocketFrameDecoderBenchmark(dataSize: 16, runCount: 100_000))
+
+try measureAndPrint(desc: "websocket_decode_16b_with_а_masking_кey", benchmark: WebSocketFrameDecoderBenchmark(dataSize: 16, runCount: 100_000, withMaskKey: true))
+
+try measureAndPrint(desc: "websocket_decode_64b", benchmark: WebSocketFrameDecoderBenchmark(dataSize: 64, runCount: 10_000))
+
+try measureAndPrint(desc: "websocket_decode_64b_with_а_masking_кey", benchmark: WebSocketFrameDecoderBenchmark(dataSize: 64, runCount: 10_000, withMaskKey: true))
```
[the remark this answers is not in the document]
emarashliev: Ah this switching from the Cyrillic...

glbrntt:
@Lukasa can you kick the bot?

Lukasa:

Sources/NIOPerformanceTester/WebSocketFrameDecoderBenchmark.swift
```diff
@@ -15,36 +15,68 @@
...
         }
         return 1
     }
 }
+
+extension ByteBufferAllocator {
+    fileprivate func buffer(size: Int) -> ByteBuffer {
+        var data = self.buffer(capacity: size)
+        let bytes = size / 8
+        data.writeBytes([0x81] + Array(repeatElement(0, count: bytes - 1)))
```
Lukasa: So this is not quite right: the protocol doesn't really work this way.

You want something like this:

```swift
fileprivate func buffer(size: Int, maskingKey: WebSocketMaskingKey?) -> ByteBuffer {
    var data = self.buffer(capacity: size)

    // Calculate some information about the mask.
    let maskBitMask: UInt8 = maskingKey != nil ? 0x80 : 0x00

    // Time to add the extra bytes. To avoid checking this twice, we also start writing stuff out here.
    switch size {
    case 0...125:
        data.writeInteger(UInt8(0x81))
        data.writeInteger(UInt8(size) | maskBitMask)
    case 126...Int(UInt16.max):
        data.writeInteger(UInt8(0x81))
        data.writeInteger(UInt8(126) | maskBitMask)
        data.writeInteger(UInt16(size))
    case (Int(UInt16.max) + 1)...:
        data.writeInteger(UInt8(0x81))
        data.writeInteger(UInt8(127) | maskBitMask)
        data.writeInteger(UInt64(size))
    }

    if let maskingKey = maskingKey {
        data.writeBytes(maskingKey)
    }

    data.writeBytes(repeatElement(0, count: size))
    return data
}
```

You then want to have one test for each of the following cases:

- size = 125, masking key = nil
- size = 125, masking key = [0x80, 0x08, 0x10, 0x01]
- size = UInt16.max, maskingKey = nil
- size = UInt16.max, maskingKey = [0x80, 0x08, 0x10, 0x01]
- size = UInt16.max + 1, maskingKey = nil
- size = UInt16.max + 1, maskingKey = [0x80, 0x08, 0x10, 0x01]

Sources/NIOPerformanceTester/WebSocketFrameDecoderBenchmark.swift
```diff
@@ -15,36 +15,86 @@
...
+            self.data.webSocketUnmask(maskingKey)
             try self.channel.writeInbound(self.data)
             let _: WebSocketFrame? =  try self.channel.readInbound()
         }
         return 1
     }
 }
+
+extension ByteBufferAllocator {
+    fileprivate func buffer(size: Int, maskingKey: WebSocketMaskingKey?) -> ByteBuffer {
```
Lukasa: Can I propose we call this method `webSocketFrame`?

Sources/NIOPerformanceTester/WebSocketFrameDecoderBenchmark.swift
```diff
@@ -15,36 +15,86 @@
...
+        for _ in 0..<self.runCount {
+            try self.channel.writeInbound(self.data)
+            let _: WebSocketFrame? =  try self.channel.readInbound()
+        }
+        return 1
+    }
+
+    private func run(with maskingKey: WebSocketMaskingKey) throws -> Int {
+        for _ in 0..<self.runCount {
+            self.data.webSocketUnmask(maskingKey)
```
Lukasa: We actually don't need to call this function: the frame decoder should correctly handle this for us. So we don't need the two separate `run` functions, they can have the same implementation.

Lukasa:

Lukasa:

Lukasa:

Lukasa:
Nice one @emarashliev! Thanks!

tomerd:
love all the new benchmarks. good stuff. <3

PR #1220, opened 2019-11-04
Add inlinability to ByteBuffer getters.

Lukasa:
Motivation:

In complex ByteBuffer parsing code a surprising amount of cost comes
from retain/release operations that are emitted by the compiler around
calls to readerIndex/writerIndex, readableBytes/writableBytes, and
ByteBufferSlice._lowerBound. These are all straightforward computed
properties based on usableFromInline operations, and so they can be made
inlinable.

Modifications:

Joannis' sample benchmark runtime is halved by eliminating
retains/releases.

Result:

Better, faster parsing code.


Lukasa:

PR #1224, opened 2019-11-07
EventLoopFuture: save one allocation per future

weissi:
Motivation:

Every EventLoopFuture was allocating an atomic boolean (which needs to
heap allocate because Swift) literally just to make the leakage checker
work. But the leakage checker is only active in debug mode so there's no
reason to be so wasteful. Furthermore, NIO doesn't require this `isFulfilled`
property in normal operation.

Modifications:

Remove the `isFulfilled` atomic which removes the allocation.
 
Result:

Fewer allocations

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -1380,7 +1380,6 @@ public final class ChannelHandlerContext: ChannelInvoker {
 
     fileprivate func invokeRegister(promise: EventLoopPromise<Void>?) {
         self.eventLoop.assertInEventLoop()
-        assert(promise.map { !$0.futureResult.isFulfilled } ?? true, "Promise \(promise!) already fulfilled")
 
         if let outboundHandler = self.outboundHandler {
             outboundHandler.register(context: self, promise: promise)
@@ -1391,7 +1390,6 @@ public final class ChannelHandlerContext: ChannelInvoker {
 
    fileprivate func invokeBind(to address: SocketAddress, promise: EventLoopPromise<Void>?) {
         self.eventLoop.assertInEventLoop()
-        assert(promise.map { !$0.futureResult.isFulfilled } ?? true, "Promise \(promise!) already fulfilled")
 
         if let outboundHandler = self.outboundHandler {
             outboundHandler.bind(context: self, to: address, promise: promise)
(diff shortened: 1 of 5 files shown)


weissi:

weissi:
CC @Davidde94 can't add you yet as a reviewer but please take a look

tomerd:
very nice

Sources/NIO/EventLoopFuture.swift
```diff
@@ -401,7 +404,7 @@ public final class EventLoopFuture<Value> {
     internal init(_eventLoop eventLoop: EventLoop, value: Result<Value, Error>?, file: StaticString, line: UInt) {
         self.eventLoop = eventLoop
         self._value = value
-        self._isFulfilled = UnsafeEmbeddedAtomic(value: value != nil)
+        self._isFulfilled = _isDebugAssertConfiguration() ? UnsafeEmbeddedAtomic(value: value != nil) : nil
```
Lukasa: This should go in a `debugOnly` block.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -374,20 +374,23 @@ public final class EventLoopFuture<Value> {
...
 
     /// The `EventLoop` which is tied to the `EventLoopFuture` and is used to notify all registered callbacks.
     public let eventLoop: EventLoop
 
     /// Whether this `EventLoopFuture` has been fulfilled. This is a thread-safe
     /// computed-property.
     internal var isFulfilled: Bool {
-        return self._isFulfilled.load()
+        precondition(_isDebugAssertConfiguration())
+        return self._isFulfilled!.load()
```
Lukasa: One note here: this makes it impossible to run our unit tests in release mode. I don't care about this, but others might. We should document the fact that this operations crashes when run in release mode.

weissi: ah shit, that's right. I'll remove this.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1380,7 +1380,6 @@ public final class ChannelHandlerContext: ChannelInvoker {
 
     fileprivate func invokeRegister(promise: EventLoopPromise<Void>?) {
         self.eventLoop.assertInEventLoop()
-        assert(promise.map { !$0.futureResult.isFulfilled } ?? true, "Promise \(promise!) already fulfilled")
```
weissi: these totally didn't make any sense anyway because in all other places, we allow pre-fulfilled futures. This could also cause racy crashes.

Sources/NIO/EventLoopFuture.swift
weissi: @Lukasa this wouldn't have worked because it's a `let` property and needs to be initialised before `self` is published which it does when you capture it in the `debugOnly` closure.

But it doesn't matter anymore, I killed the whole atomic now, in debug and release as we totally don't need it.

Sources/NIO/EventLoopFuture.swift
weissi: killed this completely.

weissi:

PR #1229, opened 2019-11-10
NIOPerformanceTester: more precision and resiliency

weissi:
Motivation:

Previously, NIOPerformanceTester would use Foundation.Date twice to
calculate the time spend in a given function. This is less precise as it
can be but more importantly, it's not resilient against the user
changing the clock or the computer going to sleep.

Modifications:

Use DispatchTime.now().uptimeNanoseconds

Result:

More precise and resilient benchmark measurements.


weissi:

PR #1230, opened 2019-11-11
Disallow zero EventLoop threads

### LINKED ISSUE #1223: Fatal error using an ELG with zero threads
While developing changes to an HTTP client library built on top of `async-http-client`, I wanted to be able to accept a user-supplied `EventLoopGroup`, and in my tests, to assert that I have correctly supplied the user's group to the `HTTPClient` instead of a default one.

`MultiThreadedEventLoopGroup` is not `Equatable`, so I cannot compare the value directly.  Instead, I thought I'd try creating one with zero threads, with the naive assumption that attempts to use it would time out.

However, although I can create such a group, attempts to use it cause a fatal error.  Having read into the code a little, I can see that it isn't reasonable to expect blocking behaviour, but I think the failure ought to be at the point of creation rather than when the group is first used.

### Expected behavior
_[what you expected to happen]_
An (ideally catchable) failure at the point of creation.

### Actual behavior
_[what actually happened]_
`Fatal error: Division by zero in remainder operation` when a request is attempted ([`HTTPClient.execute()`](https://github.com/swift-server/async-http-client/blob/1.0.0/Sources/AsyncHTTPClient/HTTPClient.swift#L191) calls `eventLoopGroup.next()`).

### Steps to reproduce

1. `let myELG = MultiThreadedEventLoopGroup(numberOfThreads: 0)`
2. `let eventLoop = myELG.next()` 

2.9.0

Swift version 5.1 (swift-5.1-RELEASE)
Linux needletail 4.15.0-66-generic #75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

richardalow:
Motivation:

Zero EventLoop threads makes no sense and leads to a crash (fixes #1223)

Modifications:

Add a precondition

Result:

Initializing EventLoop with zero threads will fail precondition rather than crash later.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -1130,6 +1130,7 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
     /// - arguments:
     ///     - numberOfThreads: The number of `Threads` to use.
     public convenience init(numberOfThreads: Int) {
+        precondition(numberOfThreads > 0, "numberOfThreads must be positive")
         let initializers: [ThreadInitializer] = Array(repeating: { _ in }, count: numberOfThreads)
         self.init(threadInitializers: initializers)
     }


Lukasa:

Lukasa:

PR #1233, opened 2019-11-11
Update ByteBuffer docs to advance write index on set operations

### LINKED ISSUE #1232: ByteBuffer documentation is out of date for set operations
### Expected behavior
ByteBuffer documentation for set operations no longer works e.g.

```
buf.setString("Hello World", at: 0)
let helloWorld = buf.getString(at: 0, length: 11)
// expected helloWorld == "Hello World"
```

### Actual behavior
helloWorld is nil.

Documentation should be updated to advance the writer index.

richardalow:
Motivation:

Docs give unexpected results since behavior was changed in NIO 2
to only see readable bytes. Fixes #1232.

Modifications:

Documentation updated.

Result:

Documentation gives expected results with no errors.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -133,10 +133,12 @@ public struct ByteBufferAllocator {
 ///
 ///     var buf = ...
 ///     buf.setString("Hello World", at: 0)
+///     buf.moveWriterIndex(to: 11)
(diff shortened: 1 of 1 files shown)


richardalow:
@glbrntt I believe the other docs are all correct

glbrntt:
> @glbrntt I believe the other docs are all correct

Awesome -- thanks for checking!

PR #1235, opened 2019-11-12
don't use variable initialisation outside of `init`

weissi:
Motivation:

The Swift compiler actively tries to make our live hard and puts all
initialisations of variables that are initialised using the shorthand
syntax

    @usableFromInline var myVar: MyVarType = .init()

into a _non-inlinable_ function

     variable initialization expression of MyModule.MyType.myVar : MyVarType

This means that the following pattern

    public struct MyType {
        @usableFromInline var myVar: MyVarType = .init()

        @inlinable public init() {
        }
    }

leads to code for the `init` that looks like

    CALL variable_initialization_expression_...

instead of also inlining the the code for the variable initialiser.

Modifications:

Don't use `@usableFromInline var myVar: MyVarType = ...` anywhere,
instead put it explicitly into the @inlinable `init`s which makes
actually inlinable.
I even removed that syntax from places where it doesn't matter
(eg. non-@inlinable inits) for consistency and in case someone were to
make the init inlinable later.

Result:

Minor performance wins working around https://bugs.swift.org/browse/SR-11768

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -59,9 +59,9 @@ public final class ServerBootstrap {
     private var serverChannelInit: ((Channel) -> EventLoopFuture<Void>)?
     private var childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
     @usableFromInline
-    internal var _serverChannelOptions = ChannelOptions.Storage()
+    internal var _serverChannelOptions: ChannelOptions.Storage
     @usableFromInline
-    internal var _childChannelOptions = ChannelOptions.Storage()
+    internal var _childChannelOptions: ChannelOptions.Storage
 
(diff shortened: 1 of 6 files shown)


Lukasa:

slavapestov:
Try declaring the structs as `@frozen` instead; this will cause SILGen to generate inlinable initializers.

weissi:
@slavapestov sadly we have to support Swift 5.0 and

```
-1- johannes:/tmp
$ jw-swift-5.0 swiftc -version
Apple Swift version 5.0 (swift-5.0-RELEASE)
Target: x86_64-apple-darwin19.0.0
johannes:/tmp
$ jw-swift-5.0 swiftc t4.swift 
t4.swift:1:2: error: unknown attribute 'frozen'
@frozen
 ^
```

PR #1244, opened 2019-11-13
perf tests: switch printed metric to mean

weissi:
Motivation:

We should print the same metric that we compare with.

Modifications:

Switch the printed metric to the one selected in `::METRIC`.

Result:

More correct printouts.


tomerd:
probably need to do the same for h2 and ssl?

PR #1246, opened 2019-11-13
HTTP1TestServer: close accepted channel on `stop`

weissi:
Motivation:

The HTTP1TestServer forgot to actually close the accepted channel on
`stop`.

Modifications:

Close the accepted Channel on stop.

Result:

Fewer bugs.

### DIFF
--- a/Sources/NIOTestUtils/NIOHTTP1TestServer.swift
+++ b/Sources/NIOTestUtils/NIOHTTP1TestServer.swift
@@ -276,6 +276,8 @@ extension NIOHTTP1TestServer {
                 guard self.inboundBuffer.isEmpty else {
                     throw NonEmptyInboundBufferOnStop()
                 }
+            }.always { _ in
+                self.currentClientChannel?.close(promise: nil)
             }
         }.wait()
     }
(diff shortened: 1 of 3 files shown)


weissi:
CC @mariosangiorgio

Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -271,6 +271,7 @@ extension NIOHTTP1TestServer {
             case .stopped:
                 preconditionFailure("double stopped NIOHTTP1TestServer")
             }
+            self.currentClientChannel?.close(promise: nil)
```
Lukasa: Should this close be chained off the server channel? Otherwise there’s a race in which the current client channel may be closed, the server channel may then accept another, and then that channel will not be closed directly. Probably we should confirm we’ve closed the firehose _first_, and then close the remaining connection.

weissi: @Lukasa great idea, done!

weissi: definitely easier to reason about now.

PR #1252, opened 2019-11-18
use Optional<T> instead of T? to workaround SR-11777

weissi:
Motivation:

In Swift, writing

    var something: T?

    init() {
        self.something = someValue
    }

means that the compiler will first set `self.something` to `nil` and
then in the init override it with `self.someValue`
(https://bugs.swift.org/browse/SR-11777). Unfortunately, because of
https://bugs.swift.org/browse/SR-11768 , stored property initialisation
cannot be made `@inlinable` (short of using `@frozen` which isn't
available in Swift 5.0).

The combination of SR-11768 and SR-11777 leads to `var something: T?`
having much worse code than `var something: Optional<T>` iff the `init`
is `public` and `@inlinable`.

Modifications:

Change all `var something: T?` to `var something: Optional<T>`

Result:

Faster code, sad NIO developers.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -12,6 +12,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+/// The type of all `channelInitializer` callbacks.
+internal typealias ChannelInitializerCallback = (Channel) -> EventLoopFuture<Void>
+
 /// A `ServerBootstrap` is an easy way to bootstrap a `ServerSocketChannel` when creating network servers.
 ///
 /// Example:
@@ -56,8 +59,8 @@ public final class ServerBootstrap {
 
     private let group: EventLoopGroup
     private let childGroup: EventLoopGroup
-    private var serverChannelInit: ((Channel) -> EventLoopFuture<Void>)?
-    private var childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
+    private var serverChannelInit: Optional<ChannelInitializerCallback>
+    private var childChannelInit: Optional<ChannelInitializerCallback>
     @usableFromInline
     internal var _serverChannelOptions: ChannelOptions.Storage
     @usableFromInline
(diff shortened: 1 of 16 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -354,11 +356,11 @@ private extension Channel {
 public final class ClientBootstrap {
 
     private let group: EventLoopGroup
-    private var channelInitializer: ((Channel) -> EventLoopFuture<Void>)?
+    private var channelInitializer: Optional<(Channel) -> EventLoopFuture<Void>>
```
Davidde94: Nit, maybe out of scope, now that this line is a little more "complex" a typealias might be nice

Sources/NIO/EventLoop.swift
```diff
@@ -58,14 +58,15 @@ public final class RepeatedTask {
     private let delay: TimeAmount
     private let eventLoop: EventLoop
     private let cancellationPromise: EventLoopPromise<Void>?
-    private var scheduled: Scheduled<EventLoopFuture<Void>>?
-    private var task: ((RepeatedTask) -> EventLoopFuture<Void>)?
+    private var scheduled: Optional<Scheduled<EventLoopFuture<Void>>>
+    private var task: Optional<(RepeatedTask) -> EventLoopFuture<Void>>
```
Davidde94: Nit: See previous comment

Sources/NIO/Bootstrap.swift
weissi: @Davidde94 hmm, what name would you propose for a typealias for `(Channel) -> EventLoopFuture<Void>`?

Davidde94: `ChannelInitializer`?

weissi: doesn't that make it harder to see what that is? CC @Lukasa I'm open to making it `Optional<ChannelInitializer>` but I think that's more opaque than `Optional<(Channel) -> ELF<Void>>` but maybe it's me?

Lukasa: I don't love `ChannelInitializer` as a typealias: `ChannelInitializerCallback` might be better.

Davidde94: Either sound good to me - note this was a readability nit, I think the good is good to go regardless :)

weissi: @Davidde94 / @Lukasa fixed.

Sources/NIO/EventLoop.swift
weissi: done too

PR #1253, opened 2019-11-19
Create a ClientBootstrap protocol

Yasumoto:
Create a simple protocol for Client-side bootstrap implementations to conform to.

### Motivation:

As described in https://github.com/apple/swift-nio/issues/674, we want to make it easier to switch between `ClientBootstrap` and `NIOTSConnectionBootstrap`.

This is one approach that I'm intrigued by, but not necessarily overly fond of. We're still allowing folks to use the wrong type of `EventLoopGroup` with the bootstrap they choose. However, this simplifies the ability to use both types of bootstrap objects, and at the least serves as a strawman proposal.

### Modifications:

Created a `ClientTransportBootstrap` protocol. I think it makes sense to separate out a client vs. server since the usecases and surface area are so different.

### Result:

Users will be able to choose between either `ClientBootstrap` or `NIOTSConnectionBootstrap` much more easily.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -330,6 +330,40 @@ private extension Channel {
     }
 }
 
+/// `NIOTCPClientBootstrap` is implemented by various underlying transport mechanisms. Typically,
+/// this will be the BSD Sockets API implemented by `ClientBootstrap`.
+public protocol NIOTCPClientBootstrap {
+    /// Initialize the connected `Channel` with `initializer`. The most common task in initializer is to add
+    /// `ChannelHandler`s to the `ChannelPipeline`.
+    ///
+    /// - parameters:
+    ///     - handler: A closure that initializes the provided `Channel`.
+    func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> Self
+
+    /// Specifies a `ChannelOption` to be applied to the `Channel`.
+    ///
+    /// - parameters:
+    ///     - option: The option to be applied.
+    ///     - value: The value for the option.
+    @inlinable
+    func channelOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> Self
+
+    /// Specifies a timeout to apply to a connection attempt.
+    //
+    /// - parameters:
+    ///     - timeout: The timeout that will apply to the connection attempt.
+    func connectTimeout(_ timeout: TimeAmount) -> Self
+
+    /// Specify the `host` and `port` to connect to for the `Channel` that will be established.
+    ///
+    /// - parameters:
+    ///     - host: The host to connect to.
+    ///     - port: The port to connect to.
+    /// - returns: An `EventLoopFuture<Channel>` to deliver the `Channel` when connected.
+    func connect(host: String, port: Int) -> EventLoopFuture<Channel>
+}
+
+
 /// A `ClientBootstrap` is an easy way to bootstrap a `SocketChannel` when creating network clients.
 ///
 /// Usually you re-use a `ClientBootstrap` once you set it up and called `connect` multiple times on it.
@@ -356,7 +390,7 @@ private extension Channel {
 /// ```
 ///
 /// The connected `SocketChannel` will operate on `ByteBuffer` as inbound and on `IOData` as outbound messages.
-public final class ClientBootstrap {
+public final class ClientBootstrap: NIOTCPClientBootstrap {
 
     private let group: EventLoopGroup
     private var channelInitializer: Optional<ChannelInitializerCallback>
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -332,7 +332,7 @@ private extension Channel {
 
 /// `NIOTCPClientBootstrap` is implemented by various underlying transport mechanisms. Typically,
 /// this will be the BSD Sockets API implemented by `ClientBootstrap`.
-public protocol NIOTCPClientBootstrap {
+internal protocol NIOTCPClientBootstrap {
     /// Initialize the connected `Channel` with `initializer`. The most common task in initializer is to add
     /// `ChannelHandler`s to the `ChannelPipeline`.
     ///
--- a/Tests/NIOTests/BootstrapTest.swift
+++ b/Tests/NIOTests/BootstrapTest.swift
@@ -134,11 +134,11 @@ class BootstrapTest: XCTestCase {
             .channelInitializer { channel in
                 XCTAssert(channel.eventLoop.inEventLoop)
                 return self.freshEventLoop().makeSucceededFuture(())
-        }
-        .bind(host: "127.0.0.1", port: 0)
-        .wait()
-        .close()
-        .wait())
+            }
+            .bind(host: "127.0.0.1", port: 0)
+            .wait()
+            .close()
+            .wait())
     }
 
     func testPreConnectedClientSocketToleratesFuturesFromDifferentEventLoopsReturnedInInitializers() throws {
@@ -204,5 +204,46 @@ class BootstrapTest: XCTestCase {
         }
         XCTAssertNoThrow(try childChannelDone.futureResult.wait())
         XCTAssertNoThrow(try serverChannelDone.futureResult.wait())
+    }
+
+    func testTCPClientBootstrapAllowsConformanceCorrectly() throws {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        func restrictBootstrapType(clientBootstrap: NIOTCPClientBootstrap) throws {
+            let serverAcceptedChannelPromise = group.next().makePromise(of: Channel.self)
+            let serverChannel = try assertNoThrowWithValue(ServerBootstrap(group: group)
+                .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+                .childChannelInitializer { channel in
+                    serverAcceptedChannelPromise.succeed(channel)
+                    return channel.eventLoop.makeSucceededFuture(())
+            }.bind(host: "127.0.0.1", port: 0).wait())
+
+            let clientChannel = try assertNoThrowWithValue(clientBootstrap
+                .channelInitializer({ (channel: Channel) in channel.eventLoop.makeSucceededFuture(()) })
+                .connect(host: "127.0.0.1", port: serverChannel.localAddress!.port!).wait())
+
+            var buffer = clientChannel.allocator.buffer(capacity: 1)
+            buffer.writeString("a")
+            try clientChannel.writeAndFlush(NIOAny(buffer)).wait()
+
+            let serverAcceptedChannel = try serverAcceptedChannelPromise.futureResult.wait()
+
+            // Start shutting stuff down.
+            XCTAssertNoThrow(try clientChannel.close().wait())
+
+            // Wait for the close promises. These fire last.
+            XCTAssertNoThrow(try EventLoopFuture.andAllSucceed([clientChannel.closeFuture,
+                                                                serverAcceptedChannel.closeFuture],
+                                                                on: group.next()).wait())
+        }
+
+        if let bootstrap = group.makeTCPClientBootstrap() {
+            try restrictBootstrapType(clientBootstrap: bootstrap)
+        } else {
+            XCTFail("Did not generate a valid Bootstrap")
+        }
     }
 }
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -60,6 +60,11 @@ public final class EmbeddedEventLoop: EventLoop {
     /// Initialize a new `EmbeddedEventLoop`.
     public init() { }
 
+    /// Provide a valid `ClientBootstrap` to setup an `EmbeddedEventLoop`.
+    public func makeTCPClientBootstrap() -> NIOTCPClientBootstrap {
+        return ClientBootstrap(group: self)
+    }
+
     /// - see: `EventLoop.scheduleTask(deadline:_:)`
     @discardableResult
     public func scheduleTask<T>(deadline: NIODeadline, _ task: @escaping () throws -> T) -> Scheduled<T> {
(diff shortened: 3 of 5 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -325,6 +325,47 @@ private extension Channel {
...
 
+/// `ClientTransportBootstrap` is implemented by various underlying transport mechanisms. Typically,
+/// this will be the BSD Sockets API implemented by `ClientBootstrap`.
+public protocol ClientTransportBootstrap {
+
+    /// Create a `Bootstrap` on the `EventLoopGroup` `group`.
+    ///
+    /// - parameters:
+    ///     - group: The `EventLoopGroup` to use.
+    init(group: EventLoopGroup)
```
Lukasa: In an ideal world we'd have a proper type here, but I am really struggling to come up with a way to do it. I'm going to think on it some more and maybe an idea will come to me.

Sources/NIO/Bootstrap.swift
```diff
@@ -325,6 +325,47 @@ private extension Channel {
     }
 }
 
+/// `ClientTransportBootstrap` is implemented by various underlying transport mechanisms. Typically,
+/// this will be the BSD Sockets API implemented by `ClientBootstrap`.
+public protocol ClientTransportBootstrap {
```
Lukasa: This would need to be called `NIOClientTransportBootstrap` to meet our API naming guidelines.

Sources/NIO/Bootstrap.swift
Yasumoto: Agreed, please take your time 🙇‍♂️

I was wondering if we could use an associated type and force something, but I don't think that's correct. I'm hoping to figure out how to save folks from accidentally using a `MultiThreadedEventLoopGroup` with `NIOTSConnectionBootstrap. 🤔

Tests/NIOTests/BootstrapTest.swift
```diff
@@ -205,4 +205,38 @@ class BootstrapTest: XCTestCase {
         XCTAssertNoThrow(try childChannelDone.futureResult.wait())
         XCTAssertNoThrow(try serverChannelDone.futureResult.wait())
     }
+
+    func testClientTransportBootstrapAllowsConformanceCorrectly() throws {
```
Yasumoto: I was inspired by `ChannelTests.testBasicLifecycle`, happy to add more/different test cases if anyone has other ideas.

Tests/NIOTests/BootstrapTest.swift
```diff
@@ -205,4 +205,38 @@ class BootstrapTest: XCTestCase {
...
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+        let clientBootstrap: NIOClientTransportBootstrap = ClientBootstrap(group: group)
```
Lukasa: It might be nicer to write a function that takes the bootstrap generically and write the test in that function. That way it's much harder to accidentally break this test by changing only one line in a seemingly-innocuous way.

Sources/NIO/Bootstrap.swift
weissi: I think we should remove the `init` here and instead add a `public static func makeTCPClientBootstrap() -> NIOClientTransportBootstrap` to `EventLoopGroup`.

Sure, for API compatibility we need to implement a default implementation for this which would be

```
extension EventLoopGroup {
    public static func makeTCPClientBootstrap() -> NIOClientTransportBootstrap {
        struct CannotBootstrap: NIOClientTransportBootstrap {
            // fail everything with a good error
        }
        return CannotBootstrap()
    }
}
```

and in NIO 3 we can remove that. WDYT @Lukasa / @Yasumoto ?

weissi: The reason for removing the `init` is because there's less room for error. The new way would basically be to choose your favourite `EventLoopGroup` and from there you could peel off a TCP client bootstrap.

Sources/NIO/Bootstrap.swift
```diff
@@ -325,6 +325,47 @@ private extension Channel {
     }
 }
 
+/// `NIOClientTransportBootstrap` is implemented by various underlying transport mechanisms. Typically,
+/// this will be the BSD Sockets API implemented by `ClientBootstrap`.
+public protocol NIOClientTransportBootstrap {
```
weissi: naming nit: maybe `NIOTCPClientBootstrap`? Because this bootstrap will only work for TCP stuff. But not strong on this one, @Lukasa ?

Sources/NIO/Bootstrap.swift
weissi: We would then just implement this for `SelectableEventLoop`, `MultiThreadedEventLoopGroup`, `EmbeddedEventLoop`, and `NIOTSEventLoop(Group)` which is 99.999% of the actual use-cases so nobody will ever see the error.

Lukasa: I'm definitely open to doing that. An important note would be that we would _need_ to ensure that libraries don't ask for an event loop group, and instead ask for a bootstrap, when they don't care about which event loop is driving the bootstrap. The reason for that is that if you wanted to, for example, write a HTTP/2 client and server bootstrap that bootstraps stream channels, you wouldn't be hanging that off an event loop group.

It may be better to have a separate protocol to which all `EventLoopGroup`s conform that does this job. That would allow other types, such as `HTTP2StreamMultiplexer`, to expose conformances to that protocol as well. How does that seem?

Sources/NIO/Bootstrap.swift
Lukasa: I might prefer NIOStreamClientBootstrap, to enable the possibility of encapsulating other things.

Sources/NIO/Bootstrap.swift
weissi: @Lukasa that sounds good but we would need to remove the `connect(...)` then, or would just have it fail for the H2 multiplexer use-case?

Lukasa: That's a good question. In an ideal world `connect` would simply create a new stream, but the requirement that connect take a complex data type causes us some limitations. The biggest limitation there is that you need to put a host/port combo in the H2 multiplexer, which clearly doesn't work.

I don't know how to design our way around that either. In principle I guess we could use the host/port for authority information? But it raises some real questions about what all of this means that I don't think we can address today, so maybe we should drop that idea for now.

Lukasa: Regardless, I think having a separate protocol is still the right choice, as it allows more flexibility for future evolution and implementations than tying the protocol directly to ELGs.

weissi: @Lukasa ok, new idea

```
public struct NIOClientStreamBootstrap<ConnectType> {
     // all the methods

    public init<EL: EventLoop, ELG: EventLoopGroup>(group: ELG, eventLoopType: EL.Type = EL.self) { ... }

    public func connect(to: ConnectType)
}
```

plus

```
public protocol NIOClientStreamBootstrapper {
    associatedtype ConnectType

    // all the methods
}
```

```
extension ClientBootstrap: NIOClientStreamBootstrapper {
    typealias ConnectType = SocketAddress
}

extension HTTP2Multiplexer: NIOClientStreamBootstrapper {
    typealias ConnectType = H2StreamID
}
```

weissi: that is more along the lines I was thinking before this PR came up but I never got to try it out to see if everything works... I'm happy to draft something more concrete up?

weissi: the nice thing is that with `ConnectType`, we can also make `NIOPipeBootstrap` conform to it :)

Lukasa: I think we'd need the connect type of `ClientBootstrap` to be an enum to allow for happy-eyeballing our way to victory, but that idea could work. The biggest downside is only that I suspect users will often constrain that associated type in a way that is somewhat troublesome. That was the advantage of requiring `host:port:` in the original implementation, it pushes you down the maximal feature route.

weissi: @Lukasa what about

```
extension NIOClientStreamBootstrap where ConnectType == SocketAddress {
    public func connect(host: String, port: Int) {
        // happy eyeballs
    }
}
```

do you think this may work?

Lukasa: It _does_, but that doesn't apply to NIOTransportServices.

Again, we can protocol our way out of this by having _another_ protocol that covers "can connect with host and port", or a type for "represents a host and port".

weissi: (but besides that, I'd also be happy to make `ConnectType = SomeEnumThatCoversWhatWeHaveToday`

Lukasa: At a certain point though I wonder if we're over-solving this. For now it may be sufficient to exclude everything but TCP clients and servers and to solve those cases today. We can try to generalise further later.

weissi: @Lukasa agreed. We could just fly with basically what's there and name it accordingly leaving the more general name available for the next version?

weissi: I would however remove the `init` and make users peel off the bootstrap from the ELG maybe?

Lukasa: I'm fine with users peeling off from the ELG for now. Let's go down that road and worry about the best possible ergonomics later.

weissi: sounds good to me! @Yasumoto are you okay with this and does this make sense?

Yasumoto:
### Naming

> name it accordingly leaving the more general name available for the next version?

I assumed this meant go with `NIOTCPStreamClientBootstrap` vs. `NIOStreamClientBootstrap` (or something else) so it was very specific, if that's not right lemme know.

### Default Implementation

I'm not sure the proposal for a default `CannotBootstrap` is correct, since none of the methods throw. Here's what I came up with:

```swift
extension EventLoopGroup {
    public static func makeTCPClientBootstrap() -> NIOTCPStreamBootstrap {
        struct CannotBootstrap: NIOTCPStreamBootstrap {
            func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> CannotBootstrap {
                return self
            }

            func channelOption<Option>(_ option: Option, value: Option.Value) -> CannotBootstrap where Option : ChannelOption {
                return self
            }

            func connectTimeout(_ timeout: TimeAmount) -> CannotBootstrap {
                return self
            }

            func connect(host: String, port: Int) -> EventLoopFuture<Channel> {
               /* return ???? */
            }
        }
        return CannotBootstrap()
    }
}
```

### Static

I also thought we do not want this to be `static`, as we want to provide the actual ELG to the bootstrap.

### Where to implement `makeTCPClientBootstrap`

I assumed we only want this on `MultiThreadedEventLoopGroup` and `NIOTSEventLoop`, but should we also add it to `SelectableEventLoop`, `EmbeddedEventLoop`, and `NIOTSEventLoop`?

#### Merge commit

🤦‍♂ I'll squash after we get the questions sorted a bit more, I rebased instead of merged.

weissi:
Thanks for looking into this!!

> ### Naming
> > name it accordingly leaving the more general name available for the next version?
> 
> I assumed this meant go with `NIOTCPStreamClientBootstrap` vs. `NIOStreamClientBootstrap` (or something else) so it was very specific, if that's not right lemme know.

Right now, it only supports TCP so I think it should be `NIOTCPClientBootstrap` or something like that. @Lukasa ?

 
> ### Default Implementation
> I'm not sure the proposal for a default `CannotBootstrap` is correct, since none of the methods throw. Here's what I came up with:
> 
> ```swift
> extension EventLoopGroup {
>     public static func makeTCPClientBootstrap() -> NIOTCPStreamBootstrap {
>         struct CannotBootstrap: NIOTCPStreamBootstrap {
>             func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> CannotBootstrap {
>                 return self

I'd return a failed future here

>             }
> 
>             func channelOption<Option>(_ option: Option, value: Option.Value) -> CannotBootstrap where Option : ChannelOption {
>                 return self
>             }
> 
>             func connectTimeout(_ timeout: TimeAmount) -> CannotBootstrap {
>                 return self
>             }
> 
>             func connect(host: String, port: Int) -> EventLoopFuture<Channel> {
>                /* return ???? */

return self.next().makeFailedFuture(CannotBootstrapError(#"Your EventLoopGroup of type \(type(of: self)) is not compatible with "NIOTCPStreamBootstrap", please implement "public func makeTCPClientBootstrap() -> NIOTCPStreamBootstrap" "#))


>             }
>         }
>         return CannotBootstrap()
>     }
> }
> ```
> 
> ### Static
> I also thought we do not want this to be `static`, as we want to provide the actual ELG to the bootstrap.

Agreed, shouldn't be static.

 
> ### Where to implement `makeTCPClientBootstrap`
> I assumed we only want this on `MultiThreadedEventLoopGroup` and `NIOTSEventLoop`, but should we also add it to `SelectableEventLoop`, `EmbeddedEventLoop`, and `NIOTSEventLoop`?

First, declare it in the `EventLoopGroup` protocol, then, add a default implementation using `extension EventLoopGroup` to retain API compatibility.


> #### Merge commit
> 🤦‍♂ I'll squash after we get the questions sorted a bit more, I rebased instead of merged.

We can also do that in the github UI at the very end.

Lukasa:
`NIOTCPClientBootstrap` sounds fine.

kylebrowning:
How is this going? Im desperately looking forward to this for https://github.com/apple/swift-nio-transport-services

Yasumoto:
> How is this going? Im desperately looking forward to this for https://github.com/apple/swift-nio-transport-services

Thanks for the nudge, @kylebrowning ! 😅 

Went through the feedback again, and I believe I've incorporated it, so this should be ready for a fresh look.

weissi:
@Yasumoto I'm happy taking this if we make it `internal` for the time being, sorry about that.

The reason we can't use this as-is is because it doesn't support TLS at all. That would mean that we have a very easy to use and compatible version for plain text (which nobody should ever use) and if you need TLS, you need to go back to the old, `#if canImport(Network)...` way. I have some ideas how we could solve this but it's too much for this PR. So if you'd be happy with that, we could start with what we have here but all `internal`, then I add what's necessary for TLS and then we make it all `public` in one go. CC @Lukasa does that sound like a plan?

Lukasa:
Works for me!

Yasumoto:
Awesome, totally a great save. Thanks for helping get this over the finish line!

slam dunk

Lukasa:

PR #1259, opened 2019-11-20
don't use `public extension`

weissi:
Motivation:

`public extension`s are very dangerous because the make it so easy to
slip in public API by accident.

Modifications:

Don't use `public extension`.

Result:

Fewer chances to accidentally make stuff public.


Davidde94:
LGTM

PR #1260, opened 2019-11-20
Define convenience socket options

Davidde94:
Added some convenience constants to provide a cleaner API when configuring a channel.

### Motivation:

The API could be a little difficult to use.

### Modifications:

Defined a struct `Socket` inside `ChannelOptions` that provides some convenience constants that produce `ChannelOptions.Types.SocketOption`.

Sample usage:
**old** `ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_DEBUG)`
**new** `ChannelOptions.Socket.enableDebugging`

### Result:

Additions to API, no modifications.

### DIFF
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -226,7 +226,7 @@ extension ChannelOptions {
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
 public struct ChannelOptions {
     
-    @available(*, deprecated, message: "Please use `ChannelOptions.Types.SocketOption` directly")
+    @available(*, deprecated, renamed: "Types.SocketOption.other")
     public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
         .init(level: level, name: name)
     }
@@ -333,25 +333,25 @@ extension ChannelOptions {
 extension ChannelOptions.Types.SocketOption {
     
     /// Turn on debugging info recording
-    public static let enableDebugging = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_DEBUG)
+    public static let enableDebugging = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_DEBUG)
     
     /// Allow local address reusue
-    public static let allowLocalAddressReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR)
+    public static let allowLocalAddressReuse = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_REUSEADDR)
     
     /// Keep connections alive
-    public static let keepAlive = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_KEEPALIVE)
+    public static let keepAlive = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_KEEPALIVE)
     
     /// Permit sending of broadcast messages
-    public static let enableBroadcastMessages = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_BROADCAST)
+    public static let enableBroadcastMessages = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_BROADCAST)
     
     /// Bypass hardware when possible
-    public static let useLoopback = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_USELOOPBACK)
+    public static let useLoopback = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_USELOOPBACK)
     
     /// Allow local address *and* port reuse
-    public static let allowLocalPortReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT)
+    public static let allowLocalPortReuse = ChannelOptions.Types.SocketOption(level: SocketOptionLevel(SOL_SOCKET), name: SO_REUSEPORT)
     
-    public static let other = { (name: SocketOptionName) -> ChannelOptions.Types.SocketOption in
-        ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), name)
+    public static let other = { (level: SocketOptionLevel, name: SocketOptionName) -> ChannelOptions.Types.SocketOption in
+        ChannelOptions.Types.SocketOption(level: level, name: name)
     }
     
 }
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -331,7 +331,7 @@ extension ChannelOptions {
 // Provides convenience methods for `ChannelOption.socket`
 extension ChannelOptions {
     
-    public struct Socket {
+    public enum Socket {
         
         /// Turn on debugging info recording
         public static let enableDebugging = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_DEBUG)
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -225,7 +225,8 @@ extension ChannelOptions {
 
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
 public struct ChannelOptions {
-    /// - seealso: `SocketOption`.
+    
+    @available(*, deprecated, message: "Please use `ChannelOptions.Types.SocketOption` directly")
     public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
         .init(level: level, name: name)
     }
@@ -329,28 +330,28 @@ extension ChannelOptions {
 }
 
 // Provides convenience methods for `ChannelOption.socket`
-extension ChannelOptions {
+extension ChannelOptions.Types.SocketOption {
     
-    public enum Socket {
-        
-        /// Turn on debugging info recording
-        public static let enableDebugging = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_DEBUG)
-        
-        /// Allow local address reusue
-        public static let allowLocalAddressReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR)
-        
-        /// Keep connections alive
-        public static let keepAlive = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_KEEPALIVE)
-        
-        /// Permit sending of broadcast messages
-        public static let enableBroadcastMessages = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_BROADCAST)
-        
-        /// Bypass hardware when possible
-        public static let useLoopback = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_USELOOPBACK)
-        
-        /// Allow local address *and* port reuse
-        public static let allowLocalPortReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT)
-        
+    /// Turn on debugging info recording
+    public static let enableDebugging = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_DEBUG)
+    
+    /// Allow local address reusue
+    public static let allowLocalAddressReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR)
+    
+    /// Keep connections alive
+    public static let keepAlive = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_KEEPALIVE)
+    
+    /// Permit sending of broadcast messages
+    public static let enableBroadcastMessages = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_BROADCAST)
+    
+    /// Bypass hardware when possible
+    public static let useLoopback = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_USELOOPBACK)
+    
+    /// Allow local address *and* port reuse
+    public static let allowLocalPortReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEPORT)
+    
+    public static let other = { (name: SocketOptionName) -> ChannelOptions.Types.SocketOption in
+        ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), name)
     }
     
 }
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -334,22 +334,22 @@ extension ChannelOptions {
     public struct Socket {
         
         /// Turn on debugging info recording
-        public static let enableDebugging = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_DEBUG)
+        public static let enableDebugging = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_DEBUG)
         
         /// Allow local address reusue
-        public static let allowLocalAddressReuse = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_REUSEADDR)
+        public static let allowLocalAddressReuse = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR)
         
         /// Keep connections alive
-        public static let keepAlive = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_KEEPALIVE)
+        public static let keepAlive = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_KEEPALIVE)
         
         /// Permit sending of broadcast messages
-        public static let enableBroadcastMessages = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_BROADCAST)
+        public static let enableBroadcastMessages = ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_BROADCAST)
         
         /// Bypass hardware when possible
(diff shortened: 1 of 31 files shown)


Sources/NIO/ChannelOption.swift
```diff
@@ -327,3 +327,39 @@ extension ChannelOptions {
...
+        public static let acceptConnections = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_ACCEPTCONN)
+        
+        /// Allow local address reusue
+        public static let allowLocalAddressReuse = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_REUSEADDR)
+        
+        /// Keep connections alive
+        public static let keepAlive = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_KEEPALIVE)
+        
+        /// Just use interface addresses
+        public static let dontRoute = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_DONTROUTE)
```
Lukasa: I don't think we need `dontRoute`, no-one ever sets it.

Sources/NIO/ChannelOption.swift
```diff
@@ -327,3 +327,39 @@ extension ChannelOptions {
...
+        
+        /// Permit sending of broadcast messages
+        public static let enableBroadcastMessages = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_BROADCAST)
+        
+        /// Bypass hardware when possible
+        public static let useLoopback = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_USELOOPBACK)
+        
+        /// Linger on close if data present
+        public static let linger = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_LINGER)
```
Lukasa: Let's not bother with `SO_LINGER`, as `SocketOptionProvider` is a better API for setting it.

Sources/NIO/ChannelOption.swift
```diff
@@ -327,3 +327,39 @@ extension ChannelOptions {
...
+// Provides convenience methods for `ChannelOption.socket`
+extension ChannelOptions {
+    
+    public struct Socket {
+        
+        /// Turn on debugging info recording
+        public static let enableDebugging = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_DEBUG)
+        
+        /// Socket has had listen()
+        public static let acceptConnections = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_ACCEPTCONN)
```
Lukasa: No-one ever sets this so let's skip that.

Sources/NIO/ChannelOption.swift
```diff
@@ -327,3 +327,39 @@ extension ChannelOptions {
...
     }
 }
+
+// Provides convenience methods for `ChannelOption.socket`
+extension ChannelOptions {
+    
+    public struct Socket {
+        
+        /// Turn on debugging info recording
+        public static let enableDebugging = ChannelOptions.Types.SocketOption(level: SOL_SOCKET, name: SO_DEBUG)
```
weissi: I'm happy with this, but this way, users still need to specify `1` as 'true' instead of `true`. We _could_ also define a new type in `ChannelOptions.Types` which abstracts that away. @Lukasa WYDT?

    ...Options(ChannelOptions.Socket.enableDebugging, value: 1)

vs

    ...Options(ChannelOptions.Socket.enableDebugging, value: true)

at the expense of more code in this PR. I favour the latter but I'm ok going with the first. If we were to choose the second, then we could make the `ChannelOptions.Types.BSDSocketOption<ValueType>`.

Lukasa: How does that interact with the pre-existing `SocketOptionProvider` protocol?

Sources/NIO/ChannelOption.swift
Davidde94: Change made, however what would be the reasoning for not supporting a simple use case, no matter how rare?

Sources/NIO/ChannelOption.swift
Davidde94: I'm ok with either, but would lean toward putting in a separate PR...

Davidde94: Scratch that, this is a super-simple change, so I will add in this PR

Sources/NIO/ChannelOption.swift
Lukasa: Lots of reasons, but here are a few:

- all code imposes maintenance costs. By default, my heuristic for writing any line of code begins with "don't" unless there is a use-case served by that code. Once written, we must support that code for the foreseeable future: code we didn't write is therefore code we don't have to support, and much cheaper than the code we did write.
- code no-one uses is silently broken. If no-one is using some code we wrote, the odds of it silently breaking or, worse, being or becoming useless are quite high. This is a double-whammy on top of the problem of having the maintenance burden of the code.
- extra code that exercises an API makes it harder to evolve that API. The more code there is that exercises a given API, the greater the cost of attempting to deprecate and replace that API. This doesn't affect the code we're writing today, but in general whenever someone is trying to evolve an API the default position is that all use-cases currently served by the existing API must be served by the new one. Having extra code that uses that API, even where we don't actually care about the use-case, restricts the way we can evolve the API by forcing us to justify why we don't want to support the feature.

All of this is a lengthy way of saying: code is a liability, not an asset. This is commonly misunderstood, because supported use cases are assets, and supported use-cases are enabled by code, so it can seem as though code is an asset. The important thing to realise is that the code itself is always a liability: it can just generate an asset that offsets its costs. This is why it's a bad idea to write code just because it would be easy to do so. In some sense, implementing simple-but-rare features is like eating junk food: it feels good at first, and doing so occasionally doesn't hurt much, but the long term consequences of frequently doing so are catastrophic.

In this particular instance, then, my reasoning goes like this:

1. This is new code, initial score is -1.
2. For a rare use-case, new score is -2.
3. This use-case can already be achieved today, albeit in an ugly way, so this does not add new _capabilites_ to NIO, new score is -2.5.
4. But the aesthetic is definitely better, new score is -1.

Better aesthetics are only ever worth it in cases where the use-case is either not supported at all or where we are saving huge chunks of code. Here we do neither. This is also why I'm ok with the other bits of code here: they are common paths, and having nice spellings for them is a quality of life improvement.

Davidde94: I agree for the most part, however I think this is a slightly different case. Here we're providing a convenience API for setting socket options. In my mind this should be "all or nothing". If a TP developer is using the newer/cleaner API but is unable to find an option (e.g. `acceptConnections`) then they may assume we don't support it at all, causing confusion and/or frustration until they give up and use some other technology or spend significant time hunting through the NIO source.

Admittedly this can be resolved by adding some documentation, however:
- Which developer really jumps to documentation as their first port of call? (we're a lazy bunch)
- Code should be largely self-documenting (opinion)

Anyway, an interesting discussion for perhaps another time. Meanwhile I've removed :)

Davidde94: I think a middle ground is to deprecate the older API and "newify" it somewhat, so that it's a little easier for other engineers to discover via code complete

Lukasa: > I agree for the most part, however I think this is a slightly different case. Here we're providing a convenience API for setting socket options. In my mind this should be "all or nothing".

Sure, that makes sense, but then I have bad news, because the current `SocketOption` solution you're building on is not capable of representing all socket options. As @weissi has [hit upon here](https://github.com/apple/swift-nio/pull/1260#discussion_r348530477), the current SocketOption type cannot set generic value types for arbitrary option types, and is not type safe. This means it's a bad fit for "wide" socket options (such as the ones required for the various multicast groups) and it's a bad fit for type safety (because each option is not currently a separate type and so cannot communicate what type it wants).

On top of all of that, to be all-or-nothing requires a fairly complete understanding of what the available socket options are. As an example, [here](https://git.musl-libc.org/cgit/musl/tree/include/sys/socket.h#n199) is a list of the socket options supported by musl libc. There are at least 60 in the `SO_` space, plus many more in the [TCP namespace](https://git.musl-libc.org/cgit/musl/tree/include/netinet/tcp.h#n7), the [IP namespace](https://git.musl-libc.org/cgit/musl/tree/include/netinet/in.h#n178), as well as a couple for IPv6 and UDP.

To do a truly all-or-nothing solution requires a complete enumeration of all socket options the users may want to set, a platform-specific description of the type required by each one, and then sufficient code to get and set those. This then needs to be repeated on Linux with all relevant libc implementations as well as on the Apple platforms, including any version-specific annotations we might need to deal with there. We then need to also cover doing this in newer OS versions whenever new socket options come along, or whenever we add support for a new socket type (e.g. ethernet, raw sockets).

This is not necessarily a bad idea, to be clear: I just want to caution against the idea of doing an "all or nothing" socket option API that explicitly names each socket option. I don't think we want that: I think we want good names for the common options, and an escape hatch for everything else.

Tests/NIOTests/ChannelOptionStorageTest.swift
```diff
@@ -61,6 +61,45 @@ class ChannelOptionStorageTest: XCTestCase {
         XCTAssertEqual([SocketOptionValue(2)],
                        (optionsCollector.allOptions as! [(ChannelOptions.Types.SocketOption, SocketOptionValue)]).map { $0.1 })
```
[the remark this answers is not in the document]
Lukasa: I think in particular @weissi wants us to use the ones with good names where we can (i.e. where we're using `SO_REUSEADDR`).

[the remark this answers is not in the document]
Davidde94: Yep, doing now :)

Sources/NIO/ChannelOption.swift
Davidde94: @weissi I've done quite a bit of experimenting for the second option, the code changes are significant so I think this is best placed in another PR that I can work on. What do you think?

Sources/NIOChatClient/main.swift
```diff
@@ -44,7 +44,7 @@ private final class ChatHandler: ChannelInboundHandler {
 let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 let bootstrap = ClientBootstrap(group: group)
     // Enable SO_REUSEADDR.
-    .channelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+    .channelOption(ChannelOptions.Types.SocketOption.allowLocalAddressReuse, value: 1)
```
weissi: we shouldn't expose the `Types` thing in actual code unless really necessary. I think we should have something like

    ChannelOptions.allowLocalAddressReuse, value: true

weissi: The `Types` is just an artefact of how this is implemented under the hood but the user shouldn't ever see this (unless they're defining their own channel options).

Sources/NIO/ChannelOption.swift
weissi: @Lukasa I don't think we need the more complicated `SocketOptionProvider` protocol. Here's what I mean with the example of just a replacement for `SO_REUSEADDR`:

```
diff --git a/Sources/NIO/BaseSocketChannel.swift b/Sources/NIO/BaseSocketChannel.swift
index 7bbd5ac9..0df8b2ce 100644
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -482,6 +482,8 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
         switch option {
         case let option as ChannelOptions.Types.SocketOption:
             try self.setSocketOption0(level: option.level, name: option.name, value: value)
+        case let option as ChannelOptions.Types.IntAsBoolSocketOption:
+            try self.setSocketOption0(level: option.level, name: option.name, value: value as! Bool ? 1: 0)
         case _ as ChannelOptions.Types.AllocatorOption:
             bufferAllocator = value as! ByteBufferAllocator
         case _ as ChannelOptions.Types.RecvAllocatorOption:
diff --git a/Sources/NIO/ChannelOption.swift b/Sources/NIO/ChannelOption.swift
index 483d1f3d..fb5af6c6 100644
--- a/Sources/NIO/ChannelOption.swift
+++ b/Sources/NIO/ChannelOption.swift
@@ -66,6 +66,23 @@ public typealias AllowRemoteHalfClosureOption = ChannelOptions.Types.AllowRemote
 extension ChannelOptions {
     public enum Types {
 
+        public struct IntAsBoolSocketOption: ChannelOption, Equatable {
+            public typealias Value = Bool
+
+            public var level: SocketOptionLevel
+            public var name: SocketOptionName
+
+            /// Create a new `SocketOption`.
+            ///
+            /// - parameters:
+            ///     - level: The level for the option as defined in `man setsockopt`, e.g. SO_SOCKET.
+            ///     - name: The name of the option as defined in `man setsockopt`, e.g. `SO_REUSEADDR`.
+            public init(level: SocketOptionLevel, name: SocketOptionName) {
+                self.level = level
+                self.name = name
+            }
+        }
+
         /// `SocketOption` allows users to specify configuration settings that are directly applied to the underlying socket file descriptor.
         ///
         /// Valid options are typically found in the various man pages like `man 4 tcp`.
@@ -230,6 +247,8 @@ public struct ChannelOptions {
         .init(level: level, name: name)
     }
 
+    public static let allowLocalAddressReUse = Types.IntAsBoolSocketOption(level: .init(SOL_SOCKET), name: SO_REUSEADDR)
+
     /// - seealso: `AllocatorOption`.
     public static let allocator = Types.AllocatorOption()
 
diff --git a/Sources/NIOChatClient/main.swift b/Sources/NIOChatClient/main.swift
index b03af9ea..26135af2 100644
--- a/Sources/NIOChatClient/main.swift
+++ b/Sources/NIOChatClient/main.swift
@@ -44,7 +44,7 @@ private final class ChatHandler: ChannelInboundHandler {
 let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 let bootstrap = ClientBootstrap(group: group)
     // Enable SO_REUSEADDR.
-    .channelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+    .channelOption(ChannelOptions.allowLocalAddressReUse, value: true)
         channel.pipeline.addHandler(ChatHandler())
     }
diff --git a/Sources/NIOChatServer/main.swift b/Sources/NIOChatServer/main.swift
index 9924723f..cc0e6769 100644
--- a/Sources/NIOChatServer/main.swift
+++ b/Sources/NIOChatServer/main.swift
@@ -121,7 +121,7 @@ let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
 let bootstrap = ServerBootstrap(group: group)
     // Specify backlog and enable SO_REUSEADDR for the server itself
     .serverChannelOption(ChannelOptions.backlog, value: 256)
-    .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
+    .serverChannelOption(ChannelOptions.allowLocalAddressReUse, value: true)
 
     // Set the handlers that are applied to the accepted Channels
```

@Davidde94 is this really much more code?

weissi: for all the actually boolean ones, we need only one type (`BoolAsIntSocketOption`) and then we can go ahead and simply define

```swift
public static let allowLocalAddressReUse = Types.IntAsBoolSocketOption(level: .init(SOL_SOCKET), name: SO_REUSEADDR)
public static let allowPortReUse = Types.IntAsBoolSocketOption(level: .init(SOL_SOCKET), name: SO_REUSEPORT)
public static let enableDebugging = Types.IntAsBoolSocketOption(level: .init(SOL_SOCKET), name: SO_DEBUG)
...
```

Lukasa: @weissi I was responding only to the proposal to have `ChannelOptions.Types.BSDSocketOption<ValueType>`. Special-casing bool isn't a big deal to my eye, but widening the scope raises questions about whether we're killing `SocketOptionProvider` or not.

weissi: @Lukasa ah sorry, at first I thought we might want to use one type (`BSDSocketOption<ValueType>`) instead of specialised ones (`BoolAsIntSocketOption`) because I thought it's easier in the code but it's not because we still have to match each type separately (`BSDSocketOption<Bool>`, `BSDSocketOption<...>`). That's why I then quickly switched to `BoolAsIntSocketOption` (terrible name...).

Regarding `SocketOptionProvider`: I thought we need that if someone wants to set an option for which we don't have special support and that is wider than a `CInt`?

Lukasa: We do, but once we go down the road of `BSDSocketOption<Value>` there is precious little reason not to allow `Value` to be essentially any type and just bitwise-copy it, which would cover the use-case for `SocketOptionProvider` as well.

weissi: @Lukasa aah I see. I don't _think_ it would work with a truly generic `BSDSocketOption<Value>` because we need to match the type in this switch

```
        switch option {
        case let option as ChannelOptions.Types.SocketOption:
            try self.setSocketOption0(level: option.level, name: option.name, value: value)
        case let option as ChannelOptions.Types.IntAsBoolSocketOption:
            try self.setSocketOption0(level: option.level, name: option.name, value: value as! Bool ? 1: 0)
```

and we can't just write

```
    case let option as ChannelOptions.Types.BSDSocketOption<T>:
```

we need a concrete `T` here and can't just match all of them. We could match `BSDSocketOption<Bool>`, and `BSDSocketOption<CInt>`, and `BSDSocketOption<linger>`, ... but that way we wouldn't support arbitrary wide stuff. Am I missing anything?

What we _could_ probably do is something like

```
        public struct GenericBSDSocketOption: ChannelOption, Equatable {
            public typealias Value = [UInt8] // byte representation of the value

            public var level: SocketOptionLevel
            public var name: SocketOptionName

            public init(level: SocketOptionLevel, name: SocketOptionName) {
                self.level = level
                self.name = name
            }
        }
```

and then value would just be the bytes (bitwise-copied) for the value to set.

weissi: and we'd set the value by doing

```
var crazyOptionValue: linger = linger(...)
.channelOption(GenericBSDSocketOption(level: SOL_SOCKET, name: SO_LINGER), value: withUnsafeBytes(of: &crazyOptionValue, Array.init))
```

Davidde94:
Having to close, will reopen in another PR

Davidde94:
@PeterAdams-A has taken care of this

PR #1261, opened 2019-11-20
join the threads of a MultiThreadedEventLoopGroup on shutdown

### LINKED ISSUE #581: join the threads of the MultiThreadedEventLoopGroup when shut down
we should use `pthread_join` on the threads of the `MultiThreadedEventLoopGroup`'s threads in the `shutdownGracefully` method.

weissi:
Motivation:

We should join the threads that we spawned which has the extra benefit
of TSan being able to detect our thread leaks.

Modifications:

Join the SelectableEventLoop's thread after it's shut down.

Result:

- we join all the threads
- fixes #581

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -642,28 +642,48 @@ private func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
     #endif
 }
 
-/// The different state in the lifecycle of an `EventLoop`.
-private enum EventLoopLifecycleState {
-    /// `EventLoop` is open and so can process more work.
-    case open
-    /// `EventLoop` is currently in the process of closing.
-    case closing
-    /// `EventLoop` is closed.
-    case closed
-}
-
 /// `EventLoop` implementation that uses a `Selector` to get notified once there is more I/O or tasks to process.
 /// The whole processing of I/O and tasks is done by a `NIOThread` that is tied to the `SelectableEventLoop`. This `NIOThread`
 /// is guaranteed to never change!
 @usableFromInline
 internal final class SelectableEventLoop: EventLoop {
+    /// The different state in the lifecycle of an `EventLoop` seen from _outside_ the `EventLoop`.
+    private enum ExternalState {
+        /// `EventLoop` is open and so can process more work.
+        case open
+        /// `EventLoop` is currently in the process of closing.
+        case closing
+        /// `EventLoop` is closed.
+        case closed
+        /// `EventLoop` is closed and is currently trying to reclaim resources (such as the EventLoop thread).
+        case reclaimingResources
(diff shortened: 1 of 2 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -658,7 +658,7 @@ private enum EventLoopLifecycleState {
 @usableFromInline
 internal final class SelectableEventLoop: EventLoop {
     private let selector: NIO.Selector<NIORegistration>
-    private let thread: NIOThread
+    internal let thread: NIOThread
```
Lukasa: Instead of blowing up this abstraction, should we have `SelectableEventLoop` have a function that can be called that calls `join()`? Maybe a `destroy` function or similar?

weissi: yeah, I'll probably to a refactor there. I'll try not to go nuts because there's a lot that could be refactored :P

Sources/NIO/EventLoop.swift
```diff
@@ -952,33 +981,76 @@ internal final class SelectableEventLoop: EventLoop {
...
-    /// Gently close this `SelectableEventLoop` which means we will close all `SelectableChannel`s before finally close this `SelectableEventLoop` as well.
-    public func closeGently() -> EventLoopFuture<Void> {
-        func closeGently0() -> EventLoopFuture<Void> {
-            guard self.lifecycleState == .open else {
-                return self.makeFailedFuture(EventLoopError.shutdown)
+            self.selector.closeGently(eventLoop: self).whenComplete { result in
+                self.assertInEventLoop()
+                assert(self.internalState == .runningButNotAcceptingNewRegistrations)
+                self.internalState = .noLongerRunning
+                self.execute {}
```
Lukasa: What's this for?

Sources/NIO/EventLoop.swift
```diff
@@ -1227,41 +1292,47 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
...
-                    fatalError("runState was \(self.runState), expected .closing")
-                }
-
-                self.runState = .closed(error)
-
-                return callbacks
+            for loop in self.eventLoops {
+                loop.syncFinaliseClose()
             }
+            self.shutdownLock.withLock {
```
Lukasa: Are we deliberately changing the scope of this lock?

Sources/NIO/EventLoop.swift
weissi: adding a comment: this is to force the event loop to create a new tick and then realise that it is no longer running.

Sources/NIO/EventLoop.swift
weissi: great catch, we're definitely not. fixing

Sources/NIO/EventLoop.swift
weissi: this is now done.

PR #1263, opened 2019-11-21
Fast atomics

2bjake:
Motivation:

The existing Atomic class holds an UnsafeEmbeddedAtomic which holds an OpaquePointer to raw memory for the C atomic. This results in multiple allocations on init. The new FastAtomic class uses ManagedBufferPointer which tail allocates memory for the C atomic as part of the FastAtomic class allocation.

Modifications:

Created FastAtomic class that uses ManagedBufferPointer to allocate memory for the C atomic. Created version of catmc_atomic_* C functions that expect memory to be allocated/deallocated by caller. Created FastAtomicPrimitive protocol for the new version of catmc_atomic_* functions.

Result:

Purely additive. FastAtomic is available as an alternative to Atomic which results in fewer memory allocations.


Sources/CNIOAtomics/include/CNIOFastAtomics.h
```diff
@@ -0,0 +1,280 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa: Mind setting this date to `2019` instead?

Sources/CNIOAtomics/src/c-fastatomics.c
```diff
@@ -0,0 +1,76 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa: Mind setting this date to 2019 as well?

Sources/CNIOAtomics/include/CNIOFastAtomics.h
```diff
@@ -0,0 +1,280 @@
+//===----------------------------------------------------------------------===//
```
Lukasa: The addition of this file is going to break the `swift package generate-xcodeproj` ability to generate Xcode modules (I know it's weird, just trust me on this point). Can we add these definitions to `CNIOAtomics.h` instead?

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,271 @@
...
+
+public class FastAtomic<T: FastAtomicPrimitive> {
+    @usableFromInline
+    typealias Manager = ManagedBufferPointer<Void, T.AtomicWrapper>
+
+    /// Create an atomic object with `value`
+    @inlinable
+    static func makeAtomic(value: T) -> FastAtomic {
+        let manager = Manager(bufferClass: self, minimumCapacity: 1) { _, _ in }
+        let fastAtomic = unsafeDowncast(manager.buffer, to: FastAtomic<T>.self)
```
Lukasa: I don't love `unsafeDowncast` here: is there any reason not to have the type checker validate we got this right with `as!`?

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,271 @@
...
+
+        Manager(unsafeBufferObject: fastAtomic).withUnsafeMutablePointerToElements {
```
Lukasa: I don't think we need to create a new `Manager` object here: can't we use the `manager` we got from the initialization?

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,271 @@
...
+import CNIOAtomics
+
+/// The protocol that all types that can be made atomic must conform to.
+///
+/// **Do not add conformance to this protocol for arbitrary types**. Only a small range
+/// of types have appropriate atomic operations supported by the CPU, and those types
+/// already have conformances implemented.
+public protocol FastAtomicPrimitive {
+    associatedtype AtomicWrapper
+    static var fast_atomic_create_with_existing_storage: (UnsafeMutableRawPointer, Self) -> Void { get }
```
Lukasa: Is there any reason this can't take a typed pointer as its first argument?

Sources/CNIOAtomics/include/CNIOFastAtomics.h
2bjake: I updated CNIOAtomics.h since this one is getting removed.

Sources/CNIOAtomics/src/c-fastatomics.c
2bjake: no problem

Sources/CNIOAtomics/include/CNIOFastAtomics.h
2bjake: Sure, moved.

Sources/NIOConcurrencyHelpers/fastatomics.swift
2bjake: Yeah, I originally wrote it that way but it seemed a little weird to be setting the first element in the makingHeaderWith factory, so I moved it out. But I can switch it back to that style.

2bjake: Ah, right. Will change.

Sources/NIOConcurrencyHelpers/fastatomics.swift
2bjake: ```as!``` should be fine. I'll change it.

Sources/NIOConcurrencyHelpers/fastatomics.swift
2bjake: When I started, the ```AtomicWrapper``` type wasn't available on the Swift side, so I couldn't type it. But now that ```FutureAtomicPrimitive``` makes the type available, it shouldn't be a problem to change it.

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,269 @@
...
+}
+
+public class FastAtomic<T: FastAtomicPrimitive> {
+    @usableFromInline
+    typealias Manager = ManagedBufferPointer<Void, T.AtomicWrapper>
+
+    /// Create an atomic object with `value`
+    @inlinable
+    static func makeAtomic(value: T) -> FastAtomic {
+        let manager = Manager(bufferClass: self, minimumCapacity: 1) { _, _ in }
```
weissi: @atrick the only reason we can use an `UInt8` buffer of the right size is that we need memory that’s sufficiently aligned for an atomic. I assume there are no alignment guarantees for this?

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,269 @@
...
+    /// This implementation uses a *relaxed* memory ordering. This guarantees nothing
+    /// more than that this operation is atomic: there is no guarantee that any other
+    /// event will be ordered before or after this one.
+    ///
+    /// - Parameter rhs: The value to add to this object.
+    /// - Returns: The previous value of this object, before the addition occurred.
+    @inlinable
+    public func add(_ rhs: T) -> T {
+        return Manager(unsafeBufferObject: self).withUnsafeMutablePointerToElements {
+            return T.fast_atomic_add($0, rhs)
```
weissi: @atrick just to double check, we can rely on the tail allocated storage in our class to always have a const address, right?

Sources/CNIOAtomics/src/c-fastatomics.c
```diff
@@ -0,0 +1,75 @@
...
+#include <stdatomic.h>
+#include <stdbool.h>
+#include <inttypes.h>
+#include <stdio.h>
+
+#include "../include/CNIOAtomics.h"
+#include "cpp_magic.h"
+
+#define MAKE(type) /*
+*/ void catmc_fast_atomic_##type##_create_with_existing_storage(struct catmc_fast_atomic_##type *storage, type value) { /*
```
weissi: @2bjake instead of duplicating everything here, couldn’t we just declare the existing structs in the header like you do for the fast atomics? There’s no actual need to only forward declare them. I’m sure we can get away with just one copy of the C stuff

2bjake: I originally tried this but declaring the structs in the header causes the generated Swift interface of these functions to change. This means the ```AtomicPrimitive``` protocol would have to change as well (compare ```AtomicPrimitive``` and ```FastAtomicPrimitive```). Wouldn't this result in a backward incompatible API change?

2bjake: Wait, I see a way where the pointer type of ```AtomicPrimitive``` wouldn't have to change pointer types by moving away from pointfree conformance and wrapping the c functions. Something like: 
```swift
public static let opaque_add: (OpaquePointer, Self) -> Self = { catmc_fast_atomic__Bool_add(UnsafeMutablePointer<AtomicWrapper>($0), $1) }
```

But there will still need to be a ```FastAtomicPrimitive``` protocol, right? So I can add the create_with_storage method and the AtomicWrapper associatedtype in a non-breaking way.

Sources/NIOConcurrencyHelpers/fastatomics.swift
atrick: @weissi I'm not sure what you're asking. `ManagedBuffer` gives you the alignment guarantee per its `_alignmentMask`. `Unsafe[Buffer]Pointer<T>.allocate` would give you the same guarantee. Or you can explicitly provide alignment to `UnsafeRawPointer.allocate`

Sources/NIOConcurrencyHelpers/fastatomics.swift
atrick: Yeah, there's no way to move a class object. IIRC the way `realloc` would work is by providing a new reference (old references will never see the realloc'd class).

Sources/CNIOAtomics/src/c-fastatomics.c
2bjake: OK, I was able to remove the duplicate c code. I was also able to dramatically reduce the boilerplate for ```AtomicPrimitive``` conformance by providing default implementation computed properties which rely on ```FastAtomicPrimitive```!

2bjake:
@weissi I'm not sure how I can make ```Atomic``` be implemented by ```FastAtomic``` since they're generic over different types (```AtomicPrimitive``` vs ```FastAtomicPrimitive```). I was able to convert ```AtomicBox``` over to ```FastAtomic``` though.

Sources/NIOConcurrencyHelpers/fastatomics.swift
```diff
@@ -0,0 +1,269 @@
...
+    /// This implementation uses a *relaxed* memory ordering. This guarantees nothing
+    /// more than that this operation is atomic: there is no guarantee that any other
+    /// event will be ordered before or after this one.
+    ///
+    /// - Parameter value: The new value to set this object to.
+    /// - Returns: The value previously held by this object.
+    @inlinable
+    public func exchange(with value: T) -> T {
+        return Manager(unsafeBufferObject: self).withUnsafeMutablePointerToElements {
+            return T.fast_atomic_exchange($0, value)
```
2bjake: @weissi @Lukasa What's your stance on implicit returns? Both returns in each of these functions can be elided but I included them since I don't know the preferred style.

Sources/NIOConcurrencyHelpers/fastatomics.swift
Lukasa: I'm a bit confused by @weissi's question here: we're not allocating a `UInt8` buffer, we're allocating a buffer of type `T.AtomicWrapper` of size 1. This does not seem like it would have any alignment issues at all.

Sources/NIOConcurrencyHelpers/fastatomics.swift
Lukasa: Mixed. We tend to allow them within single-statement closures, but as we support Swift 5 we forbid them everywhere else.

Sources/NIOConcurrencyHelpers/fastatomics.swift
weissi: @atrick / @Lukasa sorry, I was on a phone only on Friday. What we really want to express here is tail allocation for an opaque (forward declared) struct in C. Of course because it's opaque, we don't know it's true length/alignment requirements. We could however expose getters for length & alignment.

That way, we could make a `ManagedBufferPointer` that holds `length` `UInt8`s. Then we could pass the `UnsafeMutableRawBufferPointer` to these `UInt8`s to the C function that initialises the atomic. The only issue here is that our `UnsafeMutableRawBufferPointer` isn't correctly aligned to fit the struct that contains the atomic.

And I was wondering if we could ask `ManagedBufferPointer` to give us tail allocated storage aligned to a user-chosen alignment. So something like

    let manager = Manager<Void, UInt8>(bufferClass: self, minimumCapacity: get_width_for_opaque_struct(), alignment: get_alignment_requirement_for_opaque_struct()) { _, _, in }

But this API doesn't seem to exist. We could obviously ask for a possibly larger size and align ourselves.

weissi:

Sources/NIOConcurrencyHelpers/fastatomics.swift
atrick: @weissi Ah. I get it. I think it would be reasonable to propose adding an (over)alignment option to ManagedBuffer.create. I'm not sure if we can add an underscored, defaulted argument to a public API. Might need to add a new underscored API instead. Without an API I think you need to allocate extra space.

weissi:
@2bjake unfortunately, the API breakage checker found breakages here: https://ci.swiftserver.group/job/swift-nio2-api-breakage-prb/406/console 

Essentially, we need to keep the `OpaquePointer` APIs depite the fact we now accept `UnsafePointer<concrete_type>`

2bjake:
@weissi The types of the original ```AtomicPrimitive``` APIs are unchanged and still deal in ```OpaquePointer```s. Is it an issue that they are now provided as default implementations on a protocol extension?

weissi:
@2bjake 

```
[...]
 17:12:39 /* Type Changes */
17:12:39 Accessor Bool.atomic_add.Get() has return type change from (OpaquePointer, Bool) -> Bool to (UnsafeMutablePointer<catmc_atomic__Bool>, Bool) -> Bool
17:12:39 Accessor Bool.atomic_compare_and_exchange.Get() has return type change from (OpaquePointer, Bool, Bool) -> Bool to (UnsafeMutablePointer<catmc_atomic__Bool>, Bool, Bool) -> Bool
17:12:39 Accessor Bool.atomic_exchange.Get() has return type change from (OpaquePointer, Bool) -> Bool to (UnsafeMutablePointer<catmc_atomic__Bool>, Bool) -> Bool
17:12:39 Accessor Bool.atomic_load.Get() has return type change from (OpaquePointer) -> Bool to (UnsafeMutablePointer<catmc_atomic__Bool>) -> Bool
17:12:39 Accessor Bool.atomic_store.Get() has return type change from (OpaquePointer, Bool) -> Void to (UnsafeMutablePointer<catmc_atomic__Bool>, Bool) -> Void
17:12:39 Accessor Bool.atomic_sub.Get() has return type change from (OpaquePointer, Bool) -> Bool to (UnsafeMutablePointer<catmc_atomic__Bool>, Bool) -> Bool
17:12:39 Accessor Int.atomic_add.Get() has return type change from (OpaquePointer, Int) -> Int to (UnsafeMutablePointer<catmc_atomic_long>, Int) -> Int
17:12:39 Accessor Int.atomic_compare_and_exchange.Get() has return type change from (OpaquePointer, Int, Int) -> Bool to (UnsafeMutablePointer<catmc_atomic_long>, Int, Int) -> Bool
17:12:39 Accessor Int.atomic_exchange.Get() has return type change from (OpaquePointer, Int) -> Int to (UnsafeMutablePointer<catmc_atomic_long>, Int) -> Int
17:12:39 Accessor Int.atomic_load.Get() has return type change from (OpaquePointer) -> Int to (UnsafeMutablePointer<catmc_atomic_long>) -> Int
17:12:39 Accessor Int.atomic_store.Get() has return type change from (OpaquePointer, Int) -> Void to (UnsafeMutablePointer<catmc_atomic_long>, Int) -> Void
17:12:39 Accessor Int.atomic_sub.Get() has return type change from (OpaquePointer, Int) -> Int to (UnsafeMutablePointer<catmc_atomic_long>, Int) -> Int
17:12:39 Accessor Int16.atomic_add.Get() has return type change from (OpaquePointer, int_least16_t) -> int_least16_t to (UnsafeMutablePointer<catmc_atomic_int_least16_t>, int_least16_t) -> int_least16_t
```

weissi:
@2bjake the problem is that a `some_struct *` in imported into Swift in two different ways:

1. `UnsafePointer<some_struct>` if `struct some_struct` is fully declared in the header file
2. `OpaquePointer` if `struct some_struct` is forward declared

Previously, we had situation (2) but now it's (1). That's not a problem per se but the types did change unfortunately.

weissi:
@2bjake sorry, I think the real problem is this

```
17:12:39 /* Renamed Decls */
17:12:39 Var Bool.atomic_add has been renamed to Var fast_atomic_add
17:12:39 Var Bool.atomic_compare_and_exchange has been renamed to Var fast_atomic_compare_and_exchange
17:12:39 Var Bool.atomic_exchange has been renamed to Var fast_atomic_exchange
17:12:39 Var Bool.atomic_load has been renamed to Var fast_atomic_load
17:12:39 Var Bool.atomic_store has been renamed to Var fast_atomic_store
17:12:39 Var Bool.atomic_sub has been renamed to Var fast_atomic_sub
17:12:39 Var Int.atomic_add has been renamed to Var fast_atomic_add
17:12:39 Var Int.atomic_compare_and_exchange has been renamed to Var fast_atomic_compare_and_exchange
17:12:39 Var Int.atomic_exchange has been renamed to Var fast_atomic_exchange
17:12:39 Var Int.atomic_load has been renamed to Var fast_atomic_load
17:12:39 Var Int.atomic_store has been renamed to Var fast_atomic_store
17:12:39 Var Int.atomic_sub has been renamed to Var fast_atomic_sub
17:12:39 Var Int16.atomic_add has been renamed to Var fast_atomic_add
17:12:39 Var Int16.atomic_compare_and_exchange has been renamed to Var fast_atomic_compare_and_exchange
17:12:39 Var Int16.atomic_exchange has been renamed to Var fast_atomic_exchange
17:12:39 Var Int16.atomic_load has been renamed to Var fast_atomic_load
17:12:39 Var Int16.atomic_store has been renamed to Var fast_atomic_store
17:12:39 Var Int16.atomic_sub has been renamed to Var fast_atomic_sub
17:12:39 Var Int32.atomic_add has been renamed to Var fast_atomic_add
17:12:39 Var Int32.atomic_compare_and_exchange has been renamed to Var fast_atomic_compare_and_exchange
17:12:39 Var Int32.atomic_exchange has been renamed to Var fast_atomic_exchange
17:12:39 Var Int32.atomic_load has been renamed to Var fast_atomic_load
```

so for example `Int32.atomic_load` is no longer present but it was before. That's why the api-digester thinks it was renamed (but really it was removed)

2bjake:
@weissi Right, I saw those. ```But Int32.atomic_load``` and all the rest _are_ still present. The ```AtomicPrimitive``` protocol APIs are unchanged, and all of the primitives still conform it. The only difference is that each primitive's ```AtomicPrimitive``` extension doesn't explicitly list each API. They get conformance from the protocol extension that has default implementations that rely on the ```FastAtomicPrimitive``` APIs.

weissi:
@2bjake hmm, I see what you're saying. That would mean that this is a `swift-api-digester` bug then though. Would you mind filing a bug report there?

weissi:
@2bjake actually, I can do the bug report filing.

weissi:
@2bjake I think `api-digester` is right. Try this program:

```
public protocol FP {
    associatedtype Foo

    static var fastFoo: (String) -> String { get }
}

extension P where Self: FP {
    public static var foo: (String) -> String {
        return {
            return Self.fastFoo($0)
        }
    }
}

public protocol P {
    static var foo: (String) -> String {get}
}

public struct S: FP {
    public typealias Foo = String
    public static let fastFoo: (String) -> String = { $0 }
}

public struct T: P {
    public static let foo: (String) -> String = { $0 }
}

public func use() {
    let s = S()
    let t = T()

    print(S.foo("sss"))
    print(S.fastFoo("ccc"))
    print(T.foo("asdf"))
}
```

which will fail compilation with

```
$ swift x.swift 
x.swift:32:13: error: type 'S' has no member 'foo'
    print(S.foo("sss"))
          ~ ^~~
```

weissi:
@2bjake the error is that nothing conforms for example `Bool` to `AtomicPrimitive`, right?

2bjake:
@weissi The problem with that sample program is that nothing conforms S to P. Add ```extension S: P {}``` and it compiles. If you look in atomics.swift, conformance to ```AtomicPrimitive``` is added to all of the primitives, for example:
```swift
extension Bool: AtomicPrimitive {
    public static let atomic_create  = { OpaquePointer(catmc_atomic__Bool_create($0)) }
    public static let atomic_destroy = { (p: OpaquePointer) in catmc_atomic__Bool_destroy(UnsafeMutablePointer(p)) }
}
```

2bjake:
@weissi Here's an example that is just a reduced version of the actual code:
```swift
// fastatomics.swift
public protocol FastAtomicPrimitive {
    associatedtype AtomicWrapper
    static var fast_atomic_add: (UnsafeMutablePointer<AtomicWrapper>, Self) -> Self { get }
}

public struct catmc_atomic__Bool {}

extension Bool: FastAtomicPrimitive {
    public typealias AtomicWrapper = catmc_atomic__Bool
    public static let fast_atomic_add: (UnsafeMutablePointer<AtomicWrapper>, Self) -> Self = { _, _ in true }
}

// atomics.swift
public protocol AtomicPrimitive {
    static var atomic_create: (Self) -> OpaquePointer { get }
    static var atomic_add: (OpaquePointer, Self) -> Self { get }
}

extension AtomicPrimitive where Self: FastAtomicPrimitive {
    public static var atomic_add: (OpaquePointer, Self) -> Self { return { fast_atomic_add(UnsafeMutablePointer($0), $1) } }
}

extension Bool: AtomicPrimitive {
    public static var atomic_create: (Bool) -> OpaquePointer = { _ in
        OpaquePointer(bitPattern: 123)!
    }
}
```

weissi:
thanks @2bjake , I created https://bugs.swift.org/browse/SR-11859 . Do you have a user on that JIRA so I can credit you for the proper repro?

Do you see a way to make API digester happy? Because if not, all our tooling will now forever tell us that there's a problem between 2.10 and 2.11, that's not the end of the world but still not great.

weissi:
@2bjake oh, also: the allocations per HTTP request went up by 9 allocations per request:

```
17:20:18 ++ assert_less_than_or_equal 39000 30970
17:20:18 ++ [[ ! 39000 -le 30970 ]]
17:20:18 ++ fail 'assertion '\''39000'\'' <= '\''30970'\'' failed'
17:20:18 ++ echo 'FAILURE: assertion '\''39000'\'' <= '\''30970'\'' failed'
17:20:18 FAILURE: assertion '39000' <= '30970' failed
```

that's a 30% regression with an expected improvement... So somewhere, the new code must introduce new allocations, I suspect that the computed properties to conform the `FastAtomicPrimitive`s to the `AtomicPrimitives` cause those :(

2bjake:
@weissi The computed properties are for adapting the existing ```AtomicPrimitive``` to the new Swift interface of the c functions. If we went back to the duplicated c code solution, it would sidestep this issue. If the idea is to deprecate and eventually remove the original ```Atomic``` code, then the duplicate code is only temporary.

The duplicate c code solution would also make the API digester happy.

weissi:
@2bjake alright, let's do the copy & pasting then :). Sorry about the back and forth.

By the way, you can run the allocation counter tests yourself on your machine by running `IntegrationTests/run-tests.sh -f allocation`.

2bjake:
@weissi No worries. I agreed with you that it was worth a try to eliminate the duplicate code. It was fun coming up with the solution to do it with computed properties in the protocol extension which eliminated a bunch of the ```AtomicPrimitive``` conformance code. Unfortunately it does seem like a pattern in my experience that elegant code solutions don't always jive with perf & strict backward compatibility requirements.

Thanks for the tip about running the allocation tests. I was wondering how to do that! :)

2bjake:
Ok, with the deduplication commit reverted, allocations go from 554000 down to 548000 because ```AtomicBox``` is using ```FastAtomic```. I then converted the usages of ```Atomic``` throughout the codebase to ```FastAtomic``` and allocations dropped further to 543000.

2bjake:
@weissi My username on that JIRA is jakef.

weissi:

2bjake:
The```Atomic``` unit tests are getting deprecation warnings which causes validation to fail. Any thoughts on how to handle this?

weissi:
@2bjake there seem to be warnings in the code now, mind just switching those to `FastAtomic<T>`?

```
21:24:38 [13/85] Compiling NIO BaseSocket.swift
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:34:40: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38         let ai = NIOConcurrencyHelpers.Atomic<UInt64>(value: 0)
21:24:38                                        ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:47:18: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38         let ab = Atomic<Bool>(value: true)
21:24:38                  ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:60:18: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38         let ab = Atomic<Bool>(value: false)
21:24:38                  ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:80:22: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38             let ab = Atomic<T>(value: max)
21:24:38                      ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:110:22: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38             let ab = Atomic<T>(value: upperBound)
21:24:38                      ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:140:22: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38             let ab = Atomic<T>(value: zero)
21:24:38                      ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:171:22: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38             let ab = Atomic<T>(value: zero)
21:24:38                      ^
21:24:38 /code/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift:202:22: error: 'Atomic' is deprecated: please use FastAtomic instead
21:24:38             let ab = Atomic<T>(value: zero)
21:24:38                      ^
21:24:39 Build step 'Execute shell' marked build as failure
```

2bjake:
@weissi those are the unit tests for ```Atomic```. Switching that code to ```FastAtomic``` makes them worthless (we already have ```FastAtomic``` unit tests). Should I remove them? Or do you have a way to have tests against deprecated code without causing validation errors?

weissi:
@2bjake if you `rebase` your branch on top of master, you can use this trick: https://github.com/apple/swift-nio/blob/master/Tests/NIOTests/ByteBufferTest.swift#L2445

Just deprecate the tests that use deprecated functions :)

weissi:

2bjake:
@weissi I think the 5.1 validation error is a false alarm, it's failing because total_allocations is lower than expected, which is a good thing (this is what I saw on my machine). I'm not sure what's going on with the 5.0 validation error, allocs went up by 5, from 30970 to 30975.

weissi:
@2bjake awesome work! To make this actually pass CI, you can now do what is for me the favourite part of a PR that optimises allocations: You can edit the [two](https://github.com/apple/swift-nio/blob/master/docker/docker-compose.1604.51.yaml) [docker](https://github.com/apple/swift-nio/blob/master/docker/docker-compose.1804.50.yaml) files in your PR and subtract the right number of allocations your PR saves :).

So if you save 4 allocations, you typically subtract 4000 from the limits (because we usually run stuff 1000 times). Usually, there's some extra slack at the end to account for random stray allocations. So for example if we have a benchmark that allocates 12 times per round and we run it 1000 times, then I'd usually set a limit of 12050 == (12 times * 1000 allocations) + 50 slack.

Does that make sense?

weissi:
@2bjake for the one where it went up a tiny little bit (from 30790 to 30795, feel free to increase the slack to 30799 or something). Just needs to be below 31000.

weissi:

2bjake:
@weissi Allocations in the conn tests went down in 5.1 but went up in 5.0.

```
/// 5.1 ///
MAX_ALLOCS_ALLOWED_1000_reqs_1_conn=30550
test_1000_reqs_1_conn.total_allocations: 30492

MAX_ALLOCS_ALLOWED_1_reqs_1000_conn=533050
test_1_reqs_1000_conn.total_allocations: 522000

MAX_ALLOCS_ALLOWED_ping_pong_1000_reqs_1_conn=4450
test_ping_pong_1000_reqs_1_conn.total_allocations: 4434

/// 5.0 ///
MAX_ALLOCS_ALLOWED_1000_reqs_1_conn=30970
test_1000_reqs_1_conn.total_allocations: 30975

MAX_ALLOCS_ALLOWED_1_reqs_1000_conn=994050
test_1_reqs_1000_conn.total_allocations: 1005000

MAX_ALLOCS_ALLOWED_ping_pong_1000_reqs_1_conn=4480
test_ping_pong_1000_reqs_1_conn.total_allocations: 4489
```

I poked around online but didn't find anything about the 5.1 release that would explain it. A total guess, but I'm wondering if the capturing closures that are passed to ```withUnsafeMutablePointerToElements``` cause allocations in 5.0 but are optimized in 5.1? The ```FastAtomic``` implementation has a bunch of them.

weissi:
@2bjake that's great! We still care about 5.0 working but performance-wise 5.1 is definitely the target. So I think we're more than happy with 5.1 becoming better for 5.0 becoming worse :).

Just adjust the docker files accordingly and then this is good to go :).

weissi:
@2bjake oh damn, one annoying thing: according to our [public API declaration](https://github.com/apple/swift-nio/blob/master/docs/public-api.md) we have to prefix _all_ new global type/protocol names with `NIO`. So we'd need to rename them to `NIOFastAtomicPrimitive` and `NIOFastAtomic`. That makes me think: let's just use `NIOAtomic` maybe? 

The reason for the `NIO` prefixes is to "guarantee" that we won't clash with other people's types which would break SemVer in Swift :(

PR #1268, opened 2019-11-26
Out of pipeline byte stream decoding

richardalow:
Add NIOSingleStepByteToMessageDecoder and NIOSingleStepByteToMessageProcessor for processing byte streams outside of the channel pipeline.

Motivation:

ByteToMessageHandler works well when you can embed it in your pipeline, but if
you have a wrapped protocol (e.g. sending a protocol over HTTP), you can't use
it without losing the context of the outer protocol (e.g. HTTP headers/trailers).

Modifications:

Added NIOSingleStepByteToMessageDecoder and NIOSingleStepByteToMessageProcessor.

Result:

New functionality above.


Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,270 @@ extension MessageToByteHandler {
...
+            let message: Decoder.InboundOut?
+            if decodeMode == .normal {
+                message = try self.decoder.decode(buffer: &self.buffer!)
+                if message == nil, let maximumBufferSize = self.maximumBufferSize, self.buffer!.readableBytes > maximumBufferSize {
+                    throw ByteToMessageDecoderError.PayloadTooLargeError()
+                }
+            } else {
+                message = try self.decoder.decodeLast(buffer: &self.buffer!, seenEOF: seenEOF)
+            }
+            if message != nil {
```
Lukasa: This can be `if let` to avoid the force unwrap on the next line.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,270 @@ extension MessageToByteHandler {
...
+                if self.decoder.shouldReclaimBytes(buffer: self.buffer!) {
+                    self.buffer!.discardReadBytes()
+                }
+            }
+        }
+
+        if self.buffer!.readableBytes == 0 {
+            self.buffer!.discardReadBytes()
+        }
+    }
```
Lukasa: If we do some careful reordering of the body here we should be able to write this code in such a way that we can use conditional binding to get a non-optional copy of `buffer` instead of having to force unwrap the buffer all the time. This should lead to better code generation because we won't have to add separate branches and `ud2`s for each time we access `self.buffer`.

The trick here will be twofold. Firstly we'll need a local helper function:

```swift
private func withNonCoWBuffer(_ body: (inout ByteBuffer) -> Decoder.InboundOut?) -> Decoder.InboundOut? {
    guard var buffer = self.buffer else {
        return nil
    }

    self.buffer = nil  // To avoid CoW
    let result = body(&buffer)
    self.buffer = buffer
    return result
}
```

Then we can rewrite the function to move all of this logic _except_ the `if message != nil...else...` section into a call to `withNonCoWBuffer`. Then, finally, we can do the `if message != nil` check.

This should remove all the optionality around the buffer, remove the force unwraps, and lead to cleaner looking internal code as well as better code generation.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,270 @@ extension MessageToByteHandler {
...
+
+        // buffer can only be nil if we're called from finishProcessing which is handled above
+        assert(self.buffer != nil)
+
+        var decodingState: DecodingState = .continue
+        while self.buffer!.readableBytes > 0 && decodingState == .continue {
+            let message: Decoder.InboundOut?
+            if decodeMode == .normal {
+                message = try self.decoder.decode(buffer: &self.buffer!)
+                if message == nil, let maximumBufferSize = self.maximumBufferSize, self.buffer!.readableBytes > maximumBufferSize {
```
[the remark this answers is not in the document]
richardalow: That's not unreasonable, but I was attempting to copy the behavior of `ByteToMessageHandler`. It allows storing more than `maximumBufferSize` as long as it is read before returning `.needMoreData`. I'm not sure if that was intended though. It does mean that if you get a large buffer and process many messages out of it immediately, you don't throw. This seems reasonable since the memory was already allocated anyway in the `ByteBuffer` given to the handler, although it may be copied in `startProcessing`.

A more correct copy of the `ByteToMessageHandler` behavior for `SingleStepByteToMessageProcessor` would be to place this check at the end of `decodeLoop`. I propose to do this unless you feel strongly this should be in `append`.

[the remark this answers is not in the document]
Lukasa: I'm on the fence. @weissi?

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,266 @@ extension MessageToByteHandler {
...
+    private enum DecodeMode {
+        /// This is a usual decode, ie. not the last chunk
+        case normal
+        /// Last chunk
+        case last
+    }
+
+    private var decoder: Decoder
+    private let maximumBufferSize: Int?
+    internal private(set) var buffer: ByteBuffer?
```
glbrntt: Is this only `internal` for testing purposes? If so then please prefix `buffer` with an underscore.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,266 @@ extension MessageToByteHandler {
...
+            context.fireChannelRead(self.wrapInboundOut(message))
+            return .continue
+        } else {
+            return .needMoreData
+        }
+    }
+}
+
+
+/// `NIOSingleStepByteToMessageProcessor` uses a `NIOSingleStepByteToMessageDecoder` to produce messages
```
glbrntt: Great job on the documentation! 🥇

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,270 @@ extension MessageToByteHandler {
...
+    /// This method will be called in a loop until either the input `ByteBuffer` has nothing to read left or `nil` is returned. If non-`nil` is
+    /// returned and the `ByteBuffer` contains more readable bytes, this method will immediately be invoked again, unless `decodeLast` needs
+    /// to be invoked instead.
+    ///
+    /// - parameters:
+    ///     - buffer: The `ByteBuffer` from which we decode.
+    /// - returns: A message if one can be decoded or `nil` if it should be called again once more data is present in the `ByteBuffer`.
+    mutating func decode(buffer: inout ByteBuffer) throws -> InboundOut?
+
+    /// Decode from a `ByteBuffer` when no more data is incoming. This method is called in a loop only once.
```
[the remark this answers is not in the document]
richardalow: I agree, although I copied it from the `ByteToMessageDecoder` docs. I'll make it clearer.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,266 @@ extension MessageToByteHandler {
         }
     }
 }
+
+
+/// A simplified version of `ByteToMessageDecoder` that can generate zero or one messages for each invocation of `decode` or `decodeLast`.
+/// Having `decode` and `decodeLast` return an optional message avoids re-entrancy problems, since the functions relinquish exclusive access
+/// to the `ByteBuffer` when returning. This allows for greatly simplified processing.
+/// Many `ByteToMessageDecoder`'s can trivially be translated to `NIOSingleStepByteToMessageDecoder`'s.
+public protocol NIOSingleStepByteToMessageDecoder: ByteToMessageDecoder {
```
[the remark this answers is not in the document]
richardalow: As discussed offline, you can't add an inheritance clause in a protocol extension.

Sources/NIO/Codec.swift
```diff
@@ -980,52 +983,69 @@ public class NIOSingleStepByteToMessageProcessor<Decoder: NIOSingleStepByteToMes
...
+                        if self.decoder.shouldReclaimBytes(buffer: buffer) {
+                            buffer.discardReadBytes()
+                        }
+                    }
+                }
+                if decodeMode == .normal {
+                    return try self.decoder.decode(buffer: &buffer)
+                } else {
+                    return try self.decoder.decodeLast(buffer: &buffer, seenEOF: seenEOF)
                 }
```
Lukasa: If you split the body of the closure passed to `withNonCoWBuffer` into a separate function, we can rewrite this loop to use `while let` instead of `repeat...while`, which will remove yet another check on the nullity of `messageOpt`.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,284 @@ extension MessageToByteHandler {
...
+
+        while let message = try self.withNonCoWBuffer(decodeOnce) {
+            try messageReceiver(message)
+        }
+
+        if let maximumBufferSize = self.maximumBufferSize, self._buffer!.readableBytes > maximumBufferSize {
+            throw ByteToMessageDecoderError.PayloadTooLargeError()
+        }
+
+        if self._buffer!.readableBytes == 0 {
```
Lukasa: Can you place a comment above these two blocks explaining why these two force unwraps are safe? (This is an SSWG best practice)

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,270 @@ extension MessageToByteHandler {
...
 }
+
+
+/// A simplified version of `ByteToMessageDecoder` that can generate zero or one messages for each invocation of `decode` or `decodeLast`.
+/// Having `decode` and `decodeLast` return an optional message avoids re-entrancy problems, since the functions relinquish exclusive access
+/// to the `ByteBuffer` when returning. This allows for greatly simplified (and more performant) processing.
+/// Many `ByteToMessageDecoder`'s can trivially be translated to `NIOSingleStepByteToMessageDecoder`'s.
+public protocol NIOSingleStepByteToMessageDecoder: ByteToMessageDecoder {
+    /// The type of the messages this `NIOSingleStepByteToMessageDecoder` decodes to.
+    associatedtype InboundOut
```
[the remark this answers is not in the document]
richardalow: It seems that this isn't possible, at least without compiler warnings. I raised [a ticket](https://bugs.swift.org/browse/SR-11868) since I think this is a swift bug.

Tests/NIOTests/CodecTest.swift
```diff
@@ -1561,30 +1588,364 @@ public final class MessageToByteEncoderTest: XCTestCase {
...
+        XCTAssertNoThrow(try processor.finishProcessing(seenEOF: false, messageReceiver.receiveMessage))
+
+        XCTAssertEqual("12", messageReceiver.retrieveMessage().map {
+            String(decoding: $0.readableBytesView, as: Unicode.UTF8.self)
+        })
+        XCTAssertEqual("34", messageReceiver.retrieveMessage().map {
+            String(decoding: $0.readableBytesView, as: Unicode.UTF8.self)
+        })
+        XCTAssertEqual("56", messageReceiver.retrieveMessage().map {
+        String(decoding: $0.readableBytesView, as: Unicode.UTF8.self)
```
glbrntt:
```suggestion
            String(decoding: $0.readableBytesView, as: Unicode.UTF8.self)
```

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,285 @@ extension MessageToByteHandler {
...
+///
+/// When your stream ends, call `finishProcessing` to ensure all buffered data is passed to your decoder. This will call `decodeLast`
+/// one or more times with any remaining data.
+///
+/// ### Example
+///
+/// Below is an example of a protocol decoded by `TwoByteStringCodec` that is sent over HTTP. `RawBodyMessageHandler` forwards the headers
+/// and trailers directly and uses `NIOSingleStepByteToMessageProcessor` to send whole decoded messages.
+///
+///     class TwoByteStringCodec: NIOSingleStepByteToMessageDecoder {
```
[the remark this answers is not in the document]
richardalow: I think the extra spaces cause XCode to render as code - as it's currently written it shows as the code font and renders as code in the option-click popup.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,286 @@ extension MessageToByteHandler {
         }
     }
 }
+
+
+/// A simplified version of `ByteToMessageDecoder` that can generate zero or one messages for each invocation of `decode` or `decodeLast`.
```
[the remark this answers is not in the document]
weissi: I think we should also mention that you are _not_ supposed to implement the other `decode` methods.

Sources/NIO/Codec.swift
```diff
@@ -790,3 +790,286 @@ extension MessageToByteHandler {
...
+///
+///     let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+///     let bootstrap = ServerBootstrap(group: group).childChannelInitializer({channel in
+///         channel.pipeline.configureHTTPServerPipeline(withPipeliningAssistance: true, withErrorHandling: true).flatMap { _ in
+///             channel.pipeline.addHandlers([RawBodyMessageHandler(), DecodedBodyHTTPHandler()])
+///         }
+///     })
+///     let channelFuture = bootstrap.bind(host: "127.0.0.1", port: 0)
+///
+public class NIOSingleStepByteToMessageProcessor<Decoder: NIOSingleStepByteToMessageDecoder> {
```
weissi: could we make that `final`

Tests/NIOTests/CodecTest.swift
```diff
@@ -1561,30 +1588,361 @@ public final class MessageToByteEncoderTest: XCTestCase {
...
+        let decoder = Decoder()
+        let processor = NIOSingleStepByteToMessageProcessor(decoder, maximumBufferSize: max)
+        let messageReceiver: MessageReceiver<String> = MessageReceiver()
+
+        var buffer = allocator.buffer(capacity: max + 1)
+        buffer.writeString(String(repeating: "*", count: max + 1))
+        XCTAssertNoThrow(try processor.process(buffer: buffer, messageReceiver.receiveMessage))
+        XCTAssertNoThrow(try processor.finishProcessing(seenEOF: false, messageReceiver.receiveMessage))
+        XCTAssertEqual(0, processor._buffer!.readableBytes)
+        XCTAssertGreaterThan(decoder.decodeCalls, 0)
```
weissi: could we add one test that does cause re-entrancy? I know it's not an issue but we should prove it.

PR #1272, opened 2019-11-27
don't constantly check the heap property in debug mode

weissi:
Motivation:

So far we've been checking the heap property of our heaps _twice per
heap operation_. I think that's a bit much.

Modifications:

Remove the heap operation checks.

Result:

Runs much faster in debug mode.


Lukasa:
The assert is backup for “assuming the code is correct and tested”. The easiest way to grow confidence that the code is correct is to have it crash in debug mode if it isn’t. Using assertions this way (as a backstop for correctness) is a nice thing to do when obtaining complete test coverage is hard, as it tends to be for data structure work.

PR #1273, opened 2019-11-27
remove random print("hi \(path)") message

weissi:


weissi:
gotta say hi to those paths :)

PR #1277, opened 2019-11-28
document debugging with allocation tests

weissi:
Motivation:

The tooling around the allocation counter tests is pretty opaque and
also not easy to use.

Modifications:

Document what interested folks should know.

Result:

More people will be able to debug allocations in SwiftNIO.

### DIFF
--- a/docs/debugging-allocations.md
+++ b/docs/debugging-allocations.md
@@ -1,0 +1,181 @@
+# How to debug allocation regressions
+
+Usually, your journey will start by detecting that an allocation test regressed. So let's first start by looking at how to run the allocation counter tests.
+
+## Running allocation counter tests
+
+The allocation counter tests are usually run by NIO's integration test framework which lives in `IntegrationTests/`. In there, you first want to identify the test suite and test case that runs all the allocation counter tests. For the `swift-nio` repository that is
+
+    IntegrationTests/tests_04_performance/test_01_allocation_counts.sh
+
+Next to the test case (`test_01_allocation_counts.sh`) that runs the allocation counter tests should be a resource directory which contains the individual allocation counter tests. For `swift-nio` that is
+
+    IntegrationTests/tests_04_performance/test_01_resources
+
+From here on, this document expects that you are within this directory, so you might want to start with
+
+    cd IntegrationTests/tests_04_performance/test_01_resources/
+
+To run all the allocation tests, simply invoke the runner script in there
+
+    ./run-nio-alloc-counter-tests.sh
+
+The invocation of this script will take a while because it has to build all of SwiftNIO in release mode first, then it compiles the integration tests and runs them all, multiple times...
+
+### Understanding the output of the allocation counter tests
+
+The output of this script will look something like
+
+```
+- /Users/johannes/devel/swift-nio/IntegrationTests/tests_04_performance/test_01_resources/test_future_lots_of_callbacks.swift
+test_future_lots_of_callbacks.remaining_allocations: 0
+test_future_lots_of_callbacks.total_allocations: 75001
+test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+DEBUG: [["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "remaining_allocations": 0, "total_allocated_bytes": 4138056], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "remaining_allocations": 0, "total_allocated_bytes": 4138056], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["total_allocations": 75001, "total_allocated_bytes": 4138056, "remaining_allocations": 0], ["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "total_allocated_bytes": 4138056, "remaining_allocations": 0]]
+```
+
+with this kind of block repeated for each allocation counter test. Let's go through and understand each line. The first line is the name of the specific allocation test. The most relevant part is the file name (`test_future_lots_of_callbacks.swift`). For this test, we seem to be testing how many allocations futures with many callbacks are doing.
+
+```
+- /Users/johannes/devel/swift-nio/IntegrationTests/tests_04_performance/test_01_resources/test_future_lots_of_callbacks.swift
+```
+
+Next, we see
+
+```
+test_future_lots_of_callbacks.remaining_allocations: 0
+test_future_lots_of_callbacks.total_allocations: 75001
+test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+```
+
+which are the aggregate values. The first line says `remaining_allocations: 0`, which means that this allocation test didn't leak, which is good! Then, we see
+
+    test_future_lots_of_callbacks.total_allocations: 75001
+
+which means that by the end of the test, we saw 75001 allocations in total. As a rule, we usually run the workload of every allocation test 1000 times. That means you want to divide 75001 by the 1000 runs to get the number of allocations per run. In other words: each iteration of this allocation test allocated 75 times. The extra one allocation is just some noise that we have to ignore: these are usually inserted by operations in the Swift runtime that need to initialize some state. In many (especially multi-threaded test cases) there is some noise, which is why we run the workload repeatedly, which makes it easy to tell the signal from the noise.
+
+Finally, we see
+
+    test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+
+Which is the total number of bytes allocated by this test.
+
+
+Last of all, we see
+
+```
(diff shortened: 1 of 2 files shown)


weissi:
sorry guys,  this is still a bit rough but I wanted to get something out ASAP for @2bjake :)

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+which this kind of block repeated for each allocation counter test. But let's go through each line. The first line is the name of the allocation test, most relevant is the file name (`test_future_lots_of_callbacks.swift`). So we seem to be testing how many allocations futures with many callbacks are doing.
```
Lukasa:
```suggestion
with this kind of block repeated for each allocation counter test. Let's go through and understand each line. The first line is the name of the specific allocation test. The most relevant part is the file name (`test_future_lots_of_callbacks.swift`). For this test, we seem to be testing how many allocations futures with many callbacks are doing.
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+
+Next, we see
+
+```
+test_future_lots_of_callbacks.remaining_allocations: 0
+test_future_lots_of_callbacks.total_allocations: 75001
+test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+```
+
+which are the aggregates values: At the end of the test, there are `remaining_allocations: 0` which means that this allocation test didn't leak, good! Then, we see
```
Lukasa:
```suggestion
which are the aggregate values. The first line says `remaining_allocations: 0`, which means that this allocation test didn't leak, which is good! Then, we see
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+
+    test_future_lots_of_callbacks.total_allocations: 75001
+    
+Which means that at the end of the test, we saw 75001 allocations in total. As a rule, we usually run the workload of every allocation test 1000 times. That means, you want to divide 75001 by the 1000 runs. In other words: Each iteration of this allocation test allocated 75 times. The extra one allocation is just some noise that we have to ignore. In many (especially multi-threaded test cases) there is some noise which is why we run the workload usually 1000 times which makes it easy to tell the signal from the noise.
```
Lukasa:
```suggestion
which means that by the end of the test, we saw 75001 allocations in total. As a rule, we usually run the workload of every allocation test 1000 times. That means you want to divide 75001 by the 1000 runs to get the number of allocations per run. In other words: each iteration of this allocation test allocated 75 times. The extra one allocation is just some noise that we have to ignore: these are usually inserted by operations in the Swift runtime that need to initialize some state. In many (especially multi-threaded test cases) there is some noise, which is why we run the workload repeatedly, which makes it easy to tell the signal from the noise.
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+
+Finally, we see
+
+    test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+
+Which is the total number of bytes allocated by this test.
+
+
+Below, we see
```
Lukasa:
```suggestion
Last of all, we see
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+which is literally just the exact numbers of each of the 10 runs we considered. If you see a lot of fluctuation between the runs, then the allocation test doesn't allocate deterministically which would be something you want to fix. In the example above, we're totally stable to the last byte of allocated memory though.
```
Lukasa:
```suggestion
which is literally just the exact numbers of each of the 10 runs we considered. If you see a lot of fluctuation between the runs, then the allocation test doesn't allocate deterministically which would be something you want to fix. In the example above, we're totally stable to the last byte of allocated memory, so we have nothing to worry about with this test.
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+### Debugging with Instruments
+
+1. Start Instruments
+2. Choose the "Allocations" instrument
+3. Tap the executable selector in the toolbar
+4. Click "Choose Target..."
+5. Tick "Show Hidden Files"
+6. Press Cmd+Shift+G
+7. Enter the temporary directory path of a non-hooked version of the allocation counter tests (for example `/private/tmp/.nio_alloc_counter_tests_0UDNXS`)
+8. select `.build/release/`
```
Lukasa:
```suggestion
8. Select `.build/release/`
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+9. find the binary that is named like your allocation counter test (eg. `test_lots_of_future_callbacks`)
```
Lukasa:
```suggestion
9. Find the binary that is named like your allocation counter test (eg. `test_lots_of_future_callbacks`)
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+10. Click "Choose"
+11. Press the record button
+
+Now you should have a full Instruments allocations trace of the test run. To make sense of what's going on, switch the allocation lifespan from "Created & Persisted" to "Created & Destroyed". After that you should be able to see the type of the allocation and how many times it got allocated. Clicking on the little arrow next to the type name will reveal each individual allocation of this type and the responsible stack trace.
+
+### Debugging with `malloc-aggregation.d` (macOS only)
+
+SwiftNIO ships a `dtrace` script called `dev/malloc-aggregation.d` which can give you an aggregation of all allocations by stack trace. This means that you will see the stack trace of each allocation that happened and how many times this particular stack trace allocated.
+
+But let's see an example on how to use that:
```
Lukasa:
```suggestion
Here's an example of how to use this script:
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+              libswiftObjectiveC.dylib`autoreleasepool<A>(invoking:)+0x2e
+              test_future_lots_of_callbacks`measureAll(_:)+0x159
+              test_future_lots_of_callbacks`measureAndPrint(desc:fn:)+0x2d
+              test_future_lots_of_callbacks`main+0x9
+              libdyld.dylib`start+0x1
+              test_future_lots_of_callbacks`0x1
+            30000
+```
+
+this means that the above stack trace is responsible for `30000` allocations. Remember that the allocation counter tests run each test 10 times, so this means that in each test run, the above stack trace allocated 3000 times. Also remember that we usually run the workload we want to test 1000 times. That means that in the real world, the above stack trace accounts for 3 out of the 75 total allocations (30000 allocations / 10 runs / 1000 executions per run = 3).
```
Lukasa:
```suggestion
repeated many times. Each block means that the specific stack trace is responsible for some number of allocations: in the example above, `30000` allocations. Remember that the allocation counter tests run each test 10 times, so this means that in each test run, the above stack trace allocated 3000 times. Also remember that we usually run the workload we want to test 1000 times. That means that in the real world, the above stack trace accounts for 3 out of the 75 total allocations (30000 allocations / 10 runs / 1000 executions per run = 3).
```

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+
+    test_future_lots_of_callbacks.total_allocated_bytes: 4138056
+
+Which is the total number of bytes allocated by this test.
+
+
+Below, we see
+
+```
+DEBUG: [["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "remaining_allocations": 0, "total_allocated_bytes": 4138056], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "remaining_allocations": 0, "total_allocated_bytes": 4138056], ["total_allocated_bytes": 4138056, "total_allocations": 75001, "remaining_allocations": 0], ["total_allocations": 75001, "total_allocated_bytes": 4138056, "remaining_allocations": 0], ["remaining_allocations": 0, "total_allocations": 75001, "total_allocated_bytes": 4138056], ["total_allocations": 75001, "total_allocated_bytes": 4138056, "remaining_allocations": 0]]
```
glbrntt: Is the `DEBUG:` just so we can identify that line so we can validate the numbers? Might be worth mentioning it's not related to the build configuration.

docs/debugging-allocations.md
```diff
@@ -0,0 +1,187 @@
...
+Fetching /private/tmp/.nio_alloc_counter_tests_5jMMhk/HookedFunctions
+Fetching /private/tmp/.nio_alloc_counter_tests_5jMMhk/AtomicCounter
+```
+
+If you change your working directory to common directory above (`/private/tmp/.nio_alloc_counter_tests_5jMMhk`), then you will find a regular Swift package there and you can just re-run the allocation counter tests using
+
+```
+cd /tmp/.nio_alloc_counter_tests_5jMMhk
+swift run -c release
+test_future_lots_of_callbacks.total_allocated_bytes: 4138056
```
glbrntt: I don't think the test output is necessary here; the important info is `swift run -c release`

docs/debugging-allocations.md
weissi: good point, added

docs/debugging-allocations.md
weissi: good point, done

weissi:
thanks so much @Lukasa and @glbrntt for the reviews, took all feedback.

richardalow:
Very nice, it all makes sense to a newbie.

PR #1282, opened 2019-12-04
Add first(name:) to HTTPHeaders

glbrntt:
Motivation:

In some cases you only expect (or care about) a single value for a
header. However, getting this value without allocating an array or
traversing the whole array requres using first(where:) provided by
Sequence. This is not very ergonomic as it requires users to do the
case-insensitive comparison and then exctact the value from the
HTTPHeaders. Additionally, users cannot use NIO's case insensitive
ASCII comparison.

Modifications:

- Provide a first(name:) method on HTTPHeaders which returns the value
  from the first header whose name is a case insentive match of the
  subscript parameter.

Result:

It's easier to get the first value from HTTPHeaders matching a given
name.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -420,6 +420,25 @@ public struct HTTPHeaders: CustomStringConvertible, ExpressibleByDictionaryLiter
         }
     }
 
+    /// Retrieves the first value for a given header field name from the block.
+    ///
+    /// This method uses case-insensitive comparisons for the header field name. It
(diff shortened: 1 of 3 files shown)


glbrntt:
See also: https://github.com/apple/swift-nio-http2/pull/178

Lukasa:
Huh, segfault in the tests. That's interesting.

glbrntt:
> Huh, segfault in the tests. That's interesting.

Yup https://github.com/apple/swift-nio/issues/1283 -- running it in a loop with lldb, let's see what happens!

glbrntt:

Issue #1275, opened 2019-11-28
use unchecked arithmetic operations in many places in CircularBuffer


glbrntt:
[@weissi with George's account]

CircularBuffer uses checked arithmetic (`+`, `-`) in many places where we know that an under/overflow cannot happen. We should replace those operations with `&+` and `&-` to kill all the branches/extra code of the over/underflow checking where possible safely.

PR #1284, opened 2019-12-04
Replaced CircularBuffer's uses of checked arithmetic (+, -)

3a4oT:
Replaced CircularBuffer's uses of checked arithmetic (+, -) with `&+` and `&-`. 
PS. This is my first attempt to contribute so I'm open to any suggestions especially I do not understand how I can verify it makes performance better.

### Motivation:

trying to address  #1275

### Modifications:

Replaced checked arithmetic operations (+, -) with unchecked *&-* and  *&+*

### Result:

better performance

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -27,7 +27,7 @@ public struct CircularBuffer<Element>: CustomStringConvertible {
 
     @inlinable
     internal var mask: Int {
-        return self._buffer.count - 1
+        return self._buffer.count &- 1
     }
 
     @inlinable
@@ -52,7 +52,7 @@ public struct CircularBuffer<Element>: CustomStringConvertible {
 
     @inlinable
     internal func indexAdvanced(index: Int, by: Int) -> Int {
-        return (index + by) & self.mask
+        return (index &+ by) & self.mask
     }
 
     /// An opaque `CircularBuffer` index.
@@ -211,13 +211,13 @@ extension CircularBuffer: Collection, MutableCollection {
 
         switch (start.isIndexGEQHeadIndex, end.isIndexGEQHeadIndex) {
         case (true, true):
-            return end.backingIndex - start.backingIndex
+            return end.backingIndex &- start.backingIndex
         case (true, false):
-            return backingCount - (start.backingIndex - end.backingIndex)
+            return backingCount &- (start.backingIndex &- end.backingIndex)
         case (false, true):
-            return -(backingCount - (end.backingIndex - start.backingIndex))
+            return -(backingCount &- (end.backingIndex &- start.backingIndex))
         case (false, false):
-            return end.backingIndex - start.backingIndex
+            return end.backingIndex &- start.backingIndex
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Sources/NIO/CircularBuffer.swift
```diff
@@ -589,7 +589,7 @@ extension CircularBuffer: RangeReplaceableCollection {
             self.removeSubrange(subrange)
         } else {
             var newBuffer: ContiguousArray<Element?> = []
-            let neededNewCapacity = self.count + newElements.count - subrangeCount + 1 /* always one spare */
+            let neededNewCapacity = self.count &+ newElements.count &- subrangeCount &+ 1 /* always one spare */
```
weissi: I think in this case we probably want the checks as we're growing. I mean we will never be able to get to that many elements but I think it'd still be the right thing to do.

Sources/_NIO1APIShims/NIO1APIShims.swift
```diff
@@ -395,12 +395,12 @@ extension CircularBuffer {
 
     @available(*, deprecated, message: "please use CircularBuffer.Index instead of Int")
     public func index(after: Int) -> Int {
-        return after + 1
+        return after &+ 1
```
weissi: these ones are deprecated, so I think we might want to take this out

Sources/_NIO1APIShims/NIO1APIShims.swift
```diff
@@ -395,12 +395,12 @@ extension CircularBuffer {
...
     }
 
     @available(*, deprecated, message: "please use CircularBuffer.Index instead of Int")
     public func index(before: Int) -> Int {
-        return before - 1
+        return before &- 1
```
weissi: these ones are deprecated, so I think we might want to take this out

Sources/NIO/CircularBuffer.swift
3a4oT: fixed in a follow-up commit. Thanks

Sources/NIO/CircularBuffer.swift
```diff
@@ -227,7 +227,7 @@ extension CircularBuffer: RandomAccessCollection {
     /// Returns the index offset by `distance` from `index`.
     @inlinable
     public func index(_ i: Index, offsetBy distance: Int) -> Index {
-        return .init(backingIndex: (i.backingIndex + distance) & self.mask,
+        return .init(backingIndex: (i.backingIndex &+ distance) & self.mask,
```
glbrntt: I don't think this one can be left unchecked since `distance` be arbitrarily large. cc @weissi

weissi: @glbrntt I _think_ it's undefined behaviour if you ask for an OOB index, let me check the docs.

weissi: @glbrntt yes, quoting the [docs](https://developer.apple.com/documentation/swift/bidirectionalcollection/3024450-index)

> The value passed as distance must not offset i beyond the bounds of the collection.

weissi: so I think we're fine here.

glbrntt: ah okay, sounds good to me

weissi:

PR #1287, opened 2019-12-04
fix AtomicBox by deprecating it and implementing it with a CAS loop

weissi:
Motivation:

Atomic references cannot easily be maintained correctly because the
`load` operation gets a pointer value that might (before the loader gets
a change to retain the returned reference) be destroyed by another
thread.

Modifications:

- Deprecate `AtomicBox`
- Implement `NIOCASLoopBox` with functionality similar to AtomicBox (but
  implemented using a CAS loop :( ).

Result:

- fixes #1286
- fixes #1283

Issue #1286, opened 2019-12-04
AtomicBox.load() suffers from ABA problem


weissi:
`AtomicBox.load()` suffers from an ABA problem where the loaded value may be dead before it gets to increase the reference count in

https://github.com/apple/swift-nio/blob/master/Sources/NIOConcurrencyHelpers/atomics.swift#L462-L466

PR #1289, opened 2019-12-05
stop using AtomicBox

weissi:
Motivation:

AtomicBox doesn't properly work (#1286) and despite fixing it in #1287,
it's really not worth using AtomicBox (which as of #1287 instroduces a
CAS loop) anymore.

Modifications:

Remove the two uses of AtomicBox.

Result:

No usage of broken/slow data types.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -205,18 +205,84 @@ private struct SocketChannelLifecycleManager {
 class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, ChannelCore {
     typealias SelectableType = SocketType.SelectableType
 
-    // MARK: Stored Properties
-    // Visible to access from EventLoop directly
+    struct AddressCache {
+        // deliberately lets because they must always be updated together (so forcing `init` is useful).
+        let local: Optional<SocketAddress>
+        let remote: Optional<SocketAddress>
+
+        init(local: SocketAddress?, remote: SocketAddress?) {
+            self.local = local
+            self.remote = remote
+        }
+    }
+
+    // MARK: - Stored Properties
+    // MARK: Constants & atomics (accessible everywhere)
     public let parent: Channel?
     internal let socket: SocketType
     private let closePromise: EventLoopPromise<Void>
     internal let selectableEventLoop: SelectableEventLoop
-    private let addressesCached: AtomicBox<Box<(local:SocketAddress?, remote:SocketAddress?)>> = AtomicBox(value: Box((local: nil, remote: nil)))
-    private let bufferAllocatorCached: AtomicBox<Box<ByteBufferAllocator>>
+    private let _offEventLoopLock = Lock()
     private let isActiveAtomic: NIOAtomic<Bool> = .makeAtomic(value: false)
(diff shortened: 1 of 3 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -205,19 +205,82 @@ private struct SocketChannelLifecycleManager {
...
 
+    // MARK: Variables, on EventLoop thread only
+    var readPending = false
+    var pendingConnect: Optional<EventLoopPromise<Void>>
+    var recvAllocator: RecvByteBufferAllocator
+    var maxMessagesPerRead: UInt = 4
+    private var inFlushNow: Bool = false // Guard against re-entrance of flushNow() method.
+    private var autoRead: Bool = true
+
+    // MARK: Varibles that are really constants
```
Lukasa: Nit: missing 'a' in `variables`

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -205,19 +205,84 @@ private struct SocketChannelLifecycleManager {
...
+    private var _pipeline: ChannelPipeline! = nil // this is really a constant (set in .init) but needs `self` to be constructed and therefore a `var`. Do not change as this needs to accessed from arbitrary threads
+
+    // MARK: Special variables, please read comments.
+    // For reads guarded by _offEventLoopLock/EL thread; writes guarded by _offEventLoopLock (and on EL thread only).
```
Lukasa: Can we reword this slightly to make it clearer? What we want to say, as I understand it, is:

> "Reads require holding either the `_offEventLoopLock` or EL thread, writes require holding both.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -205,19 +205,84 @@ private struct SocketChannelLifecycleManager {
...
+    // PLEASE don't use these directly and use the non-underscored computed properties instead.
+    private var _addressCache = AddressCache(local: nil, remote: nil) // please use `self.addressesCached` instead
+    private var _bufferAllocatorCache: ByteBufferAllocator  // please use `self.bufferAllocatorCached` instead.
+
+    // MARK: - Computed properties
+    // This is called from arbirary threads.
```
Lukasa: Nit: "arbitrary"

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -205,19 +205,84 @@ private struct SocketChannelLifecycleManager {
...
+            if self.eventLoop.inEventLoop {
+                return self._addressCache
+            } else {
+                return self._offEventLoopLock.withLock {
+                    return self._addressCache
+                }
+            }
+        }
+        set {
+            assert(self.eventLoop.inEventLoop)
```
Lukasa: Please use `assertInEventLoop`, but in this case given how important this is to the safety of the program I think this should be `preconditionInEventLoop`.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -205,19 +205,84 @@ private struct SocketChannelLifecycleManager {
...
+            if self.eventLoop.inEventLoop {
+                return self._bufferAllocatorCache
+            } else {
+                return self._offEventLoopLock.withLock {
+                    return self._bufferAllocatorCache
+                }
+            }
+        }
+        set {
+            self.eventLoop.assertInEventLoop()
```
Lukasa: Make this `preconditionInEventLoop`

PR #1292, opened 2019-12-05
optimise future callback collection [1]

weissi:
Motivation:

The futures implementation used a particularly inefficient way to
accumulate all callbacks.

Modifications:

Improved this by allocating fewer arrays.

Result:

Fewer allocations.


weissi:
Closing for #1293 which is better

Issue #1257, opened 2019-11-20
Library Evolution support


AMatecki:
Hi, I'm using SwiftNIO 2.10.1 as a library attached to tvOS app's UI test to provide mocked backend. Since I'm distributing it as a framework I would really like not to have it rebuild with every new Swift version. To have it working SwiftNIO would have to support Library Evolution. I tried turning it on, but had a lot of errors like this:
`'let' property '_storage' may not be initialized directly; use "self.init(...)" or "self = ..." instead`
in multiple files. I'm not sure how to fix it by myself. Is there a plan to have Library Evolution implemented in SwiftNIO?

weissi:
@AMatecki are you exporting SwiftNIO types or functions as public API in _your framework_?

I would assume no and if that's the case, can you try `@_implementationOnly import NIO`?

We will still have a look into why it fails to compile.

AMatecki:
Thanks for your suggestion, I'm new to working with Swift libraries, didn't knew about `@_implementationOnly`. Unfortunately I'm exporting some types from SwiftNIO, so I can't compile with `@_implementationOnly`.

weissi:
@AMatecki Right. So right now, SwiftNIO's primary and only distribution mechanism is as a SwiftPM package and SwiftPM right now only supports source distribution. SwiftNIO also doesn't guarantee an ABI because in a source-only world, having an ABI guarantee doesn't mean much because everybody will recompile every time.

Are you making an `.xcodeproj` from SwiftNIO or are you depending on SwiftNIO using Xcode's SwiftPM Package support (File -> Swift Package -> Add Package Dependency)?

AMatecki:
My library was created before Xcode's SPM support, so I'm creating .xcodeproj from command line. I could include SwiftNIO package to my testing target, but it would complicate my project's build pipeline, so I prefer to attach it as a framework. Would including it using Xcode change anything?

weissi:
@AMatecki just to be clear: What you're doing is unsupported but given that you're using it in tests that may be fine. It's unsupported because you're relying on a stable binary interface (ABI) for a source-only package that is only guaranteeing stable APIs. This will only work if you're recompiling everything whenever you update SwiftNIO. So updates of the SwiftNIO containing framework _without_ recompiling the binaries that use it may break.

But in order for you to make progress on this issue, here are a few options that you may want to do.

## Options

There are three categories of options and you only need to pick one options out of one category. Each of the options should be able to resolve your immediate issue.

[...]

PR #1298, opened 2019-12-09
remove phantom publics from EventLoop and Heap

weissi:
Motivation:

In SelectableEventloop, Heap, PriorityQueue, etc there were quite a few
phantom publics left which cause issues with #1257 and are confusing.

Modifications:

- remove the phantom publics
- move SelectableEventLoop to its own file

Result:

- clearer what's going on (public means public)

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -632,454 +632,6 @@ enum NIORegistration: Registration {
     }
 }
 
-/// Execute the given closure and ensure we release all auto pools if needed.
-private func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
-    #if os(Linux)
-    return try execute()
-    #else
-    return try autoreleasepool {
-        try execute()
-    }
-    #endif
-}
-
-/// `EventLoop` implementation that uses a `Selector` to get notified once there is more I/O or tasks to process.
-/// The whole processing of I/O and tasks is done by a `NIOThread` that is tied to the `SelectableEventLoop`. This `NIOThread`
-/// is guaranteed to never change!
-@usableFromInline
-internal final class SelectableEventLoop: EventLoop {
-    /// The different state in the lifecycle of an `EventLoop` seen from _outside_ the `EventLoop`.
-    private enum ExternalState {
-        /// `EventLoop` is open and so can process more work.
-        case open
-        /// `EventLoop` is currently in the process of closing.
-        case closing
-        /// `EventLoop` is closed.
-        case closed
-        /// `EventLoop` is closed and is currently trying to reclaim resources (such as the EventLoop thread).
-        case reclaimingResources
-        /// `EventLoop` is closed and all the resources (such as the EventLoop thread) have been reclaimed.
-        case resourcesReclaimed
-    }
-
-    /// The different state in the lifecycle of an `EventLoop` seen from _inside_ the `EventLoop`.
-    private enum InternalState {
-        case runningAndAcceptingNewRegistrations
-        case runningButNotAcceptingNewRegistrations
-        case noLongerRunning
-    }
-
-    private let selector: NIO.Selector<NIORegistration>
-    private let thread: NIOThread
-    private var scheduledTasks = PriorityQueue<ScheduledTask>(ascending: true)
-    private var tasksCopy = ContiguousArray<() -> Void>()
-
-    private let tasksLock = Lock()
-    private let _externalStateLock = Lock()
-    private var externalStateLock: Lock {
-        // The assert is here to check that we never try to read the external state on the EventLoop unless we're
-        // shutting down.
-        assert(!self.inEventLoop || self.internalState != .runningAndAcceptingNewRegistrations,
-               "lifecycle lock taken whilst up and running and in EventLoop")
-        return self._externalStateLock
-    }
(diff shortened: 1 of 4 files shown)


Sources/NIO/SelectableEventLoop.swift
```diff
@@ -0,0 +1,455 @@
...
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+import NIOConcurrencyHelpers
+
+/// `EventLoop` implementation that uses a `Selector` to get notified once there is more I/O or tasks to process.
+/// The whole processing of I/O and tasks is done by a `NIOThread` that is tied to the `SelectableEventLoop`. This `NIOThread`
+/// is guaranteed to never change!
+@usableFromInline
+internal final class SelectableEventLoop: EventLoop {
```
weissi: this stuff is just moved apart from `s/public/internal` and fallout like `@usableFromInline` etc.

Sources/NIO/EventLoop.swift
```diff
@@ -633,7 +633,8 @@ enum NIORegistration: Registration {
 }
 
 /// Execute the given closure and ensure we release all auto pools if needed.
-private func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
+@inlinable
+internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
```
Lukasa: Rather than make this `internal` and `@inlinable`, is there any reason not simply to move it to the file with its new usage?

Sources/NIO/EventLoop.swift
```diff
@@ -1366,16 +938,19 @@ private final class ScheduledTask {
...
 extension ScheduledTask: CustomStringConvertible {
+    @usableFromInline
     var description: String {
-        return "ScheduledTask(readyTime: \(self.readyTime))"
+        return "ScheduledTask(readyTime: \(self._readyTime))"
     }
 }
 
 extension ScheduledTask: Comparable {
+    @inlinable
```
Lukasa: Why does this need to be `inlinable`? This object is purely internal and only ever compared in `PriorityQueue.push`, which should be automatically specialised inside NIO.

Sources/NIO/EventLoop.swift
```diff
@@ -1366,16 +938,19 @@ private final class ScheduledTask {
...
     static func < (lhs: ScheduledTask, rhs: ScheduledTask) -> Bool {
-        return lhs.readyTime < rhs.readyTime
+        return lhs._readyTime < rhs._readyTime
     }
 
+    @inlinable
```
Lukasa: Similarly, why does this need to be `inlinable`?

Sources/NIO/PriorityQueue.swift
```diff
@@ -12,46 +12,56 @@
...
-    public mutating func remove(_ key: Element) {
-        self.heap.remove(value: key)
+    @inlinable
+    internal mutating func remove(_ key: Element) {
+        self._heap.remove(value: key)
     }
 
-    public mutating func push(_ key: Element) {
-        self.heap.append(key)
+    @inlinable
```
Lukasa: Similarly, what are the gains of making this function `@inlinable`?

Lukasa: In fact isn't this never called from anywhere that is itself `@inlinable`? `SelectableEventLoop.schedule0` is `@usableFromInline`, and `EmbeddedEventLoop`'s `push`/`pop` callers aren't `@inlinable` either, so I'd have assumed this attribute does nothing.

Sources/NIO/EventLoop.swift
weissi: @Lukasa I can revert it but I didn't see a reason not to make it available in tests etc.

Sources/NIO/EventLoop.swift
weissi: semi-automatic ;). These days we don't really think about every single inlinable anymore but I can take it out if you want.

Sources/NIO/PriorityQueue.swift
weissi: I'd leave this one. I just haven't done the work yet for everything in the heap to be inlinable too because right now it doesn't help with anything.

Sources/NIO/EventLoop.swift
weissi: @Lukasa also it now needs to be usableFromInline/inlinable which requires internal, I'll move it to the SelectableEventLoop file though

Sources/NIO/EventLoop.swift
weissi: removed, was  semi-automatic replacement

Sources/NIO/PriorityQueue.swift
Lukasa: I mean, I don't care about leaving it in, but I don't think it'll ever help. The heap is only ever used with one concrete type, defined in NIO so it'll always be specialised, and these call sites are several layers deep in the call stack such that their inability to be inlined is basically irrelevant.

weissi: @Lukasa yes, only if we'd ever re-expose PriorityQueue or if someone copied the code. I thought if we need to stick an attribute here, `@inlinable` is the better one. But 100% agreed it doesn't help with anything right now.

PR #1301, opened 2019-12-11
SelectableEventLoop: Call wakeup less frequently

weissi:
Motivation:

Previously, we unconditionally woke up the selector for every task that
was added. But this is only necessary if we're not already on the
EventLoop. Because if we are, we'll have another look at the outstanding
tasks anyway.

Modifications:

Remove selector wakeups.

Result:

Faster executes.


weissi:

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -202,7 +202,7 @@ internal final class SelectableEventLoop: EventLoop {
             self._tasksLock.withLockVoid {
                 self._scheduledTasks.remove(task)
             }
-            self._wakeupSelector()
+            // we don't need to wake up the selector here, the scheduled task will never be picked up.
```
Lukasa: The wake up here was to potentially recalculate the timeout. It might be nice to clarify that in this comment: even if this was the next task we were going to process, we’ll wake up once unnecessarily which isn’t too expensive, whereas waking up the selector every time is very expensive.

weissi: great idea! done.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -202,7 +202,11 @@ internal final class SelectableEventLoop: EventLoop {
             self._tasksLock.withLockVoid {
                 self._scheduledTasks.remove(task)
             }
-            self._wakeupSelector()
+            // We don't need to wake up the selector here, the scheduled task will never be picked up. Waking up the
+            // selector would mean that we may be able to recalculate the shutdown to a later date. The cost of not
+            // doing the recalculation is one potentially unnecessary wakeup which is exactly what we're
+            // saving here. So in the worst case, we didn't do a performance optimisation, in the best case, we saved
+            // one.
```
Davidde94: Nit: Saved one calculation?

weissi: one wakeup, added.

Issue #1217, opened 2019-11-01
Remove the binaries from the list of products.


Lukasa:
For some reason, in the past, when we started writing example binaries we added them all to the `.products` for our package. That isn’t really right: they aren’t a formal product of swift-nio. Having them there implies that they are versioned in some way. And it can lead to unnecessary noise in Xcode.

We should remove them from the `products`. This will have the effect of preventing them from being automatically built by `swift build`, but that’s ok: it’ll probably break one or two of our scripts, but it’s not the end of the world.

shekhar-rajak:
Hi,

I am interested to contribute to the project and trying to understand the codebase. 
For this issue do we just have to remove the `Products`  folder from the root directory and check that all the test cases are passing?

Lukasa:
Sorry @shekhar-rajak, I believe this was resolved by #1303.

PR #1303, opened 2019-12-12
Remove example binaries from products

johnlinvc:
Motivation:

Avoid examples being build when using as library.
Address https://github.com/apple/swift-nio/issues/1217

Modifications:

Remove example binaries from products.
Update example section in README, show how to run them.

Result:

Examples are not being built when running `swift build`


Lukasa:
@weissi how do we want to rate the semver impact of this?

Lukasa:

weissi:
@Lukasa let's do minor :)

PR #1305, opened 2019-12-14
fix EVFILT_EXCEPT loop if autoRead is off on Darwin

weissi:
Motivation:

Due to rdar://53656794, we sometimes (especially on UNIX Domain Sockets)
see EVFILT_EXCEPT firing when data is available to read depite the fact
that we're setting the low watermark to CInt.max. This means that
EVFILT_EXCEPT becomes essentially unusable.

Modifications:

- Don't use EVFILT_EXCEPT on Darwin
- We lose early EOF delivery on Darwin

Result:

Fewer possibilities of a spin loop on Darwin.

### DIFF
--- a/Sources/NIO/SelectableEventLoop.swift
+++ b/Sources/NIO/SelectableEventLoop.swift
@@ -53,7 +53,7 @@ internal final class SelectableEventLoop: EventLoop {
         case noLongerRunning
     }
 
-    private let selector: NIO.Selector<NIORegistration>
+    /* private but tests */ internal let _selector: NIO.Selector<NIORegistration>
     private let thread: NIOThread
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>(ascending: true)
@@ -105,7 +105,7 @@ internal final class SelectableEventLoop: EventLoop {
     }
 
     internal init(thread: NIOThread) throws {
-        self.selector = try NIO.Selector()
+        self._selector = try NIO.Selector()
         self.thread = thread
         self._iovecs = UnsafeMutablePointer.allocate(capacity: Socket.writevLimitIOVectors)
         self._storageRefs = UnsafeMutablePointer.allocate(capacity: Socket.writevLimitIOVectors)
@@ -156,7 +156,7 @@ internal final class SelectableEventLoop: EventLoop {
             throw EventLoopError.shutdown
         }
 
-        try channel.register(selector: self.selector, interested: channel.interestedEvent)
+        try channel.register(selector: self._selector, interested: channel.interestedEvent)
     }
 
     /// Deregister the given `SelectableChannel` from this `SelectableEventLoop`.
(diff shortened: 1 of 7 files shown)


Sources/NIO/SelectableEventLoop.swift
```diff
@@ -53,7 +53,7 @@ internal final class SelectableEventLoop: EventLoop {
         case noLongerRunning
     }
 
-    private let selector: NIO.Selector<NIORegistration>
+    internal let _selector: NIO.Selector<NIORegistration>
```
normanmaurer: consider adding a comment that this is not private just because of testing.

Tests/NIOTests/StreamChannelsTest.swift
```diff
@@ -246,6 +247,85 @@ class StreamChannelTest: XCTestCase {
...
         XCTAssertNoThrow(try forEachCrossConnectedStreamChannelPair(runTest))
     }
+
+    func testLotsOfWritesWhilstOtherSideNotReading() {
+        // This is a regression test for a problem where we would spin on EVFILT_EXCEPT despite the fact that there
+        // was no EOF or any other exceptional event present. So this is a regression test for rdar://53656794 and https://github.com/apple/swift-nio/pull/526.
+        class FailOnReadHandler: ChannelInboundHandler {
+            typealias InboundIn = ByteBuffer
+
+            let areReadOkayNow: NIOAtomic<Bool>
```
normanmaurer: nit: /areReadOkayNow/areReadsOkayNow/

Tests/NIOTests/StreamChannelsTest.swift
```diff
@@ -246,6 +247,85 @@ class StreamChannelTest: XCTestCase {
...
+                    // (rdar://53656794) for UNIX Domain Sockets and the additional 1 byte send loop will also pretty
+                    // reliably trigger it for TCP sockets.
+                    let myBuffer = allBuffer.readSlice(length: sends.load() == 0 ? receiveBufferSize : 1)!
+                    sender.writeAndFlush(myBuffer).map {
+                        _ = sends.add(1)
+                        sender.eventLoop.scheduleTask(in: .microseconds(1)) {
+                            send()
+                        }
+                    }.whenFailure { error in
+                        guard case .some(.ioOnClosedChannel) = error as? ChannelError else {
```
normanmaurer: nit: I think it is kind of odd to use a guard here if there is nothing defined outside of the scope of the guard itself. So maybe just use a `if let` ?

weissi: discussed offline, no really good way to spell with `if` because there's no `if not case`

Tests/NIOTests/StreamChannelsTest.swift
weissi: thanks, fixing

Sources/NIO/SelectableEventLoop.swift
weissi: thanks, done

PR #1306, opened 2019-12-16
accept: don't do unnecessary dodgy pointer things

weissi:
Motivation:

We never look at the socket address that accept(4) returns so we
shouldn't ask for it. Additionally, it wasn't always the right size (in
IPv6 case) and also the pointer operations were dodgy.

Thanks to @milseman for the pointer.

Modifications:

Remove the whole socket address pointer stuff for accept.

Result:

Cleaner code.

### DIFF
--- a/Sources/NIO/ServerSocket.swift
+++ b/Sources/NIO/ServerSocket.swift
@@ -66,24 +66,17 @@
     /// - throws: An `IOError` if the operation failed.
     func accept(setNonBlocking: Bool = false) throws -> Socket? {
         return try withUnsafeFileDescriptor { fd in
-            var acceptAddr = sockaddr_in()
-            var addrSize = socklen_t(MemoryLayout<sockaddr_in>.size)
-
-            let result = try withUnsafeMutablePointer(to: &acceptAddr) { (ptr) throws -> CInt? in
-                try ptr.withMemoryRebound(to: sockaddr.self, capacity: 1) { ptr in
-                    #if os(Linux)
-                    let flags: Int32
-                    if setNonBlocking {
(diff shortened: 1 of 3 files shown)


Sources/NIO/ServerSocket.swift
```diff
@@ -66,24 +66,18 @@
...
-                    return try Posix.accept(descriptor: fd, addr: ptr, len: &addrSize)
-                    #endif
-                }
+            var zero = socklen_t(0)
+            #if os(Linux)
+            let flags: Int32
+            if setNonBlocking {
+                flags = Linux.SOCK_NONBLOCK
+            } else {
+                flags = 0
```
Davidde94: Nit: For simple ifs like this I'm a fan of ternaries, but not sure what the NIO standard is

weissi: @Davidde94 that code is unchanged, just reindented.

Davidde94: I saw - just a note, feel free to merge :)

Sources/NIO/ServerSocket.swift
```diff
@@ -66,24 +66,18 @@
...
             }
+            let result = try Linux.accept4(descriptor: fd, addr: nil, len: &zero, flags: flags)
```
Lukasa: The Linux manual page says that if we pass NULL for the address we should pass NULL for the address length as well.

weissi: thanks @Lukasa , done

Lukasa:
Gotta fix that compile error.

PR #1308, opened 2019-12-17
close kqueue/epoll fds in Selector.close

### LINKED ISSUE #1302: close the kqueue/epoll fds in SelectableEventLoop before it gets deinitialised
we should close the kqueue/epoll fds in SelectableEventLoop's close.

weissi:
Motivation:

There was always this one big embarrassment in the NIO codebase: We
violated the iron rule of 'no fd or scarce resource management in deinit'
in just one place: The most long-lived thing there is in NIO: The
Selector for the SelectableEventLoopGroup. Sure enough, even that tiny
violation lead to issues.

Modifications:

Close the fds in `close` rather than `deinit`.

Result:

- I'm feeling better
- makes @ktoso happy
- fixes #1302

### DIFF
--- a/Sources/NIO/SelectableEventLoop.swift
+++ b/Sources/NIO/SelectableEventLoop.swift
@@ -209,7 +209,12 @@ internal final class SelectableEventLoop: EventLoop {
             // one wakeup.
         })
 
-        self._schedule0(task)
+        do {
+            try self._schedule0(task)
+        } catch {
+            scheduled._promise.fail(error)
+        }
+
         return scheduled
     }
 
@@ -222,14 +227,15 @@ internal final class SelectableEventLoop: EventLoop {
     // - see: `EventLoop.execute`
     @inlinable
     internal func execute(_ task: @escaping () -> Void) {
-        self._schedule0(ScheduledTask(task, { error in
+        // nothing we can do if we fail enqueuing here.
+       try? self._schedule0(ScheduledTask(task, { error in
             // do nothing
         }, .now()))
     }
 
     /// Add the `ScheduledTask` to be executed.
     @usableFromInline
-    internal func _schedule0(_ task: ScheduledTask) {
+    internal func _schedule0(_ task: ScheduledTask) throws {
         self._tasksLock.withLockVoid {
             self._scheduledTasks.push(task)
         }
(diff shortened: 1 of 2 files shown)


Sources/NIO/Selector.swift
```diff
@@ -614,48 +629,75 @@ final class Selector<R: Registration> {
...
+        assert(self.myThread == NIOThread.current)
         guard self.lifecycleState == .open else {
             throw IOError(errnoCode: EBADF, reason: "can't close selector as it's \(self.lifecycleState).")
         }
         self.lifecycleState = .closed
 
         self.registrations.removeAll()
 
-        /* note, we can't close `self.fd` (on macOS) or `self.eventfd` (on Linux) here as that's read unprotectedly and might lead to race conditions. Instead, we abuse ARC to close it for us. */
+        /* note, we can't close `self.fd` (on macOS) or `self.eventFD` (on Linux) here as that's read unprotectedly and might lead to race conditions. Instead, we abuse ARC to close it for us. */
```
ktoso: comment not valid anymore, we do close here 👀

Sources/NIO/Selector.swift
```diff
@@ -614,48 +629,75 @@ final class Selector<R: Registration> {
...
-        var event = kevent()
-        event.ident = 0
-        event.filter = Int16(EVFILT_USER)
-        event.fflags = UInt32(NOTE_TRIGGER | NOTE_FFNOP)
-        event.data = 0
-        event.udata = nil
-        event.flags = 0
-        try withUnsafeMutablePointer(to: &event) { ptr in
-            try kqueueApplyEventChangeSet(keventBuffer: UnsafeMutableBufferPointer(start: ptr, count: 1))
+        assert(NIOThread.current != self.myThread)
```
ktoso: some message for the assert or nah?

weissi: @ktoso don't think it adds much value, we know it's the wrong thread and we get which one from the backtrace.

Sources/NIO/Selector.swift
weissi: thank you, fixed

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -222,14 +227,15 @@ internal final class SelectableEventLoop: EventLoop {
     // - see: `EventLoop.execute`
     @inlinable
     internal func execute(_ task: @escaping () -> Void) {
-        self._schedule0(ScheduledTask(task, { error in
+        // nothing we can do if we fail enqueuing here.
+       try? self._schedule0(ScheduledTask(task, { error in
```
Lukasa: Can you add a TODO for this to throw in NIO 3?

Sources/NIO/Selector.swift
ktoso: Also true, your call :) I always make a point of doing including a message, a style thing more than anything I guess

Sources/NIO/SelectableEventLoop.swift
weissi: filed as https://github.com/apple/swift-nio/issues/1309

weissi:

PR #1317, opened 2019-12-19
add descriptions to EventLoop and NIOThread

weissi:
Motivation:

Certain issues are easier debugged if we can print the thread names.

Modifications:

- add the thread name to SelectableEventLoop's description
- make EventLoop and NIOThread CustomStringConvertible

Result:

easier debugging.

### DIFF
--- a/Sources/NIO/Thread.swift
+++ b/Sources/NIO/Thread.swift
@@ -36,6 +36,7 @@ private func sys_pthread_setname_np(_ p: pthread_t, _ pointer: UnsafePointer<Int
 ///
 /// All methods exposed are thread-safe.
 final class NIOThread {
+    private let desiredName: String?
 
     /// The pthread_t used by this instance.
     private let pthread: pthread_t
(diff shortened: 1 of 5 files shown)


Sources/NIO/Thread.swift
```diff
@@ -60,14 +62,19 @@ final class NIOThread {
     }
 
     /// Get current name of the `NIOThread` or `nil` if not set.
-    var name: String? {
+    var currentName: String? {
         get {
-            // 64 bytes should be good enough as on Linux the limit is usually 16 and it's very unlikely a user will ever set something longer anyway.
```
glbrntt: Can we keep this comment?

Sources/NIO/Thread.swift
```diff
@@ -126,7 +133,36 @@ final class NIOThread {
...
+
+        switch (desiredName, actualName) {
+        case (.some(let desiredName), .some(desiredName)):
+            // We know the current, actual name and the desired name and they match. This is hopefully the most common
+            // situation.
+            return "NIOThread(name = \(desiredName))"
+        case (.some(let desiredName), .some(let actualName)):
+            // We know both names but they're not equal. That's odd but not impossible, some misbehaved library might
+            // have changed the name.
+            return "NIOThread(desiredName = \(desiredName), actualThreadName = \(actualName))"
```
glbrntt:
```suggestion
            return "NIOThread(desiredName = \(desiredName), actualName = \(actualName))"
```

Sources/NIO/Thread.swift
weissi: done

Sources/NIO/Thread.swift
weissi: thanks

PR #1323, opened 2019-12-20
universal bootstrap

weissi:
Motivation:

Currently, it's very hard to use SwiftNIO on BSD Sockets as well as Network.framework with NIOTransportServices.

Modifications:

Offer a "universal bootstraping mechanism" for TCP clients called `NIOClientTCPBootstrap` which allows you to combine a regular bootstrap and a TLS implementation into a "universal bootstrap".

Result:

Make it easier to use NIO in more heterogenous environments.

This is the universal bootstrap work captured in the following four PRs:

- https://github.com/apple/swift-nio/pull/1323
- https://github.com/apple/swift-nio-transport-services/pull/69
- https://github.com/apple/swift-nio-ssl/pull/185
- https://github.com/apple/swift-nio-examples/pull/47


Sources/NIO/UniversalBootstrapSupport.swift
```diff
@@ -0,0 +1,198 @@
...
+}
+
+public protocol NIOBootstrapWithProtocolHandlers: NIOTCPClientBootstrapProtocol {
+    func protocolHandlers(_ handlers: [() -> ChannelHandler]) -> Self
+}
+
+extension ClientBootstrap: NIOBootstrapWithProtocolHandlers {}
+
+public protocol NIOTLSProvider {
+    associatedtype Bootstrap: NIOTCPClientBootstrapProtocol
```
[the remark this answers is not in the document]
weissi: good point, fixed!

Sources/NIO/Bootstrap.swift
```diff
@@ -439,7 +417,13 @@ public final class ClientBootstrap: NIOTCPClientBootstrap {
     /// - parameters:
     ///     - handler: A closure that initializes the provided `Channel`.
     public func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> Self {
-        self.channelInitializer = handler
+        self._channelInitializer = handler
+        return self
+    }
+
+    public func protocolHandlers(_ handlers: [() -> ChannelHandler]) -> Self {
```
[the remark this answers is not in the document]
weissi: thanks, fixed.

Sources/NIO/Bootstrap.swift
```diff
@@ -439,7 +417,19 @@ public final class ClientBootstrap: NIOTCPClientBootstrap {
...
+        self._channelInitializer = handler
+        return self
+    }
+
+    /// Sets the protocol handlers that will be added to the front of the `ChannelPipeline` right after the
+    /// `channelInitializer` has been called.
+    ///
+    /// Per bootstrap, you can only set the `protocolHandlers` once. Typically, `protocolHandlers` are used for the TLS
+    /// implementation. Most notably, `NIOClientTCPBootstrap`, NIO's "universal bootstrap" abstraction, uses
+    /// `protocolHandlers` to add the requires `ChannelHandler`s for many TLS implementations.
```
glbrntt:
```suggestion
    /// `protocolHandlers` to add the required `ChannelHandler`s for many TLS implementations.
```

Sources/NIO/UniversalBootstrapSupport.swift
```diff
@@ -0,0 +1,184 @@
...
+    /// - parameters:
+    ///     - handler: A closure that initializes the provided `Channel`.
+    func channelInitializer(_ handler: @escaping (Channel) -> EventLoopFuture<Void>) -> Self
+
+    /// Sets the protocol handlers that will be added to the front of the `ChannelPipeline` right after the
+    /// `channelInitializer` has been called.
+    ///
+    /// Per bootstrap, you can only set the `protocolHandlers` once. Typically, `protocolHandlers` are used for the TLS
+    /// implementation. Most notably, `NIOClientTCPBootstrap`, NIO's "universal bootstrap" abstraction, uses
+    /// `protocolHandlers` to add the requires `ChannelHandler`s for many TLS implementations.
```
glbrntt:
```suggestion
    /// `protocolHandlers` to add the required `ChannelHandler`s for many TLS implementations.
```

Sources/NIO/Bootstrap.swift
```diff
@@ -439,7 +417,19 @@ public final class ClientBootstrap: NIOTCPClientBootstrap {
...
+    public func protocolHandlers(_ handlers: @escaping () -> [ChannelHandler]) -> Self {
+        precondition(self.protocolHandlers == nil, "protocol handlers can only be set once")
```
glbrntt: Why do we enforce this for protocol handlers but not for the channel initializer?

glbrntt:
@weissi The CI is very unhappy with this change :)

Sources/NIO/Bootstrap.swift
weissi: it's a good q. My rationale was that because the new `NIOTLSProvider`s may use the `protocolHandlers` it may lead to something unexpected if the users would use the `protocolHandlers` for something and then later on call `enableTLS` (which for NIOSSL will set the `protocolHandlers` without the user seeing that)...

I _think_ that's more likely to happen than the _user_ calling `channelInitializer` twice themselves.

CC @Lukasa

weissi:
@glbrntt rebased, hopefully it likes it more now :)

Sources/NIO/Bootstrap.swift
Lukasa: Seems reasonable to me.

glbrntt: Makes sense; thanks for clarifying.

Sources/NIO/Bootstrap.swift
```diff
@@ -398,10 +364,21 @@ internal protocol NIOTCPClientBootstrap {
...
+public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
     private let group: EventLoopGroup
-    private var channelInitializer: Optional<ChannelInitializerCallback>
+    private var protocolHandlers: Optional<() -> [ChannelHandler]>
+    private var _channelInitializer: ChannelInitializerCallback
+    private var channelInitializer: ChannelInitializerCallback {
+        if let protocolHandlers = self.protocolHandlers {
+            return { channel in
+                self._channelInitializer(channel).flatMap {
+                    channel.pipeline.addHandlers(protocolHandlers(), position: .first)
```
Lukasa: What's the reasoning for doing initializer followed by protocol handlers? I worry about channel handlers issuing writes in their `handlerAdded`.

weissi: @Lukasa good Q but the reasoning is that we can't reverse it because if the user does `addHandler(..., position: .first)` in the `channelInitializer`, then that would break TLS. Protocol handlers need to go at the very front... Writes from `handlerAdded` are indeed a problem, writes from `channelActive` are not

Lukasa: I figured. Ok, good to know, my approval from earlier stands.

PR #1324, opened 2019-12-20
NonBlockingFileIO: support and use pread

weissi:
Motivation:

Using `pread` and calculating file offsets should usually be preferred to using the kernel's seek pointer because it's often faster and there is no interference from reads happening on other threads. For some reason, we didn't support this yet in NIO.

Modifications:

Use & support pread where possible.

Result:

Faster, better, and happier @vlm

### DIFF
--- a/Sources/NIO/NonBlockingFileIO.swift
+++ b/Sources/NIO/NonBlockingFileIO.swift
@@ -62,6 +62,9 @@ public struct NonBlockingFileIO {
     ///
     /// The allocation and reading of a subsequent chunk will only be attempted when `chunkHandler` succeeds.
     ///
+    /// This method will not use the file descriptor's seek pointer which means there is no danger of reading from the
+    /// same `FileRegion` in multiple threads.
+    ///
     /// - parameters:
     ///   - fileRegion: The file region to read.
     ///   - chunkSize: The size of the individual chunks to deliver.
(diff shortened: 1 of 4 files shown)


Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -189,10 +261,17 @@ public struct NonBlockingFileIO {
...
-                    let res = try fileHandle.withUnsafeFileDescriptor { descriptor in
-                        try Posix.read(descriptor: descriptor,
-                                       pointer: ptr.baseAddress!,
-                                       size: byteCount - bytesRead)
+                    let res = try fileHandle.withUnsafeFileDescriptor { descriptor -> IOResult<ssize_t> in
+                        if let offset = fromOffset {
+                            return try Posix.pread(descriptor: descriptor,
+                                                   pointer: ptr.baseAddress!,
+                                                   size: byteCount - bytesRead,
+                                                   offset: off_t(offset + bytesRead))
```
vlm: Perhaps, something like `offset.addingReportingOverflow(bytesRead)` would be more correct?

Sources/NIO/System.swift
```diff
@@ -375,12 +376,19 @@ internal enum Posix {
     }
 
     @inline(never)
-    public static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<Int> {
+    public static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<ssize_t> {
```
vlm: That's a source breaking change, I think. Why not keep it at `Int`, which is the same width and signedness with `ssize_t` on our platforms?

weissi: This is a phantom public, public func on internal type, ie not actually public. We do have swift-api-digester running which checks api breakages.

Sources/NIO/NonBlockingFileIO.swift
weissi: Hmm, Int as well as off_t are 64 bit on all platforms we really care about, right? So this should never overflow with reasonable `offset` and if it does I think crashing is fine. On 32 bit this is a bit more nuanced so we could do `off_t(offset) + off_t(bytesRead)`

weissi: changed to the latter

Sources/NIO/System.swift
weissi: @vlm btw, we also sorta support 32 bit platforms (like some androids and raspberry pies)

vlm: @weissi on 32-bits both `Int` and `ssize_t` are 32-bits.

weissi: True, but still ssize_t is more correct here because that’s the actual return type. And this function is really not exported because the type is internal. We try to match the types as closely as we can with the actual ones in posix/libc.

milseman:
> Using pread and calculating file offsets should usually be preferred

Are you planning on updating the docs for the existing non-offset APIs to note this? For NIO3 or later, are you planning or considering a renaming/restructuring/deprecation of the non-offset API?

Lukasa:
We should probably update the docs, yeah. I don't think we plan a deprecation at this stage, but we certainly could.

Sources/NIO/System.swift
```diff
@@ -58,7 +58,8 @@ private let sysConnect: @convention(c) (CInt, UnsafePointer<sockaddr>?, socklen_
 private let sysOpen: @convention(c) (UnsafePointer<CChar>, CInt) -> CInt = open
 private let sysOpenWithMode: @convention(c) (UnsafePointer<CChar>, CInt, mode_t) -> CInt = open
 private let sysWrite: @convention(c) (CInt, UnsafeRawPointer?, CLong) -> CLong = write
```
ianpartridge: Fix up the return type of this to `ssize_t` at the same time?

weissi:
@Lukasa please re-review, I address @milseman and @ianpartridge's comments.

Issue #1320, opened 2019-12-20
ByteBuffer should be hashable


weissi:
ByteBuffers should conform to `Hashable` which should be pretty straightforward as we can just call `withUnsafeReadableBytes` and hand those bytes to the `Hasher`.

fabianfett:
Hi @weissi, 

I just looked at this. I came up with your proposed really simple solution:

```swift
extension ByteBuffer: Hashable {
    /// The hash value for the readable bytes.
    public func hash(into hasher: inout Hasher) {
        withUnsafeReadableBytes { ptr in
            hasher.combine(bytes: ptr)
        }
    }
}
```

But adding a test that ensures that different buffers with the same content but different reserved capacities are hash equal showed me that this maybe is not as simple as thought.

```swift
    func testHasherUsesReadBuffersOnly() {
        // Only cares about the read buffer
        self.buf.writeInteger(Int8.max)
        self.buf.writeString("oh hi")

        var hasher = Hasher()
        hasher.combine(self.buf)
        let hash = hasher.finalize()
        
        var otherBuffer = allocator.buffer(capacity: 32)
        otherBuffer.writeString("oh hi")

        var otherHasher = Hasher()
        otherHasher.combine(otherBuffer)
        let otherHash = otherHasher.finalize()
        
        XCTAssertEqual(hash, otherHash)
    }
```

Do you maybe have a hint on how to proceed from here?

Lukasa:
Have you cleared `self.buf`? That is, are you confident there are no readable bytes in there from the previous test?

fabianfett:
@Lukasa I made it explicit and it still fails. If you want me to I can upload my code and create a PR so that you can have a look maybe?

My first guess with the reserved capacities is obviously not the case. Sorry dumb me. 

If I set a breakpoint for the line:
```
hasher.combine(bytes: ptr)
```

And I `po` the pointer for each run, I get the following output:
```
(lldb) po ptr
▿ UnsafeRawBufferPointer(start: 0x0000000105004000, count: 5)
  ▿ _position : Optional<UnsafeRawPointer>
    ▿ some : 0x0000000105004000
      - pointerValue : 4378869760
  ▿ _end : Optional<UnsafeRawPointer>
    ▿ some : 0x0000000105004005
      - pointerValue : 4378869765

(lldb) po ptr
▿ UnsafeRawBufferPointer(start: 0x0000000100f00f80, count: 5)
[...]

PR #1326, opened 2019-12-21
ByteBuffer implements Hashable

fabianfett:
Motivation:

`ByteBuffer` should implement the `Hashable` protocol as stated by: https://github.com/apple/swift-nio/issues/1320

Modifications:

Implemented the `Hashable` protocol for `ByteBuffer` and added a test `testHasherUsesReadBuffersOnly`.

Result:

`ByteBuffer` implements the `Hashable` protocol. The test `testHasherUsesReadBuffersOnly` is failing for the moment though.

### DIFF
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -71,6 +71,31 @@ class ByteBufferTest: XCTestCase {
         otherBuffer.writeString("oh hi")
         XCTAssertEqual(otherBuffer, buf)
     }
+    
+    func testHasherUsesReadBuffersOnly() {
+        // Only cares about the read buffer
+        self.buf.clear()
+        self.buf.writeString("oh hi")
+
+        var hasher = Hasher()
+        // We need to force unwrap the implicitly unwrapped optional here in order to
+        // mark it as unwrapped for the compiler *before* the function call. Otherwise
+        // the implementation of the optional's conditional conformance is triggered,
+        // that will change the hash. For more information please see:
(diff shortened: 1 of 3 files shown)


Lukasa:

weissi:
eeehm, even if I change the code to

```
extension ByteBuffer: Hashable {
    /// The hash value for the readable bytes.
    public func hash(into hasher: inout Hasher) {
        //self.withUnsafeReadableBytes { ptr in
            hasher.combine(1)
            //hasher.combine(bytes: ptr)
        //}
    }
}
```

which is very clearly wrong but should always lead to the same hash, it does _not_

weissi:
@fabianfett / @Lukasa  OOOMMMGG, I found the issue! If you replace all appearances of `self.buf` with `self.buf!`, then it actually works :)

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -71,6 +71,25 @@ class ByteBufferTest: XCTestCase {
...
         XCTAssertEqual(otherBuffer, buf)
     }
+    
+    func testHasherUsesReadBuffersOnly() {
+        // Only cares about the read buffer
+        self.buf.clear()
+        self.buf.writeString("oh hi")
+
+        var hasher = Hasher()
+        hasher.combine(self.buf)
```
weissi:
```suggestion
        hasher.combine(self.buf!)
```

weissi:
filed it here: https://bugs.swift.org/browse/SR-11975 . Don't think this will go anywhere because it's not really a bug but let's see ;)

weissi:

Issue #1322, opened 2019-12-20
NIOWebsocket Base64 encoding - unexpected memory overhead.


fabianfett:
### Actual behavior

After the issue https://github.com/apple/swift-nio/issues/1265, I started to dig into the implementation of base64. The results of that work can be found at [fabianfett/swift-base64](https://github.com/fabianfett/swift-base64). While implementing decoding, I tested the performance of my implementation against Foundation. Since I wanted to base my work on the Base64 encoding provided by `NIOWebsocket` I also tested its performance. Surprisingly it's rather slow.

Encoding 100k times the bytes from 0 to 255 produced these results:

**On my MacBook Pro 2016 (2,7 GHz Intel Core i7)** 
```
Number of invocations: 100000
------------------------------------------
Encoding
Foundation   | took: 0.24192798137664795s
NIOWebsocket | took: 8.385756015777588s
------------------------------------------
```

**On Linux Foundation's performance isn't great either.** Therefore the performance hit compared to Foundation is not as bad.
```
------------------------------------------
Encoding
Foundation   | took: 3.710782051086426s
NIOWebsocket | took: 5.109261989593506s
Base64       | took: 0.10653805732727051s
------------------------------------------
```

### If possible, minimal yet complete reproducer code (or URL to code)

The code to reproduce can be found the branch `compare-with-niowebsocket-base64` at [fabianfett/swift-base64](https://github.com/fabianfett/swift-base64)
https://github.com/fabianfett/swift-base64/tree/compare-with-niowebsocket-base64

Performance results on GitHub actions can be seen in the `Performance Tests` (Not great for detailed measuring, but we are of the target by quite a margin in this case):
https://github.com/fabianfett/swift-base64/pull/2/checks

Looking into the issue with Instruments, I'm quite sure the performance hit is based on the use of the `Unicode.Scalar` that seem to come with "some" memory overhead.

<img width="930" alt="Screen Shot 2019-12-20 at 18 52 51" src="https://user-images.githubusercontent.com/6780861/71279796-f2c0bb00-2359-11ea-900d-c8b863220fa8.png">

### SwiftNIO version/commit hash

commit hash: `0ac7e2266bccd08ab174f6e2ef55522797b68343` code unchanged since Feb 14 2019.

### Swift & OS version (output of `swift --version && uname -a`)

```
Target: x86_64-apple-darwin18.7.0
[...]

PR #1328, opened 2019-12-23
Faster Base64 Encoding for NIOWebsocket connections

fabianfett:
Motivation:

In performance tests was shown that the current base 64 encoding function doesn’t perform very well.

https://github.com/apple/swift-nio/issues/1322

Modifications:

This commit replaces the current Base64 implementation with a more performant one (by using UInt8 bytes instead of Unicode.Scalars).

Result:

The API will stay the same. Base64 encoding is just faster.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -77,3 +77,12 @@ This product contains "ifaddrs-android.c" and "ifaddrs-android.h" from dxr.mozil
     * https://webrtc.googlesource.com/src/+/master/LICENSE
   * HOMEPAGE:
     * https://webrtc.googlesource.com/
+
+---
+
+This product contains a derivation of Fabian Fett's 'Base64.swift'.
+
+  * LICENSE (Apache License 2.0):
+    * https://github.com/fabianfett/swift-base64-kit/blob/master/LICENSE
+  * HOMEPAGE:
+    * https://github.com/fabianfett/swift-base64-kit
--- a/Sources/NIOWebSocket/Base64.swift
+++ b/Sources/NIOWebSocket/Base64.swift
@@ -12,75 +12,112 @@
 //
 //===----------------------------------------------------------------------===//
 
-// The base64 unicode table.
-private let base64Table: [Unicode.Scalar] = [
-    "A", "B", "C", "D", "E", "F", "G", "H",
-    "I", "J", "K", "L", "M", "N", "O", "P",
-    "Q", "R", "S", "T", "U", "V", "W", "X",
-    "Y", "Z", "a", "b", "c", "d", "e", "f",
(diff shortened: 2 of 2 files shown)


Sources/NIOWebSocket/Base64.swift
```diff
@@ -12,75 +12,122 @@
...
+  @inlinable
+  static func encode<Buffer: Collection>(bytes: Buffer)
+    -> String where Buffer.Element == UInt8
+  {
+    // In Base64, 3 bytes become 4 output characters, and we pad to the
+    // nearest multiple of four.
+    let newCapacity = ((bytes.count + 2) / 3) * 4
+    let alphabet = Base64.encodeBase64
+    
+    var outputBytes = [UInt8]()
```
weissi: we should put a note here that (once available) we should take advantage of [SE-263](https://github.com/apple/swift-evolution/blob/master/proposals/0263-string-uninitialized-initializer.md) to save the extra copy from `[UInt8]` to `String`

Sources/NIOWebSocket/Base64.swift
```diff
@@ -12,75 +12,122 @@
...
-            outputString.unicodeScalars.append(String.encode(secondByte: secondByte, thirdByte: thirdByte))
-            outputString.unicodeScalars.append(String.encode(thirdByte: thirdByte))
-        }
+  /// Base64 encode a collection of UInt8 to a string, without the use of Foundation.
+  ///
+  /// This function performs the world's most naive Base64 encoding: no attempts to use a larger
+  /// lookup table or anything intelligent like that, just shifts and masks. This works fine, for
+  /// now: the purpose of this encoding is to avoid round-tripping through Data, and the perf gain
+  /// from avoiding that is more than enough to outweigh the silliness of this code.
+  @inlinable
```
weissi: all the `@inlinable` in this file don't do anything because this isn't `public` and always consumed within the `NIOWebSocket` module so I think we should take them out.

Lukasa:

NOTICE.txt
```diff
@@ -77,3 +77,12 @@ This product contains "ifaddrs-android.c" and "ifaddrs-android.h" from dxr.mozil
     * https://webrtc.googlesource.com/src/+/master/LICENSE
   * HOMEPAGE:
     * https://webrtc.googlesource.com/
+
+---
+
+This product contains a derivation of the Fabian Fett's 'Base64.swift'.
```
Lukasa: Nit: not “the” Fabian Fett, just Fabian Fett. :wink:

fabianfett: @Lukasa damn it copy and paste and not reading once more. sorry. fixed now.

Lukasa:

PR #1329, opened 2019-12-30
Initializing HTTPMethod with String picks explicit value over catch all RAW(value)

fabianfett:
Motivation:

When initializing an `HTTPMethod` with a `String` we currently switch over all expected values to pick an explicit value, only to overwrite the found explicit value with a catchall `RAW(value: String)`. This doesn’t make a ton of sense. This went unnoticed because we use the conditional conformance on `RawRepresentable` to be `Equatable`. But `switch`es  don’t use the `Equatable` protocol to match enums. The issue is visible there.

Modifications:

Added a testcase with a switch that fails, if the `.RAW(value: String)` value is used for `”GET”`. Removed the final overwrite in HTTPMethod(rawValue: String). Therefore the new test passes.

Result:

We can initialize the HTTPMethod with a String and get an explicit value.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -1416,6 +1416,5 @@ extension HTTPMethod: RawRepresentable {
             default:
                 self = .RAW(value: rawValue)
         }
-        self = .RAW(value: rawValue)
     }
 }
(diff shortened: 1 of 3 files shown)


Lukasa:

fabianfett:
@Lukasa Damn it! Sorry forgot to add to Linux test suite.

Lukasa:

fabianfett:
Screen Shot 2020-01-02 at 17 03 54

Lukasa:

PR #1331, opened 2020-01-05
Update EventLoopFuture.swift

MihaelIsaev:
Add an ability to return throwing `EventLoopFuture` with `flatMap`

### Motivation:

With current methods I'm unable to return throwing `EventLoopFuture` with `flatMapThrowing`

### Modifications:

Implement `flatMapFutureThrowing ` method

### Result:

Now I'm able to return throwing `EventLoopFuture` by using `flatMapFutureThrowing`

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -509,6 +509,35 @@ extension EventLoopFuture {
             }
         }
     }
+    
+    /// When the current `EventLoopFuture<Value>` is fulfilled, run the provided callback, which
+    /// performs a synchronous computation and returns a new value of type `EventLoopFuture<NewValue>`. 
+    /// The provided callback may optionally `throw`.
+    ///
+    /// Operations performed in `flatMapThrowing` should not block, or they will block the entire
+    /// event loop. `flatMapThrowing` is intended for use when you have a data-driven function that
+    /// performs a simple data transformation that can potentially error.
+    ///
+    /// If your callback function throws, the returned `EventLoopFuture` will error.
+    ///
+    /// - parameters:
+    ///     - callback: Function that will receive the value of this `EventLoopFuture` and return
+    ///         a new value lifted into a new `EventLoopFuture`.
+    /// - returns: A future that will receive the eventual value.
+    @inlinable
+    public func flatMapFutureThrowing<NewValue>(file: StaticString = #file,
+                                line: UInt = #line,
+                                _ callback: @escaping (Value) throws -> EventLoopFuture<NewValue>) -> EventLoopFuture<NewValue> {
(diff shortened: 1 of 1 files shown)


MihaelIsaev:
@Lukasa 
> This is similar to having a function that `throws` and returns a `Result`: there are two error propagation paths in this function and it’d be better to have one.
If I understood correctly e.g. `JSONEncoder().encode` is exactly this kind of function with two propagation paths, right? If so then what's wrong with it? It can return normal value or throw.
I'm really don't understand why code smells if I have two throwing futures and call them one after another, e.g.:
```swift
// sends network request and decodes it through `try JSONDecoder().decode(...)`
func getOrGenerateAccessToken() throws -> EventLoopFuture<String> { … }

func sendMessage() -> EventLoopFuture<String> {
    return try getOrGenerateAccessToken().flatMapFutureThrowing { accessToken in
        var headers = HTTPHeaders()
        headers.add(name: "Authorization", value: "Bearer \(accessToken)")
        // throwing request initialization
        let request: HTTPClient.Request = try .init(url: “…”, method: .POST, headers: headers, body: .data(payloadData)) else { throw Abort(.internalServerError, reason: "Unable to create request for FCM") }
            return self.httpClient.execute(request: request).flatMapThrowing { res in
                struct Result: Codable {
                    let name: String
                }
                // throwing decoder
                guard let body = res.body, let result = try JSONDecoder().decode(Result.self, from: body) else {
                    throw Abort(.notFound, reason: "Data not found")
                }
                return result.name
            }
    }
}
```
So if throwing future inside throwing future smells then I should avoid any throwings, but I can't do that cause I want to know if error occurred.

in Vapor3 we had convenience wrappers for `flatMap` and `map` which allowed us to have throwing futures inside throwing futures and it was ok, but now with Vapor4 we have clean NIO without these wrappers and I stated experience these problems.

Honestly don't know what to do...

Lukasa:
There’s a very straightforward refactor here:

```swift
func sendMessage() -> EventLoopFuture<String> {
    return try getOrGenerateAccessToken().flatMapThrowing { accessToken in
        var headers = HTTPHeaders()
        headers.add(name: "Authorization", value: "Bearer \(accessToken)")
        // throwing request initialization
        guard let request = HTTPClient.Request(url: “…”, method: .POST, headers: headers, body: .data(payloadData)) else { throw Abort(.internalServerError, reason: "Unable to create request for FCM") }
        return request
    }.flatMap { request in
        return self.httpClient.execute(request: request)
    .flatMapThrowing { res in
        struct Result: Codable {
            let name: String
        }
        // throwing decoder
        guard let body = res.body, let result = try JSONDecoder().decode(Result.self, from: body) else {
            throw Abort(.notFound, reason: "Data not found")
        }
        return result.name
    }
}
```

PR #1336, opened 2020-01-13
Correct NIOHTTPWebClientSocketUpgrader to NIOHTTPWebSocketClientUpgrader

tigerpixel:
Correct 'NIOHTTPWebClientSocketUpgrader' class name to 'NIOHTTPWebSocketClientUpgrader' class name.

### Motivation:

The class name was a typo.
It is a client upgrader that upgrades WebSocket connections, so the updated name suits better.

### Modifications:

Change class name of ‘NIOHTTPWebClientSocketUpgrader’ to ‘NIOHTTPWebSocketClientUpgrader’.
Update the WebSocket client example that uses this class.
Add a typealias to allow the old class name to be used.

### Result:

Improved naming.

### DIFF
--- a/Sources/_NIO1APIShims/NIO1APIShims.swift
+++ b/Sources/_NIO1APIShims/NIO1APIShims.swift
@@ -531,3 +531,6 @@ extension WebSocketFrameDecoder {
         self.init(maxFrameSize: maxFrameSize)
     }
 }
+
+@available(*, deprecated, renamed: "NIOWebSocketClientUpgrader")
+typealias NIOWebClientSocketUpgrader = NIOWebSocketClientUpgrader
--- a/Sources/_NIO1APIShims/NIO1APIShims.swift
+++ b/Sources/_NIO1APIShims/NIO1APIShims.swift
@@ -531,6 +531,3 @@ extension WebSocketFrameDecoder {
         self.init(maxFrameSize: maxFrameSize)
     }
 }
-
-@available(*, deprecated, renamed: "NIOWebSocketClientUpgrader")
-typealias NIOWebClientSocketUpgrader = NIOWebSocketClientUpgrader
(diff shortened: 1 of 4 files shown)


Sources/_NIO1APIShims/NIO1APIShims.swift
```diff
@@ -531,3 +531,6 @@ extension WebSocketFrameDecoder {
         self.init(maxFrameSize: maxFrameSize)
     }
 }
+
+@available(*, deprecated, renamed: "NIOWebSocketClientUpgrader")
+typealias NIOWebClientSocketUpgrader = NIOWebSocketClientUpgrader
```
Lukasa: This is not a NIO 1 shim: we have to carry this name with us for the rest of the project. It must be imported and available alongside any import of NIOWebSocket. Can you move this into `NIOWebSocketClientUpgrader.swift`?

tigerpixel: Done.

Lukasa:

Lukasa:
@weissi is this a known failure mode of the API compatibility tool?

weissi:
@Lukasa hmm, I was pretty sure it handled that one correctly. Maybe we’re using an old version? I think we should repro with 5.2 and if still a problem file a new bug, I filed a few issues already but they’re different

Lukasa:
Yup, I can reproduce this with the Swift 5.2 snapshots, so I think this bug exists upstream. I'll file a bugs.swift.org and see how we go.

weissi:
Oh wow! Thanks for properly checking @Lukasa

tigerpixel:
Nice tooling, done now.

Lukasa:

Lukasa:

Lukasa:

Lukasa:

Issue #1243, opened 2019-11-13
Example in docs for ByteToMessageDecoderVerifier.verifyDecoder is incorrect


Csloan7597:
The example in [the docs for this function](https://apple.github.io/swift-nio/docs/current/NIOTestUtils/Enums/ByteToMessageDecoderVerifier.html#/s:12NIOTestUtils28ByteToMessageDecoderVerifierO06verifyF016inputOutputPairs14decoderFactoryySay3NIO0C6BufferV_Say10InboundOutQzGtG_xyctKAG0cdeF0RzSQAKRQlFZ) supplies an argument which does not match the function's signature. The example won't compile. 

> For ExampleDecoder that produces `ExampleDecoderOutput`s you would use this method the following way:

``` swift 
let expectedInOuts = [(exampleInput1, ExampleDecoderOutput("1")),
                      (exampleInput2, ExampleDecoderOutput("2"))
                     ]
XCTAssertNoThrow(
    try ByteToMessageDecoderVerifier.verifyDecoder(inputOutputPairs: expectedInOuts, decoderFactory: { ExampleDecoder() })
)
```

`expectedInOuts` is `[(ByteBuffer, ExampleDecoderOutput)]` when it should in fact be `[(ByteBuffer, [ExampleDecoderOutput])]`, as per the function signature:

``` swift
func verifyDecoder<Decoder: ByteToMessageDecoder>(inputOutputPairs: [(ByteBuffer, [Decoder.InboundOut])],
                                                                decoderFactory: @escaping () -> Decoder) throws where Decoder.InboundOut: Equatable
```

weissi:
thank you for the report @Csloan7597

PeterAdams-A:
I believe this was fixed in #1338

PR #1338, opened 2020-01-14
Change the byte to message decoder verifier example to be correct. Bug #1243

tigerpixel:
Change the byte to message decoder verifier example to be correct.

### Motivation:

The byte to message decoder example on the ‘verifyDecoder’ method was not compiling.

### Modifications:

Added the example output objects to an array so that they match the method signature.

### Result:

The example is now compiling and can serve as the basis for creating an implementation from the example.


Lukasa:

Issue #1337, opened 2020-01-14
100% CPU usage [after sending message] on Swift ARM


Bouke:
### Expected behavior
When there's not much going on, little CPU usage. Increase in CPU while there's activity.

### Actual behavior
Continuously high CPU usage, even when there's no network traffic.

### Steps to reproduce

I'm not sure what exactly triggers the problem in my application, but I've observed the same problem with the chat client. So including those steps here:

1. Install SwiftARM on a Raspberry Pi (https://swift-arm.com/install-swift/)
2. Run both chat server and client (`swift run NIOChat(Server|Client)`)
3. Notice 100% cpu usage on NIOChatClient

### SwiftNIO version/commit hash

Tested against version 2.12.0 and master (f44fe0b2042924e3f0392febd8716bbcfc8f36a6).

### Swift & OS version (output of `swift --version && uname -a`)
```
$ swift --version
Swift version 5.1.3 (swift-5.1.3-RELEASE)
Target: armv6-unknown-linux-gnueabihf

$ uname -a
Linux pi3.local 4.19.66-v7+ #1253 SMP Thu Aug 15 11:49:46 BST 2019 armv7l GNU/Linux
```

# Flame graph
Some symbols are missing due to `perf` version mismatch (I think). Both the chat client and my application ("my-home") are included. Attached are the zoomable flame graphs: [flame-graphs.zip](https://github.com/apple/swift-nio/files/4060957/flame-graphs.zip).

<img width="1193" alt="image" src="https://user-images.githubusercontent.com/235882/72373681-17197880-3709-11ea-9dde-8759fa9384f9.png">

Lukasa:
The common cause for this is that we’re spinning in `epoll_wait`, which implies there’s some I/O event we haven’t processed. This is _usually_ `EPOLL_RDHUP`. I will do my best to see if I can reproduce this tomorrow, though the fact that this affects a raspberry pi makes life moderately awkward: I’ll see if we can reproduce it in another system.

weissi:
@Bouke Thanks for the detailed report! As @Lukasa points out, this points to us not properly processing some event so `epoll_wait` constantly returns again, telling us about the same event again.

If you can, would you mind to attach a few lines of the output of

```
strace -ff -e trace=epoll_ctl,epoll_wait -p PID_OF_PROCESS
```

whilst in the 100% CPU spin loop? Also, is this 32 or 64 bit?

Lukasa:
[...]

PR #1340, opened 2020-01-14
Read exactly UInt64's size from timerfd

Bouke:
On 32-bit systems we would try to read 4 byte, which is incorrect. From `man timerfd_read`:
> (...) then the buffer given to read(2) returns an unsigned 8-byte integer (uint64_t) containing the number of expirations that  have  occurred (...)
> A  read(2) will fail with the error EINVAL if the size of the supplied buffer is less than 8 bytes.

See #1337.

### DIFF
--- a/Sources/NIO/Selector.swift
+++ b/Sources/NIO/Selector.swift
@@ -547,9 +547,9 @@ final class Selector<R: Registration> {
                 _ = try EventFd.eventfd_read(fd: self.eventFD, value: &val)
             case self.timerFD:
                 // Consume event
-                var val: UInt = 0
+                var val: UInt64 = 0
                 // We are not interested in the result
-                _ = Glibc.read(self.timerFD, &val, MemoryLayout<UInt>.size)
+                _ = Glibc.read(self.timerFD, &val, MemoryLayout.size(ofValue: val))
 
                 // Processed the earliest set timer so reset it.
                 self.earliestTimer = .distantFuture


Sources/NIO/Selector.swift
```diff
@@ -549,7 +549,7 @@ final class Selector<R: Registration> {
                 // Consume event
                 var val: UInt = 0
```
weissi: sorry, my bad, the `var val` needs to become `UInt64` too.

Sources/NIO/Selector.swift
```diff
@@ -549,7 +549,7 @@ final class Selector<R: Registration> {
...
                 // We are not interested in the result
-                _ = Glibc.read(self.timerFD, &val, MemoryLayout<UInt>.size)
+                _ = Glibc.read(self.timerFD, &val, 8)
```
weissi: instead of a hard-coded `8` we should use `MemoryLayout.size(ofValue: val)` here.

weissi:
@Lukasa usually, we'd ask for a test here but this ought to be tested by lots of tests already and writing a targeted one isn't super easy. WDYT?

Sources/NIO/Selector.swift
Bouke: Ah yes, we're writing to `&val` indeed.

Lukasa:
I’m happy to skip having a test for this case, we set timers all over the shop.

weissi:

Lukasa:

PR #1343, opened 2020-01-16
docker: support Swift 5.2 dev snapshots

weissi:
Motivation:

We need to test on 5.2.

Modifications:

Add 5.2 dev snapshots to the docker setup.

Result:

More testing.

### DIFF
--- a/docker/docker-compose.1804.52.yaml
+++ b/docker/docker-compose.1804.52.yaml
@@ -1,0 +1,43 @@
+version: "3"
+
+services:
+
+  runtime-setup:
+    image: swift-nio:18.04-5.2
+    build:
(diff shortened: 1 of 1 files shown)


tomerd:
@weissi unfortunately this will not work as intended: the underlying Dockerfile no longer supports `swift_flavour` and other more fine grained configuration required for snapshot since our transition to the official swift docker images. in order to support this, we will need a separate Dockerfile like the one we had in the past

weissi:
@tomerd oh weird, I did run the tests with this file locally and they worked. Did they silently run a different swift version?

If that’s the case I think we should roll back our Dockerfile to support the flavours again :(

Issue #1332, opened 2020-01-08
ByteBuffer's `setString`/`writeString` etc should support `StringProtocol`


helje5:
The ByteBuffer String API is tied to the concrete `String` type. Would be nice to allow arbitrary `StringProtocol` values, most importantly substrings.

### SwiftNIO version/commit hash

2.12.0

### Swift & OS version (output of `swift --version && uname -a`)

5.1 any

Lukasa:
From my racy issue in #1333: We should add new versions of these functions that take a generic StringProtocol. These must be inlinable, of course. Ideally we'd verify that calling the generic function with a String performs just as well as calling the String-specific one.

ianpartridge:
I'm curious why the existing functions couldn't be made generic, rather than adding additional generic ones? Couldn't we use `@_specialize` to hint that `String` and `Substring` should be specialized?

helje5:
Would it change the ABI?

jckarter:
IMO `StringProtocol` is pretty silly. You could have the implementation take a `Substring` with an inlinable overload that takes a String and passes down `string[...]` automatically, maybe.

helje5:
That doesn't really answer the question ;-)

But regardless, for APIs you essentially want to enable both always, and having to write everything twice is pretty silly too.
Personally I'd prefer to have just String, but that's besides the point here.

jckarter:
Well, yes, changing a nongeneric function to a generic one is going to change the ABI, and `@_specialize` won't provide entry points ABI-compatible with the original concrete entry points.

I was just also saying that in general, making something generic over a protocol to cover two types, one of which has a nearly free conversion to the other, is going to add a lot of performance overhead. The whole String/Substring/StringProtocol thing is…not the way I'd have done things, but given the world we built, if the goal is to make something accept both `String` and `Substring`, it's much more practical to just take `Substring` IMO.

helje5:
> making something generic over a protocol to cover two types, one of which has a nearly free conversion to the other, is going to add a lot of performance overhead
Even if it is `@inlinable`?

jckarter:
If it's inlinable, the generic version still has to exist and take up space in case it doesn't get inlined.

Lukasa:
So while @jckarter's idea is good, we certainly can't take that approach for any current Swift version, because `Substring` has no fast access to its backing storage: see [SR-11999](https://bugs.swift.org/browse/SR-11999).

[...]

PR #1346, opened 2020-01-21
Add setSubstring/writeSubstring to ByteBuffer

cpriebe:
### Motivation:

The `ByteBuffer` API currently has `setString`/`writeString` support for the `String`
type but no support for the `Substring` type (see #1332).

### Modifications:

Add `setSubstring`/`writeSubstring` functions to `ByteBuffer`. We do not change
the exisiting set-/writeString API to take a generic `StringProtocol` argument
as it would change the ABI and adds performance overhead. We also do not let
`setString` call into `setSubstring` for now, as Substring currently (as of
Swift 5.1.3) has no fast access to the backing storage and this therefore would
introduce a performance regression.

### Result:

ByteBuffer API for writing Substrings.

### DIFF
--- a/Sources/NIO/ByteBuffer-aux.swift
+++ b/Sources/NIO/ByteBuffer-aux.swift
@@ -112,6 +112,9 @@ extension ByteBuffer {
     @discardableResult
     @inlinable
     public mutating func setString(_ string: String, at index: Int) -> Int {
+        // Do not implement setString via setSubstring. As of Swift version 5.1.3,
+        // Substring.UTF8View does not implement withContiguousStorageIfAvailable
+        // and therefore has no fast access to the backing storage.
         if let written = string.utf8.withContiguousStorageIfAvailable({ utf8Bytes in
             self.setBytes(utf8Bytes, at: index)
         }) {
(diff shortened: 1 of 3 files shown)


Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -152,6 +155,51 @@ extension ByteBuffer {
...
+        let written = self.setSubstring(substring, at: self.writerIndex)
+        self._moveWriterIndex(forwardBy: written)
+        return written
+    }
+    
+    @inline(never)
+    @usableFromInline
+    mutating func _setSubstringSlowpath(_ substring: Substring, at index: Int) -> Int {
+        // slow path, let's try to force the substring to be native
+        if let written = (substring + "").utf8.withContiguousStorageIfAvailable({ utf8Bytes in
```
Lukasa: @weissi You're the relevant expert here: is this a good idea on a `Substring`? It'll force an allocation and copy, so maybe we want to avoid this?

Lukasa:

Lukasa:
Hey @cpriebe, do you mind running `scripts/generate_linux_tests.rb` and adding the result to this commit? That's why the CI is suffering here.

cpriebe:
@Lukasa Done.

Sources/NIO/ByteBuffer-aux.swift
weissi: unfortunately, this won't work at all on `Substring` (which is a bug). See this:

```swift
import Foundation
let nsString = NSString(data: Data([0x42, 0, 0x42, 0]), encoding: String.Encoding.utf16.rawValue)!

print("hello"                    .utf8.withContiguousStorageIfAvailable({ _ in "string direct ok" }) ?? "string direct NOT okay")
print((nsString as String)       .utf8.withContiguousStorageIfAvailable({ _ in "nsstring direct ok" }) ?? "nsstring direct NOT okay")
print(((nsString as String) + "").utf8.withContiguousStorageIfAvailable({ _ in "nsstring + '' direct ok" }) ?? "nsstring + '' direct NOT okay")
print("hello"[...]               .utf8.withContiguousStorageIfAvailable({ _ in "substring direct ok" }) ?? "substring direct NOT okay")
print(("hello"[...] + "")        .utf8.withContiguousStorageIfAvailable({ _ in "substring + '' direct ok" }) ?? "substring + '' direct NOT okay")
print(String("hello"[...])       .utf8.withContiguousStorageIfAvailable({ _ in "substring through string direct ok" }) ?? "substring through string direct NOT okay")
```

outputs

```
string direct ok
nsstring direct NOT okay
nsstring + '' direct ok
substring direct NOT okay
substring + '' direct NOT okay
substring through string direct ok
```

so I think the only way we can make sub strings behave is to always do `String(subString)` and then go through `writeString` for now. We should also file a bugs.swift.org about this.

Lukasa: There’s already a bug: this is covered under the fact that substrings don’t have a `withContiguousStorageIfAvailable` implementation. But they _will_, and it’d be nice to not have to rewrite this when they do. 

The question is, is it acceptable to allocate a new `String` to get this fast path?

Lukasa: I guess it’s interesting that `Substring + Substring` returns `Substring`.

weissi: > The question is, is it acceptable to allocate a new `String` to get this fast path?

good question. Really, it depends on the length of the string. But I'd say we _should_ have a `writeSubstring` and for now we implement it as `String(sub)`. We can always improve later without breaking the API, right?

Lukasa: That would work for me

cpriebe: Sounds good. I'll make that change then.

cpriebe:
@weissi @Lukasa I have removed `_setSubstringSlowpath` for now and implement `setSubstring` by calling `setString` as discussed above.

Sources/NIO/ByteBuffer-aux.swift
weissi: thank you!

PR #1348, opened 2020-01-21
improve flushNow re-entrancy protection

weissi:
## Motivation

The fix in #1344 wasn't complete, there is another, quite complicated scenario where we would still forget to send flushed bytes until we see another `flush` call. The scenario is

         1: writable()
         2: --> flushNow (result: .writtenCompletely)
         3:     --> writabilityChanged callout
         4:         --> flushNow because user calls writeAndFlush (result: .couldNotWriteEverything)
         5:         --> registerForWritable (because line 4 could not write everything and flushNow returned .register)
         6: --> unregisterForWritable (because line 2 wrote everything and flushNow returned .unregister)

Line 6 undoes the registeration in line 5.

## Modification

This fix makes sure that flushNow never re-enters and therefore the problem described above cannot happen anymore.

## Result

Hopefully actually fixes rdar://58571521

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -415,6 +415,11 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
         fatalError("this must be overridden by sub class")
     }
 
+    /// Returns if there are any flushed, pending writes to be sent over the network.
+    func hasFlushedPendingWrites() -> Bool {
+        fatalError("this must be overridden by sub class")
+    }
+
     /// Buffer a write in preparation for a flush.
     func bufferPendingWrite(data: NIOAny, promise: EventLoopPromise<Void>?) {
         fatalError("this must be overridden by sub class")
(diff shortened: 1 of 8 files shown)


Sources/NIO/BaseSocketChannel.swift
```diff
@@ -469,32 +475,82 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
...
+                assert(!self.isWritePending())
+                ()
+            }
+        }
+    }
+
     /// Flush data to the underlying socket and return if this socket needs to be registered for write notifications.
     ///
-    /// - returns: If this socket should be registered for write notifications. Ie. `IONotificationState.register` if _not_ all data could be written, so notifications are necessary; and `IONotificationState.unregister` if everything was written and we don't need to be notified about writability at the moment.
+    /// - warning: Only call this method i
```
normanmaurer: there is something missing... I guess "only call this method in the eventloop" ?

weissi: whoops, thank you!

weissi: fixed.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -469,32 +475,83 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
...
+        assert(self.isOpen)
+        assert(self.lifecycleManager.isActive)
+
+        switch self.flushNowState {
+        case .inFlushNowCallingOut:
+            assert(mayHaveReentered)
+            // We're calling out from `flushNow` after `writeSocket`'s write loop has finished. We therefore need to
+            // replay the `flushNow` later because we otherwise might not write a flushed write.
+            self.eventLoop.execute {
+                self.flushNowAndRegisterIfNecessary(mayHaveReentered: false)
```
Lukasa: I think this call here is exposing us to all kinds of harm, so we need to step back and rethink the runtime model here.

There are two assertions at the top of this function call: `assert(self.isOpen)` and `assert(self.lifecycleManager.isActive)`. Neither of those assertions can be guaranteed to hold after an event loop tick has occurred: arbitrary deregistrations and closures can occur entirely synchronously in between this current write operation and the next event loop tick.

We need to come to a deterministic strategy for `flushNow` to actually flush _now_. Otherwise we need to back right out and push the `flush` call into the next event loop tick.

weissi: @Lukasa thanks! That's right

weissi: will quickly add a test just so we're sure to have this.

weissi: @Lukasa we need a method like `flushNowAndRegisterIfNecessary` though (possibly with a better name). `flushNow` is confusing as it doesn't actually do what `flush` does (setting a flush marker) but it tries to send as much flushed data as possible.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -469,32 +475,83 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
...
             do {
-                assert(!self.inFlushNow)
-                self.inFlushNow = true
+                assert(self.flushNowState == .notInFlushNow)
+                self.flushNowState = .inFlushNowWriting
                 defer {
-                    self.inFlushNow = false
+                    assert(self.flushNowState == .inFlushNowWriting)
+                    self.flushNowState = .notInFlushNow
                 }
```
Lukasa: There are way too many defer blocks floating around now, I think this code needs to be refactored. It's extremely hard to follow the ordering of operations in this relatively straightforward function. I think the call flow is straightforward enough that we can rewrite this with only a single defer block (specifically, the one that sets `flushNowState` to `.notInFlushNow`, which should be able to be set unconditionally at the top of the function.

weissi: good points, all defers and `flushNowState` gone. Changing the abstraction as you suggested instead of trying to keep the old (and busted) one alive was a much better call.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -905,9 +915,14 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
...
-        if self.flushNow() == .unregister {
-            // Everything was written or connect was complete
+
+        switch self.flushNow() {
+        case .unregister:
+            // Everything was written or connect was complete, let's unregister from writable.
             self.finishWritable()
+        case .register:
+            assert(!self.isOpen || self.interestedEvent.contains(.write))
+            () // nothing to do because givent that we just received `writable`, we're still registered for writable.
```
glbrntt:
```suggestion
            () // nothing to do because given that we just received `writable`, we're still registered for writable.
```

weissi: thanks, applied!

PR #1352, opened 2020-01-22
add SAL (Syscall Abstraction Layer) and new SALChannelTests

weissi:
## Motivation:

In SwiftNIO's codebase, most things were always testable just as well as user code is. With one big missing piece: We could never make the `Selector` do arbitrary things to the `Channel` implementations which lead to super expensive to (create, read, understand) integration tests like [`testWriteAndFlushFromReentrantFlushNowTriggeredOutOfWritabilityWhereOuterSaysAllWrittenAndInnerDoesNot `](https://github.com/apple/swift-nio/blob/5dc9083c6a78b10432b4b55b95704e61ca08f09d/Tests/NIOTests/StreamChannelsTest.swift#L468-L760).

## Modification

Add SAL, a 'syscall abstraction layer' which makes it possible to assert all syscalls that come through a `Socket` or the `Selector` by mocking them. It lets the NIO programmer play the kernel in these "SAL tests" which makes it possible to assert the exact syscalls (with their parameters) and decide the syscalls' returns.

## Result

Finally, we're able to write proper unit tests for scenarios where the `Channel` drove the `Selector` incorrectly.


Tests/NIOTests/SALChannelTests.swift
```diff
@@ -0,0 +1,236 @@
...
+            try self.assertWaitingForNotification(result: .some(.init(io: [.write],
+                                                                      registration: .socketChannel(channel, [.write]))))
+            try self.assertWrite(expectedFD: .max,
+                                 expectedBytes: buffer.getSlice(at: 1, length: 1)!,
+                                 return: .processed(1))
+            buffer.clear()
+            buffer.writeString("ABC") // expected
+
+            // This time, the write again, just writes one byte, so we should remain registered for writable.
+            // This write
```
glbrntt: This line is either incomplete or not meant to be here.

Tests/NIOTests/SyscallAbstractionLayer.swift
```diff
@@ -0,0 +1,665 @@
...
+        } else {
+            throw UnexpectedKernelReturn(ret)
+        }
+    }
+
+    override func remoteAddress() throws -> SocketAddress {
+        try self.userToKernel.waitForEmptyAndSet(.remoteAddress)
+        let ret = try self.waitForKernelReturn()
+        if case .returnSocketAddress(let localAddress) = ret {
+            return localAddress
```
glbrntt: s/local/remote

Tests/NIOTests/SyscallAbstractionLayer.swift
```diff
@@ -0,0 +1,665 @@
...
+            }
+        }
+        try hookedSelector.assertWakeup(file: file, line: line)
+        try syscallAssertions()
+        return try box.takeValue().get()
+    }
+}
+
+extension EventLoopFuture {
+    internal func teleportWait() throws -> Value {
```
glbrntt: Why the 'teleport' part of the name, because we're jumping over to the main thread?

Tests/NIOTests/SyscallAbstractionLayer.swift
```diff
@@ -0,0 +1,665 @@
...
+        if case .whenReady = syscall {
+            return
+        } else {
+            XCTFail("unexpected syscall \(syscall)", file: file, line: line)
+        }
+    }
+
+    func assertWaitingForNotification(result: SelectorEvent<NIORegistration>?,
+                                      file: StaticString = #file, line: UInt = #line) throws {
+        SAL.printIfDebug("\(#function)")
```
glbrntt: Any value in including the `result` here?

Tests/NIOTests/SyscallAbstractionLayer.swift
```diff
@@ -0,0 +1,665 @@
...
+            if case .whenReady = syscall {
+                return true
+            } else {
+                return false
+            }
+        }
+    }
+
+    func assertWakeup(file: StaticString = #file, line: UInt = #line) throws {
+        try self.selector.assertWakeup(file: file, line: line)
```
glbrntt: `printIfDebug`?

Tests/NIOTests/SALChannelTests.swift
```diff
@@ -0,0 +1,236 @@
...
+            return
+        }
+
+        var buffer = ByteBufferAllocator().buffer(capacity: 5)
+        buffer.writeString("hello")
+
+        let g = DispatchGroup()
+        g.enter()
+
+        // So we can exhaust it with 2 bytes.
```
Lukasa: This comment seems unrelated to the test.

Tests/NIOTests/SALChannelTests.swift
```diff
@@ -0,0 +1,236 @@
...
+                                 return: .processed(1))
+            buffer.clear()
+            buffer.writeString("ABC") // expected
+
+            // This time, the write again, just writes one byte, so we should remain registered for writable.
+            // This write
+            try self.assertWrite(expectedFD: .max,
+                                 expectedBytes: buffer,
+                                 return: .processed(1))
+            XCTAssertEqual(UInt8(ascii: "A"), buffer.readInteger(as: UInt8.self))
```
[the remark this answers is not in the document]
weissi: @Lukasa just my dislike of `_ =` :) ( `_ = buffer.readInteger(as: UInt8.self)` :) ). I'll just move the reader index forward by 1

Tests/NIOTests/SALChannelTests.swift
weissi: weird, yes, thanks.

Tests/NIOTests/SyscallAbstractionLayer.swift
weissi: renamed `salWait` and added a comment.

Tests/NIOTests/SyscallAbstractionLayer.swift
weissi: that's done by `self.selector.assertWakeup`

weissi:
Thanks @Lukasa & @glbrntt , addressed all comments

Tests/NIOTests/SyscallAbstractionLayer.swift
glbrntt: Oops, looks like this broke your PR :)

weissi: fixed

PR #1353, opened 2020-01-22
Recursive file reading not building up EventLoopFuture chain

fabianfett:
### Motivation:

After a long discussion with @adam-fowler today in aws-sdk-swift about how to recursively build a streaming interface (see https://github.com/swift-aws/aws-sdk-swift-core/pull/181), we came to the conclusion that it’s probably for the best not to build up an EventLoopFuture chain, since all EventLoopFutures are kept in memory while iterating. It would be better to create a promise that is captured by the recursive function, because in this case the internal EventLoopFutures can be released.

The memory footprint should be O(1) instead O(n) with a EventLoopFuture chain.

Later today, @adam-fowler made me aware of this code in NIO and that it uses a chain.

### Modifications:

```swift
_read(remainingReads: Int, bytesReadSoFar: Int64)
```

The internal read function no longer returns an EventLoopFuture. All Futures end with an `whenSuccess` or `whenFailure` case.

### Result:

From my understanding the memory footprint should be reduced when reading large files. The interface stays completely the same.

### DIFF
--- a/Sources/NIO/NonBlockingFileIO.swift
+++ b/Sources/NIO/NonBlockingFileIO.swift
@@ -174,33 +174,42 @@ public struct NonBlockingFileIO {
         precondition(chunkSize > 0, "chunkSize must be > 0 (is \(chunkSize))")
         let remainingReads = 1 + (byteCount / chunkSize)
         let lastReadSize = byteCount % chunkSize
+      
+        let promise = eventLoop.makePromise(of: Void.self)
 
-        func _read(remainingReads: Int, bytesReadSoFar: Int64) -> EventLoopFuture<Void> {
+        func _read(remainingReads: Int, bytesReadSoFar: Int64) {
             if remainingReads > 1 || (remainingReads == 1 && lastReadSize > 0) {
                 let readSize = remainingReads > 1 ? chunkSize : lastReadSize
                 assert(readSize > 0)
-                return self.read0(fileHandle: fileHandle,
-                                  fromOffset: fromOffset.map { $0 + bytesReadSoFar },
-                                  byteCount: readSize,
-                                  allocator: allocator,
-                                  eventLoop: eventLoop).flatMap { buffer in
+                let readFuture = self.read0(fileHandle: fileHandle,
+                                            fromOffset: fromOffset.map { $0 + bytesReadSoFar },
+                                            byteCount: readSize,
+                                            allocator: allocator,
+                                            eventLoop: eventLoop)
+                readFuture.whenSuccess { (buffer: ByteBuffer) in
                     guard buffer.readableBytes > 0 else {
                         // EOF, call `chunkHandler` one more time.
-                        return chunkHandler(buffer)
+                        let handlerFuture = chunkHandler(buffer)
+                        handlerFuture.cascade(to: promise)
+                        return
                     }
                     let bytesRead = Int64(buffer.readableBytes)
-                    return chunkHandler(buffer).flatMap { () -> EventLoopFuture<Void> in
+                    let handlerFuture = chunkHandler(buffer)
+                    handlerFuture.whenSuccess {
                         eventLoop.assertInEventLoop()
-                        return _read(remainingReads: remainingReads - 1,
-                                     bytesReadSoFar: bytesReadSoFar + bytesRead)
+                        _read(remainingReads: remainingReads - 1,
+                              bytesReadSoFar: bytesReadSoFar + bytesRead)
                     }
+                    handlerFuture.whenFailure { promise.fail($0) }
                 }
+                readFuture.whenFailure { promise.fail($0) }
             } else {
-                return eventLoop.makeSucceededFuture(())
+                promise.succeed(())
             }
         }
+        _read(remainingReads: remainingReads, bytesReadSoFar: 0)
 
-        return _read(remainingReads: remainingReads, bytesReadSoFar: 0)
+        return promise.futureResult
     }
 
     /// Read a `FileRegion` in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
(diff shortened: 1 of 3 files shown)


Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -174,33 +174,43 @@ public struct NonBlockingFileIO {
...
+                                            fromOffset: fromOffset.map { $0 + bytesReadSoFar },
+                                            byteCount: readSize,
+                                            allocator: allocator,
+                                            eventLoop: eventLoop)
+                readFuture.whenSuccess { (buffer: ByteBuffer) in
                     guard buffer.readableBytes > 0 else {
                         // EOF, call `chunkHandler` one more time.
-                        return chunkHandler(buffer)
+                        let handlerFuture = chunkHandler(buffer)
+                        handlerFuture.whenSuccess { promise.succeed(()) }
```
weissi: these two lines should be `handlerFuture.cascade(promise)`

Sources/NIO/NonBlockingFileIO.swift
```diff
@@ -174,33 +174,43 @@ public struct NonBlockingFileIO {
...
+                        handlerFuture.whenFailure { promise.fail($0) }
+                        return
                     }
                     let bytesRead = Int64(buffer.readableBytes)
-                    return chunkHandler(buffer).flatMap { () -> EventLoopFuture<Void> in
+                    let handlerFuture = chunkHandler(buffer)
+                    handlerFuture.whenSuccess { (_) in
```
weissi: please remove the `(_) in`, that should be unnecessary

fabianfett:
> If you take the test from #1356 and make it do 100_000 chunks then for me it gets about 6% worse.

@weissi and I talked about this in the chat... One theory is: Currently every iterations has an `init` and `retain`. The new approach would change this to `init` and `deinit`. Potentially `deinit` is more expensive than `retain`. This might be the reason for the decrease in performance.

weissi:
> > If you take the test from #1356 and make it do 100_000 chunks then for me it gets about 6% worse.
> 
> @weissi and I talked about this in the chat... One theory is: Currently every iterations has an `init` and `retain`. The new approach would change this to `init` and `deinit`. Potentially `deinit` is more expensive than `retain`. This might be the reason for the decrease in performance.

Maybe I misunderstood you earlier. I thought you said that before we had 

- ARC traffic (retains/releases) + array appends per chunk

and now we have

- allocations & deallocations (ie. `init`+`malloc`+`deinit`+`free`) + ARC traffic

per chunk. If that's the case, then the performance difference makes sense.

Above however you write that "Currently every iterations has an `init` and `retain`" which isn't possible. Every `init` must be matched with a `deinit` and every `retain` has to be matched with a `release`.

Maybe we should just add the allocation counter test in this PR and we can stop guessing :)

weissi:
I'll write one quickly.

weissi:
@fabianfett done in #1356

weissi:
@fabianfett yes, confirmed. On Swift 5.1 (on Linux), the allocations look like this:

- `master`: `"total_allocated_bytes": 18090585, "total_allocations": 250014` (ie. 25 allocations per chunk which amount to 1.76 kiB of allocations per chunk)
- `ff-recursive-file-read-without-future-chain`: `"total_allocated_bytes": 20810713, "total_allocations": 270017` (ie. 27 allocations per chunk which amounts to 2.03 kiB of allocations per chunk)

that could explain the about 6% performance difference because in this test, the actual file IO is totally free.

I'm happy with this though because in the real world, the disk IO will not be free.

fabianfett:
I have an offer for you, that you hopefully can not reject:

```bash
- /src/IntegrationTests/tests_04_performance/test_01_resources/test_read_10000_chunks_from_file.swift
test_read_10000_chunks_from_file.remaining_allocations: 0
test_read_10000_chunks_from_file.total_allocations: 210014
test_read_10000_chunks_from_file.total_allocated_bytes: 16010585
DEBUG: [["remaining_allocations": 0, "total_allocations": 210014, "total_allocated_bytes": 16010585], ["remaining_allocations": 0, "total_allocations": 210014, "total_allocated_bytes": 16010585], ["total_allocations": 210014, "total_allocated_bytes": 16010585, "remaining_allocations": 0], ["remaining_allocations": 0, "total_allocations": 210014, "total_allocated_bytes": 16010585], ["remaining_allocations": 0, "total_allocations": 210014, "total_allocated_bytes": 16010585], ["total_allocations": 210014, "remaining_allocations": 0, "total_allocated_bytes": 16010585], ["total_allocations": 210014, "remaining_allocations": 0, "total_allocated_bytes": 16010585], ["remaining_allocations": 0, "total_allocations": 210014, "total_allocated_bytes": 16010585], ["total_allocations": 210014, "remaining_allocations": 0, "total_allocated_bytes": 16010585], ["total_allocated_bytes": 16010585, "remaining_allocations": 0, "total_allocations": 210014]]
```

It looks like using `whenFailure` and `whenSuccess` comes with some memory overhead, I achieved this by just using `whenComplete`. With this we would reduce allocations by 16%. Maybe this would even result in a performance boost.

fabianfett:
> Maybe I misunderstood you earlier. I thought you said that before we had
>
> - ARC traffic (retains/releases) + array appends per chunk
>
> and now we have
>
> - allocations & deallocations (ie. init+malloc+deinit+free) + ARC traffic
>
> per chunk. If that's the case, then the performance difference makes sense.

This is what I meant! Sorry for my miscommunication. I have implied that we leaked memory before, which of course is not the case. We allocated the memory before within the loop and freed in one giant memory flush at the end. Now we have a way more continuous process.

weissi:

weissi:
wow, that's amazing @fabianfett !! It's now much faster (and allocates less):

master:
```
	 Executed 1 test, with 0 failures (0 unexpected) in 2.027 (2.028) seconds
```

your branch:
```
	 Executed 1 test, with 0 failures (0 unexpected) in 1.688 (1.688) seconds
```

very nice, now I'm extra happy with this PR :)

PR #1356, opened 2020-01-24
adding an extra test for reading many chunks

weissi:
### Motivation:

We got a great PR #1353 and I thought it'd be great to add a test that specifically tests the 'many chunks' case.

### Modifications:

- add a case that reads a file in 3000 chunks.
- add an allocation counter test reading 10000 chunks.

### Result:

more tests.


weissi:
@Lukasa probably yes, I thought about that but didn’t do it yet because #1353 doesn’t actually reduce allocations, it does reduce chaining. But you’re right, I should do it, I’ll add it to this PR, happy for you to request changes to do it so we don’t accidentally merge it

weissi:
@Lukasa done

PR #1362, opened 2020-01-28
first Swift 5.2 (on 18.04) allocation limits

weissi:
## Motivation

We need good allocation count levels for all platforms.

## Modification

This PR seeds 5.2 with a full set that is bootstrapped from the existing 5.1 set and modified where necessary (alongside comments)

## Result

Got a baseline

PR #1366, opened 2020-01-29
fix 5.2 builds

weissi:
motivation: tsan is broken on ubuntu 18.04.use 16.04 so 5.2 build can pass

changes:
* change 5.2 docker setup to use 16.04 instead of 18.04
* add the ability to pipe arguments to the integration tests via the docker setup
* skip tests_02 integration tests suite since it broken on 5.2 (swift-pm issue?)

take the new limits from #1362

Co-authored-by: tomerd@apple.com


weissi:

Issue #1358, opened 2020-01-26
too many tests that are longer than 0.5s


weissi:
We have about 50 (out of 1100+) test cases, that take more than 0.5s in CI. Many (not all) of them should/could be made faster. Here's a complete list:

 - [ ] 53.411 seconds DatagramChannelTests.testSendmmsgLotsOfData
 - [ ] 46.795 seconds ChannelTests.testWritevLotsOfData
 - [ ] 30.003 seconds NIOConcurrencyHelpersTests.testLargeContendedNIOAtomicSum
 - [ ] 29.045 seconds NIOConcurrencyHelpersTests.testLargeContendedAtomicSum
 - [ ] 11.374 seconds HTTPTest.testHTTPPipelining
 - [ ] 10.942 seconds ChannelTests.testAcceptsAfterCloseDontCauseIssues
 - [ ] 10.788 seconds HTTPServerClientTest.testMassiveResponseByteBuffer
 - [ ] 10.151 seconds NIOConcurrencyHelpersTests.testLoadAndCASHammering
 - [ ] 8.53 seconds HTTPServerClientTest.testMassiveResponseFileRegion
 - [ ] 7.012 seconds HTTPTest.testHTTPPipeliningWithBody
 - [ ] 6.965 seconds ByteBufferTest.testLargeSliceBegin16MBIsOkayAndDoesNotCopy
 - [ ] 6.954 seconds ByteBufferTest.testLargeSliceBeginMoreThan16MBIsOkay
 - [ ] 6.898 seconds NonBlockingFileIOTest.testChunkReadingWorksForIncrediblyLongChain
 - [ ] 4.315 seconds NIOConcurrencyHelpersTests.testLoadAndExchangeHammering
 - [ ] 3.844 seconds StreamChannelTest.testWriteAndFlushFromReentrantFlushNowTriggeredOutOfWritabilityWhereOuterSaysAllWrittenAndInnerDoesNot
 - [ ] 3.741 seconds NIOConcurrencyHelpersTests.testLoadAndStoreHammering
 - [ ] 3.605 seconds NIOConcurrencyHelpersTests.testMultipleLoadsRacingWhilstStoresAreGoingOn
 - [ ] 2.794 seconds SystemTest.testSystemCallWrapperPerformance
 - [ ] 2.035 seconds PriorityQueueTest.testBuildAndRemoveFromRandomPriorityQueues
[...]

PR #1367, opened 2020-01-29
make more tests faster

weissi:
Motivation:

Some tests are slow (#1358), they should be faster.

Modifications:

Make more tests faster.

Result:

Faster test suite


weissi:

PR #1368, opened 2020-01-29
include clang modularised headers last

weissi:
Motivation:

Clang modules seem to mess with stuff like _GNU_SOURCE as soon as
they're included. We do need _GNU_SOURCE (but don't like the
`#define __USE_GNU` hack).

Modifications:

Include the modularised headers last.

Result:

Alloc counter tests work without `#define __USE_GNU`.


weissi:

PR #1371, opened 2020-01-30
SR-12094 is fixed, no need to skip tests

weissi:


weissi:

PR #1373, opened 2020-01-31
dev/stackdiff-dtrace: stack aggregation diffing script

weissi:
Motivation:

NIO has allocation counter tests, NIO also has dev/malloc-aggregation.d
which can give you aggregated stacks that allocated alongside how many
times they allocated. What's missing is a script that can diff two
aggregated stack traces.

Modifications:

Add dev/stacktrace-dtrace which allows aggregated stack trace diffing.

Result:

Easier to track performance regressions.

### DIFF
--- a/dev/stackdiff-dtrace.py
+++ b/dev/stackdiff-dtrace.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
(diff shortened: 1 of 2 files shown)


dev/stackdiff-dtrace.py
```diff
@@ -0,0 +1,94 @@
...
+
+num_regex = "^ +([0-9a-fx]+)$"
+stack_regex = "^ +(.*)$"
+
+def put_in_dict(path):
+    dictionary = {}
+    with open(path, "r") as f:
+        current_text = ""
+        while True:
+            line = f.readline()
```
Lukasa: You can just do `for line in f`.

weissi: thanks done

weissi:

PR #1374, opened 2020-01-31
Avoid curried thunks (workaround SR-12115)

glbrntt:
Motivation:

Calling `function(otherFunction)` sometimes allocates, but
`function({ otherFunction($0) }` does not.

See: SR-12115 and SR-12116.

Modifications:

Avoid passing functions directly; pass them in new closures instead.

Result:

Fewer allocations.

### DIFF
--- a/docs/workarounds.md
+++ b/docs/workarounds.md
@@ -1,0 +1,13 @@
+# Workarounds
+
+The following list of PRs and commits were applied in order to work around bugs
+or cases where the Swift compiler was unable to sufficiently optimize the code:
+
+- https://github.com/apple/swift-nio/pull/1374
+- https://github.com/apple/swift-nio-ssl/pull/176
+- https://github.com/apple/swift-nio/pull/1325
+- https://github.com/apple/swift-nio/pull/1299
+- https://github.com/apple/swift-nio/pull/1252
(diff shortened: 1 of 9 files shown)


docs/workarounds.md
```diff
@@ -0,0 +1,11 @@
+# Workarounds
+
+The following list of PRs and commits were applied in order to work around bugs
+or cases where the Swift compiler was unable to sufficiently optimize the code:
+
+- https://github.com/apple/swift-nio/pull/1325
```
glbrntt:
```suggestion
- https://github.com/apple/swift-nio/pull/1374
- https://github.com/apple/swift-nio/pull/1325
```

weissi:

ktoso:
Whoa that's good to know 👀

Lukasa:
As an FYI @ktoso, most instances of this pattern do _not_ trigger extra allocations: however, we haven’t isolated the common trait of the kinds that do.

weissi:
@ktoso we believe that you can find all the bad ones by

  - compile in release mode
   - `otool -tV path-to-binary | swift demangle | grep “curry thunk”`

If that matches something that might allocate, if not you’re good

PR #1375, opened 2020-02-01
improve docker security

tomerd:
motivation: more secured ci setup

changes:
* enable :z selinux flag on bind mounts so we can enable selinux on ci
* drop potentially exploitable capabilities from docker-compose


tomerd:

PR #1376, opened 2020-02-01
get IOError down to 24 bytes (#1355)

### LINKED ISSUE #1355: IOError is 26 bytes long, should be 3 words (24) or less
### Expected behavior

All `Error` types should be 3 words or less. On 64 bit platforms that's 24 bytes. This is important because Errors get boxed in Error existentials which have 3 words of builtin storage. If that's not enough, it'll heap allocate. So we'd expect

```swift
  3> MemoryLayout<IOError>.size
$R1: Int = 24 or less
```

### Actual behavior

```swift
  3> MemoryLayout<IOError>.size
$R1: Int = 26
```

### Steps to reproduce

1. `swift run --repl`
2. `import NIO`
3. `MemoryLayout<IOError>.size`

master on 24th Jan 2020

### Suggested fix

The very likely culprit is `StaticString` which is just a bad type. It's 2 words + 1 byte wide. Given that we also store an `errno` value, it's obvious how we blow through the 24 bytes. The fix should be really straightforward: Now that `String` is good, we should just always store a `String` instead of either a `String` or a `StaticString`.

realdoug:
resolves #1355 shaving off 2 bytes from `IOError` struct

### Modifications:

This one had a bit of a curveball.  Replacing `StaticString` shaves off one byte.  The other extra byte is saved by switching the order that the `enum` and `Int32` properties are declared.  Seems like a compiler bug to me.

For example:

```
struct Good {
    enum FailureDescription {
        case function(String)
        case reason(String)
    }

    let reason: FailureDescription
    let errnoCode: Int32
}

struct Bad {
    enum FailureDescription {
        case function(String)
        case reason(String)
    }

    let errnoCode: Int32
    let reason: FailureDescription
}

print(MemoryLayout<Good>.size)
print(MemoryLayout<Bad>.size)

// 24
// 25
```

I'm curious now why that might be. 

### Result:

New condition:

`print(MemoryLayout<IOError>.size) // 24`


Lukasa:
Hmm, I wonder where the enum discriminator went in the “good” case. The only excuse for shrinking this by one byte when `FailureDescription` is clearly 20 bytes wide is because the enum discriminator got stuffed somewhere in the `Good` case that it did not in the `Bad`, but I can’t really see why that optimisation wasn’t available in both cases.

weissi:
First of all, thanks very much for this PR @realdoug ! As @Lukasa points out this is public API breaking right now. That's my mistake, I didn't quite appreciate how much we exported in our API here but I have a proposal how we can fix it all and remain API stable: Internally we just stop using the enum case that uses the `StaticString` and we can convert

```
public let reason: FailureDescription
```

to

```
private let failureDescription: String

public var reason: FailureDescription {
    return .reason(self.failureDescription
}
```

we should then also deprecate this `init`

```
public init(errnoCode: Int32, function: StaticString)
```

and rewrite it as

```
@available(*, deprecated, renamed: "init(errnoCode:reason:)")
public init(errnoCode: Int32, function: StaticString) {
    self.failureDescription = "\(function)"
}
```

realdoug:
Gotcha.  I figured you may want to paper over it somehow.  Pushed a new version w/ updates @Lukasa @weissi `Int32` + `String` results in the correct 20 bytes.

It's worth noting that this implementation loses the information distinguishing between reason/function.  Perhaps that's ok, though if so, the enum & the computed var that returns it should be deprecated as well, right?

Alternatively, I could add a private `Bool` and conditionally return the correct `FailureDescription` value.  Or perhaps the correct solution is both: use a private bool _and_ deprecate both the enum and the computed prop that returns it, and add a non-deprecated public accessor which returns a string.

realdoug:
Also, the deprecation causes many tests to emit warnings, which need to be cleaned up, but I wanted to validate the approach first.

weissi:
@realdoug nice once! Great job! The information loss is okay, the old one was just there because in Swift before 5.0, Strings were very bad so StaticString made sense.

Regarding the warnings, can’t we just make NIO use `reason: ` everywhere instead of `function: ` ? I think that could be done with a big ol’ find&replace :). 

We might want to add one single test which tests that the old `function:` still works. You can just deprecate the tests (you’ll find examples) then it won’t complain about it using deprecated functionality

Sources/NIO/IO.swift
```diff
@@ -14,15 +14,22 @@
...
         case reason(String)
     }
-    /// The `errno` that was set for the operation.
-    public let errnoCode: Int32
 
     /// The actual reason (in an human-readable form) for this `IOError`.
-    public let reason: FailureDescription
+    private var failureDescription: String
+
+    @available(*, deprecated)
```
weissi: We should add `message: “NIO no longer uses FailureReasons, [...]`

Sources/NIO/IO.swift
```diff
@@ -14,15 +14,22 @@
 
 /// An `Error` for an IO operation.
 public struct IOError: Swift.Error {
+    @available(*, deprecated)
```
weissi: A message (see below) would be great here too

Sources/NIO/IO.swift
```diff
@@ -14,15 +14,22 @@
...
-    public let reason: FailureDescription
+    private var failureDescription: String
+
+    @available(*, deprecated)
+    public var reason: FailureDescription {
+        return .reason(self.failureDescription)
+    }
+
+    /// The `errno` that was set for the operation.
+    public let errnoCode: Int32
```
weissi: I know this is unrelated to this PR but the correct type should be `CInt` instead of `Int32`. So if you fancy, you could make that change (it’s API compatible on all supported platforms)

weissi:

Tests/NIOTests/IOErrorTest+XCTest.swift
```diff
@@ -0,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: @realdoug sorry, tiny nit, can you put 2020 in the copyright instead of 2017-2018 😬

weissi:
CI passed, that failure is a spurious perf test. So it’s all good from ky PoV apart from the copyright years

realdoug:
ah my bad @weissi copyright year is fixed now.

weissi:

weissi:

weissi:

Tests/NIOTests/IOErrorTest+XCTest.swift
realdoug: did you mean to set this one back to 2017-2018?  That was my fault for setting IO.swift, should have been just the new files IIUC.  I'll push the correct update if so.

weissi: @realdoug yeah, sorry, the autogenerated files need to remain as they are. Our CI gets unhappy otherwise (it checks the files aren’t edited). I know I commented (by accident) on the auto generated file :). Sorry about that

weissi:

weissi:

realdoug:
Awesome!

realdoug:
Awesome!

PR #1380, opened 2020-02-04
ByteBuffer: make getInteger for UInt8 faster

weissi:
Motivation:

getInteger used the normal, generic implementation for UInt8s, however
they can be made considerably faster by using a specialised method.
ByteBufferIterator was also affected by this.

Modifications:

Use a specialised version for UInt8s for getInteger.

Result:

- ByteBufferView can be iterated about 20% faster.

### DIFF
--- a/Sources/NIOPerformanceTester/main.swift
+++ b/Sources/NIOPerformanceTester/main.swift
@@ -762,6 +762,8 @@ try measureAndPrint(desc: "websocket_decode_64kb_+1_100k_frames",
 try measureAndPrint(desc: "websocket_decode_64kb_+1_with_а_masking_key_100k_frames",
                     benchmark: WebSocketFrameDecoderBenchmark(dataSize: Int(UInt16.max) + 1, runCount: 100_000, maskingKey: [0x80, 0x08, 0x10, 0x01]))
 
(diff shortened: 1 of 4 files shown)


Sources/NIOPerformanceTester/main.swift
```diff
@@ -762,6 +762,8 @@ try measureAndPrint(desc: "websocket_decode_64kb_+1_100k_frames",
 try measureAndPrint(desc: "websocket_decode_64kb_+1_with_а_masking_key_100k_frames",
                     benchmark: WebSocketFrameDecoderBenchmark(dataSize: Int(UInt16.max) + 1, runCount: 100_000, maskingKey: [0x80, 0x08, 0x10, 0x01]))
 
-try measureAndPrint(desc: "circular_buffer_into_byte_buffer_1kb", benchmark: ByteBufferBenchmark(iterations: 10000, bufferSize: 1024))
+try measureAndPrint(desc: "circular_buffer_into_byte_buffer_1kb", benchmark: ByteBufferIntoCircularBufferBenchmark(iterations: 10000, bufferSize: 1024))
```
glbrntt: nit: the `desc` and benchmark disagree with who is writing into who

weissi: thanks @glbrntt , fixed

PR #1386, opened 2020-02-06
add Swift 5.2 to the readme

weissi:
Motivation:

Swift 5.2 is already available as developer snapshots and Xcode 11.4
beta got released which also includes Swift 5.2.

Modifications:

- add 5.2 support to the readme
- showcase SwiftPM's package syntax for 5.2

Result:

the future is here.


Lukasa:
Where we're going we don't need roads.

PR #1387, opened 2020-02-10
Update http-parser for CVE.

Lukasa:
Motivation:

http-parser shipped a patche for node.js CVE-2019-15605, which allowed
HTTP request smuggling. This affected SwiftNIO as well, and so we need
to immediately ship an update to help protect affected users.

A CVE for SwiftNIO will follow, but as this patch is in the wild and
SwiftNIO is known to be affected we should not delay shipping this fix.

Modifications:

- Update http-parser.
- Add regression tests to validate this behaviour.

Result:

Close request smugging vector.

### DIFF
--- a/Sources/CNIOHTTPParser/c_nio_http_parser.c
+++ b/Sources/CNIOHTTPParser/c_nio_http_parser.c
@@ -384,7 +384,10 @@ enum header_states
   , h_transfer_encoding
   , h_upgrade
 
+  , h_matching_transfer_encoding_token_start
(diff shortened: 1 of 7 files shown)


Lukasa:
This patch is against master, but we have shipped an emergency 2.13 release with df9390006bce7da1b6273f804d3acbbfdfcc6154. That's a backport of the fix we provided here.

Lukasa:

Lukasa:

PR #1388, opened 2020-02-10
Update http-parser for CVE.

Lukasa:
Motivation:

http-parser shipped a patche for node.js CVE-2019-15605, which allowed
HTTP request smuggling. This affected SwiftNIO as well, and so we need
to immediately ship an update to help protect affected users.

A CVE for SwiftNIO will follow, but as this patch is in the wild and
SwiftNIO is known to be affected we should not delay shipping this fix.

Modifications:

- Update http-parser.
- Add regression tests to validate this behaviour.

Result:

Close request smugging vector.

(cherry picked from commit f94b22b506e3557cb1b325534fa9bbcd39c90246)

### DIFF
--- a/Sources/CNIOHTTPParser/c_nio_http_parser.c
+++ b/Sources/CNIOHTTPParser/c_nio_http_parser.c
@@ -28,6 +28,8 @@
 #include <string.h>
 #include <limits.h>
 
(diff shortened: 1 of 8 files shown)


Lukasa:
The 1.14 CI seems to be entirely busted (cc @tomerd) but we've validated this fix locally, so we're going to ship it anyway.

PR #1394, opened 2020-02-13
Add custom index methods to ByteBufferView

PeterAdams76:
Motivation:

The performance of firstIndex and lastIndex is
dramatically improved by addding custom
indexing methods as detailed in issue 1385.

Modifications:

Added methods to ByteBufferView:
   _customIndexOfEquatableElement
       for firstIndex
   _customLastIndexOfEquatableElement
       for lastIndex

Added a unit test for firstIndex and lastIndex
to ensure functional behaviour is unchanged.

Result:

firstIndex and lastIndex behave as before.

Performance is improved by approximately 20% for
the shortest searches.
When tested searching over 1024 byte buffer
with either no match or a match at the end
I observed approximately 15 times performance
improvement.

### DIFF
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -86,23 +86,15 @@ public struct ByteBufferView: RandomAccessCollection {
     }
 
     public func _customIndexOfEquatableElement(_ element : Element) -> Index?? {
-        let indexInRange = self.withUnsafeBytes { (ptr) -> Index? in
-            return ptr.firstIndex(of: element)
-        }
-        if let indexFoundInRange = indexInRange {
-            return Optional(Optional(indexFoundInRange + self._range.lowerBound))
-        }
-        return Optional(Optional(nil))
+        return .some(self.withUnsafeBytes { ptr -> Index? in
+            return ptr.firstIndex(of: element).map { $0 + self._range.lowerBound }
+        })
     }
 
     public func _customLastIndexOfEquatableElement(_ element: Element) -> Index?? {
-        let indexInRange = self.withUnsafeBytes { (ptr) -> Index? in
-            return ptr.lastIndex(of: element)
-        }
-        if let indexFoundInRange = indexInRange {
-            return Optional(Optional(indexFoundInRange + self._range.lowerBound))
-        }
-        return Optional(Optional(nil))
+        return .some(self.withUnsafeBytes { ptr -> Index? in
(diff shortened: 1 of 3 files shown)


Sources/NIO/ByteBuffer-views.swift
```diff
@@ -84,6 +84,26 @@ public struct ByteBufferView: RandomAccessCollection {
             return try body(bytes.bindMemory(to: UInt8.self))
         }
     }
+
+    public func _customIndexOfEquatableElement(_ element : Element) -> Index?? {
+        let indexInRange = self.withUnsafeBytes { (ptr) -> Index? in
```
weissi: I think this could be written more easily as

    return .some(self.withUnsafeBytes { ptr -> Index? in
        return ptr.firstIndex(of: element).map { $0 + self._range.lowerBound }
    })

but the generated code should be identical :).

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -84,6 +84,26 @@ public struct ByteBufferView: RandomAccessCollection {
...
+            return ptr.firstIndex(of: element)
+        }
+        if let indexFoundInRange = indexInRange {
+            return Optional(Optional(indexFoundInRange + self._range.lowerBound))
+        }
+        return Optional(Optional(nil))
+    }
+
+    public func _customLastIndexOfEquatableElement(_ element: Element) -> Index?? {
+        let indexInRange = self.withUnsafeBytes { (ptr) -> Index? in
```
weissi: similar here

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1939,6 +1939,34 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual([], self.buf.readBytes(length: self.buf.readableBytes))
     }
 
+    func testBufferViewFirstIndex() {
+          self.buf.clear();
```
weissi: nit: extra semicolon

PeterAdams76: Yes, 6 of them.  I will remove.  I also notice my indent doesn't match the file either which I will also resolve.

Sources/NIO/ByteBuffer-views.swift
PeterAdams76: Agree - your way is much neater.  I will adopt.

weissi:

weissi:

weissi:

PR #1395, opened 2020-02-13
Prevent execution of tasks on an already shutdown EL

### LINKED ISSUE #1383: NIO should assert/precondition is work is execute/submit/flatSubmitted so a closed EventLoop
When anybody enqueues work on an EventLoop that has been shut down, currently, you'll likely just hang without any feedback. Closing an EventLoop(Group) prematurely is programmer error so we should crash.

realdoug:
Resolve #1383 ; crash if a task is attempted on a closed Event Loop

One note: I created some test cases [here](https://github.com/apple/swift-nio/issues/1383#issuecomment-585754801), but wasn't able to figure out how to integrate them into IntegrationTests but perhaps you can point me in the right direction and I'll add them.

### Modifications:
Make `SelectableEventLoop._scheduledTasks` an optional, set it to null when shutting down and throw a precondition failure if task enqueuing is attempted while it is null (i.e. after shutdown).

### Result:

[see test cases](https://github.com/apple/swift-nio/issues/1383#issuecomment-585754801)


Sources/NIO/SelectableEventLoop.swift
```diff
@@ -146,13 +146,22 @@ internal final class SelectableEventLoop: EventLoop {
         }
     }
 
+    public var isOpen: Bool {
+        switch self.externalState {
+            case .open, .closing:
```
realdoug: `execute` gets called when `self.externalState` = `.closing` [here](https://github.com/apple/swift-nio/issues/1383#issuecomment-585754801).  We *could* create an backdoor that one case, but I figured since there are currently no checks, this still covers what I imagine to be the majority of cases.

Sources/NIO/EventLoop.swift
```diff
@@ -596,6 +602,10 @@ extension EventLoop {
     public func preconditionInEventLoop(file: StaticString = #file, line: UInt = #line) {
         precondition(self.inEventLoop, file: file, line: line)
     }
+
+    public var isOpen: Bool {
```
Lukasa: Can we mark this as the default implementation of this protocol requirement, and to be removed in 3.0? It's a bit of a nasty hack, and we'd like to avoid it being used if we can.

Sources/NIO/EventLoop.swift
```diff
@@ -596,6 +602,10 @@ extension EventLoop {
...
+        true
```
Lukasa: This won't compile in our set of Swift versions: we need the explicit `return` statement.

Sources/NIO/EventLoop.swift
```diff
@@ -463,6 +467,8 @@ extension EventLoop {
     /// - returns: An `EventLoopFuture` containing the result of `task`'s execution.
     @inlinable
     public func submit<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T> {
+        precondition(self.isOpen, "cannot submit work to a closed EventLoop")
```
Lukasa: `isOpen` here has some undocumented thread safety assumptions that I suspect it fails to meet. In fact, the implementation on `MultithreadedEventLoopGroup` is definitely thread _unsafe_ for this usage model. Remember that `submit` is valid to call from any thread, so it cannot safely read data that is unsynchronised.

In this case, I think the design cannot solve this problem generically: it's a requirement of the event loop implementation specifically to solve this problem. This is because the baseline function that needs to be correct is `eventLoop.execute`, which is entirely up to the individual event loop to implement. We therefore can only fix it for `MultithreadedEventLoopGroup`, and document that other event loops should also endeavour to meet this requirement.

Does that make sense?

Sources/NIO/EventLoop.swift
realdoug: Yes I agree with that, ideally we don’t have to do this at all.  Will do.

realdoug: @lukasa actually after reading your other comment, we may not need to change this protocol at all, if the `submit` function isn’t calling isOpen.  Perhaps this PR should just solve the issue in the context of MultithreadedEventLoop and not touch EventLoop protocol.

Lukasa: Sounds like a good idea to me.

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -869,7 +869,7 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             return
         }
 
-        guard self.selectableEventLoop.isOpen else {
+        guard self.selectableEventLoop.acceptingRegistrations else {
```
realdoug: This rename isn't strictly necessary anymore, but I thought it was a better name. Can change it back if preferred.

realdoug:
@Lukasa @weissi Alrighty, made some updates to this one, lmk what you think.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -136,7 +136,7 @@ internal final class SelectableEventLoop: EventLoop {
     }
 
     /// Is this `SelectableEventLoop` still open (ie. not shutting down or shut down)
-    internal var isOpen: Bool {
+    internal var acceptingRegistrations: Bool {
```
realdoug: This rename isn't strictly necessary anymore, but I thought it was a better name. Can change it back if preferred.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -146,13 +146,28 @@ internal final class SelectableEventLoop: EventLoop {
         }
     }
 
+    public var safeToExecute: Bool {
```
Lukasa: I'm inclined to make this `internal`: the type is `internal` so the access control of this computed property doesn't _really_ matter, but I don't think we'd normally want people to be able to use it.

Additionally, we'd normally want it to be `private` but we want access to it for testing purposes. The convention in NIO to annotate this is to use:

```swift
/* private but internal */ var _safeToExecute: Bool
```

This comments why the access modifier is higher than we want, and also prefixes it with an underscore to indicate to other parts of the code that we probably shouldn't use it.

Sources/NIO/SelectableEventLoop.swift
Lukasa: I think `.closing` is fine: once we're actually _closed_ we want to reject new work, but while we're merely `closing` that's fine

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -146,13 +146,28 @@ internal final class SelectableEventLoop: EventLoop {
...
+        if self.inEventLoop {
+            return true
+        }
```
Lukasa: I don't think this escape hatch is right. We can be in the event loop but at a stage where further enqueues of work will fail. In fact, the obvious situation where that happens is when the event loop is shut down. At that point we'll dequeue all pending tasks that have not yet executed and fail them. Failing those tasks dispatches callbacks, and those callbacks may try to call back around to add work to the task queue. That's a bad pattern, and we should catch it, but we won't if this hook is here.

I think we only want the external state check.

Lukasa: Actually, I think we can do better.

The `MultithreadedEventLoopGroup` currently uses a `_scheduledTasks` `PriorityQueue` to store pending work items. This queue is guarded by `_tasksLock`. I propose that we wrap the `PriorityQueue` in a data structure that allows us to "scribble" it: to mark it permanently as unusable once we're sure we aren't going to touch it again.

We can do this just by making the `PriorityQueue` optional, I think, and force unwrapping it whenever we try to modify it. That will allow us to catch the case where tasks are enqueued too early. It doesn't give a good error message, but that might be fine.

If we really want a good error message, a custom enum that is basically an optional with the method we want would also work.

This avoids the need for a new computed property, and also avoids synchronisation concerns: we don't have to worry about communication between two locks. It also allows us to use the point in the actual code where we know no further tasks can execute (the `defer` block at the top of `func run()`) to guarantee that outcome.

realdoug: Ah, makes sense!  Looking into it.

Sources/NIO/SelectableEventLoop.swift
weissi: I'm fine with this, maybe you should expand the comment a little though.

Sources/NIO/SelectableEventLoop.swift
weissi: I agree

Sources/NIO/SelectableEventLoop.swift
realdoug: Ok, pushed up a new approach with _scheduledTasks as an optional

realdoug:
> Ok cool, I think I like this. I'd like to see what @weissi thinks.

Awesome! Happy to make updates if needed.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -56,7 +56,7 @@ internal final class SelectableEventLoop: EventLoop {
     /* private but tests */ internal let _selector: NIO.Selector<NIORegistration>
     private let thread: NIOThread
     @usableFromInline
-    internal var _scheduledTasks = PriorityQueue<ScheduledTask>(ascending: true)
+    internal var _scheduledTasks: PriorityQueue<ScheduledTask>? = PriorityQueue<ScheduledTask>(ascending: true)
```
weissi: @realdoug one question here: Why do we make this optional instead of say adding a state to the `InternalState`?

Lukasa: That was my suggestion: https://github.com/apple/swift-nio/pull/1395#discussion_r380049404

Sources/NIO/SelectableEventLoop.swift
realdoug: @weissi @Lukasa I think you could break this task into 3 parts:

1) identify the minimum set of places where a check needs to be made before executing
2) identify what to query as part of the check
3) based on step 2, possibly write some data on close that can correctly fulfill the query

Since tasks can be enqueued from inside and outside the thread, I don't _think_ there is any solution to step 1 that would allow us to use just `internalState` or just `externalState` safely.

My original approach branched on `inEventLoop` (though as @Lukasa pointed out I missed a test case), alternatively, current approach adds a bit of info that can be queried from `tasksLock` and thus all cases from step 1.

weissi: @realdoug hmm, I'm somewhat lost here. I'm pretty sure we can find out what state we're in wherever we are because:

- if we're on the `EventLoop`, we can learn that we won't be processing new tasks if the `internalState` is `noLongerRunning`
- if we're off the `EventLoop`, then we should be in `.closing`/`.closed` when we're shutting down

If we need another internal (or external) state, for example `.aboutToExitThread ` or something, then I'm happy for you to add such a state. So for example at the very point where you currently set `self._scheduledTasks = nil` you could

    assert(self.state == .noLongerRunning, "illegal state: \(self.state)")
    self.state = . aboutToExitThread

Then in `schedule0` you could

```swift
if self.inEventLoop {
    switch self.internalState {
        ...
    }
} else {
    switch self.externalState {
        ...
    }
}
```

and in those `switch`es you can then do the appropriate checks depending on whether we're inside or outside the eventloop.

Does that make sense?

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -228,7 +236,7 @@ internal final class SelectableEventLoop: EventLoop {
     @inlinable
     internal func execute(_ task: @escaping () -> Void) {
         // nothing we can do if we fail enqueuing here.
-       try? self._schedule0(ScheduledTask(task, { error in
+        try? self._schedule0(ScheduledTask(task, { error in
```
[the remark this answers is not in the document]
realdoug: i think this was an existing typo, I added a 4th space here.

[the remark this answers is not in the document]
weissi: Apologies!

Sources/NIO/SelectableEventLoop.swift
realdoug: Sorry, I think I got excited and wrote too much :)  My only point was that there are two solutions: an if statement or an optional.  Using just one of the two states won't work.

I pushed up a new version w/ an if statement inside `_schedule0` based on your description.  Let me know what you think.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -104,6 +105,26 @@ internal final class SelectableEventLoop: EventLoop {
         }
     }
 
+    @usableFromInline
+    internal var _validInternalStateToScheduleTasks: Bool {
+        switch self.internalState {
+        case .exitingThread:
+            return false
+        default:
```
weissi: could you change this from `default:` to enumerating the cases explicitly like `case .running..., .running..., .noLongerRunning`? It's important that we get the compiler errors if we add a case.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -104,6 +105,26 @@ internal final class SelectableEventLoop: EventLoop {
...
+            return true
+        }
+    }
+
+    @usableFromInline
+    internal var _validExternalStateToScheduleTasks: Bool {
+        switch self.externalState {
+        case .open, .closing:
```
weissi: could you change this from `default:` to enumerating the cases explicitly like `case .running..., .running..., .noLongerRunning`? It's important that we get the compiler errors if we add a case.

realdoug: @weissi ok done & done

weissi:

weissi:
CC @Lukasa

weissi:

PR #1398, opened 2020-02-18
gitignore: ignore vim swap files

compnerd:

### Motivation:

### Modifications:

### Result:


Lukasa:

Issue #1377, opened 2020-02-03
Replace CNIOSHA1 with SwiftCrypto


weissi:
I'd be fantastic to remove the `CNIOSHA1` module and replace it with `Crypto.Insecure.SHA1` (which sadly is mandated by the WebSocket spec). Unfortunately, SwiftNIO 2 supports Swift 5.0+ and [`SwiftCrypto`](https://github.com/apple/swift-crypto) requires Swift 5.1.

0xTim:
@weissi @Lukasa so why does this require a major version bump?

I looked through one of the [old posts](https://forums.swift.org/t/semantic-versioning-in-swift/12720/29) on the forums about SemVer and Swift and saw no discussion about language version changes. IMO changing the language version is a minor release and not a major release for these reasons:
* There are no API changes (even adding Crypto, any existing functions that would be replaced can be wrapped to maintain the public API). This is all that SemVer guarantees
* SwiftPM should stop anyone building with Swift 5 from pulling in a Swift 5.1 package, so again nothing should break
* Releasing a minor allows security updates and other major bug fixes to be released in patch versions.

Alamofire are also [currently considering](https://forums.swift.org/t/requiring-swift-5-1/33866) a minor release to bump the minimum Swift version.

Completely understand if it's a team policy or similar, it's your code after all! 😄 But would be good to know the reasons.

Lukasa:
The risk isn't the raise in the Swift version number, but the OS constraints added by Swift Crypto. This is shown [in SR-11489](https://bugs.swift.org/browse/SR-11489), but the core problem is that to depend on Swift Crypto we need to raise the floor of our minimum deployment targets. This immediately breaks _all_ downstreams, which have to raise their floors as well. SwiftPM does not bring the minimum OS deployment targets into the dependency resolution phase, so it will happily choose a NIO version whose minimum deployment target is not suitable for the leaf package and then fail that build.

0xTim:
Ahhh got you, that makes sense. Thanks!

PR #1399, opened 2020-02-18
CNIOSHA1: make the code more portable

compnerd:
Replace the non-portable `u_*_t` types with the C standard `u*_t` types
instead.  Remove the Unix headers as the only headers really needed are
the C standard headers.  Provide shims for deprecated `strings.h`
functions in terms of "modern" (C89+) functions from `string.h`.  Move
an android only include into the implementation from the interface.

This enables building this library on Windows.

### Motivation:

### Modifications:

### Result:


weissi:
@compnerd agree with Cory here. Your changes however seem scriptable maybe? Couldn’t we apply these changes in the updater script using `sed`? I’d like to continue being able to update from upstream with the script if at all possible

0xTim:
Would make NIO rely on Crypto and use the SHA1 function from there be pulling in too much?

Lukasa:
That's #1377. As noted in that issue, it's a semver major.

compnerd:
That would also require bringing Crypto to Windows first.  I should be able to script it, it just may not be pretty :)

weissi:
@compnerd this looks really good, unfortunately there are now some warnings:

```
03:26:40 In file included from /code/Sources/CNIOSHA1/c_nio_sha1.c:45:
03:26:40 /code/Sources/CNIOSHA1/include/CNIOSHA1.h:74:66: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
03:26:40 extern void c_nio_sha1_loop(struct sha1_ctxt *, const uint8_t *, size_t);
03:26:40                                                                  ^
03:26:40 /code/Sources/CNIOSHA1/c_nio_sha1.c:232:1: error: conflicting types for 'c_nio_sha1_loop'
03:26:40 c_nio_sha1_loop(ctxt, input, len)
03:26:40 ^
03:26:40 /code/Sources/CNIOSHA1/include/CNIOSHA1.h:74:13: note: previous declaration is here
03:26:40 extern void c_nio_sha1_loop(struct sha1_ctxt *, const uint8_t *, size_t);
03:26:40             ^
03:26:40 1 warning and 1 error generated.
```

compnerd:
It's unclear if the failures are related to this change, but they are appearing on all the versions.

Sources/CNIOSHA1/c_nio_sha1.c
```diff
@@ -44,15 +42,15 @@
  * implemented by Jun-ichiro itojun Itoh <itojun@itojun.org>
  */
 
-#include <sys/cdefs.h>
-
+#include "include/CNIOSHA1.h"
+#include <string.h>
+#define bzero(b,l) memset((b), '\0', (l))
```
Lukasa: These redefines emit warnings on macOS due to redefining the macro. I think we'll need a different solution.

Lukasa:
Yeah, this looks like the SHA1 hashing isn't right anymore. We simply get the wrong value out of the hasher. This is pretty wild: the hasher is entirely stable, always generating the _same_ wrong answer, but the answer is definitely wrong.

ASAN and UBSAN don't reveal anything.

Lukasa:
Yeah, ok, this seems to be an endianness issue. For some reason Linux is incorrectly determining the endianness of the system as big-endian. Let me dig into why.

compnerd:
@Lukasa - I *think* that we can just include `<sys/types.h>` on `__linux__` instead of `__ANDROID__` and that should fix the endianness.

Lukasa:
WFM.

compnerd:
I generalized it even further to `__unix__` so that BSDs are correct as well

Sources/CNIOSHA1/c_nio_sha1.c
Lukasa: @compnerd I assume these warnings are still appearing.

Lukasa: Yup, they are. I'd like us to address this before we merge.

compnerd: I completely missed this bit.  Um, Im not sure what a good way to handle that is.  Personally, I think that we should just silence the warning - macro redefinition is portable - it is part of the standard as of C99.  We could also simply just set the C standard to C99 rather than ignoring the warning.  Thoughts?

Lukasa: I'm fine with setting the C standard to C99, I cannot imagine it would break anything.

compnerd:
Bleh, seems that s-p-m is too old to support the language version.  I guess time to make the script uglier.

compnerd:
@Lukasa - okay, addressed the warning, its not pretty, but works

weissi:

compnerd:
I'm not sure I understand how this is impacting the syscall wrapping :-(

Sources/CNIOSHA1/update_and_patch_sha1.sh
```diff
@@ -41,17 +39,21 @@ for f in sha1.c sha1.h; do
...
-            -e 's%#ifdef _KERNEL%#define __min_size(x)	static (x)%g' \
-            -e 's%#endif /\* _KERNEL \*/%%g' \
-            -e 's%__FBSDID("$FreeBSD$");%%g' \
             "$here/c_nio_$f"
     done
 done
 
-gsed -i $'/#define _CRYPTO_SHA1_H_/a #ifdef __ANDROID__\\\n#include <sys/endian.h>\\\n#endif' "$here/c_nio_sha1.h"
-gsed -i '/#define _CRYPTO_SHA1_H_/a #include <sys/types.h>' "$here/c_nio_sha1.h"
+gsed -e $'/#define _CRYPTO_SHA1_H_/a #include <stdint.h>\\\n#include <stddef.h>' \
```
glbrntt: Drive by comment:

```suggestion
$sed -e $'/#define _CRYPTO_SHA1_H_/a #include <stdint.h>\\\n#include <stddef.h>' \
```

Sources/CNIOSHA1/update_and_patch_sha1.sh
```diff
@@ -41,17 +39,21 @@ for f in sha1.c sha1.h; do
...
+     -e 's/u_int\([0-9]\+\)_t/uint\1_t/g'                                        \
+     -e 's%#ifdef _KERNEL%#define __min_size(x)	static (x)%g'                    \
+     -e 's%#endif /\* _KERNEL \*/%%g'                                            \
+     -i "$here/c_nio_sha1.h"
+
+gsed -e 's/u_int\([0-9]\+\)_t/uint\1_t/g'                                        \
```
glbrntt:
```suggestion
$sed -e 's/u_int\([0-9]\+\)_t/uint\1_t/g'                                        \
```

Sources/CNIOSHA1/update_and_patch_sha1.sh
glbrntt: Does that matter here? Doesn't `$sed` always refer to GNU sed but it's spelt `gsed` on Darwin and `sed` elsewhere?

Sources/CNIOSHA1/update_and_patch_sha1.sh
weissi: oh sorry, you're right!!

Sources/CNIOSHA1/update_and_patch_sha1.sh
weissi: oh sorry, you're right. Yeah, we need to make this `$sed`

PR #1400, opened 2020-02-18
CNIOAtomics: declare atomic operations using a macro

compnerd:
Use a macro to define the interface for the typed atomics.  This reduces
the boiler plate that is expanded by hand.  More importantly, it makes
it easier to add new types in the future.

### Motivation:

### Modifications:

### Result:


Lukasa:

Lukasa:

PR #1403, opened 2020-02-18
NIOConcurrencyHelpers: port to Windows

compnerd:
Add support for Windows to the `NIOConcurrencyHelpers`, replacing the
`pthreads` usage for Windows threading primitives.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIOConcurrencyHelpers/atomics.swift
+++ b/Sources/NIOConcurrencyHelpers/atomics.swift
@@ -19,6 +19,12 @@ import Darwin
 fileprivate func sys_sched_yield() {
     pthread_yield_np()
 }
+#elseif os(Windows)
+import ucrt
+import WinSDK
+fileprivate func sys_sched_yield() {
+  Sleep(0)
+}
 #else
 import Glibc
 fileprivate func sys_sched_yield() {
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -14,6 +14,9 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
+#elseif os(Windows)
+import ucrt
+import WinSDK
 #else
 import Glibc
 #endif
@@ -24,21 +27,35 @@ import Glibc
 /// of lock is safe to use with `libpthread`-based threading models, such as the
 /// one used by NIO.
 public final class Lock {
-    fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> = UnsafeMutablePointer.allocate(capacity: 1)
+#if os(Windows)
+    fileprivate let mutex: UnsafeMutablePointer<SRWLOCK> =
+        UnsafeMutablePointer.allocate(capacity: 1)
+#else
+    fileprivate let mutex: UnsafeMutablePointer<pthread_mutex_t> =
+        UnsafeMutablePointer.allocate(capacity: 1)
+#endif
 
     /// Create a new lock.
     public init() {
+#if os(Windows)
+        InitializeSRWLock(self.mutex)
+#else
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr);
         pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK));
 
         let err = pthread_mutex_init(self.mutex, &attr)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
+#endif
     }
 
     deinit {
+#if os(Windows)
+        // mutexes do not need to be explicitly destroyed
+#else
         let err = pthread_mutex_destroy(self.mutex)
         precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
+#endif
         mutex.deallocate()
     }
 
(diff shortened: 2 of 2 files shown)


Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -160,6 +204,9 @@ public final class ConditionLock<T: Equatable> {
     public func lock(whenValue wantedValue: T, timeoutSeconds: Double) -> Bool {
         precondition(timeoutSeconds >= 0)
 
+#if os(Windows)
+          fatalError()
```
compnerd: This needs to be ported over, but cannot be accurate as `SleepConditionVariableSRW` takes a relative time rather than absolute time.  It also would have a +/-10ns time-frame as timers on Windows have a 10ns granularity.

Sources/NIOConcurrencyHelpers/atomics.swift
```diff
@@ -19,11 +19,17 @@ import Darwin
 fileprivate func sys_sched_yield() {
     pthread_yield_np()
 }
-#else
+#elseif os(Android) || os(Linux)
```
Lukasa: Might be better to leave this in an `else` block as a catch-all, and just bring the `elseif os(Windows)` to the top.

Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -14,8 +14,11 @@
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
 import Darwin
-#else
+#elseif os(Android) || os(Linux)
```
Lukasa: Same here.

Sources/NIOConcurrencyHelpers/lock.swift
Lukasa: I don't think I'm overly worried about timer granularity, so I don't really consider that to be a problem. I'm not sure I follow the relative vs absolute time concerns though.

Sources/NIOConcurrencyHelpers/atomics.swift
compnerd: I think that the catch all is incorrect - why assume that `glibc` is the underlying libc?  Why not `musl`?  or `bionic`?  or `ucrt`?  Explicitly stating the underlying libc is a better approach IMO.

Lukasa: Because the decision matrix works out better. If we don't guess, on any unrecognised platform we 100% fail to build. If we do guess, some amount of the time the backing libc actually _will_ be `glibc` and we'll compile. That's a strictly better outcome.

Additionally, I'm deeply unconvinced by the idea that other libcs are going to be exposed as different imports. They could be, but if they are then we're going to have a _really_ hard time adopting new libcs on Linux, which unconditionally assumes `glibc` is the name for libc. I suspect for better or worse we've accidentally frozen the idea that Posix `libc` is imported as `glibc` even when it isn't.

compnerd: I think that we will need to revisit that decision as it doesn't hold.  I would like to be able to build against musl as well.  In the mean time, sure, I can change the condition to prevent a break for the time being.

Lukasa: I totally view building against musl as a concrete goal for Swift. I just suspect that we'll end up saying "musl is glibc" rather than forcing the whole Swift on Linux ecosystem to pretend otherwise. That's what C does, after all. :grin:

compnerd: There are header differences, so it will have to be `musl` ... the module layout is just not compatible enough.

Sources/NIOConcurrencyHelpers/atomics.swift
```diff
@@ -19,11 +19,17 @@ import Darwin
...
 import Glibc
 fileprivate func sys_sched_yield() {
     _ = sched_yield()
 }
+#elseif os(Windows)
+import ucrt
+import WinSDK
+fileprivate func sys_sched_yield() {
```
weissi: this is only used in deprecated functionality, so if you want, just do nothing on Windows

Sources/NIOConcurrencyHelpers/lock.swift
compnerd: Just that the sleep will not be exactly precise.  I will have some overheads due to adjustments for the remaining sleep just in case of a spurious wakeup.

Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -160,6 +204,28 @@ public final class ConditionLock<T: Equatable> {
     public func lock(whenValue wantedValue: T, timeoutSeconds: Double) -> Bool {
         precondition(timeoutSeconds >= 0)
 
+#if os(Windows)
+        var dwMilliseconds: DWORD = DWORD(timeoutSeconds * 1000)
+        while true {
+          if self._value == wantedValue {
```
weissi: I think this is reading `_value` without a lock which is undefined behaviour

compnerd: Well, thats not the only issue - the mutex was never locked ... this is missing a `self.lock`.

weissi: yep, that's what I mean :)

compnerd: Nice catch!  I may have to ask you to do this once more ... the MSVCRT port for System.swift is ... challenging.

PR #1404, opened 2020-02-18
NIO: excise UDS support for Windows

compnerd:
This removes support for Unix Domain Sockets from NIO on Windows.
Although Windows 10.17063 introduced basic support for UDS, the support
is incomplete and has differences from the tradtional UDS sockets in
terms of behaviour.  Excise the support initially to get NIO to build on
Windows.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -52,10 +52,12 @@ extension UnsafeMutablePointer where Pointee == sockaddr {
             return self.withMemoryRebound(to: sockaddr_in6.self, capacity: 1) {
                 SocketAddress($0.pointee, host: $0.pointee.addressDescription())
             }
+#if !os(Windows)
         case Posix.AF_UNIX:
             return self.withMemoryRebound(to: sockaddr_un.self, capacity: 1) {
                 SocketAddress($0.pointee)
             }
+#endif
         default:
             return nil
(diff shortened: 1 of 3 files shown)


compnerd:
One question that comes to mind: should this be a general option rather than platform conditional?  I imagine that WASM may not implement UDS either and so more than just windows would want the `UDS` support made optional.

Lukasa:
> I imagine that WASM may not implement UDS either and so more than just windows would want the UDS support made optional.

WASM will want a lot more removed than just UDS, I think: NIO won't function at all on WASM as it relies on far too much of the system.

When WASM becomes something we really want to worry about we may need to consider a substantial refactor of the APIs. In particular, we probably will need to separate the bits of NIO that are truly platform agnostic (things like protocols, `ChannelPipeline`, `ChannelHandler`, various helper objects) from the things that are actually secretly "NIOPOSIX" (`SelectableEventLoop`, `MultithreadedEventLoopGroup`, and so-on).

I don't think we want to go too far down that road today, because it's a pretty substantial breaking change, but if we want to get serious about WASM it'll necessarily come up.

Lukasa:
I'm marking this as a Semver patch because, well, we never worked on Windows before so this can't break existing APIs.

Lukasa:

PR #1405, opened 2020-02-18
[WIP] NIO: port to Windows

compnerd:
Define the socket types in terms of WinSock types.  This allows building
NIO for Windows.

### Motivation:

### Modifications:

### Result:


Sources/NIO/BSDSocket.swift
```diff
@@ -0,0 +1,897 @@
...
+    }
+    return .processed(size_t(iResult))
+#else
+#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+    let recv: (CInt, UnsafeMutableRawPointer, size_t, CInt) -> ssize_t = Darwin.recv
+#else
+    let recv: (CInt, UnsafeMutableRawPointer?, size_t, CInt) -> ssize_t = Glibc.recv
+#endif
+    return try syscall(blocking: true) {
+      read(s, buf, len, 0)
```
karwa, line 494: Looks like a typo

compnerd: No, this is not a typo - the `read` call was used previously - but the new wrapping here uses the original BSD API which called it `recv` since it was a sockets-only API rather than a file descriptor based API.

compnerd: Yeah, Im planning on changing the way that this is done, and that should help with that.

compnerd:
This should be more approachable for others now with swift-package-manager support on Windows.  Seems that there are is a fair amount of regression that has occurred in the mean time.  I'll try to throw up some patches to improve the current state on Windows, but after bumping to swift 5.3 for s-p-m and a couple of small tweaks, it is just a `swift build -c release` to get started to repair the regressions.

Dev1an:
Any updates on NIO for windows?

compnerd:
I don't think that this PR has anything valuable left at this point. There has been a fair amount of change to the project in the interleaving time.

@Dev1an: while many people seem interested in using NIO on Windows, no one else is willing to work on the underlying primitives.  My personal interests lean more towards compilers and programming language design rather than networking, so this makes this work low priority.  All the work that I've done for this I've continuously made available, but no one seems to be motivated to even split up the work and get it merged.

There are at least two different types of issues to consider here:

1. Get this to build with the Windows APIs and semantic differences.  This is fairly mechanical and should just a matter of working through the build.

2. Get NIO to be usable, which is a fair amount more complicated.  The underlying problem here is that while getting NIO to build would be a significant accomplishment, it would not necessarily leave NIO in a state where one could easily use it because there are places where we would be source incompatible.  This means that any user of NIO would need to have two paths for calls to NIO to make sure that the arguments are correct.  While likely that it could be done piecemeal as one starts building up usage, there are obvious places where this is a problem which I've brought up in the past (e.g. #1673), but again, there has been very little engagement from the community to address that concern.

Issue #1226, opened 2019-11-08
XOR the pointer value in the promise creation store to hide it from Memory Graph Debugger


weissi:
Memory Graph Debugger and probably other memory debugging tools (like LSan/valgrind/...) will scan the heap for pointer values. Currently, the promise creation store contains `ObjectIdentifier(promise)` which means we will have the pointer value of `promise` in the heap without actually retaining that.
This leads to Mem Graph Debugger in debug mode showing reference cycles that aren't actually there.

The fix is easy: We should just make this an `UInt(bitPattern: ObjectIdentifier(promise))` and XOR that with a known value that fits into both 32 and 64 bit address spaces, for example `0xdeadbef`. That way, Mem Graph Debugger and friends won't see this as a reference and the tools will work better on NIO :).

shekhar-rajak:
Hi,

I want to work on this. But I didn't understand this line:

> Currently, the promise creation store contains ObjectIdentifier(promise) which means we will have the pointer value of promise in the heap without actually retaining that.

Can you please explain a bit, in which file this happens?

Lukasa:
Yup, it's a property of the [`SelectableEventLoop`](https://github.com/apple/swift-nio/blob/16ab4d657e1ad4e77bd5f8b94af8538561643053/Sources/NIO/SelectableEventLoop.swift#L89-L105).

shekhar-rajak:
Thanks @Lukasa , I am able to find out the method [`promise creation store`](https://github.com/apple/swift-nio/blob/master/Sources/NIO/SelectableEventLoop.swift#L93).

Should we use 

```
private var _promiseCreationStore: [UInt: (file: StaticString, line: UInt)] = [:]
```

instead

```
private var _promiseCreationStore: [ObjectIdentifier: (file: StaticString, line: UInt)] = [:]
```
?

So code like ` self._promiseCreationStore[ObjectIdentifier(future)] = (file: file, line: line)` should be changed to ` self._promiseCreationStore[UInt(bitPattern: ObjectIdentifier(future))] = (file: file, line: line)`?

weissi:
@shekhar-rajak sounds good!

Lukasa:
I suspect that's what we'll have to do, yes.

shekhar-rajak:
Thanks! I will work on this.

PR #1411, opened 2020-02-24
Use UInt instead of ObjectIdentifier in promiseCreationStore

shekhar-rajak:
### Motivation:

Use `UInt` data type instead of `ObjectIdentifier` in `_promiseCreationStore` variable.

### Modifications:

Using 

```
private var _promiseCreationStore: [UInt: (file: StaticString, line: UInt)] = [:]
```

instead

```
private var _promiseCreationStore: [ObjectIdentifier: (file: StaticString, line: UInt)] = [:]
```

### Result:

Fixes: https://github.com/apple/swift-nio/issues/1226

### DIFF
--- a/Sources/NIO/SelectableEventLoop.swift
+++ b/Sources/NIO/SelectableEventLoop.swift
@@ -87,20 +87,20 @@ internal final class SelectableEventLoop: EventLoop {
     /// Creates a new `SelectableEventLoop` instance that is tied to the given `pthread_t`.
 
     private let promiseCreationStoreLock = Lock()
-    private var _promiseCreationStore: [ObjectIdentifier: (file: StaticString, line: UInt)] = [:]
+    private var _promiseCreationStore: [UInt: (file: StaticString, line: UInt)] = [:]
 
     @usableFromInline
     internal func promiseCreationStoreAdd<T>(future: EventLoopFuture<T>, file: StaticString, line: UInt) {
         precondition(_isDebugAssertConfiguration())
         self.promiseCreationStoreLock.withLock {
-            self._promiseCreationStore[ObjectIdentifier(future)] = (file: file, line: line)
+            self._promiseCreationStore[UInt(bitPattern: ObjectIdentifier(future))] = (file: file, line: line)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/NIO/SelectableEventLoop.swift
```diff
@@ -87,20 +87,20 @@ internal final class SelectableEventLoop: EventLoop {
...
     private let promiseCreationStoreLock = Lock()
-    private var _promiseCreationStore: [ObjectIdentifier: (file: StaticString, line: UInt)] = [:]
+    private var _promiseCreationStore: [UInt: (file: StaticString, line: UInt)] = [:]
 
     @usableFromInline
     internal func promiseCreationStoreAdd<T>(future: EventLoopFuture<T>, file: StaticString, line: UInt) {
         precondition(_isDebugAssertConfiguration())
         self.promiseCreationStoreLock.withLock {
-            self._promiseCreationStore[ObjectIdentifier(future)] = (file: file, line: line)
+            self._promiseCreationStore[UInt(bitPattern: ObjectIdentifier(future))] = (file: file, line: line)
```
weissi: I think we should have a

```swift
private func obfuscatePointerValue<T>(_ future: EventLoopFuture<T>) -> UInt {
    return UInt(bitPattern: ObjectIdentifier(future)) ^ 0xbf15ca5d
}
```

to convert the pointers to `UInt`s. Also, we should XOR it with a value (I don't mind which value we use, the above proposes `0xbf15ca5d` which reads a little bit like _obfuscated_) so that it's hidden from Memory Graph debugger which always (incorrectly) claims that there's a reference cycle :)

shekhar-rajak:
Thanks @Lukasa @weissi for the review. I have done the changes but I am still trying to understand how XOR with a 32 bit value `0xbf15ca5d` will ensure that it will hide the reference cycle ?

Lukasa:
We used to have an object on the heap that looked kinda pointer-like, because it was the integer representation of a pointer. However, it's not _actually_ a pointer: we can't dereference it, and it doesn't participate in retain cycles. However, memory graph debugger doesn't know that.

Now this object, while it still looks like a pointer, now points somewhere totally elsewhere in memory. It's highly unlikely that this will appear to be a Future anymore, so we don't look anywhere near as likely to be participating in a retain cycle. That's good, as we aren't!

PR #1413, opened 2020-02-25
Stop copying Strings into Arrays for SHA1

### LINKED ISSUE #1409: Stop copying Strings into Arrays for SHA1
This code is a very Swift 4.2 block:

https://github.com/apple/swift-nio/blob/2c555d144a3699b0f267d71092c01e60491b9615/Sources/NIOWebSocket/SHA1.swift#L48-L53

We can use `String.utf8.withContiguousStorageIfAvailable` as a fast-path here, and only fall back to the Array case if that fails. In most cases this will save us an allocation for each hash function update.

trungducc:
### Motivation:

Resolve #1409

### Modifications:

Using `withContiguousStorageIfAvailable` if `string.utf8` supports an internal representation in a form. Otherwise, fall back to use `withUnsafeBufferPointer`.

### Result:

Save an allocation for each hash function update

### DIFF
--- a/Sources/NIOWebSocket/SHA1.swift
+++ b/Sources/NIOWebSocket/SHA1.swift
@@ -46,6 +46,12 @@ internal struct SHA1 {
     ///     - string: The string that will be UTF-8 encoded and fed into the
     ///         hash context.
     mutating func update(string: String) {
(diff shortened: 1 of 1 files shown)


Lukasa:

Lukasa:

trungducc:
Great! Thanks for your time @Lukasa !

PR #1417, opened 2020-02-26
Allow discardable results on NIOAtomic for mutating functions

kylebrowning:
### Motivation:
Make the NIOAtomic API look a little cleaner
Currently when using add or sub, we have code the tlooks like this,
```swift
_ = self.activeConns.add(1)
channel.closeFuture.whenComplete { _ in
    _ = self.activeConns.sub(1)
}
```
### Modifications:
Adding @disrcardableResult to the to following function signatures.
- `public func add(_ rhs: T) -> T {`
- `public func sub(_ rhs: T) -> T {`

### Result:
If the API consumer wishes to discard results, the above looks much cleaner.
```swift
self.activeConns.add(1)
channel.closeFuture.whenComplete { _ in
    self.activeConns.sub(1)
}
```

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOAtomic.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOAtomic.swift
@@ -223,6 +223,7 @@ public final class NIOAtomic<T: NIOAtomicPrimitive> {
     /// - Returns: `True` if the exchange occurred, or `False` if `expected` did not
     ///     match the current value and so no exchange occurred.
     @inlinable
+    @discardableResult
     public func compareAndExchange(expected: T, desired: T) -> Bool {
         return Manager(unsafeBufferObject: self).withUnsafeMutablePointerToElements {
             return T.nio_atomic_compare_and_exchange($0, expected, desired)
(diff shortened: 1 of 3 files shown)


Sources/NIOConcurrencyHelpers/NIOAtomic.swift
```diff
@@ -223,6 +223,7 @@ public final class NIOAtomic<T: NIOAtomicPrimitive> {
     /// - Returns: `True` if the exchange occurred, or `False` if `expected` did not
     ///     match the current value and so no exchange occurred.
     @inlinable
+    @discardableResult
```
weissi: this shouldn't be discardable. If people don't care about whether this succeeds, they should use `store`

Sources/NIOConcurrencyHelpers/NIOAtomic.swift
```diff
@@ -268,6 +271,7 @@ public final class NIOAtomic<T: NIOAtomicPrimitive> {
     /// - Parameter value: The new value to set this object to.
     /// - Returns: The value previously held by this object.
     @inlinable
+    @discardableResult
```
weissi: this also shouldn't be discardable because people should use `store` if they don't care about the result.

weissi:

PR #1418, opened 2020-02-27
Add Collection customisation points to ByteBufferView

### LINKED ISSUE #1385: ByteBufferView: implement most Collection/Sequence customisation points
`ByteBufferView` is a `Collection` of `UInt8` which "views" into a `ByteBuffer`. Right now, we basically only implement the `subscript` which vends individual bytes. That's good enough to implement `Collection`/`Sequence` (and a bunch of more specialised sub-protocols) but it's not very fast at all.

Especially methods like `firstIndex(of:)` or `firstIndex(where:)` will be very slow because it'll repeatedly call the subscript for every byte. It would be _much_ faster to just forward the `Collection`/`Sequence` customisation points directly to the underlying bytes.

Generally, this will look something like

```
public func _customTHEFUNCTION(ARGUMENTS) {
        return try self._buffer.withVeryUnsafeBytes { ptr in
            let ptr = UnsafeRawBufferPointer(start: ptr.baseAddress!.advanced(by: self._range.lowerBound), count: self._range.count))
            return ptr.THEFUNCTION(ARGUMENTS)
        }
}
```

so essentially we just forward all operations into the `UnsafeRawBufferPointer` to the bytes this `ByteBufferView` contains.

The `Collection` customisation points we should implement are

- [x] [`_customIndexOfEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L523) using `ptr.firstIndex(of:)` [done in #1394]
- [x] [`_customLastIndexOfEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L535) using `ptr.lastIndex(of:)` [done in #1394]
- [x] all three variants of [`_failEarlyRangeCheck`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L646) by just checking the `self._range` [done in #2226]
- [x] [`_customContainsEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L352) using `ptr.contains()`
- [x] [`_copyToContiguousArray`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L358)
- [x] [`_copyContents`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L362)

We also don't need to implement all of those in one go, this can be done in multiple PRs.

Some notes:
- we should back the gains up by benchmarks
- this requires basically no SwiftNIO knowledge but intermediate Swift knowledge about how protocols/Collection/Sequence

trungducc:
### Motivation:

Resolve #1385

### Modifications:

Added methods
- all three variants of `_failEarlyRangeCheck`
- `_customContainsEquatableElement`
- `_copyToContiguousArray`
- `_copyContents`
- `testBufferViewContains`

### Result:

Better performance

### DIFF
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
 
     public subscript(position: Index) -> UInt8 {
         get {
-            guard position >= self._range.lowerBound && position < self._range.upperBound else {
-                preconditionFailure("index \(position) out of range")
-            }
+            _failEarlyRangeCheck(position, bounds: self._range)
             return self._buffer.getInteger(at: position)! // range check above
         }
         set {
-            guard position >= self._range.lowerBound && position < self._range.upperBound else {
-                preconditionFailure("index \(position) out of range")
-            }
+            _failEarlyRangeCheck(position, bounds: self._range)
             self._buffer.setInteger(newValue, at: position)
         }
     }
 
     public subscript(range: Range<Index>) -> ByteBufferView {
         get {
+            _failEarlyRangeCheck(range, bounds: self._range)
             return ByteBufferView(buffer: self._buffer, range: range)
         }
         set {
+            _failEarlyRangeCheck(range, bounds: self._range)
             self.replaceSubrange(range, with: newValue)
         }
     }
@@ -96,6 +94,52 @@ public struct ByteBufferView: RandomAccessCollection {
             return ptr.lastIndex(of: element).map { $0 + self._range.lowerBound }
         })
     }
+
+    public func _customContainsEquatableElement(_ element: Element) -> Bool? {
+        return .some(self.withUnsafeBytes { bytes -> Bool in
+            return bytes.contains(element)
+        })
+    }
+
+    public func _copyToContiguousArray() -> ContiguousArray<Element> {
+        return self.withUnsafeBytes { bytes in
+            return bytes._copyToContiguousArray()
+        }
+    }
+
+    public func _copyContents(initializing buffer: UnsafeMutableBufferPointer<UInt8>) -> (Iterator, UnsafeMutableBufferPointer<Element>.Index) {
+        return self.withUnsafeBytes { bytes in
+            let (_, idx) = bytes._copyContents(initializing: buffer)
+            return (IndexingIterator(_elements: self, _position: idx), idx)
+        }
+    }
+
+    public func _failEarlyRangeCheck(_ index: Index, bounds: Range<Index>) {
+        precondition(
+            self.startIndex <= bounds.lowerBound && bounds.upperBound <= self.endIndex,
+            "Bounds out of buffer's range")
+        precondition(
+            bounds.lowerBound <= index && index < bounds.upperBound,
+            "Index \(index) out of bounds")
+    }
+
+    public func _failEarlyRangeCheck(_ index: Index, bounds: ClosedRange<Index>) {
+        precondition(
+            self.startIndex <= bounds.lowerBound && bounds.upperBound <= self.endIndex,
+            "Bounds out of buffer's range")
+        precondition(
+            bounds.lowerBound <= index && index <= bounds.upperBound,
+            "Index \(index) out of bounds")
+    }
+
+    public func _failEarlyRangeCheck(_ range: Range<Index>, bounds: Range<Index>) {
+        precondition(
+            self.startIndex <= bounds.lowerBound && bounds.upperBound <= self.endIndex,
+            "Bounds out of buffer's range")
+        precondition(
+            bounds.lowerBound <= range.lowerBound && range.upperBound <= bounds.upperBound,
+            "Range out of bounds")
+    }
 }
 
 extension ByteBufferView: MutableCollection {}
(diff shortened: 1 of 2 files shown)


Sources/NIO/ByteBuffer-views.swift
```diff
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
 
     public subscript(position: Index) -> UInt8 {
         get {
-            guard position >= self._range.lowerBound && position < self._range.upperBound else {
-                preconditionFailure("index \(position) out of range")
-            }
+            _failEarlyRangeCheck(position, bounds: self._range)
```
Lukasa: NIO style is to use explicit self everywhere.

```suggestion
            self._failEarlyRangeCheck(position, bounds: self._range)
```

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
...
             return self._buffer.getInteger(at: position)! // range check above
         }
         set {
-            guard position >= self._range.lowerBound && position < self._range.upperBound else {
-                preconditionFailure("index \(position) out of range")
-            }
+            _failEarlyRangeCheck(position, bounds: self._range)
```
Lukasa:
```suggestion
            self._failEarlyRangeCheck(position, bounds: self._range)
```

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
...
             self._buffer.setInteger(newValue, at: position)
         }
     }
 
     public subscript(range: Range<Index>) -> ByteBufferView {
         get {
+            _failEarlyRangeCheck(range, bounds: self._range)
```
Lukasa:
```suggestion
            self._failEarlyRangeCheck(range, bounds: self._range)
```

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
...
             return ByteBufferView(buffer: self._buffer, range: range)
         }
         set {
+            _failEarlyRangeCheck(range, bounds: self._range)
```
Lukasa:
```suggestion
            self._failEarlyRangeCheck(range, bounds: self._range)
```

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -96,6 +94,52 @@ public struct ByteBufferView: RandomAccessCollection {
...
+        return self.withUnsafeBytes { bytes in
+            let (_, idx) = bytes._copyContents(initializing: buffer)
+            return (IndexingIterator(_elements: self, _position: idx), idx)
+        }
+    }
+
+    public func _failEarlyRangeCheck(_ index: Index, bounds: Range<Index>) {
+        precondition(
+            self.startIndex <= bounds.lowerBound && bounds.upperBound <= self.endIndex,
+            "Bounds out of buffer's range")
```
Lukasa: I don't believe this precondition is necessary: I think Swift guarantees that we shouldn't need it. @lorentey, do we?

Sources/NIO/ByteBuffer-views.swift
trungducc: I got it. Thanks for pointing out.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -56,24 +56,22 @@ public struct ByteBufferView: RandomAccessCollection {
 
     public subscript(position: Index) -> UInt8 {
         get {
-            guard position >= self._range.lowerBound && position < self._range.upperBound else {
-                preconditionFailure("index \(position) out of range")
-            }
+            self._failEarlyRangeCheck(position, bounds: self._range)
```
weissi: I don't think you're supposed to _call_ those underscored methods? I think you're just meant to implement them. Am I wrong here?

weissi: CC @lorenty

trungducc: @weissi Can't we use underscored methods? I read collection classes in stblib and they often use `_failEarlyRangeCheck`  in `subscript` method. I thought it's ok so I replaced guard statement with `_failEarlyRangeCheck` to remove duplicate code.

weissi: @trungducc I think the underscored methods are meant to be internal to the standard library and are not supposed to be called by users outside the standard library, @lorentey to confirm this

weissi:
friendly ping @trungducc , are you still interested in pushing this PR over the line?

Sources/NIO/ByteBuffer-views.swift
atrick: The underscored functions are usually for some functionality the stdlib needs but no one is sure yet how to best expose it to the language, or whether we want users doing it at all. So they haven't gone through Swift evolution process, and aren't part of the API. They need to be in the ABI so that inlined stdlib functions can still cal them. It's too bad we at least don't have a warning for calling them outside the stdlib. Maybe @lorentey or @airspeedswift  can explain more. If it's something you think non-stdlib code needs and isn't easily reimplemented outside the stdlib, then someone should create an evolution proposal.

Sources/NIO/ByteBuffer-views.swift
trungducc: @weissi I added this precondition as you suggested. But if it's not necessary, I think we don't need to implement `_failEarlyRangeCheck` methods because `index` and `range` parameter are already checked for lying within `bounds` in `Collection`. How do you think?

Lukasa:
Hey @trungducc, given the answers above are you happy to rework this patch to avoid using the underscored methods?

PR #1420, opened 2020-02-28
list_topsorted_deps: multiple fixes

weissi:
Motivation:

list_topsorted_dependencies had multiple issues:
- it would list some executable targets (even in libs-only mode)
- it confused packages with modules
- it would also list dependencies from outside the current package

Modifications:

Address all of the above.

Result:

Script more useful, inside and outside of swift-nio.


weissi:
CC @Jake-Prickett  , I think this should fix all your issues. For the record, this is the output in swift-nio-ssl:

```
$ scripts/list_topsorted_dependencies.sh 
NIOSSLPerformanceTester
NIOSSLHTTP1Client
NIOTLSServer
NIOSSL
CNIOBoringSSLShims
CNIOBoringSSL
$ scripts/list_topsorted_dependencies.sh  -l
NIOSSL
CNIOBoringSSLShims
CNIOBoringSSL
```

Jake-Prickett:
👏 This can also be added into Swift NIO HTTP2! Thanks @weissi!

PR #1424, opened 2020-03-02
Thread: port to Windows

compnerd:
This restructures `Thread` to have a more generic representation of a
thread handle.  Using that it is possible to have an implementation of
the threading APIs for Windows.

Similarly, introduce an abstraction over the Thread Specific Key.  This
allows us to abstract away the key type and the usage.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/Thread.swift
+++ b/Sources/NIO/Thread.swift
@@ -13,15 +13,15 @@
 //===----------------------------------------------------------------------===//
 
 import CNIOLinux
-
-private typealias ThreadBoxValue = (body: (NIOThread) -> Void, name: String?)
-private typealias ThreadBox = Box<ThreadBoxValue>
+#if os(Windows)
+import WinSDK
+#endif
 
 
 #if os(Linux)
 private let sys_pthread_getname_np = CNIOLinux_pthread_getname_np
 private let sys_pthread_setname_np = CNIOLinux_pthread_setname_np
-#else
+#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
 private let sys_pthread_getname_np = pthread_getname_np
 // Emulate the same method signature as pthread_setname_np on Linux.
 private func sys_pthread_setname_np(_ p: pthread_t, _ pointer: UnsafePointer<Int8>) -> Int32 {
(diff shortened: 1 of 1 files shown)


Sources/NIO/Thread.swift
```diff
@@ -57,31 +80,46 @@ final class NIOThread {
...
-
-                return String(decoding: UnsafeRawBufferPointer(UnsafeBufferPointer<CChar>(rebasing: ptr.prefix { char in
-                                  char != 0
-                              })),
-                              as: Unicode.UTF8.self)
+#if os(Windows)
+        var pszBuffer: PWSTR?
+        GetThreadDescription(self.handle, &pszBuffer)
+        guard let buffer = pszBuffer else { return nil }
+        let string: String = String(decodingCString: buffer, as: UTF16.self)
```
weissi: don't we need `Unicode.UTF16.self` here? I thought with Swift 5 you always need the `Unicode.` prefix?

Sources/NIO/Thread.swift
```diff
@@ -90,51 +128,86 @@ final class NIOThread {
...
-            // Cast to UnsafeMutableRawPointer? and force unwrap to make the same code work on macOS and Linux.
-            let b = Unmanaged<ThreadBox>.fromOpaque((p as UnsafeMutableRawPointer?)!).takeRetainedValue()
 
-            let body = b.value.body
-            let name = b.value.name
+        let argv0 = Unmanaged.passRetained(box).toOpaque()
 
-            let pt = pthread_self()
+#if os(Windows)
+        // FIXME(compnerd) this should use the `stdcall` calling convention
```
weissi: FIXME left

weissi: is this for this PR or a followup?

Sources/NIO/Thread.swift
```diff
@@ -176,21 +249,66 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
...
+      public typealias KeyDestructor =
+          @convention(c) (UnsafeMutableRawPointer) -> Void
+#else
+      public typealias KeyDestructor =
+          @convention(c) (UnsafeMutableRawPointer?) -> Void
+#endif
+
+      public init(destructor: KeyDestructor?) {
+#if os(Windows)
+        self.value = FlsAlloc(destructor)
```
weissi: nit: 4 space indentation

Sources/NIO/Thread.swift
```diff
@@ -176,21 +249,66 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
...
+        precondition(dwResult, "FlsFree: \(GetLastError())")
+#else
+        let result = pthread_key_delete(self.value)
+        precondition(result == 0, "pthread_key_delete failed: \(result)")
+#endif
+      }
+
+      public static func get(for key: Key) -> UnsafeMutableRawPointer? {
+#if os(Windows)
+        return FlsGetValue(key.value)
```
weissi: 4 spaces

Sources/NIO/Thread.swift
```diff
@@ -176,21 +249,66 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
     /* the actual type in there is `Box<(ThreadSpecificVariable<T>, T)>` but we can't use that as C functions can't capture (even types) */
     private typealias BoxedType = Box<(AnyObject, AnyObject)>
 
-    private let key: pthread_key_t
+    private class Key {
```
weissi: Are we making use of `Key` being reference counted? Aren't we just wrapping a handle which is destroyed explicitly and not though a `deinit`? I think we should make this a `struct Key`

Sources/NIO/Thread.swift
compnerd: Hmm, I've never needed the `Unicode` prefix.  I'm happy to add it if thats the style that the project is using, but this builds fine (and a similar pattern is used extensively in Foundation).

Sources/NIO/Thread.swift
compnerd: It will have to be a follow up.  This requires some new work in the compiler - `stdcall` is not a supported calling convention right now.  This largely is a pedantic concern atm - Windows x86 is broken for other reasons right now.

Sources/NIO/Thread.swift
compnerd: Ugh, you're right.  I had overlooked the manual key destruction when I first wrote that, and only realized that later (since so much code was blindly removed) and forgot to update this.

Sources/NIO/Thread.swift
weissi: ha, maybe they used to issue a warning but saw too many breakages and removed it again :) . happy if it doesn't warn today .

Sources/NIO/Thread.swift
compnerd: Oh, wait, I think Im mixing up another site.  Here, the original code was adding an explicit `deinit` that deleted the key.  I can certainly make the wrapper a struct and add a static method to just dispatch the delete.

weissi: sorry, I think this is fine. If we do use a `deinit`, then `class` is great. Thought the code wasn't using `deinit` nor any of the reference properties and it was still a `class` but I think I misread.

compnerd: Yes, the `deinit` destroys the slot.  Its a tradeoff ... now the slot itself is managed, so now `ThreadSpecificVariable` could be a `struct`, but its `public` so I didn't want to touch it.  IMO this is nicer because it reflects the reality - the value itself is not what was refcounted, it was the slot.

weissi:

PR #1425, opened 2020-03-02
DNM: delete NIO

compnerd:
Largely delete NIO - this is meant to allow others to help with porting to Windows, it identifies the locations that need to be adjusted in order to enable Windows support


compnerd:
Oh, the other reason to have this - I lost this three times!  its a pain to identify the various sites :-( - Is there some way to ensure that this doesn't get merged accidentally?

Lukasa:
Mark it as a draft PR. :grin:

compnerd:
Too late to do that, but this is now far less useful given how much of this is no longer necessary.

Issue #1416, opened 2020-02-26
ByteToMessageDecoder: Default implementation for decodeLast should be provided


weissi:
For all but EOF framed protocols, the right implementation for `decodeLast` is

            public func decodeLast(context: ChannelHandlerContext,
                            buffer: inout ByteBuffer,
                            seenEOF: Bool) throws -> DecodingState {
                while try self.decode(context: context, buffer: &buffer) == .continue {}
                return .needMoreData
            }

And yet, we require users to always provide a `decodeLast` implementation. Instead, we should just provide the one above.

shekhar-rajak:
So we have to change the current `decodeLast` method in  `NIOChatServer/main.swift:`, `NIOHTTP1/HTTPDecoder.swift`,  `NIOTLS/SNIHandler.swift`, `NIOWebSocket/WebSocketFrameDecoder.swift` ?

weissi:
@shekhar-rajak Thanks, great questions:

-  `HTTPDecoder` definitely needs its own implementation that's different to the one suggested above because HTTP/1 can be EOF framed sadly.
- `SNIHandler` is also special because (see comment on `decode`), its decode never returns `.continue`
- `NIOChatServer/main.swift`'s decodeLast should be removed indeed
- `NIOWebSocket`'s `decodeLast` should also be removed and replaced with the above, @Lukasa to confirm this.

Lukasa:
Confirm for `WebSocketFrameDecoder`.

PR #1426, opened 2020-03-02
ByteToMessageDecoder: Default implementation for decodeLast should be provided 

shekhar-rajak:
### Motivation:

Corrected the `decodeLast` function  in EOF framed protocols.

### Modifications:

- Removed the `decodeLast` function from `NIOChatServer/main.swift`
- Updated the `decodeLast` function  in `NIOWebSocket/WebSocketFrameDecoder.swift`

### Result:

Fixes: https://github.com/apple/swift-nio/issues/1416

### DIFF
--- a/Sources/NIOWebSocket/WebSocketFrameDecoder.swift
+++ b/Sources/NIOWebSocket/WebSocketFrameDecoder.swift
@@ -270,6 +270,7 @@ public final class WebSocketFrameDecoder: ByteToMessageDecoder {
     }
 
     public func decodeLast(context: ChannelHandlerContext, buffer: inout ByteBuffer, seenEOF: Bool) throws -> DecodingState {
+        while try self.decode(context: context, buffer: &buffer) == .continue {}
         // EOF is not semantic in WebSocket, so ignore this.
         return .needMoreData
     }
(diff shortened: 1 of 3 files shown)


Sources/NIOWebSocket/WebSocketFrameDecoder.swift
```diff
@@ -270,6 +270,7 @@ public final class WebSocketFrameDecoder: ByteToMessageDecoder {
     }
 
     public func decodeLast(context: ChannelHandlerContext, buffer: inout ByteBuffer, seenEOF: Bool) throws -> DecodingState {
```
shekhar-rajak: I think, we can remove this method definition which is same as the default implementation. Am I right ?

shekhar-rajak:
Should we also update [`NIOSingleStepByteToMessageDecoder.decodeLast`](https://github.com/apple/swift-nio/blob/master/Sources/NIO/SingleStepByteToMessageDecoder.swift#L66) function ?

Sources/NIOWebSocket/WebSocketFrameDecoder.swift
weissi: Yes, agreed

shekhar-rajak:
> Let’s remove the websocket frame decoder decodeLast.

Thanks for the review, I have done the changes.

PR #1427, opened 2020-03-02
Enable manual unwrapping of `NIOAny`

Davidde94:
Expose public methods to unwrap `NIOAny` safely (or unsafely if you're brave).

### Motivation:
It's difficult to work with `NIOAny` outside of a `Channel`. These methods should make handling a little easier.

### Modifications:
Add public methods to either force-unwrap or safely-unwrap some `NIOAny` value.
### Result:
It's now possible to unwrap `NIOAny` as and when you please.


Sources/NIO/NIOAny.swift
```diff
@@ -255,6 +255,22 @@ public struct NIOAny {
             return o
         }
     }
+    
+    /// Force-unwraps the wrapped message as a given type `T`
+    ///
+    /// returns: The wrapped `T`, or crashes if the wrapped message is not a `T`
+    @inlinable
+    public func unwrap<T>(as type: T.Type = T.self) -> T {
```
glbrntt: If we're okay adding public API for unwrapping I think it would be better for it to return `T?` (in which case it might just be better to `tryAs` public). I think this will help to discourage people from using it over `unwrap{In,Out}boundIn` in handlers. 

If we do have a force unwrap API as well then I think it should be obvious from the name that it can blow up.

Davidde94: You make a good point. If we go down that route we can even just mark the existing functions as `public` and the job is done. Hoping to spin up a discussion to see if the general consensus is for/against this.

Davidde94: I misread what you first said, I didn't mean to just repeat what you said back at you haha!

Davidde94:
@weissi IMO I don't think using `Any` would be the right/best approach, but should I close this down for now?

weissi:
@Davidde94 if you’re in a handler that really needs to handle anything (which isn’t recommended), then I think using Any is the right approach. Of course if you can type it, you should point out your type.

Your patch however enables unwrapping literally anything from a NIOAny. And if you need to do that; then you should be upfront and say `Any` I think. Imagine we merged your patch, someone could now say `InboundIn = String` and then unwrap an `Int`. If anyone wants to do this (again very very not recommend), then should say upfront that they accept `Any` and not String

Davidde94:
I think I found myself in a rather strange/unusual use case where I wasn't inside a handler, and yet was passed a `NIOAny`. I completely agree that if you're inside a handler then this would be redundant. Anyway, I'll close this and we can continue discussion on the other PR?

weissi:
@Davidde94 NIOAnys should never leave a handler. If they do, that’s a bug in the handler.

PR #1430, opened 2020-03-03
tests: improve tests that are supposed to throw

weissi:
Motivation:

The

    do {
        try someOperation()
	XCTFail("should throw") // easy to forget
    } catch error as SomethingError {
        XCTAssertEqual(.something, error as? SomethingError)
    } catch {
        XCTFail("wrong error")
    }

pattern is not only very long, it's also very error prone. If you forget
any of the XCTFails, you might not tests what it looks like

    XCTAssertThrowsError(try someOperation) { error in
        XCTAssertEqual(.something, error as? SomethingError)
    }

is much safer and shorter.

Modifcations:

Do many of the above replaces.

Result:

Cleaner, shorter, and safer tests.

### DIFF
--- a/Tests/NIOTests/HappyEyeballsTest.swift
+++ b/Tests/NIOTests/HappyEyeballsTest.swift
@@ -32,8 +32,9 @@ private extension Array where Element == Channel {
         self.forEach {
             do {
                 _ = try($0 as! EmbeddedChannel).finish()
+                // We're happy with no error
             } catch ChannelError.alreadyClosed {
-                return
+                return // as well as already closed.
             } catch {
                 XCTFail("Finishing got error \(error)")
             }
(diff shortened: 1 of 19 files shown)


Tests/NIOTests/HappyEyeballsTest.swift
```diff
@@ -32,8 +32,9 @@ private extension Array where Element == Channel {
         self.forEach {
             do {
                 _ = try($0 as! EmbeddedChannel).finish()
+                // We're happy with no error
             } catch ChannelError.alreadyClosed {
-                return
+                return // as well as already closed.
```
glbrntt: Is there any value in something like `assertMaybeThrows` which validates the error if one is thrown but doesn't fail if no error is thrown? I think I've only seen one or two occurrences of this elsewhere here so probably not...

weissi: There are a few but unfortunately our `TestUtils.swift` doesn't really scale. It's always just available in one test module but we have lots of modules and like 5 different repos :|. So I now prefer to stick with plain XCTest whenever possible...

PR #1431, opened 2020-03-04
Thread: repair macOS build

compnerd:
The port to Windows built on Linux and Windows but regressed the macOS
build.  Repair that by handling the optionality of the thread handle
through a cast to optional and then unwrapping the result.

### Motivation:

### Modifications:

### Result:


weissi:
sorry @compnerd , only seeing this now. I'll close this in favour of #1433

PR #1433, opened 2020-03-04
fix macOS build & cleanup Thread

weissi:
Motivation:

The macOS build was broken by #1424, also the Thread.swift file became
a bit of a mess with `#if os(...)/#else` everywhere.

Modifications:

- fix macOS build
- cleanup Thread by splitting it into ThreadPosix & ThreadWindows

Result:

- build working again
- cleaner code

### DIFF
--- a/Sources/NIO/Thread.swift
+++ b/Sources/NIO/Thread.swift
@@ -13,62 +13,45 @@
 //===----------------------------------------------------------------------===//
 
 import CNIOLinux
-#if os(Windows)
-import WinSDK
-#endif
 
-
-#if os(Linux)
-private let sys_pthread_getname_np = CNIOLinux_pthread_getname_np
-private let sys_pthread_setname_np = CNIOLinux_pthread_setname_np
-#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
-private let sys_pthread_getname_np = pthread_getname_np
-// Emulate the same method signature as pthread_setname_np on Linux.
-private func sys_pthread_setname_np(_ p: pthread_t, _ pointer: UnsafePointer<Int8>) -> Int32 {
-    assert(pthread_equal(pthread_self(), p) != 0)
-    pthread_setname_np(pointer)
-    // Will never fail on macOS so just return 0 which will be used on linux to signal it not failed.
-    return 0
+enum LowLevelThreadOperations {
+    
 }
-#endif
 
-fileprivate extension NIOThread.NIOThreadHandle {
-  static var invalid: NIOThread.NIOThreadHandle {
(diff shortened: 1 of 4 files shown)


Sources/NIO/Thread.swift
```diff
@@ -249,56 +144,23 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
...
-
-#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
-        public typealias KeyDestructor =
-            @convention(c) (UnsafeMutableRawPointer) -> Void
-#else
-        public typealias KeyDestructor =
-            @convention(c) (UnsafeMutableRawPointer?) -> Void
-#endif
+    internal class Key {
+        private var value: ThreadOpsSystem.ThreadSpecificKey
```
glbrntt: Can we name this `key` instead?

It's misleading in function calls: `getThreadSpecificValue(self.value)`, `setThreadSpecificValue(key: self.value, value: value)`

Sources/NIO/Thread.swift
```diff
@@ -80,46 +63,17 @@ final class NIOThread {
     /// - parameters:
     ///     - body: The closure that will accept the `pthread_t`.
     /// - returns: The value returned by `body`.
-#if !os(Windows)
-    internal func withUnsafePthread<T>(_ body: (pthread_t) throws -> T) rethrows -> T {
+    internal func withUnsafePthread<T>(_ body: (ThreadOpsSystem.ThreadHandle) throws -> T) rethrows -> T {
```
glbrntt: nit: `withUnsafeThreadHandle`? (and update documentation)

Sources/NIO/ThreadPosix.swift
```diff
@@ -0,0 +1,150 @@
...
+// Emulate the same method signature as pthread_setname_np on Linux.
+private func sys_pthread_setname_np(_ p: pthread_t, _ pointer: UnsafePointer<Int8>) -> Int32 {
+    assert(pthread_equal(pthread_self(), p) != 0)
+    pthread_setname_np(pointer)
+    // Will never fail on macOS so just return 0 which will be used on linux to signal it not failed.
+    return 0
+}
+private typealias ThreadDestructor = @convention(c) (UnsafeMutableRawPointer) -> UnsafeMutableRawPointer?
+
+#endif
```
glbrntt: Instead of nesting in `#if !os(Windows)` can we just have a separate `#if os(Linux) || os(iOS) || os(macOS) || os(tvOS) || os(watchOS)` below here; it's closer to what we actually want, right?

weissi:

PR #1434, opened 2020-03-05
Thread: repair build on Windows after #1433

compnerd:
This repairs the build on Windows after the thread implementation was
split out.

### Motivation:

### Modifications:

### Result:


compnerd:
CC: @weissi

compnerd:
No problem at all, there are a lot of paths to test here.  My pleasure :)

PR #1438, opened 2020-03-10
Fix NIOThreadPool type name in NonBlockingFileIO.init

kmahar:
Updates `NonBlockingFileIO.init` docs to use the new type name.


Lukasa:

PR #1439, opened 2020-03-11
NIO: use `Posix` "namespace" more aggressively

compnerd:
Convert many more cases of the raw address family constant to the
definition in the System version.

### Motivation:

### Modifications:

### Result:


compnerd:
Im very confused why I didn't need the conversion :/

Lukasa:

Lukasa:
I think they're only necessary on Linux due to a mismatch in the way the constants are declared and the way they're used.

PR #1440, opened 2020-03-11
Split System modules

compnerd:

### Motivation:

### Modifications:

### Result:


Sources/NIO/SystemGlibc.swift
```diff
@@ -0,0 +1,467 @@
...
+fileprivate extension IOError where T: FixedWidthInteger {
+  var result: T {
+    switch self {
+    case .processed(let value):
+      return value
+    case .wouldBlock(_):
+      fatalError("cannot unwrap IOError")
+    }
+  }
+}
```
Lukasa: I think the indentation is off here.

Sources/NIO/SystemDarwin.swift
```diff
@@ -0,0 +1,454 @@
...
+internal extension IOResult where T: FixedWidthInteger {
+  var result: T {
+    switch self {
+    case .processed(let value):
+      return value
+    case .wouldBlock(_):
+      fatalError("cannot unwrap IOError")
+    }
+  }
+}
```
Lukasa: The indentation is off here. Also, why is this `internal` when the Linux version is `fileprivate`?

compnerd: Because of a very special reason: not building on Linux ;-).  The extension is used elsewhere, and so needs to be `internal`.

Lukasa: 😁 Hah ok then.

Lukasa:
I don't know if you missed it, but there are still a chunk of Linux test failures here.

compnerd:
Ill try to run this on a Linux environment and re-upload (with a rebase)

Lukasa:
Looks like the syscall wrapper test has to be rewritten. It looks for files with the wrong name.

Lukasa:
Specifically this line needs to be updated to bring across all the relevant `System.swift` files:

https://github.com/apple/swift-nio/blob/648a6712ff8a62f5c85274f7b8e9df43ba8c1058/IntegrationTests/tests_02_syscall_wrappers/defines.sh#L42

Sources/NIO/SystemDarwin.swift
```diff
@@ -0,0 +1,443 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: can we make it `2017-2020`?

Sources/NIO/SystemDarwin.swift
```diff
@@ -0,0 +1,443 @@
...
+
+        if case .processed(let fd) = result {
+            return fd
+        } else {
+            return nil
+        }
+    }
+
+    @inline(never)
+    public static func connect(descriptor: CInt, addr: UnsafePointer<sockaddr>, size: socklen_t) throws -> Bool {
```
weissi: why are all these now duplicated? Couldn't we do a `SystemPosix.swift` instead of having so much duplication between `SystemLinux/Darwin`?

compnerd: These aren't really POSIX per se (some of them are BSD - the sockets one), and some of them are C standard (ISO), and some are POSIX (ISO).  If we want split them up across those boundaries, thats okay, but I think that this is *far* more legible and much easier to see what is going on than all the indirection.

This became even more important as I started adding a third path through this code.  It was extremely difficult to reason about the changes and follow along as I was making the changes to support MSVCRT.

weissi: but the code is identical for Linux & Mac, right? Regarding the naming: I'm happy with `SystemUNIX` or whatever but I'd like to avoid duplicating the whole file for Linux/Mac. For the _actual_ differences I think a `SystemDarwin`/`SystemLinux` does make sense but stuff like `connect` is the same for both, right?

compnerd: I think that the aliasing through sys... makes it far less readable.  I suppose we could just `#if os()` the implementations.

compnerd: Okay, so, I've opted instead to split out `System` into `System*` and `BSDSocket`.  There is duplication in `BSDSocket` still, but, collapsing those cases can be done and we can judge which is more readable.

This basically comes at a cost: you get a much larger change in a single commit.  But, the refactoring will just take a single larger step rather than the multiple intermediate ones.

compnerd:
This took too long, ended up just merging it into a single ginormous patch.

PR #1444, opened 2020-03-13
NIO: split out `FileDescriptor` and `Selectable`

compnerd:
A `FileDescriptor` is not necessarily selectable (e.g. Windows), and a
selectable is not necessarily a `FileDescriptor` (e.g. Windows).

This will allow a basic `Selector` on Windows which will use `WSASelect`
to select on sockets for some basic functionality even though
IOCompletionPorts are preferred.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -206,19 +206,12 @@ extension sockaddr_storage {
 /// Base class for sockets.
 ///
 /// This should not be created directly but one of its sub-classes should be used, like `ServerSocket` or `Socket`.
-class BaseSocket: Selectable, BaseSocketProtocol {
+class BaseSocket: BaseSocketProtocol {
     typealias SelectableType = BaseSocket
 
     private var descriptor: CInt
     public var isOpen: Bool {
         return descriptor >= 0
-    }
-
-    func withUnsafeFileDescriptor<T>(_ body: (CInt) throws -> T) throws -> T {
-        guard self.isOpen else {
-            throw IOError(errnoCode: EBADF, reason: "file descriptor already closed!")
-        }
-        return try body(descriptor)
     }
 
     /// Returns the local bound `SocketAddress` of the socket.
@@ -244,8 +237,8 @@ class BaseSocket: Selectable, BaseSocketProtocol {
         try addr.withMutableSockAddr { addressPtr, size in
             var size = socklen_t(size)
 
-            try withUnsafeFileDescriptor { fd in
-                try body(fd, addressPtr, &size)
+            try self.withUnsafeHandle {
+                try body($0, addressPtr, &size)
             }
         }
         return addr.convert()
@@ -272,7 +265,7 @@ class BaseSocket: Selectable, BaseSocketProtocol {
         #if !os(Linux)
         if setNonBlocking {
             do {
-                try BaseSocket.setNonBlocking(fileDescriptor: sock)
+                try Posix.setNonBlocking(socket: sock)
             } catch {
                 // best effort close
                 try? Posix.close(descriptor: sock)
@@ -325,8 +318,8 @@ class BaseSocket: Selectable, BaseSocketProtocol {
     ///
     /// throws: An `IOError` if the operation failed.
     final func setNonBlocking() throws {
-        return try withUnsafeFileDescriptor { fd in
-            try BaseSocket.setNonBlocking(fileDescriptor: fd)
+        return try self.withUnsafeHandle {
+            try Posix.setNonBlocking(socket: $0)
         }
     }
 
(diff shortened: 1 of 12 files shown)


Sources/NIO/BaseSocket.swift
```diff
@@ -203,24 +203,31 @@ extension sockaddr_storage {
     }
 }
 
+private func SetNonBlocking(socket: CInt) throws {
```
Lukasa: I don't love having this as a top-level function: it'd be really good to find an appropriate type to hang it off.

Sources/NIO/BaseSocket.swift
```diff
@@ -244,8 +251,8 @@ class BaseSocket: Selectable, BaseSocketProtocol {
         try addr.withMutableSockAddr { addressPtr, size in
             var size = socklen_t(size)
 
-            try withUnsafeFileDescriptor { fd in
-                try body(fd, addressPtr, &size)
+            try withUnsafeHandle {
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
```diff
@@ -325,8 +332,8 @@ class BaseSocket: Selectable, BaseSocketProtocol {
     ///
     /// throws: An `IOError` if the operation failed.
     final func setNonBlocking() throws {
-        return try withUnsafeFileDescriptor { fd in
-            try BaseSocket.setNonBlocking(fileDescriptor: fd)
+        return try withUnsafeHandle {
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
```diff
@@ -345,11 +352,11 @@ class BaseSocket: Selectable, BaseSocketProtocol {
             // most socket options settings so for the time being we'll just ignore this. Let's revisit for NIO 2.0.
             return
         }
-        return try withUnsafeFileDescriptor { fd in
+        return try withUnsafeHandle {
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
```diff
@@ -366,7 +373,7 @@ class BaseSocket: Selectable, BaseSocketProtocol {
     ///     - name: The name of the option to set.
     /// - throws: An `IOError` if the operation failed.
     final func getOption<T>(level: Int32, name: Int32) throws -> T {
-        return try withUnsafeFileDescriptor { fd in
+        return try withUnsafeHandle { fd in
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
```diff
@@ -390,7 +397,7 @@ class BaseSocket: Selectable, BaseSocketProtocol {
     ///     - address: The `SocketAddress` to which the socket should be bound.
     /// - throws: An `IOError` if the operation failed.
     final func bind(to address: SocketAddress) throws {
-        try withUnsafeFileDescriptor { fd in
+        try withUnsafeHandle { fd in
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
```diff
@@ -425,10 +432,19 @@ class BaseSocket: Selectable, BaseSocketProtocol {
     ///
     /// - throws: An `IOError` if the operation failed.
     final func takeDescriptorOwnership() throws -> CInt {
-        return try withUnsafeFileDescriptor { fd in
+        return try withUnsafeHandle {
```
weissi: missing `self.`

Sources/NIO/BaseSocket.swift
compnerd: What about placing this in `System`?

Sources/NIO/BaseSocket.swift
compnerd: Just want to point out, not my fault!  The code was pre-existing in that state.  But, sure, happy to adjust that while Im here.

weissi: oh sorry, indeed. Much appreciated if you could fix it :)

Sources/NIO/BaseSocket.swift
Lukasa: I'd be fine with that.

compnerd:
I don't understand why `NIOFailedToSetSocketNonBlockingError` is not resolved on 5.0/5.1

Lukasa:
The syscall wrapper module is constructed by hand:

https://github.com/apple/swift-nio/blob/ad0c671acfa0a50cf51f4e33c0d4305ae8c1e70b/IntegrationTests/tests_02_syscall_wrappers/defines.sh#L41-L46

The error in question is not in any of those files, so it cannot be thrown. You can just move its definition.

Sources/NIO/BaseSocket.swift
```diff
@@ -272,7 +265,7 @@ class BaseSocket: Selectable, BaseSocketProtocol {
         #if !os(Linux)
         if setNonBlocking {
             do {
-                try BaseSocket.setNonBlocking(fileDescriptor: sock)
+                try Posix.SetNonBlocking(socket: sock)
```
weissi: should be `setNonBlocking` (lower-case `s`)

PR #1447, opened 2020-03-15
Selector: further default to Linux behaviour

compnerd:
This furthers the behaviour uniformity of special casing behaviour for
other platforms (e.g. Darwin, Windows) and defaulting to the Linux
behaviour as the rest of the code base assumes that the Linux behaviour
is a reasonable fallback.  Ensuring that the paths taken are always
uniformly taken helps identify the changes necessary for a port and make
it easier to reason about the changes.

### Motivation:

### Modifications:

### Result:


compnerd:
Sure, that’s not a problem to change.  However there’s a mixture in the code, should everything be reindented?

Lukasa:
Let's only do this for the ones we're changing: no need to add noise to this diff.

compnerd:
Oh, missed the bit about the additional indentation level.

compnerd:
Okay, reformatted to put the `#if` at the code level, code itself indented an additional level for the places where I've changed things, the rest of the file still has a few different styles.

Lukasa:

compnerd:
Anything else left for this one?

Sources/NIO/Selector.swift
```diff
@@ -663,26 +662,26 @@ internal class Selector<R: Registration> {
...
-            event.filter = Int16(EVFILT_USER)
-            event.fflags = UInt32(NOTE_TRIGGER | NOTE_FFNOP)
-            event.data = 0
-            event.udata = nil
-            event.flags = 0
-            try withUnsafeMutablePointer(to: &event) { ptr in
-                try self.kqueueApplyEventChangeSet(keventBuffer: UnsafeMutableBufferPointer(start: ptr, count: 1))
-            }
-        #endif
+            #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
```
weissi: nit: indentation isn't quite right here.

PR #1448, opened 2020-03-15
NIO: implement core counting on Windows

compnerd:
Implement the core counting logic for Windows, which does not support
`sysconf`.  It uses the Windows SDK to get the processor topology and
computes the core count from the topology.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/Utilities.swift
+++ b/Sources/NIO/Utilities.swift
@@ -19,6 +19,17 @@
 /// https://forums.swift.org/t/support-debug-only-code/11037 for a discussion.
 import CNIOLinux
 
+#if os(Windows)
+import let WinSDK.RelationProcessorCore
+
+import func WinSDK.GetLastError
+import func WinSDK.GetLogicalProcessorInformation
+
+import struct WinSDK.SYSTEM_LOGICAL_PROCESSOR_INFORMATION
+
+import typealias WinSDK.DWORD
+#endif
+
 @inlinable
 internal func debugOnly(_ body: () -> Void) {
     assert({ body(); return true }())
@@ -41,7 +52,37 @@ public enum System {
     ///
     /// - returns: The logical core count on the system.
     public static var coreCount: Int {
+#if os(Windows)
+        var dwLength: DWORD = 0
+        _ = GetLogicalProcessorInformation(nil, &dwLength)
+
+        let alignment: Int =
+            MemoryLayout<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>.alignment
+        let pBuffer: UnsafeMutableRawPointer =
+            UnsafeMutableRawPointer.allocate(byteCount: Int(dwLength),
+                                             alignment: alignment)
+        defer {
+            pBuffer.deallocate()
+        }
+
+        let dwSLPICount: Int =
+            Int(dwLength) / MemoryLayout<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>.stride
+        let pSLPI: UnsafeMutablePointer<SYSTEM_LOGICAL_PROCESSOR_INFORMATION> =
+            pBuffer.bindMemory(to: SYSTEM_LOGICAL_PROCESSOR_INFORMATION.self,
+                               capacity: dwSLPICount)
+
+        let bResult: Bool = GetLogicalProcessorInformation(pSLPI, &dwLength)
+        assert(bResult, "GetLogicalProcessorInformation: \(GetLastError())")
+
+        return UnsafeBufferPointer<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>(start: pSLPI,
+                                                                         count: dwSLPICount)
+            .filter { $0.Relationship == RelationProcessorCore }
+            .map { $0.ProcessorMask.nonzeroBitCount }
+            .reduce(0, +)
+
+#else
         return sysconf(CInt(_SC_NPROCESSORS_ONLN))
+#endif
     }
 
     /// A utility function that enumerates the available network interfaces on this machine.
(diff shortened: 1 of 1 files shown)


Sources/NIO/Utilities.swift
```diff
@@ -41,7 +52,34 @@ public enum System {
     ///
     /// - returns: The logical core count on the system.
     public static var coreCount: Int {
+#if os(Windows)
+        var dwLength: DWORD = 0
+        _ = GetLogicalProcessorInformation(nil, &dwLength)
+
+        let alignment: Int =
+            MemoryLayout<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>.alignment
+        let pBuffer: UnsafeMutableRawPointer =
```
weissi: aren't we missing a `defer { ...deallocate() }` here?

compnerd: Yeah, this is missing the defer ... thats what I get for trying to push this into a separate change (I have that locally).

Sources/NIO/Utilities.swift
```diff
@@ -41,7 +52,37 @@ public enum System {
...
+            UnsafeMutableRawPointer.allocate(byteCount: Int(dwLength),
+                                             alignment: alignment)
+        defer {
+            pBuffer.deallocate()
+        }
+
+        let dwSLPICount: Int =
+            Int(dwLength) / MemoryLayout<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>.size
```
Lukasa: I think we have to use `.stride` here, not `.size`, as we're trying to treat this as an array.

(Apropos of nothing it's super annoying that `GetLogicalProcessorInformation` returns the size of the buffer in _bytes_ instead of in element count.

compnerd: Hmm, MSDN agrees about the size.  The returned value doesn’t talk about the stride, it is talking about size.  Consider tail allocations or alignment which would not be included.  This is not swift-esque, this is C more or less.  I think that size is much more clear.

weissi: @compnerd MSDN refers to C's `sizeof` which gives you the _stride_. In Swift, you definitely have to use `.stride` and not `.size`.

Lukasa: +1. This is also entirely natural when you consider that the return value _must_ be aligned to use safely, and so this must be a standard C-style array, which lays out objects based on their stride, not their size.

weissi: Swift:
```
$ cat test.swift
struct a {
    var x: CInt
    var y: CChar
}

print(MemoryLayout<a>.size, MemoryLayout<a>.stride)
$ swift test.swift
5 8
```

C:
```
#include <stdio.h>

struct a {
    int x;
    char y;
};

int main() {
    printf("%lu\n", sizeof(struct a));
    return 0;
}
$ clang test.c && ./a.out
8
```

Lukasa: Though it should be noted that on at least some versions of Swift it seems that Swift "knows" that a struct came from C and just lies. Consider this:

---
`lib.h`

```C
#ifndef LIB_H
#define LIB_H

struct LibStruct {
    int foo;
    char bar;
};

int szof_lib(void);

#endif
```

---

`lib.c`

```C
#include "lib.h"

int szof_lib(void) {
    return sizeof(struct LibStruct);
}
```

---

`main.swift`

```swift
import lib

print(MemoryLayout<LibStruct>.size)
print(MemoryLayout<LibStruct>.stride)
print(szof_lib())
```

---

The output of this program is:

```
8
8
8
```

weissi: yeah, it's the padding from C...

weissi: C adds padding (which Swift imports correctly) to make stride == size. Still, in Swift `.stride` is correct :)

weissi: you can see it with `-Wpadded`:

```
$ clang -Wpadded test.c && ./a.out 
test.c:5:8: warning: padding size of 'struct a' with 3 bytes to alignment boundary [-Wpadded]
struct a {
       ^
1 warning generated.
8, 4
```

Lukasa: Oh sure, I agree that `.stride` is right, it just so happens that in this case `.size` was going to get the right outcome anyway.

weissi: but why not use `.stride` which works everywhere in Swift? If you use `.size`, then you need the programmer who reads this to understand that for C structs `.size == .stride` and that your struct is in fact imported from C, ...

Sources/NIO/Utilities.swift
```diff
@@ -41,7 +52,37 @@ public enum System {
...
+        }
+
+        let dwSLPICount: Int =
+            Int(dwLength) / MemoryLayout<SYSTEM_LOGICAL_PROCESSOR_INFORMATION>.stride
+        let pSLPI: UnsafeMutablePointer<SYSTEM_LOGICAL_PROCESSOR_INFORMATION> =
+            pBuffer.bindMemory(to: SYSTEM_LOGICAL_PROCESSOR_INFORMATION.self,
+                               capacity: dwSLPICount)
+
+        let bResult: Bool = GetLogicalProcessorInformation(pSLPI, &dwLength)
+        assert(bResult, "GetLogicalProcessorInformation: \(GetLastError())")
```
Lukasa:
```suggestion
        precondition(bResult, "GetLogicalProcessorInformation: \(GetLastError())")
```

Accessing the memory if this does not return `true` leads to reading uninitialised memory, which is undefined behaviour. We must crash here.

Lukasa:
@weissi want to re-review?

Sources/NIO/Utilities.swift
compnerd: Because I've found that knowing where the struct is defined has been extremely useful in porting - there are differences in the generated code patterns and generally the debugging technique that works for swift currently is print or read disassembly.

PR #1450, opened 2020-03-15
NIO: add module constructor for Windows

compnerd:
This adds the necessary code to add a module constructor for NIO.  On
Windows, the socketing subsystem (WinSock) must be explicitly
initialized unlike Unices.  Add a module constructor that will be
handled by the C runtime to initialize the subsystem before the module
code is executed so that the user does not need to worry about this.

Since there is no way to inject this in Swift, write a tiny helper in C
and link that into the final NIO binary.

### Motivation:

### Modifications:

### Result:


compnerd:
*sigh* seems that s-p-m doesn't support mixed sources in a directory.  What do you want to do @Lukasa and @weissi?

Lukasa:
Time to define a `CNIOWindows` module! Wrap that in `#ifdef` and move all the definition to there.

compnerd:
Oh, you think that CNIOWindows should be statically linked?  Are the other platform ones statically linked?

Lukasa:
Yup, SwiftPM statically links everything.

compnerd:
That don’t work for Windows :). Static linking of swift code is not supported.

Lukasa:
I'm afraid it's the only option. Mixed source is not supported in SwiftPM and there is no plan to do so.

compnerd:
Okay moved the code into `CNIOWindows`

Sources/CNIOWindows/include/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CNIOWindows {
+  header "CNIOWindows.h"
+  export *
+}
```
Lukasa: Is the module map required? We shouldn't need it for SwiftPM, but I'm acutely aware that the CMake build may have other requirements.

Sources/CNIOWindows/shim.c
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#include "CNIOWindows.h"
```
Lukasa: There's nothing magic about the `shim.c` file, so we can just remove it.

Sources/CNIOWindows/include/module.modulemap
compnerd: It is needed for the cmake based build.  I’ve had it locally for some time.

Sources/CNIOWindows/shim.c
compnerd: This is going to be needed soon (the sendmmsg Abe recvmmsg functions will reside here.  I suppose we can delay adding it.

Lukasa: Ok, we can leave it then.

Sources/CNIOWindows/include/module.modulemap
Lukasa: Ok, I think this shouldn't cause a problem with `generate-xcodeproj` because we shouldn't be importing this on any platform where it matters. /cc @weissi to get his input.

weissi: Hmm, I'm kind of scared we break this again: https://bugs.swift.org/browse/SR-9024

weissi: We need to test that the following setup works:

- creating an Xcode project with NIO with this
- creating a new Xcode project for an app
- then dragging the generated NIO Xcode project as a sub-project into the app Xcode project
- make sure everything still works

So far, this always broke as soon as a modulemap was added anywhere.

Lukasa: Who's volunteering to do that screwing around? :wink:

compnerd: As soon as theres an Xcode on Windows :-p

Lukasa: I think we still need to validate this. I'll do so today.

Lukasa: No observed problem on Xcode 11.4.

compnerd: Thanks @Lukasa for testing it

Sources/CNIOWindows/WSAStartup.c
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#include <Winsock2.h>
```
weissi: doesn't this need to be `#if`'d for windows only?

PR #1451, opened 2020-03-16
Provide a means to control the underlying header storage capacity.

ldewailly:
### Motivation:

It's not always possible to initialize `HTTPHeaders` with all the headers one needs to set. In such case, the underlying array will be expanded as and when needed upon headers being appended. `Array` provides the means to reserve capacity to optimize such use case. This PR exposes this functionality in `HTTPHeaders`.

### Modifications:

Exposed method `reserveCapacity(_ minimumCapacity: Int)` and computed variable `capacity: Int` in `HTTPHeaders` and added some tests.

### Result:

Developers will be able to control the capacity of the underlying headers array as such:

```swift
var headers = HTTPHeaders()
headers.reserveCapacity(5)
```


weissi:

PR #1457, opened 2020-03-20
Convenient initialisers from `ByteBuffer` 

Davidde94:
Added simple initialisers to various types that consume a `ByteBuffer`.

### Motivation:
Common operations to read a `ByteBuffer` as some different type `T` were a little too verbose.

### Modifications:
New initialisers for:
- Data
- String
- [UInt8]

### Result:
```
var buffer = ...
let string = String(from: &buffer)
let data = Data(from: &buffer)
let array = Array<UInt8>(from: &buffer)
```


Sources/NIO/Array+UInt8+ByteBuffer.swift
```diff
@@ -0,0 +1,23 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: wrong copyright

Sources/NIO/String+ByteBuffer.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension String {
+    
+    /// Creates a `String` from a given `ByteBuffer`. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    public init(from buffer: inout ByteBuffer) {
```
weissi: this shouldn't be `inout`

Sources/NIO/Array+UInt8+ByteBuffer.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension Array where Element == UInt8 {
+    
+    /// Creates a `[UInt8]` from the given buffer. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    init(from buffer: inout ByteBuffer) {
```
weissi: this shouldn't be `inout`

Sources/NIO/String+ByteBuffer.swift
```diff
@@ -0,0 +1,23 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi: wrong copyright

Sources/NIOFoundationCompat/Data+ByteBuffer.swift
```diff
@@ -0,0 +1,26 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftNIO project authors
```
weissi: wrong copyright

Sources/NIOFoundationCompat/Data+ByteBuffer.swift
```diff
@@ -0,0 +1,26 @@
...
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import NIO
+
+extension Data {
+    
+    /// Creates a `Data` from a given `ByteBuffer`. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    public init(from buffer: inout ByteBuffer) {
```
weissi: shouldn't be `inout`

weissi: you might want to promote the other arguments to `readData` to here to

Sources/NIOFoundationCompat/Data+ByteBuffer.swift
```diff
@@ -0,0 +1,26 @@
...
+        self = Data(buffer.readBytes(length: buffer.readableBytes)!)
```
weissi: should use `readData`

Tests/LinuxMain.swift
```diff
@@ -105,6 +107,7 @@ class LinuxMainRunnerImpl: LinuxMainRunner {
              testCase(SocketChannelTest.allTests),
              testCase(SocketOptionProviderTest.allTests),
              testCase(StreamChannelTest.allTests),
+             testCase(String_ByteBuffer_Tests.allTests),
```
weissi: can these all go into one test suite please? Maybe `ByteBufferToOtherTypeConversionTest` or so?

Sources/NIO/Array+UInt8+ByteBuffer.swift
```diff
@@ -0,0 +1,23 @@
+//===----------------------------------------------------------------------===//
```
weissi: can we put the `String` and the `Array` stuff in a file `ByteBuffer-conversions.swift` or so? Otherwise they're in totally weird places in the project and not next to the other ByteBuffer ones. Also, fewer files is better if they're that small.

Sources/NIOFoundationCompat/Data+ByteBuffer.swift
```diff
@@ -0,0 +1,26 @@
+//===----------------------------------------------------------------------===//
+//
```
weissi: this one can go into `ByteBuffer-foundation.swift` or whatever it's name is

Sources/NIO/ByteBuffer-conversions.swift
```diff
@@ -0,0 +1,33 @@
+//===----------------------------------------------------------------------===//
```
tomerd: rename to `ByteBuffer+Conversions.swift`?

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
```diff
@@ -212,3 +212,15 @@ extension ByteBufferView: DataProtocol {
...
 extension ByteBufferView: MutableDataProtocol {}
+
+// MARK: - Data
+extension Data {
+    
+    /// Creates a `Data` from a given `ByteBuffer`. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    public init(from buffer: ByteBuffer, byteTransferStrategy: ByteBuffer.ByteTransferStrategy = .automatic) {
+        var buffer = buffer
+        self = buffer.readData(length: buffer.readableBytes, byteTransferStrategy: byteTransferStrategy)!
```
tomerd: guard with nicer error message better than force unwrap?

Sources/NIO/ByteBuffer-conversions.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+    
+}
+
+extension String {
+    
+    /// Creates a `String` from a given `ByteBuffer`. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    public init(from buffer: ByteBuffer) {
+        self = String(decoding: buffer.readableBytesView, as: Unicode.UTF8.self)
```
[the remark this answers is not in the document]
tomerd: what does this return when buffer is empty?

Sources/NIO/ByteBuffer-conversions.swift
Davidde94: I have no preference, but there's a precedence set by `ByteBuffer-aux` and `ByteBuffer-core`, etc

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
Davidde94: I wasn't sure what to do. Having a `guard` seemed a little off as while it's safer the code *should never* execute, if we can't read `buffer.readableBytes` bytes from the buffer then something has gone horribly wrong.

tomerd: agreed it *should never* but if and when it does, do we prefer a clear error message from a guard or a cryptic one from force unwrap? are there any performance overhead in a guard?

Sources/NIO/ByteBuffer-conversions.swift
tomerd: no strong opinion either, its more of a swift pattern to use `+`

Davidde94: @weissi @Lukasa

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
Davidde94: There'll be a minimal performance hit using a guard. What about
```
var buffer = buffer
let string = buffer.readData(length: buffer.readableBytes, byteTransferStrategy: byteTransferStrategy)!
assert(string != nil)
return string!
```

This should avoid the performance issue but still tell us something meaningful if we're in debug mode.

Sources/NIO/ByteBuffer-conversions.swift
weissi: I think we should use `+` in the future but for now I'd stick with existing conventions. Not strong at all

Sources/NIO/ByteBuffer-conversions.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension Array where Element == UInt8 {
+    
+    /// Creates a `[UInt8]` from the given buffer. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    init(from buffer: ByteBuffer) {
```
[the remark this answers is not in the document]
weissi: agreed

Sources/NIO/ByteBuffer-conversions.swift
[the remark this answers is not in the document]
weissi: empty string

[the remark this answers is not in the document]
weissi: this should also probably be faster using `read/getString`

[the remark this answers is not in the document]
weissi: not sure we need encodings tbh

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
weissi: Please leave the `!` and add a comment on why is is here. This will never happen and we don't need to blow up the code size for something that never happens.

weissi: @tomerd I don't think that's a good argument. Certain things are preconditions and we don't need a message to debug. This can not and will not happen

Sources/NIO/ByteBuffer-conversions.swift
Davidde94: Agreed

Sources/NIO/ByteBuffer-conversions.swift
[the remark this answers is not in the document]
Davidde94: I actually expected this method to be faster, which is why I used it. If you're confident readString is faster I can switch :)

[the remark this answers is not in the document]
weissi: very much, https://bugs.swift.org/browse/SR-12125

Sources/NIO/ByteBuffer-conversions.swift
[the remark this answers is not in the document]
Davidde94: Sure

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
tomerd: @weissi I believe we had similar discussion in another thread, if readData should never return nil when passing readableBytes, then having it crash instead of the call site is probably a better API design. we even mentioned the idea of adding an API which is readAllData of sorts that does not return an optional and would be used in call sites like this which would make the API design even more ergonomic. with that said, I dont think this should be holding up this PR but its a repeating theme that could/should be addressed at some point imo

weissi: @tomerd it's impossible to do, you need the `!`. `read*` _can_ return `nil` _unless_ the parameter is `buffer.readableBytes`, in that case it cannot return `nil`. If we implemented `readAllData`, then the `!` would live inside the body of `readAllData` implementation.

 The PR we're discussing here is a PR to `swift-nio` and does what you suggests, it implements the functionality that you propose for `readAllData`, ie. never returning `nil`. Except that it uses Swift's API convention which is: Conversions from `Foo` to `Bar` are not implemented as `Foo.asBar` but rather as `Bar.init(foo: Foo)`. So consider the methods that are implemented here `readAll*`.

Davidde94: So where did we land? Keep the bang?

Sources/NIO/ByteBuffer-conversions.swift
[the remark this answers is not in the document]
Davidde94: @glbrntt I don't think you can specify the encoding in `readString`, unless I'm not looking in the right place...

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
Lukasa: FWIW, I'm also with @weissi, keep the bang. It's right here.

Sources/NIO/ByteBuffer-conversions.swift
```diff
@@ -0,0 +1,34 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
```
Lukasa:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
```

Sources/NIO/ByteBuffer-conversions.swift
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension Array where Element == UInt8 {
+    
+    /// Creates a `[UInt8]` from the given buffer. The entire readable portion of the buffer will be read.
+    /// - parameter buffer: The buffer to read.
+    init(buffer: ByteBuffer) {
+        self = Array(buffer.readableBytesView)
```
Lukasa: I _think_ this generates moderately less efficient code than taking a local `var` and using `readBytes`.

Sources/NIO/ByteBuffer-conversions.swift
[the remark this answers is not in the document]
Davidde94: `readString` with encoding is there, but isn't useful to us because it's in the FoundationCompat target. We're going to punt for a while and address in a future PR

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
Davidde94: Alright, I'll merge as is

ktoso:
Whoop, missed review here -- sorry fixed my email settings / filters a bit recently :-)

LGTM, really nice to see this one

PR #1460, opened 2020-03-22
IO: extend `IOError` for Windows

compnerd:
There are three different error domains that we deal with in NIO, which
need to be differentiated.  Store the error with the domain to ensure
that we can identify the error domain.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/IO.swift
+++ b/Sources/NIO/IO.swift
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(Windows)
+import typealias WinSDK.DWORD
+#endif
+
 /// An `Error` for an IO operation.
 public struct IOError: Swift.Error {
     @available(*, deprecated, message: "NIO no longer uses FailureDescription.")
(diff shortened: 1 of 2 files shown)


Sources/NIO/IO.swift
```diff
@@ -28,16 +32,47 @@ public struct IOError: Swift.Error {
         return .reason(self.failureDescription)
     }
 
+    private enum Error {
+#if os(Windows)
+    case WindowsError(DWORD)
+    case WinSockError(CInt)
+#endif
+    case errno(CInt)
```
Lukasa: Nit: we indent enum `case` declarations.

compnerd: Re-indented.

Sources/NIO/IO.swift
```diff
@@ -28,16 +32,47 @@ public struct IOError: Swift.Error {
...
+              return code
+            #if os(Windows)
+                default:
+                    fatalError("IOError domain is not `errno`")
+            #endif
+        }
+    }
+
+#if os(Windows)
+    public init(WindowsError code: DWORD, reason: String) {
```
weissi: the labels should be lower case, so `windowsError code:...`

Sources/NIO/IO.swift
```diff
@@ -28,16 +32,47 @@ public struct IOError: Swift.Error {
...
+        self.error = .WindowsError(code)
+        self.failureDescription = reason
+    }
+
+    public init(WinSockError code: CInt, reason: String) {
```
weissi: same here

Sources/NIO/IO.swift
```diff
@@ -28,16 +32,47 @@ public struct IOError: Swift.Error {
...
     /// The `errno` that was set for the operation.
-    public let errnoCode: CInt
+    public var errnoCode: CInt {
+        switch self.error {
+            case .errno(let code):
+              return code
+            #if os(Windows)
+                default:
+                    fatalError("IOError domain is not `errno`")
+            #endif
```
Lukasa: I apologise in advance for this, but...we indent the `case` _declaration_, not the `case` statement in a `switch`. That is:

```swift
enum Foo {
    case bar
    case baz
}

switch aFoo {
case bar:
    break
case baz:
    break
}
```

Do you mind unindenting the `case` statements in the `switch`? (While we're here the `return code` is also under-indented, it should be 4 spaces not 2.

PR #1461, opened 2020-03-22
Add new `BSDSocket` namespace

compnerd:
This starts the split of the POSIX/Linux/Darwin/BSD interfaces and the
BSD Socket interfaces in order to support Windows.  The constants on
Windows are not part of the C standard library and need to be explicitly
prefixed.  Use the import by name to avoid the `#if` conditions on the
wrapped versions.  This will allow us to cleanly cleave the dependency
on the underlying C library.

This change adds a `BSDSocket.OptionLevel` and `BSDSocket.Option` types
which allow us to get proper enumerations for these values and pipe them
throughout the NIO codebase.


Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+import let WinSDK.SO_RCVTIMEO
+import let WinSDK.SO_REUSEADDR
+import let WinSDK.SO_REUSE_UNICASTPORT
+#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+import Darwin
+#else
+import Glibc
+#endif
+
+public enum BSDSocket {
```
Lukasa: For public types we need doc comments.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+}
+
+public extension BSDSocket {
```
Lukasa: We are trying to avoid using access modifiers on extensions, and prefer to put them on the individual declarations (it makes it easier to evaluate what the actual access modifier is for a given declaration when you don't have to pop the stack, especially for large extensions).

This applies to all extensions added here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+        public let rawValue: RawValue
+        public init(rawValue: RawValue) {
+            self.rawValue = rawValue
+        }
+    }
+
+    /// Defines configuration option names.
+    struct Option: RawRepresentable {
+        public typealias RawValue = CInt
+        public let rawValue: RawValue
```
Lukasa: This may as well be `var`.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+#endif
+
+public enum BSDSocket {
+}
+
+public extension BSDSocket {
+    /// Defines socket option levels.
+    struct OptionLevel: RawRepresentable {
+        public typealias RawValue = CInt
+        public let rawValue: RawValue
```
Lukasa: This may as well be `var`, as no invariant is violated by changing it.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+            self.rawValue = rawValue
+        }
+    }
+}
+
+extension BSDSocket.OptionLevel: Equatable {
+}
+
+extension BSDSocket.Option: Equatable {
+}
```
Lukasa: Any reason these shouldn't be `Hashable`?

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+
+// Option Level
+public extension BSDSocket.OptionLevel {
+    /// Socket options apply only to IP sockets.
```
Lukasa:
```suggestion
    /// Socket options that apply only to IP sockets.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    #if os(Linux)
+      static let ip: BSDSocket.OptionLevel =
+              BSDSocket.OptionLevel(rawValue: CInt(IPPROTO_IP))
+    #else
+      static let ip: BSDSocket.OptionLevel =
+              BSDSocket.OptionLevel(rawValue: IPPROTO_IP)
+    #endif
+
+    /// Socket options apply only to IPv6 sockets.
```
Lukasa:
```suggestion
    /// Socket options that apply only to IPv6 sockets.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+                BSDSocket.OptionLevel(rawValue: CInt(IPPROTO_IPV6))
+    #elseif os(Windows)
+        static let ipv6: BSDSocket.OptionLevel =
+                BSDSocket.OptionLevel(rawValue: IPPROTO_IPV6.rawValue)
+    #else
+        static let ipv6: BSDSocket.OptionLevel =
+                BSDSocket.OptionLevel(rawValue: IPPROTO_IPV6)
+    #endif
+
+    /// Socket options apply only to TCP sockets.
```
Lukasa:
```suggestion
    /// Socket options that apply only to TCP sockets.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+                BSDSocket.OptionLevel(rawValue: CInt(IPPROTO_TCP))
+    #elseif os(Windows)
+        static let tcp: BSDSocket.OptionLevel =
+                BSDSocket.OptionLevel(rawValue: IPPROTO_TCP.rawValue)
+    #else
+        static let tcp: BSDSocket.OptionLevel =
+                BSDSocket.OptionLevel(rawValue: IPPROTO_TCP)
+    #endif
+
+    /// Socket options apply to all sockets.
```
Lukasa:
```suggestion
    /// Socket options that apply to all sockets.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let socket: BSDSocket.OptionLevel =
+            BSDSocket.OptionLevel(rawValue: SOL_SOCKET)
+}
+
+// IPv4 Options
+public extension BSDSocket.Option {
+    /// Add an IP group membership.
```
Lukasa:
```suggestion
    /// Add a multicast group membership.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let ip_add_membership: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IP_ADD_MEMBERSHIP)
+
+    /// Drop an IP group membership.
```
Lukasa:
```suggestion
    /// Drop a multicast group membership.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let ip_drop_membership: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IP_DROP_MEMBERSHIP)
+
+    /// Set the interface for outgoing multicast packets.
+    static let ip_multicast_if: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IP_MULTICAST_IF)
+
+    /// An IP multicast loopback.
```
Lukasa:
```suggestion
    /// Control multicast loopback.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let ip_multicast_loop: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IP_MULTICAST_LOOP)
+
+    /// An IP multicast time-to-live.
```
Lukasa:
```suggestion
    /// Control multicast time-to-live.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// IPv6 Options
+public extension BSDSocket.Option {
+    static let ipv6_join_group: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IPV6_JOIN_GROUP)
+    static let ipv6_leave_group: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IPV6_LEAVE_GROUP)
+    static let ipv6_multicast_hops: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IPV6_MULTICAST_HOPS)
```
Lukasa: These three need to have doc comments.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    /// Set the interface for outgoign multicast packets.
```
Lukasa:
```suggestion
    /// Set the interface for outgoing multicast packets.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    static let ipv6_multicast_if: BSDSocket.Option =
+            BSDSocket.Option(rawValue: IPV6_MULTICAST_IF)
+
+    /// An IP multicast loopback.
```
Lukasa:
```suggestion
    /// Control multicast loopback.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+// TCP Options
+public extension BSDSocket.Option {
+    /// Disables the Nagle algorithm for send coalescing.
+    static let nodelay: BSDSocket.Option =
+            BSDSocket.Option(rawValue: TCP_NODELAY)
+}
+
+#if os(Linux) || os(FreeBSD)
+public extension BSDSocket.Option {
+    static let info: BSDSocket.Option =
```
Lukasa: We need a doc comment here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+            BSDSocket.Option(rawValue: TCP_INFO)
+}
+#endif
+
+#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+public extension BSDSocket.Option {
+    static let connection_info: BSDSocket.Option =
```
Lukasa: We need a doc comment here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    /// Linger on close if unsent data is present.
+    static let linger: BSDSocket.Option =
+            BSDSocket.Option(rawValue: SO_LINGER)
+
+    /// Specifies the total per-socket buffer space reserved for receives.
+    static let rcvbuf: BSDSocket.Option =
+            BSDSocket.Option(rawValue: SO_RCVBUF)
+
+    /// Specifies the recieve timeout.
```
Lukasa:
```suggestion
    /// Specifies the receive timeout.
```

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -0,0 +1,228 @@
...
+    #endif
+
+    /// Allows the socket to be bound to an address that is already in use.
+    static let reuseaddr: BSDSocket.Option =
+            BSDSocket.Option(rawValue: SO_REUSEADDR)
+}
+
+#if !os(Windows)
+public extension BSDSocket.Option {
+    static let timestamp: BSDSocket.Option =
```
Lukasa: This needs a doc comment.

Sources/NIO/ChannelOption.swift
```diff
@@ -226,7 +226,7 @@ extension ChannelOptions {
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
 public struct ChannelOptions {
     /// - seealso: `SocketOption`.
-    public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
+    public static let socket = { (level: BSDSocket.OptionLevel, name: BSDSocket.Option) -> Types.SocketOption in
```
Lukasa: We cannot change these types this way: we need to deprecate-and-replace.

Sources/NIO/ChannelOption.swift
```diff
@@ -72,15 +72,15 @@ extension ChannelOptions {
...
+            public var level: BSDSocket.OptionLevel
+            public var name: BSDSocket.Option
 
             /// Create a new `SocketOption`.
             ///
             /// - parameters:
             ///     - level: The level for the option as defined in `man setsockopt`, e.g. SO_SOCKET.
             ///     - name: The name of the option as defined in `man setsockopt`, e.g. `SO_REUSEADDR`.
-            public init(level: SocketOptionLevel, name: SocketOptionName) {
+            public init(level: BSDSocket.OptionLevel, name: BSDSocket.Option) {
```
Lukasa: We cannot change these types this way: we need to deprecate-and-replace.

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -70,7 +70,7 @@ public protocol SocketOptionProvider {
     ///     - name: The name of the socket option, e.g. `SO_REUSEADDR`.
     /// - returns: An `EventLoopFuture` containing the value of the socket option, or
     ///     any error that occurred while retrieving the socket option.
-    func unsafeGetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName) -> EventLoopFuture<Value>
+    func unsafeGetSocketOption<Value>(level: BSDSocket.OptionLevel, name: BSDSocket.Option) -> EventLoopFuture<Value>
```
Lukasa: We cannot change these types this way: we need to deprecate and replace.

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -263,11 +263,11 @@ extension BaseSocketChannel: SocketOptionProvider {
...
 
-    func setSocketOption0<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) throws {
-        try self.socket.setOption(level: Int32(level), name: name, value: value)
+    func setSocketOption0<Value>(level: BSDSocket.OptionLevel, name: BSDSocket.Option, value: Value) throws {
+        try self.socket.setOption(level: level, name: name, value: value)
     }
 
-    func getSocketOption0<Value>(level: SocketOptionLevel, name: SocketOptionName) throws -> Value {
-        return try self.socket.getOption(level: Int32(level), name: name)
+    func getSocketOption0<Value>(level: BSDSocket.OptionLevel, name: BSDSocket.Option) throws -> Value {
```
Lukasa: Similarly all these changes need to be deprecate-and-replace.

weissi:
```
23:21:29 ==============================
23:21:29 ERROR: public API change in NIO.json
23:21:29 ==============================
23:21:29 
23:21:29 /* Generic Signature Changes */
23:21:29 
23:21:29 /* RawRepresentable Changes */
23:21:29 
23:21:29 /* Removed Decls */
23:21:29 
23:21:29 /* Moved Decls */
23:21:29 
23:21:29 /* Renamed Decls */
23:21:29 
23:21:29 /* Type Changes */
23:21:29 Accessor ChannelOptions.Types.SocketOption.level.Get() has return type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Accessor ChannelOptions.Types.SocketOption.level.Set() has parameter 0 type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Accessor ChannelOptions.Types.SocketOption.name.Get() has return type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Accessor ChannelOptions.Types.SocketOption.name.Set() has parameter 0 type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Accessor ChannelOptions.socket.Get() has return type change from (NIO.SocketOptionLevel, NIO.SocketOptionName) -> NIO.ChannelOptions.Types.SocketOption to (NIO.BSDSocket.OptionLevel, NIO.BSDSocket.Option) -> NIO.ChannelOptions.Types.SocketOption
23:21:29 Constructor ChannelOptions.Types.SocketOption.init(level:name:) has parameter 0 type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Constructor ChannelOptions.Types.SocketOption.init(level:name:) has parameter 1 type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Func SocketOptionProvider.unsafeGetSocketOption(level:name:) has parameter 0 type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Func SocketOptionProvider.unsafeGetSocketOption(level:name:) has parameter 1 type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Func SocketOptionProvider.unsafeSetSocketOption(level:name:value:) has parameter 0 type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Func SocketOptionProvider.unsafeSetSocketOption(level:name:value:) has parameter 1 type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Var ChannelOptions.Types.SocketOption.level has declared type change from NIO.SocketOptionLevel to NIO.BSDSocket.OptionLevel
23:21:29 Var ChannelOptions.Types.SocketOption.name has declared type change from NIO.SocketOptionName to NIO.BSDSocket.Option
23:21:29 Var ChannelOptions.socket has declared type change from (NIO.SocketOptionLevel, NIO.SocketOptionName) -> NIO.ChannelOptions.Types.SocketOption to (NIO.BSDSocket.OptionLevel, NIO.BSDSocket.Option) -> NIO.ChannelOptions.Types.SocketOption
23:21:29 
23:21:29 /* Decl Attribute changes */
23:21:29 
23:21:29 /* Fixed-layout Type Changes */
23:21:29 
23:21:29 /* Protocol Conformance Change */
23:21:29 
23:21:29 /* Protocol Requirement Change */
23:21:29 
23:21:29 /* Class Inheritance Change */
23:21:29 Checking NIOTestUtils.json... OK
23:21:29 Checking NIOTLS.json... OK
23:21:29 Checking NIOWebSocket.json... OK
23:21:29 done
```

Sources/NIO/BSDSocketAPI.swift
compnerd: Sure, it could conform to `Hashable` as well.  I don't think that the default type'd enumerations conform to `Hashable` which is why I didn't add the conformance, but I've got no problems with that.

Sources/NIO/BSDSocketAPI.swift
weissi: and new public types need to be `NIO` prefixed

weissi: (see https://github.com/apple/swift-nio/blob/master/docs/public-api.md)

compnerd: So should this be called `NIOBSDSocket`?  Or perhaps add it inside another enum to make it `NIO.BSDSocket`?

Lukasa: `NIOBSDSocket` is fine.

Sources/NIO/ChannelOption.swift
```diff
@@ -72,17 +72,37 @@ extension ChannelOptions {
...
-            public var name: SocketOptionName
+            public var optionLevel: NIOBSDSocket.OptionLevel
+            public var optionName: NIOBSDSocket.Option
+
+            public var level: SocketOptionLevel {
+                return SocketOptionLevel(optionLevel.rawValue)
+            }
+            public var name: SocketOptionName {
+                return SocketOptionName(optionName.rawValue)
+            }
```
Lukasa: These two need setters as well to be API compliant.

Sources/NIO/ChannelOption.swift
```diff
@@ -225,9 +245,21 @@ extension ChannelOptions {
...
 public struct ChannelOptions {
+    #if !os(Windows)
+        @available(*, deprecated)
+        public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
+            .init(level: NIOBSDSocket.OptionLevel(rawValue: CInt(level)), name: NIOBSDSocket.Option(rawValue: CInt(name)))
+        }
+    #endif
+
+    /// - seealso: `SocketOption`.
+    public static let sock = { (name: NIOBSDSocket.Option) -> Types.SocketOption in
```
Lukasa: Let's call this `socketOption` instead?

Sources/NIO/ChannelOption.swift
```diff
@@ -225,9 +245,21 @@ extension ChannelOptions {
 
 /// Provides `ChannelOption`s to be used with a `Channel`, `Bootstrap` or `ServerBootstrap`.
 public struct ChannelOptions {
+    #if !os(Windows)
+        @available(*, deprecated)
```
Lukasa: Can we use `renamed:` to help point people at the new one?

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -56,7 +74,23 @@ public protocol SocketOptionProvider {
...
+        /// and thereby read uninitialized or invalid memory. If at all possible, please use one of
+        /// the safe functions defined by this protocol.
+        ///
+        /// - parameters:
+        ///     - level: The socket option level, e.g. `SOL_SOCKET` or `IPPROTO_IP`.
+        ///     - name: The name of the socket option, e.g. `SO_REUSEADDR`.
+        /// - returns: An `EventLoopFuture` containing the value of the socket option, or
+        ///     any error that occurred while retrieving the socket option.
+        func unsafeGetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName) -> EventLoopFuture<Value>
+    #endif
```
Lukasa: We can't add new protocol requirements without having default implementations. Mind providing a default implementation here?

Sources/NIO/ChannelOption.swift
compnerd: Right, added.  I'm just dumb.

Sources/NIO/SocketOptionProvider.swift
compnerd: I thought I had added that, did I miss something?

Sources/NIO/ChannelOption.swift
compnerd: Yeah

Sources/NIO/SocketOptionProvider.swift
compnerd: https://github.com/apple/swift-nio/pull/1461/files#diff-0472f0322d1f2b7085ece7bd3f5ed3d5R271-R275 and https://github.com/apple/swift-nio/pull/1461/files#diff-0472f0322d1f2b7085ece7bd3f5ed3d5R292-R294 provide the default implementation (or am I misunderstanding something?)

Lukasa: I see implementations on `SocketChannel`, but I mean there must be a default implementation in a protocol extension. Otherwise types that previously conformed now will not.

compnerd: Oh, different types >.< ... I get it now.

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -70,9 +104,20 @@ public protocol SocketOptionProvider {
...
+#if !os(Windows)
+    extension SocketOptionProvider {
+        func unsafeSetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName, value: Value) -> EventLoopFuture<Void> {
+            return unsafeSetSocketOption(level: NIOBSDSocket.OptionLevel(rawValue: CInt(level)), name: NIOBSDSocket.Option(rawValue: CInt(name)), value: value)
+        }
+
+        func unsafeGetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName) -> EventLoopFuture<Value> {
+            return unsafeGetSocketOption(level: NIOBSDSocket.OptionLevel(rawValue: CInt(level)), name: NIOBSDSocket.Option(rawValue: CInt(name)))
+        }
+    }
```
Lukasa: Unfortunately these are backwards: we need default implementations for the _new_ protocol requirements, in terms of the _old_ ones.

The goal is that, if someone else has conformed their type to `SocketOptionProvider`, that their type does not break when they update NIO. To achieve that, there must be no new protocol requirements that do not have default implementations.

compnerd: Ah, good point.  Inverted.

Lukasa:
There appear to be compilation errors here that may be worth addressing: specifically, references to `BSDSocket` where the name is `NIOBSDSocket`. Mind fixing those up?

compnerd:
Ugh, I probably mixed up revisions of this patch :-(

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -70,9 +103,20 @@ public protocol SocketOptionProvider {
...
 }
 
+#if !os(Windows)
+    extension SocketOptionProvider {
+        func unsafeSetSocketOption<Value>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option, value: Value) -> EventLoopFuture<Void> {
+            return unsafeSetSocketOption(level: SocketOptionLevel(level.rawValue), name: SocketOptionName(name.rawValue), value: value)
+        }
+
+        func unsafeGetSocketOption<Value>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option) -> EventLoopFuture<Value> {
+            return unsafeGetSocketOption(level: SocketOptionLevel(level.rawValue), name: SocketOptionName(name.rawValue))
```
Lukasa:
```suggestion
            return self.unsafeGetSocketOption(level: SocketOptionLevel(level.rawValue), name: SocketOptionName(name.rawValue))
```

Sources/NIO/SocketOptionProvider.swift
```diff
@@ -70,9 +103,20 @@ public protocol SocketOptionProvider {
...
     /// - returns: An `EventLoopFuture` containing the value of the socket option, or
     ///     any error that occurred while retrieving the socket option.
-    func unsafeGetSocketOption<Value>(level: SocketOptionLevel, name: SocketOptionName) -> EventLoopFuture<Value>
+    func unsafeGetSocketOption<Value>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option) -> EventLoopFuture<Value>
 }
 
+#if !os(Windows)
+    extension SocketOptionProvider {
+        func unsafeSetSocketOption<Value>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option, value: Value) -> EventLoopFuture<Void> {
+            return unsafeSetSocketOption(level: SocketOptionLevel(level.rawValue), name: SocketOptionName(name.rawValue), value: value)
```
Lukasa:
```suggestion
            return self.unsafeSetSocketOption(level: SocketOptionLevel(level.rawValue), name: SocketOptionName(name.rawValue), value: value)
```

Sources/NIO/ChannelOption.swift
```diff
@@ -225,9 +255,21 @@ extension ChannelOptions {
...
+
+    /// - seealso: `SocketOption`.
+    public static let socketOption = { (name: NIOBSDSocket.Option) -> Types.SocketOption in
+        .init(level: .socket, name: name)
+    }
+
     /// - seealso: `SocketOption`.
-    public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
-        .init(level: level, name: name)
+    public static let tcp = { (name: NIOBSDSocket.Option) -> Types.SocketOption in
```
Lukasa: I wonder if this should be `tcpOption` for consistency with the new `socketOption`.

compnerd: Sadly I have to agree with you (but I can do so begrudgingly right?).  I really like `tcp`, but it is nicer to have the uniform `tcpOption` and `socketOption`.

Sources/NIO/ChannelOption.swift
```diff
@@ -225,9 +255,21 @@ extension ChannelOptions {
...
 public struct ChannelOptions {
+    #if !os(Windows)
+        @available(*, deprecated, renamed: "ChannelOptions.socketOption(_:)")
+        public static let socket = { (level: SocketOptionLevel, name: SocketOptionName) -> Types.SocketOption in
+            .init(level: NIOBSDSocket.OptionLevel(rawValue: CInt(level)), name: NIOBSDSocket.Option(rawValue: CInt(name)))
+        }
+    #endif
+
+    /// - seealso: `SocketOption`.
+    public static let socketOption = { (name: NIOBSDSocket.Option) -> Types.SocketOption in
```
Lukasa: I'm wondering if we need a third one of these that does not default the `level` parameter so that folks have an escape hatch. Thoughts?

compnerd: Isn’t that what they have through the old API?  I agree that’s not as convenient.

Lukasa: The old API uses different types, and is deprecated. But there is more than just `SOL_SOCKET` and `IPPROTO_TCP`. For example, there's `IPPROTO_IP`. Rather than try to cover them all in one go, we should just provide a place where users can provide both the level and name.

weissi: @Lukasa I don't think we should / can deprecate the old API. I thought the new stuff is a shorthand for `SOL_SOCKET`. Before we deprecate the old API (which will break every NIO app), we should have an actually _nice_ one like https://github.com/apple/swift-nio/pull/1260 has started to do. Once we have migrated 99% of our users to "nice" `ChannelOptions` that are also usable for Network.framework, then we can start deprecating `.socket(.., ...)` I think

Lukasa: @weissi You may want to re-review this PR then: it adds a completely new namespace for socket options and fundamentally changes their types.

compnerd: I thought that they old API just was unavailable on Windows but not marked as deprecated.

weissi: agree with @compnerd

weissi: @Lukasa the way I understand this patch is that it gives us a new, fully compatible (but any nicer) way of spelling `.socket(.init(SOL_SOCKET), *)` for BSD Sockets, nothing more or less. So it would be 100% compatible with `.socket(.init(SOL_SOCKET), ...)` on UNIX.

On Windows `.socket` would be entirely unavailable and for now we only offer `.socketOption`.

As a separate patch (not done by @compnerd probably), we'd actually offer nice APIs that don't contain the word "socket" for example and look actually "swifty".

@compnerd is that a fair summary?

Lukasa: The patch as it currently stands makes the old API unavailable on Windows and deprecated everywhere else:

https://github.com/apple/swift-nio/blob/0161bd76b2a6eedebc955738deb0478d0fcbd1af/Sources/NIO/ChannelOption.swift#L95-L106

Moreover, to achieve @compnerd's goals we _must_ deprecate the old API. If we do not deprecate the old API, nothing will work on Windows because no-one will move off the current API. While we can move all the NIO examples, the wide range of third-party projects will not move because nothing will tell them that they could move, forcing @compnerd to chase them all. I do not think @compnerd would consider that a success.

Lukasa: Sorry, wrong snippet, I wanted this one:

https://github.com/apple/swift-nio/blob/0161bd76b2a6eedebc955738deb0478d0fcbd1af/Sources/NIO/ChannelOption.swift#L258-L263

weissi: Okay, I'm not happy with the deprecation. We can totally start migrating all the examples to using `.socketOption` instead of `.socket(SOL_SOCKET, ...)` but I don't think we can deprecate.

The reason is: We already did some changes to the socket options, now we're thinking about more changes but what I really want is an _actually nice_ API. So I think it's not cool to deprecate whatever was the last best mechanism all too often.

compnerd: I think that waiting on the subsequent patch that @weissi is talking about for the deprecation to follow in short order is acceptable.  I'm okay with removing the `@available` on `socket` for the time being with the assumption that there is ongoing work to deprecate it.

Lukasa: Great, works for me. @weissi is anyone actually _working_ on that patch?

weissi: ping @Davidde94

Davidde94: Yeh I need to go back to the PR and get it merged. I can try and get this done tomorrow

compnerd: Okay, I think that we should try to get this merged in the mean time - it will allow me to start getting a whole slew of other changes drained.

Davidde94:
Let's hold until after today's meeting - TL;DR I'm fine with this replacing my patch though

Lukasa:
@Davidde94 I don't think this replaces your patch: I think your patch supplements this one. This one just introduces new types for the "not nice" interface so that Windows can work.

Sources/NIO/ChannelOption.swift
weissi: @compnerd totally agreed, I think this is just on hold because it's still breaking public API (says `swift-api-digester`)

weissi:
@Lukasa happy but `swift-api-digester` says this is breaking API :|

Davidde94:
Error is regarding added protocol requirements, could we get away with adding default implementations?

Lukasa:
@weissi @Davidde94 This continues to be the same bug in API digester failing to notice default implementations. They're already there:

https://github.com/apple/swift-nio/blob/e9c9552c4fcbd5e60eeecb6e77a9de0105422cab/Sources/NIO/SocketOptionProvider.swift#L109-L119

Davidde94:
Weird...

compnerd:
Thanks!  I'll start to see if I can rip out some of the other changes so we can get it quickly to the point where others can start contributing :)

PR #1464, opened 2020-03-31
bootstraps: offer ELG validation

weissi:
Motivation:

Today, we just expect the ELGs passed to the bootstraps to be the
correct ones, if not, we crash.

Modifications:

Offer an alternative `validatingGroup:` `init` that just returns `nil`
if the ELGs are of the wrong types.

Result:

Easier to work with multi-stack systems for example when the user might
pass an ELG for either NIO on Sockets or NIO on Network.framework.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -14,6 +14,13 @@
 
 /// The type of all `channelInitializer` callbacks.
 internal typealias ChannelInitializerCallback = (Channel) -> EventLoopFuture<Void>
+
+/// Common functionality for all NIO on sockets bootstraps.
+internal enum NIOOnSocketsBootstraps {
+    internal static func isCompatible(group: EventLoopGroup) -> Bool {
+        return group is SelectableEventLoop || group is MultiThreadedEventLoopGroup
+    }
+}
 
 /// A `ServerBootstrap` is an easy way to bootstrap a `ServerSocketChannel` when creating network servers.
 ///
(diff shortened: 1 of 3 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -384,11 +411,34 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
     private var connectTimeout: TimeAmount = TimeAmount.seconds(10)
     private var resolver: Optional<Resolver>
 
+    private static func isCompatible(group: EventLoopGroup) -> Bool {
+        return group is SelectableEventLoop || group is MultiThreadedEventLoopGroup
```
Davidde94: Minor: Is it possible to extract this code out to avoid duplication?

Sources/NIO/Bootstrap.swift
```diff
@@ -66,20 +66,47 @@ public final class ServerBootstrap {
...
+    /// Create a `ServerBootstrap` on the `EventLoopGroup` `group`.
+    ///
+    /// The `EventLoopGroup` `group` must be compatible, otherwise the program will crash. `ServerBootstrap` is
+    /// compatible only with `MultiThreadedEventLoopGroup` as well as the `EventLoop`s returned by
+    /// `MultiThreadedEventLoopGroup.next`.
     ///
     /// - parameters:
-    ///     - group: The `EventLoopGroup` to use for the `ServerSocketChannel`.
+    ///     - group: The `EventLoopGroup` to use for the `bind` of the `ServerSocketChannel` and to accept new `SocketChannel`s with.
+    ///     - childGroup: The `EventLoopGroup` to run the accepted `SocketChannel`s on.
```
glbrntt: `childGroup` isn't here

Sources/NIO/Bootstrap.swift
```diff
@@ -66,20 +66,47 @@ public final class ServerBootstrap {
...
+            preconditionFailure("ServerBootstrap is only compatible with MultiThreadedEventLoopGroup and " +
+                                "SelectableEventLoop. You tried constructing one with \(group) which is incompatible.")
+        }
+        self.init(validatingGroup: group, childGroup: group)!
+    }
+
+    public convenience init(group: EventLoopGroup, childGroup: EventLoopGroup) {
+        guard ServerBootstrap.isCompatible(group: group) && ServerBootstrap.isCompatible(group: childGroup) else {
+            preconditionFailure("ServerBootstrap is only compatible with MultiThreadedEventLoopGroup and " +
+                                "SelectableEventLoop. You tried constructing one with \(group) which is incompatible.")
```
glbrntt: nit: this is misleading if only `childGroup` is incompatible

Sources/NIO/Bootstrap.swift
```diff
@@ -66,20 +66,47 @@ public final class ServerBootstrap {
...
-    /// Create a `ServerBootstrap` for the `EventLoopGroup` `group`.
+    private static func isCompatible(group: EventLoopGroup) -> Bool {
+        return group is SelectableEventLoop || group is MultiThreadedEventLoopGroup
+    }
+
+    /// Create a `ServerBootstrap` on the `EventLoopGroup` `group`.
+    ///
+    /// The `EventLoopGroup` `group` must be compatible, otherwise the program will crash. `ServerBootstrap` is
+    /// compatible only with `MultiThreadedEventLoopGroup` as well as the `EventLoop`s returned by
+    /// `MultiThreadedEventLoopGroup.next`.
```
glbrntt: Might be worth highlighting the failable init as another option here and on the other bootstraps.

Sources/NIO/Bootstrap.swift
```diff
@@ -66,20 +66,47 @@ public final class ServerBootstrap {
...
+        guard ServerBootstrap.isCompatible(group: group) && ServerBootstrap.isCompatible(group: childGroup) else {
+            preconditionFailure("ServerBootstrap is only compatible with MultiThreadedEventLoopGroup and " +
+                                "SelectableEventLoop. You tried constructing one with \(group) which is incompatible.")
+        }
+        self.init(validatingGroup: group, childGroup: childGroup)!
+
     }
 
-    /// Create a `ServerBootstrap`.
+    /// Create a `ServerBootstrap` on the `EventLoopGroup` `group`, validating that `group` is compatible.
```
Lukasa, line 99: I think this addition has actually made this more confusing, given that this init takes two groups.

PR #1465, opened 2020-03-31
[WIP] [HOLD] Define convenient socket options

Davidde94:
Provided a slightly nicer and easier-to-use way of accessing socket options.

### Motivation:
Socket options were very verbose to set, and difficult to find if you didn't know the exact spelling, prefix, etc. Old usage: `ChannelOptions.socket(SOL_SOCKET, SO_REUSEADDR)`

### Modifications:
Created wrapper options in `ChannelOptions.swift`. New method of use: `ChannelOptions.Socket.allowLocalAddressReuse`

### Result:
Deprecated the old method, with a recommendation to use the new.


Davidde94:
@Lukasa yep, apologies I didn't mark as WIP. We have a meeting later today to discuss the exact API.

Davidde94:
Shutting this down as a result of our meeting

PR #1466, opened 2020-03-31
shorten NIOThreadPool's thread names

weissi:
Motivation:

Previously, NIOThreadPool would use super long thread names. Linux
doesn't allow more than 15 bytes (+ \0 char) so in the end they would
never get set at all.

Modifications:

- Shorten NIOThreadPool's thread names
- Only take the first 15 characters of thread names in general on Linux

Result:

We don't lose thread names on Linux.


Sources/NIO/NIOThreadPool.swift
```diff
@@ -175,7 +175,8 @@ public final class NIOThreadPool {
 
         for id in 0..<self.numberOfThreads {
             group.enter()
-            NIOThread.spawnAndRun(name: "NIOThreadPool thread #\(id)", detachThread: false) { thread in
+            // We should keep thread names under 16 characters because Linux doesn't allow more.
+            NIOThread.spawnAndRun(name: "TP-#\(id)", detachThread: false) { thread in
```
glbrntt, line 179: I know I missed the boat on this one, but there may be some value in having NIO in the name: `"NIOTP-#\(id)"` perhaps?

weissi: @glbrntt this time I deliberately left NIO out because we got so many bug reports about crashes if there's NIO in it :D . People thing that all crashes on NIO* threads are NIO's fault :|

glbrntt: Oh, very sensible then! 😅

PR #1467, opened 2020-04-03
Add `BSDSocket.SocketType` to represent the socket type

compnerd:
This adds a new enumeration which encapsulates the socket type values.
This enables ports to provide the same surface with a different name as
necessary.


Sources/NIO/BSDSocketAPI.swift
```diff
@@ -49,6 +52,23 @@ import Glibc
 public enum NIOBSDSocket {
 }
 
+extension NIOBSDSocket {
+    /// Specifies the type of socket.
+    public struct SocketType: RawRepresentable {
```
Lukasa, line 57: Does this need to be public? We don't appear to use it in any public API surface area.

compnerd: mm, I don't remember why I made it public, lets make it internal and then we can change that if needed.

PR #1468, opened 2020-04-03
Add `BSDSocket.ProtocolFamily`

compnerd:
This encapsulates the enumeration of protocol families rather than
relying on the underlying C library to provide the value.

### Motivation:

### Modifications:

### Result:


Sources/NIO/SocketAddresses.swift
```diff
@@ -119,14 +119,14 @@ public enum SocketAddress: CustomStringConvertible {
     }
 
     /// Returns the protocol family as defined in `man 2 socket` of this `SocketAddress`.
-    public var protocolFamily: Int32 {
+    public var protocolFamily: NIOBSDSocket.ProtocolFamily {
```
Lukasa, line 122: This change is incompatible: we need to deprecate-and-replace.

PR #1469, opened 2020-04-03
Add `BSDSocket.AddressFamily`

compnerd:
This finishes up the encapsulation that is required to ensure that we
are not leaking libc interfaces to clients but instead vend an
enumeration that they can use.  The enumeration is open so although we
may not have the full interface enumerated, users will not be prevented
from using alternate values.

### Motivation:

### Modifications:

### Result:


weissi:
@compnerd merged the PF patch but now this conflicts :|

compnerd:
@weissi rebased :)

PR #1470, opened 2020-04-03
fix accidental deprecation [MUST go into next release]

weissi:
Motivation:

In one of the previous patches, we accidentally merged the deprecation
of `ChannelOptions.Types.SocketOption.init` which will break a lot of
NIO programs.

Modifications:

Undo the deprecation.

Result:

NIOTS builds without warnings.


weissi:
@compnerd no worries at all, that's why we've got the release tools :)

PR #1471, opened 2020-04-05
NIO: Implement BSD sockets APIs for Windows

compnerd:

### Motivation:

### Modifications:

### Result:


Sources/NIO/BSDSocketAPI.swift
```diff
@@ -12,11 +12,20 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if os(Windows)
+#if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+@_exported
+import Darwin.C
+#elseif os(Windows)
+@_exported
```
Lukasa, line 19: Should we be doing an exported import here? I think we regret doing it for libc.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -41,15 +54,107 @@ import let WinSDK.SO_RCVTIMEO
...
+let cShutdown: @convention(c) (CInt, CInt) -> CInt = shutdown
+let cSocket: @convention(c) (CInt, CInt, CInt) -> CInt = socket
+
+let cPread: @convention(c) (CInt, UnsafeMutableRawPointer?, size_t, off_t) -> ssize_t = pread
+let cPwrite: @convention(c) (CInt, UnsafeRawPointer?, size_t, off_t) -> ssize_t = pwrite
+
+let cPoll: @convention(c) (UnsafeMutablePointer<pollfd>?, nfds_t, CInt) -> CInt = poll
+
+let cInet_ntop: @convention(c) (CInt, UnsafeRawPointer?, UnsafeMutablePointer<CChar>?, socklen_t) -> UnsafePointer<CChar>? = inet_ntop
+let cInet_pton: @convention(c) (CInt, UnsafePointer<CChar>?, UnsafeMutableRawPointer?) -> CInt = inet_pton
```
Lukasa, line 98: I'm pretty sure these all used to be `sysX`: why the change?

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+    @inline(never)
+    internal static func getsockopt(socket: NIOBSDSocket.Handle,
+                                    level: NIOBSDSocket.OptionLevel,
+                                    option_name optname: NIOBSDSocket.Option,
+                                    option_value optval: UnsafeMutableRawPointer,
+                                    option_len optlen: UnsafeMutablePointer<socklen_t>)
+            throws {
+        #if os(Windows)
+            if WinSDK.getsockopt(socket, level.rawValue, optname.rawValue,
+                                 optval.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 609: We need a comment here about how `assumingMemoryBound` is wrong, but as we're missing [SR-10246](https://bugs.swift.org/browse/SR-10246) we have very little choice. In practice this limited use is safe because there is no risk of an aliasing violation thanks to the pointer immediately moving into C.

Lukasa: Though actually the "preferred" NIO way to solve this is to write a stub function in C that has the signature we want and then perform the cast in C, where it's explicitly safe.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+        #endif
+    }
+
+    @inline(never)
+    internal static func recv(socket s: NIOBSDSocket.Handle,
+                              buffer buf: UnsafeMutableRawPointer,
+                              length len: size_t) throws -> IOResult<size_t> {
+        #if os(Windows)
+            let iResult: CInt =
+                WinSDK.recv(s, buf.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 640: This has the same pointer concerns above: given the multiple incidences of this patten, it may be better to shim them through a helper.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+    @inline(never)
+    internal static func recvfrom(socket s: NIOBSDSocket.Handle,
+                                  buffer buf: UnsafeMutableRawPointer,
+                                  length len: size_t,
+                                  address from: UnsafeMutablePointer<sockaddr>,
+                                  address_len fromlen: UnsafeMutablePointer<socklen_t>)
+            throws -> IOResult<size_t> {
+        #if os(Windows)
+            let iResult: CInt =
+                WinSDK.recvfrom(s, buf.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 662: Same note about pointer types here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+        #endif
+    }
+
+    @inline(never)
+    internal static func send(socket s: NIOBSDSocket.Handle,
+                              buffer buf: UnsafeRawPointer,
+                              length len: size_t) throws -> IOResult<size_t> {
+        #if os(Windows)
+            let iResult: CInt =
+                WinSDK.send(s, buf.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 681: Pointer types here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+
+    @inline(never)
+    internal static func setsockopt(socket: NIOBSDSocket.Handle,
+                                    level: NIOBSDSocket.OptionLevel,
+                                    option_name optname: NIOBSDSocket.Option,
+                                    option_value optval: UnsafeRawPointer,
+                                    option_len optlen: socklen_t) throws {
+        #if os(Windows)
+            if WinSDK.setsockopt(socket, level.rawValue, optname.rawValue,
+                                 optval.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 702: Pointer types here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+    // which is the same size, but is unsigned.
+    @inline(never)
+    internal static func sendto(socket s: NIOBSDSocket.Handle,
+                                buffer buf: UnsafeRawPointer, length len: size_t,
+                                dest_addr to: UnsafePointer<sockaddr>,
+                                dest_len tolen: socklen_t)
+            throws -> IOResult<size_t> {
+        #if os(Windows)
+            let iResult: CInt =
+                WinSDK.sendto(s, buf.assumingMemoryBound(to: CChar.self),
```
Lukasa, line 724: Pointer types here.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -284,3 +461,533 @@ extension NIOBSDSocket.Option {
...
+    @inline(never)
+    internal static func sendfile(socket s: NIOBSDSocket.Handle, fd: CInt,
+                                  offset: off_t,
+                                  len nNumberOfBytesToWrite: off_t)
+            throws -> IOResult<Int> {
+        #if os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+            var written: off_t = 0
+            do {
+                _ = try syscall(blocking: false) { () -> ssize_t in
+                    var w: off_t = off_t(nNumberOfBytesToWrite)
```
Lukasa, line 914: Can we avoid the one-letter variable name here?

Sources/NIO/SocketAddresses.swift
```diff
@@ -119,14 +119,14 @@ public enum SocketAddress: CustomStringConvertible {
     }
 
     /// Returns the protocol family as defined in `man 2 socket` of this `SocketAddress`.
-    public var protocolFamily: Int32 {
+    public var protocolFamily: NIOBSDSocket.ProtocolFamily {
```
Lukasa, line 122: This type change is incompatible: we need to preserve the old API where possible as well as the new one. Feel free to deprecate-and-replace though.

Sources/NIO/SocketAddresses.swift
```diff
@@ -263,15 +263,15 @@ public enum SocketAddress: CustomStringConvertible {
...
-            if inet_pton(AF_INET, $0, &ipv4Addr) == 1 {
+            if inet_pton(NIOBSDSocket.AddressFamily.inet.rawValue, $0, &ipv4Addr) == 1 {
                 var addr = sockaddr_in()
-                addr.sin_family = sa_family_t(AF_INET)
+                addr.sin_family = sa_family_t(NIOBSDSocket.AddressFamily.inet.rawValue)
                 addr.sin_port = in_port_t(port).bigEndian
                 addr.sin_addr = ipv4Addr
                 return .v4(.init(address: addr, host: ""))
-            } else if inet_pton(AF_INET6, $0, &ipv6Addr) == 1 {
+            } else if inet_pton(NIOBSDSocket.AddressFamily.inet6.rawValue, $0, &ipv6Addr) == 1 {
```
Lukasa, line 272: Do you want these raw pton conversions here, or do you want to use the helpers?

Tests/NIOTests/HappyEyeballsTest.swift
```diff
@@ -129,15 +129,15 @@ private extension SocketAddress {
...
-        if inet_pton(AF_INET, ipAddress, &v4addr) == 1 {
+        if inet_pton(NIOBSDSocket.AddressFamily.inet.rawValue, ipAddress, &v4addr) == 1 {
             var sockaddr = sockaddr_in()
-            sockaddr.sin_family = sa_family_t(AF_INET)
+            sockaddr.sin_family = sa_family_t(NIOBSDSocket.AddressFamily.inet.rawValue)
             sockaddr.sin_port = in_port_t(port).bigEndian
             sockaddr.sin_addr = v4addr
             self = .init(sockaddr, host: host)
-        } else if inet_pton(AF_INET6, ipAddress, &v6addr) == 1 {
+        } else if inet_pton(NIOBSDSocket.AddressFamily.inet6.rawValue, ipAddress, &v6addr) == 1 {
```
Lukasa, line 138: Similarly do you want to use the helpers in this file?

Sources/NIO/BSDSocketAPI.swift
weissi: As @Lukasa points out, `assumingMemoryBound` is not right here. We should use a C wrapper function or if _really_ necessary, we can re-`bind` the memory but that's only valid assuming nobody else holds a `Unsafe*Pointer<original_type>`.

C wrapper is really the way to go. I really don't want NIO to deliberately do the wrong thing. People do look into the NIO codebase for inspiration and I don't think we should do definitely wrong things if there's a way around and in this case, there is. We even already have a `CNIOWindows` module.

Sources/NIO/BSDSocketAPI.swift
compnerd: Yeah, this is a difficult one - maintain parity or do something unfortunate.  I suppose that maintaining parity here is a little less important, particularly since most of the interfaces for networking are not part of the C library anyways.

Sources/NIO/BSDSocketAPI.swift
compnerd: I generally am not fond of the prefixed aliases, but @weissi is adamant about them.  So, this works as a decent enough compromise I think - its a short prefix and indicates that they are from the C world.  Id be okay with `pfn` as well as they are closer to pointers to functions if you like.

Lukasa: I was proposing keeping them as `sys`.

Sources/NIO/BSDSocketAPI.swift
compnerd: That levies a hefty penalty though - I don't think that the function can be fully inlined, plus the multi-language support in s-p-m leaves much to be desired.  I think that the rebinding here is safe - and actually better than in C since the compiler will not assume any aliasing.

Sources/NIO/BSDSocketAPI.swift
compnerd: This is just moved code.  I don't really care - if you want me to change the name, what would you like it changed to?

Sources/NIO/SocketAddresses.swift
compnerd: Yeah, that change is from the previous `ProtocolFamily` change.  This will disappear when rebased.

Sources/NIO/SocketAddresses.swift
compnerd: No, these should go through `NIOBSDSocket`.  Thanks for catching that!

Tests/NIOTests/HappyEyeballsTest.swift
compnerd: Yeap, this should be through the wrappers.

Sources/NIO/BSDSocketAPI.swift
Lukasa: It's not a hefty penalty: it's a non-dynamic direct call to a thunk that will dispatch out to the syscall. The syscall cost is hefty, the indirection is minor.

The rebinding is not safe here because it violates Swift's pointer aliasing model. The fact that the pointer does not escape into Swift but is only passed into C makes this defensible, but it's better to simply be above reproach by not violating the aliasing model at all.

Sources/NIO/BSDSocketAPI.swift
Lukasa: Gross, let's fix it. The name of the parameter in the manual page is `len`, that would be appropriate.

compnerd: Hey, that's exactly what I did for the BSD socket APIs - the man page gives the names.

Sources/NIO/BSDSocketAPI.swift
compnerd: Right, I'm suggesting that `c` is nicer than `sys` as a prefix - although, my personal preference would be to duplicate the code and use the module qualified name.

Sources/NIO/BSDSocketAPI.swift
compnerd: The current thing on Windows is that its not just a single indirection - since static linking on Swift doesn't work on Windows - so its actually ~5 indirections completely destroying the trace cache in the process.

Lukasa: While I grant the concern, I don't think we're expecting the winsock implementation to be a speed demon _anyway_: the goal is more for a future IOCP implementation to be fast, no?

compnerd: The WinSock implementation should be relatively fast, just not scale to 1000s of connections.

Sources/NIO/Bootstrap.swift
```diff
@@ -575,17 +586,29 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+        }
+    #endif
+
     /// Use the existing connected socket file descriptor.
     ///
     /// - parameters:
     ///     - descriptor: The _Unix file descriptor_ representing the connected stream socket.
     /// - returns: an `EventLoopFuture<Channel>` to deliver the `Channel`.
-    public func withConnectedSocket(descriptor: CInt) -> EventLoopFuture<Channel> {
+    public func withConnectedSocket(socket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel> {
```
weissi, line 606: the word socket is duplicated. This should be `withConnectedSocket(_ socket: ...)`

Sources/NIO/Bootstrap.swift
```diff
@@ -575,17 +586,29 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
         }
     }
 
+    #if !os(Windows)
+        /// Use the existing connected socket file descriptor.
+        ///
+        /// - parameters:
+        ///     - descriptor: The _Unix file descriptor_ representing the connected stream socket.
+        /// - returns: an `EventLoopFuture<Channel>` to deliver the `Channel`.
+        @available(*, deprecated, renamed: "ClientBoostrap.withConnectedsocket(socket:)")
```
weissi, line 595: casing is wrong in the `renamed` thing which probably breaks the fixits

Sources/NIO/SocketAddresses.swift
```diff
@@ -264,27 +264,35 @@ public enum SocketAddress: CustomStringConvertible {
...
                 var addr = sockaddr_in6()
                 addr.sin6_family = sa_family_t(NIOBSDSocket.AddressFamily.inet6.rawValue)
                 addr.sin6_port = in_port_t(port).bigEndian
                 addr.sin6_flowinfo = 0
                 addr.sin6_addr = ipv6Addr
                 addr.sin6_scope_id = 0
                 return .v6(.init(address: addr, host: ""))
-            } else {
-                throw SocketAddressError.failedToParseIPString(ipAddress)
+            } catch {
```
weissi, line 292: empty `catch` block

Sources/NIO/SocketAddresses.swift
```diff
@@ -264,27 +264,35 @@ public enum SocketAddress: CustomStringConvertible {
...
+                try NIOBSDSocket.inet_pton(af: .inet, src: $0, dst: &ipv4Addr)
+
                 var addr = sockaddr_in()
                 addr.sin_family = sa_family_t(NIOBSDSocket.AddressFamily.inet.rawValue)
                 addr.sin_port = in_port_t(port).bigEndian
                 addr.sin_addr = ipv4Addr
+
                 return .v4(.init(address: addr, host: ""))
-            } else if inet_pton(NIOBSDSocket.AddressFamily.inet6.rawValue, $0, &ipv6Addr) == 1 {
+            } catch {
```
weissi, line 278: empty `catch` block

Sources/NIO/SocketChannel.swift
```diff
@@ -40,9 +40,16 @@ final class SocketChannel: BaseStreamSocketChannel<Socket> {
         try super.init(socket: socket, parent: nil, eventLoop: eventLoop, recvAllocator: AdaptiveRecvByteBufferAllocator())
     }
 
-    init(eventLoop: SelectableEventLoop, descriptor: CInt) throws {
-        let socket = try Socket(descriptor: descriptor, setNonBlocking: true)
-        try super.init(socket: socket, parent: nil, eventLoop: eventLoop, recvAllocator: AdaptiveRecvByteBufferAllocator())
+    #if !os(Windows)
+        @available(*, deprecated, renamed: "SocketChannel.init(eventLoop:socket:)")
```
weissi, line 44: `SocketChannel` isn't public API, we don't need the deprecation here.

Sources/NIO/SocketChannel.swift
```diff
@@ -147,9 +154,16 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
                        recvAllocator: AdaptiveRecvByteBufferAllocator())
     }
 
-    convenience init(descriptor: CInt, eventLoop: SelectableEventLoop, group: EventLoopGroup) throws {
-        let socket = try ServerSocket(descriptor: descriptor, setNonBlocking: true)
-        try self.init(serverSocket: socket, eventLoop: eventLoop, group: group)
+    #if !os(Windows)
```
weissi, line 157: `ServerSocketChannel` isn't public API

Sources/NIO/SocketChannel.swift
```diff
@@ -343,8 +357,15 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
         return super.isOpen
     }
 
-    convenience init(eventLoop: SelectableEventLoop, descriptor: CInt) throws {
-        let socket = try Socket(descriptor: descriptor)
+    #if !os(Windows)
+        @available(*, deprecated, renamed: "DatagramChannel.init(eventLoop:socket:)")
```
weissi, line 361: not public API

Sources/NIO/BSDSocketAPI.swift
weissi: My primary concern is to twofold:
- keep the diff as small as possible
- no duplication of code. Darwin & Linux match almost perfectly so we don't need to duplicate all code

Tests/NIOTests/BootstrapTest.swift
```diff
@@ -196,7 +197,7 @@ class BootstrapTest: XCTestCase {
                 }
                 return self.freshEventLoop().makeSucceededFuture(())
             }
-            .withBoundSocket(descriptor: socket)
+            .withBoundSocket(socket: socket)
```
weissi, line 200: we need new tests that test _all_ deprecated functionality. It's super risky for us to merge this because no code is exercising those. 

The tests that test deprecated functionality should look like

```
    @available(*, deprecated, message: "deprecated because it tests deprecated functionality")
    func testDeprecatedXYZ() {
```

Sources/NIO/BSDSocketAPI.swift
weissi: Why can't they also stay in `System.swift`? I always really liked to have just one place where I can find all the system calls that NIO makes.

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -365,3 +460,533 @@ extension NIOBSDSocket.Option {
...
+
+            return fd
+        #endif
+    }
+
+    @inline(never)
+    internal static func bind(socket s: NIOBSDSocket.Handle,
+                              address addr: UnsafePointer<sockaddr>,
+                              address_len namelen: socklen_t) throws {
+        #if os(Windows)
```
weissi, line 511: I really struggle to read this `#if os` code. Couldn't we just rename `System.swift` to `SystemUNIX.swift` and make a Windows version that defines exactly the same ones without

```
#if os(Windows)
    complicatedCode()
#else if os(...) 
    moreComplicatedCode()
#endif
```

NIO is in active development and we have to maintain all this so I'd really appreciate if we could try to make the code easy to read and follow.

Sources/NIO/BSDSocketAPI.swift
weissi: NIO is always correctness over speed, we really can't just violate the memory model for some potential efficiency gains.

Sources/NIO/BSDSocketAPI.swift
weissi: I think `ThreadPosix.swift` & `ThreadWindows.swift` are good examples how the code can be still easy to read and yet portable. It's very clear which file you have to open, there's no performance overhead, and because of the (unused) protocol we still have clarity on what has to be implemented with what exact types.

Sources/NIO/BSDSocketAPI.swift
compnerd: I think that having the BSD sockets API separated from the C library surface helps immensely in understanding the codebase (its read more than written), it helps discoverability, it helps porting, etc.  System.swift was immensely large, and it is going to have to grow to at least 3x its current size.

Sources/NIO/BSDSocketAPI.swift
compnerd: I am curious as to how this actually results in incorrect behaviour.  But, I think that we *should* actually change the signature from the BSD API to the Windows API then - because taking a `char *` is much safer for a buffer than a `void *` (aliasing and alignment rules).

Sources/NIO/BSDSocketAPI.swift
compnerd: Doesn't that exactly violate your idea of you having one place to find the code? :)  I think that it would be better to introduce the indirection in a follow up change much like you did with the threading.  I think that it is going to be really odd having `BSDSocketAPIUnix.swift` and `BSDSocketWindows.swift` though, but thats just a minor point.  IMO this current structure does make it easier to compare the behaviour which is valuable.

Sources/NIO/SocketAddresses.swift
compnerd: The empty `catch` is much nicer to follow - it prevents the embedded, deeply nested approach.  Is there a better way to write this?  The `guard ...` approach is just as nested.

Lukasa: If you're going to do that you have to put a comment in it pointing out that it's intentional.

Sources/NIO/BSDSocketAPI.swift
weissi: I personally find code that is full of `#if os(...)` impossible to read. I would just go to `BSDSocketAPIUnix.swift` and that gives me all the code. If something's only not working on Windows, I'd go to `BSDSocketAPIWindows.swift`.

Sure, for the porting effort it's very valuable to see both behaviours at once but after the porting is done I think it's more important that the code is readable and I really struggle hard with lots of `#if os(...)`. I'm fine with them if all is super trivial and restricted to very few files. But the code sockets API full of `#if os(...)`es would be a touch pill for me to swallow.

CC @Lukasa / @glbrntt / @PeterAdams-A / @Davidde94 how do you feel about lots of `#if os(...)` interspersed in code?

glbrntt: I'm a +1 on not having lots of `#if os(...)`: splitting into unix/windows specific files is my preferred option here.

Davidde94: In general I think the more `#if`s then the harder the code is to read and maintain. 2 files would work, but I'm sure it would also be possible to elegantly group code within one file but with far fewer directives.

Sources/NIO/BSDSocketAPI.swift
weissi: It's undefined behaviour in Swift and there's talk about adding tooling to catch this. The compiler is free to exploit the undefined behaviour in any way it wants.

- `char *` is aligned to 1 byte IIRC, ie. 'no special alignment', how can this be better than `void *`
- what are the aliasing problems you're thinking of? I can only think of aliasing problems if we _do_ alias the same pointer as `UnsafePointer<CChar>` and `UnsafePointer<CInt>` for example (for a `CInt` socket option). This is illegal in Swift (legal in C I know). 

`void *` is AFAIK the only correct type here. We offer this API a pointer that is bound to a certain type, the only valid thing that you can do in Swift is to either leave it bound to that type, or make a raw pointer out of it. Making a raw pointer out of it doesn't change the alignment.
`char *` isn't the right type as it gets imported as a pointer to memory bound to `CChar.self` which isn't what we need here. If we did a `CChar.self` based API we would always need to `bind` it to the correct type and that is best avoided.

The right solution is to shim the Windows API into `void *` to get the `Unsafe*Raw*Pointer` we actually need.

Sources/NIO/SocketAddresses.swift
compnerd: Sure, happy to add a comment, that request wasn't clear.  If there is a better way to write this, I'm also okay with that, I just don't know of one.

Sources/NIO/BSDSocketAPI.swift
compnerd: @Davidde94 what did you have in mind?  I can try to regroup things for clarity.

Sources/NIO/BSDSocketAPI.swift
compnerd: I vehemently disagree with you on the `void *` bit.  The correct type *is* `char *`.  The semantics of the C model make it the only correct type as it avoids all the potential problems with alignment and aliasing.  You do not know how the pointer is going to be used, but you do know that it is a buffer.  The access pattern can not be identified.  I agree that we will need to shim everything to get to `char *`, but as you said, we already have `CNIODarwin` and `CNIOLinux` and it is possible to have shims for this.

I think that the Windows signature is far safer here, and should be preferred, but, I'm willing to live with re-binding.

Lukasa: I think you two are talking past each other. In C, `char *` is the right type for this API, but in Swift, `void *` is. This is unavoidably because the "avoids alignment and aliasing" pointer type in Swift is _not_ `UnsafePointer<CChar>`, it's `UnsafeRawPointer`, and to get an `UnsafeRawPointer` you _must_ have a `void *` from C.

We can leave aside the question of whether that was an acceptable design decision or not, but it remains true. In Swift today `UnsafePointer<CChar>` is forbidden from aliasing any other bound pointer.

It is this impedance mismatch that is why we are recommending a C thunk: the thunk allows Swift to hold pointers that match its memory model, while C can perform the cast that is defined as being _entirely_ safe in C.

compnerd: I think that exposing the C signature makes sense though - and we can do the thunk for Linux/macOS just as well.  (I don't think that aliasing/casting is entirely safe in C either ... clang and LLVM can do a lot of clever things).

Lukasa: The problem with using the C signature in Swift is that it forces us to make an unsafe cast _in Swift_.

> (I don't think that aliasing/casting is entirely safe in C either ... clang and LLVM can do a lot of clever things).

This specific cast is what is safe in C. Specifically, the C strict aliasing rule explicitly allows `void *` and `char *` to alias pointers to other types. Thus, in C it is always safe to take any pointer and cast it to `void *` or `char *`. However, in Swift it is only safe to take any pointer and create a _raw_ pointer from it.

That's why we're pushing for the thunk in C: C is _always_ allowed to cast a `void *` to a `char *` without violating aliasing rules. If clang punishes you for this, clang is buggy. However, Swift is not, and if we make this cast here and `swiftc` punishes us for it, `swiftc` will be right.

compnerd: Sure, but, surely thunking where it is cheaper (Linux/macOS) is also reasonable right?

weissi: @compnerd which aliasing/alignment concerns are you referring to? What makes `char *` “far safer”?

compnerd: `char *` has the proper alignment requirements: 1.  `void *` has different alignment requirements.

weissi: Right, so `char *` is effectively unaligned since every possible pointer is correctly aligned for a `char *`. So every `void *` must also be a correctly aligned `char *`, agreed?

weissi: I can also not see any aliasing problems because C excludes `void *` and `char *` from the aliasing do-not-alias-a-pointer-of-different-type alias rule.

compnerd: Im just suggesting that the Swift side of the API be `char *` and we can thunk through C if we must.

weissi: Okay, let's look at some facts:

- The `get`/`setsockopt` APIs work with a multitude of types, eg. `CInt`, `linger`, ... We cannot control what they have to be because the OS decides.
- The APIs want pointers to the values that are in different types

I guess we all agree that if we could design this API in C, we would pick `void *`. However, in C (because the aliasing rules do not apply for `char *` & `void *`), we have one extra option which is `char *`. Picking `char *` is Microsoft did is highly questionable (because it's not character data) but at least in C it's not wrong. Ie. the API can be used without violating the aliasing rules.

In Swift however, the aliasing rules are _stricter_ than in C. There's precisely one way to safely alias any typed pointer and that is `Unsafe*Raw*Pointer`. This decides for us that the Swift API has to be `Unsafe*Raw*Pointer`. There is no other option. Even if we accepted explicit rebinding of pointers as an option. Rebinding is only allowed if we can prove that there are no other pointers left that are assuming the old binding.

Example:

Valid Swift program:

```swift
func iTakeCharPointers(_ ptr: UnsafePointer<CChar>) { ... }

let iAmAnIntPointer = UnsafeMutablePointer<CInt>.allocate(count: 1)
// assumingMemoryBound(to: CChar.self) would be just wrong here because it's bound to CInt
// withMemoryRebound(to: CChar.self) is also not right because it requires the size of the old and the new type to be the same (but CChar and CInt aren't)
let reboundPtr = UnsafeMutableRawPointer(iAmAnIntPointer).bind(to: CChar.self, capacity: MemoryLayout<CInt>.size)
iTakeCharPointers(reboundPtr)
[...] // crucially no more use of `iAmAnIntPointer` unless we also bind it back to `CInt.self`
```

Invalid Swift program:

```swift
func iTakeCharPointers(_ ptr: UnsafePointer<CChar>) { ... }

class PointerHolder {
    internal let iAmAnIntPointer: UnsafeMutablePointer<Int>

   init(...) { ... }

   deinit {
        self.iAmAnIntPointer.deallocate(...)
   }

   func foo() {
       let reboundPtr = UnsafeMutableRawPointer(iAmAnIntPointer).bind(to: CChar.self, capacity: MemoryLayout<CInt>.size)
       iTakeCharPointers(reboundPtr)
       [...]
    }
}
```

This program snippet rebinds the pointer but can't make sure no pointers bound the the old type (`CInt`) is still around.

If however, we had a function `iTakeRawPointers` instead of `iTakeCharPointers` the problem would totally disappear:

Valid Swift programs:

```swift
func iTakeRawPointers(_ ptr: UnsafeRawBufferPointer) { ... }

let iAmAnIntPointer = UnsafeMutablePointer<CInt>.allocate(count: 1)
let rawPtr = UnsafeMutableRawBufferPointer(start: iAmAnIntPointer, byteCount: MemoryLayout<CInt>.self)
iTakeRawPointers(reboundPtr)
[...] // other uses of `iAmAnIntPointer` are not a problem

class PointerHolder {
    internal let iAmAnIntPointer: UnsafeMutablePointer<Int>

   init(...) { ... }

   deinit {
        self.iAmAnIntPointer.deallocate(...)
   }

   func foo() {
       let rawPtr = UnsafeMutableRawBufferPointer(start: iAmAnIntPointer, byteCount: MemoryLayout<CInt>.self)
        iTakeRawPointers(reboundPtr)
        [...]
    }
}
```

I think this clearly shows that `UnsafeRawBufferPointer` is the only acceptable answer in Swift, for 2 reasons:
- getting from any typed or raw pointer to an `UnsafeRawBufferPointer` is very easy (in fact to get from one typed pointer to another typed one, you need to first go through a raw pointer)
- we don't need to consider the whole program which is likely impossible to check what other types a pointer may still be bound to

## Solving the actual problem

After having figured out what the pretty much only possible type in Swift is (`UnsafeRaw(Buffer)Pointer`) we need to see how we can now interface with the C functions. The C functions take a `void *` (UNIX) or a `char *` (C).

Given the C aliasing rules both are acceptable types for such an API so the only issue we've got left to solve is how to get the C functions imported with an `UnsafeRaw(Buffer)Pointer`.

Sadly, we cannot control how C functions are getting imported which leaves us with:

- `void *` --> `UnsafeMutableRawPointer` ✅
- `char *` --> `UnsafeMutablePointer<CChar>` ❌

The only solution in Swift today is to add a little

```C
static inline int c_nio_swift_getsockopt(int socket, int level, int option_name, void *option_value, socklen_t *restrict option_len) {
    return getsockopt(socket, level, option_name, (char *)option_value, option_len);
}
```

in a `shim.h` or similar. The shim is legal in C because any pointer can validly be casted to a `char *`.

@compnerd to you agree with this? Btw, this is what we [do for the HTTP parser too](https://github.com/apple/swift-nio/blob/master/Sources/CNIOHTTPParser/include/CNIOHTTPParser.h#L22-L27) despite the fact that for HTTP `char *` is a lot closer to the reality (character data) than for socket options but still working with `CChar` wouldn't actually correctly work.

compnerd: The part that we disagree on is that char * is a totally valid and arguably better choice in C.

The problem with the skin is that s-p-m doesn’t allow multi-language libraries and static linking isn’t change in Windows for swift.  We can however accomplish that for the other platforms.

Let’s flip the signature string and use the thunk on Linux/macOS.  Alternatively, let’s add a CMake based build system into the tree and I should be able to coerce that into doing something reasonable.

weissi: @compnerd I don’t quite see how linkage matters for `static inline`, I haven’t checked but I would expect it to work fine.

I also don’t understand  why `char *` would be a better type. It’s meant for string data which this isn’t. This is essentially a “pointer to some unknown type” which `void *` is perfect for. What would an advantage of `char *` be? Apart obviously on Windows where `get/setsockopt` take the non-standard `char *` argument so you can save one cast.

weissi: Update: yes, I can confirm `static inline` works as you'd expect.

I created (in module `CTest`) a `CTest.h` file with the following (terrible) content:

```C
#include <string.h>

static inline void *my_memcpy(char *dst, char *src, size_t n) {
	src[n] = 0;
	strcpy(dst, src);
	return dst;
}
```

The compiler was too clever if I put `memcpy` in there because it replaced it by a direct copy of 1 byte, so I put something more terrible in there...

My Swift program is

```swift
import CTest

@inline(never)
func foo() -> CChar {
    let dst = UnsafeMutablePointer<CChar>.allocate(capacity: 2)
    let src = UnsafeMutablePointer<CChar>.allocate(capacity: 2)
    src[0] = 0x41

    my_memcpy(dst, src, 1)
    return dst[0]
}

print(foo())
```

and the disassembly is:

```asm
testp.foo() -> Swift.Int8:
0000000100001090	pushq	%rbp
0000000100001091	movq	%rsp, %rbp
0000000100001094	pushq	%rbx
0000000100001095	pushq	%rax
0000000100001096	movl	$0x2, %edi
000000010000109b	movq	$-0x1, %rsi
00000001000010a2	callq	0x1000035f6 ## symbol stub for: _swift_slowAlloc
00000001000010a7	movq	%rax, %rbx
00000001000010aa	movl	$0x2, %edi
00000001000010af	movq	$-0x1, %rsi
00000001000010b6	callq	0x1000035f6 ## symbol stub for: _swift_slowAlloc
00000001000010bb	movw	$0x41, (%rax)
00000001000010c0	movq	%rbx, %rdi
00000001000010c3	movq	%rax, %rsi
00000001000010c6	callq	0x1000035c6 ## symbol stub for: _strcpy
00000001000010cb	movb	(%rbx), %al
00000001000010cd	addq	$0x8, %rsp
00000001000010d1	popq	%rbx
00000001000010d2	popq	%rbp
00000001000010d3	retq
```

As you can see, it inlined the whole C code and calls `strcpy` directly from Swift.

So in `CNIOWindows`'s `CNIOWindows.h` we should just add a 

```C
static inline int c_nio_swift_getsockopt(int socket, int level, int option_name, void *option_value, socklen_t *restrict option_len) {
    return getsockopt(socket, level, option_name, (char *)option_value, option_len);
}
```

and we're all good.

compnerd: > I also don’t understand why `char *` would be a better type. It’s meant for string data which this isn’t. This is essentially a “pointer to some unknown type” which `void *` is perfect for. What would an advantage of `char *` be? Apart obviously on Windows where `get/setsockopt` take the non-standard `char *` argument so you can save one cast.

It isn't meant for _just_ string data, its an arbitrary byte array.  `void *` has different alignment requirements than `char *`.

weissi: @compnerd `char *` has the weakest alignment requirements you can have. `void *` cannot be less than 1-byte aligned, they’re fully compatible. The standard allows any pointer to be casted to `void *` and `char *` so I respectfully disagree that alignment issues are an argument here. Neither are aliasing concerns.

compnerd: `void *` is pointer alignment, which is usually word alignment, which is why there are potential alignment issues.

weissi: So having something _more_ aligned is never a problem. Despite that, it's also not accurate that void pointer is more aligned than a char pointer.

Maybe the following is confusing you? The alignment of `void *` _the pointer itself_ is indeed a word! But so is `char *` and every other pointer type, you can ask the compiler about `_Alignof(void *)` and `_Alignof(char *)` which will both be 8 on 64-bit platforms. Of course, this isn't what we've been talking about because what matters in our discussion is if the alignment of what the pointer _is pointing to_ is aligned enough for the data. Ie. we'd need to compare the alignment of `void` vs `char`. Now `char` is 1 byte aligned of course and the question doesn't really make sense for `void` because it's not a data type that has size so alignment, size, stride, ... are all questions that don't make sense because you need to know what _actual_ underlying type is underneath. So for socket options this could be `int`, `struct linger`, ... If anything your argument would be one against `char *` because you could assume that a 1-byte aligned memory is enough for `char *` but that is no longer true if we're aliasing say an `int *` as a `char *`.

Also for all practical purposes, `void` has exactly the same alignment as `char`. Again, if you accept the warning asking the wrong question, the compiler will tell you too:

```
printf("%lu %lu %lu %lu\n", _Alignof(void), _Alignof(char), _Alignof(void *), _Alignof(char *));
```

is

```
warning: invalid application of 'alignof' to a void type [-Wpointer-arith]
1 1 8 8
```

Which I hope finally busts the myth that `void *` has any builtin alignment requirements. If we pass a `void *` to a `setsockopt`/`getsockopt` we must make sure it's aligned correctly for the _actual_ type (`int`, `struct linger`, ...). The same caveat applies to `char *` in the very same way, except that we could make your argument that `char` is only 1-byte aligned so it may be deceiving if we require say 4 bytes for an `int` but alias it as a `char *`. Again, `void *` feels much better because it doesn't have any alignment requirements on its own, that's why I call it "unaligned". For practical purposes however unaligned == 1 byte aligned because in both cases every possible pointer value is aligned enough.

Sources/NIO/BSDSocketAPI.swift
Lukasa: Hey @weissi @Davidde94 can you revisit this? I'd like to unblock this work as it will also unblock #1473 and subsequent work.

weissi: I don't think we can take lots of `#if os(...)` in non-trivial code.

weissi: In my opinion, this file does the exact opposite of what I think makes maintainable code. It clumps together different implementations for the same interface.

I think we should create the interface and then have two files implementing it. One for Windows, one for Posix.

I'm also not happy with the syscalls being taken out of `System.swift`. `System.swift` is our (poorly named) place where we define all Posix syscalls. We have many times done refactorings to exactly how we call into those syscalls and I think it's crucial that they stay in one place which is `System.swift`. I'm happy renaming it `PosixSystem.swift` or so.

Lukasa:
Ok, I've updated this branch with a change that I believe addresses @weissi's primary concern. @compnerd when you get a moment it'd be good if you could validate that this appears to be ok. FYI, I rebased this on top of the current master, which means that this branch is now out of step with your local branch, you may need to rebase your own local branch to get this to apply cleanly.

PeterAdams-A:
Hi @compnerd.  I was wondering what the status of your windows port is - I know the NIO team was delaying adding recvmsg and friends to try and give you a chance to work on this on a stable base.  It would be useful add this extra functionality soon though - could this be done or would it be too disruptive?

compnerd:
@PeterAdams-A - ah, so @Lukasa and I spoke about this, I think that we agreed to try to get this change in and then merge the other changes that you are talking about to prevent the change from being too disruptive (as this largely should allow that to be a pretty easy transition).

PR #1473, opened 2020-04-07
Move DatagramChannel over to using recvmsg.

Lukasa:
Motivation:

When attempting to obtain metadata about a read, we need to use recvmsg
in order to obtain that data. While we're using recvmmsg for vector
reads right now, our scalar reads use recvfrom, which does not provide
us with that metadata.

While we're not extracting metadata right now, we may well do so in
future, so we can apply it now.

Modifications:

- Move recvfrom usage to recvmsg.

Result:

We'll be able to extend recvmsg to extract metadata.


Sources/NIO/Socket.swift
```diff
@@ -170,6 +170,33 @@ typealias IOVector = iovec
         }
     }
 
+    func recvmsg(pointer: UnsafeMutableRawBufferPointer, storage: inout sockaddr_storage, storageLen: inout socklen_t, controlBytes: inout UnsafeMutableRawBufferPointer) throws -> IOResult<Int> {
```
weissi, line 173: I don't think we should do `inout UnsafeMutableRawBufferPointer`. That's pretty dangerous because the user can't know that we don't mess with the base address. If we were to mess with the base address, then the user couldn't free this anymore.

The right type for this IMHO is: `inout Slice<UMRBP>`. Then it's clear that the user gives us a slice and we may mess with the slice but the user is always in charge of allocating/freeing the UMRBP and we can't mess with that one.

Sources/NIO/SocketChannel.swift
```diff
@@ -464,14 +464,17 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+
         for i in 1...self.maxMessagesPerRead {
             guard self.isOpen else {
                 throw ChannelError.eof
             }
             buffer.clear()
 
             let result = try buffer.withMutableWritePointer {
-                try self.socket.recvfrom(pointer: $0, storage: &rawAddress, storageLen: &rawAddressLength)
+                try self.socket.recvmsg(pointer: $0, storage: &rawAddress, storageLen: &rawAddressLength, controlBytes: &controlBytes)
```
weissi, line 477: I think we should start migrating off the auto-magic stuff and replace `&rawAddress` and `&rawAddressLength` by `withUnsafe*`. WDYT?

Sources/NIO/Socket.swift
```diff
@@ -170,6 +170,33 @@ typealias IOVector = iovec
...
+        var vec = iovec(iov_base: pointer.baseAddress, iov_len: pointer.count)
+
+        return try withUnsafeMutablePointer(to: &vec) { vecPtr in
+            return try storage.withMutableSockAddr { (sockaddrPtr, _) in
+                var messageHeader = msghdr(msg_name: sockaddrPtr,
+                                           msg_namelen: storageLen,
+                                           msg_iov: vecPtr,
+                                           msg_iovlen: 1,
+                                           msg_control: controlBytes.baseAddress,
+                                           msg_controllen: .init(controlBytes.count),
```
weissi, line 183: thanks UNIX for providing such fantastic APIs :)

Lukasa: It's bloody annoying. `msg_controllen` is `socklen_t` on Darwin and `ssize_t` on Linux.

Sources/NIO/SocketChannel.swift
Lukasa: Sure, works for me.

Lukasa: Oh, no, doesn't work for me. `recvmsg` is asking for inouts, not pointers, so there's no magic here at all.

Sources/NIO/Socket.swift
weissi: @Lukasa oh sorry! Mis-reviewed then. My opinion is `&` is (obviously) fine and good for `inout` but bad for `Unsafe*Pointer`. Fairly sure we share that opinion anyway :D. With Swift had a sigil that _only_ worked on `inout`s so that I won't misreview like this in the future :)

Sources/NIO/Socket.swift
Lukasa: Addressed.

compnerd:
Thanks for the heads up @Lukasa.  I think I would rather get #1471 settled and then address the new interface to be added to the `NIOBSDSocket` interface.  This should be possible to support on Windows, but will need a little bit of work.  Given the complexity of that change, Id rather that it settle first - since that is renaming/moving all of the BSD Sockets API to make it easier to deal with other ports and implementing the Windows port of that API surface as well.

Lukasa:
Ok, I've updated this patch to use `NIOBSDSocket` instead of `Posix` to abstract `recvmsg`. Right now we don't have `recvmsg` working in Windows so it just fatal errors.

Sources/NIO/Socket.swift
```diff
@@ -201,6 +201,45 @@ typealias IOVector = iovec
         }
     }
 
+    /// Receive data from the socket, along with aditional control information.
```
glbrntt, line 204:
```suggestion
    /// Receive data from the socket, along with additional control information.
```

Sources/NIO/Socket.swift
```diff
@@ -201,6 +201,45 @@ typealias IOVector = iovec
...
+            return try storage.withMutableSockAddr { (sockaddrPtr, _) in
+                var messageHeader = msghdr(msg_name: sockaddrPtr,
+                                           msg_namelen: storageLen,
+                                           msg_iov: vecPtr,
+                                           msg_iovlen: 1,
+                                           msg_control: localControlBytePointer.baseAddress,
+                                           msg_controllen: .init(localControlBytePointer.count),
+                                           msg_flags: 0)
+                defer {
+                    // We need to write back the length of the message and the control bytes.
```
glbrntt, line 229: Should we still write these back if we throw?

Lukasa: Yes. We only throw _after_ the syscall occurs, and the syscall may well have set the values we're reading. To clarify that, we ensure that the values are written back. Note that if these values were _not_ set, they'll be defaulted to the original values: `messageHeader.msg_namelen` is initialised above to `storageLen`, and `messageHeader.msg_controllen` is initialised to `localControlBytePointer.count`.

Sources/NIO/Socket.swift
```diff
@@ -201,6 +201,45 @@ typealias IOVector = iovec
...
+    /// - parameters:
+    ///     - pointer: The pointer (and size) to the storage into which the data should be read.
+    ///     - storage: The address from which the data was received
+    ///     - storageLen: The size of the storage itself.
+    ///     - controlBytes: A region of a buffer into which control data can be written. This parameter will be modified on return to be
+    ///         the slice of the data actually written into, if any.
+    /// - returns: The `IOResult` which indicates how much data could be received and if the operation returned before all the data could be received
+    ///     (because the socket is in non-blocking mode)
+    /// - throws: An `IOError` if the operation failed.
+    func recvmsg(pointer: UnsafeMutableRawBufferPointer, storage: inout sockaddr_storage, storageLen: inout socklen_t, controlBytes: inout Slice<UnsafeMutableRawBufferPointer>) throws -> IOResult<Int> {
```
PeterAdams-A, line 215: You should put the arguments on separate lines - see comment on later PR ;-)

Sources/NIO/SocketChannel.swift
```diff
@@ -490,14 +490,18 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+        var controlByteSlice = controlBytes[...]
+
         for i in 1...self.maxMessagesPerRead {
             guard self.isOpen else {
                 throw ChannelError.eof
             }
             buffer.clear()
 
             let result = try buffer.withMutableWritePointer {
-                try self.socket.recvfrom(pointer: $0, storage: &rawAddress, storageLen: &rawAddressLength)
```
PeterAdams-A, line 500: Should we delete the now unused recvfrom method?

Lukasa:
Feedback addressed. Deleting `recvfrom` becomes a pretty large diff, so asking for a re-review from Peter.

PR #1475, opened 2020-04-08
Draft - Shorthand way of setting options in server bootstrap.

PeterAdams-A:
Motivation:

Less typing is good.

Modifications:

Add a new collection of types to provide shorthand channel options.
A new method on ServerBootstrap to set an array of these shorthand options.
A test case that long hand does the same as short hand.
Change to HTTPServer to demonstrate the change.

Result:

There will be an new way of setting options involving less typing.

_[One line description of your change]_

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_

PR #1489, opened 2020-04-23
Allow bootstraps to override TCP shorthand options from universal.

PeterAdams-A:
Allow bootstraps to override TCP shorthand options from universal.

### Motivation:

Some underlying bootstraps may know how to interpret the common shorthand options in a better way than translating to the default socket options.

### Modifications:

Add a method and default implementation to NIOClientTCPBootstrapProtocol to optionally interpret options.  Default implementation always reports failure.
Add a test to prove it is possible to consume options in this way and that the default is still to arrive as channel options.

### Result:

It is now possible to override interpretation of shorthand channel options when applied through universal bootstrap.

### DIFF
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -1050,7 +1050,7 @@ extension ShorthandServerBootstrapOption {
     
     /// `BacklogOption` allows users to configure the `backlog` value as specified in `man 2 listen`. This is only useful for `ServerSocketChannel`s.
     /// - See: ChannelOptions.backlog
-    public static func backlog(_ value : ChannelOptions.Types.BacklogOption.Value) -> ShorthandServerBootstrapOption {
+    public static func backlog(_ value: ChannelOptions.Types.BacklogOption.Value) -> ShorthandServerBootstrapOption {
         return ShorthandServerBootstrapOption(.backlog(value))
     }
 }
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -494,7 +494,7 @@ extension ServerBootstrap.ShorthandChildBootstrapOption {
     /// - See: `AllowRemoteHalfClosureOption`.
     public static func allowRemoteHalfClosure(_ value: Bool) ->
         ServerBootstrap.ShorthandChildBootstrapOption {
-        ServerBootstrap.ShorthandChildBootstrapOption(.allowRemoteHalfClosure(value))
+        return ServerBootstrap.ShorthandChildBootstrapOption(.allowRemoteHalfClosure(value))
     }
 }
 
@@ -843,7 +843,7 @@ extension ClientBootstrap.ShorthandClientBootstrapOption {
     
     /// - See: `AllowRemoteHalfClosureOption`.
     public static func allowRemoteHalfClosure(_ value: Bool) -> ClientBootstrap.ShorthandClientBootstrapOption {
-        ClientBootstrap.ShorthandClientBootstrapOption(.allowRemoteHalfClosure(value))
+        return ClientBootstrap.ShorthandClientBootstrapOption(.allowRemoteHalfClosure(value))
     }
 }
 
@@ -1281,6 +1281,6 @@ extension NIOPipeBootstrap.ShorthandChannelBootstrapOption {
     /// - See: `AllowRemoteHalfClosureOption`.
     public static func allowRemoteHalfClosure(_ value: Bool) ->
         NIOPipeBootstrap.ShorthandChannelBootstrapOption {
-        NIOPipeBootstrap.ShorthandChannelBootstrapOption(.allowRemoteHalfClosure(value))
+        return NIOPipeBootstrap.ShorthandChannelBootstrapOption(.allowRemoteHalfClosure(value))
     }
 }
--- a/Sources/NIO/Bootstrap.swift
+++ b/Sources/NIO/Bootstrap.swift
@@ -179,6 +179,7 @@ public final class ServerBootstrap {
     /// - See: serverChannelOption
     /// - Parameter options: List of shorthand options to apply.
     /// - Returns: The updated server bootstrap (`self` being mutated)
+    @inlinable
     public func serverOptions(_ options: [ShorthandServerBootstrapOption]) -> Self {
         for option in options {
             option.applyOption(to: self)
@@ -201,6 +202,7 @@ public final class ServerBootstrap {
     /// - See: childChannelOption
     /// - Parameter options: List of shorthand options to apply.
     /// - Returns: The update server bootstrap (`self` being mutated)
+    @inlinable
     public func childChannelOptions(_ options: [ShorthandChildBootstrapOption]) -> Self {
         for option in options {
             option.applyOption(to: self)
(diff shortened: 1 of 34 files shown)


Sources/NIO/Bootstrap.swift
```diff
@@ -175,6 +174,18 @@ public final class ServerBootstrap {
         self._serverChannelOptions.append(key: option, value: value)
         return self
     }
+    
+    /// Specifies some `ChannelOption`s to be applied to the `ServerSocketChannel`.
+    /// - See: serverChannelOption
+    /// - Parameter options: List of shorthand options to apply.
+    /// - Returns: The updated server bootstrap (`self` being mutated)
+    @inlinable
+    public func serverOptions(_ options: [ServerOption]) -> Self {
```
weissi, line 183: there are two API discrepancies here:

1. it's `serverOptions` vs. `childChannelOptions` (one has `...Channel...` the other one doesn't)
2. `ServerOption` seems `ServerBootstrap` specific (what about `NIOTSListenerBootstrap`) vs `NIOTCPShorthandOption` which is more generic

Sources/NIO/Bootstrap.swift
```diff
@@ -828,6 +935,54 @@ public final class DatagramBootstrap {
...
+                    _ = bootstrap.channelOption(ChannelOptions.autoRead, value: false)
+                
+                }
+            }
+        }
+    }
+}
+
+// Hashable for the convenience of users.
+extension DatagramBootstrap.Option: Hashable {}
```
weissi, line 973: why is it `DatagramBootstrap.Option` for UDP but `NIOTCPShorthandOption` for TCP?

weissi: what about NIOTS's upcoming UDP support? https://github.com/apple/swift-nio-transport-services/pull/68

Sources/NIO/Bootstrap.swift
```diff
@@ -966,4 +1133,134 @@ public final class NIOPipeBootstrap {
...
+        return NIOPipeBootstrap.Option(.allowRemoteHalfClosure(value))
+    }
+}
+
+
+
+// ------------------------
+
+public protocol NIOTCPOptionAppliable {
+    func applyOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> Self
```
weissi, line 1195: @Lukasa / @glbrntt we need a good name for this. This is just a _secondary_ way to set good old `ChannelOption`s. Usually they are set using `channelOption` and `clientChannelOption`. This function duplicates what they're doing. I think the actual purpose is to apply an option to the TCP bit?

So maybe `tcpChannelOption`? What about the protocol name?

Sources/NIO/Bootstrap.swift
```diff
@@ -966,4 +1133,134 @@ public final class NIOPipeBootstrap {
             }
         }
     }
+    
+    /// A channel option which can be applied to pipe bootstrap using shorthand notation.
+    /// - See: NIOPipeBootstrap.channelOptions(_ options: [Option])
+    public struct Option {
```
weissi, line 1139: Not sure what @Lukasa / @glbrntt think but should we move all the new stuff into a separate file? Like `UniversalBootstrapSupport.swift` we could have `ConvenienceOptionSupport.swift`?

Lukasa: I think a separate file would be a good idea.

Sources/NIO/Bootstrap.swift
Lukasa: I mean, they do apply to TCP channels. Maybe `tcpChannelOption` is fine? If we use that then `NIOTCPChannelOptionAppliable` could work, though it's a gross name. I guess most users will never have to deal with the name though.

Sources/NIO/Bootstrap.swift
```diff
@@ -966,4 +1133,134 @@ public final class NIOPipeBootstrap {
...
+    private let data: ShorthandOption
+    
+    private init(_ data: ShorthandOption) {
+        self.data = data
+    }
+    
+    /// Apply the contained option to the supplied object (almost certainly bootstrap) using the default mapping.
+    /// - Parameter to: object to apply this option to.
+    /// - Returns: the modified object
+    public func applyOption<T : NIOTCPOptionAppliable>(with: T) -> T {
```
weissi, line 1222: `T` isn't a great generic parameter name it's it doesn't apply to literally anything. Could be `<Bootstrap: ...>` so it's clear what it is. Then you can also drop the `with:` label

Sources/NIO/Bootstrap.swift
```diff
@@ -966,4 +1133,134 @@ public final class NIOPipeBootstrap {
...
+    
+    fileprivate enum ShorthandOption {
+        case reuseAddr
+        case disableAutoRead
+        case allowRemoteHalfClosure(Bool)
+        
+        func applyOption<T : NIOTCPOptionAppliable>(with: T) -> T {
+            switch self {
+            case .reuseAddr:
+                return with.applyOption(ChannelOptions.socketOption(.reuseaddr), value: 1)
```
weissi, line 1234: doesn't this _always_ map `.allowImmediateEndpointAddressReuse` to `.socketOption`?

Wasn't the idea that NIOTS can _not_ go via the socket API (because it doesn't have sockets?). Socket API is cool as the (inevitable) fallback but for NIOTS we don't need the fallback because we can add the support.

Maybe I'm missing something?

Sources/NIO/Bootstrap.swift
```diff
@@ -364,6 +388,65 @@ public final class ServerBootstrap {
...
+}
+
+// Hashable for the convenience of users.
+extension ServerBootstrap.ServerOption: Hashable {}
+extension ServerBootstrap.ServerOption.ShorthandServerOption: Hashable {}
+
+/// Approved shorthand server options.
+extension ServerBootstrap.ServerOption {
+    /// Option to reuse address.
+    /// - See:  NIOBSDSocket.Option.reuseaddr
```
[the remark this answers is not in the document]
PeterAdams-A: Made all docs sound more authoritative.

Sources/NIO/Bootstrap.swift
PeterAdams-A: OK, I have pulled all the option types out to be top level classes so they can be reused in any other implementations which might exist in the future.  Added Channel to name.

Sources/NIO/Bootstrap.swift
PeterAdams-A: Renamed to top level NIOUDPShorthandOption

Sources/NIO/Bootstrap.swift
PeterAdams-A: Rethought a little as  NIOChannelOptionAppliable and used more like a closure.

Sources/NIO/Bootstrap.swift
PeterAdams-A: Now all in separate file.

Sources/NIO/Bootstrap.swift
PeterAdams-A: Changed everything to be 
applyOptionDefaultMapping<OptionApplier : NIOChannelOptionAppliable>

Sources/NIO/Bootstrap.swift
PeterAdams-A: NIOTS will override by implementing NIOClientTCPBootstrapProtocol.applyChannelOption.
I've renamed function to be called applyOptionDefaultMapping

PeterAdams-A:
Combining into https://github.com/apple/swift-nio/pull/1475

Issue #1487, opened 2020-04-21
Update the Resolver interface


Lukasa:
NIO has spent the majority of its lifetime with a pluggable DNS resolver via the `Resolver` protocol:

https://github.com/apple/swift-nio/blob/197664726f402a90c69d9e2fbba885fc76fa6001/Sources/NIO/Resolver.swift#L16-L49

This resolver is a vital part of our connection establishment logic which uses Happy Eyeballs v2 to race outbound TCP connections against each other. The goal is to allow an arbitrary DNS resolver to inform NIO of DNS resolution asynchronously, which we will use to manage racing outbound TCP connections. The forward-looking goal was to allow NIO to be easily updated to use asynchronous DNS where available.

This `Resolver` interface bakes has two major limitations. The first is that it bakes in a strong knowledge of DNS: it's phrased in terms of A/AAAA. The second is that it does not allow result streaming: there are only two opportunities to deliver results.

These limitations manage to constrain the current `Resolver` interface both in terms of how good it is for DNS _and_ how general it can be. More general service discovery use-cases have to awkwardly contort themselves into the shape of DNS. However, truly streaming DNS resolvers such as dns-sd are required to force themselves to give all-or-nothing answers.

We should update the `Resolver` interface to be substantially more general. Here's what the shape would likely be:

```swift
protocol NIOStreamingResolver {
    func resolve(name: String, destinationPort: Int, session: NIONameResolutionSession)

    func cancel(_ session: NIONameResolutionSession)
}

class NIONameResolutionSession {
    var hints: Hints

    func deliverResults(_: [SocketAddress])

    func resolutionComplete(_: Result<Void, Error>)

    struct Hints {
        var localAddress: SocketAddress?
    }
}
```

There are three parts to this new interface. The first is the `NIOStreamingResolver` protocol. This is the protocol that new resolvers will adopt. It is now phrased entirely in terms of resolving _names_ to socket addresses, without any implication that DNS is the only way to do this. This unlocks powerful new tools such as the ability to plug in arbitrary service discovery tools into the resolver interface. For example, you could teach NIO to use Redis for service discovery, and the Happy Eyeballs implementation would seamlessly transition to using results from Redis. The destination port is still provided because it is necessary to properly construct a `SocketAddress`, and it allows the resolver to potentially override that port choice if needed.

[...]

PR #1490, opened 2020-04-23
ClientBootstrap: allow binding sockets

### LINKED ISSUE #1484: Binding ClientBootstrap to a specific Interface
### Expected behavior
_I want to bind a client socket to a particular interface._

### Actual behavior
_Currently ClientBootstrap doesn't provide any options to do so._

### Steps to reproduce

1. Create a ClientBootstrap object
2. try to bind the channel in channelInitializer

### SwiftNIO version/commit hash : 2.16.0
Tag: _2.16.0_
Commit hash: _e876fb37410e0036b98b5361bb18e6854739572b_

```
Apple Swift version 5.2 (swiftlang-1103.0.32.1 clang-1103.0.32.29)
Target: x86_64-apple-darwin19.4.0
Darwin Bikramjeets-MacBook-Pro.local 19.4.0 Darwin Kernel Version 19.4.0: Wed Mar  4 22:28:40 PST 2020; root:xnu-6153.101.6~15/RELEASE_X86_64 x86_64
```

weissi:
### Motivation:

It can be useful to `bind` client sockets as well but previously NIO wasn't allowing you to do that.

### Modifications:

Add `ClientBootstrap.bind` which allows you to `bind` sockets.

### Result:

- fixes #1484

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -337,7 +337,7 @@ class BaseSocket: BaseSocketProtocol {
     ///     - name: The name of the option to set.
     ///     - value: The value for the option.
     /// - throws: An `IOError` if the operation failed.
-    final func setOption<T>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option, value: T) throws {
+    func setOption<T>(level: NIOBSDSocket.OptionLevel, name: NIOBSDSocket.Option, value: T) throws {
         if level == .tcp && name == .tcp_nodelay && (try? self.localAddress().protocol) == Optional<NIOBSDSocket.ProtocolFamily>.some(.unix) {
             // setting TCP_NODELAY on UNIX domain sockets will fail. Previously we had a bug where we would ignore
             // most socket options settings so for the time being we'll just ignore this. Let's revisit for NIO 2.0.
@@ -387,7 +387,7 @@ class BaseSocket: BaseSocketProtocol {
     /// - parameters:
     ///     - address: The `SocketAddress` to which the socket should be bound.
     /// - throws: An `IOError` if the operation failed.
-    final func bind(to address: SocketAddress) throws {
+    func bind(to address: SocketAddress) throws {
         try self.withUnsafeHandle { fd in
             func doBind(ptr: UnsafePointer<sockaddr>, bytes: Int) throws {
                 try Posix.bind(descriptor: fd, ptr: ptr, bytes: bytes)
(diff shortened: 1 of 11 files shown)


weissi:
@Lukasa not 100% sure what you're proposing. Do you want this to wait until for #1487 ?

Lukasa:
I'm proposing just call `bind` in the channels constructed by the `HappyEyeballsConnector`. Yeah, a bunch of those binds will cause subsequent `connect`s to fail, but only until we do #1487. That lets us throw away _loads_ of code in this PR, that we'll only want to throw away when #1487 lands.

weissi:
@Lukasa won't that delay IPv4 a little bit?

weissi:
@Lukasa hang on, also the `bind` will fail because we'd try to bind the same address/port/family

Lukasa:
Ugh damnit. Yeah, the only way around this is to modify this so that instead of binding a local address you specify a local interface. Not an easy path to go down.

weissi:
> If the user wants to bind a specific address _and port_, then there are very limited options for us. We can try to connect sequentially, but right now if they do that they will get the same behaviour as SadEyeballs (albeit with more work).

@Lukasa  Ah, so the idea is that say we bind to `127.0.0.1:11111` but connect to `localhost:22222`. Then HE may first give us a `Channel` to connect to `[::1]:22222` which we would then try to `bind` to `127.0.0.1:11111`. This would result in `EAFNOSUPPORT` (`Address family not supported by protocol family`) and we'd stop. Subsequently, HE would make us make a channel for `127.0.0.1:22222` which again we'd bind to `127.0.0.1:11111` which then could succeed.

Is that what you're thinking? I think I could buy into this. The delay that HE gives us for IPv4 could be slightly annoying here but I think that's okay because most users who want to `bind` would connect to an IP anyway, so there won't be any resolving going on and therefore no HE either.

Lukasa:
Yes, that’s exactly what I’m thinking.

weissi:
@Lukasa let's do it!

Sources/NIO/SadEyeballs.swift
```diff
@@ -0,0 +1,119 @@
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1: Didn't we kill this whole file?

weissi: whooooops, yes

weissi: great catch, fixed

PR #1493, opened 2020-04-28
Make some vars lets.

### LINKED ISSUE #1492: BaseSocket.swift contains var that could be let
Swift 5.3 now warns on:

https://github.com/apple/swift-nio/blob/197664726f402a90c69d9e2fbba885fc76fa6001/Sources/NIO/BaseSocket.swift#L368

This is a `var` that is not mutated, and so can be a `let`. This never used to be caught, but it is now, and breaks downstream test cases (see apple/swift-nio-http2#197).

Lukasa:
Motivation:

Newer Swift 5.3 snapshots have spotted more vars that could be lets.

Modifications:

Made two vars lets.

Result:

The word `let` appears more frequently; the word `var`, less so.

Fixes #1492


weissi:

PR #1496, opened 2020-04-29
UDP based allocation tests

PeterAdams-A:
UDP based allocation tests

### Motivation:

With future work being considered in the UDP space it will be useful to have a baseline measure of allocations.

### Modifications:

New allocation tests for UDP - 
Ideas inspired by the TCP allocation tests. 
Code a modified version of UDPEchoServer and UDPEchoClient examples.

### Result:

There are UDP based allocation tests.


IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,133 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+        public typealias InboundIn = AddressedEnvelope<ByteBuffer>
+        public typealias OutboundOut = AddressedEnvelope<ByteBuffer>
+        private var numBytesOutstanding = 0
+        private var repetitionsRemaining : Int
+        
+        private let remoteAddressInitializer: () throws -> SocketAddress
+        
+        init(remoteAddressInitializer: @escaping () throws -> SocketAddress,
+             numberOfRepetitions: Int) {
+            self.remoteAddressInitializer = remoteAddressInitializer
```
Lukasa, line 180: Why do we delay this?

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,133 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            }
+        }
+        
+        public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+            let envelope = self.unwrapInboundIn(data)
+            let byteBuffer = envelope.data
+            
+            self.numBytesOutstanding -= byteBuffer.readableBytes
+            
+            if self.numBytesOutstanding <= 0 {
```
Lukasa, line 224: We should probably remove this check: a datagram-oriented transport should expect one read per write, and not necessarily assume that all the bytes it wrote will come back.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,133 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            context.close(promise: nil)
+        }
+    }
+
+    static func doUdpRequests(group: EventLoopGroup, number numberOfRequests: Int) throws -> Int {
+        let serverChannel = try DatagramBootstrap(group: group)
+            .channelOption(ChannelOptions.socketOption(.reuseaddr), value: 1)
+            // Set the handlers that are applied to the bound channel
+            .channelInitializer { channel in
+                // Ensure we don't read faster than we can write by adding the BackPressureHandler into the pipeline.
```
Lukasa, line 244: Probably worth deleting this comment.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,133 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+                return channel.pipeline.addHandler(EchoHandler())
+            }
+            .bind(to: localhostPickPort).wait()
+
+        defer {
+            try! serverChannel.close().wait()
+        }
+        
+        let remoteAddress = { () -> SocketAddress in serverChannel.localAddress! }
```
Lukasa, line 253: Why is this a closure, instead of just passing it directly?

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: I just copied the logic from the existing example - ie, I don't know.
I could try without and see what happens.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: As above - cloned from existing example - will try without.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Made direct and it worked just fine.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Done.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Made direct and it worked.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
     }
     #endif
 }
+
+// MARK: UDP Shared
+
+/// Repeatedly send, echo and receive a number of UDP requests.
+/// Based on the UDPEchoClient/Server example.
+enum UdpShared {
```
weissi, line 144: nit: in Swift, abbreviations go all caps (or all lowercase, depending on where they are). This would be `UDPShared`

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+        }
+
+        public func channelReadComplete(context: ChannelHandlerContext) {
+            // As we are not really interested getting notified on success or failure we just pass nil as promise to
+            // reduce allocations.
+            context.flush()
+        }
+
+        public func errorCaught(context: ChannelHandlerContext, error: Error) {
+            print("error: ", error)
```
weissi, line 162: do we really expect errors? If not, we should just crash on error. `print` would also allocate loads

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            self.repetitionsRemaining = numberOfRepetitions
+        }
+        
+        public func channelActive(context: ChannelHandlerContext) {
+            // Channel is available. It's time to send the message to the server to initialize the ping-pong sequence.
+            sendSomeDataIfDesiredOrClose(context: context)
+        }
+        
+        private func sendSomeDataIfDesiredOrClose(context: ChannelHandlerContext) {
+            if (repetitionsRemaining > 0) {
```
weissi, line 189: nit: extra `(`s

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+                // Set the transmission data.
+                let line = "Something to send there and back again."
+                var buffer = context.channel.allocator.buffer(capacity: line.utf8.count)
+                buffer.writeString(line)
+                
+                // Forward the data.
+                let envolope = AddressedEnvelope<ByteBuffer>(remoteAddress: remoteAddress, data: buffer)
+                
+                context.writeAndFlush(self.wrapOutboundOut(envolope), promise: nil)
+            }
```
weissi, line 201: nit: `else` fell off the `}`

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            else {
+                // We're all done - hurrah!
+                context.close(promise: nil)
+            }
+        }
+        
+        public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+            let envelope = self.unwrapInboundIn(data)
+            _ = envelope.data
```
weissi, line 210: line doesn't doesn't do anything

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            
+            // Got back a response - maybe send some more.
+            sendSomeDataIfDesiredOrClose(context: context)
```
weissi, line 213: missing `self.`

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+        }
+        
+        public func errorCaught(context: ChannelHandlerContext, error: Error) {
+            print("error: ", error)
```
weissi, line 217: same question regarding errors here

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,119 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+            
+            // As we are not really interested getting notified on success or failure we just pass nil as promise to
+            // reduce allocations.
+            context.close(promise: nil)
+        }
+    }
+
+    static func doUdpRequests(group: EventLoopGroup, number numberOfRequests: Int) throws -> Int {
```
weissi, line 225: nit: `doUDPRequests`

IntegrationTests/tests_04_performance/test_01_resources/test_udp_1000_reqs_1_conn.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+func run(identifier: String) {
+    measure(identifier: identifier) {
+        let numberDone = try! UdpShared.doUdpRequests(group: group, number: 1000)
```
weissi, line 19: I think you could pull the `Channel` setup out of the `measure` block and this would become much more stable. The TCP tests possibly get that wrong. When the alloc tests were first written this wouldn't have been possible :)

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Ok, will go capitals.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Nope.  Will change.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Will kill.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Will fix - Spent 25 years avoiding KnR bracketing.  Hate it :-(

weissi: haha, this will be all over once the Swift community decided on the "official style". Then we can all type now we prefer and the formatter will make it "right".

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: will fatal.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: ok.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: ok.

IntegrationTests/tests_04_performance/test_01_resources/test_udp_1000_reqs_1_conn.swift
PeterAdams-A: As discussed.  If possible I'd like to keep everything in.  Will give some grace in the checked numbers to make it stable in error reporting at least.

docker/docker-compose.1604.52.yaml
```diff
@@ -33,6 +33,8 @@ services:
       - MAX_ALLOCS_ALLOWED_modifying_byte_buffer_view=2010 # 5.2 improvement 4000
       - MAX_ALLOCS_ALLOWED_schedule_10000_tasks=90050
       - MAX_ALLOCS_ALLOWED_read_10000_chunks_from_file=210050
+      - MAX_ALLOCS_ALLOWED_udp_1000_reqs_1_conn=16211
+      - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=226000
```
weissi, line 37: Oh, do we understand why 5.2 is worse than 5.1? Seems to be 3 allocations worse in 5.2, we should analyse why that is and either fix it on our side or file a Swift bug. Newer Swifts should only make it better :).

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: Changed.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: done.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: added.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: gone to fatal.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: done.

docker/docker-compose.1604.52.yaml
PeterAdams-A: Instability as noted below could be the cause.  Could well be a swift bug but that's a separate issue.

IntegrationTests/tests_04_performance/test_01_resources/test_udp_1000_reqs_1_conn.swift
PeterAdams-A: Decided to leave it in there for now but give a little grace in numbers.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: killed both.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: fixed.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: now fatal.

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
PeterAdams-A: It is gone.

Lukasa:

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,108 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
     #endif
 }
+
+// MARK: UDP Shared
+
+/// Repeatedly send, echo and receive a number of UDP requests.
+/// Based on the UDPEchoClient/Server example.
+enum UDPShared {
+    private final class EchoHandler: ChannelInboundHandler {
+        public typealias InboundIn = ByteBuffer
```
Lukasa, line 146:
```suggestion
        public typealias InboundIn = AddressedEnvelope<ByteBuffer>
```

IntegrationTests/tests_04_performance/test_01_resources/shared.swift
```diff
@@ -136,3 +136,108 @@ func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+        public typealias OutboundOut = ByteBuffer
```
Lukasa, line 147:
```suggestion
        public typealias OutboundOut = AddressedEnvelope<ByteBuffer>
```

IntegrationTests/tests_04_performance/test_01_resources/test_udp_1000_reqs_1_conn.swift
PeterAdams-A: Grace seems to make things stable.

PeterAdams-A:
Re memory regression 5.1 -> 5.2.

I believe this is a known (to the nio team) issue - see https://github.com/PeterAdams-A/swift-nio/blob/197664726f402a90c69d9e2fbba885fc76fa6001/docker/docker-compose.1604.52.yaml#L21

It is definitely related to channel options - if I comment out the two lines which are setting channel options the allocations become the same between 5.1 and 5.2

The dtrace output is a nightmare to read due to different inlining etc but I think I've managed to pair all significant allocations except the following.

# Different.
AFTER >> KEY:                test_udp_1_reqs_1000_conn`partial apply for thunk for @escaping @callee_guaranteed (@guaranteed Channel) -> (@owned @escaping @callee_guaranteed (@in_guaranteed Any, @in_guaranteed Any) -> (@owned EventLoopFuture<()>))
AFTER >> 22000
BEFOR >> KEY:                test_udp_1_reqs_1000_conn`specialized thunk for @escaping @callee_guaranteed (@guaranteed Channel) -> (@owned @escaping @callee_guaranteed (@in_guaranteed Any, @in_guaranteed Any) -> (@owned EventLoopFuture<()>))
BEFOR >> 44000

### only in AFTER
AFTER >> KEY:                test_udp_1_reqs_1000_conn`specialized ChannelOptions.Storage.append<A>(key:value:)
AFTER >> 22000
AFTER >> KEY:                test_udp_1_reqs_1000_conn`specialized applier #1 <A>(_:) in ChannelOptions.Storage.append<A>(key:value:)
AFTER >> 22000

docker/docker-compose.1804.51.yaml
```diff
@@ -34,6 +34,9 @@ services:
       - MAX_ALLOCS_ALLOWED_modifying_byte_buffer_view=6010
       - MAX_ALLOCS_ALLOWED_schedule_10000_tasks=90050
       - MAX_ALLOCS_ALLOWED_read_10000_chunks_from_file=210050
+      - MAX_ALLOCS_udp_1000_reqs_1_conn=16250
+      - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=224050
+      - SANITIZER_ARG=--sanitize=thread
```
Davidde94, line 39: Remove this sanitizer thread and everything should pass :)

PeterAdams-A: Thanks - I messed up the merge.

PR #1499, opened 2020-05-01
allow completely single-threaded NIO programs

weissi:
Motivation:

Usually, you spawn a new MultiThreadedEventLoopGroup to run NIO
programs. But if you're a very simple command line utility, that may not
be necessary, so why not just re-use the main thread?

Modifications:

Allow taking over threads and making them EventLoops.

Result:

Fewer threads if you want.


Sources/NIO/EventLoop.swift
```diff
@@ -914,6 +930,17 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
             }
         }
     }
+
+    public static func makeEventLoopFromCallingThread(_ callback: @escaping (EventLoop) -> Void) {
```
Lukasa, line 934: Can we document acceptable uses of this and explain what will happen? I think it's especially important to explain what the exit conditions are (i.e. that this call will block the thread until the event loop is shut down).

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -369,11 +371,17 @@ internal final class SelectableEventLoop: EventLoop {
             }
         }
         var nextReadyTask: ScheduledTask? = nil
+        self._tasksLock.withLock {
+            if let firstTask = self._scheduledTasks.peek() {
+                // We're only getting here if something was enqueued before `self.run` was called.
+                nextReadyTask = firstTask
+            }
+        }
```
Lukasa, line 379: Why do we need this?

weissi: @Lukasa I'll expand on the comment. But basically: If we enqueue a ready task before `run` is called, we'd block until the next task is enqueued from another thread. This just checks if there's a task already and then only blocks until the ready time for that task.

Technically, this was a race we always had. But given that it's SUPER unlikely that we would publish the `EventLoop` reference before `run` is called, it was impossible to hit :). But now, if you enqueue from the `callback` in `makeEventLoopFromCallingThread`, you can totally do that.

Sources/NIO/EventLoop.swift
weissi: whoops, forgot the docs.

Sources/NIO/EventLoop.swift
```diff
@@ -725,6 +725,30 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
...
+    private static func runTheLoop(thread: NIOThread,
+                                   canEventLoopBeShutdownIndividually: Bool,
+                                   selectorFactory: @escaping () throws -> NIO.Selector<NIORegistration>,
+                                   initializer: @escaping ThreadInitializer,
+                                   _ callback: @escaping (SelectableEventLoop) -> Void) {
+        assert(NIOThread.current == thread)
+        initializer(thread)
+
+        do {
+            /* we try! this as this must work (just setting up kqueue/epoll) or else there's not much we can do here */
```
glbrntt, line 737: Any reason to `try!` instead of relying on the `fatalError` below?

Sources/NIO/EventLoop.swift
glbrntt: It also feels weird calling this from MTELG. I wonder if making it internal here and surfacing it via some other namespace would be a better way to go: `<Something>.withCurrentThreadAsEventLoop { ... }`

Sources/NIO/EventLoop.swift
weissi: none, fixed. thanks

weissi:
@glbrntt / @Lukasa I agree that it's weird to call this on `MultiThreadedEventLoopGroup` but I didn't find a better place. I'm happy adding a new `NIOTakeOverEventLoop` namespace and nest it under there?

I felt it's not really worth adding a new top-level namespace for a function that'll be as rarely used as this one so I put it into MTELG and thought we can always (deprecate) move it elsewhere in the future. But 100% open to moving it.

Sources/NIO/EventLoop.swift
t089: How about naming it something like ˋmakeSingleThreadedGroupFromCallingThreadˋ?

Sources/NIO/EventLoop.swift
```diff
@@ -725,6 +725,31 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
...
+    private static func runTheLoop(thread: NIOThread,
+                                   canEventLoopBeShutdownIndividually: Bool,
+                                   selectorFactory: @escaping () throws -> NIO.Selector<NIORegistration>,
+                                   initializer: @escaping ThreadInitializer,
+                                   _ callback: @escaping (SelectableEventLoop) -> Void) {
+        assert(NIOThread.current == thread)
+        initializer(thread)
+
+        do {
+            let l = SelectableEventLoop(thread: thread,
```
Davidde94, line 737: Nit: `l` doesn't seem like a great name IMHO, could we at least go for `loop` or something?

Davidde94:

tomerd:
this is great @weissi, can be very useful in one of our projects. thanks!

Sources/NIO/EventLoop.swift
glbrntt: > I felt it's not really worth adding a new top-level namespace for a function that'll be as rarely used as this one so I put it into MTELG and thought we can always (deprecate) move it elsewhere in the future. But 100% open to moving it.

Yeah I think that's fair: I'm fine with leaving it where it is. Although I do think `with` is a better prefix than `make` here. Make – at least in my mind – suggests returning something to the caller, with suggests it's providing an object scoped to the callback.

weissi: @glbrntt yes, `with` over `make` is right! `make` is a factory.

@Lukasa happy with leaving it in MTELG?

Lukasa: Yup.

weissi: @glbrntt done & thanks Cory

Sources/NIO/EventLoop.swift
weissi: fixed.

weissi: (was just moving code but should still use better variable names)

Issue #1445, opened 2020-03-13
flaky test EventLoopTest.testScheduleRepeatedTaskCancelFromDifferentThread


weissi:
this looks like a race in the recent change

```
15:15:30 Test Case 'EventLoopTest.testScheduleRepeatedTaskCancelFromDifferentThread' started at 2020-03-13 15:15:30.460
15:15:30 Precondition failed: cannot schedule tasks on an EventLoop that has already shut down: file /code/Sources/NIO/SelectableEventLoop.swift, line 264
15:15:30 Current stack trace:
15:15:30 0    libswiftCore.so                    0x00007f579844432b <unavailable> + 3953451
15:15:30 1    libswiftCore.so                    0x00007f57984443a0 swift_reportError + 50
15:15:30 2    libswiftCore.so                    0x00007f57984b5090 _swift_stdlib_reportFatalErrorInFile + 115
15:15:30 3    libswiftCore.so                    0x00007f57983d94fe <unavailable> + 3515646
15:15:30 4    libswiftCore.so                    0x00007f57983d9677 <unavailable> + 3516023
15:15:30 5    libswiftCore.so                    0x00007f57981ceddd <unavailable> + 1375709
15:15:30 6    libswiftCore.so                    0x00007f57983b0458 <unavailable> + 3347544
15:15:30 7    libswiftCore.so                    0x00007f57981ce179 <unavailable> + 1372537
15:15:30 8    swift-nioPackageTests.xctest       0x0000563f2cd9050a <unavailable> + 3601674
15:15:30 9    swift-nioPackageTests.xctest       0x0000563f2cd9839c <unavailable> + 3634076
15:15:30 10   swift-nioPackageTests.xctest       0x0000563f2cbc7960 <unavailable> + 1730912
15:15:30 11   swift-nioPackageTests.xctest       0x0000563f2ce39da8 <unavailable> + 4296104
15:15:30 12   swift-nioPackageTests.xctest       0x0000563f2ce39a20 <unavailable> + 4295200
15:15:30 13   swift-nioPackageTests.xctest       0x0000563f2ce39c0f <unavailable> + 4295695
15:15:30 14   swift-nioPackageTests.xctest       0x0000563f2cd8fd52 <unavailable> + 3599698
15:15:30 15   swift-nioPackageTests.xctest       0x0000563f2cd90101 <unavailable> + 3600641
[...]

PR #1500, opened 2020-05-01
Create new write repeating byte method

### LINKED ISSUE #1495: Add support for writing repeated bytes into a byte buffer.
I was working on a unit test where I needed `2**24` bytes in a byte buffer. Writing this the naive way is: `buffer.writeBytes(repeatElement(0, count: 1 << 24))`. This took around 16s. I replaced this with an unsafe pointer call to `memset` and it took more like 8ms.

We could solve this with a `writeRepeatedByte(_:count:)` that hides the unsafe code.

Davidde94:
Resolves #1495 

Add a function to support writing an arbitrary number of identical bytes.

### Motivation:

Sometimes it's necessary to do this, see #1495 

### Modifications:

Add a function that allocates `n` bytes, and uses `memset` to write a provided `UInt8` value to each byte. Also added a test.

### Result:

It's possible to quickly write `n` identical bytes to a buffer.
`buffer.writeRepeatingByte(9, count: 1 << 100)` (if you have enough memory)


Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,22 @@ extension ByteBuffer {
...
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    @inlinable
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
+        let pointer = malloc(count)
+        memset(pointer!, Int32(byte), count)
```
weissi, line 386: You’re allocating memory here, then setting then bytes, and then copying them all into the target buffer. We can skip the first and the third operation (allocation & copy) and write straight into the target.

Check out [`withUnsafeWritableBytes(minimumWritableBytes:)`](https://apple.github.io/swift-nio/docs/current/NIO/Structs/ByteBuffer.html#/s:3NIO10ByteBufferV27writeWithUnsafeMutableBytes015minimumWritableH0_S2i_SiSwKXEtKF)
M

weissi: Actually, [`writeWithUnsafeWritableBytes(minimumWritableBytes:)`](https://apple.github.io/swift-nio/docs/current/NIO/Structs/ByteBuffer.html#/s:3NIO10ByteBufferV27writeWithUnsafeMutableBytes015minimumWritableH0_S2i_SiSwKXEtKF) is even better, that moves the writer index for you too.

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,22 @@ extension ByteBuffer {
...
     }
+    
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    @inlinable
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
```
weissi, line 384: We also need a precondition that this is `>= 0`

Sources/NIO/ByteBuffer-aux.swift
Davidde94: Thanks a lot - I was looking for a function like that but clearly missed it

Davidde94: As for the leak - oops, just idiocy on my part

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,21 @@ extension ByteBuffer {
...
         return written
     }
+    
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    @inlinable
```
weissi, line 383: can we take off this `@inlinable`? The parameters here are `UInt8` and `Int`, if anything, inlining this would make it worse.

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,21 @@ extension ByteBuffer {
...
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
+        precondition(count >= 0, "Can't write less than 0 bytes")
+        return self.writeWithUnsafeMutableBytes(minimumWritableBytes: count) { (_pointer) -> Int in
```
weissi, line 386: - extra set of parens
- `_pointer` is underscored, should be `ptr` or `pointer`

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -2579,6 +2579,21 @@ class ByteBufferTest: XCTestCase {
         self.buf.moveWriterIndex(forwardBy: 5)
         XCTAssertEqual("hellohello", self.buf.readString(length: 10))
     }
+    
+    func testRepeatingBytes() {
+
```
weissi, line 2584: we could make this test more readable by writing:

```
func testRepeatingBytes() {
    let written = self.buf.writeRepeatingByte(9, count: 1024)
    XCTAssertEqual(1024, written)
    XCTAssertEqual(Array(repeating: UInt8(9), count: 1024), Array(self.buf.readableBytesView))
}
```

- we should also add a test case that tests the edge condition (`count: 0`)

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,21 @@ extension ByteBuffer {
...
     }
+    
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    @inlinable
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
```
weissi, line 384: usually when we have a `write`, we also have a corresponding `set` method

Tests/NIOTests/ByteBufferTest.swift
Davidde94: Looks good, I'll also bump the number of bytes up I think

Sources/NIO/ByteBuffer-aux.swift
Davidde94: Extra set of parens?

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
...
     }
+    
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
+        precondition(count >= 0, "Can't write less than 0 bytes")
```
Lukasa, line 384: I mean, it couldn’t possibly matter less, but...

```suggestion
        precondition(count >= 0, "Can't write fewer than 0 bytes")
```

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
...
+    }
+    
+    /// Sets the given `byte` `repeated` times at the given `index`. Will reserve more memory if necessary. Does not move the writer index.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func setRepeatingByte(_ byte: UInt8, count: Int, at index: Int) -> Int {
+        precondition(count >= 0, "Can't write less than 0 bytes")
```
Lukasa, line 398: Again, not that it matters in the slightest.

```suggestion
        precondition(count >= 0, "Can't write fewer than 0 bytes")
```

Lukasa, line 398: This method is missing a `isKnownUniquelyReferenced` check, so it doesn’t correctly maintain the value semantics of the type. Take a look at `_setBytes` for an example of what needs to be done here.

Lukasa: The fact that this is passing its tests also means that there are no tests to confirm that this function actually does maintain the value semantics of the type: mind adding some?

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
         self._moveWriterIndex(forwardBy: written)
         return written
     }
+    
+    /// Writes `byte` `repeated` times. Moves the writer index forward by the number of bytes written.
```
glbrntt, line 377:
```suggestion
    /// Writes `byte` `count` times. Moves the writer index forward by the number of bytes written.
```

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
...
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
+        precondition(count >= 0, "Can't write less than 0 bytes")
+        return self.writeWithUnsafeMutableBytes(minimumWritableBytes: count) { (pointer) -> Int in
```
glbrntt, line 385: Can this just be written in terms of `setRepeatingByte` to remove some duplication?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -2579,6 +2579,37 @@ class ByteBufferTest: XCTestCase {
...
+            let written = self.buf.writeRepeatingByte(9, count: count)
+            XCTAssertEqual(count, written)
+            XCTAssertEqual(Array(repeating: UInt8(9), count: count), Array(self.buf.readableBytesView))
+        }
+        
+        write(count: 1_000_000)
+        write(count: 0)
+    }
+    
+    func testSetBytes() {
```
glbrntt, line 2595: nit: this should including "repeating" (`setBytes` is its own thing)

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
...
+            memset(pointer.baseAddress!, Int32(byte), count)
+            return count
+        }
+    }
+    
+    /// Sets the given `byte` `repeated` times at the given `index`. Will reserve more memory if necessary. Does not move the writer index.
```
glbrntt, line 391:
```suggestion
    /// Sets the given `byte` `count` times at the given `index`. Will reserve more memory if necessary. Does not move the writer index.
```

Sources/NIO/ByteBuffer-aux.swift
Lukasa: This thread is not outdated: it still has not been fixed.

Davidde94: GH marked it as outdated automatically because there was another change.

Davidde94: (When it's done I'll mark as resolved)

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,39 @@ extension ByteBuffer {
...
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func setRepeatingByte(_ byte: UInt8, count: Int, at index: Int) -> Int {
+        precondition(count >= 0, "Can't write fewer than 0 bytes")
+        let bytesCount = count
+        let newEndIndex: _Index = UInt32(index) + _toIndex(bytesCount)
+        if !isKnownUniquelyReferenced(&self._storage) {
```
weissi, line 400: - We shouldn't do the CoW-thing manually here yet, this is too dangerous.
- Please use

```swift
self.reserveCapacity(index + count)
self.withVeryUnsafeMutableBytes { destCompleteStorage in
            assert(destCompleteStorage.count >= index + count)
            let dest = destCompleteStorage[index ..< index + count]
            memset(dest.baseAddress!, byte, count: count)
        }
        return count
}
```

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,37 @@ extension ByteBuffer {
...
     }
+    
+    /// Writes `byte` `count` times. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func writeRepeatingByte(_ byte: UInt8, count: Int) -> Int {
+        precondition(count >= 0, "Can't write fewer than 0 bytes")
```
weissi, line 384: kill this precondition please, it's repeated in `set` just below

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,37 @@ extension ByteBuffer {
...
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func setRepeatingByte(_ byte: UInt8, count: Int, at index: Int) -> Int {
+        precondition(count >= 0, "Can't write fewer than 0 bytes")
+        self.reserveCapacity(index + count)
+        self.withVeryUnsafeMutableBytes { pointer in
+            precondition(pointer.count >= count)
+            let dest = pointer[index ..< index + count].base
+            memset(dest.baseAddress!, Int32(byte), count)
+//            memset(<#T##__b: UnsafeMutableRawPointer!##UnsafeMutableRawPointer!#>, <#T##__c: Int32##Int32#>, <#T##__len: Int##Int#>)
```
weissi, line 403: leftover comment

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,35 @@ extension ByteBuffer {
...
+    ///
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func setRepeatingByte(_ byte: UInt8, count: Int, at index: Int) -> Int {
+        precondition(count >= 0, "Can't write fewer than 0 bytes")
+        self.reserveCapacity(index + count)
+        self.withVeryUnsafeMutableBytes { pointer in
+            precondition(pointer.count >= count)
```
weissi, line 399: this `precondition` is in the wrong place. It needs to be right in front of `memset` asserting the correct pointer has enough bytes left.

weissi: Probably something like

```swift
let dest = pointer[index...].base
preconditon(dest.count >= count)
memset(dest.baseAddress!, CInt(byte), size_t(count))
```

@Davidde94 also make sure to use the `CInt` (and not `Int32`) constructor as well as the `size_t` one. If you don't you're assuming the following things to be true:

- `CInt` == `Int32`
- `size_t` == `Int`

those are true for 64bit macOS but not on 32 bit systems usually.

weissi:
hit #1445 on 5.0

weissi:

Sources/NIO/ByteBuffer-aux.swift
```diff
@@ -373,6 +373,34 @@ extension ByteBuffer {
...
+    /// - parameter byte: The `UInt8` byte to repeat.
+    /// - parameter count: How many times to repeat the given `byte`
+    /// - returns: How many bytes were written.
+    @discardableResult
+    public mutating func setRepeatingByte(_ byte: UInt8, count: Int, at index: Int) -> Int {
+        precondition(count >= 0, "Can't write fewer than 0 bytes")
+        self.reserveCapacity(index + count)
+        self.withVeryUnsafeMutableBytes { pointer in
+            let dest = pointer.baseAddress?.advanced(by: index)
+            _ = dest!.initializeMemory(as: UInt8.self, repeating: byte, count: count)
```
weissi, line 400: can you move the `!` up to `pointer.baseAddress!`? Otherwise we may be doing extra work.

Davidde94: Done

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -249,10 +249,7 @@ final class DatagramChannelTests: XCTestCase {
             // will cause EMSGSIZE.
             let bufferSize = 1024 * 5
             var buffer = self.firstChannel.allocator.buffer(capacity: bufferSize)
-            buffer.writeWithUnsafeMutableBytes(minimumWritableBytes: bufferSize) {
-                _ = memset($0.baseAddress!, 4, $0.count)
-                return $0.count
-            }
+            buffer.writeRepeatingByte(4, count: buffer.capacity)
```
weissi, line 252: this isn't a correct translation of the old code. Should be `count: bufferSize`. `ByteBufferAllocator` may over-allocate.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -723,21 +723,16 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(otherBuf, buf)
 
         // Write to both buffers.
-        let firstResult = buf!.withUnsafeMutableWritableBytes { (ptr: UnsafeMutableRawBufferPointer) -> Bool in
-            XCTAssertEqual(cap, ptr.count)
-            memset(ptr.baseAddress!, 0, ptr.count)
-            return false
-        }
-        let secondResult = otherBuf!.withUnsafeMutableWritableBytes { (ptr: UnsafeMutableRawBufferPointer) -> Bool in
+        self.buf.setRepeatingByte(0, count: self.buf.capacity, at: 0)
```
weissi, line 726: - `at: 0` isn't right, that should be `buf.writerIndex`
- `count: self.buf.capacity` also isn't right, that should be `buf.capacity - buf.writerIndex`

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -723,21 +723,16 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(otherBuf, buf)
 
         // Write to both buffers.
-        let firstResult = buf!.withUnsafeMutableWritableBytes { (ptr: UnsafeMutableRawBufferPointer) -> Bool in
```
weissi, line 726: actually, sorry, this needs to be undone. This test is specifically testing `withUnsafeMutableWritableBytes`

Davidde94:
@glbrntt ping

PR #1501, opened 2020-05-02
UDPEchoClient to close channel after response received.

PeterAdams-A:
Close channel after UCPEchoClient received a response.

### Motivation:

The application does not shutdown without doing this.

### Modifications:

Close channel after receiving data.

### Result:

Application exits after receiving data.


Lukasa:

Issue #1472, opened 2020-04-07
Support propagating ECN information on `AddressedEnvelope`s


Lukasa:
It is possible that users of datagram channels may be interested in ECN information associated with the IP packets that carry the datagrams. It's possible to obtain this information from the OS and for us to propagate it through the channel, but today we do not.

The proposal for how to do so comes in multiple parts, each of which is probably a separate chunk of work:

- [ ] Add support for `AddressedEnvelope` to carry arbitrary metadata in a performant way

    This probably involves adding a metadata field to `AddressedEnvelope` that is usually empty, and only ever filled when necessary. This metadata will need to be keyed and typed. This is a straightforward Swift-level data structure problem, and requires no NIO-specific knowledge.

- [ ] Move datagram channels over to using `recvmsg` in the single-read case, instead of `recvfrom`. (#1473)

    This allows us to extract metadata from socket reads.

- [ ] Wire up extracting metadata of known types from `recvmsg`.

    Just a straightforward extension. We need to minimise allocations here, so we should aim to be careful about how we build this. Ideally we'll use the stack.

- [ ] Add a `ChannelOption` that allows users to request ECN information, and then make sure we extract and propagate that information.

- [ ] Update `swift-nio-transport-services` to support this new flag, as `Network.framework` automatically provides this information for us.

PeterAdams-A:
Solved for swift-nio iin #1587, #1590, #1591, #1593 , #1594 and #1596.

I've made an issue in transport services - https://github.com/apple/swift-nio-transport-services/issues/100 to track adding this support there.

ovidiuurotaru:
Can we also read ECN flags using NIO?

Lukasa:
Yes, this is implemented. Set `ChannelOption.explicitCongestionNotification` on your `DatagramBootstrap` to enable it.

PR #1502, opened 2020-05-04
Add Optional ECN MetaData to Addressed Envelope

PeterAdams-A:
Add Optional ECN MetaData to Addressed Envelope

### Motivation:

This is a very small step along the path to https://github.com/apple/swift-nio/issues/1472

### Modifications:

Add an optional metadata field to addressed envelope.  Currently the only possible data is ECN information.  This is not currently populated anywhere.

### Result:

Metadata in place for future work.

### DIFF
--- a/Sources/NIO/AddressedEnvelope.swift
+++ b/Sources/NIO/AddressedEnvelope.swift
@@ -20,7 +20,7 @@
 public struct AddressedEnvelope<DataType> {
     public var remoteAddress: SocketAddress
     public var data: DataType
-    public var metaData : MetaData? = .none
+    public var metaData : MetaData? = nil
 
     public init(remoteAddress: SocketAddress, data: DataType) {
         self.remoteAddress = remoteAddress
@@ -28,7 +28,7 @@ public struct AddressedEnvelope<DataType> {
     }
     
     public struct MetaData {
-        public var ecnEventInProgress : Bool
+        public var ecnEventInProgress : NIOEcnState
     }
 }
 
@@ -36,4 +36,16 @@ extension AddressedEnvelope: CustomStringConvertible {
     public var description: String {
         return "AddressedEnvelope { remoteAddress: \(self.remoteAddress), data: \(self.data) }"
     }
+}
+
+/// Possible ECN States
+public enum NIOEcnState {
+    /// Non-ECN Capable Transport.
+    case nonECT
+    /// ECN Capable Transport (flag 0).
+    case ect0
+    /// ECN Capable Transport (flag 1).
+    case ect1
+    /// Congestion Experienced.
+    case ce
 }
--- a/Sources/NIO/AddressedEnvelope.swift
+++ b/Sources/NIO/AddressedEnvelope.swift
@@ -20,14 +20,16 @@
 public struct AddressedEnvelope<DataType> {
     public var remoteAddress: SocketAddress
     public var data: DataType
-    internal var metaData : MetaData? = nil
+    /// Any MetaData associated with this `AddressedEnvelope`
+    public var metaData : MetaData? = nil
 
     public init(remoteAddress: SocketAddress, data: DataType) {
         self.remoteAddress = remoteAddress
         self.data = data
     }
     
-    internal struct MetaData {
+    /// Any MetaData associated with an AddressedEnvelope
+    public struct MetaData {
         /// Details of any congestion state.
         public var ecnState : NIOEcnState
     }
@@ -40,7 +42,7 @@ extension AddressedEnvelope: CustomStringConvertible {
 }
 
 /// Possible ECN States
-internal enum NIOEcnState {
+public enum NIOEcnState {
     /// Non-ECN Capable Transport.
     case nonECT
     /// ECN Capable Transport (flag 0).
--- a/Sources/NIO/AddressedEnvelope.swift
+++ b/Sources/NIO/AddressedEnvelope.swift
@@ -31,7 +31,7 @@ public struct AddressedEnvelope<DataType> {
     /// Any metadata associated with an `AddressedEnvelope`
     public struct Metadata {
         /// Details of any congestion state.
-        public var ecnState: NIOECNState
+        public var ecnState: NIOExplicitCongestionNotificationState
     }
 }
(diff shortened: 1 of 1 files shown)


Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
...
+    /// Any MetaData associated with this `AddressedEnvelope`
+    public var metaData : MetaData? = nil
 
     public init(remoteAddress: SocketAddress, data: DataType) {
         self.remoteAddress = remoteAddress
         self.data = data
     }
+    
+    /// Any MetaData associated with an AddressedEnvelope
+    public struct MetaData {
```
Lukasa, line 32: `Metadata` is one word, and so should be spelled...well, the way I just spelled it.

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
...
 }
 
 extension AddressedEnvelope: CustomStringConvertible {
     public var description: String {
         return "AddressedEnvelope { remoteAddress: \(self.remoteAddress), data: \(self.data) }"
     }
 }
+
+/// Possible ECN States
+public enum NIOEcnState {
```
Lukasa, line 45: As "ECN" is an acronym this should read `NIOECNState`.

Lukasa: Can this be conformed to `Hashable`?

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
 public struct AddressedEnvelope<DataType> {
     public var remoteAddress: SocketAddress
     public var data: DataType
+    /// Any MetaData associated with this `AddressedEnvelope`
+    public var metaData : MetaData? = nil
```
Lukasa, line 24: Similarly, this should be `metadata`.

PeterAdams-A: done.

Sources/NIO/AddressedEnvelope.swift
PeterAdams-A: done.

Sources/NIO/AddressedEnvelope.swift
PeterAdams-A: done.

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
 public struct AddressedEnvelope<DataType> {
     public var remoteAddress: SocketAddress
     public var data: DataType
+    /// Any metadata associated with this `AddressedEnvelope`
+    public var metadata : Metadata? = nil
```
Lukasa, line 24: Mild nit

```suggestion
    public var metadata: Metadata? = nil
```

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
...
 
     public init(remoteAddress: SocketAddress, data: DataType) {
         self.remoteAddress = remoteAddress
         self.data = data
     }
+    
+    /// Any metadata associated with an `AddressedEnvelope`
+    public struct Metadata {
+        /// Details of any congestion state.
+        public var ecnState : NIOECNState
```
Lukasa, line 34: Mild nit

```suggestion
        public var ecnState: NIOECNState
```

Sources/NIO/AddressedEnvelope.swift
glbrntt, line 24:
```suggestion
    public var metadata: Metadata? = nil
```

Sources/NIO/AddressedEnvelope.swift
glbrntt, line 34:
```suggestion
        public var ecnState: NIOECNState
```

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
...
+/// Possible ECN States
+public enum NIOECNState: Hashable {
+    /// Non-ECN Capable Transport.
+    case nonECT
+    /// ECN Capable Transport (flag 0).
+    case ect0
+    /// ECN Capable Transport (flag 1).
+    case ect1
+    /// Congestion Experienced.
+    case ce
```
glbrntt, line 53: nit: `congestionEncountered`/`congestionExperienced`

PeterAdams-A: I was heavily influenced by
https://developer.apple.com/documentation/network/nwprotocolip/ecn which has experienced.
Wikipedia https://en.wikipedia.org/wiki/Explicit_Congestion_Notification goes for encountered.
RFC https://tools.ietf.org/html/rfc3168 goes for experienced.

I think experienced wins 2:1 :-)

Lukasa: I think George is proposing changing the name of the enum case. ;)

glbrntt: Correct. We should spell it out either way to gain some clarity here.

PeterAdams-A: Ahh - I should probably spell them all out then.

PeterAdams-A: I have expanded the naming in this area.

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -20,15 +20,35 @@
...
+    /// Any metadata associated with this `AddressedEnvelope`
+    public var metadata: Metadata? = nil
 
     public init(remoteAddress: SocketAddress, data: DataType) {
         self.remoteAddress = remoteAddress
         self.data = data
     }
+    
+    /// Any metadata associated with an `AddressedEnvelope`
+    public struct Metadata {
```
Lukasa, line 32: Let's make this struct `Hashable` too, while we're here.

PeterAdams-A: Done.

PR #1508, opened 2020-05-06
Add inverse of `preconditionInEventLoop()` and ability to add messages to the assertion

gwynne:
Add `preconditionNotInEventLoop(_:file:line:)` and `assertNotInEventLoop(_:file:line:)`, and add optional message parameter to the existing counterparts.

**WARNING**: There are no tests for the existing or new precondition APIs, as unit tests can not test crash conditions and the integration tests currently can not build all of NIO when testing crashes. An issue will be filed to track addressing this shortcoming.

### Motivation:

1. It is valuable to be able to provide additional context for any kind of assertion failure, such as obvious points of API misuse or when the purpose of the assertion itself is unclear. The explainer message provided on precondition failure in `EventLoopFuture.wait()` is an ideal example.

1. As it is useful to have a precondition to assert that calling code _is_ on the given `EventLoop`, it is also useful to assert that calling code is _not_ on that `EventLoop`. Precedent is provided by the `dispatchPrecondition()` API.

### Modifications:

- Added a distinct overload of `preconditionInEventLoop()` that accepts a `message` `@autoclosure` parameter in the same fashion as does `Swift.precondition(_:file:line:)`. The overload's implementation calls that very function with the appropriate inputs, including the forwarded message (if any). The implementation of the original version now calls through to the new one with an empty message, providing 100% source compatibility.
- Added the same new parameter to `assertInEventLoop()`. Didn't bother with a forwarder, as the method is not specified on the `EventLoop` protocol; just gave the new param a default value.
- Rewrote the comments of all aforementioned methods to be both more grammatical and more technically precise.
- Add the two new inverse methods to `EventLoop`, providing the precondition form as a protocol requirement and the assertion form only as an extension method, to match their counterparts.
- Provide defaulted implementations of the new methods as appropriate to their defined behaviors. The new methods also accept `message` autoclosure inputs but, being newly added, default the argument rather than adding forwarders.
- Finally, replaced various uses of `precondition(!eventLoop.inEventLoop)` and `assert(!eventLoop.inEventLoop)` by `EventLoopFuture` and `NIOHTTP1TestServer` with `eventLoop.preconditionNotInEventLoop()` and `eventLoop.assertNotInEventLoop()` respectively.

### Result:

- It is now possible to optionally provide explanatory messages to be included when "in/not in event loop" preconditions fail, as with any other precondition failure, matching the behavior of the `Swift.precondition()` API.
- It is now possible to canonically assert that a given execution path is _not_ executing in the thread context belonging to a given event loop. Improves parity with `dispatchPrecondition()`, allows replacing direct access to `EventLoop.inEventLoop` with direct statements of intent, and enables overriding the assertion behaviors on a per-loop basis (such as the use of `dispatchPrecondition()` by `NIOTSEventLoop`, not shown in this PR of course),


gwynne:
Filed #1509 to track crash-detecting integration tests that cover all of NIO's codebase.

gwynne:
The failure of the API breakage test is expected given the new APIs added and the source-compatible change of `assertInEventLoop()`'s signature, I think?

Sources/NIO/EventLoop.swift
```diff
@@ -244,9 +244,17 @@ public protocol EventLoop: EventLoopGroup {
     @discardableResult
     func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T>
 
-    /// Checks that this call is run from the `EventLoop`. If this is called from within the `EventLoop` this function
-    /// will have no effect, if called from outside the `EventLoop` it will crash the process with a trap.
+    /// Asserts that the current thread is the one tied to this `EventLoop`. Has no effect if `inEventLoop` is `true`.
```
Lukasa, line 247: This statement is worth removing: while it’s true, it implies that the behaviour of this function is in terms of `inEventLoop`. However, it isn’t.

Sources/NIO/EventLoop.swift
```diff
@@ -244,9 +244,17 @@ public protocol EventLoop: EventLoopGroup {
...
+    /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
     func preconditionInEventLoop(file: StaticString, line: UInt)
+    
+    /// Asserts that the current thread is the one tied to this `EventLoop`. Has no effect if `inEventLoop` is `true`.
```
Lukasa, line 251: Same note here.

Sources/NIO/EventLoop.swift
```diff
@@ -244,9 +244,17 @@ public protocol EventLoop: EventLoopGroup {
...
+    /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
+    func preconditionInEventLoop(_ message: @autoclosure() -> String, file: StaticString, line: UInt)
+    
+    /// Asserts that the current thread is _not_ the one tied to this `EventLoop`. Has no effect if `inEventLoop` is `false`.
```
Lukasa, line 255: Again, let’s remove the comment about `inEventLoop`

gwynne:
> Generally looks good. Please also file a NIOTS issue to adopt these APIs in terms of `dispatchPrecondition`

A PR addressing the implementation for NIOTS is in progress.

gwynne:
First draft of changes for NIOTS is up. Needs significant additional work.

Sources/NIO/EventLoop.swift
```diff
@@ -244,9 +244,17 @@ public protocol EventLoop: EventLoopGroup {
...
 
-    /// Checks that this call is run from the `EventLoop`. If this is called from within the `EventLoop` this function
-    /// will have no effect, if called from outside the `EventLoop` it will crash the process with a trap.
+    /// Asserts that the current thread is the one tied to this `EventLoop`.
+    /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
     func preconditionInEventLoop(file: StaticString, line: UInt)
+    
+    /// Asserts that the current thread is the one tied to this `EventLoop`.
+    /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
+    func preconditionInEventLoop(_ message: @autoclosure() -> String, file: StaticString, line: UInt)
```
weissi, line 253: We need to take out the new parameter here, we can't change `preconditionInEventLoop`

Sources/NIO/EventLoop.swift
```diff
@@ -580,21 +588,45 @@ extension EventLoop {
...
-    /// have no effect, if called from outside the EventLoop it will crash the process with a trap if run in debug mode.
-    /// In release mode this function never has any effect.
+    /// Asserts that the current thread is the one tied to this `EventLoop`.
+    /// Otherwise, if running in debug mode, the process will be abnormally terminated as per the semantics of
+    /// `preconditionFailure(_:file:line:)`. Never has any effect in release mode.
     ///
     /// - note: This is not a customization point so calls to this function can be fully optimized out in release mode.
     @inlinable
-    public func assertInEventLoop(file: StaticString = #file, line: UInt = #line) {
+    public func assertInEventLoop(_ message: @autoclosure() -> String = "", file: StaticString = #file, line: UInt = #line) {
```
weissi, line 597: This new parameter should be taken out too.

Sources/NIO/EventLoop.swift
```diff
@@ -580,21 +588,45 @@ extension EventLoop {
...
+        preconditionInEventLoop("", file: file, line: line)
+    }
+
+    /// Checks the necessary condition of currently running on the called `EventLoop` for making forward progress.
+    public func preconditionInEventLoop(_ message: @autoclosure() -> String, file: StaticString = #file, line: UInt = #line) {
+        precondition(self.inEventLoop, message(), file: file, line: line)
+    }
+
+    /// Checks the necessary condition of currently _not_ running on the called `EventLoop` for making forward progress.
+    public func preconditionNotInEventLoop(_ message: @autoclosure() -> String = "", file: StaticString = #file, line: UInt = #line) {
```
weissi, line 628: This should be inlinable, or else we get needless allocations in Swift 5.0 for example.

Sources/NIO/EventLoop.swift
```diff
@@ -580,21 +588,45 @@ extension EventLoop {
...
     /// Checks the necessary condition of currently running on the called `EventLoop` for making forward progress.
+    /// Forwarder to preserve API while adding message parameter.
+    @inlinable
     public func preconditionInEventLoop(file: StaticString = #file, line: UInt = #line) {
-        precondition(self.inEventLoop, file: file, line: line)
+        preconditionInEventLoop("", file: file, line: line)
+    }
+
+    /// Checks the necessary condition of currently running on the called `EventLoop` for making forward progress.
+    public func preconditionInEventLoop(_ message: @autoclosure() -> String, file: StaticString = #file, line: UInt = #line) {
```
weissi, line 623: This should be inlinable, or else we get needless allocations in Swift 5.0 for example.

Sources/NIO/EventLoop.swift
Lukasa: We're not changing it, the original is above.

Sources/NIO/EventLoop.swift
Lukasa: _This_ one is wrong, and needs to be done the same way `preconditionInEventLoop` is done: add a new method.

weissi: You're right. I didn't notice it's a new method but I think we should probably keep to the old customisation point (even the newer would be nicer). It's confusing enough that you're supposed to implement them

Lukasa:
Cool, I think we think the API breakage checker is wrong for known reasons, so I'm going to merge this.

PR #1510, opened 2020-05-07
Adjust names of BSDSocket option values.

Lukasa:
Motivation:

The names of most of the BSD socket option values were brought over into
their NIO helper properties, with their namespace removed. This vastly
increases the risk of collision, particularly for things like TCP_INFO,
which just became .info.

Modifications:

- Added the prefixes back, e.g. `.info` is now `.tcp_info`.
- Renamed socket type `.dgram` to `.datagram`, as this is the nice clean
  API and we should use nice clean names.

Result:

Better APIs


Lukasa:
Please ignore the test failure on the API checker: yes, I broke the API, but we hadn't shipped that API yet.

Also cc @compnerd.

Lukasa:
Brief ping of @weissi

Lukasa:
The API breakage is expected: we haven't shipped this API yet.

PR #1513, opened 2020-05-09
Allocation diffing script no longer drops data.

PeterAdams-A:
Allocation diffing script no longer drops data.

### Motivation:

Looking at allocations is hard enough without the script getting the allocation counts wrong.

### Modifications:

Where multiple allocations resolved to the same "key" only the last allocation was accounted for.  I have changed the script to store all allocations - still keyed in the same way.  When diffing the total number of allocations is used - if they are different the total and all contributing stack traces are output.

Added a total for all allocations everywhere at the end together with a difference number.

Output all stack traces before and after contributing to a detected diff.

Update the documentation to reflect changes.

Change the threshold for reporting diffs from > 1000 to >= 1000.  This means if allocations go from 1000 to 2000 they are reported as a difference rather than a new allocation.

### Result:

It is now easier - if somewhat more verbose to compare allocations.


PeterAdams-A:

PeterAdams-A:

PR #1515, opened 2020-05-13
survive having disableSIGPIPE fail

weissi:
Motivation:

We know that on Darwin (eg. when the other end of a TCP connection reset
_really_ early on) F_SETNOSIGPIPE sometimes fails. We thought we could
handle it but we weren't able to handle everywhere.

Modifications:

- make us survive it everywhere
- tests

Result:

- Fewer crashes with ClientBootstrap.withConnectedSocket


weissi:

PR #1517, opened 2020-05-13
Update supported version in README

glbrntt:
Motivation:

NIO can run on macOS 10.10+, but the README states 10.12+.

Modifications:

Update readme.

Result:

More accurate docs.


glbrntt:

weissi:
@glbrntt we could even add macOS 10.9 and iOS 7 :)

glbrntt:
@weissi ah neat, will add. We should make our pods consistent with that as well: https://github.com/apple/swift-nio/blob/c1d86289ed7b2f44f71071da2e1a776a19ed956e/scripts/build_podspecs.sh#L104-L106

Issue #1512, opened 2020-05-08
fix cores count in containers


gautierdelorme:
The current `System.coreCount` implementation relies on `_SC_NPROCESSORS_ONLN` so isn't cgroups aware which might have a bad impact for apps runnings in containers (e.g. Docker, Kubernetes, Amazon ECS...).
This issue might affect many users since SwiftNIO examples and documentation recommend the usage of `System.coreCount` to start event loops through `MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)`. It seems like the community is following these guidelines and relies heavily on `System.coreCount`. For example, Vapor follows the guidelines and also relies on `System.coreCount` (see https://github.com/vapor/vapor/blob/633b2a81f547383aa6becb6ba1ad011dc7174779/Sources/Vapor/Application.swift#L71).

### Expected behavior

`System.coreCount` to return the number of cores available to containers.

### Actual behavior

`System.coreCount` returns the containers' host cores count instead of the number of cores available to containers

### Steps to reproduce

1. Build a `swift-nio` image
   ```console
   $ docker build -t swift-nio:18.04-5.1 -f docker/Dockerfile --build-arg ubuntu_version=bionic --build-arg swift_version=5.1 .
   ```
1. Print `System.coreCount` when a container is not restricted
   ```console
   $ docker run --rm -it --privileged -v $PWD:/code -w /code swift-nio:18.04-5.1 swift run --repl
   Welcome to Swift version 5.1.5 (swift-5.1.5-RELEASE).
     1> import NIO; System.coreCount
   $R0: Int = 4
   ```
1. Print `System.coreCount` when a container is restricted to 1 core using cpusets
   ```console
   $ docker run --cpuset-cpus=0 --rm -it --privileged -v $PWD:/code -w /code swift-nio:18.04-5.1 swift run --repl
   Welcome to Swift version 5.1.5 (swift-5.1.5-RELEASE).
     1> import NIO; System.coreCount
   $R0: Int = 4
   ```
1. Print `System.coreCount` when a container is restricted to 2 cores using CFS quota
   ```console
   $ docker run --cpus=2 --rm -it --privileged -v $PWD:/code -w /code swift-nio:18.04-5.1 swift run --repl
   Welcome to Swift version 5.1.5 (swift-5.1.5-RELEASE).
     1> import NIO; System.coreCount
   $R0: Int = 4
   ```

### SwiftNIO commit hash

[...]

PR #1518, opened 2020-05-13
fix cores count in containers

gautierdelorme:
### Motivation:

Fixes https://github.com/apple/swift-nio/issues/1512.
The current System.coreCount implementation relies on _SC_NPROCESSORS_ONLN so isn't cgroups aware which might have a bad impact for apps runnings in containers (e.g. Docker, Kubernetes, Amazon ECS...).

### Modifications:

- Changed System.coreCount on Linux only to make it read from CFS quotas and cpusets when present.
- Removed incorrect precondition in Sources/NIO/LinuxCPUSet.swift

### Result:

System.coreCount returns correct values when apps run in containers.


Lukasa:

Sources/NIO/Linux.swift
```diff
@@ -15,6 +15,7 @@
 // This is a companion to System.swift that provides only Linux specials: either things that exist
 // only on Linux, or things that have Linux-specific extensions.
 import CNIOLinux
+import Foundation
```
weissi, line 18: SwiftNIO doesn't use `Foundation` and must continue not to. Would you mind removing this import?

Tests/NIOTests/LinuxTest.swift
```diff
@@ -0,0 +1,65 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5: this should become a `2020` :)

Tests/NIOTests/LinuxTest+XCTest.swift
```diff
@@ -0,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5: (this copyright can remain because it's auto-gen'd)

Sources/NIO/Linux.swift
```diff
@@ -132,5 +137,35 @@ internal enum Linux {
...
+            let first = ids.first.flatMap(Int.init),
+            let last = ids.last.flatMap(Int.init),
+            last >= first
+        else { throw CpusetError() }
+        return 1 + last - first
+    }
+
+    static func coreCount(cpuset cpusetPath: String) -> Int? {
+        guard
+            let cpuset = try? String(contentsOfFile: cpusetPath, encoding: .utf8).trimmingCharacters(in: .whitespacesAndNewlines).components(separatedBy: ","),
```
weissi, line 153: I know it's annoying but we'd have to redo the file reading using `NIOFileHandle`'s open/close functions and then probably the `Posix.read` function.

I know this is very annoying and we don't have high-level API to support this, SwiftNIO only offers asynchronous APIs in `NonBlockingFileIO` but those aren't appropriate here :'(

Sources/NIO/Linux.swift
```diff
@@ -113,6 +113,10 @@ internal enum Epoll {
 }
 
 internal enum Linux {
+    static let CFS_QUOTA_PATH = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
```
weissi, line 116: It'd be quite nice to swiftify those a bit, maybe `cfsQuotaPath`, `cfsPeriodPath`, and `cpuSetPath`?

Sources/NIO/Linux.swift
```diff
@@ -132,5 +136,60 @@ internal enum Linux {
         }
         return fd
     }
+
+    private static func contentsOfFile(path: String) throws -> Substring {
+        let fh = try NIOFileHandle(path: path)
+        defer { try? fh.close() }
```
weissi, line 142: this should be `try!`. The close needs to work here.

Sources/NIO/Linux.swift
```diff
@@ -132,5 +136,60 @@ internal enum Linux {
...
+        // linux doesn't properly report /sys/fs/cgroup/* files lengths so we use a reasonable limit
+        let buffSize = 1024
+        var buf = ByteBufferAllocator().buffer(capacity: buffSize)
+        try buf.writeWithUnsafeMutableBytes(minimumWritableBytes: buffSize) { ptr in
+            let res = try fh.withUnsafeFileDescriptor { fd -> IOResult<ssize_t> in
+                return try Posix.read(descriptor: fd, pointer: ptr.baseAddress!, size: buffSize)
```
weissi, line 148: instead of `buffSize` you should use `ptr.count`. `ByteBufferAllocator` reserves the right to allocate more so if it does, we can make use of that :).

Sources/NIO/Linux.swift
```diff
@@ -132,5 +136,60 @@ internal enum Linux {
...
+            }
+            switch res {
+            case .processed(let n):
+                return n
+            case .wouldBlock:
+                throw LinuxError()
```
weissi, line 154: You can put a

    preconditionFailure("read returned EWOULDBLOCK despite a blocking fd")

here. This will never happen, and if it does, this needs to crash because it would then be programmer error (someone passed `O_NONBLOCK`).

Sources/NIO/Linux.swift
```diff
@@ -132,5 +136,60 @@ internal enum Linux {
...
+            }
+        }
+        guard let content = buf.readString(length: buf.readableBytes) else { throw LinuxError() }
```
weissi, line 157: You can do

```
let content = String(buffer: buf)
```

Sources/NIO/Linux.swift
```diff
@@ -132,5 +136,60 @@ internal enum Linux {
...
+    }
+
+    private static func toInt<S: StringProtocol>(_ s: S) -> Int? {
+        return Int(s)
+    }
+
+    private static func countCoreIds(cores: Substring) throws -> Int {
+        let ids = cores.split(separator: "-")
+        guard
+            let first = ids.first.flatMap(toInt),
```
weissi, line 168: instead of `toInt`, you can just write `ids.first.flatMap { Int($0) }` or possibly `ids.first.flatMap(Int.init)`

Sources/NIO/Linux.swift
```diff
@@ -113,6 +113,10 @@ internal enum Epoll {
 }
 
 internal enum Linux {
+    static let CFS_QUOTA_PATH = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
+    static let CFS_PERIOD_PATH = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
+    static let CPUSET_PATH = "/sys/fs/cgroup/cpuset/cpuset.cpus"
+    struct LinuxError: Error { }
```
weissi, line 119: Pretty much everything that's a `LinuxError` should never happen so we can remove this error I think.

Sources/NIO/Linux.swift
gautierdelorme: @weissi this is what I was initially doing but it doesn't work with `Substring`. I first went with `ids.first.map(String.init).flatMap(Int.init)` but then I thought the `toInt` alternative would look better. I'm fine either way so let me know if you prefer the other way. :)

gautierdelorme: I went with `ids.first.map(String.init).flatMap(Int.init)` again :)

Sources/NIO/Linux.swift
```diff
@@ -132,5 +135,54 @@ internal enum Linux {
...
+                return try Posix.read(descriptor: fd, pointer: ptr.baseAddress!, size: ptr.count)
+            }
+            switch res {
+            case .processed(let n):
+                return n
+            case .wouldBlock:
+                preconditionFailure("read returned EWOULDBLOCK despite a blocking fd")
+            }
+        }
+        return String(buffer: buf).prefix(while: { $0 != "\n" })
```
Lukasa, line 155: Maybe this function should only return `String` instead of `Substring` and this `prefix(while:)` should move elsewhere? Otherwise it's a bit weird that this is called `contentsOfFile`, when it's really `firstLineOfFile`.

Sources/NIO/Linux.swift
```diff
@@ -132,5 +135,54 @@ internal enum Linux {
...
+    }
+
+    private static func countCoreIds(cores: Substring) -> Int {
+        let ids = cores.split(separator: "-")
+        guard
+            let first = ids.first.map(String.init).flatMap(Int.init),
```
Lukasa, line 161: The transformation through `String` is unnecessary, there's an [initializer from `Substring`](https://developer.apple.com/documentation/swift/int/2924481-init).

Sources/NIO/Linux.swift
```diff
@@ -132,5 +135,54 @@ internal enum Linux {
...
+                return n
+            case .wouldBlock:
+                preconditionFailure("read returned EWOULDBLOCK despite a blocking fd")
+            }
+        }
+        return String(buffer: buf).prefix(while: { $0 != "\n" })
+    }
+
+    private static func countCoreIds(cores: Substring) -> Int {
+        let ids = cores.split(separator: "-")
```
Lukasa, line 159: `maxSplits: 1` should be passed here.

Sources/NIO/Linux.swift
```diff
@@ -132,5 +135,54 @@ internal enum Linux {
...
+        guard
+            let first = ids.first.map(String.init).flatMap(Int.init),
+            let last = ids.last.map(String.init).flatMap(Int.init),
+            last > first || !cores.contains("-")
```
Lukasa, line 163: If we adopt `maxSplits: 1` we can remove these `cores.contains` logic and just use `last >= first`.

Sources/NIO/Linux.swift
gautierdelorme: I renamed it to `firstLineOfFile` to be more correct. :)

weissi:
CC @neonichu / @abertelrud SwiftPM needs that too. We often see OOM kills in containers because SwiftPM spawns way too many jobs. For example if the actual hardware has 48 cores but the container gets assigned just 1 core, then SwiftPM will still spawn 48 compilers I think. (Unless this has been fixed recently)

PR #1524, opened 2020-05-18
HTTPEncoder: fix 0 length chunks

weissi:
Motivation:

Previously, if the user sent us a 0 length chunk, we would encode it as
`0\r\n\r\n` which means the _end_ of the body in chunked encoding. But
the end in NIO is send by sending `.end(...)`.

Modifications:

Don't write anything for 0 length writes.

Result:

More correct behaviour with 0 length body chunks.


Lukasa:

PR #1526, opened 2020-05-22
silence the #file used instead of #filepath warning

weissi:
Motivation:

Very recent versions of the Swift compiler complain if you pass `#file`
to a parameter that is defaulted to `#filePath`. You can silence the
warning by using `(#file)`. We cannot migrate to `#filepath` because we
support Swift 5.0 and 5.1 which don't have `#filepath` yet.

Modifications:

- replace all occurances of `#file` by a special (test `internal`)
function `fullFilePath()` which gives you the full path and is defined
differently depending on the compiler version.

Result:

No warnings even on the latest compiler versions.


weissi:
weird, hit #1535

weissi:

Lukasa:
Allocation count problem.

Lukasa:
Per @weissi the 5.3 failure is the result of a regression in 5.3, not this change. Merging over the failure.

weissi:

PR #1530, opened 2020-05-27
don't blow up if schedule happens during EL shutdown

weissi:
Motivation:

Previously, if we scheduled EventLoop executions from the promises we
failed during EventLoop shutdown, we'd blow up. Thanks to @fabianfett
for the bug report.

Modifications:

Set the EL state to 'exiting thread' only after we failed all
outstanding tasks.

Result:

Fewer crashes.

### DIFF
--- a/Sources/NIO/SelectableEventLoop.swift
+++ b/Sources/NIO/SelectableEventLoop.swift
@@ -354,21 +354,28 @@ internal final class SelectableEventLoop: EventLoop {
         self.preconditionInEventLoop()
         defer {
             var scheduledTasksCopy = ContiguousArray<ScheduledTask>()
-            _tasksLock.withLockVoid {
-                // reserve the correct capacity so we don't need to realloc later on.
-                scheduledTasksCopy.reserveCapacity(self._scheduledTasks.count)
-                while let sched = self._scheduledTasks.pop() {
-                    scheduledTasksCopy.append(sched)
+            var iterations = 0
+            repeat { // We may need to do multiple rounds of this because failing tasks may lead to more work.
+                scheduledTasksCopy.removeAll(keepingCapacity: true)
+
+                self._tasksLock.withLockVoid {
+                    // reserve the correct capacity so we don't need to realloc later on.
+                    scheduledTasksCopy.reserveCapacity(self._scheduledTasks.count)
+                    while let sched = self._scheduledTasks.pop() {
+                        scheduledTasksCopy.append(sched)
+                    }
                 }
-            }
+
+                // Fail all the scheduled tasks.
+                for task in scheduledTasksCopy {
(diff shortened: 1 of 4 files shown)


Sources/NIO/SelectableEventLoop.swift
```diff
@@ -354,21 +354,25 @@ internal final class SelectableEventLoop: EventLoop {
...
+                        scheduledTasksCopy.append(sched)
+                    }
                 }
-            }
+
+                // Fail all the scheduled tasks.
+                for task in scheduledTasksCopy {
+                    task.fail(EventLoopError.shutdown)
+                }
+            } while scheduledTasksCopy.count > 0
```
Lukasa, line 372: This can loop infinitely if users are repeatedly scheduling tasks in the teardown. Should we crash if the iteration count gets too high?

weissi: @Lukasa Yeah, I thought about that one too. I didn't think it was too likely that a user would _accidentally_ run into this situation and therefore didn't think it was worthwhile putting a certain limit on it (I can totally be convinced differently).

Also, we have the same kind of unlimited logic in the event loop itself already.

Lukasa: Yeah, but event loops are _loops_ so I don't mind so much. I was just thinking a really high iteration count (like 1000) that then blows up if we exceed it so that it's easy to diagnose.

glbrntt: I'm slightly in favour of having a limit here, I think. Is there any downside to having it?

weissi: Let's assume you do continuously schedule new work _during_ shutdown.

Wouldn't the user hang both with and without the limit?

- With the limit we would likely either hang (because the user's work doesn't happen) or crash (because some promises won't get fulfilled).
- Without the limit we would just spin and therefore hang (because we continuously run the new, incoming work).

So I'm kinda worried adding fairly random limits without an actual real-world example where that would've helped us.

WDYT?

weissi: Or @Lukasa did you want to crash if that situation happens?

Lukasa: I want to crash. The explicit goal is to say: "hey, you have a work pattern that forces a hang on shutdown". It's basically a livelock condition, and if anyone is waiting for the process to exit then they won't ever see it.

weissi: Ah okay, crash is fine, sorry!

weissi: done and now even tested :)

weissi:
thanks @glbrntt , put a workaround in

PR #1531, opened 2020-05-28
Finer grained allocation tests

PeterAdams-A:
Finer grained allocation tests.

### Motivation:

The existing tcp allocation tests cover everything from bootstrap creation through to sending data in one massive test.  The parameters of how many iterations control the focus of the test.  It would be better to be more explicit about what we're trying to test.

### Modifications:

3 new tests
1) client bootstrap creation.
2) connection establishment.
3) data transfer.

### Result:

Easier to assess the impact of changes to memory allocation patterns.


docker/docker-compose.1604.53.yaml
```diff
@@ -35,6 +35,9 @@ services:
       - MAX_ALLOCS_ALLOWED_read_10000_chunks_from_file=200500 #5.3 improvement 210050
       - MAX_ALLOCS_ALLOWED_udp_1000_reqs_1_conn=16250
       - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=226050
+      - MAX_ALLOCS_ALLOWED_1000_reqs=6100
+      - MAX_ALLOCS_ALLOWED_1000_tcpbootstraps=4100
```
Lukasa, line 39: This is a sad regression, doubly so given that it seems to have come in 5.2.

PeterAdams-A: I think it's the same one we've discovered 3 times and every time feel sad about it.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
```diff
@@ -0,0 +1,125 @@
...
+        context.write(data, promise: nil)
+    }
+
+    // Flush it out. This can make use of gathering writes if multiple buffers are pending
+    public func channelReadComplete(context: ChannelHandlerContext) {
+        context.flush()
+    }
+
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        print("error: ", error)
```
weissi, line 33: let's remove this print and make it a `fatalError` instead. There are only a couple of outcomes:

- we hit this always: test numbers will be skewed because `print` allocates a lot
- we never hit this: `fatalError` is cool
- we sometimes hit this: shows that the test has issues (or possibly the error needs to be ignored, depending on what it is)

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
```diff
@@ -0,0 +1,125 @@
...
+    let pileOfBytes: [UInt8] = .init(repeating: 5, count: 1000)
+    
+    var iterationsOutstanding = 0
+    var bytesOutstanding = 0
+    var whenDone: EventLoopPromise<()>? = nil
+    
+    private func sendBytes(clientChannel: Channel) {
+        bytesOutstanding += pileOfBytes.count
+        var buffer = clientChannel.allocator.buffer(capacity: pileOfBytes.count)
+        buffer.writeBytes(pileOfBytes)
```
weissi, line 82: sending more than 1 byte will make this more unstable than necessary. The don't necessarily get sent/received over the "wire" in one chunk. So we can't know the number of `read`s we need to perform to read them.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
```diff
@@ -0,0 +1,125 @@
...
+        // As we are not really interested getting notified on success or failure we just pass nil as promise to
+        // reduce allocations.
+        context.close(promise: nil)
+    }
+    
+    let pileOfBytes: [UInt8] = .init(repeating: 5, count: 1000)
+    
+    var iterationsOutstanding = 0
+    var bytesOutstanding = 0
+    var whenDone: EventLoopPromise<()>? = nil
```
weissi, line 77: we moved to `ELP<Void>` a long time ago because the stdlib team told us that that's preferred (for whatever reason :P )

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+fileprivate final class ClientHandler: ChannelInboundHandler {
+    public typealias InboundIn = ByteBuffer
+    public typealias OutboundOut = ByteBuffer
+    
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        // Count the data in.
+        let byteBuffer = self.unwrapInboundIn(data)
```
weissi, line 47: I think we should modify this to only ever expect exactly 1 byte and then reply with that one byte, or else it'll be unstable

docker/docker-compose.1604.53.yaml
weissi: which is the regression? We know about a 3 allocation regression _per connection_.

docker/docker-compose.1604.53.yaml
```diff
@@ -35,6 +35,9 @@ services:
       - MAX_ALLOCS_ALLOWED_read_10000_chunks_from_file=200500 #5.3 improvement 210050
       - MAX_ALLOCS_ALLOWED_udp_1000_reqs_1_conn=16250
       - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=226050
+      - MAX_ALLOCS_ALLOWED_1000_reqs=6100
```
weissi, line 38: isn't `1000_reqs` similar to `ping_pong_1000_reqs_1_conn`?

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
PeterAdams-A: good point.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_reqs.swift
PeterAdams-A: Fine.

docker/docker-compose.1604.53.yaml
PeterAdams-A: Yeah - I'll ditch this one.

PeterAdams-A: which in fact makes the other problems go away.

docker/docker-compose.1604.53.yaml
PeterAdams-A: @weissi - recall we found that there was a regression that was something to do with channel options but which was very hard to look at when we did https://github.com/apple/swift-nio/pull/1496.
Given how hard it is to look at regressions across compiler versions I'm reluctant dive in again.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_tcpbootstraps.swift
```diff
@@ -0,0 +1,36 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
```

IntegrationTests/tests_04_performance/test_01_resources/test_1000_tcpconnections.swift
```diff
@@ -0,0 +1,50 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5:
```suggestion
// Copyright (c) 2020 Apple Inc. and the SwiftNIO project authors
```

Issue #554, opened 2018-08-06
test testScheduleRepeatedTaskCancelFromDifferentThread flaky


weissi:
https://ci.swiftnio.io/job/swift-nio-master-410-prb/416/console :

```
16:56:42 	 Executed 36 tests, with 0 failures (0 unexpected) in 2.484 (2.484) seconds
16:56:42 Test Suite 'EventLoopTest' started at 2018-08-06 15:56:42.630
16:56:42 Test Case 'EventLoopTest.testSchedule' started at 2018-08-06 15:56:42.630
16:56:43 Test Case 'EventLoopTest.testSchedule' passed (1.002 seconds)
16:56:43 Test Case 'EventLoopTest.testScheduleWithDelay' started at 2018-08-06 15:56:43.632
16:56:44 Test Case 'EventLoopTest.testScheduleWithDelay' passed (1.003 seconds)
16:56:44 Test Case 'EventLoopTest.testScheduleCancelled' started at 2018-08-06 15:56:44.635
16:56:46 Test Case 'EventLoopTest.testScheduleCancelled' passed (2.011 seconds)
16:56:46 Test Case 'EventLoopTest.testScheduleRepeatedTask' started at 2018-08-06 15:56:46.646
16:56:46 Test Case 'EventLoopTest.testScheduleRepeatedTask' passed (0.068 seconds)
16:56:46 Test Case 'EventLoopTest.testScheduleRepeatedTaskCancelFromDifferentThread' started at 2018-08-06 15:56:46.714
16:56:50 Exited with signal code 4
```

so we crash in testScheduleRepeatedTaskCancelFromDifferentThread

weissi:
CC @FranzBusch  too

weissi:
ok, I identified the race condition for this, working on a fix.

PR #1532, opened 2020-05-28
Make SocketAddress Hashable.

Lukasa:
Motivation:

In general there's no good justification for `SocketAddress` not to be
`Hashable`: it's a well-formed value type that can easily have a
clearly-defined version of hashability defined. In particular, we have
a clear definition of equatability for this type already present in the
code.

This makes SocketAddress more broadly useful in a range of
circumstances.

Modifications:

- Defined `Hashable` for `SocketAddress`.

Result:

Users can obtain the hash value of `SocketAddress`.


Sources/NIO/SocketAddresses.swift
```diff
@@ -363,6 +363,35 @@ extension SocketAddress: Equatable {
...
 
+/// We define an extension on `SocketAddress` that gives it an elementwise hashable conformance, using
+/// only the elements defined on the structure in their man pages (excluding lengths).
+extension SocketAddress: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        switch self {
+        case .unixDomainSocket(let uds):
+            hasher.combine(0)
+            hasher.combine(uds.address.sun_family)
+            withUnsafeBytes(of: uds.address.sun_path) {
```
weissi, line 374: This is actually incorrect in both hashable and `==`. This needs to only compare the bytes until the first `0` byte I think.

weissi: (ignore this is we know for sure that all trailing bytes are always `0` for example  in which case we should add an `assert`).

Lukasa: Yeah, I felt pretty weird about it as I was writing it tbh.

Lukasa: Updated, please take a look.

Sources/NIO/SocketAddresses.swift
```diff
@@ -354,15 +354,62 @@ extension SocketAddress: Equatable {
...
-            return memcmp(&sunpath1, &sunpath2, MemoryLayout.size(ofValue: sunpath1)) == 0
+            let bufferSize = MemoryLayout.size(ofValue: addr1.address.sun_path)
+
+            // These uses of withMemoryRebound(to:) are fine: the `strncmp` call cannot possibly re-enter Swift code, and so we cannot possibly violate
+            // Swift's strict aliasing rules by way of this type-pun.
+            return withUnsafePointer(to: addr1.address.sun_path) { sunpath1 in
+                return withUnsafePointer(to: addr2.address.sun_path) { sunpath2 in
+                    return sunpath1.withMemoryRebound(to: Int8.self, capacity: bufferSize) { sunpath1 in
+                        return sunpath2.withMemoryRebound(to: Int8.self, capacity: bufferSize) { sunpath2 in
+                            return strncmp(sunpath1, sunpath2, MemoryLayout.size(ofValue: addr1.address.sun_path)) == 0
```
glbrntt, line 365: Isn't this just another `bufferSize`?

Lukasa:
Hit now-closed issue #554. @weissi are we sure that fix worked?

Lukasa:

Sources/NIO/SocketAddresses.swift
Lukasa: Yup, I just missed this one.

weissi:
@Lukasa pretty sure the original fix fixed one issue (which we hit frequently) but there might be a new one. Mind filing a fresh one with the printout?

Lukasa:
See #1535.

PR #1534, opened 2020-05-28
Fine grained UDP allocation tests

PeterAdams-A:
Fine grained UDP allocation tests

### Motivation:

The existing UDP allocation tests count allocations everywhere.

Split out 3 tests to cover:
1) Making connections.
2) Making bootstraps.
3) Transferring data.

### Modifications:

Three new tests.

### Result:

Easier to debug changes to allocation profiles in UDP.


IntegrationTests/tests_04_performance/test_01_resources/test_1000_udp_reqs.swift
```diff
@@ -0,0 +1,124 @@
...
+        context.write(data, promise: nil)
+    }
+
+    // Flush it out. This can make use of gathering writes if multiple buffers are pending
+    public func channelReadComplete(context: ChannelHandlerContext) {
+        context.flush()
+    }
+
+    public func errorCaught(context: ChannelHandlerContext, error: Error) {
+        print("error: ", error)
```
weissi, line 33: same concern with `print`

IntegrationTests/tests_04_performance/test_01_resources/test_1000_udpconnections.swift
```diff
@@ -0,0 +1,56 @@
...
+    measure(identifier: identifier) {
+        let numberOfIterations = 1000
+        for _ in 0 ..< numberOfIterations {
+            let clientChannel = try! clientBootstrap.bind(to: localhostPickPort).wait()
+            defer {
+                try! clientChannel.close().wait()
+            }
+            
+            // Send a byte to make sure everything is really open.
+            var buffer = clientChannel.allocator.buffer(capacity: 1)
```
weissi, line 48: I still think we should just move these two lines out. `clientChannel.allocator.buffer(capacity: 1)` has to allocate twice (the `class` and the bytes) and we don't need to change this buffer.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_udp_reqs.swift
PeterAdams-A: Gah - I felt sure I'd changed that.

weissi: You did a very similar change in the other PR I think

PR #1536, opened 2020-05-30
crash tester test suite

weissi:
Motivation:

It's often very important to test that in certain cases we actually
crash with a certain error message whilst having access to all of NIO.

Modifications:

Add a crash tester test suite.

Result:

We can test crashers.


Sources/NIOCrashTester/CrashTests.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import Foundation
+
+let group = MultiThreadedEventLoopGroup(numberOfThreads: 2)
+
+struct Foo {
+    let multiThreadedELGCrashesOnZeroThreads = CrashTest { requireCrashRegex in
```
weissi, line 21: note to reviewers: this is how crash tests look. For each thing you'd like to test you add a

```swift
let whatYouWantToTest = CrashTest { requireCrashRegex in
    requireCrashRegex("^regex of what message you want to see")

    whatMakesItCrash()
}
```

Sources/NIOCrashTester/EventLoopCrashTests.swift
```diff
@@ -0,0 +1,35 @@
...
+fileprivate let group = MultiThreadedEventLoopGroup(numberOfThreads: 2)
+
+struct EventLoopCrashTests {
+    let testMultiThreadedELGCrashesOnZeroThreads = CrashTest(
+        regex: "^Precondition failed: numberOfThreads must be positive"
+    ) {
+        try? MultiThreadedEventLoopGroup(numberOfThreads: 0).syncShutdownGracefully()
+    }
+
+    let testWaitCrashesWhenOnEL = CrashTest(
```
weissi, line 27: note to reviewers: these files are the core. To add a new crash test, the only thing you need to do is

```
let testWhatYouAreTesting = CrashTest(regex: "^outputYouExpect") {
    codeThatWillCrash()
}
```

weissi: the will automatically be collected through the `Mirror` of the test struct. If you add a new test suite, then you'll need to edit `main.swift`.

Sources/NIOCrashTester/CrashTests+EventLoop.swift
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    let testPreconditionInEventLoop = CrashTest(
+        regex: "^Precondition failed: file .*CrashTests\\+EventLoop.swift, line \(#line + 2)"
+    ) {
+        group.next().preconditionInEventLoop()
+    }
+
+    let testAssertNotInEventLoop = CrashTest(
+        regex: "^Precondition failed: file .*CrashTests\\+EventLoop.swift, line \(#line + 4)"
```
weissi, line 49: CC @gwynne / @Davidde94

Sources/NIOCrashTester/CrashTests+ByteBuffer.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import Foundation
```
glbrntt, line 16: Seems unnecessary? (and elsewhere)

Sources/NIOCrashTester/CrashTests+EventLoop.swift
```diff
@@ -0,0 +1,65 @@
...
+        regex: #"^Precondition failed: BUG DETECTED: wait\(\) must not be called when on an EventLoop"#
+    ) {
+        let promise = group.next().makePromise(of: Void.self)
+        try? group.next().submit {
+            try? promise.futureResult.wait()
+        }.wait()
+    }
+
+    let testAssertInEventLoop = CrashTest(
+        regex: "^Precondition failed: file .*CrashTests\\+EventLoop.swift, line \(#line + 2)"
```
glbrntt, line 37: Super minor, this feels slightly brittle setting the expected line like this, would it be better to just manually override the line in `assertInEventLoop` etc.?

Sources/NIOCrashTester/CrashTests+ByteBuffer.swift
```diff
@@ -0,0 +1,27 @@
...
+
+fileprivate let group = MultiThreadedEventLoopGroup(numberOfThreads: 2)
```
Lukasa, line 18: We don't seem to be using this at all.

Sources/NIOCrashTester/CrashTests+ByteBuffer.swift
ktoso: (There isn't a linter which could fail if an import isn't used, is there? If there is I'd love to know -- Foundation imports creeping into more files than necessary is a bit annoying to me in my work :-/)

Sources/NIOCrashTester/CrashTests+EventLoop.swift
weissi: good idea!

weissi: done

Sources/NIOCrashTester/CrashTests+ByteBuffer.swift
weissi: thanks, removed

Sources/NIOCrashTester/CrashTests+ByteBuffer.swift
weissi: removed the unused imports

ktoso: I see so you test it based on what's linked hm... Was hoping there's some linter that per source could check it and I could "permit" it sometimes. No worries, just asking

Davidde94:
This is really cool :)

PR #1543, opened 2020-06-03
PipeBootstrap: fix file type change & accept 1 fd

weissi:
Motivation:

Sometimes it can be useful to use NIOPipeBootstrap with a single file
descriptor, like a serial line. Previously however we would only accept
two distinct fds (without checking they're distinct...).

Also, we didn't correctly check file types.

Modifications:

- also support NIOPipeChannel for streams that have just 1 fd
- fix file type check
- check that the 2 fds are distinct in withPipe(fds...)

Result:

- more functionality
- fewer bugs


glbrntt:
Isn't this a minor version bump?

Lukasa:
Yes.

PR #1547, opened 2020-06-08
Shorthand way of setting options in universal bootstrap

PeterAdams-A:
Motivation:

Guide novice users towards the options they are most likely to want to use.

Modifications:

New type to represent shorthand TCP client options containing only the options
which are sensible for novice users.
New type to represent a collection of these options.
A method to allow underlying bootstraps to override universal behaviour - default
implementation is to do nothing.

A test case to demonstrate that shorthand options match old style options.
A test case to demonstrate that overriding option behaviour is possible.

Result:

For universal botostrap there will be an new way of setting options involving
less typing which guides the user towards the common options.

### DIFF
--- a/Sources/NIO/ConvenienceOptionSupport.swift
+++ b/Sources/NIO/ConvenienceOptionSupport.swift
@@ -97,7 +97,7 @@ extension ChannelOptions {
 /// Approved shorthand options.
 extension ChannelOptions.NIOTCPShorthandOption {
     /// Allow immediately reusing a local address.
-    public static let allowImmediateLocalAddressReuse = ChannelOptions.NIOTCPShorthandOption(.reuseAddr)
+    public static let allowLocalEndpointReuse = ChannelOptions.NIOTCPShorthandOption(.reuseAddr)
     
     /// The user will manually call `Channel.read` once all the data is read from the transport.
     public static let disableAutoRead = ChannelOptions.NIOTCPShorthandOption(.disableAutoRead)
@@ -114,7 +114,7 @@ extension ChannelOptions.NIOTCPShorthandOption {
 extension ChannelOptions {
     /// A set of `NIOTCPShorthandOption`s
     public struct NIOTCPShorthandOptions: ExpressibleByArrayLiteral, Hashable {
-        var allowImmediateLocalAddressReuse = false
+        var allowLocalEndpointReuse = false
         var disableAutoRead = false
         var allowRemoteHalfClosure = false
         
@@ -130,7 +130,7 @@ extension ChannelOptions {
         mutating func add(_ element: NIOTCPShorthandOption) {
             switch element.data {
             case .reuseAddr:
-                self.allowImmediateLocalAddressReuse = true
+                self.allowLocalEndpointReuse = true
             case .allowRemoteHalfClosure:
                 self.allowRemoteHalfClosure = true
             case .disableAutoRead:
@@ -138,14 +138,14 @@ extension ChannelOptions {
             }
         }
         
-        /// Caller is consuming the knowledge that allowImmediateLocalAddressReuse was set or not.
+        /// Caller is consuming the knowledge that `allowLocalEndpointReuse` was set or not.
         /// The setting will nolonger be set after this call.
-        /// - Returns: If allowImmediateLocalAddressReuse was set.
-        public mutating func consumeAllowImmediateLocalAddressReuse() -> Types.NIOOptionValue<Void> {
+        /// - Returns: If `allowLocalEndpointReuse` was set.
+        public mutating func consumeAllowLocalEndpointReuse() -> Types.NIOOptionValue<Void> {
             defer {
-                self.allowImmediateLocalAddressReuse = false
+                self.allowLocalEndpointReuse = false
             }
-            return Types.NIOOptionValue<Void>(flag: self.allowImmediateLocalAddressReuse)
+            return Types.NIOOptionValue<Void>(flag: self.allowLocalEndpointReuse)
         }
         
         /// Caller is consuming the knowledge that disableAutoRead was set or not.
@@ -170,7 +170,7 @@ extension ChannelOptions {
         
         func applyFallbackMapping(_ universalBootstrap: NIOClientTCPBootstrap) -> NIOClientTCPBootstrap {
             var result = universalBootstrap
-            if self.allowImmediateLocalAddressReuse {
+            if self.allowLocalEndpointReuse {
                 result = result.channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
             }
             if self.allowRemoteHalfClosure {
--- a/Sources/NIO/ConvenienceOptionSupport.swift
+++ b/Sources/NIO/ConvenienceOptionSupport.swift
@@ -168,15 +168,15 @@ extension ChannelOptions {
             return Types.NIOOptionValue<Void>(flag: self.allowRemoteHalfClosure)
         }
         
-        func applyFallbackMapping(_ universalBootstrap: NIOClientTCPBootstrap) -> NIOClientTCPBootstrap {
+        mutating func applyFallbackMapping(_ universalBootstrap: NIOClientTCPBootstrap) -> NIOClientTCPBootstrap {
             var result = universalBootstrap
-            if self.allowLocalEndpointReuse {
+            if self.consumeAllowLocalEndpointReuse().isSet {
                 result = result.channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
             }
-            if self.allowRemoteHalfClosure {
+            if self.consumeAllowRemoteHalfClosure().isSet {
                 result = result.channelOption(ChannelOptions.allowRemoteHalfClosure, value: true)
             }
-            if self.disableAutoRead {
+            if self.consumeDisableAutoRead().isSet {
                 result = result.channelOption(ChannelOptions.autoRead, value: false)
             }
             return result
--- a/Sources/NIO/ConvenienceOptionSupport.swift
+++ b/Sources/NIO/ConvenienceOptionSupport.swift
@@ -14,22 +14,23 @@
 
 // MARK:  Universal Client Bootstrap
 extension NIOClientTCPBootstrapProtocol {
-    /// Apply any understood shorthand options to the bootstrap, removing them from the set of options if they are consumed.
+    /// Apply any understood convenience options to the bootstrap, removing them from the set of options if they are consumed.
     /// - parameters:
     ///     - options:  The options to try applying - the options applied should be consumed from here.
     /// - returns: The updated bootstrap with and options applied.
-    public func _applyChannelConvenienceOptions(_ options: inout ChannelOptions.NIOTCPShorthandOptions) -> Self {
+    public func _applyChannelConvenienceOptions(_ options: inout ChannelOptions.TCPConvenienceOptions) -> Self {
         // Default is to consume no options and not update self.
         return self
     }
 }
 
 extension NIOClientTCPBootstrap {
-    /// Specifies some `ChannelOption`s to be applied to the channel.
-    /// - See: channelOption
-    /// - Parameter options: Set of shorthand options to apply.
+    /// Specifies some `TCPConvenienceOption`s to be applied to the channel.
+    /// These are preferred over regular channel options as they are easier to use and restrict
+    /// options to those with a normal user would consider.
+    /// - Parameter options: Set of convenience options to apply.
     /// - Returns: The updated bootstrap (`self` being mutated)
-    public func channelConvenienceOptions(_ options: ChannelOptions.NIOTCPShorthandOptions) -> NIOClientTCPBootstrap {
+    public func channelConvenienceOptions(_ options: ChannelOptions.TCPConvenienceOptions) -> NIOClientTCPBootstrap {
         var optionsRemaining = options
         // First give the underlying a chance to consume options.
         let withUnderlyingOverrides =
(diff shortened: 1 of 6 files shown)


PeterAdams-A:
I have extracted the universal component from https://github.com/apple/swift-nio/pull/1475. 
I have made the fallback application use the universal type directly.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+            NIOClientTCPBootstrap(self, withUpdated: underlyingBootstrap._applyOptions(&optionsRemaining))
+        // Default apply any remaining options.
+        return optionsRemaining.applyFallbackMapping(withUnderlyingOverrides)
+    }
+}
+
+// MARK: Utility
+/// Has an option been set?
+/// Option has a value of generic type T.
+public enum NIOOptionValue<T> {
```
weissi, line 45: @Davidde94 / @glbrntt / @Lukasa shouldn't we nest this under `ChannelOptions.Types`?

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+    /// The option was not set.
+    case notSet
+    /// The option was set with a value of type T.
+    case set(T)
+}
+
+public extension NIOOptionValue where T == () {
```
weissi, line 52: please don't use `public extension`. Someone might add a (meant to be `internal`) things below and it becomes public by accident without a `public` appearing in code review.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+            case .notSet:
+                return false
+            case .set(()):
+                return true
+            }
+        }
+    }
+}
+
+private extension NIOOptionValue where T == () {
```
weissi, line 66: please migrate the `private` in front of the `init`. Any word in front of `extension` is an anti-pattern because it hides the actual access modifier from the func/property.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+            self = .set(())
+        } else {
+            self = .notSet
+        }
+    }
+}
+
+// MARK: TCP - data
+/// A TCP channel option which can be applied to a bootstrap using shorthand notation.
+public struct NIOTCPShorthandOption: Hashable {
```
weissi, line 78: @Davidde94 / @glbrntt / @Lukasa do we actually want/need this top-level?

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+    /// peer shuts down its send stream, or whether it will remain open. If set to `false` (the default), the `Channel`
+    /// will be closed automatically if the remote peer shuts down its send stream. If set to true, the `Channel` will
+    /// not be closed: instead, a `ChannelEvent.inboundClosed` user event will be sent on the `ChannelPipeline`,
+    /// and no more data will be received.
+    public static let allowRemoteHalfClosure =
+        NIOTCPShorthandOption(.allowRemoteHalfClosure)
+}
+
+/// A set of `NIOTCPShorthandOption`s
+public struct NIOTCPShorthandOptions : ExpressibleByArrayLiteral, Hashable {
```
weissi, line 110: extra space before `:`

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+        return self
+    }
+}
+
+extension NIOClientTCPBootstrap {
+    /// Specifies some `ChannelOption`s to be applied to the channel.
+    /// - See: channelOption
+    /// - Parameter options: Set of shorthand options to apply.
+    /// - Returns: The updated bootstrap (`self` being mutated)
+    public func options(_ options: NIOTCPShorthandOptions) -> NIOClientTCPBootstrap {
```
weissi, line 32: @Davidde94 / @glbrntt / @Lukasa this function name needs to be bike-shed. It's a new recommended-to-use public API. I think it needs to have the word `channel` in there.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+        case reuseAddr
+        case disableAutoRead
+        case allowRemoteHalfClosure
+    }
+}
+
+/// Approved shorthand options.
+extension NIOTCPShorthandOption {
+    /// Allow immediately reusing a local address.
+    public static let allowImmediateLocalEndpointAddressReuse = NIOTCPShorthandOption(.reuseAddr)
```
weissi, line 95: @Davidde94 / @glbrntt / @Lukasa these three names need bike-shedding.

Sources/NIO/ConvenienceOptionSupport.swift
weissi: `channelOptions` is indeed not good imho because it's confusing to have `channelOption` and `channelOptions` being totally different things.

But we could have `channelConvenienceOptions` or something like that.

Sources/NIO/ConvenienceOptionSupport.swift
Davidde94: Not sure I'm a fan of `endpoint` being in the name, I think the others are fine IMO

Sources/NIO/ConvenienceOptionSupport.swift
Davidde94: Makes sense

Sources/NIO/ConvenienceOptionSupport.swift
Davidde94: Maybe invert `disableAutoRead` to `enableAutoRead`. It feels a bit weird to have "allow", "allow", "don't allow"

Sources/NIO/ConvenienceOptionSupport.swift
Lukasa: Some thoughts:

- `generalChannelOptions`
- `genericChannelOptions`
- `universalChannelOptions`
- `shorthandChannelOptions`

Sources/NIO/ConvenienceOptionSupport.swift
Lukasa: Yes. Down with top-level types that users should never actually use.

Sources/NIO/ConvenienceOptionSupport.swift
Lukasa: Nope, under `ChannelOptions` it goes.

Sources/NIO/ConvenienceOptionSupport.swift
Lukasa: I think endpoint and address are redundant, pick one. Network.framework uses `Endpoint` which may be prior art justifying that name.

Lukasa: Given that `autoRead` is on by default, I don't think we can have an option to enable it without changing that default, which we cannot safely do in a semver minor.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,178 @@
...
+        defer {
+            self.disableAutoRead = false
+        }
+        return NIOOptionValue<Void>(flag: self.disableAutoRead)
+    }
+    
+    /// Caller is consuming the knowledge that allowRemoteHalfClosure was set or not.
+    /// The setting will nolonger be set after this call.
+    /// - Returns: If allowRemoteHalfClosure was set.
+    public mutating func consumeAllowRemoteHalfClosure() -> NIOOptionValue<Void> {
```
Lukasa, line 158: Are these three consume functions necessary? The fallback just directly looks at the stored properties on the struct: why shouldn't other use-cases do the same?

Sources/NIO/ConvenienceOptionSupport.swift
PeterAdams-A: Agree - ship has sailed autoRead defaults.
I think I'll go with Address - I like it more than Endpoint - `sockaddr_in` feels like prior art for it :-)

Sources/NIO/ConvenienceOptionSupport.swift
PeterAdams-A: They're not strictly necessary but they do push the user towards doing the right thing of removing them from the set when applying.

Sources/NIO/ConvenienceOptionSupport.swift
PeterAdams-A: I think general, generic and universal are all words used elsewhere to mean something else and don't feel like the right fit.
Shorthand or Convenience both feel fine.  Happy with either.

Sources/NIO/ConvenienceOptionSupport.swift
PeterAdams-A, line 110: No-one mentioned pushing NIOTCPShorthandOptions into ChannelOptions - I'll assume that was an oversight as it feels like it should logically live there with everything else.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+        // Default apply any remaining options.
+        return optionsRemaining.applyFallbackMapping(withUnderlyingOverrides)
+    }
+}
+
+// MARK: Utility
+extension ChannelOptions.Types {
+    /// Has an option been set?
+    /// Option has a value of generic type T.
+    public enum NIOOptionValue<T> {
```
weissi, line 47: - doesn't need the `NIO` prefixed because now nested
- should contain "convenience". So maybe `ConvenienceOptionValue`?
- `T` isn't a great generic parameter name, maybe `Value` or `ValueType` or so?

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            self = .notSet
+        }
+    }
+}
+
+// MARK: TCP - data
+extension ChannelOptions {
+    /// A TCP channel option which can be applied to a bootstrap using shorthand notation.
+    public struct NIOTCPShorthandOption: Hashable {
```
weissi, line 82: shorthand vs convenience

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+        fileprivate var data: ShorthandOption
+        
+        private init(_ data: ShorthandOption) {
+            self.data = data
+        }
+        
+        fileprivate enum ShorthandOption: Hashable {
+            case reuseAddr
```
weissi, line 90: these differ from the public names which is confusing

PeterAdams-A:
Applied consensus name of allowLocalEndpointReuse

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+        return self
+    }
+}
+
+extension NIOClientTCPBootstrap {
+    /// Specifies some `ChannelOption`s to be applied to the channel.
+    /// - See: channelOption
+    /// - Parameter options: Set of shorthand options to apply.
+    /// - Returns: The updated bootstrap (`self` being mutated)
+    public func channelConvenienceOptions(_ options: ChannelOptions.NIOTCPShorthandOptions) -> NIOClientTCPBootstrap {
```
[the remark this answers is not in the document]
PeterAdams-A: @Lukasa has a weak preference for shorthand,
David liked convenience as it pairs with the term convenience initialisers.
I'll vote convenience too making @glbrntt vote irrelevant.

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+                return false
+            case .set(()):
+                return true
+            }
+        }
+    }
+}
+
+extension ChannelOptions.Types.ConvenienceOptionValue where ValueType == () {
+    fileprivate init(flag: Bool) {
```
weissi, line 71: [just in case you're making more changes anyway], I'd remove the `flag:` label, ie `_ flag: Bool`. It doesn't add value

Sources/NIO/UniversalBootstrapSupport.swift
```diff
@@ -158,6 +165,11 @@ public struct NIOClientTCPBootstrap {
         self.underlyingBootstrap = bootstrap
         self.tlsEnablerTypeErased = tlsEnabler
     }
+    
+    internal init(_ original : NIOClientTCPBootstrap, withUpdated underlying : NIOClientTCPBootstrapProtocol) {
```
weissi, line 169: [just in case you're making more changes anyway], `with` is typically not used for those argument labels. I guess more standard would be `updating underlying: ...`

Sources/NIO/ConvenienceOptionSupport.swift
```diff
@@ -0,0 +1,186 @@
...
+    public struct TCPConvenienceOptions: ExpressibleByArrayLiteral, Hashable {
+        var allowLocalEndpointReuse = false
+        var disableAutoRead = false
+        var allowRemoteHalfClosure = false
+        
+        /// Construct from an array literal.
+        @inlinable
+        public init(arrayLiteral elements: TCPConvenienceOption...) {
+            for element in elements {
+                add(element)
```
weissi, line 126: missing `self.`

Sources/NIO/ConvenienceOptionSupport.swift
PeterAdams-A: It distinguishes from having ValueType == Bool.

Tests/NIOTests/UniversalBootstrapSupportTest.swift
```diff
@@ -135,4 +135,98 @@ class UniversalBootstrapSupportTest: XCTestCase {
...
+        let bothFake = FakeBootstrap()
+        _ = NIOClientTCPBootstrap(bothFake, tls: NIOInsecureNoTLS())
+            .channelConvenienceOptions([.allowRemoteHalfClosure, .allowLocalEndpointReuse])
+        XCTAssertTrue(bothFake.convenienceOptionConsumed)
+        XCTAssertTrue(bothFake.regularOptionsSeen)
+    }
+}
+
+// Prove we've not broken implementors of NIOClientTCPBootstrapProtocol by adding a new method.
+private class UniversalWithoutNewMethods : NIOClientTCPBootstrapProtocol {
```
PeterAdams-A, line 204: @weissi - Class missing _applyChannelConvenienceOptions function to prove the API change will not break anyone.

weissi: @Lukasa I'm guessing that you'll eventually merge this PR as your approval is still outstanding so I guess this is most relevant to you, this is the proof that we didn't actually break API despite what api-digester is saying.

PR #1552, opened 2020-06-09
Explicit Congestion Notification receive path

PeterAdams-A:
Explicit Congestion Notification receive path

### Motivation:

Surfacing ECN data from a UDP channel will be important for future protocols.

### Modifications:

Add methods to process cmsghdrs to both linux and Darwin C modules. 
Iterate the headers received to find ECN data.
Report upwards through AddressedEnvelope.Metadata

### Result:

It is possible to receive explicit congestion notifications


Sources/CNIODarwin/shim.c
```diff
@@ -50,4 +50,25 @@ int CNIODarwin_recvmmsg(int sockfd, CNIODarwin_mmsghdr *msgvec, unsigned int vle
...
 
+struct cmsghdr *CNIODarwin_CMSG_FIRSTHDR(const struct msghdr *mhdr) {
+    return CMSG_FIRSTHDR(mhdr);
+}
+
+struct cmsghdr *CNIODarwin_CMSG_NXTHDR(const struct msghdr *mhdr, const struct cmsghdr *cmsg) {
+    return CMSG_NXTHDR(mhdr, cmsg);
+}
+
+const unsigned char *CNIODarwin_CMSG_DATA(const struct cmsghdr *cmsg) {
```
Lukasa, line 61: Can we add assertions that the pointers passed to these three functions aren't `NULL`?

Sources/CNIOLinux/shim.c
```diff
@@ -112,4 +112,24 @@ int CNIOLinux_CPU_ISSET(int cpu, cpu_set_t *set) {
...
+
+struct cmsghdr *CNIOLinux_CMSG_FIRSTHDR(const struct msghdr *mhdr) {
+    return CMSG_FIRSTHDR(mhdr);
+}
+
+struct cmsghdr *CNIOLinux_CMSG_NXTHDR(struct msghdr *mhdr, struct cmsghdr *cmsg) {
+    return CMSG_NXTHDR(mhdr, cmsg);
+}
+
+const unsigned char *CNIOLinux_CMSG_DATA(const struct cmsghdr *cmsg) {
```
Lukasa, line 124: Same note on null pointers here too.

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -21,11 +21,12 @@ public struct AddressedEnvelope<DataType> {
     public var remoteAddress: SocketAddress
     public var data: DataType
     /// Any metadata associated with this `AddressedEnvelope`
-    public var metadata: Metadata? = nil
+    public var metadata: Metadata?
 
-    public init(remoteAddress: SocketAddress, data: DataType) {
+    public init(remoteAddress: SocketAddress, data: DataType, metadata: Metadata? = nil) {
```
Lukasa, line 26: This change is not backwards compatible, we need to add a new initializer.

Sources/NIO/Socket.swift
```diff
@@ -170,6 +170,59 @@ typealias IOVector = iovec
...
+                if case .processed = result {
+                    controlMessageReceiver.map { foreachControlMessage(in: messageHeader, $0)}
+                }
+                
+                return result
+            }
+        }
+    }
+    
+    private func foreachControlMessage(in messageHeader: msghdr,
```
Lukasa, line 211: It would be much nicer to design this to have a Swift wrapper type around the control message data. The properties of this type would be:

1. It would be a `Collection`. The `Index` type of this collection would be a pointer to the individual control message, and it would be able to implement `index(after:)` by way of `CMSG_NXTHDR`.
2. The element types would be a type that provides the level, type, and pointer to the data. This type should itself be named and have names for these two fields (mystery tuples are gross).

This then allows us to reimplement `foreachControlMessage` with simple straightforward iteration.

Sources/NIO/SocketChannel.swift
```diff
@@ -489,15 +500,28 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
         var rawAddressLength = socklen_t(MemoryLayout<sockaddr_storage>.size)
         var buffer = self.recvAllocator.buffer(allocator: self.allocator)
         var readResult = ReadResult.none
+        
+        let controlBytes = self.reportExplicitCongestionNotifications ?
+                // Guess at max 4 int32 payload messages.
+                UnsafeMutableRawBufferPointer.allocate(
```
Lukasa, line 506: This allocation can actually be hoisted up to be on the event loop. We do this already for writing (see `SelectableEventLoop` which has a number of these pointers). As only one read can be active on an event loop at any given time we can guarantee that these pointers will not be clobbered during the read operation.

Sources/NIO/SocketChannel.swift
```diff
@@ -520,6 +548,47 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
         }
         return readResult
     }
+    
+    private struct ControlMessageReceiver {
```
Lukasa, line 552: If we add a collection type for control messages, this becomes a simple struct that takes that collection as input and consumes it, turning it into simple values.

Sources/NIO/SocketChannel.swift
```diff
@@ -520,6 +548,47 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+        
+        mutating func receiveMessage(level: Int32, type: Int32, data: UnsafeBufferPointer<UInt8>?) {
+            #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+            let ipv4TosType = IP_RECVTOS
+            #else
+            let ipv4TosType = IP_TOS    // Linux
+            #endif
+            if level == IPPROTO_IP && type == ipv4TosType {
+                if let data = data {
+                    assert(data.count == 1)
```
Lukasa, line 563: Can these `assert`s have companion `precondition`s that enforce that `data.count` is _at least_ the size we want? Having the exact assert is good for debug time to spot assumptions in our code, but the `precondition` should be present so that instead of invalid reads we crash.

Sources/NIO/System.swift
```diff
@@ -98,6 +99,25 @@ private let sysSendMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_m
 private let sysRecvMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_mmsghdr>?, CUnsignedInt, CInt, UnsafeMutablePointer<timespec>?) -> CInt = CNIODarwin_recvmmsg
 #endif
 
+#if os(Linux)
```
Lukasa, line 102: We can re-use the existing `#if` block above.

Sources/CNIODarwin/shim.c
PeterAdams-A: Will do - interestingly it is valid to pass `cmsg = NULL` to `CMSG_NXTHDR` on Darwin but not on Linux.

Sources/NIO/SocketChannel.swift
PeterAdams-A: For discussion - I've kept the collection in the recvmsg function for now to minimise the chances of the control bytes pointer escaping this scope.

Sources/NIO/Socket.swift
```diff
@@ -170,6 +170,111 @@ typealias IOVector = iovec
...
+            get {
+                var messageHeader = self.messageHeader
+                return withUnsafePointer(to: &messageHeader) { messageHeaderPtr in
+                    let firstCMsg = Posix.cmsgFirstHeader(inside: messageHeaderPtr)
+                    return ControlMessageIndex(cmsgPointer: firstCMsg)
+                }
+            }
+        }
+        
+        let endIndex: Index = ControlMessageIndex(cmsgPointer: nil)
```
Lukasa, line 252:
```suggestion
        let endIndex = ControlMessageIndex(cmsgPointer: nil)
```

Sources/NIO/Socket.swift
```diff
@@ -170,6 +170,111 @@ typealias IOVector = iovec
...
+        
+        func index(after: Index) -> Index {
+            var msgHdr = messageHeader
+            return withUnsafeMutablePointer(to: &msgHdr) { messageHeaderPtr in
+                return ControlMessageIndex(cmsgPointer:
+                                            Posix.cmsgNextHeader(inside: messageHeaderPtr, from: after.cmsgPointer))
```
Lukasa, line 258: The indentation here is a bit wild!

Sources/NIO/SocketChannel.swift
```diff
@@ -12,6 +12,14 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import CNIODarwin
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+import CNIOLinux
+#else
+let badOS = { fatalError("unsupported OS") }()
+#endif
```
Lukasa, line 21: Do we need this? I think it'd be better to fatal error only if you try to _use_ the new functionality.

Sources/NIO/SocketChannel.swift
```diff
@@ -415,10 +424,12 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
             let valueAsInt: Int32 = value as! Bool ? 1 : 0
             switch self.localAddress?.protocol {
             case .some(.inet):
+                reportExplicitCongestionNotifications = true
```
Lukasa, line 427:
```suggestion
                self.reportExplicitCongestionNotifications = true
```

Sources/NIO/SocketChannel.swift
```diff
@@ -415,10 +424,12 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
                 try self.socket.setOption(level: .ip,
                                           name: .ip_recv_tos,
                                           value: valueAsInt)
             case .some(.inet6):
+                reportExplicitCongestionNotifications = true
```
Lukasa, line 432:
```suggestion
                self.reportExplicitCongestionNotifications = true
```

Sources/NIO/SocketChannel.swift
```diff
@@ -489,36 +500,94 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
-            }
-            buffer.clear()
-
-            let result = try buffer.withMutableWritePointer {
-                try self.socket.recvfrom(pointer: $0, storage: &rawAddress, storageLen: &rawAddressLength)
+        
+        return try self.selectableEventLoop.withControlMessageBytes {
+            let controlBytes = self.reportExplicitCongestionNotifications ?
+                    $0:
+                    UnsafeMutableRawBufferPointer(start: nil, count: 0)
```
Lukasa, line 507: I think if you're splitting the ternary over three lines it's not really aiding clarity anymore. Better to replace with `if`/`else` I think.

Sources/NIO/SocketChannel.swift
Lukasa: I think there's no risk: so long as we have this essentially "consume" the collection in its init we're safe, there's only one place to audit.

Sources/NIO/SocketChannel.swift
```diff
@@ -489,36 +500,94 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+                case .processed(let bytesRead):
+                    assert(bytesRead > 0)
+                    assert(self.isOpen)
+                    let mayGrow = recvAllocator.record(actualReadBytes: bytesRead)
+                    readPending = false
+
+                    let msg = AddressedEnvelope(remoteAddress: rawAddress.convert(),
+                                                data: buffer,
+                                                metadata: self.reportExplicitCongestionNotifications ?
+                                                    AddressedEnvelope.Metadata(ecnState: controlMessageReceiver.ecnValue) :
```
Lukasa, line 533: Again, this ternary kinda sucks. Can `ControlMessageReceiver` take over the ternary for us in both cases?

Sources/NIO/SocketChannel.swift
Lukasa: I'd also like to see this move out into another file with the collection we just created.

Sources/NIO/SocketChannel.swift
```diff
@@ -489,36 +500,94 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+    
+    private struct ControlMessageReceiver {
+        var ecnValue: NIOExplicitCongestionNotificationState = .transportNotCapable // Default
+        
+        mutating func receiveMessage(_ controlMessage: Socket.ControlMessage) {
+            #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+            let ipv4TosType = IP_RECVTOS
+            #else
+            let ipv4TosType = IP_TOS    // Linux
+            #endif
```
Lukasa, line 558: Let's make this a `private static let` just to pull it out of the main method body.

Sources/CNIODarwin/shim.c
```diff
@@ -50,4 +51,27 @@ int CNIODarwin_recvmmsg(int sockfd, CNIODarwin_mmsghdr *msgvec, unsigned int vle
     errx(EX_SOFTWARE, "recvmmsg shim not implemented on Darwin platforms\n");
 }
 
+struct cmsghdr *CNIODarwin_CMSG_FIRSTHDR(const struct msghdr *mhdr) {
+    assert(mhdr != NULL);
+    return CMSG_FIRSTHDR(mhdr);
+}
+
+struct cmsghdr *CNIODarwin_CMSG_NXTHDR(const struct msghdr *mhdr, const struct cmsghdr *cmsg) {
+    assert(mhdr != NULL);
```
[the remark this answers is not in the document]
PeterAdams-A: Yes - NULL is valid for the second one.  
We don't use this so I could assert anyway.
RFC 2292 says that CMSG_NXTHDR(mhdr, NULL) is equivalent to CMSG_FIRSTHDR(mhdr)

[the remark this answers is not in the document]
Lukasa: There's inconsistency here: the assert is present on Linux and not on Darwin. I don't much mind which you want to go for, but they should be the same.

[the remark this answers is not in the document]
PeterAdams-A: Yeah - sadly we'll have to go the Linux way as it isn't null safe like Darwin, much like it also is less good for constness than Darwin.  Pesky Linux.

PR #1569, opened 2020-06-23
alloc tests: implement malloc_zone_*

weissi:
Motivation:

The Swift runtime is now using malloc_zone_*, we need to implement
replacements for these too. This is just a first pass, eventually, we
should implement _all_ replacements as `malloc_zone_memalign` which is
powerful enough to implement all others.

Modifications:

Provide new replacements.

Result:

Alloc tests work again on macOS.


Lukasa:
Looks like the new integration test suite doesn't build cleanly on Linux though. Sadly the script doesn't seem to actually give us the build output so you'll need to repro locally.

weissi:
@Lukasa requesting re-review because I now split Darwin & non-Darwin UNIX. This makes everything work on Darwin because on Darwin I now _always_ jump into the right functions (which isn't possible on Linux because there's no easy way to get hold of the original function whilst in the replacement for that function (`dlsym` etc allocate ...))

PR #1574, opened 2020-06-25
Add BaggageContext property to ChannelHandlerContext

slashmo:
⚠️ This PR is not intended to ever be merged, but rather meant as a playground to inspire a "real" PR once the `BaggageContext` has matured and was accepted by the SSWG.

Publicly expose a `BaggageContext` through the `ChannelHandlerContext`.

### Motivation:

As already discussed in slashmo/gsoc-swift-tracing#48, in order to share a `BaggageContext` between multiple handlers on the same channel, we want to store a `BaggageContext` on a `Channel`, which it publicly accessible through the `ChannelHandlerContext`.

### Modifications:

- NIO depends on [slashmo/gsoc-swift-baggage-context](https://github.com/slashmo/gsoc-swift-baggage-context)

### Result:

Users of `BaggageContext` and [related APIs](https://github.com/slashmo/gsoc-swift-tracing) will be able to access a `BaggageContext` through the `ChannelHandlerContext`. This has the side-effect of NIO now depending on [slashmo/gsoc-swift-baggage-context](slashmo/gsoc-swift-baggage-context).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,8 @@ var targets: [PackageDescription.Target] = [
                            "CNIODarwin",
                            "NIOConcurrencyHelpers",
                            "CNIOAtomics",
-                           "CNIOSHA1"]),
+                           "CNIOSHA1",
+                           "Baggage"]),
     .target(name: "NIOFoundationCompat", dependencies: ["NIO"]),
     .target(name: "CNIOAtomics", dependencies: []),
     .target(name: "CNIOSHA1", dependencies: []),
@@ -94,6 +95,7 @@ let package = Package(
         .library(name: "NIOTestUtils", targets: ["NIOTestUtils"]),
     ],
     dependencies: [
+        .package(url: "https://github.com/slashmo/gsoc-swift-baggage-context", .branch("main"))
     ],
     targets: targets
 )


slashmo:
@ktoso & @weissi This should now be a working version of what we discussed. I left out documentation until now if we want to move things around a bit first. For the tests I noticed that it'd be the first using the `XCTExpectation` API. Do you not use that on purpose? If so, I'm happy to adapt my test as well.

Sources/NIO/DeadChannel.swift
```diff
@@ -12,10 +12,14 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Baggage
+
 /// A `DeadChannelCore` is a `ChannelCore` for a `DeadChannel`. A `DeadChannel` is used as a replacement `Channel` when
 /// the original `Channel` is closed. Given that the original `Channel` is closed the `DeadChannelCore` should fail
 /// all operations.
 private final class DeadChannelCore: ChannelCore {
+    var baggage = BaggageContext()
```
ktoso, line 21: not sure about usage of DeadChannelCore, but sounds like it may be safer to never store baggage values in here, there's no real purpose to propagate using a dead channel it feels like?

ktoso:
@swift-server-bot add to whitelist

PR #1584, opened 2020-07-01
fix incorrect case when removing an element from a circular buffer

luoxiu:
### Motivation:

```swift
// CircularBuffer.swift
public mutating func remove(at position: Index) -> Element {
  // ...
  switch bufferIndex {
  // ...
  case self.indexBeforeHeadIdx():  // incorrect case, will never be executed.
    self.advanceTailIdx(by: -1)
    self.tailBackingIndex = self.indexBeforeTailIdx()
    self._buffer[bufferIndex] = nil
  }
}
```

### Modifications:

Should be:

```swift
  case self.indexBeforeTailIdx():
    self.advanceTailIdx(by: -1)
    self._buffer[bufferIndex] = nil
```

### Result:

more efficient to remove the last element.

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -656,9 +656,8 @@ extension CircularBuffer: RangeReplaceableCollection {
         case self.headBackingIndex:
             self.advanceHeadIdx(by: 1)
             self._buffer[bufferIndex] = nil
-        case self.indexBeforeHeadIdx():
+        case self.indexBeforeTailIdx():
             self.advanceTailIdx(by: -1)
-            self.tailBackingIndex = self.indexBeforeTailIdx()
             self._buffer[bufferIndex] = nil
         default:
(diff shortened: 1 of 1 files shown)


Lukasa:

luoxiu:
In fact, there are already tests that cover the use case:

```swift
func testRemoveAtLastPosition() {
    var ring = CircularBuffer<Int>(initialCapacity: 4)
    for idx in 0..<7 {
        ring.prepend(idx)
    }

    let last = ring.remove(at: ring.index(ring.endIndex, offsetBy: -1))
    XCTAssertEqual(0, last)
    XCTAssertEqual(1, ring.last)
}

// and
func testRemoveAtTailIdx0() {
  // ...
  XCTAssertEqual(99, ring.remove(at: ring.index(ring.endIndex, offsetBy: -1)))
  // ...
}
```

The old path is never executed.

Lukasa:
Heh, so there are.

PR #1585, opened 2020-07-06
Explicit Congestion Notification for UDP

PeterAdams-A:
Explicit Congestion Notification for UDP

### Motivation:

Network congestion can be controlled without packet loss if explicit congestion notification messages are understood.

### Modifications:

Add a ChannelOption to turn on receiving notifications via AddressedEnvelope.Metadata.
Receive single and vector paths to get notification and surface.
Send paths for both single and vector sends.
Round trip tests.

I have tested allocations and they seem to be fine but the allocation tests for vector read and write were not stable enough to sensibly add to CI.

### Result:

Explicit Congestion Notifications can be both sent and received.


PeterAdams-A:
@Lukasa This contains https://github.com/apple/swift-nio/pull/1552 which in turn contains https://github.com/apple/swift-nio/pull/1473 - might be easier to review separately as it's a bit of a monster change - up to you.

Lukasa:
Yeah, I think I'm gonna delay reviewing this until we can land the other patches.

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -35,17 +35,12 @@ struct DatagramVectorReadManager {
...
+            self.controlMessageBuffers.deallocate()
 
             self.messageVector = .allocateAndInitialize(repeating: MMsgHdr(msg_hdr: msghdr(), msg_len: 0), count: newValue)
             self.ioVector = .allocateAndInitialize(repeating: IOVector(), count: newValue)
             self.sockaddrVector = .allocateAndInitialize(repeating: sockaddr_storage(), count: newValue)
-            self.controlMessageBuffers = .init(repeating: .init(start: nil, count: 0), count: messageCount)
-            for controlIndex in 0..<self.controlMessageBuffers.count {
-                self.controlMessageBuffers[controlIndex] = DatagramChannel.allocateControlMessageBuffer()
-            }
+            self.controlMessageBuffers = DatagramChannel.allocateControlMessageBuffer(msghdrCount: newValue)
```
Lukasa, line 43: Why are we placing this static method on `DatagramChannel`?

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -78,25 +86,40 @@ struct DatagramVectorReadManager {
...
+    ///     - reportExplicitCongestionNotifications: Should explicit congestion notifications be reported up using metadata.
+    func readFromSocket(socket: Socket,
+                        buffer: inout ByteBuffer,
+                        reportExplicitCongestionNotifications: Bool) throws -> ReadResult {
         assert(buffer.readerIndex == 0, "Buffer was not cleared between calls to readFromSocket!")
 
         let messageSize = buffer.capacity / self.messageCount
+        let controlBytesPerMessage =
+            (self.controlMessageBuffers.count / self.messageCount / MemoryLayout<cmsghdr>.alignment)
+            * MemoryLayout<cmsghdr>.alignment
```
Lukasa, line 98: Why does this divide out and then remultiply by the alignment of the cmsghdr? What's the goal, to pad out to that alignment value? Whatever the reason, let's throw a comment in here explaining why we did it.

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -148,9 +178,18 @@ struct DatagramVectorReadManager {
             // Next we extract the remote peer address.
             precondition(self.messageVector[i].msg_hdr.msg_namelen != 0, "Unexpected zero length peer name")
             let address: SocketAddress = self.sockaddrVector[i].convert()
+            
+            // Extract congestion information if requested.
+            let metadata: AddressedEnvelope<ByteBuffer>.Metadata?
+            if reportExplicitCongestionNotifications {
+                let controlMessagesReceived = UnsafeControlMessageCollection(messageHeader: messageVector[i].msg_hdr)
```
Lukasa, line 185: Why aren't we just using `self.messageVector` like we do on line 179, and removing the extra argument?

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -66,12 +66,15 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
...
                                                    addresses: UnsafeMutableBufferPointer<sockaddr_storage>,
                                                    storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>,
+                                                   controlMessageBuffers: UnsafeMutableRawBufferPointer,
                                                    _ body: (UnsafeMutableBufferPointer<MMsgHdr>) throws -> IOResult<Int>) throws -> IOResult<Int> {
     assert(msgs.count >= Socket.writevLimitIOVectors, "Insufficiently sized buffer for a maximal sendmmsg")
 
     // the numbers of storage refs that we need to decrease later.
     var c = 0
     var toWrite: Int = 0
+    let controlBytesPerMessage = (controlMessageBuffers.count / msgs.count / MemoryLayout<cmsghdr>.alignment)
```
Lukasa, line 76: Same note here about explaining the divide/multiply dance.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -66,12 +66,15 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
                                                    msgs: UnsafeMutableBufferPointer<MMsgHdr>,
                                                    addresses: UnsafeMutableBufferPointer<sockaddr_storage>,
                                                    storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>,
+                                                   controlMessageBuffers: UnsafeMutableRawBufferPointer,
```
Lukasa, line 69: Nit: the parameter name here is now a bit misleading.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -99,13 +102,19 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
             storageRefs[c] = storageRef.retain()
             let addressLen = p.copySocketAddress(addresses.baseAddress! + c)
             iovecs[c] = iovec(iov_base: UnsafeMutableRawPointer(mutating: ptr.baseAddress!), iov_len: numericCast(toWriteForThisBuffer))
+            
+            let controlBytesRaw = UnsafeMutableRawBufferPointer(
+                rebasing: controlMessageBuffers[(c * controlBytesPerMessage)..<((c+1) * controlBytesPerMessage)])
```
Lukasa, line 107: Can we write ourselves a helper for this slicing operation? It'd be helpful for avoiding repetition.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -375,14 +386,17 @@ final class PendingDatagramWritesManager: PendingWritesManager {
     ///     - iovecs: A pre-allocated array of `IOVector` elements
     ///     - addresses: A pre-allocated array of `sockaddr_storage` elements
     ///     - storageRefs: A pre-allocated array of storage management tokens used to keep storage elements alive during a vector write operation
+    ///     - controlMessageBuffers: A pre-allocated memory buffers for storing cmsghdr data during a vector write operation.
     init(msgs: UnsafeMutableBufferPointer<MMsgHdr>,
          iovecs: UnsafeMutableBufferPointer<IOVector>,
          addresses: UnsafeMutableBufferPointer<sockaddr_storage>,
-         storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>) {
+         storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>,
+         controlMessageBuffers: UnsafeMutableRawBufferPointer) {
```
Lukasa, line 394: Nit: parameter name here isn't terribly clear either.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -375,14 +386,17 @@ final class PendingDatagramWritesManager: PendingWritesManager {
     ///     - iovecs: A pre-allocated array of `IOVector` elements
     ///     - addresses: A pre-allocated array of `sockaddr_storage` elements
     ///     - storageRefs: A pre-allocated array of storage management tokens used to keep storage elements alive during a vector write operation
+    ///     - controlMessageBuffers: A pre-allocated memory buffers for storing cmsghdr data during a vector write operation.
```
Lukasa, line 389:
```suggestion
    ///     - controlMessageBuffers: A pre-allocated memory buffer for storing cmsghdr data during a vector write operation.
```

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -555,6 +561,11 @@ internal final class SelectableEventLoop: EventLoop {
             }
         }
     }
+    
+    func withControlMessageBytes<ReturnType>(
+        _ body: (UnsafeMutableRawBufferPointer) throws -> ReturnType) rethrows -> ReturnType {
+        return try body(self.controlMessageBuffers)
+    }
```
Lukasa, line 568: How useful is this abstraction method? It doesn't seem to add any clarity at all.

Sources/NIO/Socket.swift
```diff
@@ -240,6 +240,46 @@ typealias IOVector = iovec
         }
     }
 
+    func recvmsg(pointer: UnsafeMutableRawBufferPointer,
+                 storage: inout sockaddr_storage,
+                 storageLen: inout socklen_t,
+                 controlBytes: inout Slice<UnsafeMutableRawBufferPointer>,
+                 controlMessagesReceived: inout UnsafeControlMessageCollection?) throws -> IOResult<Int> {
```
Lukasa, line 247: Is it really better to have two of these rather than an optional parameter to the above method? Do we even ever call the above method now?

PR #1590, opened 2020-07-13
Add methods to create and parse cmsghdr structured data.

PeterAdams-A:
Motivation:

cmsghdrs can be used to send an receive extra data on UDP packets.
For example ECN data.

Modifications:

Map in Linux and Darwin versions of cmsghdr macros as functions.
Create strctures for holding a received collection; parsing ecn
data from a received collection and building a collection suitable
for sending.

Result:

Functions to manipulate cmsghdr and data exist.


PeterAdams-A:
@Lukasa The obvious points for discussion are
1)  Send and receive collections are currently different - this is because I felt more confortable using the msghdr based functions when I had received them that way.
2) No windows support - should I implement blank functions which just assert on windows?

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,224 @@
...
+    mutating func appendControlMessage<PayloadType>(level: Int32,
+                                          type: Int32,
+                                          payload: PayloadType) {
+        let writableBuffer = UnsafeMutableRawBufferPointer(rebasing: self.controlBytes[writePosition...])
+        
+        let requiredSize = Posix.cmsgSpace(payloadSize: MemoryLayout.stride(ofValue: payload))
+        precondition(writableBuffer.count >= requiredSize, "Insufficient size for cmsghdr and data")
+        
+        let bufferBase = writableBuffer.baseAddress!
+        let cmsghdrPtr = bufferBase.bindMemory(to: cmsghdr.self, capacity: 1)
```
Lukasa, line 171: Any use of the memory binding APIs requires considerable caution and some commenting to explain why it's fine (or indeed, not fine).

In this case, we want the bound pointer because we need it for `cmsgData`, so we need to bind it. However, we drop the bound pointer after this function call. As we own the original pointer, if we're confident we never bind this to anything _else_ at the same time we're good to go. We know we don't (no other bound pointer to this memory can exist at the same time), so we're safe, but we need to write all of this down.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,224 @@
...
+        cmsghdrPtr.pointee.cmsg_level = level
+        cmsghdrPtr.pointee.cmsg_type = type
+        cmsghdrPtr.pointee.cmsg_len = .init(Posix.cmsgLen(payloadSize: MemoryLayout.size(ofValue: payload)))
+        
+        let dataPointer = Posix.cmsgData(for: .some(cmsghdrPtr))
+        precondition(dataPointer!.count >= MemoryLayout<PayloadType>.stride)
+        let dataPointerBase = dataPointer!.baseAddress!
+        let dataPointerTyped = dataPointerBase.bindMemory(to: PayloadType.self, capacity: 1)
```
Lukasa, line 179: Relatedly, this binding is probably not the right choice. `UnsafeMutableRawBufferPointer.store` is a better choice here.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,224 @@
...
+        }
+        return UnsafeMutableRawBufferPointer(rebasing: self.controlBytes[0 ..< self.writePosition])
+    }
+    
+}
+
+extension UnsafeOutboundControlBytes {
+    /// - address:  Either local or remote will do, we just use it for extracting the right protocol.
+    internal mutating func appendExplicitCongestionState(metadata: AddressedEnvelope<ByteBuffer>.Metadata?,
+                                                         address: SocketAddress?) {
```
Lukasa, line 198: Rather than take the fairly heavy `SocketAddress` struct, can we just take the `NIOBSDSocket.ProtocolFamily` type? That's the question we actually have anyway.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,224 @@
...
+        if let metadata = metadata {
```
Lukasa, line 199: Replace this with `guard let`.

Sources/NIO/System.swift
```diff
@@ -373,7 +392,9 @@ internal enum Posix {
     }
     
     @inline(never)
-    public static func sendmsg(descriptor: CInt, msgHdr: UnsafePointer<msghdr>, flags: CInt) throws -> IOResult<ssize_t> {
+    public static func sendmsg(descriptor: CInt,
+                               msgHdr: UnsafePointer<msghdr>,
+                               flags: CInt) throws -> IOResult<ssize_t> {
```
Lukasa, line 397: Let's revert this, we don't need the churn.

Sources/NIO/System.swift
```diff
@@ -535,6 +556,50 @@ internal extension Posix {
     }
 }
 
+internal extension Posix {
```
Lukasa, line 559: No need for this to be an extension.

Sources/NIO/System.swift
```diff
@@ -535,6 +556,50 @@ internal extension Posix {
...
+    static func cmsgFirstHeader(inside msghdr: UnsafePointer<msghdr>?) -> UnsafeMutablePointer<cmsghdr>? {
+        return sysCmsgFirstHdr(msghdr)
+    }
+    
+    static func cmsgNextHeader(inside msghdr: UnsafeMutablePointer<msghdr>?, from: UnsafeMutablePointer<cmsghdr>?)
```
Lukasa, line 564: Let's break the line in the arguments as well.

Sources/NIO/System.swift
```diff
@@ -535,6 +556,50 @@ internal extension Posix {
...
+        -> UnsafeMutablePointer<cmsghdr>? {
+        return sysCmsgNxtHdr(msghdr, from)
+    }
+    
+    static func cmsgData(for header: UnsafePointer<cmsghdr>?) -> UnsafeRawBufferPointer? {
+        let dataPointer = sysCmsgData(header)
+        if let header = header {
```
Lukasa, line 571: This `if let` is really in error: we aren't allowed to call `sysCmsgData` with a nil pointer. Let's remove the nullability in the API to reflect that, and save ourselves the branch.

Sources/NIO/System.swift
```diff
@@ -535,6 +556,50 @@ internal extension Posix {
     }
 }
 
+internal extension Posix {
+    static func cmsgFirstHeader(inside msghdr: UnsafePointer<msghdr>?) -> UnsafeMutablePointer<cmsghdr>? {
```
Lukasa, line 560: This argument is forbidden to be `nil` (assert in the C layer), so let's not accept `nil` here.

Sources/NIO/System.swift
Lukasa: These pointers are forbidden to be `nil` (asserts in the C layer), so let's forbid `nil` here as well.

Sources/NIO/System.swift
```diff
@@ -535,6 +556,50 @@ internal extension Posix {
...
+            // Linux and Darwin use different types for cmsg_len.
+            let length = size_t(header.pointee.cmsg_len) - cmsgLen(payloadSize: 0)
+            let buffer = UnsafeRawBufferPointer(start: dataPointer, count: Int(length))
+            return buffer
+        } else {
+            return nil
+        }
+    }
+    
+    static func cmsgData(for header: UnsafeMutablePointer<cmsghdr>?) -> UnsafeMutableRawBufferPointer? {
```
Lukasa, line 581: Same note here about `nil`, which kills this branch for us as well.

Tests/NIOTests/ControlMessageTests.swift
```diff
@@ -0,0 +1,70 @@
...
+                            msg_flags: 0)
+        self.init(messageHeader: msgHdr)
+    }
+}
+
+class ControlMessageTests: XCTestCase {
+    var encoder: UnsafeOutboundControlBytes!
+    
+    override func setUp() {
+        let encoderBytes = UnsafeMutableRawBufferPointer.allocate(byteCount: 1000,
```
Lukasa, line 35: These tests leak 1kb of memory each, as they never free this pointer.

Tests/NIOTests/ControlMessageTests.swift
```diff
@@ -0,0 +1,70 @@
...
+    
+    func testEncodeDecode1() {
+        self.encoder.appendControlMessage(level: 1, type: 2, payload: 3)
+        let encodedBytes = self.encoder.validControlBytes
+        
+        let decoder = UnsafeControlMessageCollection(controlBytes: encodedBytes)
+        XCTAssertEqual(decoder.count, 1)
+        XCTAssertEqual(decoder.first!.level, 1)
+        XCTAssertEqual(decoder.first!.type, 2)
+        XCTAssertEqual(decoder.first!.data!.count, MemoryLayout<Int>.size)
```
Lukasa, line 52: It might be better to turn this into a test that gathers an `Array` of decoded messages, as that'll better test that the iterator works.

We should also confirm the value is right.

Tests/NIOTests/ControlMessageTests.swift
```diff
@@ -0,0 +1,70 @@
...
+        let encodedBytes = self.encoder.validControlBytes
+        
+        let decoder = UnsafeControlMessageCollection(controlBytes: encodedBytes)
+        XCTAssertEqual(decoder.count, 2)
+        XCTAssertEqual(decoder.first!.level, 1)
+        XCTAssertEqual(decoder.first!.type, 2)
+        XCTAssertEqual(decoder.first!.data!.count, MemoryLayout<Int>.size)
+        XCTAssertEqual(decoder[decoder.index(after: decoder.startIndex)].level, 4)
+        XCTAssertEqual(decoder[decoder.index(after: decoder.startIndex)].type, 5)
+        XCTAssertEqual(decoder[decoder.index(after: decoder.startIndex)].data!.count, MemoryLayout<Int>.size)
```
Lukasa, line 67: Same note here, let's just use an iterator instead.

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+        0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, // IPPROTO_IP, IP_TOS
+        0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00  // ECT-1 (1 byte)
+    ]
+    private static let cmsghdr_secondStartPosition = 32
+    private static let cmsghdr_firstDataStart = 16
+    private static let cmsghdr_firstDataCount = 12
+    private static let cmsghdr_secondDataCount = 1
+    private static let cmsghdr_firstType = IP_PKTINFO
+    private static let cmsghdr_secondType = IP_TOS
+    #endif
```
Lukasa, line 83: Can we add a catch-all `#else` branch that emits a compile error with `#error`? That's a lot better than the missing symbols.

That, or we should skip these tests on all platforms not covered in the above branch.

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+
+    func testCmsgFirstHeader() {
+        var exampleCmsgHrd = SystemTest.cmsghdrExample
```
Lukasa, line 86:
```suggestion
        var exampleCmsgHdr = SystemTest.cmsghdrExample
```

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+        exampleCmsgHrd.withUnsafeMutableBytes { pCmsgHdr in
```
Lukasa, line 87: Is there any reason to be using mutable bytes here?

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+            withUnsafePointer(to: msgHdr) { pMsgHdr in
+                let result = Posix.cmsgFirstHeader(inside: pMsgHdr)
+                XCTAssertEqual(pCmsgHdr.baseAddress, result)
+            }
+        }
+    }
+    
+    func testCMsgNextHeader() {
+        var exampleCmsgHrd = SystemTest.cmsghdrExample
+        exampleCmsgHrd.withUnsafeMutableBytes { pCmsgHdr in
```
Lukasa, line 101: Same mutable note here too.

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+                XCTAssertEqual(expectedSecondSlice.baseAddress, second)
+                let third = Posix.cmsgNextHeader(inside: pMsgHdr, from: second)
+                XCTAssertEqual(third, nil)
+            }
+        }
+    }
+    
+    func testCMsgData() {
+        var exampleCmsgHrd = SystemTest.cmsghdrExample
+        exampleCmsgHrd.withUnsafeMutableBytes { pCmsgHdr in
```
Lukasa, line 120: Same mutable note here too.

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+                                        SystemTest.cmsghdr_firstDataStart + SystemTest.cmsghdr_firstDataCount)])
+                XCTAssertEqual(expecedFirstData.baseAddress, firstData?.baseAddress)
+                XCTAssertEqual(expecedFirstData.count, firstData?.count)
+            }
+        }
+    }
+    
+    func testCMsgCollection() {
+        var exampleCmsgHrd = SystemTest.cmsghdrExample
+        exampleCmsgHrd.withUnsafeMutableBytes { pCmsgHdr in
```
Lukasa, line 139: Same mutable note here too.

Sources/NIO/System.swift
Lukasa: Should we rename `from` to `after`?

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+            var msgHdr = msghdr()
+            msgHdr.msg_control = pCmsgHdr.baseAddress
+            msgHdr.msg_controllen = .init(pCmsgHdr.count)
+
+            withUnsafeMutablePointer(to: &msgHdr) { pMsgHdr in
+                let first = Posix.cmsgFirstHeader(inside: pMsgHdr)
+                let second = Posix.cmsgNextHeader(inside: pMsgHdr, from: first)
+                let expectedSecondSlice = UnsafeMutableRawBufferPointer(
+                    rebasing: pCmsgHdr[SystemTest.cmsghdr_secondStartPosition...])
+                XCTAssertEqual(expectedSecondSlice.baseAddress, second)
```
Lukasa, line 111: Let's not bother creating a slice if we don't need one, let's just use `pCmsgHdr.baseAddress! + SystemTest.cmsghdr_secondStartPosition`

Tests/NIOTests/SystemTest.swift
```diff
@@ -46,4 +46,115 @@ class SystemTest: XCTestCase {
...
+    }
+    
+    func testCMsgNextHeader() {
+        var exampleCmsgHrd = SystemTest.cmsghdrExample
+        exampleCmsgHrd.withUnsafeMutableBytes { pCmsgHdr in
+            var msgHdr = msghdr()
+            msgHdr.msg_control = pCmsgHdr.baseAddress
+            msgHdr.msg_controllen = .init(pCmsgHdr.count)
+
+            withUnsafeMutablePointer(to: &msgHdr) { pMsgHdr in
```
Lukasa, line 106: Do we need a mutable pointer here?

PeterAdams-A:
@Lukasa I have taken most of your suggested changes.  There are a few where I have not done but have replied to your comment.

Lukasa:
There are no visible comment replies @PeterAdams-A: did you submit them?

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,224 @@
...
+    var level: Int32
+    var type: Int32
+    var data: UnsafeRawBufferPointer?
+}
+
+/// Collection representation of `cmsghdr` structures and associated data from `recvmsg`
+/// Unsafe as captures pointers held in msghdr structure which must not escape scope of validity.
+struct UnsafeControlMessageCollection: Collection {
+    typealias Index = ControlMessageIndex
+    typealias Element = UnsafeControlMessage
```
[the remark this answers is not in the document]
Lukasa: Sorry, I should have been clearer: remove the `typealias` and just write the return value on the `subscript` instead. Irregardless, it doesn't matter and we can leave it here, I don't mind it much.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+struct ControlMessageParser {
+    var ecnValue: NIOExplicitCongestionNotificationState = .transportNotCapable // Default
+    
+    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    private static let ipv4TosType = IP_RECVTOS
+    #else
+    private static let ipv4TosType = IP_TOS    // Linux
+    #endif
+
+    static func readCInt(data: UnsafeRawBufferPointer) -> CInt {
```
Lukasa, line 101: Prefix with an underscore to indicate this is not supposed to be part of the public API of this type.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+            self = .congestionExperienced
+        default:
+            self = .transportNotCapable
+        }
+    }
+}
+
+extension CInt {
+    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    fileprivate static let notCapableValue = IPTOS_ECN_NOTECT
```
Lukasa, line 146: This can be `private`.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+            self = IPTOS_ECN_ECT1
+        case .congestionExperienced:
+            self = IPTOS_ECN_CE
+        }
+    }
+}
+
+struct UnsafeOutboundControlBytes {
+    private var controlBytes: UnsafeMutableRawBufferPointer
+    private var writePosition: UnsafeMutableRawBufferPointer.Index = 0
```
Lukasa, line 168: For the sake of keeping this as correct as possible, let's initialise this in `init` with `self.controlBytes.startIndex`.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+        precondition(writableBuffer.count >= requiredSize, "Insufficient size for cmsghdr and data")
+        
+        let bufferBase = writableBuffer.baseAddress!
+        // Binding to cmsghdr is safe here as this is the only place where we bind to non-Raw.
+        let cmsghdrPtr = bufferBase.bindMemory(to: cmsghdr.self, capacity: 1)
+        cmsghdrPtr.pointee.cmsg_level = level
+        cmsghdrPtr.pointee.cmsg_type = type
+        cmsghdrPtr.pointee.cmsg_len = .init(Posix.cmsgLen(payloadSize: MemoryLayout.size(ofValue: payload)))
+        
+        let dataPointer = Posix.cmsgData(for: cmsghdrPtr)
```
Lukasa, line 196: We can bring the force-unwrap to here and save ourselves two more below.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+            break
+        }
+    }
+}
+
+extension AddressedEnvelope.Metadata {
+    /// It's assumed the caller has checked that congestion information is required before calling.
+    internal init(from controlMessagesReceived: UnsafeControlMessageCollection) {
+        var controlMessageReceiver = ControlMessageParser()
+        controlMessagesReceived.forEach { controlMessage in controlMessageReceiver.receiveMessage(controlMessage) }
```
Lukasa, line 242:
```swift
for message in controlMessagesReceived {
    controlMessageReceiver.receiveMessage(message)
}
```

In general `forEach` is not really worth using: it's kinda needless sugar over the more straightforward looping operation.

Lukasa: Also, should this loop move into an `init` on `ControlMessageParser`? That object can really only be used with one thing _anyway_.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,245 @@
...
+            valuePtr.copyMemory(from: data)
+        }
+        return readValue
+    }
+    
+    mutating func receiveMessage(_ controlMessage: UnsafeControlMessage) {
+        if controlMessage.level == IPPROTO_IP && controlMessage.type == ControlMessageParser.ipv4TosType {
+            if let data = controlMessage.data {
+                assert(data.count == 1)
+                precondition(data.count >= 1)
```
glbrntt, line 115: Seems like at least one of these is wrong/unnecessary!

Lukasa: Nope, they're doing different things.

The `assert` codifies something we believe to be true, but that is fundamentally outside our control. The `precondition` covers something that we _require_ to be true for safety. We _believe_ we'll only ever have one byte of data to read from, but it doesn't actually matter if we have more: we're just going to ignore the others. So we use `assert` to guarantee our programming expectations, but at runtime we tolerate machines that do something different so long as they don't lead to us doing something unsafe.

Sources/NIO/ControlMessage.swift
PeterAdams-A: ok

Sources/NIO/ControlMessage.swift
glbrntt: Oh I see now, thanks for the explanation.

Sources/NIO/ControlMessage.swift
```diff
@@ -0,0 +1,251 @@
...
+        return withUnsafePointer(to: &messageHeader) { messageHeaderPtr in
+            let firstCMsg = Posix.cmsgFirstHeader(inside: messageHeaderPtr)
+            return Index(cmsgPointer: firstCMsg)
+        }
+    }
+    
+    var endIndex: Index { return Index(cmsgPointer: nil) }
+    
+    func index(after: Index) -> Index {
+        precondition(after.cmsgPointer != nil)
```
Lukasa, line 75: This precondition is unnecessary: we force-unwrap this pointer before its first use anyway, so the correctness of the code isn't enforced here. If you wanted to match the message below you can move the force-unwrap to this line instead, if that helps.

PR #1591, opened 2020-07-14
Allow Metadata to be constructed and set on AddressedEnvelope

PeterAdams-A:
Motivation:

This needs to be possible so users can specify what
ECN status they wish to send.  The auto-generated
initialiser is internal.

Modifications:

Add public initialiser to AddressedEnvelope.Metadata
Add public initialiser to AddressedEnvelope taking Metadata

Result:

Users will be able to construct AddressedEnvelopes with Metadata set.


Lukasa:

PR #1596, opened 2020-07-15
Explicit Congestion Notification for UDP

PeterAdams-A:
Motivation:

Network congestion can be controlled without packet loss if explicit congestion notification messages are understood.

Modifications:

Receive single and vector paths to get notification and surface.
Send paths for both single and vector sends.
Round trip tests.
Storage for control messages added in a reusable way.

I have tested allocations and they seem to be fine but the allocation tests for vector read and write were not stable enough to sensibly add to CI.

Result:

Explicit Congestion Notifications can be both sent and received.


PeterAdams-A:
@swift-nio-bot test this please.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -66,6 +66,7 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
                                                    msgs: UnsafeMutableBufferPointer<MMsgHdr>,
                                                    addresses: UnsafeMutableBufferPointer<sockaddr_storage>,
                                                    storageRefs: UnsafeMutableBufferPointer<Unmanaged<AnyObject>>,
+                                                   controlMessageStorage: UnsafeControlMessageStorage,
                                                    _ body: (UnsafeMutableBufferPointer<MMsgHdr>) throws -> IOResult<Int>) throws -> IOResult<Int> {
     assert(msgs.count >= Socket.writevLimitIOVectors, "Insufficiently sized buffer for a maximal sendmmsg")
```
Lukasa, line 71: Can we add a new assertion on the size of `controlMessageStorage` that also does this? An extra round of defense would be good.

Sources/NIO/Socket.swift
```diff
@@ -206,13 +206,15 @@ typealias IOVector = iovec
...
+    ///     - controlMessagesReceived: If data is received this is populated using the received control bytes.
     /// - returns: The `IOResult` which indicates how much data could be received and if the operation returned before all the data could be received
     ///     (because the socket is in non-blocking mode)
     /// - throws: An `IOError` if the operation failed.
     func recvmsg(pointer: UnsafeMutableRawBufferPointer,
                  storage: inout sockaddr_storage,
                  storageLen: inout socklen_t,
-                 controlBytes: inout Slice<UnsafeMutableRawBufferPointer>) throws -> IOResult<Int> {
+                 controlBytes: inout Slice<UnsafeMutableRawBufferPointer>,
+                 controlMessagesReceived: inout UnsafeControlMessageCollection?) throws -> IOResult<Int> {
```
Lukasa, line 217: So this interface isn't really doing its job anymore: the `controlBytes` were `inout` because the plan had been to return the `controlBytes` and let the parsing happen outside it. That's clearly not happening anymore with this refreshed API, so instead we should consider at the _very least_ dropping `Slice` and `inout` because we don't need that slicing mess.

More broadly it'd be nicer to be able to pass a struct in here that more clearly communicates the behaviour we want (namely that the control message collection is tied to the lifetime of this pointer.

Tests/NIOTests/ControlMessageTests.swift
```diff
@@ -90,4 +90,29 @@ class ControlMessageTests: XCTestCase {
         }
         XCTAssertEqual(expected, decoded)
     }
+
+    private func assertBuffersNonOverlapping(_ b1: UnsafeMutableRawBufferPointer,
+                                             _ b2: UnsafeMutableRawBufferPointer,
+                                             file: StaticString = #file,
+                                             line: UInt = #line) {
```
Lukasa, line 97: No point passing these around if we aren't going to use them.

Sources/NIO/ControlMessage.swift
```diff
@@ -18,6 +18,49 @@ import CNIODarwin
...
+/// Memory for use as `cmsghdr` and associated data.
+/// Supports multiple messages each with enough storage for multiple `cmsghdr`
+struct UnsafeControlMessageStorage: Collection {
+    let bytesPerMessage: Int
+    var buffer: UnsafeMutableRawBufferPointer
+
+    /// Initialise which includes allocating memory
+    /// parameter:
+    /// - msghdrCount: How many messages we need space for.
+    private init(msghdrCount: Int) {
```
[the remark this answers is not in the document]
PeterAdams-A: OK.

Sources/NIO/ControlMessage.swift
```diff
@@ -18,6 +18,49 @@ import CNIODarwin
...
+
+    /// Allocate new memory - Caller must call `deallocate` when no longer required..
+    /// parameter:
+    ///   - msghdrCount:   How many `msghdr` structures will be fed from this buffer - we assume 4 Int32 cmsgs for each.
+    static func allocate(msghdrCount: Int) -> UnsafeControlMessageStorage {
+        return UnsafeControlMessageStorage(msghdrCount: msghdrCount)
+    }
+
+    func deallocate() {
+        self.buffer.deallocate()
```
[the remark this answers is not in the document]
PeterAdams-A: OK.

Sources/NIO/PendingDatagramWritesManager.swift
PeterAdams-A: Good idea.

Tests/NIOTests/ControlMessageTests.swift
PeterAdams-A: Ahh yes, will start using them.

Sources/NIO/Socket.swift
PeterAdams-A: Still feels pretty awkward to be honest.

Sources/NIO/SocketChannel.swift
```diff
@@ -510,7 +524,17 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+                if self.reportExplicitCongestionNotifications,
+                   let controlMessagesReceived = controlBytes.receivedControlMessages {
+                    metadata = .init(from: controlMessagesReceived)
+                } else {
+                    metadata = nil
+                }
+
+                let msg = AddressedEnvelope(remoteAddress: rawAddress.convert(),
+                                            data: buffer,
+                                            metadata:  metadata)
```
Lukasa, line 537: Extreme nitpick

```suggestion
                                            metadata: metadata)
```

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -662,4 +666,85 @@ final class DatagramChannelTests: XCTestCase {
...
+            }
+
+            let expectedReads = ecnStates.count * (vectorSend ? 2 : 1)
+            let reads = try receiveChannel.waitForDatagrams(count: expectedReads)
+            XCTAssertEqual(reads.count, expectedReads)
+            for readNumber in 0..<reads.count {
+                let read = reads[readNumber]
+                XCTAssertEqual(read.metadata?.ecnState, ecnStates[readNumber / (vectorSend ? 2 : 1)])
+            }
+        } ())
```
Lukasa, line 715: This test leaks sockets: can we make sure we're adding `defer` blocks to clean up channels?

PeterAdams-A: Done.

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -662,4 +666,91 @@ final class DatagramChannelTests: XCTestCase {
...
+                .bind(host: address, port: 0)
+                .wait()
+            defer {
+                XCTAssertNoThrow(try receiveChannel.close().wait())
+            }
+            let sendChannel = try DatagramBootstrap(group: group)
+                .bind(host: address, port: 0)
+                .wait()
+            defer {
+                XCTAssertNoThrow(try! sendChannel.close().wait())
```
Lukasa, line 694: No point having an `XCTAssertNoThrow` if we crash ;)

```suggestion
                XCTAssertNoThrow(try sendChannel.close().wait())
```

PR #1597, opened 2020-07-21
Don't install Jazzy on Xenial

PeterAdams-A:
Motivation:

Ruby on Xenial is too old for the cocoapods downloader which jazzy uses.

Modifications:

Change Dockerfile to not install Jazzy on xenial.

Result:

Dockerfile will build again, but bionic or later is required for docs.


PeterAdams-A:

PR #1606, opened 2020-07-31
Add doc comment about potential of events to arrive out of order.

PeterAdams-A:
Motivation:

This has come as a surprise to at least one person and is easy to forget.

Modifications:

Doc commented added at the top of ChannelPipeline.

Result:

Docs are clearer.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -109,6 +109,11 @@
 /// - 1 and 2 don't implement `ChannelOutboundHandler`, and therefore the actual evaluation order of a outbound event will be: 5, 4, and 3.
 /// - If 5 implements both `ChannelInboundHandler` and `ChannelOutboundHandler`, the evaluation order of an inbound and a outbound event could be 125 and 543 respectively.
 ///
+/// Note:   Handlers may choose not to progagate messages down the pipeline immediaitely.  For example a handler may need to wait
+/// for additional data before sending a protocol event to the next handler in the pipeline.  Due to this you can't assume that later handlers
+/// in the pipeline will recienve the same number of events as were sent, or that events of different types will arrive in the same order.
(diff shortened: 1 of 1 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -109,6 +109,11 @@
 /// - 1 and 2 don't implement `ChannelOutboundHandler`, and therefore the actual evaluation order of a outbound event will be: 5, 4, and 3.
 /// - If 5 implements both `ChannelInboundHandler` and `ChannelOutboundHandler`, the evaluation order of an inbound and a outbound event could be 125 and 543 respectively.
 ///
+/// Note:   Handlers may choose not to progagate messages down the pipeline immediaitely.  For example a handler may need to wait
```
Lukasa, line 112:
```suggestion
/// Note:   Handlers may choose not to progagate messages down the pipeline immediately.  For example a handler may need to wait
```

yim-lee: In addition to Cory's suggestion:

```suggestion
/// Note:   Handlers may choose not to propagate messages down the pipeline immediaitely.  For example a handler may need to wait
```

ktoso: This reads good, thanks @PeterAdams-A !

The spacing is a bit weird... Why double spaces everywhere?

```suggestion
Note: Handlers may choose not to progagate messages down the pipeline immediately. For example a handler may need to wait
```

etc?

ktoso:
Resolves https://github.com/apple/swift-nio/issues/1607

PeterAdams-A:
Interestingly commit comments seem to disappear when multiple people suggest changes to the same line.  I've corrected both typos though.

The multiple space thing comes from being old - it's what we used to do.  I was going to accept the changes but then I noticed that the comments above this area formatted the same way so I'll leave for now so at least the file is consistent with itself.

PR #1609, opened 2020-08-03
Define public EventLoopGroupProvider type

### LINKED ISSUE #1608: Consider creating public EventLoopGroupProvider type
`EventLoopGroupProvider` type is used in a number of swift-server libraries, example from [AsyncHTTPClient](https://github.com/swift-server/async-http-client/blob/c9a9bf061d713c91ee9974fa8a6afe413acfd0e9/Sources/AsyncHTTPClient/HTTPClient.swift#L737-L743):

```swift
    /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
    public enum EventLoopGroupProvider {
        /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
        case shared(EventLoopGroup)
        /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
        case createNew
    }
```

Currently different libraries (have to) implement (copy and paste) the same type in different namespaces.

This is a bit of a hassle as it requires mapping of "the same" type in different namespaces, example: https://github.com/pokryfka/aws-xray-sdk-swift/pull/49#discussion_r464148735

I think it would be very convenient if `SwiftNIO` defined it  and in such way promoted the pattern.

CC @ktoso

pokryfka:
Closes #1608

Defines public `EventLoopGroupProvider` type.

### Motivation:

`EventLoopGroupProvider` type is used in a number of swift-server libraries, example from [AsyncHTTPClient](https://github.com/swift-server/async-http-client/blob/c9a9bf061d713c91ee9974fa8a6afe413acfd0e9/Sources/AsyncHTTPClient/HTTPClient.swift#L737-L743):

```swift
    /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
    public enum EventLoopGroupProvider {
        /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
        case shared(EventLoopGroup)
        /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
        case createNew
    }
```

Currently different libraries (have to) implement (copy and paste) the same type in different namespaces.

This is a bit of a hassle as it requires mapping of "the same" type in different namespaces, example: https://github.com/pokryfka/aws-xray-sdk-swift/pull/49#discussion_r464148735

I think it would be very convenient if `SwiftNIO` defined it  and in such way promoted the pattern.

### Modifications:

Created public `EventLoopGroupProvider` type.

### Result:

Nothing in `SwiftNIO` itself ;-)

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -771,6 +771,10 @@ extension EventLoopGroup {
 }
 
 /// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
+///
+/// This type is intended to be used by libraries which use NIO, and offer their users either the option
+/// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be 
+/// managed by given library and its lifecycle.
 public enum EventLoopGroupProvider {
     /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
     case shared(EventLoopGroup)
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -770,16 +770,16 @@ extension EventLoopGroup {
     }
 }
 
-/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
-///
 /// This type is intended to be used by libraries which use NIO, and offer their users either the option
-/// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be 
+/// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be
 /// managed by given library and its lifecycle.
 public enum EventLoopGroupProvider {
-    /// Use an `EventLoopGroup` provided by the user. 
+    /// Use an `EventLoopGroup` provided by the user.
     /// The owner of this group is responsible for its lifecycle.
     case shared(EventLoopGroup)
-    /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
+    /// Create a new `EventLoopGroup` when necessary.
+    /// The library which accepts this provider takes ownership of the created event loop group,
+    /// and must ensure its proper shutdown when the library is being shut down.
     case createNew
 }
 
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -770,6 +770,14 @@ extension EventLoopGroup {
     }
 }
 
+/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
+public enum EventLoopGroupProvider {
+    /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
+    case shared(EventLoopGroup)
+    /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
+    case createNew
+}
+
 private let nextEventLoopGroupID = NIOAtomic.makeAtomic(value: 0)
 
 /// Called per `NIOThread` that is created for an EventLoop to do custom initialization of the `NIOThread` before the actual `EventLoop` is run on it.
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -773,7 +773,7 @@ extension EventLoopGroup {
 /// This type is intended to be used by libraries which use NIO, and offer their users either the option
 /// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be
 /// managed by given library and its lifecycle.
-public enum EventLoopGroupProvider {
+public enum NIOEventLoopGroupProvider {
     /// Use an `EventLoopGroup` provided by the user.
     /// The owner of this group is responsible for its lifecycle.
     case shared(EventLoopGroup)
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -776,7 +776,8 @@ extension EventLoopGroup {
 /// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be 
 /// managed by given library and its lifecycle.
 public enum EventLoopGroupProvider {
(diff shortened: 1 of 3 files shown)


Tests/NIOTests/EventLoopTest.swift
```diff
@@ -1279,4 +1279,25 @@ public final class EventLoopTest : XCTestCase {
...
+        }
+
+        let provider = EventLoopGroupProvider.shared(eventLoopGroup)
+
+        if case EventLoopGroupProvider.shared(let sharedEventLoopGroup) = provider {
+            XCTAssertTrue(sharedEventLoopGroup is MultiThreadedEventLoopGroup)
+            // this is sketchy but I dont think we can do better without changing MultiThreadedEventLoopGroup
+            // note that the description contains `myGroupID`
+            XCTAssertEqual(sharedEventLoopGroup.description, eventLoopGroup.description)
+            XCTAssertNotEqual(sharedEventLoopGroup.description, eventLoopGroup2.description)
```
pokryfka, line 1298: 🤔

pokryfka:
CC @Lukasa 

Please see inline comment above and let me know of any issues.

Tried to follow code style, would be great if it was defined in `swift-format` or `swiftformat` file.

Tests/NIOTests/EventLoopTest.swift
Lukasa: We can do better:

```swift
if case .shared(let sharedEventLoopGroup) = provider, let sharedGroup = sharedEventLoopGroup as? MultiThreadedEventLoopGroup {
    XCTAssert(sharedGroup === eventLoopGroup)
}
```

pokryfka: tried, not `Equatable`

Lukasa: Note the number of `=` signs I used.

pokryfka: oh, it works; I did not know about `===`

will submit a fix

Sources/NIO/EventLoop.swift
```diff
@@ -770,6 +770,14 @@ extension EventLoopGroup {
     }
 }
 
+/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
+public enum EventLoopGroupProvider {
+    /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
+    case shared(EventLoopGroup)
+    /// `EventLoopGroup` will be created by the client. When `syncShutdown` is called, created `EventLoopGroup` will be shut down as well.
```
glbrntt, line 777: This comment needs updating to reflect the fact that this won't necessarily be used by a client nor will it  necessarily have a `syncShutdown`.

glbrntt:
@pokryfka you'll need to run `./scripts/sanity.sh` as well to generate the linux test manifest

Sources/NIO/EventLoop.swift
```diff
@@ -770,6 +770,14 @@ extension EventLoopGroup {
     }
 }
 
+/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
+public enum EventLoopGroupProvider {
```
glbrntt, line 774: This'll need to be renamed to `NIOEventLoopGroupProvider` as well (see: https://github.com/apple/swift-nio/blob/master/docs/public-api.md#1-no-global-namespace-additions-that-arent-prefixed)

Sources/NIO/EventLoop.swift
```diff
@@ -770,6 +770,14 @@ extension EventLoopGroup {
     }
 }
 
+/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
```
ktoso, line 773: Some docs on the purpose of this type could be good, e.g.

```suggestion
/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
///
/// This type is intended to be used by libraries which use NIO, and offer their users either the option
/// to `.share` an existing event loop group or create (and manage) a new one (`.createNew`) and let it be 
/// managed by given library and its lifecycle.
```

Sources/NIO/EventLoop.swift
```diff
@@ -770,6 +770,14 @@ extension EventLoopGroup {
...
+public enum EventLoopGroupProvider {
+    /// `EventLoopGroup` will be provided by the user. Owner of this group is responsible for its lifecycle.
```
ktoso, line 775: Phrasing nitpick:

```suggestion
    /// Use an `EventLoopGroup` provided by the user. 
    /// The owner of this group is responsible for its lifecycle.
```

Sources/NIO/EventLoop.swift
ktoso: Perhaps rephrased as:

```suggestion
    /// Create a new `EventLoopGroup` when necessary. 
    /// The library which accepts this provider takes ownership of the created event loop group, 
    /// and must ensure its proper shutdown when the library is being shut down.
```

?

pokryfka:
@glbrntt

>  you'll need to run ./scripts/sanity.sh as well to generate the linux test manifest

thank you for pointing it out

I run tests with `swift test --enable-test-discovery` myself (I think it works only for Swift 5+)

updated linux test manifest for `EventLoopTest+XCTest.swift`

glbrntt:

glbrntt:
> I run tests with `swift test --enable-test-discovery` myself (I think it works only for Swift 5+)

Unfortunately it was introduced in 5.1 and we support 5.0 so still need to rely on a script for this.

pokryfka:
Let me confirm the changes (including comments):

1. rename `EventLoopGroupProvider` to `NIOEventLoopGroupProvider`

2. update comments as suggested above

```swift
/// Specifies how `EventLoopGroup` will be created and establishes lifecycle ownership.
public enum NIOEventLoopGroupProvider {
    /// Use an `EventLoopGroup` provided by the user.
    /// The owner of this group is responsible for its lifecycle.
    case shared(EventLoopGroup)
    /// Create a new `EventLoopGroup` when necessary.
    /// The library which accepts this provider takes ownership of the created event loop group,
    /// and must ensure its proper shutdown when the library is being shut down.
    case createNew
}
```

Lukasa:

PR #1610, opened 2020-08-03
Deprecate and replace `ByteBuffer.clear(minimumCapacity: _Capacity)`

### LINKED ISSUE #1589: `ByteBuffer.clear(minimumCapacity:)` uses `_Capacity` as it's type for `minimumCapacity`
### Expected behavior

NIO shouldn't expose `_UnderscoreInternal` types in its public API.

### Actual behavior

`ByteBuffer.clear(minimumCapacity:)` uses `_Capacity` instead of `Int`, see:

https://github.com/apple/swift-nio/blob/ee8acd9b3d0e2f00b852743f86075c35bea9e986/Sources/NIO/ByteBuffer-core.swift#L711

Davidde94:
Resolves #1589 

Deprecate `ByteBuffer.clear(minimumCapacity: _Capacity)` and introduce `Bytebuffer.clear(minimumCapacity: Int)`.

Note - all of the tests for `clear` will automatically use the new `Int` method. I considered making new tests to explicitly cover the existing `_Capacity` method, but thought that this is probably not necessary. Open to being told otherwise.

### Motivation:

`_Capacity` is underscored and so should be used publicly.

### Modifications:

Add a deprecation warning to the `_Capacity` method, and introduce a new `Int`-based method.

### Result:

Deprecation warning if the `_Capacity` version of `clear` was used.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -709,16 +709,8 @@ public struct ByteBuffer {
     /// - parameters:
     ///     - minimumCapacity: The minimum capacity that will be (re)allocated for this buffer
     @available(*, deprecated, message: "Use an `Int` as the argument")
-    public mutating func clear(minimumCapacity: _Capacity) {
-        if !isKnownUniquelyReferenced(&self._storage) {
-            self._storage = self._storage.allocateStorage(capacity: minimumCapacity)
-        } else if minimumCapacity > self._storage.capacity {
-            self._storage.reallocStorage(capacity: minimumCapacity)
-        }
-        self._slice = self._storage.fullSlice
-
-        self._moveWriterIndex(to: 0)
-        self._moveReaderIndex(to: 0)
+    public mutating func clear(minimumCapacity: UInt32) {
+        self.clear(minimumCapacity: Int(minimumCapacity))
     }
     
     /// Set both reader index and writer index to `0`. This will reset the state of this `ByteBuffer` to the state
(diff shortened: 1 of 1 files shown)


Sources/NIO/ByteBuffer-core.swift
```diff
@@ -708,6 +708,7 @@ public struct ByteBuffer {
     ///
     /// - parameters:
     ///     - minimumCapacity: The minimum capacity that will be (re)allocated for this buffer
+    @available(*, deprecated, message: "Use an `Int` as the argument")
     public mutating func clear(minimumCapacity: _Capacity) {
```
glbrntt, line 712: Can we switch `_Capacity` to whatever its type alias is currently is?

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -708,6 +708,7 @@ public struct ByteBuffer {
...
         if !isKnownUniquelyReferenced(&self._storage) {
```
glbrntt, line 713: Can we just forward this implementation to the new one?

Davidde94: I wasn't sure how - in this case `renamed` doesn't work as the functions have the same name. Any suggestions?

Sources/NIO/ByteBuffer-core.swift
Davidde94: I can do - wasn't sure if this is technically an API change, @Lukasa any opinions?

Sources/NIO/ByteBuffer-core.swift
glbrntt: Oh I meant just call `self.clear(minimumCapacity: Int(minimumCapacity))`

Davidde94: Ah! Sure

Sources/NIO/ByteBuffer-core.swift
Lukasa: It's not, typealiases are exactly the same as their underlying type. It only matters if you remove the name altogether, which we aren't.

PR #1614, opened 2020-08-07
Reduce refcounting in Heap by removing configurability

Lukasa:
Motivation:

Right now Heap.heapify incurs quite a lot of ARC traffic when it is used
in SelectableEventLoop. Specifically, each of the invocations of
self.comparator in heapify incurs three refcount operations. These three
are all the result of the fact that the comparator is an arbitrary
function, but all three are guaranteed to be unnecessary.

The first refcount is for the closure context for self.comparator. Of
course, self.comparator has only two possible values, neither of which
needs a closure context, but the compiler doesn't know that and the type
system doesn't encode it.

The other two refcounts are for the ScheduledTask objects themselves.
Again, the compiler doesn't know what the comparator function will do,
so it must defensively ensure that a retain/release is inserted for the
arguments of that function.

If we happen to hit both branches of the code, we'll execute 6 retains
and 6 releases, and even in fairly average cases we'll execute 3 of
each.

All of this is unnecessary: the only runtime function ever used here is
a trivial comparison operation. We know what it is at compile time, we
just have to ask. While it was admirable to have the configurability of
the priority queue in place, after 3 years I think we have a pretty good
feeling that we're not going to need the ability to have a max-heap
instead of a min-heap.

We can always get this back if we need it.

Modifications:

- Removed the configurability of Heap.
- Removed HeapType.
- Removed arguments from PriorityQueue
- Removed all tests testing now-removed max-heap functionality.

Result:

Less ARC traffic in the core task scheduling loop.

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2020 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -61,7 +61,7 @@ public final class EmbeddedEventLoop: EventLoop {
     /// The current "time" for this event loop. This is an amount in nanoseconds.
     /* private but tests */ internal var _now: NIODeadline = .uptimeNanoseconds(0)
 
-    private var scheduledTasks = PriorityQueue<EmbeddedScheduledTask>(ascending: true)
+    private var scheduledTasks = PriorityQueue<EmbeddedScheduledTask>()
 
     // The number of the next task to be created. We track the order so that when we execute tasks
     // scheduled at the same time, we may do so in the order in which they were submitted for
--- a/Sources/NIO/Heap.swift
+++ b/Sources/NIO/Heap.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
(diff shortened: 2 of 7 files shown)


Lukasa:
If you hit performance issues, a better choice may be to use wrapper structs that define `Comparable` The way you want. That’ll give the compiler visibility into the implementation (as it’s always concrete for the given type) and would also have worked to elide the ARC.

ktoso:
Good idea, thanks :)

tdotclare:
Thanks for looking into this @Lukasa! Improved the tests I run to track performance on LK yesterday in order to glean a little more insight into things -

Tests are flooding the SEL and for the vast majority of tasks, are hitting a synch cache - a succeeded cached ast is serialized to a succeeded `ELF<ByteBuffer>.whenComplete()` callback to decrement the task counter, so the (pipe-serialize) figure per iteration is nearly all NIO callback work (momentary hit on a POSIX lock when success), and numbers are purely from actual *in-loop* work and don't include any setup or shutdown activity on the SEL - so this (roughly) seems to cut out a nice 15% in execution time when it's being flooded!

For ref, the numbers below are on a single SEL, late 2014 4GHz i7 iMac.

* This branch - ~ 11.7µs in non-synch pipe
```
Linear Cache Speed: 10x100,000 render
0.05% test overhead
1.480 s avg/test: 14.798 µs/3.191 µs pipe:serialize/iteration
Random Cache Speed: 10x100,000 render
0.03% test overhead
2.958 s avg/test: 29.584 µs/17.780 µs pipe:serialize/iteration
```
* 2.20.2 - ~ 13.8µs in non-synch pipe
```
Linear Cache Speed: 10x100,000 render
0.04% test overhead
1.651 s avg/test: 16.514 µs/3.126 µs pipe:serialize/iteration
Random Cache Speed: 10x100,000 render
0.03% test overhead
3.279 s avg/test: 32.787 µs/18.531 µs pipe:serialize/iteration
```

PR #1616, opened 2020-08-10
Add `EventLoop.makeResultantFuture(Result<T, Error>...)`

tdotclare:
Adds a method to`EventLoop` to allow creating a succeeded or failed `EventLoopFuture` directly from a `Result`.

```swift
public func makeResultantFuture<T>(_ value: Result<T, Error>, ...)
```

### Motivation:
In API functions that return an `EventLoopFuture`, underlying synchronous functions may themselves be already able to provide useful `Result`s that can immediately be passed to a completed promise, and it would be easier to pass that directly onwards before moving to potentially blocking branches; example:

Before:
```swift
if let result = blockingCache.insert(ast, replace: true) {
  switch result {
    case .success(let ast): return eventLoop.makeSucceededFuture(ast)
    case .failure(let err): return eventLoop.makeFailedFuture(err)
  }
}
// .... move on to handle async operations
```
After:
```swift
if let result = blockingCache.insert(ast, replace: true) {
  return eventLoop.makeResultantFuture(result) 
}
//  ... move on to handle async operations
```
### Modifications:

Add above method

### Result:

Less boilerplate switches needed in mixed sync/async code when returning ELFs

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -566,11 +566,24 @@ extension EventLoop {
     /// Creates and returns a new `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop` as execution `NIOThread`.
     ///
     /// - parameters:
-    ///     - result: the value that is used by the `EventLoopFuture`.
+    ///     - value: the value that is used by the `EventLoopFuture`.
     /// - returns: a succeeded `EventLoopFuture`.
     @inlinable
     public func makeSucceededFuture<Success>(_ value: Success, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
         return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
+    }
+    
+    /// Creates and returns a new `EventLoopFuture` with state arbitrated from a provided `Result`. Notifications will be done using this `EventLoop` as execution `NIOThread`.
+    ///
+    /// - parameters:
+    ///     - value: the `Result` whose state will arbitrate the resultant `EventLoopFuture`.
+    /// - returns: a succeeded or failed `EventLoopFuture`.
+    @inlinable
+    public func makeResultantFuture<T>(_ value: Result<T, Error>, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<T> {
+        switch value {
+            case .success(let success): return makeSucceededFuture(success, file: file, line: line)
+            case .failure(let failure): return makeFailedFuture(failure, file: file, line: line)
+        }
     }
 
     /// An `EventLoop` forms a singular `EventLoopGroup`, returning itself as the 'next' `EventLoop`.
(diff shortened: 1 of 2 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -566,12 +566,22 @@ extension EventLoop {
...
         return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
     }
+    
+    /// Creates and returns a new `EventLoopFuture` with state arbitrated from a provided `Result`. Notifications will be done using this `EventLoop` as execution `NIOThread`.
+    ///
+    /// - parameters:
+    ///     - value: the `Result` whose state will arbitrate the resultant `EventLoopFuture`.
+    /// - returns: a succeeded or failed `EventLoopFuture`.
+    @inlinable
+    public func makeResultantFuture<T>(_ value: Result<T, Error>, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<T> {
```
Lukasa, line 582: I think I’d prefer `makeCompletedFuture`.

Lukasa, line 582: Also, can we give a more descriptive name to the generic type parameter? `Success` is better than `T`.

Tests/NIOTests/EventLoopFutureTest.swift
```diff
@@ -1023,6 +1023,28 @@ class EventLoopFutureTest : XCTestCase {
...
+        let failure: Result<Int, Error> = .failure(EventLoopFutureTestError.example)
+
+        let future = (success: loop.makeResultantFuture(success),
+                      failure: loop.makeResultantFuture(failure))
+        
+        let promised = (success: loop.makePromise(of: Int.self),
+                        failure: loop.makePromise(of: Int.self))
+        
+        promised.success.completeWith(future.success)
+        promised.failure.completeWith(future.failure)
```
Lukasa, line 1041: This is quite convoluted: should we just `.wait` for the result and use the regular `XCTAssert` macros on the returned value?

Lukasa:

Sources/NIO/EventLoop.swift
tdotclare: Think I was avoiding that mostly to keep things from implying or confusing purpose with makeSucceeded

tdotclare: I toyed with that actually - just slightly concerned “completed” sounds slightly closer to “succeeded” than being agnostic about what’s in the Result?

glbrntt: I'm +1 on `makeCompletedFuture`; there's prior art in NIO around futures/promise accepting a `Result` type: `EventLoopFuture<Value>.whenComplete` and `EventLoopPromise<Value>.completeWith`, for example.

Lukasa: I don't think it'll matter: the generic type parameter is definitely the type of the success case.

glbrntt:
Closing as this got implemented as part of https://github.com/apple/swift-nio/pull/1766

PR #1625, opened 2020-08-28
Add @inlinable to all BBV methods.

Lukasa:
Motivation:

While looking at a different part of the code in swift-nio-http2 I
noticed that we can't inline many straightforward ByteBufferView
operations, including bytewise-access. That's pretty sad, so I thought
we should fix it.

Modifications:

- Slap @inlinable on all the BBV entry points.

Result:

Better code generation, particularly around repeated subscript accesses.


glbrntt (approved):
Nice find!

PR #1627, opened 2020-09-02
Remove incorrect assertion in MarkedCircularBuffer.popFirst

glbrntt:
Motivation:

`MarkedCircularBuffer.popFirst()` asserts that the backing buffer should
contain more than zero elements yet `popFirst()` allows `nil` to be
returned when there is no value to return.

Modifications:

- Move the assertion from `popFirst() -> Element?` to `removeFirst() -> Element`
- Test for `popFirst()`

Result:

- We can safely call `popFirst()` on an empty `MarkedCircularBuffer` in
  debug mode.


Davidde94 (approved):
LGTM

Issue #1485, opened 2020-04-20
ByteBuffer should give a user an answer to the question how large the underlying allocation is


weissi:
You'd think that `buffer.capacity` gives you the answer to the following question:

> How large was the underlying allocation that is backing this `ByteBuffer`?

However, that is not true. If you're a slice of a `ByteBuffer`, you simply can't get an answer to this question...

Consider:

```
var buffer = ByteBufferAllocator().buffer(capacity: 1024 * 1024 * 1024) // allocate 1GB
buffer.writeString("X")
var otherBuffer = buffer.readSlice(length: 1)!
```

`otherBuffer.capacity` is now `1`, despite the fact that it may be the last thing holding onto 1 GB of memory.

We can either repurpose `.capacity` or create a new property that gives you `self._storage.fullSlice.count` which is the size of the allocation.

gcjenkinson:
I've created a pull request for this issue as an introductory task for myself (being unfamiliar with both Swift and SwiftNIO). It should be vaguely sane and at the very least  will allow be to iron out any workflow/process misunderstanding before tackling something more challenging.

FranzBusch:
Closing this since it has been fixed in: https://github.com/apple/swift-nio/pull/1629

PR #1629, opened 2020-09-03
Distinguish capacity and no. of addressable bytes. (#1485)

gcjenkinson:
## Motivation:

Whilst the semantics of the capacity property are well defined, a
distinction between the number of addressable bytes in the buffer and
the total number of bytes allocated to the buffer would be clearer (note
that these values are not necessarily the same for example, when
constructing of COW slice from the original buffer:
buf.readSlice(length: 1)).

## Modifications:

- Repurposed the capacity property to always report the capacity of the
allocation underlying the storage.
- Added new length property to show the number of addressable bytes of
the buffer (or derived COW slice).

## Result:

The capacity of the original buffer and a COW slice with remain the same
as the original buffer until data is written to the slice. The length
property reports the number of addressable bytes e.g. the length of the
buffer obtained by buf.readSlice(length: 1) is 1.

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -441,6 +441,13 @@ public struct ByteBuffer {
         return self._slice.count
     }
 
+    /// The current capacity of the underlying storage of this `ByteBuffer`.
+    /// A COW slice of the buffer (e.g. readSlice(length: x)) will posses the same storageCapacity as the original
+    /// buffer until new data is written.
+    public var storageCapacity: Int {
+        return self._storage.fullSlice.count
+    }
+
     /// Reserves enough space to store the specified number of bytes.
     ///
     /// This method will ensure that the buffer has space for at least as many bytes as requested.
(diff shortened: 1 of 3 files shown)


Lukasa (commented):
Thanks for the patch!

In this case, given that `capacity` is used elsewhere in the implementation and has a well-defined meaning, I think I'd like to propose that we invert the patch: instead of adding a new property that does what `capacity` _used_ to do, we should add a new property that indicates how much storage is being retained by the buffer.

Lukasa:

Lukasa:
I think a new property with a well distinguished name (I like `allocatedCapacity`) is good. The test failures in the build logs demonstrate that there are parts of NIO that do rely on this information, and it also happens to be used when resizing buffers in various cases (e.g. it interacts with `reserveCapacity`)

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -441,6 +441,13 @@ public struct ByteBuffer {
         return self._slice.count
     }
 
+    /// The current capacity of the underlying storage of this `ByteBuffer`.
+    /// A COW slice of the buffer (e.g. readSlice(length: x)) will posses the same storageCapacity as the original
+    /// buffer until new data is written.
+    public var storageCapacity: _Capacity {
```
Lukasa, line 447: `_Capacity` is a strictly internal type: let's use `Int`.

Tests/NIOTests/NIOAnyDebugTest.swift
```diff
@@ -22,7 +22,8 @@ class NIOAnyDebugTest: XCTestCase {
         XCTAssertEqual(wrappedInNIOAnyBlock(123), wrappedInNIOAnyBlock("123"))
         
         let bb = ByteBuffer(string: "byte buffer string")
-        XCTAssertTrue(wrappedInNIOAnyBlock(bb).contains("NIOAny { ByteBuffer { readerIndex: 0, writerIndex: 18, readableBytes: 18, capacity: 32, slice: _ByteBufferSlice { 0..<32 }, storage: "))
+        print("\(bb)")
```
Lukasa, line 25: Presumably we want this to be removed.

Lukasa (approved):
Looks great, thanks @gcjenkinson!

PR #1632, opened 2020-09-20
build: bump tools version to 5.3

compnerd:
This is needed to gain access to Windows support.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Package.swift@5.3
+++ b/Package.swift@5.3
@@ -1,0 +1,99 @@
+// swift-tools-version:5.3
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


0xTim:
Can't this be achieved with a `Package@swift-5.3.swift` manifest?

Lukasa (changes requested):
Yeah, we cannot unconditionally bump the tools version, as it amounts to dropping support for 5.0 through 5.2. A new `Package.swift@5.3` is a workable alternative.

PeterAdams-A:
I don't think we can really take this approach as it breaks previous versions of swift.

compnerd:
Okay, I'll fork the current `package.swift` then.  Does s-p-m pickup the file automatically?  Seems that the same argument as with CMake applies to keeping the old definition: you have two definitions.

Lukasa:
Yes, SPM will pick up the appropriate version.

Lukasa:
I think we agreed we don't need this?

compnerd:
I think for now we do not.  We can always revisit this if necessary.

PR #1637, opened 2020-09-21
Cleanup unix socket pathname on server socket close or bind

AndriusA:
Resolves https://github.com/apple/swift-nio/issues/1619

### Motivation:

Server bootstrap should clean up its socket file on clean close, and additionally, we should by default try and remove an existing socket file before binding to it.

### Modifications:

- Adds `cleanupExistingSocketFile` to server bootstrap `bind` that signals whether preexisting socket file should be cleaned up before binding
- Cleans up existing socket file on server socket close

To avoid accidental loss of data, the path is only cleaned up if it points to a socket file.

### Result:

Server socket successfully binds to a UNIX Domain Socket Path if the path already exists and is a socket file. On clean close, the socket file also gets removed.

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -18,10 +18,8 @@ import NIOConcurrencyHelpers
 import let WinSDK.INVALID_SOCKET
 #endif
 
-public enum SocketSetupError: Error {
-    /// The requested UDS path exists and has wrong type (not a socket).
-    case unixDomainSocketPathWrongType
-}
+/// The requested UDS path exists and has wrong type (not a socket).
+public struct UnixDomainSocketPathWrongType: Error {}
 
 /// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
 protocol Registration {
@@ -313,7 +311,7 @@ class BaseSocket: BaseSocketProtocol {
     ///
     /// - parameters:
     ///     - unixDomainSocketPath: The pathname of the UDS.
-    /// - throws: An `SocketSetupError.unixDomainSocketPathWrongType` if the pathname exists and is not a socket.
+    /// - throws: An `UnixDomainSocketPathWrongType` if the pathname exists and is not a socket.
     static func cleanupSocket(unixDomainSocketPath: String) throws {
         do {
             var sb: stat = stat()
@@ -325,12 +323,12 @@ class BaseSocket: BaseSocketProtocol {
             if sb.st_mode & S_IFSOCK == S_IFSOCK {
                 try Posix.unlink(pathname: unixDomainSocketPath)
             } else {
-                throw SocketSetupError.unixDomainSocketPathWrongType
+                throw UnixDomainSocketPathWrongType()
             }
         } catch let err as IOError {
             // If the filepath did not exist, we consider it cleaned up
             if err.errnoCode == ENOENT {
-                return ();
+                return
             }
             throw err
         }
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -18,6 +18,11 @@ import NIOConcurrencyHelpers
 import let WinSDK.INVALID_SOCKET
 #endif
 
+public enum SocketSetupError: Error {
+    /// The requested UDS path exists and has wrong type (not a socket).
+    case unixDomainSocketPathWrongType
+}
+
 /// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
 protocol Registration {
     /// The `SelectorEventSet` in which the `Registration` is interested.
@@ -308,7 +313,7 @@ class BaseSocket: BaseSocketProtocol {
     ///
     /// - parameters:
     ///     - unixDomainSocketPath: The pathname of the UDS.
-    /// - throws: An `SocketAddressError.unixDomainSocketPathWrongType` if the pathname exists and is not a socket.
+    /// - throws: An `SocketSetupError.unixDomainSocketPathWrongType` if the pathname exists and is not a socket.
     static func cleanupSocket(unixDomainSocketPath: String) throws {
         do {
             var sb: stat = stat()
@@ -320,7 +325,7 @@ class BaseSocket: BaseSocketProtocol {
             if sb.st_mode & S_IFSOCK == S_IFSOCK {
                 try Posix.unlink(pathname: unixDomainSocketPath)
             } else {
-                throw SocketAddressError.unixDomainSocketPathWrongType
+                throw SocketSetupError.unixDomainSocketPathWrongType
             }
         } catch let err as IOError {
             // If the filepath did not exist, we consider it cleaned up
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -304,24 +304,30 @@ class BaseSocket: BaseSocketProtocol {
     
     /// Cleanup the unix domain socket.
     ///
-    /// Deletes the associated file if it exists and has socket type.
+    /// Deletes the associated file if it exists and has socket type. Does nothing if pathname does not exist.
     ///
     /// - parameters:
     ///     - unixDomainSocketPath: The pathname of the UDS.
-    /// - throws: An `IOError` if the operation failed.
+    /// - throws: An `SocketAddressError.unixDomainSocketPathWrongType` if the pathname exists and is not a socket.
     static func cleanupSocket(unixDomainSocketPath: String) throws {
-        let sb: UnsafeMutablePointer<stat> = UnsafeMutablePointer<stat>.allocate(capacity: 1)
-        if lstat(unixDomainSocketPath, sb) == 0 {
-            // Only unlink the existing file if it is a socket
-            if sb.pointee.st_mode & S_IFSOCK == S_IFSOCK {
-                let res = unlink(unixDomainSocketPath)
-                if res == -1 {
-                    let err = errno
-                    throw IOError(errnoCode: err, reason: "Failed to unlink existing socket file \(unixDomainSocketPath)")
-                }
-            } else {
-                throw IOError(errnoCode: EEXIST, reason: "Pathname \(unixDomainSocketPath) exists and is not a socket")
+        do {
+            var sb: stat = stat()
+            try withUnsafeMutablePointer(to: &sb) { sbPtr in
+                try Posix.stat(pathname: unixDomainSocketPath, outStat: sbPtr)
             }
+
+            // Only unlink the existing file if it is a socket
+            if sb.st_mode & S_IFSOCK == S_IFSOCK {
+                try Posix.unlink(pathname: unixDomainSocketPath)
+            } else {
+                throw SocketAddressError.unixDomainSocketPathWrongType
+            }
+        } catch let err as IOError {
+            // If the filepath did not exist, we consider it cleaned up
+            if err.errnoCode == ENOENT {
+                return ();
+            }
+            throw err
         }
     }
 
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
         }
         return sock
     }
+    
+    /// Cleanup the unix domain socket.
+    ///
+    /// Deletes the associated file if it exists and has socket type.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The pathname of the UDS.
+    /// - throws: An `IOError` if the operation failed.
+    static func cleanupSocket(unixDomainSocketPath: String) throws {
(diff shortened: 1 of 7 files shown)


Lukasa:

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
...
+    
+    /// Cleanup the unix domain socket.
+    ///
+    /// Deletes the associated file if it exists and has socket type.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The pathname of the UDS.
+    /// - throws: An `IOError` if the operation failed.
+    static func cleanupSocket(unixDomainSocketPath: String) throws {
+        let sb: UnsafeMutablePointer<stat> = UnsafeMutablePointer<stat>.allocate(capacity: 1)
```
Lukasa, line 313: This pointer is leaked because it's never freed.

In this case, we want a `stat` struct. We can just stack allocate it:

```swift
var sb: stat = stat()

withUnsafePointer(to: &sb) { sbPtr in
    // Do whatever
}
```

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
...
+        if lstat(unixDomainSocketPath, sb) == 0 {
```
Lukasa, line 314: We have wrapped all our syscalls in helpers to deal with things like `EINTR`. This is necessary here as well. See `System.swift` which has a definition of `stat`. If you think using `lstat` is necessary here, then we need an equivalent wrapper.

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
...
+            // Only unlink the existing file if it is a socket
+            if sb.pointee.st_mode & S_IFSOCK == S_IFSOCK {
+                let res = unlink(unixDomainSocketPath)
```
Lukasa, line 317: We need to create a wrapped version of `unlink` as well.

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
...
+                if res == -1 {
+                    let err = errno
+                    throw IOError(errnoCode: err, reason: "Failed to unlink existing socket file \(unixDomainSocketPath)")
```
Lukasa, line 320: This error handling would move into the wrapped version of `unlink`.

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +301,29 @@ class BaseSocket: BaseSocketProtocol {
...
+                }
+            } else {
+                throw IOError(errnoCode: EEXIST, reason: "Pathname \(unixDomainSocketPath) exists and is not a socket")
```
Lukasa, line 323: Fake `IOErrors` are confusing: we should create a proper error type for this.

Sources/NIO/Bootstrap.swift
```diff
@@ -219,7 +219,16 @@ public final class ServerBootstrap {
     ///
     /// - parameters:
     ///     - unixDomainSocketPath: The _Unix domain socket_ path to bind to. `unixDomainSocketPath` must not exist, it will be created by the system.
-    public func bind(unixDomainSocketPath: String) -> EventLoopFuture<Channel> {
+    ///     - cleanupExistingSocketFile: Whether to cleanup an existing socket file at `path`.
+    public func bind(unixDomainSocketPath: String, cleanupExistingSocketFile: Bool = true) -> EventLoopFuture<Channel> {
```
Lukasa, line 223: We cannot add new parameters to existing Swift functions: that's a breaking change. The reason is that functions can be referred to by fully-qualified name, e.g.

```swift
let someFunc = ServerBootstrap.bind(unixDomainSocketPath:)
```

With this change the above function no longer exists, and the above code would break. We need a _new_ function instead.

Sources/NIO/Bootstrap.swift
```diff
@@ -855,7 +864,16 @@ public final class DatagramBootstrap {
     ///
     /// - parameters:
     ///     - unixDomainSocketPath: The path of the UNIX Domain Socket to bind on. `path` must not exist, it will be created by the system.
-    public func bind(unixDomainSocketPath: String) -> EventLoopFuture<Channel> {
+    ///     - cleanupExistingSocketFile: Whether to cleanup an existing socket file at `path`.
+    public func bind(unixDomainSocketPath: String, cleanupExistingSocketFile: Bool = true) -> EventLoopFuture<Channel> {
```
Lukasa, line 868: The same note applies here.

Sources/NIO/ServerSocket.swift
```diff
@@ -109,4 +109,17 @@
...
         }
     }
+    
+    /// Close the socket.
+    ///
+    /// After the socket was closed all other methods will throw an `IOError` when called.
+    ///
+    /// - throws: An `IOError` if the operation failed.
+    override func close() throws {
+        let maybePathname = try? self.localAddress().pathname
```
Lukasa, line 119: This extra syscall on shutdown is a bit of a pain. Can we add a stored property to `Socket` that indicates whether we think we have to unlink it?

Sources/NIO/ServerSocket.swift
```diff
@@ -109,4 +109,17 @@
...
+        try super.close()
+        if let socketPath = maybePathname {
+            try BaseSocket.cleanupSocket(unixDomainSocketPath: socketPath)
```
Lukasa, line 122: With the above in mind, should we make `cleanupSocket` a method instead of a static?

Sources/NIO/SocketAddresses.swift
```diff
@@ -187,6 +177,25 @@ public enum SocketAddress: CustomStringConvertible {
             }
         }
     }
+    
+    /// Get the pathname of a UNIX domain socket as a string
+    public var pathname: String? {
+        switch self {
+        case .v4(_):
```
Lukasa, line 184:
```suggestion
        case .v4:
```

Sources/NIO/SocketAddresses.swift
```diff
@@ -187,6 +177,25 @@ public enum SocketAddress: CustomStringConvertible {
...
+            return nil
+        case .v6(_):
```
Lukasa, line 186:
```suggestion
        case .v6:
```

Tests/NIOTests/EchoServerClientTest.swift
```diff
@@ -172,6 +172,45 @@ class EchoServerClientTest : XCTestCase {
...
 
+    func testCleanupUnixDomainSocket() throws {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        try withTemporaryUnixDomainSocketPathName { udsPath in
+            let bootstrap = ServerBootstrap(group: group)
+                .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
```
Lukasa, line 183: `so_reuseaddr` is unnecessary here.

Tests/NIOTests/EchoServerClientTest.swift
```diff
@@ -172,6 +172,45 @@ class EchoServerClientTest : XCTestCase {
...
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        try withTemporaryUnixDomainSocketPathName { udsPath in
+            // Bootstrap should not overwrite an existing file unless it is a socket
+            FileManager.default.createFile(atPath: udsPath, contents: nil, attributes: nil)
+            let bootstrap = ServerBootstrap(group: group)
+                .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
```
Lukasa, line 207: Same note here about `.so_reuseaddr`.

Lukasa (changes requested):
Thanks for this patch! I've left a few notes in the diff about some things to change.

Sources/NIO/Bootstrap.swift
AndriusA: fair, should we perhaps use `ChannelOption` instead then? Or even just piggyback on `SO_REUSEADDR`?

Lukasa: I'm disinclined to piggyback on `so_reuseaddr` tbh, it isn't the semantic we want. We could use a channel option but I think for now it's better to just add the new functions to do this. If we find a second kind of socket that has similar semantics we can consider merging them together into a channel option.

Sources/NIO/ServerSocket.swift
AndriusA: we could, but with the current `Socket` only storing `descriptor` adding the full string path representation seemed like a lot to add

Lukasa: We don't need to store that, we just need to store a bool. I want to avoid us paying this syscall for _every_ socket in the program, including all ephemeral TCP sockets. I'm fine paying for the syscall for sockets that actually require it.

Lukasa: Ah, though I double-checked and this is only on `ServerSocket`. So I don't feel that bad, actually. Let's just leave this.

Sources/NIO/ServerSocket.swift
AndriusA: we could, but would probably still need a static version of that to clean up the file before binding (or pass the `cleanupExistingSocketFile` option through many steps)

Lukasa: Fair enough.

Sources/NIO/BaseSocket.swift
AndriusA: Introduced `SocketAddressError.unixDomainSocketPathWrongType` instead

Sources/NIO/BaseSocket.swift
AndriusA: introduced `lstat` wrapper - current `fstat` expects file descriptor that we don't have in this scenario

Sources/NIO/BaseSocket.swift
AndriusA: thanks, something learned!

Sources/NIO/BaseSocket.swift
AndriusA: created a very trivial wrapped version, but didn't fully follow the logic of which are made available in what platforms, e.g. `fstat` exists in Android too, but only defined under linux and and macOS/iOS/etc. `unlink` should be available across all platforms

Sources/NIO/BaseSocket.swift
AndriusA: ended up with a mix where this function still handles the `ENOENT` case

Sources/NIO/BaseSocket.swift
Lukasa: Sorry, we can't introduce new enum cases either: that's also a semver major. We need a brand new `struct` that conforms to `Error`.

Sources/NIO/BaseSocket.swift
```diff
@@ -301,6 +306,35 @@ class BaseSocket: BaseSocketProtocol {
...
+            // Only unlink the existing file if it is a socket
+            if sb.st_mode & S_IFSOCK == S_IFSOCK {
+                try Posix.unlink(pathname: unixDomainSocketPath)
+            } else {
+                throw SocketSetupError.unixDomainSocketPathWrongType
+            }
+        } catch let err as IOError {
+            // If the filepath did not exist, we consider it cleaned up
+            if err.errnoCode == ENOENT {
+                return ();
```
Lukasa, line 333:
```suggestion
                return
```

Sources/NIO/BaseSocket.swift
```diff
@@ -18,6 +18,11 @@ import NIOConcurrencyHelpers
 import let WinSDK.INVALID_SOCKET
 #endif
 
+public enum SocketSetupError: Error {
```
Lukasa, line 21: This should be a `struct`, not an `enum`.

Sources/NIO/SocketAddresses.swift
```diff
@@ -22,6 +22,8 @@ public enum SocketAddressError: Error {
     case unsupported
     /// The requested UDS path is too long.
     case unixDomainSocketPathTooLong
+    /// The requested UDS path exists and has wrong type (not a socket).
+    case unixDomainSocketPathWrongType
```
Lukasa, line 26: We need to remove this case.

Lukasa (changes requested):
Great, looking good! A few more notes.

Sources/NIO/BaseSocket.swift
AndriusA: fixed

Lukasa (approved):

PR #1639, opened 2020-09-21
CNIOWindows: preprocess away WinSock support code

compnerd:
This preprocesses away the WinSock2 support code when on non-Windows
targets.


Lukasa (approved):
LGTM.

Issue #1558, opened 2020-06-17
tweak the defaults for the RecvAllocator to be reasonable


weissi:
Currently, the default size for the adaptive rvc allocator is I think 64 bytes. We should probably start with something around MTU because a TLS connection will have pretty large Client/ServerHellos.

PR #1641, opened 2020-09-22
Increases default initial size for `AdaptiveRecvByteBufferAllocator`

AndriusA:
### Motivation:

Resolves https://github.com/apple/swift-nio/issues/1558
A common MTU on the internet is 1500 bytes, and it should fit in the initial size buffer

### Modifications:

Only the default initial size is increased

### Result:

Little is changed, but should result in fewer reallocations e.g. during TLS connection handshake


Lukasa (approved):
LGTM, thanks!

Lukasa:
Seems like there's a unit test you need to fix up.

Lukasa (approved):

Issue #1056, opened 2019-07-02
Scheduling task in "effectively infinite" future fails not nicely


ktoso:
### Expected behavior

Some nice way of failing would be nice, like "Don't schedule tasks into `Int.max` nanoseconds into the future", that's a bad idea"! Or whichever the limit inherent to kevent is.

Likely a `fatalError` is fine for this, however perhaps provide a better error message for it than currently.

### Actual behavior

Scheduling into effectively infinite future crashes not nicely, in kevent as invalid argument:

`Fatal error: unexpected error while executing EventLoop kevent(kq:changelist:nchanges:eventlist:nevents:timeout:) failed: Invalid argument (errno: 22): file /Users/ktoso/code/swift-nio/Sources/NIO/EventLoop.swift, line 1041`
`Exited with signal code 4`

### If possible, minimal yet complete reproducer code (or URL to code)

```swift
    public func testScheduleInfinite() throws {
        let nanos: NIODeadline = .now()
        let amount: TimeAmount = .nanoseconds(.max)
        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
        defer {
            XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
        }
        let value = try eventLoopGroup.next().scheduleTask(in: amount) { 
            true
        }.futureResult.wait() // should never complete or rather error nicely
    }

```

### SwiftNIO version/commit hash

master right now; 803d36b7

### Swift & OS version (output of `swift --version && uname -a`)

Any swift version

weissi:
thanks, we should definitely not hand that to `kevent` but fail the promise ourselves with a better error or maybe never schedule it.

gcjenkinson:
I've started investigating this. I see a check in `kevent_internal()` of `ts.tv_sec & 0xFFFFFFFF00000000) != 0)` this makes sense to me as the callweheel structure (at least in FreeBSD) works on a time value specified in 32.32. Though when I create an event I can only create values < 2^30 seconds (I'll keep investigating to reconcile the two). Either way though this is a kernel restriction on the `kevent` timeout  due to limitations on data types, which in term is determined by HW clock stability.

I think - at this stage - that what is needed is better error handling for the EINVAL returned from the syscall.

Lukasa:
Resolved in #1642.

PR #1642, opened 2020-09-22
Scheduling tasks Int.max ns into future results in an obtuse fatal error (#1056)

gcjenkinson:
Motivation:

When scheduling a task too far into the future a meaningful fatal error
should be produced. As the bounds on a kevent timeout are enforced in
kernel (and are undocumented), the current error only indicates that a
parameter to the kevent syscall was invalid (EINVAL).

Modifications:

Surrounded KQueue.kevent sycall with do{} catch{}. The catch clause matches
EINVAL (one or more paramters are invalid) and then checks if this is
likely to have resulted from specifying  an invalid timeout. The bounds
checks are duplicated from those in Darwin kernel interval timer.

The rethrown IOError is caught in the SelectableEventLoop run method
and the offending task is failed (.fail()). This prevents the failure of
assert conditions in the defer statement when running as debug.

Result:

The fatal error will now output a more meaningful reason indicating the
kevent timeout bounds were exceeded (due to scheduling an event too far
into the future).

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -255,6 +255,7 @@ public protocol EventLoop: EventLoopGroup {
     ///            on the completion of the task.
     ///
     /// - note: You can only cancel a task before it has started executing.
+    /// - note: The `in` value is clamped to a maximum when running on a Darwin-kernel.
     @discardableResult
     func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T>
 
--- a/Sources/NIO/Selector.swift
+++ b/Sources/NIO/Selector.swift
@@ -372,7 +372,24 @@ internal class Selector<R: Registration> {
         case .now:
             return timespec(tv_sec: 0, tv_nsec: 0)
         case .blockUntilTimeout(let nanoseconds):
-            return timespec(timeAmount: nanoseconds)
+            // A scheduledTask() specified with a zero or negative timeAmount, will be scheduled immediately
+            // and therefore SHOULD NOT result in a kevent being created with a negative or zero timespec.
+            precondition(nanoseconds.nanoseconds > 0, "\(nanoseconds) is invalid (0 < nanoseconds)")
+
+            var ts = timespec(timeAmount: nanoseconds)
+            // Check that the timespec tv_nsec field conforms to the definition in the C11 standard (ISO/IEC 9899:2011).
+            assert((0..<1_000_000_000).contains(ts.tv_nsec), "\(ts) is invalid (0 <= tv_nsec < 1_000_000_000)")
+
+            // The maximum value in seconds supported by the Darwin-kernel interval timer (kern_time.c:itimerfix())
+            // (note that - whilst unlikely - this value *could* change).
+            let sysIntervalTimerMaxSec = 100_000_000
+
+            // Clamp the timespec tv_sec value to the maximum supported by the Darwin-kernel.
+            // Whilst this schedules the event far into the future (several years) it will still be triggered provided
+            // the system stays up.
+            ts.tv_sec = min(ts.tv_sec, sysIntervalTimerMaxSec)
+
+            return ts
         }
     }
 
--- a/Tests/NIOTests/EventLoopTest.swift
+++ b/Tests/NIOTests/EventLoopTest.swift
@@ -1295,4 +1295,26 @@ public final class EventLoopTest : XCTestCase {
             XCTFail("Not the same")
         }
     }
+
+    // Test that scheduling a task at the maximum value doesn't crash.
+    // (Crashing resulted from an EINVAL/IOException thrown by the kevent
+    // syscall when the timeout value exceeded the maximum supported by
+    // the Darwin kernel #1056).
+    public func testScheduleMaximum() {
+        let eventLoop = EmbeddedEventLoop()
+        let maxAmount: TimeAmount = .nanoseconds(.max)
+        let scheduled = eventLoop.scheduleTask(in: maxAmount) { true }
+
+        var result: Bool?
+        var error: Error?
+        scheduled.futureResult.whenSuccess { result = $0 }
+        scheduled.futureResult.whenFailure { error = $0 }
+        
+        scheduled.cancel()
+
+        XCTAssertTrue(scheduled.futureResult.isFulfilled)
+        XCTAssertNil(result)
+        XCTAssertEqual(error as? EventLoopError, .cancelled)
+    }
+
 }
--- a/Tests/NIOTests/EventLoopTest+XCTest.swift
+++ b/Tests/NIOTests/EventLoopTest+XCTest.swift
@@ -78,6 +78,7 @@ extension EventLoopTest {
                 ("testWeFailOutstandingScheduledTasksOnELShutdown", testWeFailOutstandingScheduledTasksOnELShutdown),
                 ("testSchedulingTaskOnFutureFailedByELShutdownDoesNotMakeUsExplode", testSchedulingTaskOnFutureFailedByELShutdownDoesNotMakeUsExplode),
                 ("testEventLoopGroupProvider", testEventLoopGroupProvider),
+                ("testScheduleMaximum", testScheduleMaximum),
            ]
    }
 }


Sources/NIO/System.swift
```diff
@@ -625,9 +625,21 @@ internal enum KQueue {
...
+            return try syscall(blocking: false) {
+                sysKevent(kq, changelist, nchanges, eventlist, nevents, timeout)
+            }.result
+        } catch let err as IOError where err.errnoCode == EINVAL {
+            // If kevent returned EINVAL, check if the timeout exceeded bounds impose by the kernel and if so throw
+            // a more meaningful error. These bounds are enforced by the Darwin kernel interval timer.
+            if let ts = timeout{
+                if ts.pointee.tv_sec < 0  || ts.pointee.tv_sec > 100_000_000 ||
+                     ts.pointee.tv_nsec < 0 || ts.pointee.tv_nsec >= 1_000_000_000 {
+                    throw IOError(errnoCode: err.errnoCode, reason: "kevent timeout exceeded bounds (event scheduled too far in future)")
```
Lukasa, line 638: Do we need to do this on an error-handling path? It seems like it'd be useful to just clamp the timeout before we issue the call, no?

gcjenkinson: I think its debatable as the bounds themselves are trawled up from kernel code. My approach was let the kernel say it's unhappy, and then suggest that exceeding these bounds is probably why. But I'm happy to put a check prior to calling kevent with more commenting that the bounds in kernel could change?

Lukasa: I think that approach is better. It's less complex and it avoids concerns about kernel boundaries changing, which seems highly unlikely to me.

gcjenkinson: OK. If you're setting a timer 2^29 ns into the future it was probably in error anyway, regardless of whether the kernel would accept it.

Lukasa: It might be intentional, but the joys of clamping are that the penalty isn't high. If we really did sleep that long we'll wake up a few extra times, but we'll have been asleep for so long that those wakeup are essentially free.

gcjenkinson: Just to be 100% clear by "clamp" you mean values greater than the kernel accepted maximum are limited to that value. And that, for example, specifying Int.max would not generate an error?

Lukasa: Correct.

Sources/NIO/EventLoop.swift
```diff
@@ -255,6 +255,7 @@ public protocol EventLoop: EventLoopGroup {
     ///            on the completion of the task.
     ///
     /// - note: You can only cancel a task before it has started executing.
+    /// - note: The `in` value is clamped a maximum when running on Darwin kenel-based OSes.
```
Davidde94, line 258:
```suggestion
    /// - note: The `in` value is clamped a maximum when running on Darwin kernel-based OSes.
```

Sources/NIO/System.swift
```diff
@@ -625,8 +625,25 @@ internal enum KQueue {
     @inline(never)
     @discardableResult
     public static func kevent(kq: CInt, changelist: UnsafePointer<kevent>?, nchanges: CInt, eventlist: UnsafeMutablePointer<kevent>?, nevents: CInt, timeout: UnsafePointer<Darwin.timespec>?) throws -> CInt {
+        var timeoutClamped = timeout
+        // Note that a nil timeout waits indefinitely
+        if let ts = timeout {
+            assert(ts.pointee.tv_nsec >= 0 || ts.pointee.tv_nsec < 1_000_000_000, "kevent timeout value is malformed (0 <= tv_nsec < 1_000_000_000)")
```
Davidde94, line 631: This number seems a _little_ magic to me. Might be worth linking to some documents or something to say that `0 <= tv_nsec < 1_000_000_000`

Sources/NIO/System.swift
```diff
@@ -625,8 +625,25 @@ internal enum KQueue {
...
+            // Validate that the timeout parameter passed to kevent does not exceed the limits of the Darwin kernel
+            // interval timer (note that - whilst unlikely - these values could change)
+            if ts.pointee.tv_sec < 0  || ts.pointee.tv_sec > 100_000_000 {
```
Davidde94, line 634: A similarly magic number here. As someone who doesn't know the rules of Darwin kernel timeouts, its strange to see a mixture of `100_000_000` and `1_000_000_000`

gcjenkinson: It's a value splunked out of the kernel, I'm not sure it can be anything other than magic. I can put it in a constant and let kernelItervalTImerMaxiumSecs = 100_000_000?

Sources/NIO/System.swift
gcjenkinson: These are constraints on the `tv_nsec` field of the `timespec` datatype, which are latter checked in kernel. The `tv_nsec` field must hold a value [0, 999_999_999], I'll just add a ref to the C11 ISO standard timespec datatype.

Lukasa: This condition is backwards and so all integers will pass. Negative timevals will be allowed (they will pass the second clause) and timevals greater than `1_000_000_000` will be allowed (they will pass the first clause).

We need a unit-test to validate that nothing crashes (it naturally doesn't need to wait for the task to complete, just confirm that the program crashes before the fix and doesn't crash after). That unit test would have caught this invalid assertion.

Separately, it'd be good to move this logic to closer to where `kevent` is invoked, rather than in the syscall wrapper. The syscall wrapper shouldn't paper over the idiosyncrasies of the call: the higher-level code can do that.

Finally, it may be easier to use `(0..<1_000_000_000).contains(ts.pointee.tv_nsec)` rather than try to figure out the right conditional.

Sources/NIO/System.swift
```diff
@@ -625,8 +627,26 @@ internal enum KQueue {
...
     @discardableResult
     public static func kevent(kq: CInt, changelist: UnsafePointer<kevent>?, nchanges: CInt, eventlist: UnsafeMutablePointer<kevent>?, nevents: CInt, timeout: UnsafePointer<Darwin.timespec>?) throws -> CInt {
+        var timeoutClamped = timeout
+        // Note that a nil timeout waits indefinitely
+        if let ts = timeout {
+            // Assert to check that timeout conforms to the timespec definition in the C11 standard (ISO/IEC 9899:2011)
+            assert(ts.pointee.tv_nsec >= 0 || ts.pointee.tv_nsec < 1_000_000_000, "kevent timeout value is malformed (0 <= tv_nsec < 1_000_000_000)")
+            // Validate that the timeout parameter passed to kevent does not exceed the limits of the Darwin kernel
+            // interval timer (note that - whilst unlikely - these values could change)
+            if ts.pointee.tv_sec < 0  || ts.pointee.tv_sec > sysIntervalTimerMaxSec {
```
Lukasa, line 637: Same note here about swapping to `contains` as well as moving the check out of the wrapper. The conditional is right here though.

Sources/NIO/System.swift
```diff
@@ -625,8 +627,26 @@ internal enum KQueue {
...
+                let tsClamped = UnsafeMutablePointer(mutating: ts)
+
+                // Clamp the timeout tv_sec value to the maximum accepted by the kernel
+                // (note this schedules the event many years in the future)
+                tsClamped.pointee.tv_sec = sysIntervalTimerMaxSec
+
+                timeoutClamped = UnsafePointer(tsClamped)
```
Lukasa, line 644: We can't do this. This is the Swift equivalent of casting away `const` in C, with the same affects. The Swift compiler can and will assume that this function will not mutate through this pointer, and doing so may invoke undefined behaviour.

In this case, the issue will probably go away if we hoist this logic higher up in the code as we'll probably have the `timespec` struct anyway, and so can just mutate it.

Lukasa (changes requested):

Sources/NIO/Selector.swift
```diff
@@ -513,7 +513,24 @@ internal class Selector<R: Registration> {
...
+            let sysIntervalTimerMaxSec = 100_000_000
+
+            // Validate that the timespec passed to kevent does not exceed the limits of the Darwin kernel
+            // interval timer (note that - whilst unlikely - these values could change)
+            if !(0...sysIntervalTimerMaxSec).contains(ts.tv_sec) {
+                // Clamp the timeout tv_sec value to the maximum accepted by the kernel
+                // (note this schedules the event many years in the future)
+                timespec!.tv_sec = sysIntervalTimerMaxSec
+            }
+        }
```
Lukasa, line 532: This should probably be part of `toKQueueTimeSpec`.

Sources/NIO/Selector.swift
```diff
@@ -513,7 +513,24 @@ internal class Selector<R: Registration> {
 
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-        let timespec = Selector.toKQueueTimeSpec(strategy: strategy)
+        var timespec = Selector.toKQueueTimeSpec(strategy: strategy)
+        // Note that a nil timeout waits indefinitely
+        if let ts = timespec {
+            // Assert to check that timespec conforms to the timespec definition in the C11 standard (ISO/IEC 9899:2011)
+            assert((0..<1_000_000_000).contains(ts.tv_nsec), "kevent timeout value is malformed (0 <= tv_nsec < 1_000_000_000)")
```
Lukasa, line 520: Let's have this assertion also include the actual value that was set.

Tests/NIOTests/EventLoopTest.swift
```diff
@@ -1295,4 +1295,22 @@ public final class EventLoopTest : XCTestCase {
             XCTFail("Not the same")
         }
     }
+ 
+    public func testScheduleMaximum() throws {
+        let eventLoop = EmbeddedEventLoop()
+        let maxAmount: TimeAmount = .nanoseconds(.max)
+        let scheduled = eventLoop.scheduleTask(in: maxAmount) { true }
```
Lukasa, line 1302: Can we add a comment here referencing the issue we're testing and noting that before our fix this would crash?

Lukasa (changes requested):
This is going really well now! Just a few more changes and we'll be good.

Sources/NIO/Selector.swift
```diff
@@ -372,7 +372,21 @@ internal class Selector<R: Registration> {
...
+            // Check that the timespec conforms to the definition in the C11 standard (ISO/IEC 9899:2011)
+            assert((0..<1_000_000_000).contains(ts.tv_nsec), "\(ts) is malformed (0 <= tv_nsec < 1_000_000_000)")
+
+            // The maximum value in seconds supported by the Darwin-kernel interval timer (kern_time.c:itimerfix())
+            // (note that - whilst unlikely - this value *could* change).
+            let sysIntervalTimerMaxSec = 100_000_000
+
+            // Clamp the timespec tv_sec value to the maximum supported by the Darwin-kernel,
+            // (note this schedules the event many years into the future).
+            if !(0...sysIntervalTimerMaxSec).contains(ts.tv_sec) {
```
Lukasa, line 385: Hmm, sorry for flipflopping here, but I actually think we should program this defensively. While the assert above prevents us for scheduling a negative timeout, we probably _shouldn't_ turn a negative timeout into a massive positive one. This turns EINVAL (which will alert us to an issue) into a very long delay, which seems not good. So let's only check the upper bound.

Sources/NIO/Selector.swift
```diff
@@ -513,7 +527,7 @@ internal class Selector<R: Registration> {
 
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
-        let timespec = Selector.toKQueueTimeSpec(strategy: strategy)
+        var timespec = Selector.toKQueueTimeSpec(strategy: strategy)
```
Lukasa, line 530: We don't need this change.

gcjenkinson: Yeah my mistake.

Sources/NIO/Selector.swift
gcjenkinson: Let me look at that agains. tv_sec is a time_t which if I recall correctly has some slightly weird wishy washy definition like its a integral value. I suspect it could be negative but for most practical cases won't be.

Lukasa: To be clear, the kernel will reject a negative value here. That isn't going to prevent us from _creating_ a negative value. That's the case I want to defend against.

> However, it feels like it is somewhat back to square one as we're back to letting the kernel decide what is and what isn't valid.

I disagree. The two cases have different causes.

The thing we're defending against is a user scheduling a task in the far future causing a crash. This should not happen: there is nothing wrong with scheduling a task in the far future. The kernel will reject it if we simply pass the naive value of that timeout in, so we need to tolerate that error.

Similarly, the user should be able to schedule a task to occur in the past without causing a crash! _However_, the code elsewhere in the codebase handles that conversion: we should not be creating negative timeouts by the time we get here. If we see a crash this is not _user_ error, this is our programming error. That's why its acceptable to crash here on negative values.

Putting it another way, the Linux and Darwin kernels behave differently here. This code is responsible for compatibility between the two platforms. They do not differ on negative timeout values, and so we ensure we get that right on all platforms.

Sources/NIO/Selector.swift
```diff
@@ -372,7 +372,26 @@ internal class Selector<R: Registration> {
...
+            // The maximum value in seconds supported by the Darwin-kernel interval timer (kern_time.c:itimerfix())
+            // (note that - whilst unlikely - this value *could* change).
+            let sysIntervalTimerMaxSec = 100_000_000
+
+            // Clamp the timespec tv_sec value to the maximum supported by the Darwin-kernel.
+            // Whilst this schedules the event far into the future (several years) it will still be triggered provided
+            // the system stays up.
+            if ts.tv_sec > sysIntervalTimerMaxSec {
+                ts.tv_sec = sysIntervalTimerMaxSec
+            }
```
Lukasa, line 392: Maybe nicer to spell this `ts.tv_sec = max(ts.tv_sec, sysIntervalTimerMaxSec)`.

gcjenkinson: You mean `ts.tv_sec = min(ts.tv_sec, sysIntervalTimerMaxSec)` right?

Lukasa: I do indeed, yes!

Lukasa (approved):
Nice, looks good. Thanks @gcjenkinson!

Lukasa:
Looks like you have to run the linux test generation script.

Lukasa:
Yup.

Lukasa:

Issue #1620, opened 2020-08-13
NonBlockingFileIO should return FileRegion from read/write function as well


sveamarcus:
https://github.com/apple/swift-nio/blob/a0f6a48ec0f52af49cdf8274757a2bd1ec9f3295/Sources/NIO/NonBlockingFileIO.swift#L494

After discussion with @Lukasa on Discord, it's been shown that the function should not return FileRegion as can be seen from the comments in #975. Hence, just a nit to remove FileRegion from the function docs return type.

Lukasa:
Sorry, I wasn't clear: I think this function _should_ have returned `FileRegion`. Ideally we'd add a version of this that does.

sveamarcus:
I stand corrected 👍 

Sorry for any confusion.

Lukasa:
After some discussion with @AndriusA we've come to the conclusion that there is probably limited utility in adding the `FileRegion` to the return value, so we've agreed to just fix up the documentation.

Lukasa:
Resolved by #1644.

PR #1643, opened 2020-09-22
Adds NonBlockingFileIO method returning FileRegion for read/write method

AndriusA:
### Motivation:

For consistent behaviour when opening file with/without explicit flags, want a version that returns FileRegion

Resolves https://github.com/apple/swift-nio/issues/1620

### Modifications:

Additional method in NonBlockingFileIO that uses the same code as earlier `openFile` but includes file mode and flags

### Result:

Method returning both `NIOFileHandle` and `FileRegion`

*Issue: if `mode` does not include reading, attempting to read from `FileRegion` will crash the program*


Lukasa:
Closing in favour of #1644.

PR #1645, opened 2020-09-23
CNIOWindows: add stubs for `sendmmsg` and `recvmmsg`

compnerd:
These need to be filled out, but lets stub them out for the time being
to get the symbol resolution sorted out.  These are possible to
implement using the WinSDK.

### Motivation:

### Modifications:

### Result:


Lukasa (approved):
LGTM.

PR #1647, opened 2020-09-24
NIO: implement network interface enumeration for Windows

compnerd:
Partially implement network interface enumeration.  This is sufficient
to build up the structures for basic operations.  Although the interface
information is incomplete, this provides enough structure to continue
porting the rest of the NIO interfaces.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/Interfaces.swift
+++ b/Sources/NIO/Interfaces.swift
@@ -20,6 +20,21 @@
 
 import CNIOLinux
 
+#if os(Windows)
+import let WinSDK.AF_INET
+import let WinSDK.AF_INET6
+
+import let WinSDK.INET_ADDRSTRLEN
+import let WinSDK.INET6_ADDRSTRLEN
+
+import struct WinSDK.ADDRESS_FAMILY
+import struct WinSDK.IP_ADAPTER_ADDRESSES
+import struct WinSDK.IP_ADAPTER_UNICAST_ADDRESS
+
+import typealias WinSDK.UINT8
+#endif
+
+#if !os(Windows)
 private extension ifaddrs {
     var dstaddr: UnsafeMutablePointer<sockaddr>? {
         #if os(Linux)
@@ -37,6 +52,7 @@ private extension ifaddrs {
         #endif
     }
 }
+#endif
 
 /// A representation of a single network device on a system.
 public struct NIONetworkDevice {
@@ -127,6 +143,15 @@ public struct NIONetworkDevice {
     /// This constructor will fail if NIO does not understand the format of the underlying
     /// socket address family. This is quite common: for example, Linux will return AF_PACKET
     /// addressed interfaces on most platforms, which NIO does not currently understand.
+#if os(Windows)
+    internal init?(_ pAdapter: UnsafeMutablePointer<IP_ADAPTER_ADDRESSES>,
+                   _ pAddress: UnsafeMutablePointer<IP_ADAPTER_UNICAST_ADDRESS>) {
+        guard let backing = Backing(pAdapter, pAddress) else {
+            return nil
+        }
+        self.backing = backing
+    }
+#else
     internal init?(_ caddr: ifaddrs) {
         guard let backing = Backing(caddr) else {
             return nil
@@ -134,7 +159,9 @@ public struct NIONetworkDevice {
 
         self.backing = backing
     }
+#endif
 
+#if !os(Windows)
     /// Convert a `NIONetworkInterface` to a `NIONetworkDevice`. As `NIONetworkDevice`s are a superset of `NIONetworkInterface`s,
     /// it is always possible to perform this conversion.
     @available(*, deprecated, message: "This is a compatibility helper, and will be removed in a future release")
@@ -149,6 +176,7 @@ public struct NIONetworkDevice {
             interfaceIndex: interface.interfaceIndex
         )
     }
+#endif
 
     public init(name: String,
                 address: SocketAddress?,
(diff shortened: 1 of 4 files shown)


Sources/NIO/Interfaces.swift
```diff
@@ -206,6 +227,57 @@ extension NIONetworkDevice {
...
+            case ADDRESS_FAMILY(AF_INET6):
+                self.netmask = v6mask(pAddress.pointe.OnLinkPrefixLength)
+                self.interfaceIndex = Int(pAdapter.pointee.Ipv6IfIndex)
+                break
+            default:
+                nil
+            }
+
+            // TODO(compnerd) handle broadcast/ppp/multicast information
+            self.broadcastAddres = nil
```
Lukasa, line 276: This won't compile.

Sources/NIO/Interfaces.swift
```diff
@@ -206,6 +227,57 @@ extension NIONetworkDevice {
         /// This constructor will fail if NIO does not understand the format of the underlying
         /// socket address family. This is quite common: for example, Linux will return AF_PACKET
         /// addressed interfaces on most platforms, which NIO does not currently understand.
+#if os(Windows)
+        internal init?(_ pAdapter: UnsafeMutablePointer<IP_ADAPTER_ADDRESSES>,
+                       _ pAddress: UnsafeMutablePointer<IP_ADAPTER_UNICAST_ADDRESS>) {
```
Lukasa, line 232: Do we need these pointers mutable? If not it'd be nice to ask for them `const`. It's much harder to walk that back later.

Sources/NIO/Interfaces.swift
```diff
@@ -206,6 +227,57 @@ extension NIONetworkDevice {
...
+            self.name = String(decodingCString: pAdapter.pointee.FriendlyName,
+                               as: UTF16.self)
+            self.address = pAddress.pointee.Address.lpSockaddr.convert()
+
+            let v4mask: (UINT8) -> SocketAddress? = { _ in
+                var buffer: [CChar] =
+                    Array<CChar>(repeating: 0, count: Int(INET_ADDRSTRLEN))
+                var mask: sockaddr_in = sockaddr_in)
```
Lukasa, line 240: Unbalanced parenthesis here.

Sources/NIO/Interfaces.swift
```diff
@@ -206,6 +227,57 @@ extension NIONetworkDevice {
...
+                mask.sin_family = ADDRESS_FAMILY(BSDSocket.AF_INET)
+                _ = buffer.withUnsafeMutableBufferPointer {
+                    try! BSDSocket.inet_ntop(af: BSDSocket.AF_INET, src: &mask,
+                                             dst: $0.baseAddress!,
+                                             size: INET_ADDRSTRLEN)
+                }
+                return SocketAddress(mask, host: mask.addressDescription())
+            }
+            let v6mask: (UINT8) -> SocketAddress? = { _ in
```
Lukasa, line 249: Can we get TODOs on these two functions that note that they don't _actually_ calculate the mask? It took me quite a while of looking at this code to note that they are skipping the calculation.

Sources/NIO/Interfaces.swift
```diff
@@ -206,6 +227,57 @@ extension NIONetworkDevice {
...
+                var buffer: [CChar] =
+                    Array<CChar>(repeating: 0, count: Int(INET6_ADDRSTRLEN))
+                var mask: sockaddr_in6 = sockaddr_in6()
+                mask.sin6_family = ADDRESS_FAMILY(BSDSOcket.AF_INET6)
```
Lukasa, line 253: Surprise zero.

Sources/NIO/Interfaces.swift
```diff
@@ -325,6 +398,61 @@ public final class NIONetworkInterface {
...
+        default:
+            nil
+        }
+
+        // TODO(compnerd) handle broadcast/ppp/multicast information
+        self.broadcastAddres = nil
+        self.pointToPointDestinationAddress = nil
+        self.multicastSupported = false
+    }
+#else
```
Lukasa, line 455: Given that this type is deprecated, is it worth simply making it unavailable on Windows?

Sources/NIO/Utilities.swift
```diff
@@ -130,22 +175,60 @@ public enum System {
...
     public static func enumerateDevices() throws -> [NIONetworkDevice] {
+        var devices: [NIONetworkDevice] = []
+        devices.reserveCapacity(12)  // Arbitrary choice.
+
+#if os(Windows)
+        var ulSize: ULONG = 0
+        _ = GetAdaptersAddresses(ULONG(AF_UNSPEC), 0, nil, nil, &ulSize)
+
+        let alignment: Int = MemoryLayout<IP_ADAPTER_ADDRESSES>.alignment
+        let pBuffer: UnsafeMutableRawPointer =
```
Lukasa, line 186: I think we should bind this pointer to the appropriate memory type here. This avoids the invalid use of `assumingMemoryBound` elsewhere in this code.

Sources/NIO/Utilities.swift
```diff
@@ -102,23 +109,61 @@ public enum System {
     /// - throws: If an error is encountered while enumerating interfaces.
     @available(*, deprecated, renamed: "enumerateDevices")
     public static func enumerateInterfaces() throws -> [NIONetworkInterface] {
+        var interfaces: [NIONetworkInterface] = []
```
Lukasa, line 112: Given that this function is deprecated, should we just make it unavailable on Windows?

Sources/NIO/Interfaces.swift
compnerd: I suppose that we can construct the `UnsafePointer` from the `UnsafeMutablePointer`.  They must be `UnsafeMutablePointer` for the call into the Windows SDK AFAIK.

Sources/NIO/Interfaces.swift
compnerd: Sure, I don't have a problem with Windows having a different API surface where the difference is _deprecated_ interfaces are unavailable.

Sources/NIO/Utilities.swift
compnerd: Same as above.

Sources/NIO/Interfaces.swift
Lukasa: Woof, really? If they do then we should leave them as is.

compnerd: Yes, the Windows headers don't very often mark inputs as `const` so we need things to be mutable.

Sources/NIO/Interfaces.swift
compnerd: I don't think its a TODO item, its just a NOTE right?  These functions are fed the prefix which is then converted to a netmask via `inet_ntop`.

Sources/NIO/Interfaces.swift
compnerd: Im not getting any warnings for this atm.  Am I missing something?

Sources/NIO/Utilities.swift
compnerd: I don't fully know how to `bindMemory(to:capacity:)` here, the `capacity:` being the bit that I cannot immediately see how to compute.

Sources/NIO/Interfaces.swift
Lukasa: These functions explicitly do nothing with the prefix. To be clear, exactly what happens here is:

1. Allocate an empty buffer to write the address into.
2. Create an empty sockaddr of the appropriate family. (initialized to zeros, aka the all-zero address)
3. Call `inet_ntop` to convert the empty address we just allocated to a string in the empty buffer
4. Create a SocketAddress from the empty address, ignoring the buffer.

This means that a) the buffer does nothing, and b) the mask is always the all-zero address. The prefix is never used in these functions, so there is no way it can be involved in creating the netmask.

Sources/NIO/Interfaces.swift
Lukasa: There's an `s` missing.

Sources/NIO/Utilities.swift
Lukasa: The `capacity` is `Int(ulSize) / MemoryLayout<IP_ADAPTER_ADDRESSES>.stride`.

Sources/NIO/Interfaces.swift
compnerd: Impressive, I managed to miss that upon multiple readings.

Sources/NIO/Utilities.swift
compnerd: Ah right - the capacity of the buffer is in terms of the `IP_ADAPTER_ADDRESSES`.  I was tripping myself up because of the linked-list aspect of traversal.  Thanks.

Sources/NIO/Interfaces.swift
compnerd: Ah, right.  Okay, lets make this a TODO for now, getting everything building and then fixing is going to be easier here as it will allow running.

Lukasa (approved):

Lukasa:

Lukasa:
Sorry @compnerd, seems that 5.0 and 5.1 have a bug where the compiler guards cause the `@available(*, deprecated)` annotations to be ignored, which fails our warnings-as-errors build. I know I asked you a while back to just remove the deprecated type on Windows, but I think we do need to add it back and make it work. I'm very sorry about that!

Sources/NIO/Utilities.swift
Lukasa: I think we missed doing this.

compnerd: Ah, oops, fixed.

Lukasa (approved):

PR #1653, opened 2020-09-24
NIO: ignore `SIGPIPE` on Windows

compnerd:
Windows does not support Unix-style signals.  Simply exclude the SIGPIPE
path on Windows.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -351,7 +351,11 @@ class BaseSocket: BaseSocketProtocol {
         do {
             try self.ignoreSIGPIPE()
         } catch {
+#if os(Windows)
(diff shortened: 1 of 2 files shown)


Sources/NIO/BaseSocket.swift
```diff
@@ -348,21 +348,25 @@ class BaseSocket: BaseSocketProtocol {
...
+#endif
     }
 
     deinit {
         assert(!self.isOpen, "leak of open BaseSocket")
     }
 
+#if !os(Windows)
     func ignoreSIGPIPE() throws {
         try BaseSocket.ignoreSIGPIPE(descriptor: self.descriptor)
```
Lukasa, line 367: The better version of this patch is to edit `BaseSocket.ignoreSIGPIPE`, which already contains platforms-specific code.

Lukasa (changes requested):

Lukasa (approved):

PR #1654, opened 2020-09-25
NIO: implement `cleanupSocket(unixDomainSocketPath:)` for Windows

compnerd:
This adds an initial implementation for the UDS clean up path on
Windows.  Unlike Unix, you cannot simply `stat` the destination to
determine if it is a UDS endpoint.

One must first create a file handle from the path.  Assuming that we are
able to create that, we should verify that the file that we are dealing
with is a not a disk type as the subsequent checks will fail.  Now, we
can use the file handle to query the information of the file.  This will
allow us to determine if it is a reparse point.  If not, it is
impossible for it to be a UDS as a UDS endpoint is a reparse point with
the tag IO_REPARSE_TAG_AF_UNIX.  Once we know that we have a reparse
point, we must query the kernel for information about it via the
`DeviceIoControl` system call (perform an ioctl in Unix speak).  The
returned buffer's tag will identify if we have a UDS.

For the next dance, we must now _close_ the handle as the handle being
left open will cause the subsequent `DeleteFileW` to fail due to there
being an open handle.  Once we have cleaned up the file handle that we
created, we can safely attempt to remove the file via `DeleteFileW`.

This should hopefully be sufficiently robust from possible error
scenarios for the removal of a UDS endpoint.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/CNIOWindows/include/CNIOWindows.h
+++ b/Sources/CNIOWindows/include/CNIOWindows.h
@@ -20,6 +20,36 @@
 #include <WinSock2.h>
 #include <time.h>
 
+// This is a DDK type which is not available in the WinSDK as it is not part of
+// the shared, usermode (um), or ucrt portions of the code.  We must replicate
+// this datastructure manually from the MSDN references or the DDK.
+// https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_reparse_data_buffer
+typedef struct _REPARSE_DATA_BUFFER {
+  ULONG   ReparseTag;
+  USHORT  ReparseDataLength;
+  USHORT  Reserved;
+  union {
+    struct {
+      USHORT  SubstituteNameOffset;
+      USHORT  SubstituteNameLength;
+      USHORT  PrintNameOffset;
+      USHORT  PrintNameLength;
+      ULONG   Flags;
+      WCHAR   PathBuffer[1];
+    } SymbolicLinkReparseBuffer;
+    struct {
+      USHORT  SubstituteNameOffset;
+      USHORT  SubstituteNameLength;
+      USHORT  PrintNameOffset;
+      USHORT  PrintNameLength;
+      WCHAR   PathBuffer[1];
+    } MountPointReparseBuffer;
+    struct {
+      UCHAR   DataBuffer[1];
+    } GenericReparsaeBuffer;
+  } DUMMYUNIONNAME;
+} REPARSE_DATA_BUFFER, *PREPARSE_DATA_BUFFER;
+
 #define NIO(name) CNIOWindows_ ## name
 
 typedef struct {
--- a/Sources/CNIOWindows/include/CNIOWindows.h
+++ b/Sources/CNIOWindows/include/CNIOWindows.h
@@ -20,11 +20,13 @@
 #include <WinSock2.h>
 #include <time.h>
 
+#define NIO(name) CNIOWindows_ ## name
+
 // This is a DDK type which is not available in the WinSDK as it is not part of
 // the shared, usermode (um), or ucrt portions of the code.  We must replicate
 // this datastructure manually from the MSDN references or the DDK.
 // https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_reparse_data_buffer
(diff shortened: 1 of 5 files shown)


Sources/CNIOWindows/include/CNIOWindows.h
```diff
@@ -20,6 +20,36 @@
...
+      USHORT  SubstituteNameLength;
+      USHORT  PrintNameOffset;
+      USHORT  PrintNameLength;
+      WCHAR   PathBuffer[1];
+    } MountPointReparseBuffer;
+    struct {
+      UCHAR   DataBuffer[1];
+    } GenericReparsaeBuffer;
+  } DUMMYUNIONNAME;
+} REPARSE_DATA_BUFFER, *PREPARSE_DATA_BUFFER;
```
Lukasa, line 51: Is it worth us namespacing this struct? SwiftPM has an awkward habit of being really aggressive about including header files in unrelated places, and I worry that someone else may also have forward-declared this type.

Sources/NIO/BaseSocket.swift
```diff
@@ -313,6 +352,56 @@ class BaseSocket: BaseSocketProtocol {
...
+        }
+
+        var fdi: FILE_DISPOSITION_INFO_EX = FILE_DISPOSITION_INFO_EX()
+        fdi.Flags = DWORD(FILE_DISPOSITION_FLAG_DELETE | FILE_DISPOSITION_FLAG_POSIX_SEMANTICS)
+
+        if !SetFileInformationByHandle(hFile, FileDispositionInfoEx, &fdi,
+                                       DWORD(MemoryLayout<FILE_DISPOSITION_INFO_EX>.stride)) {
+            throw IOError(windows: GetLastError(), reason: "GetFileInformationByHandle")
+        }
+#else
```
Lukasa, line 404: Let's declare this function in `BSDSocketAPI` and then split the implementation into the POSIX/Windows branches, rather than add this conditional code.

Sources/CNIOWindows/include/CNIOWindows.h
compnerd: I don't think its worth namespacing really.  Ideally this would just be provided by the `import WinSDK` (or `#include <Windows.h>`) but its a km header rather than um.

Lukasa: The reason to want to namespace it is that the bugs that manifest if this does go wrong are _extremely_ gnarly to debug.

compnerd: Namespacing it makes it harder for others to immediately connect what this is to the Microsoft side IMO.

Lukasa: I don't really think so: adding `CNIOWindows_` is not likely to confuse, and [we have prior art](https://github.com/apple/swift-nio/blob/bbb0e92702e7ac549cb9a13ab86f96e83deaac63/Sources/CNIODarwin/include/CNIODarwin.h#L29-L32).

compnerd: I suppose, Im still not particularly fond of it - it does make some things more annoying.

Lukasa (approved):

Sources/NIO/BaseSocket.swift
```diff
@@ -14,6 +14,49 @@
...
+import struct WinSDK.BY_HANDLE_FILE_INFORMATION
+import struct WinSDK.DWORD
+import struct WinSDK.FILE_DISPOSITION_INFO
+import struct WinSDK.socklen_t
+
+import CNIOWindows
+#endif
+
 /// The requested UDS path exists and has wrong type (not a socket).
 public struct UnixDomainSocketPathWrongType: Error {}
```
Lukasa, line 61: Can we move this error to `BSDSocketAPI.swift`? That'll fix the compile error in CI.

compnerd: Sure, but I dont understand why this change exposes the issue which should have existed previously.

Lukasa: It's an infrastructural issue with the way we build our syscall investigation helpers, so the refactor exposed it. I wouldn't worry about it too much. :wink:

Lukasa (approved):

Issue #1577, opened 2020-06-28
A helper for easily unwrapping optional values in an EventLoopFuture


aciidb0mb3r:
I often find myself unwrapping optional values in a future but it's kind of annoying to keep writing the boilerplate code of flatMapThrowing + gaurd let + throw error. I ended up writing this extension:

```swift
extension EventLoopFuture {
    func unwrapOptional<T>(orError error: Swift.Error) -> EventLoopFuture<T> where Value == T? {
        self.flatMapThrowing { value in
            guard let value = value else {
                throw error
            }
            return value
        }
    }
}
```

And now I am able to write code that looks like:

```swift
let sourceFile = srcTree.flatMap { tree in
    tree.lookup(path: AbsolutePath("/Sources/foo/main.swift"), in: ctx.db, ctx)
}
.unwrapOptional(orError: StringError("unable to find main.swift"))
.map{ $0.id }
.map {
    LLBArtifact.source(
          shortPath: "main.swift",
          roots: [label.asRoot, "src"],
          dataID: $0
    )
 }
```

I wonder if this helper can/should be added to NIO itself?

Lukasa:
We've generally resisted adding too many helpers directly into NIO for this kind of thing, in part because they have a tendency to proliferate. After all, here you want to throw an error, but a nil-coalescing equivalent method could also be added (`unwrapOptional(orReplace:)`). Then we start ending up pretty deep into the flow control story.

_However_, for optional specifically I'm a bit tempted to say yes. Optional holds a somewhat special place in Swift, and it's not unreasonable to want to reflect this in the Future APIs. So I'm tentatively in favour. @Davidde94, @glbrntt, @PeterAdams-A, thoughts?

Davidde94:
@Lukasa agree with your points - I'd say that if we think there's a common use case for this (or there will be) then the implementation looks clean/simple enough to get it in (probably as a "good first issue").

0xTim:
FYI Vapor has an `unwrap(or:)` function that does exactly this (and I much prefer that cleaner callsite as well)

gcjenkinson:
I have been working up the following:

```
func unwrap<T>(orError error: Error) throws -> EventLoopFuture<T> where Value == T?
func unwrap<T>(orReplace replace: T) -> EventLoopFuture<T> where Value == T?
func unwrap<T>(orElse: @escaping () -> T) -> EventLoopFuture<T> where Value == T?
```

[...]

PR #1656, opened 2020-09-25
A helper for easily unwrapping Optional values in an EventLoopFuture (#1577)

gcjenkinson:
Motivation:

Unwrapping an `Optional` value from an `EventLoopFuture` is a fairly
common requirement that currently involves the client writing
boilerplate code, for example:

```
extension EventLoopFuture {
    func unwrapOptional<T>(orError error: Swift.Error) -> EventLoopFuture<T> where Value == T? {
        self.flatMapThrowing { value in
            guard let value = value else {
                throw error
            }
            return value
        }
    }
}
```

As this is a fairly common requirement adding an extension of
`EventLoopFuture` to unwrap `Optional` values would remove this
burden from clients.

Modifications:

Added Extension to `EventLoopFuture` containing the following functions:

- `unwrap<T>(or error: Error)`: Unwraps a future returning a new
  `EventLoopFuture` with the same value as the resolved future when
  its value is Optional.some(...)`, otherwise the `Error` passed in
  the `or` parameter is thrown

- func unwrap<T>(or replace: T)`: Unwraps a future returning a new
  `EventLoopFuture` with either: the value passed in the `or`
  parameter when the future resolved with value `Optional.none`, or
  the same value otherwise.

- func unwrap<T>(or orElse: @escaping ()- > T): Unwraps a future
  returning a new `EventLoopFuture` with either: the value returned
  by the colused passed in the `or` parameter when the future
  resolved with value `Optional.none`, or the same value otherwise.

Added new unit tests for each new `unwrap(or:)` function.

Result:

Client's no longer have to write their own boilerplate code.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -1325,3 +1325,80 @@ extension EventLoopFuture {
         return self
     }
 }
+
+// MARK: unwrap
+
+extension EventLoopFuture {
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwrap a future returning a new `EventLoopFuture`. When the resolved future's value is `Optional.some(...)`
+    /// the new future is created with the identical value. Otherwise the `Error` passed in the `orError` parameter
+    /// is thrown. For example:
+    /// ```
+    /// do {
+    ///     try promise.futureResult.unwrap(orError: ErrorToThrow).wait()
+    /// } catch ErrorToThrow {
+    ///     ...
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - orError: the `Error` that is thrown when then resolved future's value is `Optional.none`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and the same value as the resolved
+    ///     future.
+    /// - throws: the `Error` passed in the `orError` parameter when the resolved future's value is `Optional.none`.
+    @inlinable
+    public func unwrap<NewValue>(orError error: Error) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
+        return self.flatMapThrowing { (value) throws -> NewValue in
+            guard let value = value else {
+                throw error
+            }
+            return value
+        }
+    }
+
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwraps a future returning a new `EventLoopFuture` with either: the value passed in the `orReplace`
+    /// parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
+    /// ```
+    /// promise.futureResult.unwrap(orReplace: 42).wait()
+    /// ```
+    ///
+    /// - parameters:
+    ///     - orReplace: the value of the returned `EventLoopFuture` when then resolved future's value is `Optional.some()`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and the value passed in the `orReplace` parameter.
+    @inlinable
+    public func unwrap<NewValue>(orReplace replacement: NewValue) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
+        return self.map { (value) -> NewValue in
+            guard let value = value else {
+                return replacement
+            }
+            return value 
+        }
+    }
+
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwraps a future returning a new `EventLoopFuture` with either: the value returned by the closure passed in
+    /// the `orElse` parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
+    /// ```
+    /// var x = 2
+    /// promise.futureResult.unwrap(orElse: { x * 2 }).wait()
+    /// ```
+    ///
+    /// - parameters:
+    ///     - orElse: a closure that returns the value of the returned `EventLoopFuture` when then resolved future's value
+    ///         is `Optional.some()`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and with the value returned by the closure
+    ///     passed in the `orElse` parameter.
+    @inlinable
+    public func unwrap<NewValue>(orElse callback: @escaping () -> NewValue) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
+        return self.map { (value) -> NewValue  in
+            guard let value = value else {
+                return callback()
+            }
+            return value 
+        }
+    }
+}
--- a/Tests/NIOTests/EventLoopFutureTest.swift
+++ b/Tests/NIOTests/EventLoopFutureTest.swift
@@ -1207,5 +1207,62 @@ class EventLoopFutureTest : XCTestCase {
         
         try exitPromise.futureResult.wait()
     }
+ 
+    func testEventLoopFutureOrErrorNoThrow() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(42)
+        promise.completeWith(result)
+
+        XCTAssertEqual(try promise.futureResult.unwrap(orError: EventLoopFutureTestError.example).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrThrows() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(nil)
+        promise.completeWith(result)
+
+        XCTAssertThrowsError(try _ = promise.futureResult.unwrap(orError: EventLoopFutureTestError.example).wait()) { (error) -> Void in
+            XCTAssertEqual(error as! EventLoopFutureTestError, EventLoopFutureTestError.example)
+        }
+    }
+  
+     func testEventLoopFutureOrNoReplacement() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(42)
+        promise.completeWith(result)
+
+        XCTAssertEqual(try! promise.futureResult.unwrap(orReplace: 41).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrReplacement() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(nil)
+        promise.completeWith(result)
+
+        XCTAssertEqual(try! promise.futureResult.unwrap(orReplace: 42).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrNoElse() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(42)
+        promise.completeWith(result)
+
+        XCTAssertEqual(try! promise.futureResult.unwrap(orElse: { 41 } ).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrElse() {
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(4)
+        promise.completeWith(result)
+
+        let x = 2
+        XCTAssertEqual(try! promise.futureResult.unwrap(orElse: { x * 2 } ).wait(), 4)
+    }
 
 }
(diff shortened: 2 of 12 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,77 @@ extension EventLoopFuture {
...
+    ///     ...
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - or: the `Error` that is thrown when then resolved future's value is `Optional.none`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `T` and the same value as the resolved
+    ///     future.
+    /// - throws: the `Error` passed in the `or` parameter when the resolved future's value is `Optional.none`.
+    func unwrap<T>(or error: Error) -> EventLoopFuture<T> where Value == T? {
```
Lukasa, line 1349: Two minor nits: I'd rather spell this `T?` as `Optional<T>`, and then can we use `NewValue` as the generic type parameter? Single letter type parameters aren't great.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,77 @@ extension EventLoopFuture {
...
+        self.flatMapThrowing { value in
```
Lukasa, line 1350: We need explicit `return` as we support 5.0.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,77 @@ extension EventLoopFuture {
...
+    /// Unwraps a future returning a new `EventLoopFuture` with either: the value passed in the `or`
+    /// parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
+    /// ```
+    /// promise.futureResult.unwrap(or: 42).wait()
+    /// ```
+    ///
+    /// - parameters:
+    ///     - or: the value of the returned `EventLoopFuture` when then resolved future's value is `Optional.some()`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `T` and the value passed in the `or` parameter.
+    func unwrap<T>(or replace: T) -> EventLoopFuture<T> where Value == T? {
```
Lukasa, line 1369:
```suggestion
    func unwrap<T>(or replacement: T) -> EventLoopFuture<T> where Value == T? {
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,77 @@ extension EventLoopFuture {
...
+        self.flatMapThrowing { value in
```
Lukasa, line 1370: This isn't `flatMapThrowing` as the closure doesn't throw, it's just `map`.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,77 @@ extension EventLoopFuture {
...
+    /// promise.futureResult.unwrap(or: { x * 2 }).wait()
+    /// ```
+    ///
+    /// - parameters:
+    ///     - or: a closure that returns the value of the returned `EventLoopFuture` when then resolved future's value
+    ///         is `Optional.some()`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `T` and with the value returned by the closure
+    ///     passed in the `or` parameter.
+    func unwrap<T>(or orElse: @escaping () -> T) -> EventLoopFuture<T> where Value == T? {
+        self.flatMapThrowing { value in
```
Lukasa, line 1393: Similarly, this is `map`.

Tests/NIOTests/EventLoopFutureTest.swift
```diff
@@ -1207,5 +1207,63 @@ class EventLoopFutureTest : XCTestCase {
...
+        promise.completeWith(result)
+
+        XCTAssertEqual(try promise.futureResult.unwrap(or: EventLoopFutureTestError.example).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrThrows() {
+        XCTAssertThrowsError(try unwrapOrErrorThatThrows())
+    }
+
+    func unwrapOrErrorThatThrows() throws {
```
Lukasa, line 1224: It's usually nice to make helper functions like this `private`.

Lukasa: As this helper function is only in one place, can we just move it into `testEventLoopFutureOrThrows`? It's fine to make test functions throwing.

Tests/NIOTests/EventLoopFutureTest.swift
```diff
@@ -1207,5 +1207,63 @@ class EventLoopFutureTest : XCTestCase {
...
+        let eventLoop = EmbeddedEventLoop()
+        let promise = eventLoop.makePromise(of: Int?.self)
+        let result: Result<Int?, Error> = .success(42)
+        promise.completeWith(result)
+
+        XCTAssertEqual(try promise.futureResult.unwrap(or: EventLoopFutureTestError.example).wait(), 42)
+    }
+ 
+    func testEventLoopFutureOrThrows() {
+        XCTAssertThrowsError(try unwrapOrErrorThatThrows())
```
Lukasa, line 1221: Can we assert the error we expected to be thrown was thrown?

Lukasa (changes requested):
Great, broadly looks good! Left some notes in the diff.

Sources/NIO/EventLoopFuture.swift
gcjenkinson: Yeah. I did change it to `flatMap` at one stage and it through up loads of errors. I obvious forgot about resolving that, I'll look again.

Lukasa: Note that it's `map`, not `flatMap`.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,78 @@ extension EventLoopFuture {
...
+    ///     ...
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - or: the `Error` that is thrown when then resolved future's value is `Optional.none`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `T` and the same value as the resolved
+    ///     future.
+    /// - throws: the `Error` passed in the `or` parameter when the resolved future's value is `Optional.none`.
+    func unwrap<NewValue>(or error: Error) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
```
Lukasa, line 1350: Can we add an `@inlinable` declaration to these three methods, please?

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,78 @@ extension EventLoopFuture {
...
+    /// parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
+    /// ```
+    /// promise.futureResult.unwrap(or: 42).wait()
+    /// ```
+    ///
+    /// - parameters:
+    ///     - or: the value of the returned `EventLoopFuture` when then resolved future's value is `Optional.some()`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `T` and the value passed in the `or` parameter.
+    func unwrap<NewValue>(or replacement: NewValue) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
+        self.map{ (value) -> NewValue in
```
Lukasa, line 1371:
```suggestion
        self.map { (value) -> NewValue in
```

Lukasa (commented):
Two minor notes and then we're good to go I think

Sources/NIO/EventLoopFuture.swift
glbrntt: I'm a little uncomfortable with these overloads; I think `unwrap(orError:)` and `unwrap(orElse:)` would be more readable and avoid any ambiguity in the case of `Value` also being an error.

glbrntt: (Maybe `orElse` isn't necessary if we have `orError`)

Lukasa: Oh good catch, that could _absolutely_ be awkward.

Lukasa (approved):
LGTM. @glbrntt?

Lukasa:
Looks like the tests need an update.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+
+extension EventLoopFuture {
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwrap a future returning a new `EventLoopFuture`. When the resolved future's value is `Optional.some(...)`
+    /// the new future is created with the identical value. Otherwise the `Error` passed in the `or` parameter
+    /// is thrown. For example:
+    /// ```
+    /// do {
+    ///     try promise.futureResult.unwrap(or: ErrorToThrow).wait()
```
glbrntt, line 1339:
```suggestion
    ///     try promise.futureResult.unwrap(orError: ErrorToThrow).wait()
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
     }
 }
+
+// MARK: unwrap
+
+extension EventLoopFuture {
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwrap a future returning a new `EventLoopFuture`. When the resolved future's value is `Optional.some(...)`
+    /// the new future is created with the identical value. Otherwise the `Error` passed in the `or` parameter
```
glbrntt, line 1335:
```suggestion
    /// the new future is created with the identical value. Otherwise the `Error` passed in the `orError` parameter
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+            guard let value = value else {
+                throw error
+            }
+            return value
+        }
+    }
+
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwraps a future returning a new `EventLoopFuture` with either: the value passed in the `or`
```
glbrntt, line 1362:
```suggestion
    /// Unwraps a future returning a new `EventLoopFuture` with either: the value passed in the `orReplace`
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+    /// parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
+    /// ```
+    /// promise.futureResult.unwrap(or: 42).wait()
```
glbrntt, line 1365:
```suggestion
    /// promise.futureResult.unwrap(orReplace: 42).wait()
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+                return replacement
+            }
+            return value 
+        }
+    }
+
+    /// Unwrap an `EventLoopFuture` where its type parameter is an `Optional`.
+    ///
+    /// Unwraps a future returning a new `EventLoopFuture` with either: the value returned by the closure passed in
+    /// the `or` parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
```
glbrntt, line 1384:
```suggestion
    /// the `orElse` parameter when the future resolved with value Optional.none, or the same value otherwise. For example:
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+    /// ```
+    /// var x = 2
+    /// promise.futureResult.unwrap(or: { x * 2 }).wait()
```
glbrntt, line 1387:
```suggestion
    /// promise.futureResult.unwrap(orElse: { x * 2 }).wait()
```

glbrntt (approved):
Looks good minus the tests. I left some documentation fixes.

Sources/NIO/EventLoopFuture.swift
gcjenkinson: Annoyingly I made all these changes and they got lost due to multitasking confusing :(

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1325,3 +1325,81 @@ extension EventLoopFuture {
...
+    /// }
+    /// ```
+    ///
+    /// - parameters:
+    ///     - orError: the `Error` that is thrown when then resolved future's value is `Optional.none`.
+    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and the same value as the resolved
+    ///     future.
+    /// - throws: the `Error` passed in the `orError` parameter when the resolved future's value is `Optional.none`.
+    @inlinable
+    func unwrap<NewValue>(orError error: Error) -> EventLoopFuture<NewValue> where Value == Optional<NewValue> {
```
glbrntt, line 1351: These all need to be `public`!

glbrntt (changes requested):
Sorry @gcjenkinson, I just realised that these functions are all `internal`!

Lukasa (approved):
Once @glbrntt's note about visibility is addressed, this looks good to go!

glbrntt (approved):

PR #1663, opened 2020-09-30
Minor cleanups of Windows compatability code.

Lukasa:
Motivation:

Cleanup some noise that came in with some previous Windows compat
patches.

Modifications:

- Move INVALID_SOCKET to NIOBSDSocket, and refer to it by a better name
  there.
- Comment in an empty block to indicate that it's supposed to be empty.

Result:

Bit cleaner code.

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -14,10 +14,6 @@
 
 import NIOConcurrencyHelpers
 
-#if os(Windows)
-import let WinSDK.INVALID_SOCKET
-#endif
-
 /// The requested UDS path exists and has wrong type (not a socket).
 public struct UnixDomainSocketPathWrongType: Error {}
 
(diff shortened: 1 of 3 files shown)


Sources/NIO/BaseSocket.swift
```diff
@@ -218,7 +214,7 @@ class BaseSocket: BaseSocketProtocol {
     private var descriptor: NIOBSDSocket.Handle
     public var isOpen: Bool {
         #if os(Windows)
-            return descriptor != WinSDK.INVALID_SOCKET
+            return descriptor != NIOBSDSocket.invalidHandle
```
glbrntt, line 217: Can we blast away the whole `#if-#else-#endif` block here leaving just this line? Or do we expect negative descriptors which aren't `-1`?

Lukasa: I think we're just being extra cautious. We could get rid of it, but I'd prefer to do it by factoring out this code entirely but keeping the defensiveness.

glbrntt: Sounds reasonable.

glbrntt (approved):

Issue #1436, opened 2020-03-08
HTTPObjectAggregator implementation


m0wfo:
### Expected behavior
Netty's [HttpObjectAggregator](https://netty.io/4.1/api/io/netty/handler/codec/http/HttpObjectAggregator.html) significantly simplifies implementation of HTTP APIs where raw throughput is not a concern. Happy to port this if needed, just checking here that I'm not duplicating any work.

Downstream handlers would optionally be able to deal with subtypes of `FullHttpRequest` and `FullHttpResponse` rather than individual parts.

### Actual behavior

HTTP channel handlers must track state of individual parts of an HTTP message they have sent/received.

### Steps to reproduce

Write a SwiftNIO handler which deals with HTTP :)

### If possible, minimal yet complete reproducer code (or URL to code)

_[anything to help us reproducing the issue]_

### SwiftNIO version/commit hash

HEAD, AFAIK

### Swift & OS version (output of `swift --version && uname -a`)

latest, pick any OS

Lukasa:
I think this would be a good addition to the project!

PR #1664, opened 2020-09-30
HTTPObjectAggregator implementation

AndriusA:
HTTPObjectAggregator implementation

### Motivation:

Resolves https://github.com/apple/swift-nio/issues/1436

Netty's [HttpObjectAggregator](https://netty.io/4.1/api/io/netty/handler/codec/http/HttpObjectAggregator.html) significantly simplifies implementation of HTTP APIs where raw throughput is not a concern.

Downstream handlers would be able to deal with `HTTPServerRequestFull` and `HTTPClientResponseFull` rather than individual parts.

### Modifications:

Introduces new message types `HTTPServerRequestFull` `HTTPClientResponseFull` that include `HTTPRequestHead` and `HTTPResponseHead` respectively and `ByteBuffer` for message body.

The two `ChannelInboundHandler`s aggregate message parts until a message is completed, then passing the aggregated message to the next handler.

Additionally adds convenience extensions to handle `content-length` and `expect` headers.

### Result:

Library users can add `HTTPServerRequestAggregator` or `HTTPClientResponseAggregator` to their inbound channel handlers to aggregate message parts into complete `HTTPServerRequestFull` or `HTTPClientResponseFull` respectively. The handlers also take care of some error signalling back to the client on server side as well as discarding oversize messages in the case of ill-behaved remote endpoints.


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -1435,3 +1435,118 @@ extension HTTPMethod: RawRepresentable {
...
+    }
+
+    public var contentLength: Int? {
+        return headers.contentLength
+    }
+}
+
+extension HTTPHeaders {
+    var hasContentLength: Bool {
+        return self.contains(name: "content-length") && self[canonicalForm: "content-length"].count == 1
```
Lukasa, line 1475: This loops over the headers twice. We can remove `contains` and just use the `canonicalForm` check, as it's redundant with the previous one.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,408 @@
...
+            if self.shouldCloseAfterContinueResponse(response: response) {
+                self.state.closed()
+            } else {
+                // Won't close the connection, but handling an oversized message
+                self.state.handlingOversizeMessage()
+                // Remove `Expect` header from the aggregated message
+                self.fullMessageHead?.headers.remove(name: "expect")
+            }
+            return response
+        } else if request.hasContentLength && request.contentLength! > self.maxContentLength {
```
Lukasa, line 200: No need to check `hasContentLength` here either, just use the conditional binding.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -1435,3 +1435,118 @@ extension HTTPMethod: RawRepresentable {
...
+    }
+
+    var contentLength: Int? {
+        return self.hasContentLength ? Int(self.first(name: "content-length") ?? "") : nil
```
Lukasa, line 1479: This also loops redundantly: we don't need to preflight whether `hasContentLength` is true, as we'll answer that by using `self.first(name:)` anyway. Also, let's not ask Swift to try to work out if `""` is a valid `Int`:

```swift
self.first(name: "content-length").flatMap { Int($0) }
```

is just as good a spelling here.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -1435,3 +1435,118 @@ extension HTTPMethod: RawRepresentable {
...
+    func isUnsupportedExpectation(version: HTTPVersion) -> Bool {
+        if version.major == 1 && version.minor >= 1 {
+            var unsupported: Bool? = nil
+            for word in self[canonicalForm: "expect"] {
+                if word.utf8.compareCaseInsensitiveASCIIBytes(to: "100-continue".utf8) {
+                    // if we see multiple values, that's clearly bad and we default to 'unsupported'
+                    unsupported = unsupported != nil ? true : false
+                } else {
+                    unsupported = true
+                }
```
Lukasa, line 1491: Hmm, this seems to say that we expect only one header field. If we do we can skip a bunch of this mess by removing the use of `canonicalForm` and just checking the headers we get back. There should only be one, and it should exactly match the `100-continue` string.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -1435,3 +1435,118 @@ extension HTTPMethod: RawRepresentable {
...
+                if word.utf8.compareCaseInsensitiveASCIIBytes(to: "100-continue".utf8) {
+                    // if we see multiple values, that's unsupported and we default to 'false'
+                    cont = cont != nil ? false : true
+                }
+            }
+            return cont ?? false
+        } else {
+            return false
+        }
+    }
```
Lukasa, line 1512: I think we should be able to remove some redundancy between these two methods.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -1435,3 +1435,118 @@ extension HTTPMethod: RawRepresentable {
...
+}
+
+extension HTTPResponseStatus {
+    public var clientErrorClass: Bool {
```
Lukasa, line 1516: Let's not make this public if we can avoid it.

Lukasa (changes requested):
Great, thanks for this patch! I've left some notes in the diff.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,408 @@
...
+
+import NIO
+
+/// The parts of a complete HTTP message, either request or response.
+///
+/// A HTTP message is made up of a request or status line with several headers,
+/// encoded by `.head`, zero or more body parts, and optionally some trailers. To
+/// indicate that a complete HTTP message has been sent or received, we use `.end`,
+/// which may also contain any trailers that make up the mssage.
+public struct HTTPMessageFull<HeadT: Equatable, BodyT: Equatable> {
```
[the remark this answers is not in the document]
AndriusA: very little, though I would have found reusing some parts of the code if the original `HTTPClient*` and `HTTPServer*` shared parent classes

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    mutating func handlingOversizeMessage() {
+        switch self {
+        case .receiving, .idle:
+            self = .ignoringContent
+        case .ignoringContent, .closed:
+            debugOnly {
+                preconditionFailure("Received payload too large in state \(self)")
+            }
```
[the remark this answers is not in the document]
AndriusA: thought `assertionFailure` also gets removed from release builds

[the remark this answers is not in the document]
Lukasa: So does `debugOnly { preconditionFailure }`. That's what `debugOnly` means ;)

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,408 @@
...
+            // does not match the actual body, but the number of bytes that would be
+            // transmitted if a GET would have been used.
+            //
+            // See rfc2616 14.13 Content-Length
+
+            if !aggregated.hasContentLength && self.buffer.readableBytes > 0 {
+                aggregated.headers.replaceOrAdd(
+                    name: "content-length",
+                    value: String(self.buffer.readableBytes))
+            }
```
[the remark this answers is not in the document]
AndriusA: Direct copy of equivalent code in Netty's HTTPObjectAggregator, but merely signals content length to the receiver. Agree that this does not fully handle all the cases from RFC7230 when it should be excluded though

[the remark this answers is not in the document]
Lukasa: I'm inclined to say that doing this in an incomplete way is worse than not doing it at all: we run the risk of creating malformed HTTP headers, and we end up doing work we didn't need to do most of the time. I'd say for now we should just omit this code.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,408 @@
...
+    func generateContinueResponse(context: ChannelHandlerContext, request: HTTPRequestHead) -> HTTPResponseHead? {
+        if request.isUnsupportedExpectation {
+            // if the request contains an unsupported expectation, we return 417
+            context.fireUserInboundEventTriggered(HTTPObjectAggregatorEvents.httpExpectationFailedEvent)
+            return HTTPResponseHead(
+                version: .init(major: 1, minor: 1),
+                status: .expectationFailed,
+                headers: HTTPHeaders([("Content-Length", "0")]))
+        } else if request.isContinueExpected {
+            if !request.hasContentLength || request.contentLength! <= self.maxContentLength {
```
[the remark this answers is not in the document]
AndriusA: we do now

Sources/NIOHTTP1/HTTPTypes.swift
AndriusA: moved the overlapping logic out into a private function

Sources/NIOHTTP1/HTTPTypes.swift
AndriusA: just doing something like `self.first(name: "expect")...` would fail to check that there is only one, have kept canonical form for now but simplified the logic

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,422 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+/// The parts of a complete HTTP response from the view of the client.
+///
+/// Afull HTTP request is made up of a response header encoded by `.head`
```
Lukasa, line 19: Nit:

```suggestion
/// A full HTTP request is made up of a response header encoded by `.head`
```

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,422 @@
...
+        self.head = head
+        self.body = body
+    }
+}
+
+extension HTTPServerRequestFull: Equatable {}
+
+/// The parts of a complete HTTP response from the view of the client.
+///
+/// Afull HTTP response is made up of a response header encoded by `.head`
```
Lukasa, line 35: Nit: 

```suggestion
/// Afull HTTP response is made up of a response header encoded by `.head`
```

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,422 @@
...
+
+    public init(head: HTTPResponseHead, body: ByteBuffer?) {
+        self.head = head
+        self.body = body
+    }
+}
+
+extension HTTPClientResponseFull: Equatable {}
+
+public enum HTTPObjectAggregatorError: Error {
```
Lukasa, line 49: So we probably need to replace these errors with a custom struct error. This can carry an error code, but that error code cannot be an `enum`.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,422 @@
...
+    case finishedIgnoringContent
+}
+
+extension HTTPObjectAggregatorError: Equatable {}
+
+public enum HTTPObjectAggregatorEvents {
+    case httpExpectationFailedEvent
+    case httpContinueEvent
+    case httpFrameTooLongEvent
+}
```
Lukasa, line 63: The same is true here: we probably want to make this a struct that hides an `enum`. The pattern here would be something like this:

```swift
public struct HTTPObjectAggregatorEvent: Hashable {
    private enum Base {
        case httpExpectationFailedEvent
        case httpContinueEvent
        case httpFrameTooLongEvent
    }

    private var base: Base

    private init(base: Base) {
        self.base = base
    }

    public static let httpExpectationFailedEvent = HTTPObjectAggregatorEvent(base: . httpExpectationFailedEvent)
    public static let httpContinueEvent = HTTPObjectAggregatorEvent(base: . httpContinueEvent)
    public static let httpFrameTooLongEvent = HTTPObjectAggregatorEvent(base: . httpFrameTooLongEvent)
}
```

This somewhat awkward pattern is necessary because you cannot add new cases to Swift `enum`s without breaking existing source. The same is not true for the above, which is essentially an "open" enumeration.

While we're here, the word `event` is redundant and so we can probably remove it from the enum cases and `static let`s. It's in the type, so that's sufficient.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,422 @@
...
+
+    mutating func messageEndReceived() throws {
+        switch self {
+        case .receiving:
+            // Got the request end we were waiting for.
+            self = .idle
+        case .ignoringContent:
+            // Finished ignoring message contents
+            self = .idle
+            throw HTTPObjectAggregatorError.finishedIgnoringContent
```
Lukasa, line 106: Is this a real error? What are we trying to do here?

Lukasa (changes requested):
Ah, shoot. I'm sorry @AndriusA, I forgot a fairly substantial note in the previous review. While I was reading it I had a vague niggle about there being an outstanding bug that overlaps with this PR in some way, but forgot to follow up.

I have now: this interacts nastily with the fact that we don't currently have _actual_ support for handling 100 Continue in the rest of the HTTP stack. You can see #1422 for an in-progress (albeit stalled) patch that aims to address that, but there's some outstanding work still to be done there to deal with the conversation in [this thread](https://github.com/apple/swift-nio/pull/1422/files/e22a49b240ecc1b173f9df6eb0417cd9b405185c#diff-276d1584d63984ae4bd16c7a6a1549d8).

I don't think it's a good idea to block this PR on that one, so I think instead we should just _remove_ the 100-continue functionality from this PR. I appreciate it's a bit of a loss of functionality, but we can always add it back in later. If we merged it but prevented it from being used it would almost certainly rot as long as #1422 remained unfixed.

If you're super motivated, you are more than welcome to take a swing at updating #1422. I doubt @t089 would mind too much, considering their personal responsibilities! Otherwise, feel free to remove the 100-continue functionality from this PR and I can review in that context!

Sorry for that. I even only remembered partway through the review of this patch, so I have _also_ left a few notes in the diff. They're substantially less meaningful than the 100-continue functionality.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
AndriusA: this would be an expected transition when an oversize message ends from essentially an "erroring" state where any content received is getting discarded. Throwing an error avoids going into dispatching the aggregated message and instead just cleaning up

Lukasa: I'd love a comment to that effect here to explain why we are throwing an error here, and perhaps the error should be a named so as to be clear to the user who catches it. Right now I think its name favours the implementer, not the catcher.

AndriusA:
> If you're super motivated, you are more than welcome to take a swing at updating #1422. I doubt @t089 would mind too much, considering their personal responsibilities! Otherwise, feel free to remove the 100-continue functionality from this PR and I can review in that context!

I'll stash finishing #1422 away in a todo..

Though for correctness, wouldn't it be necessary for the server to respond with `expectationFailed` and close the connection if `Expect: 100-continue` is received? Might be misreading RFC 7231 here:

> An origin server MUST, upon receiving an HTTP/1.1 (or later)
>    request-line and a complete header section that contains a
>    100-continue expectation and indicates a request message body will
>    follow, either send an immediate response with a final status code,
>    if that status can be determined by examining just the request-line
>    and header fields, or send an immediate 100 (Continue) response to
>    encourage the client to send the request's message body.

If we don't do the latter bit, we should send a final status code and close?

Lukasa:
You aren't misreading that, but it is less damaging to not send the 100 than to send the immediate unsupported expectation response.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
AndriusA: used the pattern for both and made the error throwing much more verbose (including bubbling everything up via `fireErrorCaught`) - not sure if not too verbose though?

Lukasa: I think it's fine: it's the price we have to pay.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,405 @@
...
+        var serverResponse: HTTPResponseHead? = nil
+
+        do {
+            switch msg {
+            case .head(let httpHead):
+                try self.state.messageHeadReceived()
+                try serverResponse = self.beginAggregation(context: context, request: httpHead, message: msg)
+                if serverResponse?.status == .payloadTooLarge {
+                    try self.state.handlingOversizeMessage()
+                }
```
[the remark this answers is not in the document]
AndriusA: indeed, mixed things up while refactoring, thanks for the catch. Actually might be best to move the state transition to server response dispatch - `self.state.handlingOversizeMessage()` should only happen when we also send the `.payloadTooLarge` response

Sources/NIOHTTP1/HTTPObjectAggregator.swift
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    mutating func handlingOversizeMessage() throws {
+        switch self {
+        case .receiving, .idle:
+            self = .ignoringContent
+        case .ignoringContent:
+            // If we are already ignoring content, do not want to proceed to
+            // sending entity too large response
+            throw HTTPObjectAggregatorError.frameTooLong
```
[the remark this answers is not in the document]
AndriusA: In principle this is (and should be) unreachable, might better to use precondition then? Can only get into this if received either `.head` or `.body`, both of which start with a state transition that throws if already in `.ignoringContent` state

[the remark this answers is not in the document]
Lukasa: Yeah, that works for me.

Sources/NIOHTTP1/HTTPObjectAggregator.swift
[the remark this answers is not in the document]
Lukasa: Sounds good.

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
```diff
@@ -0,0 +1,400 @@
...
+                self.endAggregation(context: context, trailingHeaders: trailingHeaders)
+            }
+        } catch let error as NIOHTTPObjectAggregatorError {
+            context.fireErrorCaught(error)
+            // Won't be able to complete those
+            self.fullMessageHead = nil
+            self.buffer.clear()
+        } catch {
+            // Ignore other types of errors
+        }
```
Lukasa, line 225: Though this branch is unreachable, I think it'd be better if we threw unexpected error types down the pipeline with `fireErrorCaught`.

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
```diff
@@ -0,0 +1,400 @@
...
```
Lukasa, line 357: Same note here.

Lukasa (commented):
Two small notes otherwise this looks really good.

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
AndriusA: that's sensible, fixed

Lukasa (approved):

PR #1666, opened 2020-10-02
NIOUDPEchoServer use AddressedEnvelope as InboundIn and OutboundOut

fabianfett:
As discussed on Slack

### Motivation:

Currently the Channel types don't match between `NIOUDPEchoServer` and `NIOUDPEchoClient`. The correct type to use is `AddressedEnvelope<ByteBuffer>` in the server. 

So far this has only worked since `unwrapInboundIn(data)` has not been called in the server.

### Modifications:

The `NIOUDPEchoServer` example now uses `AddressedEnvelope<ByteBuffer>` in its `EchoHandler` as `InboundIn` and `OutboundOut` type.

### Result:

Newbies are less confused. 🤩

### DIFF
--- a/Sources/NIOUDPEchoServer/main.swift
+++ b/Sources/NIOUDPEchoServer/main.swift
@@ -14,8 +14,8 @@
 import NIO
 
 private final class EchoHandler: ChannelInboundHandler {
-    public typealias InboundIn = ByteBuffer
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Love it, great fix!

Lukasa:

PR #1668, opened 2020-10-05
NIO: silence error on Windows

compnerd:
The `ignoreSIGPIPE` does nothing on Windows.  However, we must truncate
the socket descriptor as the SIGPIPE takes a FD which is 32-bits but the
socket handle is 64-bits.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -379,7 +379,7 @@ class BaseSocket: BaseSocketProtocol {
     }
 
     func ignoreSIGPIPE() throws {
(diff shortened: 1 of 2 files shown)


Lukasa (commented):
Would it be better to define `ignoreSIGPIPE` in terms of `NIOBSDSocket.Handle`?

compnerd:
Well, really you would change the signature to indicate what you are ignoring the SIGPIPE on: a socket or a file descriptor.  Would you prefer a changed API?

Lukasa:
I think I would.

Lukasa (approved):

PR #1669, opened 2020-10-05
NIO: carve away more of `System.swift` on Windows

compnerd:
`System` interfaces should be considered deprecated and not be used.
This removes more of the interfaces from Windows as the socket
interfaces are now part of BSDSocketAPI.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/System.swift
+++ b/Sources/NIO/System.swift
@@ -49,6 +49,7 @@ extension ipv6_mreq { // http://lkml.iu.edu/hypermail/linux/kernel/0106.1/0080.h
 #endif
 
 // Declare aliases to share more code and not need to repeat #if #else blocks
+#if !os(Windows)
 private let sysClose = close
 private let sysShutdown = shutdown
 private let sysBind = bind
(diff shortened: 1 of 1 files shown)


Sources/NIO/System.swift
```diff
@@ -334,7 +335,7 @@ internal enum Posix {
     }
 
     @inline(never)
-    public static func open(file: UnsafePointer<CChar>, oFlag: CInt, mode: mode_t) throws -> CInt {
+    public static func open(file: UnsafePointer<CChar>, oFlag: CInt, mode: CInt) throws -> CInt {
```
Lukasa, line 338: This change is wrong: `mode_t` exists on the Posix platforms and is defined as `UInt32`, at least on Linux. It'd be better for this file to define `mode_t` for Windows as the right thing.

compnerd: Sure, but that wasn't meant to be part of this change, I want to try to do the `mode_t` changes together as this isn't the only site.

Lukasa (approved):

PR #1672, opened 2020-10-12
Fix incorrect new names for deprecated APIs

glbrntt:
Motivation:

Some of the deprecated APIs on the client, server and datagram
bootstraps included the type name in the "rename" field. Since the
underlying type hasn't changed, this isn't quite right results in the
typename being erroneously included when applying a fix-it in Xcode.

Modifications:

- Remove unnecessary typename from a few rename fields.

Result:

Fix-its work as expected.


PeterAdams-A (approved):
looks good

PR #1674, opened 2020-10-13
NIO: implement `sendmsg` and `recvmsg` on Windows

compnerd:
`WSASendMsg` and `WSARecvMsg` are not directly accessible to use.
Instead, one must perform an IOCTL on the socket to retrieve the
extension point and then use that function pointer to perform the
operation.  Use this to implement the functionality on Windows.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/BSDSocketAPIWindows.swift
+++ b/Sources/NIO/BSDSocketAPIWindows.swift
@@ -88,6 +88,7 @@ import func WinSDK.ReadFile
 import func WinSDK.TransmitFile
 import func WinSDK.WriteFile
 import func WinSDK.WSAGetLastError
+import func WinSDK.WSAIoctl
 
 import struct WinSDK.socklen_t
 import struct WinSDK.u_long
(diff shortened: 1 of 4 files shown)


Sources/NIO/BSDSocketAPIWindows.swift
```diff
@@ -210,15 +247,62 @@ extension NIOBSDSocket {
...
+                           &pfnWSARecvMsg,
+                           DWORD(MemoryLayout.stride(ofValue: pfnWSARecvMsg)),
+                           &cbBytesReturned, nil, nil) == SOCKET_ERROR {
+            throw IOError(winsock: WSAGetLastError(), reason: "WSAIoctl")
+        }
+
+        guard let pfnWSARecvMsg = pfnWSARecvMsg else {
+            throw IOError(windows: DWORD(ERROR_INVALID_FUNCTION),
+                          reason: "recvmsg")
+        }
```
Lukasa, line 268: So let's state up-front that this code is entirely insane. Worse, it seems to be correct, and I think having insane and correct code is the worst of all worlds.

As we noted in our Slack chat, this is a performance nightmare. Ideally we'd cache this function pointer, if we can work out what its validity period is, and we'd also deduplicate the code from the sendmsg implementation. However, for now I'm ok keeping it here, noting that it limits performance, because "working but slow" is better than "not working". Anything can be fast if it doesn't work.

That said, can we have a comment at the top of this that reminds us to find a way to dedupe this and cache the function pointer?

Sources/NIO/Socket.swift
```diff
@@ -174,14 +174,27 @@ typealias IOVector = iovec
 
         return try withUnsafeHandle { handle in
             return try withUnsafeMutablePointer(to: &vec) { vecPtr in
+#if os(Windows)
+                var messageHeader =
+                    WSAMSG(name: notConstCorrectDestinationPtr
+                                    .assumingMemoryBound(to: SOCKADDR.self),
```
Lukasa, line 180: For my own OCD, can we `.assumingMemoryBound(to: sockaddr.self)`? I'm pretty sure they typealias to the same thing, but this way I can see by inspection that the assumption is right.

Sources/NIO/Socket.swift
```diff
@@ -174,14 +174,27 @@ typealias IOVector = iovec
...
+                           namelen: destinationSize,
+                           lpBuffers: vecPtr,
+                           dwBufferCount: 1,
+                           Control: WSABUF(len: ULONG(controlBytes.count),
+                                           buf: controlBytes.baseAddress?
+                                                    .assumingMemoryBound(to: CHAR.self)),
```
Lukasa, line 186: This `assumingMemoryBound` is a bit rough. It's technically a violation of Swift's pointer aliasing model, though it's ok in C.

Normally the way we'd address this is to thunk through CNIOWindows by having it declare the API with the types we want and just doing the cast that is unsafe in Swift. However, in this case I think it's acceptable for us to bind the memory: the rest of the program assumes it's untyped, so we won't break the aliasing model. Can we replace the `assumingMemoryBound` here and below with `.bindMemory(to:)`?

Sources/NIO/Socket.swift
```diff
@@ -216,6 +230,19 @@ typealias IOVector = iovec
 
         return try withUnsafeMutablePointer(to: &vec) { vecPtr in
             return try storage.withMutableSockAddr { (sockaddrPtr, _) in
+#if os(Windows)
+                var messageHeader =
+                    WSAMSG(name: sockaddrPtr, namelen: storageLen,
+                           lpBuffers: vecPtr, dwBufferCount: 1,
+                           Control: WSABUF(len: ULONG(controlBytes.controlBytesBuffer.count),
+                                           buf: controlBytes.controlBytesBuffer.baseAddress?
+                                                    .assumingMemoryBound(to: CHAR.self)),
```
Lukasa, line 239: I think this needs to be a `bindMemory` as well.

Lukasa (approved):

PR #1679, opened 2020-10-19
Add watchOS deployment to PodSpec build script

glbrntt:
Motivation:

We support watchOS 6+ with SwiftNIO Transport Services; as such we should
include watchOS as a deployment target for our CocoaPods.

Modifications:

- Add a watchOS deployment target to `build_podspecs.sh`
- Update docs

Result:

Users can deploy to watchOS 6+ with CocoaPods.


glbrntt:
@Lukasa we actually support below watchOS 6 but not usefully since we can't do any networking. Are we okay with setting it as 6?

PeterAdams-A (approved):
Looks good.

PR #1683, opened 2020-10-23
Add multiple channel handlers in a single async call.

PeterAdams-A:
Motivation:

Curently adding multiple channel handlers makes a call to the
async version of addHandler for each handler resulting in
n+1 futures.  It feels better to use just one future and add
all the handlers synchoronously.

Modifications:

Change sync functions with take a promise to instead return a Result.
Feed this back until reaching addHandlers.

Result:

Multiple handlers can be added more quickly.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -909,6 +909,8 @@ extension ChannelPipeline {
             individualPosition = .after(handler)
         }
 
+        let promise = self.eventLoop.makePromise(of: Void.self)
+
         // Add all the handlers.
         func addAllHandlersAndComplete() {
             // Function for reducing to propagate errors.
@@ -927,7 +929,6 @@ extension ChannelPipeline {
             promise.completeWith(result)
         }
 
-        let promise = self.eventLoop.makePromise(of: Void.self)
         if self.eventLoop.inEventLoop {
             addAllHandlersAndComplete()
         } else {
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -192,8 +192,8 @@ public final class ChannelPipeline: ChannelInvoker {
     ///     - position: The position in the `ChannelPipeline` to add `handler`. Defaults to `.last`.
     /// - returns: the result of adding this handler - either success or failure with an error code if this could not be completed.
     private func _add(_ handler: ChannelHandler,
-              name: String? = nil,
-              position: ChannelPipeline.Position = .last) -> Result<Void, Error> {
+                      name: String? = nil,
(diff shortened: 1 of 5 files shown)


PeterAdams-A:
Code looks a bit worse but I think it should run better.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -907,9 +907,37 @@ extension ChannelPipeline {
...
+                case .success:
+                    return current
+                case .failure:
+                    return initial
+                }
+            }
 
+            let result = handlers.map {
+                _add($0, position: individualPosition)
+            }.reduce(.success(()), firstErrorOrSuccess)
```
Lukasa, line 928: Let’s avoid map and reduce here and just use a loop, we don’t need the intermediate arrays and if we’re aiming to speed things up anyway it’s another incremental savings.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -907,9 +907,37 @@ extension ChannelPipeline {
...
+                switch initial {
+                case .success:
+                    return current
+                case .failure:
+                    return initial
+                }
+            }
 
+            let result = handlers.map {
+                _add($0, position: individualPosition)
```
Lukasa, line 927: I think this needs explicit self.

Lukasa (changes requested):
I’m broadly happy with this! I have a couple of basic notes but otherwise this seems like a good change.

Lukasa:
Do we have allocation counter tests that cover this?

Lukasa:
Never mind, looking at the build failures tells me we do, nice work Peter!

Sources/NIO/ChannelPipeline.swift
```diff
@@ -171,52 +171,59 @@ public final class ChannelPipeline: ChannelInvoker {
...
+    /// May only be called from on the event loop.
+    ///
+    /// - parameters:
+    ///     - name: the name to use for the `ChannelHandler` when its added. If none is specified it will generate a name.
+    ///     - handler: the `ChannelHandler` to add
+    ///     - position: The position in the `ChannelPipeline` to add `handler`. Defaults to `.last`.
+    /// - returns: the result of adding this handler - either success or failure with an error code if this could not be completed.
+    private func _add(_ handler: ChannelHandler,
+              name: String? = nil,
+              position: ChannelPipeline.Position = .last) -> Result<Void, Error> {
```
glbrntt, line 196: Indentation is off here (and for all the `add0` calls in the body too)

glbrntt (changes requested):
Nice one! Indentation is off in a handful of places but looks great otherwise.

Lukasa (approved):
Once @glbrntt's concerns are addressed I'm happy.

glbrntt (approved):

PR #1684, opened 2020-10-28
NIO: implement address resolution on Windows

compnerd:
Windows has a slightly different API for address resolution.  Implement
the Windows path for addrinfo resolution.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/GetaddrinfoResolver.swift
+++ b/Sources/NIO/GetaddrinfoResolver.swift
@@ -23,6 +23,21 @@
 /// This resolver is a single-use object: it can only be used to perform a single host resolution.
 import CNIOLinux
 
+#if os(Windows)
+import let WinSDK.AF_INET
+import let WinSDK.AF_INET6
+
+import func WinSDK.FreeAddrInfoW
+import func WinSDK.GetAddrInfoW
+import func WinSDK.gai_strerrorA
+
+import struct WinSDK.ADDRESS_FAMILY
+import struct WinSDK.ADDRINFOW
+import struct WinSDK.SOCKADDR_IN
+import struct WinSDK.SOCKADDR_IN6
+#endif
+
+
 internal class GetaddrinfoResolver: Resolver {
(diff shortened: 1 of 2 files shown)


Sources/NIO/GetaddrinfoResolver.swift
```diff
@@ -102,13 +141,44 @@ internal class GetaddrinfoResolver: Resolver {
...
     /// - parameters:
     ///     - info: The pointer to the first of the `addrinfo` structures in the list.
     ///     - host: The hostname we resolved.
+#if os(Windows)
+    private func parseResults(_ pResult: UnsafeMutablePointer<ADDRINFOW>, host: String) {
+        var v4: [SocketAddress] = []
+        var v6: [SocketAddress] = []
+
+        var pResult: UnsafeMutablePointer<ADDRINFOW>? = pResult
+        while pResult != nil {
```
Lukasa, line 158: We save ourselves some force unwraps by using `while let` instead of `while != nil`. Even better we could follow the looping logic below which uses `while true` and explicitly breaks the loop if it cannot assign a non-nil pointer.

compnerd: Thanks for pointing that out.  No offense, but I kinda disagree, after looking at it when you pointed it out, I think that this entire change is just rubbish.  There is no need for the difference here - this is a private function, an internal typealias is the right thing to do here.  No one cares about internal typealiases - its the public surface that matters.  By doing this this change becomes a mere 5 lines.

Sorry.

Lukasa: Do we want to do the same for the other lookup function here?

Lukasa (approved):

Sources/NIO/GetaddrinfoResolver.swift
compnerd: Discussed offline, although for the core body we could use a `let` binding to deal with the differences in functions, the string encoding I feel like would detract from readability as the string re-encoding can be confusing.

PR #1686, opened 2020-11-02
NIO: initial pass over iov handling for Windows

compnerd:
This partially cleans up the iovec based message handling.  Create a new
typealias for `IOVector` which changes the type for the underlying
implementation's scatter/gather IO type.  This only covers the cases
where there is an implemented equivalent.  There is still work to be
done to cover some of the other cases.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/PendingDatagramWritesManager.swift
+++ b/Sources/NIO/PendingDatagramWritesManager.swift
@@ -13,6 +13,18 @@
 //===----------------------------------------------------------------------===//
 import NIOConcurrencyHelpers
 
+#if os(Windows)
+import let WinSDK.EHOSTUNREACH
+import let WinSDK.EMSGSIZE
+
+import struct WinSDK.CHAR
+import struct WinSDK.SOCKADDR
+import struct WinSDK.ULONG
+import struct WinSDK.WSABUF
+import struct WinSDK.WSAMSG
+import struct WinSDK.socklen_t
+#endif
+
 private struct PendingDatagramWrite {
     var data: ByteBuffer
     var promise: Optional<EventLoopPromise<Void>>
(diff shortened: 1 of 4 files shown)


Davidde94 (approved):
LGTM

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -101,19 +113,41 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
...
-            iovecs[c] = iovec(iov_base: UnsafeMutableRawPointer(mutating: ptr.baseAddress!), iov_len: numericCast(toWriteForThisBuffer))
+
+            let buffer: UnsafeMutablePointer<IOVector> = iovecs.baseAddress! + c
+            let address: UnsafeMutablePointer<sockaddr_storage> =
+                addresses.baseAddress! + c
+
+#if os(Windows)
+            iovecs[c] =
+                IOVector(len: numericCast(toWriteForThisBuffer),
+                         buf: UnsafeMutableRawPointer(mutating: ptr.baseAddress!).assumingMemoryBound(to: CHAR.self))
```
Lukasa, line 124: We can't make this assumption either, we have to bind it.

Sources/NIO/PendingDatagramWritesManager.swift
```diff
@@ -101,19 +113,41 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
...
+            iovecs[c] = IOVector(iov_base: UnsafeMutableRawPointer(mutating: ptr.baseAddress!), iov_len: numericCast(toWriteForThisBuffer))
+#endif
 
             var controlBytes = UnsafeOutboundControlBytes(controlBytes: controlMessageStorage[c])
             controlBytes.appendExplicitCongestionState(metadata: p.metadata, protocolFamily: p.address.protocol)
             let controlMessageBytePointer = controlBytes.validControlBytes
 
-            let msg = msghdr(msg_name: addresses.baseAddress! + c,
+#if os(Windows)
+            let msg = address.withMemoryRebound(to: SOCKADDR.self, capacity: 1) {
```
Lukasa, line 134: This `withMemoryRebound` dance, both here and above, is not right. We are escaping the rebound pointer, which violates Swift's pointer aliasing rules. Is `SOCKADDR` a meaningfully different type, or is it the same as `sockaddr_storage` in all meaningful ways?

Sources/NIO/PendingWritesManager.swift
```diff
@@ -52,7 +56,12 @@ private func doPendingWriteVectorOperation(pending: PendingStreamWritesState,
 
             buffer.withUnsafeReadableBytesWithStorageManagement { ptr, storageRef in
                 storageRefs[i] = storageRef.retain()
+#if os(Windows)
+                iovecs[i] = IOVector(len: numericCast(toWriteForThisBuffer),
+                                     buf: UnsafeMutableRawPointer(mutating: ptr.baseAddress!) .assumingMemoryBound(to: CHAR.self))
```
Lukasa, line 61: We can't make this assumption, we have to bind this memory.

Sources/NIO/Socket.swift
```diff
@@ -169,7 +186,11 @@ typealias IOVector = iovec
                  controlBytes: UnsafeMutableRawBufferPointer) throws -> IOResult<Int> {
         // Dubious const casts - it should be OK as there is no reason why this should get mutated
         // just bad const declaration below us.
-        var vec = iovec(iov_base: UnsafeMutableRawPointer(mutating: pointer.baseAddress!), iov_len: pointer.count)
+#if os(Windows)
+        var vec = IOVector(len: ULONG(pointer.count), buf: UnsafeMutablePointer<CHAR>(mutating: pointer.baseAddress!.assumingMemoryBound(to: CHAR.self)))
```
Lukasa, line 190: Ok, there are a lot of assumptions about the memory binding here. Is there any reason not to just change the type of the base pointer here on Windows?

Sources/NIO/PendingDatagramWritesManager.swift
compnerd: They are different.  The `SOCKADDR_STORAGE` is opaque content, and will need to be cast to the `SOCKADDR`.  From the documentation:

> A WSK application typically does not directly access any of the members of the SOCKADDR_STORAGE structure except for the ss_family member. Instead, a pointer to a SOCKADDR_STORAGE structure is normally cast to a pointer to the specific SOCKADDR structure type that corresponds to a particular address family.

Sources/NIO/Socket.swift
compnerd: I was trying to avoid two entire paths, but we can do that.

Lukasa: Do we get two entire paths? I think we can typealias the actual type and use that type everywhere.

compnerd: Ah, let me try that approach - using a typealias instead would certainly be better.

Sources/NIO/PendingDatagramWritesManager.swift
Lukasa: Ugh this really awkward, this is extremely hard to fix. I think we need to write a thunk version of the function we want to call that takes a structure that isn't `WSAMSG` but is _basically_ a `WSAMSG` that takes a `SOCKADDR_STORAGE` and not a `SOCKADDR`. Swift pointer aliasing forbids us from escaping the temporarily-rebound pointer to the `SOCKADDR`, so we have to rebind at a higher scope around this whole function which is also extremely difficult to do.

Thunking through C is gonna be necessary.

Sources/NIO/Socket.swift
compnerd: I opted to not do this since re-working this made this the same as every other instance.

Sources/NIO/PendingDatagramWritesManager.swift
compnerd: I don't understand what you mean by a thunk here that takes that - this is being used to construct an array of elements of these types, I don't see how to thunk the function call without sinking all of this into C.

Lukasa: I think we only need to thunk the operation at the bottom of the call stack (into libc) to take the function type we want, no? Then the changes will ripple up through Swift and allow us to remove the rebind here.

Lukasa: You are missing something. :smile:

We cannot construct a `WSAMSG` here, in this way, because to do so requires us to escape a rebound pointer and violate aliasing rules, which we may not do.

_However_, `MMsgHdr` on Windows is a type _we_ defined, and so _we_ control it. We can therefore redefine it to store, instead of a `WSAMSG`, a type almost exactly the same as `WSAMSG` but with the socket address pointer type we want to use from Swift. The C code can then cast the buffer to the appropriate type internally, where this cast is "safe".

PR #1687, opened 2020-11-11
Removes misleading docs for EventLoopFuture.whenComplete(_:)

jemmons:
Removes documentation from `EventLoopFuture.whenComplete(_:)` that is no longer accurate.

### Motivation:

Upon the addition of `Result` in the Swift standard library, #734
updated `EventLoopFuture.whenComplete(_:)` to pass a `Result<T, Error>` to
its `callback`, but the documentation still confusingly states:

> Unlike its friends… `whenComplete` does not receive the result of the
> `EventLoopFuture`.

It would appear the documentation and functionality are no longer in sync and 
could thus cause confusion.

### Modifications:

This patch removes the (now inaccurate) lines from the documentation.

### Result:

The documentation now correctly reflects the functionality of 
`EventLoopFuture.whenComplete(_:)`.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -727,10 +727,6 @@ extension EventLoopFuture {
     /// Adds an observer callback to this `EventLoopFuture` that is called when the
     /// `EventLoopFuture` has any result.
     ///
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Great catch, that comment is way out of date

Lukasa:

PR #1691, opened 2020-11-16
Conform TimeAmount to AdditiveArithmetic

### LINKED ISSUE #1690: TimeAmount does not support +=, -=, etc...
### Expected behavior
Sometimes it is useful to manipulate time amounts when building up a delay for `.scheduleTask()` for instance, and if that iteration fails, we would want to `delay += .milliseconds(5)` to add 5 milliseconds to our delay and try again.

### Actual behavior
You would get a lovely `Referencing operator function '+=' on 'FloatingPoint' requires that 'TimeAmount' conform to 'FloatingPoint'` compile time error if such a statement were attempted.

In general, this is not strictly _necessary_ (I can get by with my local extension, for instance), but is certainly a good quality of life improvement to save a future developer's marbles when they don't realize `+=` conformance from `+` isn't automatic 🙃.

I'm mostly writing this to field input to see if such a PR would be welcome or not (besides, makes for a great first issue for newcomers 😉)

```
var retryDelay = TimeAmount.milliseconds(0) // <-- A `.zero` here would also be stellar,
                                        // but maybe conforming to `AdditiveArithmetic` would be even better?

// (in your favorite accent): One Iteration Later...
retryDelay += .milliseconds(5) // Compiler error 😔
retryDelay = retryDelay + .milliseconds(5) // Works, but 🙃
```

2.23.0

```
Apple Swift version 5.3.1 (swiftlang-1200.0.41 clang-1200.0.32.8)
Target: x86_64-apple-darwin19.6.0
Darwin Dimitri-Bouniols-16-inch-MacBook-Pro.local 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64
```

joshrutkowski:
Conforms TimeAmount to AdditiveArithmetic, resolves #1690 

Motivation:
TimeAmount does not support -=, +=. Sometimes it is useful to manipulate time amounts when building up a delay and if that iteration fails, we would want to delay += .milliseconds(5) to add 5 milliseconds to our delay and try again.

Modifications:
Conformed TimeAmount to AdditiveArithmetic: added a static zero property and required operators.

Result:
TimeAmount conforms to AdditiveArithmetic.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -354,15 +354,15 @@ extension TimeAmount: AdditiveArithmetic {
     public static var zero: TimeAmount {
         return TimeAmount.nanoseconds(0)
     }
-    
+
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
-    
+
     public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
          return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
     }
-    
+
     public static func * <T: BinaryInteger>(lhs: T, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(Int64(lhs) * rhs.nanoseconds)
     }
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -358,9 +358,17 @@ extension TimeAmount: AdditiveArithmetic {
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
+    
+    public static func +=(lhs: inout TimeAmount, rhs: TimeAmount){
+        lhs = lhs + rhs
+    }
 
     public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
          return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
+    }
+    
+    public static func -=(lhs: inout TimeAmount, rhs: TimeAmount){
+        lhs = lhs - rhs
     }
 
     public static func * <T: BinaryInteger>(lhs: T, rhs: TimeAmount) -> TimeAmount {
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -277,9 +277,15 @@ extension EventLoopGroup {
 /// Represents a time _interval_.
 ///
 /// - note: `TimeAmount` should not be used to represent a point in time.
-public struct TimeAmount: Hashable {
+public struct TimeAmount: Hashable, AdditiveArithmetic {
+    
     @available(*, deprecated, message: "This typealias doesn't serve any purpose. Please use Int64 directly.")
     public typealias Value = Int64
+    
+    /// The zero value for `TimeAmount`.
+    public static var zero: TimeAmount {
+        return TimeAmount.nanoseconds(0)
+    }
 
     /// The nanoseconds representation of the `TimeAmount`.
     public let nanoseconds: Int64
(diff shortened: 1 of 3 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -353,10 +359,18 @@ extension TimeAmount {
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
+    
+    public static func += (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
```
Lukasa, line 363: This function signature doesn’t seem to match what `AdditiveArithmetic` wants: that takes the argument _inout_.

Sources/NIO/EventLoop.swift
```diff
@@ -353,10 +359,18 @@ extension TimeAmount {
...
+        return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
+    }
 
     public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
     }
+    
+    public static func -= (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
```
Lukasa, line 371: Same here.

Lukasa (changes requested):
Thanks for this! You’ve added a pair of functions that don’t exist on additive arithmetic and that behave a bit unlike the usual behaviour of `+=` and `-=`. I think it’d probably be best to remove those two.

Sources/NIO/EventLoop.swift
joshrutkowski: Added the _inout_ argument

Sources/NIO/EventLoop.swift
joshrutkowski: Added the _inout_ argument

joshrutkowski:
> Thanks for this! You’ve added a pair of functions that don’t exist on additive arithmetic and that behave a bit unlike the usual behaviour of `+=` and `-=`. I think it’d probably be best to remove those two.

Removed the functions as suggested to preserve usual behavior of `+=` and `-=`.

Sources/NIO/EventLoop.swift
```diff
@@ -349,15 +350,20 @@ extension TimeAmount: Comparable {
...
+    }
+    
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
-
+    
     public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
-        return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
+         TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
```
Lukasa, line 364: This change won’t compile in Swift 5.0, which we’re still supporting.

joshrutkowski: Fixed. Thank you for the patience.

Sources/NIO/EventLoop.swift
```diff
@@ -353,10 +359,18 @@ extension TimeAmount {
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
+    
+    public static func += (lhs: inout TimeAmount, rhs: TimeAmount) -> TimeAmount {
```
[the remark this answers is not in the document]
joshrutkowski: Would this function be required, given that we are conforming to AdditiveArithmetic? Removed in later commit, and passes with test:

```
func testTimeAmountDoesAddTime() {
        var lhs = TimeAmount.milliseconds(0)
        lhs += .milliseconds(5)
        XCTAssertEqual(lhs, .milliseconds(5))
    }
```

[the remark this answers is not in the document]
Lukasa: You are correct that this function isn’t required at all.

Lukasa (approved):
Cool, this patch looks reasonable to me!

Lukasa:

Lukasa:
Sadly it looks like the two functions do need to be implemented in older Swift versions to get the conformance. We need `static func +=(lhs: inout TimeAmount, rhs: TimeAmount)` and `static func -=(lhs: inout TimeAmount, rhs: TimeAmount)`.

Lukasa:

Lukasa (approved):
Nice patch!

Lukasa:

Issue #1648, opened 2020-09-24
How to read SocketAddress from ByteBuffer?


purkylin:
I have a ChannelInboundHandler, and want to read ip address from the ByteBuffer, can someone give me some code conduct?

Lukasa:
Can the IP address be IPv4 _or_ IPv6, or must it be one or the other? What format is it in: packed binary, dotted-decimal, colon-separated IPv6?

purkylin:
@Lukasa 
+------+----------+----------+
| ATYP | BND.ADDR | BND.PORT |
+------+----------+----------+
|  1   | Variable |    2     |
+------+----------+----------+

ATYP
1: IPV4, follow 4 bytes
4: IPV6, follow 16 bytes

Lukasa:
Ok, those are packed representations. I am assuming network byte order here.

What data type would you like to end up having? `SocketAddress`?

purkylin:
Yes, I want SocketAddress type

Lukasa:
I haven't tested it, but this code should work:

```swift
import NIO

enum PackedSocketAddressError: Error {
    case invalidTypeByte
}

extension ByteBuffer {
    mutating func readPackedRepresentation() throws -> SocketAddress? {
        guard let type = self.readInteger(as: UInt8.self) else {
            return nil
        }

        switch type {
        case 1:
            guard let packed = self.readInteger(as: UInt32.self),
                  let port = self.readInteger(as: UInt16.self) else {
                return nil
            }
            var addr = sockaddr_in()
            addr.sin_addr = in_addr(s_addr: packed)
            addr.sin_family = sa_family_t(AF_INET)
            addr.sin_port = port.littleEndian
            return SocketAddress(addr, host: "")
        case 4:
            guard let packed = self.readSlice(length: 16)?.readableBytesView,
                  let port = self.readInteger(as: UInt16.self) else {
                return nil
            }
            var addr = sockaddr_in6()
            addr.sin6_family = sa_family_t(AF_INET6)
            addr.sin6_port = port.littleEndian
            withUnsafeMutableBytes(of: &addr.sin6_addr) { $0.copyBytes(from: packed) }
            return SocketAddress(addr, host: "")
        default:
            throw PackedSocketAddressError.invalidTypeByte
        }
    }
}
```

PR #1692, opened 2020-11-18
Create SocketAddress from packed byte representation

### LINKED ISSUE #1649: Provide helper to create SocketAddress from packed byte representation
We have helpers to create `SocketAddress` objects from string representations of IP addresses, but it would be good if we had helpers to create them from packed bytes as well. An example algorithm that we could repurpose is in #1648.

ayshiff:
Closes #1649

### Motivation:

A user should be able to create SocketAddress from packed bytes representation.

### Modifications:

I added a new SocketAddress initializer which takes the IP address in ByteBuffer form. 
I have also added tests that ensure the initializer works properly.

ByteBuffer inputs should be in the [SOCKS](https://tools.ietf.org/html/rfc1928) representation:

| ATYP | DST.ADDR | DST.PORT |
|------|----------|----------|
| 1    | Variable | 2        |

As mentioned in #1648.

**ATYP** address type of following address:
- IP V4 address: X'01' -> Implemented
- IP V6 address: X'04' -> Implemented
- DOMAINNAME: X'03' -> Not implemented

**Address field** of following length:
- IP V4 address: 4 bytes
- IP V6 address: 16 bytes 
- DOMAINNAME: N Bytes

Implemented address types:
- [X] IP V4 address
- [X] IP V6 address
- [ ] DOMAINNAME

I don't know if I followed the correct method and if there are scenarios that I haven't covered (for example if we want to initialize the IP address as a packed byte representation but the port as an Integer).

I don't have much knowledge on the subject and I would like to have your opinion on this PR ! 

### Result:

We have a new way to initialize a SocketAddress from a byteBuffer.


Lukasa (changes requested):
Thanks for this patch!

As a quick note, we don't want to construct this from a packed SOCKS-formatted data structure, we'd like to construct it from just the raw address bytes themselves. The SOCKS formatting is necessary for the other use-case, but is not necessary for NIO more generically.

Sources/NIO/SocketAddresses.swift
```diff
@@ -325,6 +327,34 @@ public enum SocketAddress: CustomStringConvertible {
...
         }
     }
+    
+    /// Create a new `SocketAddress` for an IP address in ByteBuffer form.
+    ///
+    /// - parameters:
+    ///     - ipAddress: The IP address, in ByteBuffer form.
+    ///     - port: The target port.
+    /// - returns: the `SocketAddress` corresponding to this string and port combination.
+    /// - throws: may throw `SocketAddressError.failedToParseIPByteBuffer` if the IP address cannot be parsed or `SocketAddressError.unsupported` if the ATYP is wrong.
```
Lukasa, line 337: We aren't using an `ATYP` here and can never throw `unsupported`. Can we update this throws message?

Sources/NIO/SocketAddresses.swift
```diff
@@ -325,6 +327,34 @@ public enum SocketAddress: CustomStringConvertible {
...
+    public init(ipAddress: ByteBuffer, port: Int) throws {
```
Lukasa, line 338: One minor nit: can we change the argument to `packedIpAddress`? This will give us a slightly more descriptive name, and it'll avoid a type-based overload and replace it with a label-based overload.

Lukasa (commented):
This looks really good! I have two minor notes and then I think this should be good to go.

Please also run the `scripts/generate_linux_tests.rb` script to add the extra hooks for testing on Linux.

Sources/NIO/SocketAddresses.swift
```diff
@@ -38,6 +38,8 @@ public enum SocketAddressError: Error {
     case unixDomainSocketPathTooLong
     /// Unable to parse a given IP string
     case failedToParseIPString(String)
+    /// Unable to parse a given IP ByteBuffer
+    case failedToParseIPByteBuffer(ByteBuffer)
```
Lukasa, line 42: I'm afraid we can't add new cases to enumerations: doing so is a Semver major change. Can we add a new `struct` to represent this error instead?

ayshiff: I'm having trouble implementing the `struct`. 
Could you enlighten me on this point ?

This is what i did :

```swift
public enum SocketAddressError: Error {
    /// The host is unknown (could not be resolved).
    case unknown(host: String, port: Int)
    /// The requested `SocketAddress` is not supported.
    case unsupported
    /// The requested UDS path is too long.
    case unixDomainSocketPathTooLong
    /// Unable to parse a given IP string
    case failedToParseIPString(String)
    /// Unable to parse a given IP ByteBuffer
    struct failedToParseIPByteBuffer: Error {
      let address: ByteBuffer
    }
}
// ...
throw SocketAddressError.failedToParseIPByteBuffer(address: packedIpAddress)
```

Lukasa:
```swift
extension SocketAddressError {
    public struct FailedToParseIPByteBuffer: Error, Hashable {
        public var address: ByteBuffer

        public init(address: ByteBuffer) {
            self.address = address
        }
    }
}

throw SocketAddressError.FailedToParseIPByteBuffer(address: packedIpAddress)
```

ayshiff:
@Lukasa Does the code look good to you or do you want me to make other changes ?

Lukasa:

Lukasa (approved):
Great, this looks really good to me. Well done @ayshiff!

PR #1695, opened 2020-11-25
Update and finish Android port

buttaface:
Motivation:

Be able to run and test swift-nio on Android.

Modifications:

- Remove the custom ifaddrs and use the one from the Android NDK instead.
- Enable a bunch of conditionally-compiled code for Android.
- Add a handful of constants and other Android declarations.
- Cast some types because of mismatches specific to Android.

Result:

Most tests pass on Android AArch64 and ARMv7.

This pull was [developed natively on Android AArch64 using the Termux app](https://forums.swift.org/t/swift-for-android-call-for-the-community/32766/13) and then [cross-compiled using all three Android SDKs for Swift](https://github.com/buttaface/swift-android-sdk), which turned up the `arm`-specific parts of this pull. One testcase, `BootstrapTest/testClientBindWorksOnSocketsBoundToEitherIPv4OrIPv6Only`, still fails on both AArch64 and AArch32, looks like a problem with Happy Eyeballs:
```
/home/butta/swift-nio/Tests/NIOTests/BootstrapTest.swift:619: error: BootstrapTest.testClientBindWorksOnSocketsBoundToEitherIPv4OrIPv6Only : XCTAssertNoThrow failed: threw error "NIOConnectionError(host: "localhost", port: 37087, dnsAError: nil, dnsAAAAError: nil, connectionErrors: [NIO.SingleConnectionFailure(target: [IPv4]localhost/127.0.0.1:37087, error: bind(descriptor:ptr:bytes:): Address family not supported by protocol (errno: 97))])" -
```
Four more testcases fail only on AArch32: `NIOTests.SystemTest/testCMsg{Collection,Data,NextHeader}` fail because they don't seem to use the right test data for 32-bit platforms, whereas the second assert in `NIOHTTP1Tests.HTTPHeadersTest/testCapacity` returns `6` instead of `5`, probably some oddity of 32-bit arrays?

x86_64 has been built but not tested: I'll look into running those tests when I set up my Android SDK CI eventually.


Sources/NIO/Interfaces.swift
```diff
@@ -462,6 +462,11 @@ public final class NIONetworkInterface {
 #else
     internal init?(_ caddr: ifaddrs) {
         self.name = String(cString: caddr.ifa_name)
+
+        guard caddr.ifa_addr != nil else {
+            return nil
```
buttaface, line 467: Android devices have a bunch of interfaces, many without an address, which was breaking the next `guard` when this is called from `enumerateInterfaces()`. Since that function is deprecated anyway, could just ignore it and remove this change.

Sources/NIO/Linux.swift
```diff
@@ -71,12 +71,12 @@ internal enum Epoll {
...
-    public static let EPOLLERR: CUnsignedInt = numericCast(CNIOLinux.EPOLLERR)
-    public static let EPOLLRDHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLRDHUP)
-    public static let EPOLLHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLHUP)
-    public static let EPOLLET: CUnsignedInt = numericCast(CNIOLinux.EPOLLET)
+    public static let EPOLLIN: CUnsignedInt = 1 //numericCast(CNIOLinux.EPOLLIN)
+    public static let EPOLLOUT: CUnsignedInt = 4 //numericCast(CNIOLinux.EPOLLOUT)
+    public static let EPOLLERR: CUnsignedInt = 8 // numericCast(CNIOLinux.EPOLLERR)
+    public static let EPOLLRDHUP: CUnsignedInt = 8192 //numericCast(CNIOLinux.EPOLLRDHUP)
+    public static let EPOLLHUP: CUnsignedInt = 16 //numericCast(CNIOLinux.EPOLLHUP)
+    public static let EPOLLET: CUnsignedInt = 2147483648 //numericCast(CNIOLinux.EPOLLET)
```
Lukasa, line 79: I'm very nervous about removing these. Can we define the appropriate values in `CNIOLinux` conditionally on Android instead?

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -100,7 +100,11 @@ struct DatagramVectorReadManager {
                 // TODO(cory): almost all of this except for the iovec could be done at allocation time. Maybe we should?
 
                 // First we set up the iovec and save it off.
+                #if os(Android)
+                self.ioVector[i] = IOVector(iov_base: bufferPointer.baseAddress! + (i * messageSize), iov_len: __kernel_size_t(messageSize))
```
Lukasa, line 104: Rather than add this conditional code block, can we define an extension on the android iovec type that accepts the appropriate types in its initializer and does the conversion there? It'll clean up the code a lot.

Lukasa (commented):
Thanks for this patch! I think it broadly looks reasonable, but I've left a couple of notes in the diff.

Sources/NIO/Linux.swift
```diff
@@ -71,12 +71,12 @@ internal enum Epoll {
     public static let EPOLL_CTL_DEL: CInt = numericCast(CNIOLinux.EPOLL_CTL_DEL)
 
     #if os(Android)
-    public static let EPOLLIN: CUnsignedInt = numericCast(CNIOLinux.EPOLLIN)
-    public static let EPOLLOUT: CUnsignedInt = numericCast(CNIOLinux.EPOLLOUT)
-    public static let EPOLLERR: CUnsignedInt = numericCast(CNIOLinux.EPOLLERR)
-    public static let EPOLLRDHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLRDHUP)
-    public static let EPOLLHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLHUP)
-    public static let EPOLLET: CUnsignedInt = numericCast(CNIOLinux.EPOLLET)
+    public static let EPOLLIN: CUnsignedInt = 1 //numericCast(CNIOLinux.EPOLLIN)
```
buttaface, line 74: ClangImporter doesn't register these constants, apparently [because of the empty `__force` macro](https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/kernel/uapi/linux/eventpoll.h#28) as deleting that word from that header gets it to work:
```
swift-nio/Sources/NIO/Linux.swift:74:59: error: module 'CNIOLinux' has no member named 'EPOLLIN'
```
Hacked in the numeric values for now, could try something else if anyone knows any other way.

Sources/NIO/Interfaces.swift
Lukasa: Yeah, that's part of why this was deprecated. Nonetheless, a crash is a bad idea and I'm fine with this change.

Sources/NIO/SocketAddresses.swift
```diff
@@ -264,11 +264,7 @@ public enum SocketAddress: CustomStringConvertible {
             throw SocketAddressError.unixDomainSocketPathTooLong
         }
 
-#if os(Android) // in Android first byte must be zero to use abstract namespace
-        let pathBytes = [0] + Array(unixDomainSocketPath.utf8) + [0]
```
buttaface, line 268: I don't know what this was ever about: a couple tests fail with an empty UDS path with this used, but pass after this pull.

Sources/NIO/System.swift
```diff
@@ -438,7 +455,7 @@ internal enum Posix {
                     var off: off_t = offset
                     let result: ssize_t = Glibc.sendfile(descriptor, fd, &off, count)
                     if result >= 0 {
-                        written = result
+                        written = off_t(result)
```
buttaface, line 458: The only change in this pull that's not conditionally compiled only for Android, with the compiler complaining otherwise:
```
swift-nio/Sources/NIO/System.swift:458:35: error: cannot assign value of type 'ssize_t' (aka 'Int') to type 'off_t' (aka 'Int64')
```

Tests/NIOTests/TestUtils.swift
```diff
@@ -127,9 +127,7 @@ var temporaryDirectory: String {
         // for UNIX Domain Socket paths (which are limited to 103 bytes).
         return "/tmp"
 #else
-#if os(Android)
-        return "/data/local/tmp"
```
buttaface, line 131: `FileManager.default.temporaryDirectory.path` below handles this just fine, using the Termux `TMPDIR` in that app and this path otherwise.

Sources/NIO/Linux.swift
buttaface: Oh, these are only defined for Android before my pull, see the `#if` above.

Sources/NIO/DatagramVectorReadManager.swift
buttaface: Will do.

Sources/NIO/DatagramVectorReadManager.swift
```diff
@@ -100,7 +100,7 @@ struct DatagramVectorReadManager {
                 // TODO(cory): almost all of this except for the iovec could be done at allocation time. Maybe we should?
 
                 // First we set up the iovec and save it off.
-                self.ioVector[i] = IOVector(iov_base: bufferPointer.baseAddress! + (i * messageSize), iov_len: messageSize)
+                self.ioVector[i] = IOVector(iov_base: bufferPointer.baseAddress! + (i * messageSize), iov_len: numericCast(messageSize))
```
buttaface, line 103: @Lukasa, I tried adding another initializer that takes `iov_len:Int` instead, as you suggested above, but that would then cause [other invocations that used `numericCast()`](https://github.com/apple/swift-nio/blob/main/Sources/NIO/PendingWritesManager.swift#L55) to fail, as they'd get confused over which initializer to use. Since those other `numericCast()` invocations were working fine on Android, changed these three in this pull to use it too.

Lukasa (approved):
Nice patch, it seems really good! I'm happy with it.

Lukasa:

PR #1698, opened 2020-11-30
Fix casing of 'SocketAddress.init(packedIpAddress:port:)'

glbrntt:
Motivation:

A new init was added to `SocketAddress` in #1692, but the casing of
`packedIpAddress` is incorrect, it should be `packedIPAddress`. This
hasn't been released yet so let's fix it while we still can!

Modifications:

- s/packedIpAddress/packedIPAddress

Result:

More consistent API


glbrntt:
API breakage is expected, but okay because it hasn't been released yet:

```
17:03:28 /* Removed Decls */
17:03:28 Constructor SocketAddress.init(packedIpAddress:port:) has been removed
```

Lukasa (approved):

PR #1700, opened 2020-12-01
Use CircularBuffer in EmbeddedChannel.

Lukasa:
Motivation:

When running load through EmbeddedChannel we spend an enormous amount of
time screwing around with removing things from Arrays. Arrays are not a
natural data type for `removeFirst()`, and in fact that method is
linear-time on Array due to the need for Array to be zero-indexed. Let's
stop using (and indeed misusing) Array on EmbeddedChannel.

While we're here, if we add some judicious @inlinable annotations we can
also save additional work generating results that users don't need.

Modifications:

- Replace arrays with circular buffers (including marked versions).
- Avoid CoWs and extra allocations on flush.
- Make some API methods inlinable to make them cheaper.

Result:

- Much cheaper EmbeddedChannel for benchmark purposes.

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -193,6 +193,7 @@ public final class EmbeddedEventLoop: EventLoop {
     }
 }
 
+@usableFromInline
 class EmbeddedChannelCore: ChannelCore {
     var isOpen: Bool = true
     var isActive: Bool = false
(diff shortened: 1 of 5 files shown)


Davidde94 (approved):

glbrntt (commented):
Nice one. I assume the CI failure are from the expected allocation counts now being too high.

Lukasa:
Actually in most of these cases the allocation counts are too _low_. Empty arrays don't allocate, but all CircularBuffers do, so we traded better runtime performance for more allocations on EmbeddedChannel construction. I think this is a good trade, but it's a real tradeoff.

glbrntt (approved):

PR #1701, opened 2020-12-09
async/await support

weissi:
Motivation:

The async/await proposal has entered the review phase, so we may want to
start looking into NIO with async/await.

Modifications:

- Add `EventLoopFuture.get() throws await`
- Add `ChannelOutboundInvoker` `async` methods

Result:

If async/await were actually implemented, you could use NIO together
with async/wait. So far, you can only compile your projects.


ktoso (approved):
Yup, looks right :-)

Sources/NIO/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,92 @@
+#if compiler(>=5.4)
+extension EventLoopFuture {
+    public func get() async throws -> Value {
```
kirilltitov, line 3: I'm a bit concerned about the name of this method: it's too generic and wide. I would go for something more specific, like `async()`, in which case it would make more sense (awaiting async thing), however, I'm not sure if it's allowed by compiler anymore.

I think you could also add an extension for `ELF<Void>` with non-returning async method, which would make even more sense with `async()` naming (rather than `get()`). Or alternatively make this method `@discardableResult`.

MaxDesiatov: `get()` would be much more consistent with the rest of the ecosystem. For example, [in the structured concurrency proposal](https://github.com/DougGregor/swift-evolution/blob/structured-concurrency/proposals/nnnn-structured-concurrency.md#task-handles) task handles have an almost identical `get() async throws -> Success` function on them.

ktoso: We use `get()` everywhere in the language so having this do the same is the right way; however ugly the method name is 😉

ktoso: +1 on `@discardableResult` though

MaxDesiatov: Yeah, `Result` from stdlib has this too: `func get() throws -> Success`.

Sources/NIO/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,92 @@
+#if compiler(>=5.4) && $AsyncAwait
+extension EventLoopFuture {
```
weissi, line 2: for @weissi : make promise constructor that takes async function (thanks @tachyonics)

weissi: @tachyonics now done

Lukasa:
As we are not committing to an API at this time I've flagged this as outside of semver.

Sources/NIOAsyncAwaitDemo/main.swift
```diff
@@ -0,0 +1,72 @@
...
+let dg = DispatchGroup()
+dg.enter()
+let task = Task.runDetached {
+    await main()
+    dg.leave()
+}
+dg.wait()
+#else
+print("ERROR: This demo only works with async/await enabled (NIO.System.hasAsyncAwaitSupport = \(NIO.System.hasAsyncAwaitSupport))")
+print("Unfortunately, async/await is currency only supported on macOS.")
```
Lukasa, line 70:
```suggestion
print("Unfortunately, async/await is currently only supported on macOS.")
```

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,143 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,143 @@
...
+                self.succeed(value)
+            } catch {
+                self.fail(error)
+            }
+        }
+    }
+}
+
+extension Channel {
+    public func writeAndFlush<T>(_ any: T) async throws {
```
Lukasa, line 69: Is there a reason this is defined on `Channel` instead of `ChannelOutboundInvoker`?

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,143 @@
...
+        try await self.triggerUserOutboundEvent(event, file: file, line: line).get()
+    }
+}
+
+extension ChannelPipeline {
+    public func addHandler(_ handler: ChannelHandler,
+                           name: String? = nil,
+                           position: ChannelPipeline.Position = .last) async throws {
+        return try await self.addHandler(handler, name: name, position: position).get()
+    }
```
Lukasa, line 141: Is it worth adding the other pipeline operations?

Lukasa (changes requested):

Sources/NIOAsyncAwaitDemo/main.swift
weissi: that's actually a lie now, removed :)

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
weissi: good catch, don't think so :)

weissi: turns out yes: For some reason, in NIO `Channel` has the `<T>wAF(T)` and `ChannelOutboundInvoker` only has the `NIOAny` version... Mirrored that here.

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,189 @@
...
+        #endif
+        #else
+        return false
+        #endif
+    }
+}
+
+#if compiler(>=5.4) // we cannot write this on one line with `&&` because Swift 5.0 doesn't like it...
+#if compiler(>=5.4) && $AsyncAwait
+import NIO
```
kirilltitov, line 33:
```suggestion
import NIO
import _Concurrency
```

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -0,0 +1,189 @@
...
+        return try await self.context(name: name).get()
+    }
+
+    public func context<Handler: ChannelHandler>(handlerType: Handler.Type) async throws -> ChannelHandlerContext {
+        return try await self.context(handlerType: handlerType).get()
+    }
+
+    public func addHandlers(_ handlers: [ChannelHandler],
+                            position: ChannelPipeline.Position = .last) async throws {
+        try await self.addHandlers(handlers, position: position)
```
kirilltitov, line 180:
```suggestion
        try await self.addHandlers(handlers, position: position).get()
```

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
kirilltitov: It won't compile without it. However, if you're aiming for a release Swift version, which won't have this underscored module, it's fine :)

kirilltitov (commented):
I think `EventLoop` should conform to `@unchecked Sendable`, but it's extremely tricky to do that with `#if` checks (not to mention that we don't have `@unchecked` attribute yet, and `UnsafeSendable` triggers an annoying compile warning). Same applies to `EventLoopPromise` (not sure about ELFs, didn't use them in async/await context yet)

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
weissi: this should use the `async` version of `addHandlers`

weissi: you're right, sorry! done

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
weissi: Right, it only compiled if you passed `-Xfrontend -enable-experimental-concurrency` which I did. Now fixed. thanks

ktoso:
> I think EventLoop should conform to @unchecked Sendable, but it's extremely tricky to do that with #if checks (not to mention that we don't have @unchecked attribute yet, and UnsafeSendable triggers an annoying compile warning). Same applies to EventLoopPromise (not sure about ELFs, didn't use them in async/await context yet)

Yeah all those should be `Sendable`, though we don't yet have the unchecked implemented.

Note also that we won't be doing errors/warnings this year for the Sendable checks I believe...

docker/docker-compose.2004.main.yaml
```diff
@@ -20,37 +20,38 @@ services:
...
+      - MAX_ALLOCS_ALLOWED_ping_pong_1000_reqs_1_conn=4450
+      - MAX_ALLOCS_ALLOWED_read_10000_chunks_from_file=190050
       - MAX_ALLOCS_ALLOWED_schedule_10000_tasks=90050
       - MAX_ALLOCS_ALLOWED_scheduling_10000_executions=20150
-      - MAX_ALLOCS_ALLOWED_udp_1000_reqs_1_conn=16250
-      - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=204050 # regression from 5.3 which was 202050
+      - MAX_ALLOCS_ALLOWED_udp_1000_reqs_1_conn=14200
+      - MAX_ALLOCS_ALLOWED_udp_1_reqs_1000_conn=200050
       # - SANITIZER_ARG=--sanitize=thread # TSan broken still
+      - SWIFT_TEST_VERB=build # WARNING: Please remove (workaround https://bugs.swift.org/browse/SR-14268)
```
weissi, line 54: CC @Lukasa

weissi:
@Lukasa I pulled everything that isn't async/await out of this PR now (you know fixing `main` CI etc). So I think this is ready to go in once you're happy. It should be a non-functional change to the `NIO` module (or any other officially exported module).

Sources/_NIOConcurrency/Helpers.swift
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+// note: We have to define the variable `hasAsyncAwait` here because if we copy this code into the property below,
+// it doesn't compile on Swift 5.0.
+#if compiler(>=5.4)
+#if compiler(>=5.4) // we cannot write this on one line with `&&` because Swift 5.0 doesn't like it...
```
theMomax, line 20: I just stumbled across this...I think you forgot to add the ` && $AsyncAwait` here.

weissi: @theMomax good spot! thank you, I'll fix this rn

weissi: https://github.com/apple/swift-nio/pull/1791

theMomax: I'm happy to help...thanks for the quick fix!

PR #1703, opened 2020-12-10
Specialise succeeded Void futures so we allocate less

weissi:
Motivation:

Succeeded `EventLoopFuture<Void>`s are quite important in SwiftNIO, they
happen all over the place. Unfortunately, we usually allocate each time,
unnecessarily.

Modifications:

Offer `EventLoop`s the option to cache succeeded void futures.

Result:

Fewer allocations.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -266,6 +266,20 @@ public protocol EventLoop: EventLoopGroup {
     /// Asserts that the current thread is _not_ the one tied to this `EventLoop`.
     /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
     func preconditionNotInEventLoop(file: StaticString, line: UInt)
+
+    /// Return a succeeded `Void` future.
+    ///
+    /// Semantically, this function is equivalent to calling `makeSucceededFuture(())`.
+    /// Contrary to `makeSucceededFuture`, `makeSucceededVoidFuture` is a customization point for `EventLoop`s which
+    /// allows `EventLoop`s to cache a pre-succeded `Void` future to prevent superfluous allocations.
+    func makeSucceededVoidFuture() -> EventLoopFuture<Void>
+}
+
+extension EventLoop {
+    /// Default implementation of `makeSucceededVoidFuture`: Return a fresh future (which will allocate).
+    public func makeSucceededVoidFuture() -> EventLoopFuture<Void> {
+        return EventLoopFuture(eventLoop: self, value: (), file: "n/a", line: 0)
+    }
 }
 
 extension EventLoopGroup {
@@ -586,7 +600,12 @@ extension EventLoop {
     /// - returns: a succeeded `EventLoopFuture`.
     @inlinable
     public func makeSucceededFuture<Success>(_ value: Success, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
-        return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
+        if Success.self == Void.self {
+            // The as! will always succeed because we previously checked that Success.self == Void.self.
+            return self.makeSucceededVoidFuture() as! EventLoopFuture<Success>
+        } else {
+            return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
+        }
     }
 
     /// An `EventLoop` forms a singular `EventLoopGroup`, returning itself as the 'next' `EventLoop`.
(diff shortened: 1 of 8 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -584,7 +592,11 @@ extension EventLoop {
     /// - returns: a succeeded `EventLoopFuture`.
     @inlinable
     public func makeSucceededFuture<Success>(_ value: Success, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
-        return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
+        if /*!_isDebugAssertConfiguration() && */ (Success.self == Void.self) {
```
weissi, line 595: Comment should be removed, this can always run like this. We don't need `file`/`line` for  succeeded future because the `file`/`line` is only relevant for the debugging if the promise isn't fulfilled (which can't be the case here).

weissi:

weissi:

weissi:

Lukasa:
Seems like all the big wins went away :(

weissi:
> Seems like all the big wins went away :(

Yes, the big wins were old patches. But there are very few void futures allocated synchronously seemingly.

Lukasa (commented):
Do you plan to do this for embedded event loop as well?

weissi:
> Do you plan to do this for embedded event loop as well?

Yes, doing that right now

weissi:
For the record, change of plan: Not doing this for `EmbeddedEL` because there's a nasty side-effect if we implement ELF<Void> caching for EmbeddedEL:
To cache it, I naturally need a retain cycle EL -> cached future -> EL .
In SelectableEL, that's no problem because people need to shut them down anyway to not leak a thread. It may actually be beneficial to also leak the EL because if they forget to shutdown because you can spot that in Mem Graph Debugger.
For EmbeddedEL, things are much different. We never shut them down :stuck_out_tongue: .

Sources/NIO/EventLoop.swift
```diff
@@ -586,7 +600,12 @@ extension EventLoop {
     /// - returns: a succeeded `EventLoopFuture`.
     @inlinable
     public func makeSucceededFuture<Success>(_ value: Success, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
-        return EventLoopFuture<Success>(eventLoop: self, value: value, file: file, line: line)
+        if Success.self == Void.self {
+            // The as! will always succeed because we previously checked that Success.self == Void.self.
+            return self.makeSucceededVoidFuture() as! EventLoopFuture<Success>
```
glbrntt, line 605: Why is this preferred over `func makeSucceededFuture(_ value: Void, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Void>`?

Lukasa: Because this one is cached.

Lukasa: Oh, also because this one hits unspecialised generic code and doesn't potentially cause confusion in the type checker.

glbrntt: For clarity, I meant why not:

```swift
func makeSucceededFuture(_ value: Void, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Void> {
  return self.makeSucceededVoidFuture()
}
```

Anyway, "because this one hits unspecialised generic code and doesn't potentially cause confusion in the type checker" is what I was looking for, thanks!

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -59,6 +59,12 @@ internal final class SelectableEventLoop: EventLoop {
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
     private var tasksCopy = ContiguousArray<() -> Void>()
+    @usableFromInline
+    internal var _succeededVoidFuture: Optional<EventLoopFuture<Void>> = nil {
+        didSet {
+            assert(self.inEventLoop)
```
Lukasa, line 65: As an incredibly minor detail: can we use the `assertInEventLoop` function? I _know_ that on this specific event loop it doesn't matter, but it's nice to eliminate this particular anti-pattern from the codebase where we can.

Lukasa (approved):
One minor piece of feedback there: feel free to merge once it's addressed.

Sources/NIO/SelectableEventLoop.swift
weissi: oh yes!

weissi: fixed, thank you! Will merge when green

PR #1706, opened 2020-12-17
Retry on EPROTOTYPE on socket writes.

Lukasa:
Motivation:

When writing to a network socket on Apple platforms it is possible to
see EPROTOTYPE returned as an error. This is an undocumented and
special-case error code that appears to be associated with socket
shutdown, and so can fire when writing to a socket that is being shut
down by the other side. This should not be fired into the pipeline but
instead should be retried.

Modifications:

- Retry EPROTOTYPE errors on socket write methods.
- Add an (unfortunately) probabilistic test bed.

Result:

Should avoid weird error cases.
Resolves swift-server/async-http-client#322.

### DIFF
--- a/Sources/NIO/System.swift
+++ b/Sources/NIO/System.swift
@@ -150,6 +150,7 @@ private func preconditionIsNotUnacceptableErrno(err: CInt, where function: Strin
 @inline(__always)
 @discardableResult
 internal func syscall<T: FixedWidthInteger>(blocking: Bool,
+                                            eprototypeWorkaround: Bool = false,
                                             where function: String = #function,
                                             _ body: () throws -> T)
         throws -> IOResult<T> {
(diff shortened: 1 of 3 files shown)


Sources/NIO/System.swift
```diff
@@ -150,18 +150,26 @@ private func preconditionIsNotUnacceptableErrno(err: CInt, where function: Strin
...
-            switch (err, blocking) {
-            case (EINTR, _):
+            switch (err, blocking, eprototypeWorkaround) {
+            case (EINTR, _, _):
                 continue
-            case (EWOULDBLOCK, true):
+            case (EWOULDBLOCK, true, _):
                 return .wouldBlock(0)
+            #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+            case (EPROTOTYPE, _, eprototypeWorkaround):
```
glbrntt, line 167: Is this meant to be `true` rather than `eprototypeWorkaround`?

Lukasa: Uhhhhh yes.

glbrntt (approved):
Nice one!

saroar:
@Lukasa @glbrntt thanks 🧘🏻‍♂️🙏🏾

PR #1707, opened 2020-12-18
Revert "Retry on EPROTOTYPE on socket writes. (#1706)"

Lukasa:
This reverts commit 4853e910e8aca4e65f92b2d335a59da337bd3730.

While we have been able to observe the effect that this change was
trying to workaround, the change seems to interact poorly with a
different issue in Big Sur that can cause EPROTOTYPE to be consistently
emitted during socket writes on otherwise connected sockets. This would
change a connection-terminating error into a 100% CPU spin that rendered
the event loop entirely useless: a substantial regression.

For this reason, we should back this out until the issue is better
characterised.


glbrntt (approved):
Welp.

PR #1709, opened 2020-12-21
Add allocation test for adding multiple handlers

PeterAdams-A:
Motivation:

I believe there is at least 1 avoidable allocation in this area.
Even if there isn't, making sure we don't increase allocations is good.

Modifications:

Add a test of allocations when adding multiple handlers.
Set limits for docker images.

Result:

Allocations when adding multiple handlers are now checked.


PeterAdams-A:
Went in as part of #1710

PR #1710, opened 2020-12-21
Remove an allocation from addHandlers

PeterAdams-A:
Remove an allocation from addHandlers
    
Motivation:
    
Fewer allocations should improve performance.
    
Modifications:
    
Split out a sub function from addHandlers.
    
I originally thought I'd have to change the part of this
function which reads `var handlers = handlers` as there was
a surprising allocation at the beginning of this function.
It seems that breaking out some of the logic is sufficient
to remove an allocation.
    
Result:
    
1 fewer allocation.


Davidde94 (approved):

PeterAdams-A:
Sorry @Davidde94 I had to add an extra commit.

Davidde94 (approved):

PR #1711, opened 2020-12-22
DO NOT MERGE - Struct EventLoopFuture

PeterAdams-A:
An idea which was sitting in the back of my mind so I thought I'd sketch it out.

Make event loop future a struct - which for proper futures just contains the current EventLoopFuture but can directly contain the results when makeSucceeded, or makeFailed is used.

### Motivation:

Nio allocates loads of futures.  Allocation is known to be slow.

### Modifications:

Make event loop future a struct - which for proper futures just contains the current EventLoopFuture but can directly contain the results when makeSucceeded, or makeFailed is used.

### Result:

This is obviously API breaking.
It is source compatible though.
There are definitely fewer allocations when using this with current NIO applications.
**It is unfortunately also quite a lot slower** - I've not looked into why as it's just an idea for discussion.

There has previously been discussion of adding synchronous methods to NIO to avoid allocating futures which will be even faster than this.   When thinking about this from the point of view of adding channel handlers - if there was a sync method for adding handlers, you still need to allocate a future to return from the channelInitialiser method.  Of course that method could be changed too but if not careful all methods end up existing in sync and async forms with burden on the user and for maintenance.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,7 +24,8 @@ var targets: [PackageDescription.Target] = [
                            "CNIOWindows",
                            "NIOConcurrencyHelpers",
                            "CNIOAtomics",
-                           "CNIOSHA1"]),
+                           "CNIOSHA1",
+                           .product(name: "Atomics", package: "swift-atomics")]),
     .target(name: "NIOFoundationCompat", dependencies: ["NIO"]),
     .target(name: "CNIOAtomics", dependencies: []),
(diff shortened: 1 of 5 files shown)


weissi:
@PeterAdams-A can we close this or would you like to keep it open?

PeterAdams-A:
Closing - we can always still find it in history if needed.

PR #1712, opened 2020-12-23
Remove error handling for adding/removing handlers functions

PeterAdams-A:
Motivation:

The handlerAdded and handlerRemoved functions are not throwing.
I can't spot anyway way and exception could actually happen
so no point in catering for the case when one does.

Modifications:

Remove possibility that an exception was thrown when adding or
removing handlers as none of the called functions are throwing.

Result:

Fewer lines of code, less complexity.


glbrntt (approved):
Nice catch.

Lukasa (approved):
Nice cleanup.

PR #1715, opened 2021-01-07
Revert "Implement faster pointer rebasing. (#1696)"

Lukasa:
This reverts commit 4bf379b1566c04807e173862acab301aaa22b566.

While this commit sped up downstream workloads, it regressed local benchmarks, which I missed. We should pull it out until we know why.


Lukasa:

Lukasa:
Marking this as a draft, because the above performance results suggest that this did _not_ regress perf.

Lukasa:
I think we have sufficient evidence that this patch did not, in fact, regress perf.

PR #1720, opened 2021-01-14
order perf tests by name in docker files

weissi:
Motivation:

When you make a change that affects many performance tests, it's often
easier to just copy the results from CI. Unfortunately, that makes the
diff hard to read because the order is arbitrary.

Modifications:

Sort the list so you can always easily get to the same order as the
docker file by using `| sort` or `:sort` in vim.

Result:

Easier to update perf tests.


PeterAdams-A (approved):
Seems like a fine idea.

PR #1721, opened 2021-01-15
Fixed typo in EventLoopFuture docs

mr-j-tree:
"provably" becomes "probably"


Lukasa:
This is not a typo. We want the word "provably": that is, you should only not call `hop(to:)` when you can prove that the hop is unnecessary.

mr-j-tree:
My mistake! Not a word I've heard used before.

As you were! :)

PR #1725, opened 2021-01-20
Add support for replacing a channel handler

### LINKED ISSUE #1694: Add a replace decoder or replace handler function
To implement a custom protocol, you may have many states and each state require a decoder. I thinks it will be very convenient to replace below code:

```swift
        context.pipeline
            .removeHandler(name: decoderName)
            .and(context.pipeline.addHandler(newHandler, name: decoderName, position: .before(self)))
            .cascade(to: nil)
```

Davidde94:
Resolves #1694 

### Motivation:

Third party request to be able to replace a channel handler, instead of having to add then remove. This makes the API slightly nicer/easier to work with.

### Modifications:

Add 2 public methods to replace an "old" channel handler with a "new" one. One future-based, one promise-based.

### Result:

You can now replace a channel handler with a call to `channel.pipeline.replaceHandler(old, with: new)`

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -326,6 +326,8 @@ public final class ChannelPipeline: ChannelInvoker {
         existing.prev = new
         prev?.next = new
     }
+    
+    // MARK: - Remove channel handler
 
     /// Remove a `ChannelHandler` from the `ChannelPipeline`.
     ///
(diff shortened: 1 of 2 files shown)


Lukasa (commented):
I don't think this behaves sensibly if the removable channel handler in question does delayed handler removal. A removable channel handler is entitled to remove itself _after_ the call to `removeHandler`, so long as it's appropriately buffering.

I'd normally be inclined to say that this should use a `flatMap` instead of an `and`. But in either case we need to motivate what the intended order of operations should be, and document how this will manifest in a pipeline, as understanding what the intermediate states will be is really important.

Davidde94:
@Lukasa good point. I went in with the thought that it makes sense for the new handler to be inserted *before* the old one is removed, so that the new handler can immediately begin to buffer/handle data.

There should be exactly one intermediate state, where the new handler has been added but the old has not yet been removed.

PR #1726, opened 2021-01-20
[WIP] Add stream support to ChannelHandlerContext

### LINKED ISSUE #14: think about if we want easier, more high-level byte writing APIs
suggested by @ddunbar 

SwiftNIO might be appealing to developers of network servers that need good but maybe not the best network performance. In these cases, the allocate `ByteBuffer`, accumulate, `ctx.write`, ..., `ctx.flush` might be a bit awkward. A simpler API might also be a good starting point. For `ChannelHandlers` with `OutboundOut = ByteBuffer`, @ddunbar proposed the following API:

    ctx.writeWithStream { stream in
        stream <<< “this” <<< Format.asJSON(…) <<< “bla bla bla”
    }

which would basically do:

    {
        var buf = ctx.channel.allocate(capacity: 256 /* random guess */)
        buf.write(staticString: "this")
        buf.write(bytes: Format.asJSON(...)) // Format.asJSON is actually more sophisticated than that. It returns a "formatter", which is a struct which knows how to render the output as JSON directly onto the stream. This avoids creating unnecessary temporary allocations of the JSON data.
        buf.write(staticString: "bla bla bla")
        return ctx.write(self.wrapOutboundOut(.byteBuffer(buf)))
    }()

and I guess we could also offer a `ctx.writeAndFlushWithStream` API. Or maybe `ctx.writeWithStream(flush: Bool = false, (Stream) -> EventLoopPromise<()>) -> EventLoopPromise<()>` 🤔.

The custom `<<<` operator would be defined for everything we accept: `ByteBuffer`, `(Contiguous)Collection of UInt8`, `String`, `StaticString`, the various integers, ...

We should definitely evaluate having such a more high-level API.

This should probably live in some yet to be created `NIOExtras` module. And we need to think about how to make it perform well and in an unsurprising way.

Davidde94:
Resolves #14

### Motivation:

The API is easier to use when outbound writing a variety of data types.

### Modifications:

Add a new `Stream` class, and a new method `writeOutboundWithStream` to `ChannelHandlerContext`. You provide a closure as the last argument to this method, which takes  a `Stream` as it's only argument. You then write to the stream using the new `<<<` operator.

Multiple types are supported including:
- `String`
- `StaticString`
- `Substring`
- `[UInt8]`
- `FixedWidthInteger`

### Result:

```
var buffer = ByteBuffer()
buffer.writeString("Hello")
buffer.writeString("world")
buffer.writeInteger(1234)
context.writeAndFlush(NIOAny(buffer), promise: promise)
```

turns into

```
context.writeOutboundWithStream(promise: promise) { stream in
    stream <<< "hello" <<< "world" <<< 1234
}
```

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -1656,6 +1656,7 @@ extension ChannelPipeline: CustomDebugStringConvertible {
 
 extension ChannelHandlerContext {
     
+    /// Used to stream data into a `ChannelHandlerContext`
     public class Stream {
         
         fileprivate var buffer: ByteBuffer
@@ -1666,6 +1667,10 @@ extension ChannelHandlerContext {
         
     }
     
+    /// Streams data to a `ChannelHandlerContext` and then flushes.
+    /// - parameter initialCapacity: The initial capacity of the stream, defaults to 1024 bytes.
+    /// - parameter promise: An `EventLoopPromise` that will be notified when writing to the `ChannelHandlerContext` is complete.
+    /// - parameter writer: A closure that takes a `Stream` as it's first and only argument. You then write to this `Stream` using the `<<<` operator.
     public func writeOutboundWithStream(initialCapacity: Int = 1024, promise: EventLoopPromise<Void>? = nil, _ writer: (Stream) -> Void) {
         let stream = Stream(buffer: self.channel.allocator.buffer(capacity: initialCapacity))
         writer(stream)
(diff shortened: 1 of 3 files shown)


Sources/NIO/ChannelPipeline.swift
```diff
@@ -1639,3 +1639,77 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
```
Lukasa, line 1655: This seems...unlikely to be where you want it

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1639,3 +1639,77 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+
+extension ChannelHandlerContext {
+    
+    /// Used to stream data into a `ChannelHandlerContext`
+    public class Stream {
```
Lukasa, line 1660: We should avoid using a `class` here: I think a different design should make it possible to use a `struct` without too much difficulty.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1639,3 +1639,77 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+    /// - parameter writer: A closure that takes a `Stream` as it's first and only argument. You then write to this `Stream` using the `<<<` operator.
+    public func writeOutboundWithStream(initialCapacity: Int = 1024, promise: EventLoopPromise<Void>? = nil, _ writer: (Stream) -> Void) {
+        let stream = Stream(buffer: self.channel.allocator.buffer(capacity: initialCapacity))
+        writer(stream)
+        return self.writeAndFlush(NIOAny(stream.buffer), promise: promise)
+    }
+    
+}
+
+precedencegroup StreamPrecedence {
```
Lukasa, line 1682: Do we need a new precedence group?

Lukasa (commented):
This is a really interesting initial prototype, and it definitely implements the original proposal. However, I think I'd like to see the design space explored a bit. We shouldn't necessarily just import what C++ does unthinkingly here, when there may be easier and better designs for Swift.

As an example, what would the ramifications be of using function builders here? They have limited control flow, but in this mode that's not necessarily a huge issue. That design may work better, but we'd have to dive into it to see.

I'm very reluctant to add new API in this area without exploring what some of our options look like.

Sources/NIO/ChannelPipeline.swift
Davidde94: ...yep

Sources/NIO/ChannelPipeline.swift
Davidde94: I'm not sure this is possible due to mutability constraints, but I'm open to being wrong of course

Davidde94:
Closing as we've talked offline and I've written up a better solution in #1727

PR #1727, opened 2021-01-20
Support result builders in ByteBuffer

Davidde94:
Add support for creating and writing to ByteBuffers using result builders.

### Motivation:

When writing multiple data types, or a lot of the same type, this is an easier way to push data into a byte buffer.

### Modifications:

Add a new init and write method that takes a result builder as an argument, calls that builder, and then writes each part to the buffer. Also defined a `ByteBufferSerializable` protocol that can be used to enable custom types to be used in a result builder.

### Result:

Old initialisation:
```
var buffer = ByteBuffer()
buffer.writeString("hello")
buffer.writeString("world")
```

New initialisation:
```
var buffer = ByteBuffer {
    "hello"
    "world"
}

-----------------------------------

Old writing:
```
var size = 0
size += buffer.writeString("hello")
size += buffer.writeInteger(1234)
size += buffer.writeString("world")
```

New writing
```
size = buffer.write {
    "hello"
    1234
    "world"
}

### DIFF
--- a/Sources/NIO/ByteBuffer-builder.swift
+++ b/Sources/NIO/ByteBuffer-builder.swift
@@ -1,0 +1,51 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+
+public protocol ByteBufferSerializable {
+    
+    @discardableResult func write(into buffer: inout ByteBuffer) -> Int
+    
+}
+
+extension String: ByteBufferSerializable {
+    
+    public func write(into buffer: inout ByteBuffer) -> Int {
+        buffer.writeString(self)
+    }
+    
+}
+
+@resultBuilder
+public struct BufferBuilder {
+    public static func buildBlock(_ components: ByteBufferSerializable...) -> [ByteBufferSerializable] {
+        components
+    }
+}
+
+extension ByteBuffer {
+ 
+    init(@BufferBuilder contents: () -> [ByteBufferSerializable]) {
+        self.init()
+        for part in contents() {
+            part.write(into: &self)
+        }
+    }
+    
+    public mutating func write(@BufferBuilder contents: () -> [ByteBufferSerializable]) -> Int {
+        contents().map { $0.write(into: &self) }.reduce(0, +)
+    }
+    
+}
--- a/Sources/NIO/ByteBuffer-builder.swift
+++ b/Sources/NIO/ByteBuffer-builder.swift
@@ -14,6 +14,8 @@
 
 import Dispatch
 
+// MARK: - Serializable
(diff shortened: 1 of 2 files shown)


PeterAdams-A:
I assume you'll add some comments later.
Worth talking to Cory - I'm sure he has advice on extending types you don't own but I forget exactly what it is right now.

Sources/NIO/ByteBuffer-builder.swift
```diff
@@ -0,0 +1,123 @@
...
+    public static func buildBlock(_ components: ByteBufferSerializable...) -> [ByteBufferSerializable] {
+        return components
+    }
+}
+
+extension ByteBuffer {
+ 
+    /// Creates a new `ByteBuffer` using a result builder.
+    /// - parameter contents: A result builder that uses `ByteBufferSerializable` types.
+    public init(@BufferBuilder contents: () -> [ByteBufferSerializable]) {
```
weissi, line 107: this will be very inefficient because it'll allocate an array of existentials.

Sources/NIO/ByteBuffer-builder.swift
```diff
@@ -0,0 +1,123 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+
+// MARK: - Serializable
+
+/// Used to write a conforming type into a `ByteBuffer` so that it can be used in a result builder.
+public protocol ByteBufferSerializable {
```
weissi, line 20: unprefixed global protocol

Sources/NIO/ByteBuffer-builder.swift
```diff
@@ -0,0 +1,123 @@
...
+extension UInt16: ByteBufferSerializable {}
+extension UInt32: ByteBufferSerializable {}
+extension UInt64: ByteBufferSerializable {}
+
+// MARK: - Buffer builder
+
+#if compiler(>=5.4)
+
+@resultBuilder
+public struct BufferBuilder {
```
weissi, line 97: unprefixed global type

weissi (changes requested):
The idea is certainly interesting but I'm not sure if it's worth it given that it's very inefficient compared to the normal way of just writing multiple items into a BB.

Maybe we could introduce something similar in NIOExtras and bake the API a little?

weissi:
This also introduces API that's only available with a certain Swift version which is a problem because it will be non-obvious to users that if they use this, their code will work only on >=5.4.

Davidde94:
@weissi surely the same concern about Swift 5.4 applies to async await?

weissi:
> @weissi surely the same concern about Swift 5.4 applies to async await?

No because to use `await`, users have to make `async` functions and convert their whole API, this is a very very deliberate and obvious change in your program. Also, currently all that sits behind `import _NIOConcurrency` and may end up in a `import NIOConcurrency` module. But because revamping all your code to a/a is so obvious I don't think it's necessary.

Sources/NIO/ByteBuffer-builder.swift
Davidde94: Why does it need a prefix?

Davidde94:
@weissi fair enough!

Sources/NIO/ByteBuffer-builder.swift
weissi: https://github.com/apple/swift-nio/blob/main/docs/public-api.md#promises-the-nio-team-make

PR #1728, opened 2021-01-21
Use welcoming language

fabianfett:
To comply with the new [Swift code of conduct](https://swift.org/code-of-conduct/), let's make the language a little more welcoming.

### Modifications:

- Replaced usages of "sanity" with more welcoming alternatives
- Added "sanity" to the list of unacceptable terms. (Commented out, until the CI has been updated to use the new soundness check) 

### Result:

- Swift NIO is nicer.

### DIFF
--- a/Sources/CNIOSHA1/c_nio_sha1.c
+++ b/Sources/CNIOSHA1/c_nio_sha1.c
@@ -58,7 +58,7 @@
 
 
 
-/* sanity check */
+/* soundness check */
 #if BYTE_ORDER != BIG_ENDIAN
 # if BYTE_ORDER != LITTLE_ENDIAN
 #  define unsupported 1
--- a/Sources/NIOHTTP1/HTTPEncoder.swift
+++ b/Sources/NIOHTTP1/HTTPEncoder.swift
@@ -101,7 +101,7 @@ private enum BodyFraming {
 ///
 /// Note that for HTTP/1.0 if there is no Content-Length then the response should be followed
 /// by connection close. We require that the user send that connection close: we don't do it.
-private func sanitizeTransportHeaders(hasBody: HTTPMethod.HasBody, headers: inout HTTPHeaders, version: HTTPVersion) -> BodyFraming {
+private func correctlyFrameTransportHeaders(hasBody: HTTPMethod.HasBody, headers: inout HTTPHeaders, version: HTTPVersion) -> BodyFraming {
     switch hasBody {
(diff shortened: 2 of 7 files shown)


PeterAdams-A:
Should there be a scripts/soundness.sh included in this PR?

Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -101,7 +101,7 @@ private enum BodyFraming {
 ///
 /// Note that for HTTP/1.0 if there is no Content-Length then the response should be followed
 /// by connection close. We require that the user send that connection close: we don't do it.
-private func sanitizeTransportHeaders(hasBody: HTTPMethod.HasBody, headers: inout HTTPHeaders, version: HTTPVersion) -> BodyFraming {
+private func correctlyFrameTransportHeaders(hasBody: HTTPMethod.HasBody, headers: inout HTTPHeaders, version: HTTPVersion) -> BodyFraming {
```
Lukasa, line 104: As a linguistic note I don't think this change was needed: the word "sanitize" does not relate to the word "sanity" but to the word "sanitary". While they have the same latin root the meanings are different: sanity derives the meaning of mental health, while sanitary covers physical disease.

Regardless, this change doesn't _hurt_ in any meaningful way so I think we should just leave it.

Lukasa (approved):
LGTM.

fabianfett:
@PeterAdams-A 

> Should there be a scripts/soundness.sh included in this PR?

I renamed `scripts/sanity.sh` to `scripts/soundness.sh`.

Sources/CNIOSHA1/c_nio_sha1.c
```diff
@@ -58,7 +58,7 @@
 
 
 
-/* sanity check */
+/* soundness check */
```
weissi, line 61: this change needs to be done in `update_and_patch_sha1.sh`. This file can't be manually edited because it's vendored by that script. Also add it to the top-level comment which summarises the changes.

docker/docker-compose.yaml
```diff
@@ -24,7 +24,11 @@ services:
 
   sanity:
```
weissi, line 25: there's a left-over "sanity"

weissi (commented):
Thank you, very much appreciated! we'll need a follow-up PR to fix this though as it edits code that's updated by a script manually.

PR #1729, opened 2021-01-21
B2MD: Fix default memory reclamation strategy to match its documentation

fabianfett:
In cases in which only a small amount of bytes were consumed from a `ByteBuffer` during decoding a small error in our logic could lead to extensive reclaims of memory, degrading the overall performance and allocating way to much memory.

### Motivation:

- In cases in which only a small amount was decoded from the `ByteToMessageHandler`'s internal byteBuffer, we ran into reallocations based on our previous logic. If the byteBuffer was larger than 1kb and less than 50% were consumed we reallocated.
- This leads to performance problems
- But more importantly if within the decode function a byteBuffer slice is read, the internal `ByteToMessageHandler`'s byteBuffer's storage is now referenced at multiple locations a leading to fast and excessive memory growth.

### Modifications:

- A logic bug has been fixed, we reclaim memory now, if more than 50% of the byteBuffer has been read.
- The tests have been adjusted to reflect the fixed logic bug.

### Result:

- Faster decoding of small messages
- Less memory footprint when reading slices in small messages.

### DIFF
--- a/Sources/NIO/Codec.swift
+++ b/Sources/NIO/Codec.swift
@@ -222,13 +222,13 @@ extension ByteToMessageDecoder {
     public func shouldReclaimBytes(buffer: ByteBuffer) -> Bool {
         // We want to reclaim in the following cases:
         //
-        // 1. If there is more than 2kB of memory to reclaim
+        // 1. If there is at least 2kB of memory to reclaim
         // 2. If the buffer is more than 50% reclaimable memory and is at least
         //    1kB in size.
-        if buffer.readerIndex > 2048 {
+        if buffer.readerIndex >= 2048 {
             return true
         }
-        return buffer.capacity > 1024 && (buffer.capacity - buffer.readerIndex) >= buffer.readerIndex
+        return buffer.storageCapacity > 1024 && (buffer.storageCapacity - buffer.readerIndex) < buffer.readerIndex
     }
 
     public func wrapInboundOut(_ value: InboundOut) -> NIOAny {
--- a/Tests/NIOTests/CodecTest.swift
+++ b/Tests/NIOTests/CodecTest.swift
@@ -210,19 +210,43 @@ public final class ByteToMessageDecoderTest: XCTestCase {
         _ = try channel.pipeline.addHandler(decoder).wait()
 
         // We're going to send in 513 bytes. This will cause a chunk to be passed on, and will leave
-        // a 512-byte empty region in a 513 byte buffer. This will not cause a shrink.
+        // a 512-byte empty region in a byte buffer with a capacity of 1024 bytes. Since 512 empty
+        // bytes are exactly 50% of the buffers capacity and not one tiny bit more, the empty space
+        // will not be reclaimed.
         var buffer = channel.allocator.buffer(capacity: 513)
         buffer.writeBytes(Array(repeating: 0x04, count: 513))
         XCTAssertTrue(try channel.writeInbound(buffer).isFull)
 
+        XCTAssertEqual(decoder.cumulationBuffer!.capacity, 1024)
         XCTAssertEqual(decoder.cumulationBuffer!.readableBytes, 1)
         XCTAssertEqual(decoder.cumulationBuffer!.readerIndex, 512)
 
-        // Now we're going to send in another 513 bytes. This will cause another chunk to be passed in,
-        // but now we'll shrink the buffer.
+        // Next we're going to send in another 513 bytes. This will cause another chunk to be passed
+        // into our decoder buffer, which has a capacity of 1024 bytes, before we pass in another
+        // 513 bytes. Since we already have written to 513 bytes, there isn't enough space in the
+        // buffer, which will cause a resize to a new underlying storage with 2048 bytes. Since the
+        // `LargeChunkDecoder` has consumed another 512 bytes, there are now two bytes left to read
+        // (513 + 513) - (512 + 512). The reader index is at 1024. The empty space has not been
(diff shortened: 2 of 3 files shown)


Sources/NIO/Codec.swift
```diff
@@ -222,13 +222,13 @@ extension ByteToMessageDecoder {
...
-        // 1. If there is more than 2kB of memory to reclaim
+        // 1. If there is at least 2kB of memory to reclaim
         // 2. If the buffer is more than 50% reclaimable memory and is at least
         //    1kB in size.
-        if buffer.readerIndex > 2048 {
+        if buffer.readerIndex >= 2048 {
             return true
         }
-        return buffer.capacity > 1024 && (buffer.capacity - buffer.readerIndex) >= buffer.readerIndex
+        return buffer.capacity >= 1024 && (buffer.capacity - buffer.readerIndex) < buffer.readerIndex
```
weissi, line 231: - the first change `>` --> `>=` essentially halves the threshold. We only allocate ByteBuffers in powers of 2. So outside of slices, we should only ever see a `1024` capacity or a `2048` capacity BB, nothing in between.
- I'd revert this change and go back to `>` because it's unrelated.
- let's use `.storageCapacity` even though in this case it's debatable I think

Sources/NIO/Codec.swift
```diff
@@ -222,13 +222,13 @@ extension ByteToMessageDecoder {
     public func shouldReclaimBytes(buffer: ByteBuffer) -> Bool {
         // We want to reclaim in the following cases:
         //
-        // 1. If there is more than 2kB of memory to reclaim
+        // 1. If there is at least 2kB of memory to reclaim
         // 2. If the buffer is more than 50% reclaimable memory and is at least
         //    1kB in size.
-        if buffer.readerIndex > 2048 {
+        if buffer.readerIndex >= 2048 {
```
weissi, line 228: what's the motivation for this change? Seems unrelated to the bugfix and obfuscates the important change?

Tests/NIOTests/CodecTest.swift
```diff
@@ -210,19 +210,43 @@ public final class ByteToMessageDecoderTest: XCTestCase {
         _ = try channel.pipeline.addHandler(decoder).wait()
 
         // We're going to send in 513 bytes. This will cause a chunk to be passed on, and will leave
-        // a 512-byte empty region in a 513 byte buffer. This will not cause a shrink.
+        // a 512-byte empty region in a byte buffer with a capacity of 1024 bytes. Since 512 empty
```
weissi, line 213: I think we should leave the old unit tests alone and add targeted new tests that specifically test for what was broken.

weissi (changes requested):
Thanks very much! This is definitely a fix we want to have asap. Left some minor feedback.

Sources/NIO/Codec.swift
fabianfett: In `CodecTests` we currently explicitly test this behavior. The test succeeded up until now, since it used the buggy code path:
```swift
    func testMemoryIsReclaimedIfLotsIsAvailable() throws {
        defer {
            XCTAssertNoThrow(try channel.finish())
        }

        let decoder = ByteToMessageHandler(OnceDecoder()) // <--- the once decoder consumes exactly 2048 bytes.
        _ = try channel.pipeline.addHandler(decoder).wait()

        // We're going to send in 5119 bytes. This will be held.
        var buffer = channel.allocator.buffer(capacity: 5119)
        buffer.writeBytes(Array(repeating: 0x04, count: 5119))
        XCTAssertTrue(try channel.writeInbound(buffer).isEmpty)

        XCTAssertEqual(decoder.cumulationBuffer!.readableBytes, 5119)
        XCTAssertEqual(decoder.cumulationBuffer!.readerIndex, 0)

        // Now we're going to send in one more byte. This will cause a chunk to be passed on,
        // shrinking the held memory to 3072 bytes. However, memory will be reclaimed.
        XCTAssertTrue(try channel.writeInbound(buffer.getSlice(at: 0, length: 1)).isFull)
        XCTAssertEqual(decoder.cumulationBuffer!.readableBytes, 3072)
        XCTAssertEqual(decoder.cumulationBuffer!.readerIndex, 0)
    }
```

Tests/NIOTests/CodecTest.swift
fabianfett: With the fixed logic above this tests fails without modifications. For this reason, I adjusted it.

Sources/NIO/Codec.swift
fabianfett: Fixed.

Sources/NIO/Codec.swift
weissi: Thank you! That makes sense.

Tests/NIOTests/CodecTest.swift
weissi: Thank you! That makes sense.

Sources/NIO/Codec.swift
weissi: offline discussion with Fabian:

- the change in this logic is kinda arbitrary but it also won't hurt
- the documentation is just a code comment and _not_ in the public API docs so we never guaranteed what we'd do anyway so shifting the arbitrary limit by 1 byte shouldn't matter

weissi (approved):
Thank you! Really great fix!

Sources/NIO/Codec.swift
weissi: thank you

Lukasa (approved):
Nice, LGTM.

PR #1733, opened 2021-01-22
B2MD: Don't try to reclaim if continuing to parse

weissi:
Motivation:

B2MD called out to the decoder's `shouldReclaimBytes` after every
parsing attempt, even if the parser said `.continue`.

That's quite pointless because we won't add any bytes into the buffer
before we're trying the parser again.

Modifications:

Only ask the decoder if we should reclaim bytes if the decoder actually
says `.needMoreData`.

Result:

Faster, better, more sensible.

### DIFF
--- a/Tests/NIOTests/CodecTest.swift
+++ b/Tests/NIOTests/CodecTest.swift
@@ -200,14 +200,14 @@ public final class ByteToMessageDecoderTest: XCTestCase {
         XCTAssertEqual(testDecoderIsNotQuadratic_reallocs, 3)
     }
 
-    func testMemoryIsReclaimedIfMostIsConsumed() throws {
+    func testMemoryIsReclaimedIfMostIsConsumed() {
         let channel = EmbeddedChannel()
         defer {
-            XCTAssertNoThrow(try channel.finish())
+            XCTAssertNoThrow(XCTAssertTrue(try channel.finish().isClean))
         }
 
         let decoder = ByteToMessageHandler(LargeChunkDecoder())
-        _ = try channel.pipeline.addHandler(decoder).wait()
+        XCTAssertNoThrow(try channel.pipeline.addHandler(decoder).wait())
 
         // We're going to send in 513 bytes. This will cause a chunk to be passed on, and will leave
         // a 512-byte empty region in a byte buffer with a capacity of 1024 bytes. Since 512 empty
@@ -215,7 +215,9 @@ public final class ByteToMessageDecoderTest: XCTestCase {
         // will not be reclaimed.
         var buffer = channel.allocator.buffer(capacity: 513)
         buffer.writeBytes(Array(repeating: 0x04, count: 513))
-        XCTAssertTrue(try channel.writeInbound(buffer).isFull)
+        XCTAssertNoThrow(XCTAssertTrue(try channel.writeInbound(buffer).isFull))
+        XCTAssertNoThrow(XCTAssertEqual(ByteBuffer(repeating: 0x04, count: 512), try channel.readInbound()))
+        XCTAssertNoThrow(XCTAssertNil(try channel.readInbound()))
 
         XCTAssertEqual(decoder.cumulationBuffer!.capacity, 1024)
         XCTAssertEqual(decoder.cumulationBuffer!.readableBytes, 1)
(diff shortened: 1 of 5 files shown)


weissi:
CC @fabianfett

Tests/NIOTests/CodecTest.swift
```diff
@@ -200,22 +200,24 @@ public final class ByteToMessageDecoderTest: XCTestCase {
         XCTAssertEqual(testDecoderIsNotQuadratic_reallocs, 3)
     }
 
-    func testMemoryIsReclaimedIfMostIsConsumed() throws {
+    func testMemoryIsReclaimedIfMostIsConsumed() {
         let channel = EmbeddedChannel()
         defer {
-            XCTAssertNoThrow(try channel.finish())
+            XCTAssertNoThrow(XCTAssertTrue(try channel.finish().isClean))
```
fabianfett, line 206: I think we can get rid of the enclosing `XCTAssertNoThrow`...

weissi: @fabianfett but we want that?

weissi: it has a `try` and I don't want it to throw, right?

weissi:

Lukasa (approved):

Tests/NIOTests/CodecTest.swift
fabianfett: Correct, but AFAIK XCTAssertTrue also catches `try`.

fabianfett: In other word `XCTAssertTrue` has a build in `XCTAssertNoThrow`

weissi:

Tests/NIOTests/CodecTest.swift
fabianfett: Probably not available for Swift 5.0.

fabianfett (approved):
LGTM

weissi:
> circular_buffer_into_byte_buffer_1kb	0.051517237	0.041180082	previous	25%
> circular_buffer_into_byte_buffer_1mb	0.094858812	0.082505024	previous	14%
> byte_buffer_view_iterator_1mb	0.117302428	0.126364362	current	-7%

@Lukasa / @tomerd I didn't touch anything that's related to these tests, yet we get _huge_ differences there. Does that make sense to anybody? (I did retry this already and the results before were similar)

Lukasa:
Are we consistently getting huge differences?

weissi:
> Are we consistently getting huge differences?

at least twice in a row... I'll try again

weissi:

weissi:
@Lukasa safe to say that yes, in three consecutive runs, we're seeing >20% diffs here. And this PR only touches B2MD but the diffs are in unrelated tests...

Lukasa:
Seems like we need to investigate these tests and the perf environment. It could be that there is wide variance in the tests (e.g. maybe the bulk of their runtime is in the allocator), or that there is wide variance in the test environment.

weissi:
@Lukasa switching the investigation over to https://github.com/apple/swift-nio/pull/1716 which is an empty PR.

weissi:
@Lukasa don't think it's the allocator or something because the results themselves are pretty stable (std diff 0.00024521498109819677)

weissi:

weissi:

weissi:

weissi:

weissi:

weissi:

weissi:

weissi:

weissi:
@Lukasa okay, something funny's going on with the binary here. Merely adding the benchmark (with no code in there) triggers the major diffs. Will need to check inlining thresholds and other code layout changes now. This'll be a while

weissi:

weissi:

weissi:

weissi:
@fabianfett / @Lukasa we can safely ignore the performance "regression". Details are the motivating example of this new doc: https://github.com/apple/swift-nio/pull/1738

weissi:

Lukasa (approved):

PR #1735, opened 2021-01-25
Add internal perf framework

Davidde94:
Add a lightweight performance framework that can be used to run benchmarks against NIO family projects.

### Motivation:

It seems extremely counter-intuitive to have to copy our performance testing code between repositories. This makes life much easier when setting up new projects, and has the added benefit of us being able to modify the framework in one place.

### Modifications:

Define a `Benchmark` protocol inside `_NIOPerformanceFramework` that can be used in NIO family projects.

### Result:

```
import _NIOPerformanceFramework

let benchmarks = <Make some benchmarks here>
try _PerformanceTester().runBenchmarks(benchmarks)
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -65,6 +65,8 @@ var targets: [PackageDescription.Target] = [
             dependencies: ["NIO", "NIOHTTP1"]),
     .target(name: "NIOCrashTester",
             dependencies: ["NIO", "NIOHTTP1", "NIOWebSocket", "NIOFoundationCompat"]),
+    .target(name: "_NIOPerformanceFramework",
+            dependencies: ["NIO"]),
     .testTarget(name: "NIOTests",
                 dependencies: ["NIO", "NIOFoundationCompat", "NIOTestUtils", "NIOConcurrencyHelpers"]),
     .testTarget(name: "NIOConcurrencyHelpersTests",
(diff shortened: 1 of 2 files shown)


Lukasa (commented):
I think we need to have a think about what this should be and how it should look.

As a first note, I see no particular reason to tie this to SwiftNIO: it prevents us versioning the API separately. I know this patch tries to get around it by making this an underscored dependency, but if we're sharing this across projects then we still have to version it _anyway_. We may as well make it a separate thing.

Davidde94:
@Lukasa note this first iteration doesn't need to be _perfect_, just better than the current method of copying and pasting code around the place, which I think/hope you'll agree isn't fantastic

Lukasa:
I think we don't necessarily agree on how bad the "copy around code" situation is.

The strong advantage of copying around the code is that there's no versioning problem. There is no risk of breaking API and no concern about behaviours changing, because each repository has a snapshot of the benchmark tool that will work for it. Given that the benchmark code is extremely small and doesn't change much, it doesn't present much of a burden.

So I think we need to make an affirmative statement of what problem, exactly, depending on this will solve.

Davidde94:
@Lukasa making it a separate repository isn't something I'm against, but copying around code is - it's a fairly common concept that files shouldn't just be copied around like this. I'm not saying it's a major issue, but it's one that's easily solvable with minimal hassle.

PR #1736, opened 2021-01-27
Add a security policy

glbrntt:
Motivation:

Security is hugely important to us and our users yet we don't provide
guidelines on how users should report vulnerabilities to us, nor any
commitments we make to resolve these issues.

Modifications:

- Add SECURITY.md detailing how to report vulnerabilities and what
  happens when one is reported.

Result:

It's easier for users to report vulnerabilities to us.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,46 @@
+# Security
+
+This document specifies the security process for the SwiftNIO project.
+
+## Disclosures
+
+### Private Disclosure Process
+
+The SwiftNIO core team asks that known and suspected vulnerabilities be
+privately and responsibly disclosed by emailing
+[cve@forums.swift.org](mailto:cve@forums.swift.org) with the [details usually
+included with bug reports][issue-template]. **Do not file a public issue.**
+
+#### When to report a vulnerability
+
+* You think you have discovered a potential security vulnerability in SwiftNIO
+  or any of the SwiftNIO projects.
+* You are unsure how a vulnerability affects SwiftNIO or any of the SwiftNIO
+  projects.
+
+#### What happens next?
+
+* A member of the team will acknowledge receipt of the report within 3
+  working days (United States). This may include a request for additional
+  information about reproducing the vulnerability.
+* We will privately inform the Swift Server Work Group ([SSWG][sswg]) of the
+  vulnerability within 10 days of the report as per their [security
+  guidelines][sswg-security].
+* Once we have identified a fix we may ask you to validate it. We aim to do this
+  within 30 days. In some cases this may not be possible, for example when the
+  vulnerability exists at the protocol level and the industry must coordinate on
+  the disclosure process.
+* If a CVE number is required, one will be requested from [MITRE][mitre]
+  providing you with full credit for the discovery.
+* We will decide on a planned release date and let you know when it is.
+* Prior to release, we will inform major dependents that a security-related
+  patch is impending.
+* Once the fix has been released we will publish a security advisory on GitHub
+  and the [SSWG][sswg] will announce the vulnerability on the [Swift
+  forums][swift-forums-sec].
+
+[issue-template]: https://github.com/apple/swift-nio/blob/main/.github/ISSUE_TEMPLATE.md
+[sswg]: https://github.com/swift-server/sswg
+[sswg-security]: https://github.com/swift-server/sswg/blob/main/process/incubation.md#security-best-practices
+[swift-forums-sec]: https://forums.swift.org/c/server/security-updates/
+[mitre]: https://cveform.mitre.org/


SECURITY.md
```diff
@@ -0,0 +1,44 @@
...
+#### When to report a vulnerability
+
+* You think you have discovered a potential security vulnerability in SwiftNIO
+  or any of the SwiftNIO projects.
+* You are unsure how a vulnerability affects SwiftNIO or any of the SwiftNIO
+  projects.
+
+#### What happens next?
+
+* A member of the team will acknowledge receipt of the report within X
```
weissi, line 24: X=3 ?

SECURITY.md
```diff
@@ -0,0 +1,44 @@
+# Security
+
+This document specifies the CVE (Common Vulnerabilities and Exposure) process
```
weissi, line 3: maybe just "security process"? We will issue CVEs but not sure if it's called the CVE process.

SECURITY.md
```diff
@@ -0,0 +1,44 @@
...
+  projects.
+
+#### What happens next?
+
+* A member of the team will acknowledge receipt of the report within X
+  working days. This may include a request for additional information about
+  reproducing the vulnerability.
+* We will inform the Swift Server Work Group ([SSWG][sswg]) of the vulnerability
+  within 10 days of the report.
+* Once we have identified a fix we may ask you to validate it. We aim to do this
```
weissi, line 29: X=30?

We should add that sometimes it's a coordinated disclosure process (example the HTTP/2 vulns) where the industry coordinates in which case we can't guarantee much.

weissi (commented):
Thank you! This looks really good, left a few comments

SECURITY.md
glbrntt: Seems totally reasonable.

SECURITY.md
glbrntt: Ah, that's a really good point.

Lukasa (approved):
LGTM as a starting point

weissi (approved):
Awesome, thank you! Let's get @PeterAdams-A / @tomerd to have a look too

SECURITY.md
```diff
@@ -0,0 +1,45 @@
...
+#### When to report a vulnerability
+
+* You think you have discovered a potential security vulnerability in SwiftNIO
+  or any of the SwiftNIO projects.
+* You are unsure how a vulnerability affects SwiftNIO or any of the SwiftNIO
+  projects.
+
+#### What happens next?
+
+* A member of the team will acknowledge receipt of the report within 3
```
PeterAdams-A, line 23: Should we specify which country?

SECURITY.md
```diff
@@ -0,0 +1,45 @@
...
+
+This document specifies the security process for the SwiftNIO project.
+
+## Disclosures
+
+### Private Disclosure Process
+
+The SwiftNIO core team asks that known and suspected vulnerabilities be
+privately and responsibly disclosed by emailing
+[cve@forums.swift.org](mailto:cve@forums.swift.org) with the [details usually
```
PeterAdams-A, line 11: I assume you've checked this address is ok for us to use.

PeterAdams-A: Who actually gets those mails?

PeterAdams-A (commented):
Generally OK I think.  Added a couple of query comments.

SECURITY.md
glbrntt: I assume @tomerd -- but not certain

tomerd: the swift core team has access to this. we are considering creating such email alias for sswg projects which may be applicable here. cc @weissi

SECURITY.md
```diff
@@ -0,0 +1,45 @@
...
+  or any of the SwiftNIO projects.
+* You are unsure how a vulnerability affects SwiftNIO or any of the SwiftNIO
+  projects.
+
+#### What happens next?
+
+* A member of the team will acknowledge receipt of the report within 3
+  working days (United Kingdom). This may include a request for additional
+  information about reproducing the vulnerability.
+* We will inform the Swift Server Work Group ([SSWG][sswg]) of the vulnerability
```
tomerd, line 26:
```suggestion
* We will privately inform the Swift Server Work Group ([SSWG][sswg]) of the vulnerability
```

SECURITY.md
```diff
@@ -0,0 +1,45 @@
...
+  within 10 days of the report.
```
tomerd, line 27: maybe add something like "as required by the sswg security guidelines" and add link to them

tomerd (approved):
very nice, added couple of comments

SECURITY.md
glbrntt: Given emails currently go to core team I switched the country to the US.

SECURITY.md
weissi: Yeah, a NIO specific alias that posts to a private forums area would be great actually I think. WDYT other NIO devs?

Lukasa: I think that's a good idea.

weissi: @tomerd is it possible to get a NIO-specific one? @Lukasa / @glbrntt are we waiting for this or can we just merge this PR (I'm in favour of merging now).

tomerd: I reckon we use the sswg-security-reports one and merge. wdyt?

glbrntt: Fine to merge now, we can always update later if/when we get a NIO specific one.

Lukasa: I am somewhat uncomfortable with keeping these two addresses aliased together but that doesn't need to block the merge of this patch.

tomerd: this should be changes to `sswg-security-reports` (SSWG) which has more relevant people on it than `cve@forums.swift.org` (swift core team)

PR #1738, opened 2021-01-28
docs: advanced performance analysis guide

weissi:
Motivation:

Sometimes, advances performance analysis is required and we didn't have
documents describing this.

Modifications:

Add a document describing advanced performance analysis with `perf`.

Result:

More guides.

### DIFF
--- a/docs/advanced-performance-analysis.md
+++ b/docs/advanced-performance-analysis.md
@@ -1,0 +1,227 @@
+# Advanced performance analysis for CPU-bound programs
+
+Most performance problems can be handily debugged with [Instruments, `perf`, or FlameGraphs](https://github.com/swift-server/guides/blob/main/performance.md). At times however, there are changes in performance that are very hard to understand and sometimes very counter intuitive.
+
+## Motivating example (feel free to skip)
+A motivating example is [this pull request](https://github.com/apple/swift-nio/pull/1733) which is supposed to improve the performance of `ByteToMessageDecoder`s in very specific cases. The expectation would be that most performance tests remain the same and everything that uses a `ByteToMessageDecoder` either remains the same or becomes a little faster. Of course, all the low-level microbenchmarks that don't even use `ByteToMessageDecoder`s should be totally unaffected by this change.
+
+For the most part that was the case, except for two benchmarks in SwiftNIO's performance test suite for which the [results were](https://github.com/apple/swift-nio/pull/1733#issuecomment-766917189) very unexpected:
+
+<table border="1">
+  <tr>
+    <td>benchmark name</td>
+    <td>run time with PR</td>
+    <td>run time before PR</td>
+    <td>which is faster</td>
+    <td>difference</td>
+  </tr>
+  <tr>
+    <td>circular_buffer_into_byte_buffer_1kb</td>
+    <td>0.05089386</td>
+    <td>0.041163118</td>
+    <td>previous</td>
+    <td>23%</td>
+  </tr>
+  <tr>
+    <td>circular_buffer_into_byte_buffer_1mb</td>
+    <td>0.094321349</td>
+    <td>0.082286637</td>
+    <td>previous</td>
+    <td>14%</td>
+  </tr>
+</table>
+
+This means that for whatever reason, the totally unrelated benchmarks `circular_buffer_into_byte_buffer_*` got _slower_ by 23% and 14% respectively (measured in wall clock seconds) despite using none of the changed code.
+
+## Wall clock time
+
+To start out, we need to verify if the benchmark results are stable regarding wall clock time. If there is too much variation, then it is likely that too much I/O or allocations etc are involved. If a benchmark is unstable because of I/O or allocations, there are better tools available than what is discussed here, for example `strace`, Instrument's "Allocations" instrument, or Xcode's Memory Graph Debugger. This document solely discusses _CPU-bound_ and mostly single-threaded performance analysis, ie. you should see one CPU core pegged at 100% for the whole duration of the benchmark.
+
+## Executed CPU instructions
+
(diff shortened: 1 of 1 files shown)


docs/advanced-performance-analysis.md
```diff
@@ -0,0 +1,226 @@
...
+        5927358294      instructions                                                
+         308782310      uops_executed.stall_cycles                                   
+          14159547      resource_stalls.any                                         
+         288072686      cycle_activity.stalls_mem_any                                   
+
+       0.510635133 seconds time elapsed
+
+```
+
+What we can immediately see is that `resource_stalls.any` is pretty much the same in both cases, but `uops_executed.stall_cycles` and `cycle_activity.stalls_mem_any` are about an order of magnitude higher for the slow case. Remember, this is with the number of instructions being the same.
```
Lukasa, line 144:
```suggestion
What we can immediately see is that `resource_stalls.any` is pretty much the same in both cases, but `uops_executed.stall_cycles` and `cycle_activity.stalls_mem_any` are about an order of magnitude higher for the slow case. Remember, this is with the number of instructions being almost the same.
```

docs/advanced-performance-analysis.md
```diff
@@ -0,0 +1,226 @@
...
+
+So in this case, we would have identified that the slowdown comes from us getting a lot of "stall cycles" in the CPU. Those are cycles where the CPU was free to run some (micro-)instructions but could not. Unfortunately, from this picture alone, it is unclear _why_ exactly the CPU couldn't actually run some more (micro-)operations. Learned only a little so far, we need to dig deeper.
```
Lukasa, line 146:
```suggestion
So in this case, we would have identified that the slowdown comes from us getting a lot of "stall cycles" in the CPU. Those are cycles where the CPU was free to run some (micro-)instructions but could not. Unfortunately, from this picture alone, it is unclear _why_ exactly the CPU couldn't actually run some more (micro-)operations. We’ve only learned a little so far, we need to dig deeper.
```

docs/advanced-performance-analysis.md
```diff
@@ -0,0 +1,226 @@
...
...ering what `dsb2mite_switches.penalty_cycles` is, a web search revealed a nicely written [blogpost](https://easyperf.net/blog/2018/01/18/Code_alignment_issues) which discusses something very similar. Bottom line is that modern Intel CPUs have a MITE (Micro-operation Translation Engine) which seems to read the instruction stream in 16 byte chunks. So if the alignment of your code is changed (for example by adding/removing an unrelated function to/from your binary) you may incur a penalty if your instructions aren't aligned by 16 bytes (remember, Intel instructions are variable length). This does sound scary (and for some code it is) but the good news is that this should _very rarely_ matter outside of micro-benchmarks. If your whole run time is doing exactly what that micro-benchmark is doing, you should probably care. But most real-world code does much more and more diverse work than some arbitrary micro-benchmarks. So in this particular example we can safely ignore the "regression" which was just unlucky placement of the hot function in the binary. Nevertheless it was important to understand why the regression happened so we could safely decide it does not matter in this instance.
```
Lukasa, line 174: Do we really want these square brackets?

docs/advanced-performance-analysis.md
```diff
@@ -0,0 +1,226 @@
...
+
+Another place where instruction counts come in very handy is when you see this maybe impossible-sounding situation: the run time varies significantly but the number of executed instructions is _the same_. The motivating example is exactly this situation: The code executed is the _same_, confirmed by manually checking the generated assembly for the code run as well as checking that the number of instructions is almost equal. Because manual assembly analysis can easily go wrong, it's advisable to compare the number of executed instructions even if you convinced yourself that the executed code is the same. Of course, there are other options like single-stepping in a debugger but that gets tedious quickly. If the executed code and the inputs are the same, then the number of executed instructions should also be the same (ie. < 0.01% variance).
+
+## CPU stalls
```
glbrntt, line 83: How did you make the jump from runtime difference and the same instruction count to CPU stalls? Was there anything else you looked at at this point before looking at stalls?

glbrntt (approved):
Nice writeup!

Lukasa (approved):
Looks good, a few notes in the diff that you can take or leave.

docs/advanced-performance-analysis.md
weissi: @glbrntt Thanks! I'll flesh that out a little. But in general: If I know that the instructions run are the same but one is faster, then it needs to be some kind of stall/penalty really. But I'll add some text.

docs/advanced-performance-analysis.md
```diff
@@ -0,0 +1,226 @@
...
+First, we can confirm that it is nicely CPU bound (`1.001 CPUs utilized`).
+
+We can also see that we executed about 5.9B instructions during this benchmark run. The number of instructions executed itself is not very meaningful but it can be very useful when comparing between versions of the same code. For example, let's assume we changed some code and want to confirm in a benchmark that indeed we have made it faster, then it can be very useful to see that we have also reduced the number of instructions.
+
+With `perf` you could run `perf stat -e instructions -- ./benchmark-before-change` and `perf stat -e instructions -- ./benchmark-after-change` and confirm that:
+
+1. The runtime (_time elapsed_) has gone down and
+2. the number of instructions executed has gone down too.
+
+Note that the time it takes to run an instruction varies wildly so it is in theory possible to reduce the number of instructions but make performance actually worse. An example would be if you traded a lot of simple arithmetic for memory accesses. Said that, in most cases where you reduced the number of instructions significantly, you will also see an improvement in the actual runtime.
```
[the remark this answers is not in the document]
weissi: Good point! I'll add a quick sentence

docs/advanced-performance-analysis.md
weissi: no

weissi:
thanks very much @glbrntt  & @Lukasa I applied everything you suggested.

PR #1739, opened 2021-01-28
align all functions to make micro benchmarks more stable

weissi:
Motivation:

After chatting to the Swift perf team, they thought it may be a good
idea to align all functions in our performance tests which may make the
micro benchmarks more stable.

Modifications:

Align all functions/blocks.

Result:

Hopefully more stable micro benchmarks.

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -40,7 +40,7 @@ services:
 
   performance-test:
     <<: *common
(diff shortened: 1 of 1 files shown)


weissi:

Lukasa (approved):

Lukasa:
As I’d expect the immediate effect of this change is to make all the benchmarks slower. That’s likely the result of all the extra `nop` slides floating around. That’s ok, we don’t really care about the absolute numbers, just the hope that this makes relative numbers more stable.

PR #1740, opened 2021-01-31
Host header is requires for HTTP/1.1

Bouke:
The `Host` header is required for HTTP/1.1 and standard-compliant servers will refuse to connect without it, causing new users of this library unneeded debugging frustration.

### Motivation:

I tried to connect to a websocket through NIO. The current example (NIOWebSocketClient) contains a lot of boilerplate which is hard to reason about for new users of this library. Assessing the correctness of this boilerplate while getting to a bare-minimum "just works" is hard to do. Most standard-compliant webservers will simply refuse to connect without the `Host` header. The library is not giving any indication of the problem, and eventually I had to compare the http traffic between this library and the equivalent in python. That's when I finally discovered that the example lacks the required `Host` header. Adding that, I finally got to a working websocket.

### Modifications:

This adds the `Host` header in the example. The host header ideally would be taken from the user's input on the command-line. If that's desired, I'd like to get some pointers on how to pass that value along.

### Result:

The required `Host` header will be present.

### DIFF
--- a/Sources/NIOWebSocketClient/main.swift
+++ b/Sources/NIOWebSocketClient/main.swift
@@ -155,33 +155,6 @@ private final class WebSocketPingPongHandler: ChannelInboundHandler {
     }
 }
 
-let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-let bootstrap = ClientBootstrap(group: group)
-    // Enable SO_REUSEADDR.
-    .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
-    .channelInitializer { channel in
-        
-        let httpHandler = HTTPInitialRequestHandler()
-        
-        let websocketUpgrader = NIOWebSocketClientUpgrader(requestKey: "OfS0wDaT5NoxF2gqm7Zj2YtetzM=",
-                                                           upgradePipelineHandler: { (channel: Channel, _: HTTPResponseHead) in
-            channel.pipeline.addHandler(WebSocketPingPongHandler())
-        })
-        
-        let config: NIOHTTPClientUpgradeConfiguration = (
-            upgraders: [ websocketUpgrader ],
-            completionHandler: { _ in
-                channel.pipeline.removeHandler(httpHandler, promise: nil)
-        })
-
-        return channel.pipeline.addHTTPClientHandlers(withClientUpgrade: config).flatMap {
-            channel.pipeline.addHandler(httpHandler)
-        }
-}
-defer {
-    try! group.syncShutdownGracefully()
-}
-
 // First argument is the program path
 let arguments = CommandLine.arguments
 let arg1 = arguments.dropFirst().first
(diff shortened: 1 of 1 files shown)


Bouke:
Can somebody stop the bots from taking over?

Davidde94 (approved):
Looks good, thanks for the code and feedback!

Davidde94:

Davidde94:
@swift-server-bot add to whitelist

Davidde94:

Lukasa (commented):
Thanks for this! I agree that it would be better if we got the host from the arguments provided on the command-line.

The way to plumb this through would be to have the `HTTPInitialRequestHandler` take the target in its constructor.  To do that we'd want to move the block in lines 157 to 180 (the `ClientBootstrap` block) underneath the block in lines 183 to 209 (which sets `connectTarget`). At that point we know where we're connecting. The big switch in lines 196 to 209 can also set a string value for the `Host` header, which we can then pass into the constructor of the channel handler.

Does that make sense?

Bouke:
Thanks @Lukasa for the points, that sure makes sense. I've updated the PR to include these changes. For the unix path connection target I don't know what `Host` header to pass, so it's now only doing that for an IP connection target. Let me know what you think, if changes are needed.

Sources/NIOWebSocketClient/main.swift
```diff
@@ -208,6 +190,33 @@ default:
     connectTarget = .ip(host: defaultHost, port: defaultPort)
 }
 
+let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+let bootstrap = ClientBootstrap(group: group)
+    // Enable SO_REUSEADDR.
+    .channelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
+    .channelInitializer { channel in
+
+        let httpHandler = HTTPInitialRequestHandler()
```
Lukasa, line 199: I think the `init` needs to be updated here to actually take the target.

Lukasa (commented):
Looks great, and I think that doing best-effort for now is fine. UDS is definitely unclear and would require more command line arguments which just isn't worth the hassle right now. One small fix and I think this will probably be good to go.

Lukasa:

Sources/NIOWebSocketClient/main.swift
Bouke: Ah yes, that got lost when making the commits. Fixed now.

Lukasa:

Lukasa:
Looks like you'll need to hoist the definition of `ConnectTo` to the top of the file. `main.swift` is kinda weird like that, and so unlike the rest of Swift you need definitions to be "in order".

Bouke:
> Looks like you'll need to hoist the definition of `ConnectTo` to the top of the file. `main.swift` is kinda weird like that, and so unlike the rest of Swift you need definitions to be "in order".

Is that new? My Xcode has no problems building `main.swift` with the "out of order" definitions. Anyway, changed.

Lukasa:
I actually think it's _old_. We run on all versions back to 5.0.

Lukasa:

Lukasa (approved):
Great patch, thanks!

PR #1741, opened 2021-02-05
Add a synchronous pipeline view

glbrntt:
Motivation:

`ChannelPipeline` is explicitly thread-safe, any of the operations may
be called from outside of the channel's event loop. However, there are
often cases where it is known that the caller be on the right event
loop, and an asynchronous API is unnecessary.

In some cases -- such as when a pipeline is configured dynamically and
handlers are added from the 'channelRead' implementation of one handler
-- it forces the caller to write code that they might not actually need:
such as buffering events which may happen before the future completes.
This is unnecessary complexity when the caller knows that they must
already be on an event loop.

Modifications:

- Add a 'SynchronousVew' to the 'ChannelPipeline' which is only
  available to callers via 'withSynchronousViewIfAvailable'.
- An error is thrown if the caller is not on the event loop.
- The synchronous view supports: adding a handler, adding multiple
  handlers, retrieving a context via various predicates and retrieving a
  handler of a given type.
- Some of the operations in 'ChannelPipeline' were refactored to have an
  explicitly synchronous version, asynchronous versions complete their
  promise based on the result of these calls.
- Various minor documentation fixes and addition of 'self' where it was
  not used explicitly.

Result:

Users can perform synchronous operations on the 'ChannelPipeline' if
they know they are on the right event loop.

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -163,7 +163,7 @@ public final class ChannelPipeline: ChannelInvoker {
     /// Add a `ChannelHandler` to the `ChannelPipeline`.
     ///
     /// - parameters:
-    ///     - name: the name to use for the `ChannelHandler` when its added. If none is specified it will generate a name.
+    ///     - name: the name to use for the `ChannelHandler` when it's added. If none is specified it will generate a name.
     ///     - handler: the `ChannelHandler` to add
     ///     - position: The position in the `ChannelPipeline` to add `handler`. Defaults to `.last`.
     /// - returns: the `EventLoopFuture` which will be notified once the `ChannelHandler` was added.
@@ -172,10 +172,10 @@ public final class ChannelPipeline: ChannelInvoker {
                            position: ChannelPipeline.Position = .last) -> EventLoopFuture<Void> {
         let promise = self.eventLoop.makePromise(of: Void.self)
         if self.eventLoop.inEventLoop {
-            promise.completeWith(self._add(handler, name: name, position: position))
+            promise.completeWith(self.addHandlerSync(handler, name: name, position: position))
         } else {
             self.eventLoop.execute {
-                promise.completeWith(self._add(handler, name: name, position: position))
+                promise.completeWith(self.addHandlerSync(handler, name: name, position: position))
             }
         }
 
@@ -187,13 +187,13 @@ public final class ChannelPipeline: ChannelInvoker {
     /// May only be called from on the event loop.
     ///
     /// - parameters:
-    ///     - name: the name to use for the `ChannelHandler` when its added. If none is specified it will generate a name.
     ///     - handler: the `ChannelHandler` to add
+    ///     - name: the name to use for the `ChannelHandler` when it's added. If none is specified a name will be generated.
     ///     - position: The position in the `ChannelPipeline` to add `handler`. Defaults to `.last`.
     /// - returns: the result of adding this handler - either success or failure with an error code if this could not be completed.
-    private func _add(_ handler: ChannelHandler,
-                      name: String? = nil,
-                      position: ChannelPipeline.Position = .last) -> Result<Void, Error> {
+    fileprivate func addHandlerSync(_ handler: ChannelHandler,
+                                    name: String? = nil,
+                                    position: ChannelPipeline.Position = .last) -> Result<Void, Error> {
         self.eventLoop.assertInEventLoop()
 
         if self.destroyed {
@@ -278,7 +278,7 @@ public final class ChannelPipeline: ChannelInvoker {
                       operation: (ChannelHandlerContext, ChannelHandlerContext) -> Void) -> Result<Void, Error> {
         self.eventLoop.assertInEventLoop()
 
-        if destroyed {
+        if self.destroyed {
             return .failure(ChannelError.ioOnClosedChannel)
         }
 
@@ -366,9 +366,7 @@ public final class ChannelPipeline: ChannelInvoker {
     ///     - handler: the `ChannelHandler` to remove.
     ///     - promise: An `EventLoopPromise` that will complete when the `ChannelHandler` is removed.
     public func removeHandler(_ handler: RemovableChannelHandler, promise: EventLoopPromise<Void>?) {
-        let contextFuture = self.context0 {
-            return $0.handler === handler
-        }.map { context in
+        let contextFuture = self.context(handler: handler).map { context in
             self.removeHandler(context: context, promise: promise)
         }
 
@@ -381,9 +379,7 @@ public final class ChannelPipeline: ChannelInvoker {
     ///     - name: the name that was used to add the `ChannelHandler` to the `ChannelPipeline` before.
     ///     - promise: An `EventLoopPromise` that will complete when the `ChannelHandler` is removed.
     public func removeHandler(name: String, promise: EventLoopPromise<Void>?) {
-        let contextFuture = self.context0 {
-            $0.name == name
-        }.map { context in
+        let contextFuture = self.context(name: name).map { context in
             self.removeHandler(context: context, promise: promise)
         }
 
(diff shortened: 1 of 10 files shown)


Davidde94 (approved):
I had a good look through and don't see any obvious issue, but others may spot something my untrained eyes didn't see.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,163 @@ extension ChannelPipeline {
...
+    }
+
+    /// Invokes a closure with a synchronous view of this `ChannelPipeline` if this function is
+    /// called from the `EventLoop`.
+    ///
+    /// - Important: Users *must not* escape the `SynchronousView` from the closure.
+    /// - Parameter closure: A closure to operate on a synchronous view of this `ChannelPipeline`.
+    /// - Throws: `NIONotOnEventLoopError` if this function is not invoked on the event loop for
+    ///     this `ChannelPipeline`. Re-throws any error thrown from `body`.
+    public func withSynchronousViewIfAvailable(_ closure: (SynchronousView) throws -> Void) throws {
```
weissi, line 1095: - is all the extra syntax of the `with*` block worth it to make escaping the "SynchronousView" a little harder? I probably think it's not but can be convinced otherwise.
- I'd probably go for `channel.pipeline.syncOperations -> ChannelPipeline.SyncOperations` or something. Given that all operations throw _anyway_, why don't we just make them throw if you call the `SyncOperations` from off the EL? Then our interface is still thread-safe :). [problem: see below]

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,163 @@ extension ChannelPipeline {
...
+        if self.eventLoop.inEventLoop {
+            try closure(SynchronousView(_pipeline: self))
```
weissi, line 1097: So technically, this is incorrect. `self.eventLoop.inEventLoop` must only be used as an _optimisation_, not as a decider. In NIOTS, this may say `false` even if you're on it.

Tests/NIOTests/ChannelPipelineTest.swift
```diff
@@ -1294,6 +1294,141 @@ class ChannelPipelineTest: XCTestCase {
         XCTAssertNoThrow(try removal1Future.wait())
         XCTAssertNoThrow(XCTAssertTrue(try channel.finish().isClean))
     }
+
```
weissi, line 1297: I think we should add an alloc counter test that shows that it's actually cheaper :)

weissi (changes requested):
This looks really great, almost can go in as is. Just a few questions and a request for test :)

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,163 @@ extension ChannelPipeline {
...
+
+    /// Synchronously adds a sequence of `ChannelHandlers` to the pipeline at the given position.
+    ///
+    /// - Important: Must be called on the `EventLoop`.
+    /// - Parameters:
+    ///   - handlers: A sequence of handlers to add.
+    ///   - position: The position in the `ChannelPipeline` to add the handlers.
+    /// - Returns: A result representing whether the handlers were added or not.
+    @inlinable
+    internal func _addHandlersSync<Handlers: Sequence>(_ handlers: Handlers,
```
Lukasa, line 989: This `@inlinable` is pointless: the function is never called from a public context. We can strip it and make this function `private` and we'll be getting the same effect.

Sources/NIO/ChannelPipeline.swift
Lukasa: I'm inclined to say that we should get rid of the ceremony of getting one of these, and then just force all of the methods to trap if they're used off-loop.

glbrntt: The extra syntax is pretty minimal and -- I think -- makes it clearer that there are conditions to executing them synchronously (you have to think "can I really do this here?"). You also avoid having to check if you're on the event loop for each operation (and the caller only has to check whether we threw because we weren't on the event loop in one place). 

I'm not opposed `syncOperations` per se but I don't think it's any better.

weissi: @glbrntt see my comment below, you can't actually check if it's safe or not (without trapping anyway).

glbrntt: I saw, was just thinking about a solution! Seems like it's `syncOperations` and trapping :)

Tests/NIOTests/ChannelPipelineTest.swift
glbrntt: Will do, wanted to get consensus on the changes first.

Sources/NIO/ChannelPipeline.swift
weissi: @glbrntt or maybe `channel.pipeline.onEventLoop.addHandler` ?

Tests/NIOTests/ChannelPipelineTest.swift
glbrntt: Added a sync counterpart to the 'add two handlers 1000 times' and async and sync versions of 'get a typed handler from the pipeline'. Reckon we need any more?

Sources/NIO/ChannelPipeline.swift
glbrntt: I prefer `syncOperations` but fine with either.

Tests/NIOTests/ChannelPipelineTest.swift
weissi: no, that's fine, thank you.

weissi (approved):
Awesome, thank you! LGTM

Sources/NIO/ChannelPipeline.swift
```diff
@@ -187,13 +187,14 @@ public final class ChannelPipeline: ChannelInvoker {
...
     /// - parameters:
-    ///     - name: the name to use for the `ChannelHandler` when its added. If none is specified it will generate a name.
     ///     - handler: the `ChannelHandler` to add
+    ///     - name: the name to use for the `ChannelHandler` when it's added. If none is specified a name will be generated.
     ///     - position: The position in the `ChannelPipeline` to add `handler`. Defaults to `.last`.
     /// - returns: the result of adding this handler - either success or failure with an error code if this could not be completed.
-    private func _add(_ handler: ChannelHandler,
-                      name: String? = nil,
-                      position: ChannelPipeline.Position = .last) -> Result<Void, Error> {
+    @usableFromInline // should be fileprivate
```
Lukasa, line 194: This `usableFromInline` is spurious, and so this function actually _can_ be fileprivate.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -440,29 +477,41 @@ public final class ChannelPipeline: ChannelInvoker {
     ///     - handlerType: The type of the handler to search for.
     /// - returns: the `EventLoopFuture` which will be notified once the the operation completes.
     public func context<Handler: ChannelHandler>(handlerType: Handler.Type) -> EventLoopFuture<ChannelHandlerContext> {
```
Lukasa, line 479: This function should be `@inlinable`, or the other inlinability annotation won't help us at all.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -885,53 +934,17 @@ extension ChannelPipeline {
...
-            }
+        let promise = self.eventLoop.makePromise(of: Void.self)
 
-            if self.eventLoop.inEventLoop {
-                addAllHandlersAndComplete()
-            } else {
-                self.eventLoop.execute {
-                    addAllHandlersAndComplete()
-                }
+        if self.inEventLoop {
```
Lukasa, line 939: There's some inconsistency in the use of this computed property or in asking the event loop. It'd be nice to be consistent.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,141 @@ extension ChannelPipeline {
...
+
+    /// Synchronously adds the provided `ChannelHandler`s to the pipeline in the order given, taking
+    /// account of the behaviour of `ChannelHandler.add(first:)`.
+    ///
+    /// - Important: Must be called on the `EventLoop`.
+    /// - Parameters:
+    ///   - handlers: The array of `ChannelHandler`s to add.
+    ///   - position: The position in the `ChannelPipeline` to add the handlers.
+    /// - Returns: A result representing whether the handlers were added or not.
+    internal func addHandlersSync(_ handlers: [ChannelHandler],
```
Lukasa, line 971: Should this be `_addHandlersSync`, with the leading underscore to match `_addHandlerSync`?

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,141 @@ extension ChannelPipeline {
...
+        public func context(name: String) throws -> ChannelHandlerContext {
+            return try self.pipeline.contextSync(name: name).get()
+        }
+
+        /// Returns the `ChannelHandlerContext` for the handler of given type, if one exists.
+        ///
+        /// - Important: This *must* be called on the event loop.
+        /// - Parameter name: The name of the handler whose context is being fetched.
+        /// - Returns: The `ChannelHandlerContext` associated with the handler.
+        public func context<Handler: ChannelHandler>(handlerType: Handler.Type) throws -> ChannelHandlerContext {
```
Lukasa, line 1078: This method should be `@inlinable`.

Sources/NIO/ChannelPipeline.swift
```diff
@@ -944,7 +957,141 @@ extension ChannelPipeline {
...
+            return try self.pipeline._contextSync(handlerType: handlerType).get()
+        }
+
+        /// Returns the `ChannelHandler` of the given type from the `ChannelPipeline`, if it exists.
+        ///
+        /// - Important: This *must* be called on the event loop.
+        /// - Returns: A `ChannelHandler` of the given type if one exists in the `ChannelPipeline`.
+        public func handler<Handler: ChannelHandler>(type _: Handler.Type) throws -> Handler {
```
Lukasa, line 1086: And this one.

Sources/NIO/ChannelPipeline.swift
glbrntt: Oof, good catch.

Sources/NIO/ChannelPipeline.swift
glbrntt: I think `_addHandlerSync` (previously `_add`, not sure why it was underscored) should be `addHandlerSync` and this should be `fileprivate func addHandlersSync`

Lukasa (approved):
Great, LGTM.

PR #1743, opened 2021-02-08
alloc counters: Use atomic globals instead of thread-locals

weissi:
Motivation:

The alloc counters need to store the pointer to the original libc
implmentations of the hooked functions somewhere. Previously, we would
store them in thread locals. That works fine but creates quite some
overhead (and allocations) per thread (to do dlsym on every thread).

Modifications:

Instead, we now store the libc function pointers in (atomic) globals so
we need to only resolve each function once, no matter how many threads
we use.

Result:

Faster, and more accurate.

### DIFF
--- a/IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-unix.c
+++ b/IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-unix.c
@@ -37,12 +37,20 @@ static _Atomic ptrdiff_t g_recursive_malloc_next_free_ptr = ATOMIC_VAR_INIT(0);
 
 #define LIBC_SYMBOL(_fun) "" # _fun
 
+/* Some thread-local flags we use to check if we're recursively in a hooked function. */
 static __thread bool g_in_malloc = false;
 static __thread bool g_in_realloc = false;
 static __thread bool g_in_free = false;
-static __thread void *(*g_libc_malloc)(size_t) = NULL;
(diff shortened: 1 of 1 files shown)


IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-unix.c
```diff
@@ -85,25 +93,37 @@ static void recursive_free(void *ptr) {
...
 
-/* on other UNIX systems this is slightly harder. Basically we see if we already
- * have a thread local variable that is a pointer to the original libc function.
- * If yes, easy, call it. If no, we need to resolve it which we do by using
- * dlsym. There's only one slight problem: dlsym might call back into the
- * function we're just trying to resolve (it does call malloc). In that case
- * we need to emulate that function (named recursive_*). But that's all then.
+/* on other UNIX systems this is slightly harder. Basically we already have the
+ * pointer to the original libc function in our respective global (atomic)
+ * variable.  If yes, easy, call it. If no, we need to resolve it which we do by
```
Lukasa, line 98: Silly grammar nit but: you removed an "if" and now this paragraph doesn't read properly.

Lukasa (approved):
Basically fine.

IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-unix.c
weissi: Thanks, I rewrote that comment.

PR #1746, opened 2021-02-09
Add a new overloading function completeWith to EventLoopPromise class.

Hailong:
Add a new overloading function completeWith to EventLoopPromise class.

### Motivation:

- Replace global helper function executeAndComplete() with class method for better encapsulation.
- Get rid of the repeated pattern of checking whether the event loop is "current".

### Modifications:

- Removed global function executeAndComplete().
- Added a new overloading of the completeWith() method to EventLoopPromise class.

### Result:

Simplified the consumers of this functionality in code. And the new added overloading supports crossing event loop as well.

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -543,13 +543,9 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
     }
 
     public final func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> EventLoopFuture<Void> {
-        if eventLoop.inEventLoop {
-            let promise = eventLoop.makePromise(of: Void.self)
-            executeAndComplete(promise) { try self.setOption0(option, value: value) }
-            return promise.futureResult
-        } else {
-            return eventLoop.submit { try self.setOption0(option, value: value) }
-        }
+        let promise = eventLoop.makePromise(of: Void.self)
+        promise.completeWithClosure { try self.setOption0(option, value: value) }
+        return promise.futureResult
     }
 
     func setOption0<Option: ChannelOption>(_ option: Option, value: Option.Value) throws {
@@ -646,8 +642,8 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             return
         }
 
-        executeAndComplete(promise) {
-            try socket.bind(to: address)
+        promise?.completeWithClosure {
+            try self.socket.bind(to: address)
             self.updateCachedAddressesFromSocket(updateRemote: false)
         }
     }
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -217,6 +217,34 @@ public struct EventLoopPromise<Value> {
         self._resolve(value: result)
     }
 
+    /// Execute the given closure and complete the promise.
+    ///
+    /// The given closure is executed _synchronously_ or _asynchronously_ depending on whether
+    /// the current `NIOThread` is tied to the `EventLoop` that was used to create the promise.
+    ///
+    /// WARNING: You _must not_ call into any blocking code in the given closure if it is executed synchronously.
+    ///
+    /// - parameters:
+    ///     - body: The closure to execute.
+    @inlinable
+    public func completeWithClosure(_ body: @escaping () throws -> Value) {
+        func executeAndComplete() {
+            do {
+                self.succeed(try body())
+            } catch let e {
+                self.fail(e)
+            }
+        }
+
+        if self.futureResult.eventLoop.inEventLoop {
+            executeAndComplete()
+        } else {
+            self.futureResult.eventLoop.execute {
+                executeAndComplete()
+            }
+        }
+    }
+
     /// Fire the associated `EventLoopFuture` on the appropriate event loop.
     ///
     /// This method provides the primary difference between the `EventLoopPromise` and most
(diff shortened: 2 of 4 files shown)


Lukasa:
Thanks for this patch! Out of curiosity, is it necessary to hop the loop? `EventLoopPromise` is entirely thread-safe and can absolutely be executed on any thread. Unless you have a specific reason to want to do the loop hopping, I think we can safely remove that from this.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -217,6 +217,32 @@ public struct EventLoopPromise<Value> {
...
 
+    /// Execute the given function and complete the promise.
+    ///
+    /// Specifically, the function is executed synchronously on the `EventLoop`
+    /// that was used to create the promise.
+    ///
+    /// - parameters:
+    ///     - body: The function to execute.
+    @inlinable
+    public func completeWith(_ body: @escaping () throws -> Value) {
```
Lukasa, line 228: A minor note: overloading based on argument type is a bit tricky. It might be nice to give this a different name: `completeWithBlock` or similar might be good.

Hailong: Thank you for the review!! `completeWithBlock` does look better, I just change it accordingly.

Hailong:
> Thanks for this patch! Out of curiosity, is it necessary to hop the loop? `EventLoopPromise` is entirely thread-safe and can absolutely be executed on any thread. Unless you have a specific reason to want to do the loop hopping, I think we can safely remove that from this.

Lukasa:
The original logic of the exact function you're replacing didn't do `if eventLoop.inEventLoop` and then conditionally call `execute`. This is not really necessary, and can be removed.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -217,6 +217,32 @@ public struct EventLoopPromise<Value> {
...
 
+    /// Execute the given function and complete the promise.
+    ///
+    /// Specifically, the function is executed synchronously on the `EventLoop`
+    /// that was used to create the promise.
+    ///
+    /// - parameters:
+    ///     - body: The function to execute.
+    @inlinable
+    public func completeWithBlock(_ body: @escaping () throws -> Value) {
```
weissi, line 228: So despite not being a fan of overloading, in this case I think we should probably overload the `completeWith` again for throwing functions. "Block" isn't really a word that's used in Swift.

Apart from the naming, the only thing I think we need here is a big, fat warning that you must not call into any blocking code here.

weissi (changes requested):
Thank you! That looks like a good idea to me. I left a comment with some minor feedback, mostly naming & docs :)

Sources/NIO/EventLoopFuture.swift
Lukasa: Hah, that's the exact opposite code review feedback I gave. :wink:

Is there a compromise position we can get to here?

weissi: @Lukasa haha oops. So if we find a good name, I'm (always) happy not to overload. But `completeWithClosure` doesn't sound great, `completeWithBlock` sounds like ObjC. But the real reason that tipped me into "let's overload here" is that the other two versions are spelled `completeWith` and are overloaded, so I thought let's follow prior art.

But really, I'm not crazy strong here just didn't like "block" (ObjC and may even be confused with "blocking") and then looked at the other two.

Lukasa: Hmm, they are overloaded, but they're overloaded with non-confusing concrete types. Overloading with closures always makes me feel a bit more nervous. But ok, I'm fine with going back to `completeWith` I guess. Sorry @Hailong!

weissi: @Lukasa I agree that closures are extra troublesome...

weissi: `Result` uses `Result(catching: () throws -> T)` which is also weird ;)

weissi: Maybe `completeWithClosure` isn't too bad? @Lukasa / @Hailong WDYT?

Lukasa: WFM

Hailong: Thank you both! I have just changed it to `completeWithClosure`. I also reworded the doc a little bit and added the warning message, please let me know if anything is missing. 😄

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -646,8 +642,8 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             return
         }
 
-        executeAndComplete(promise) {
-            try socket.bind(to: address)
+        promise?.completeWithClosure {
```
Lukasa, line 645: This is a subtle behavioural change and I'd be surprised if the tests don't pick it up: we _must_ execute this closure, but the code here will not execute if the promise is nil. I think the original variant is still necessary here.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -217,6 +217,34 @@ public struct EventLoopPromise<Value> {
...
+    public func completeWithClosure(_ body: @escaping () throws -> Value) {
+        func executeAndComplete() {
+            do {
+                self.succeed(try body())
+            } catch let e {
+                self.fail(e)
+            }
+        }
+
+        if self.futureResult.eventLoop.inEventLoop {
```
Lukasa, line 239: I'm still unsure whether we need this. This seems to me to be conflating two different actions: one is promise completion, one is where the closure executes. I think I'd like them to be separate.

Lukasa (changes requested):

Sources/NIO/BaseSocketChannel.swift
Hailong: Yep, it's a subtle change. Let me take a look.

Sources/NIO/EventLoopFuture.swift
Hailong: It's a good point. I have just created another PR#1747, which removes the conflating, and also avoids the subtle behaviour change. Feel free let me know if you have any comments, thanks! 😄

Lukasa:
I believe we have public API to do this now.

PR #1747, opened 2021-02-11
Move the global helper function executeAndComplete() into EventLoop class.

Hailong:
Move the global helper function executeAndComplete() into EventLoop class.

### Motivation:

- For better encapsulation.
- From either the caller's prospective or their internal implementation, executeAndComplete() is more likely an alternative to the submit() method of EventLoop class.

### Modifications:

- Moved global function executeAndComplete() to EventLoop class.

### Result:

Now it's looks more natural to use executeAndComplete() from the callers' prospective.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -229,6 +229,9 @@ public protocol EventLoop: EventLoopGroup {
     /// Submit a given task to be executed by the `EventLoop`
     func execute(_ task: @escaping () -> Void)
 
+    /// Execute the given function and synchronously complete the given `EventLoopPromise` (if not `nil`).
+    func executeAndComplete<T>(_ promise: EventLoopPromise<T>?, _ body: () throws -> T)
+
     /// Submit a given task to be executed by the `EventLoop`. Once the execution is complete the returned `EventLoopFuture` is notified.
     ///
     /// - parameters:
@@ -499,6 +502,17 @@ extension NIODeadline {
 }
 
 extension EventLoop {
+    /// Execute the given function and synchronously complete the given `EventLoopPromise` (if not `nil`).
+    @inlinable
+    public func executeAndComplete<T>(_ promise: EventLoopPromise<T>?, _ body: () throws -> T) {
+        do {
+            let result = try body()
+            promise?.succeed(result)
+        } catch let err {
+            promise?.fail(err)
+        }
+    }
+
     /// Submit `task` to be run on this `EventLoop`.
     ///
     /// The returned `EventLoopFuture` will be completed when `task` has finished running. It will be succeeded with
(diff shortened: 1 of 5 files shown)


Hailong:
Hi @Lukasa  @weissi, here is an alternative to my previous PR#1746. It's based on the thoughts of answering the two concerns raised by @Lukasa in my previous PR: 1. when the promise is nil, 2. conflating two different actions. So I'm trying to make this one simple and straightforward.

This idea is also based on an diagram of Netty that I found [here](https://drek4537l1klr.cloudfront.net/maurer/Figures/07fig03_alt.jpg), so we might be able to consider to simplify the `if eventLoop.inEventLoop` pattern base on this change later on.

Sources/NIO/EventLoop.swift
```diff
@@ -499,6 +502,16 @@ extension NIODeadline {
 }
 
 extension EventLoop {
+    /// Execute the given function and synchronously complete the given `EventLoopPromise` (if not `nil`).
+    @inlinable
+    public func executeAndComplete<T>(_ promise: EventLoopPromise<T>?, _ body: () throws -> T) {
+        do {
+            promise?.succeed(try body())
```
Lukasa, line 509: Sadly you still can't use the `?` operator: you've introduced the same bug that was in #1746. Here again, if the promise is `nil`, the call to `try body()` will never be made. We need the code to look as it did before, with the call to `try body()` on a separate line.

Other than that little bug, I think this is reasonable enough, but I do think the API naming might be an issue. We'll see what @weissi thinks.

Hailong: Ah, sadly I totally missed your point 😂 It is very helpful to me, let me change it immediately. Yeah, I tried but failed to find a good name, I'd like to hear you guys opinions.

Sources/NIO/EventLoop.swift
```diff
@@ -229,6 +229,9 @@ public protocol EventLoop: EventLoopGroup {
     /// Submit a given task to be executed by the `EventLoop`
     func execute(_ task: @escaping () -> Void)
 
+    /// Execute the given function and synchronously complete the given `EventLoopPromise` (if not `nil`).
+    func executeAndComplete<T>(_ promise: EventLoopPromise<T>?, _ body: () throws -> T)
```
Lukasa, line 233: I don't think this realistically belongs on the `EventLoop`, this method probably belongs to `EventLoopPromise`.

weissi:
friendly ping @Hailong

Sources/NIO/EventLoop.swift
Hailong: @Lukasa @weissi I'm personally feel this method belongs to EventLoop other than EventLoopPromise, that's the reason I tried to create this PR#1747 to replace PR#1746. Maybe I can dig into the Netty code to get more ideas to help you guys to make a decision.

Lukasa: The reason I don't think it does is that the implementation you provide does not use `self` anywhere within it. It does not rely on the event loop in any way.

Hailong: @Lukasa I agree with you, it doesn't rely on event loop. I'm going ahead to close this PR.

I had also checked out the Netty's code, found its `bind` method and how it updates the promise. So it looks like there is no counterpart for `func executeAndComplete()` in Netty, it is doing to same thing inline. I'm not an expert of Netty, I hope I found the right place.

Anyway, I think `func executeAndComplete()` is only about executing the wrapped code, catching exception and then updating the promise correspondingly, there is nothing about EventLoop. Sorry for the confusion, and I'm going to close this PR.

image

cc @weissi

PR #1755, opened 2021-02-19
Add synchronous channel options

glbrntt:
Motivation:

The functions for getting and setting channel options are currently
asynchronous. This ensures that options are set and retrieved safely.
However, in some cases the caller knows they are on the correct event
loop but still has to pay the cost of allocating a future to either get
or set an option.

Modifications:

- Add a 'NIOSynchronousChannelOptions' protocol for getting and setting
  options
- Add a customisation point to 'Channel' to return 'NIOSynchronousChannelOptions'.
- Default implementation returns nil so as to not break API.
- Add implementations for 'EmbeddedChannel' and 'BaseSocketChannel'
- Allocation tests for getting and setting autoRead

Results:

Options can be get and set synchronously.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
@@ -15,19 +15,20 @@
 import NIO
 
 func run(identifier: String) {
+    let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+    defer {
+        try! group.syncShutdownGracefully()
+    }
+
+    let server = try! ServerBootstrap(group: group)
+      .bind(host: "127.0.0.1", port: 0)
+      .wait()
+    defer {
+        try! server.close().wait()
+    }
+
     measure(identifier: identifier) {
         let iterations = 1000
-        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-        defer {
-            try! group.syncShutdownGracefully()
-        }
-
-        let server = try! ServerBootstrap(group: group)
-          .bind(host: "127.0.0.1", port: 0)
-          .wait()
-        defer {
-            try! server.close().wait()
-        }
 
         for _ in 0..<iterations {
             let autoReadOption = try! server.getOption(ChannelOptions.autoRead).wait()
--- a/IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
@@ -1,0 +1,39 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+
+func run(identifier: String) {
+    measure(identifier: identifier) {
+        let iterations = 1000
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            try! group.syncShutdownGracefully()
+        }
+
+        let server = try! ServerBootstrap(group: group)
+          .bind(host: "127.0.0.1", port: 0)
+          .wait()
+        defer {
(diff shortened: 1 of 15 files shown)


glbrntt:
API breakage expected:

```
10:34:31 /* Protocol Requirement Change */
10:34:31 Var Channel.syncOptions has been added as a protocol requirement
```

Default implementation has been added so that _should_ be okay.

5.4 and nightly also expected to fail (5.4 compiler bug, nightly docker PR hasn't been merged yet).

IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
```diff
@@ -0,0 +1,39 @@
...
+
+func run(identifier: String) {
+    measure(identifier: identifier) {
+        let iterations = 1000
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            try! group.syncShutdownGracefully()
+        }
+
+        let server = try! ServerBootstrap(group: group)
```
weissi, line 25: this is a total nit but shouldn't we pull out everything apart from what we actually want to measure get/setOptions out of the `measure` block? Then you should get the sync get/set numbers to 0.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet_sync.swift
```diff
@@ -0,0 +1,43 @@
...
+        defer {
+            try! group.syncShutdownGracefully()
+        }
+
+        let server = try! ServerBootstrap(group: group)
+          .bind(host: "127.0.0.1", port: 0)
+          .wait()
+        defer {
+            try! server.close().wait()
+        }
```
weissi, line 30: more nits:

- same here, we could pull most of the code out of the `measure` block.
- instead of spinning up `MTELG`, we could also use `MTELG.withCurrentThreadAsEventLoop`, then you won't need an execute.

weissi (commented):
This is super awesome, thank you George! I'm basically happy as is but commented with some nits :)

IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet.swift
glbrntt: Ah, duh, of course.

IntegrationTests/tests_04_performance/test_01_resources/test_1000_autoReadGetAndSet_sync.swift
glbrntt: genius, I always forget about `withCurrentThreadAsEventLoop`!

Sources/NIO/Channel.swift
```diff
@@ -140,6 +140,30 @@ public protocol Channel: AnyObject, ChannelOutboundInvoker {
...
+
+    /// Returns a view of the `Channel` exposing synchronous versions of `setOption` and `getOption`.
+    /// The default implementation returns `nil`, and `Channel` implementations must opt in to
+    /// support this behavior.
+    var syncOptions: NIOSynchronousChannelOptions? { get }
+}
+
+extension Channel {
+    /// Default implementation: `NIOSynchronousChannelOptions` are not supported.
+    public var syncOptions: NIOSynchronousChannelOptions? {
```
Lukasa, line 152: It's a shame we can't have opaque return types for this.

Lukasa (approved):
After JW's notes, LGTM.

Sources/NIO/Channel.swift
glbrntt: Yup... life is much nicer in 5.2+

weissi: @Lukasa / @glbrntt I mean we _could_ (not saying we should) make this a 5.2+ only feature and guard it...

Lukasa: I suspect it isn't worth it, tbh. If we were really getting to the stage where we thought it was worth it I'd rather just be dropping support for 5.0 and 5.1.

weissi (approved):
Thank you George!

weissi:
Ok, this is good to go, master/5.4 are expected fails and so is API break. @glbrntt mind if I merge this?

glbrntt:
> Ok, this is good to go, master/5.4 are expected fails and so is API break. @glbrntt mind if I merge this?

Please do @weissi -- I don't have perms to override here.

weissi:
> > Ok, this is good to go, master/5.4 are expected fails and so is API break. @glbrntt mind if I merge this?
> 
> Please do @weissi -- I don't have perms to override here.

You should get them, CC @tomerd

weissi:
@glbrntt merged, thank you

PR #1758, opened 2021-02-22
surface errors when connection reset through pipeline

### LINKED ISSUE #1756: NIO doesn't necessarily send ECONNRESET and other errors on `reset()` through pipeline
In a few cases (at least), NIO doesn't send ECONNRESET and other errors on `reset()` through the pipeline.

Two scenarios are known:

- async connect
  * `channelActive`
  * receive EOF w/ half-closure on
  * `EPOLLHUP`

needs more investigation. Thanks to @fabianfett for mentioning that connect errors aren't necessarily surfaced.

weissi:
Motivation:

Previously, NIO wouldn't send errors that we pulled out of the socket on
connection reset through the pipeline. That meant that you wouldn't for
example see connect errors.

Modification:

Send errors through the pipeline if the channel is still alive after
trying to read all bytes.

Result:

- More erorrs surfaced to the user.
- fixes #1756

### DIFF
--- a/Sources/NIO/BaseSocketChannel.swift
+++ b/Sources/NIO/BaseSocketChannel.swift
@@ -68,73 +68,85 @@ private struct SocketChannelLifecycleManager {
     }
 
     @inline(__always) // we need to return a closure here and to not suffer from a potential allocation for that this must be inlined
-    internal mutating func beginRegistration() -> ((EventLoopPromise<Void>?, ChannelPipeline) -> Void) {
+    internal mutating func beginRegistration() -> ((EventLoopPromise<Void>?, Error?, ChannelPipeline) -> Void) {
         return self.moveState(event: .beginRegistration)
     }
 
     @inline(__always) // we need to return a closure here and to not suffer from a potential allocation for that this must be inlined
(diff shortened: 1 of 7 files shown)


weissi:
@Lukasa this isn't actually too bad at all. The real changes required are pretty minor. The cool thing is that my change only does much if the channel is still open after `readEOF0`. But if we do see an error there, then it'll be closed there and things should be fine because we never get to the new logic. If we do survive `readEOF0` without an error, then we should actually surface the error.

This needs much more testing and analysis, therefore still WIP. But I'll do that when I'm actually back at work :)

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -837,8 +849,19 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             self.pendingConnect = nil
             connectPromise.fail(error)
         }
+        
+        let pipelineError: Error?
+        switch error as? ChannelError {
+        case .some(ChannelError.inputClosed),
```
weissi, line 855: this is obviously ugly, we have a similar issue elsewhere that we have "errors" that aren't errors.

weissi:
Aged out, closing.

PR #1759, opened 2021-02-22
docs: debugging from the OS level

weissi:
Motivation:

When users have problems like missing data or similar, it's often useful
to debug from the OS level. The commands are a bit arcane and it saves
time if we don't have to remember them all the time.

Modification:

Add some macOS and Linux examples.

Result:

Better docs

### DIFF
--- a/docs/debug-os-level.md
+++ b/docs/debug-os-level.md
@@ -1,0 +1,74 @@
+# Debugging NIO programs from the OS level
+
+
+## List all file descriptors
+
+```
+lsof -p $(pgrep NIOEchoServer)
+```
+
+## Dump traffic to file
+
+For example to dump everything on port `12345` on interface `en0`:
+
+```
+sudo tcpdump -i en0 -w /tmp/dump.pcap '' 'port 12345'
+```
+
+## List event file descriptors
+
+### macOS
+
+```
+lskq -p $(pgrep NIOEchoServer)
+```
+
+### Linux
+
+```
+pid=$(pgrep NIOEchoServer); for fd_file in /proc/$pid/fd/*; do if [[ "$(readlink "$fd_file")" =~ eventpoll ]]; then fd=$(basename "$fd_file"); echo "Epoll $fd"; echo =====; cat "/proc/$pid/fdinfo/$fd"; fi; done
+```
+
+
+## Info about UNIX Domain Socket connections
+
+### Linux
+
+```
+sudo netstat --protocol=unix -peona
+```
+
+### macOS
+
+```
+netstat -n -a -f unix
+```
+
+## Info about TCP connections
+
+### Linux
+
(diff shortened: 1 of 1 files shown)


docs/debug-os-level.md
```diff
@@ -0,0 +1,74 @@
+# Debugging NIO programs from the OS level
+
+
+## List all file descriptors
+
+```
+lsof -p $(pgrep NIOEchoServer)
```
Lukasa, line 7: Is it worth noting that you should use the process name of the process you're trying to debug here? Won't necessarily be obvious to everyone that `NIOEchoServer` isn't a weird magic string.

docs/debug-os-level.md
```diff
@@ -0,0 +1,74 @@
...
+```
+sudo tcpdump -i en0 -w /tmp/dump.pcap '' 'port 12345'
+```
+
+## List event file descriptors
+
+### macOS
+
+```
+lskq -p $(pgrep NIOEchoServer)
```
Lukasa, line 23: Same note here about process name.

docs/debug-os-level.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### Linux
+
+```
+pid=$(pgrep NIOEchoServer); for fd_file in /proc/$pid/fd/*; do if [[ "$(readlink "$fd_file")" =~ eventpoll ]]; then fd=$(basename "$fd_file"); echo "Epoll $fd"; echo =====; cat "/proc/$pid/fdinfo/$fd"; fi; done
```
Lukasa, line 29: Same note here about process name.

Lukasa (approved):

docs/debug-os-level.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+sudo netstat --udp -peona
+```
+
+### macOS
+
+```
+netstat -n -p udp -a
+```
```
glbrntt, line 74: You already have this section above.

docs/debug-os-level.md
```diff
@@ -0,0 +1,74 @@
...
+
+## List all file descriptors
+
+```
+lsof -p $(pgrep NIOEchoServer)
+```
+
+## Dump traffic to file
+
+For example to dump everything on port `12345` on interface `en0`:
```
glbrntt, line 12:
```suggestion
For example to dump everything on port `12345` on interface `en0` to `/tmp/dump.pcap`:
```

PeterAdams-A (commented):
I guess my thought is that this is a useful list of things people can do but perhaps a couple of lines on each explaining why you might want to do it may help - we'd previously talked about some people not getting as far as knowing what they even want to do.

weissi:
Aged out, closing.

PR #1763, opened 2021-02-23
HTTP server upgrade tests: cosmetic improvements

weissi:
Motivation:

The HTTP server upgrade tests had unncessarily wide and also verbose
test assertions.

Modifications:

Rewrote them much simpler and narrowed what states are allowed.

Result:

- Neater code
- Works around https://bugs.swift.org/browse/SR-14253

### DIFF
--- a/Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
+++ b/Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
@@ -866,14 +866,7 @@ class HTTPServerUpgradeTestCase: XCTestCase {
 
         let channel = EmbeddedChannel()
         defer {
-            do {
-                let complete = try channel.finish()
-                XCTAssertTrue(complete.isClean)
-            } catch No.no {
-                // ok
-            } catch {
-                XCTFail("Unexpected error: \(error)")
-            }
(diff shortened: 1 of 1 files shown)


Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
```diff
@@ -866,14 +866,7 @@ class HTTPServerUpgradeTestCase: XCTestCase {
...
         let channel = EmbeddedChannel()
         defer {
-            do {
-                let complete = try channel.finish()
-                XCTAssertTrue(complete.isClean)
-            } catch No.no {
-                // ok
-            } catch {
-                XCTFail("Unexpected error: \(error)")
-            }
```
Lukasa, line 876: This is not a semantic replacement for the above. This used to tolerate an error that now is not tolerated.

weissi: @Lukasa indeed. That's what I meant with "tighten" in the description. It's on EmbeddedChan so it should be deterministic and the tests pass, shouldn't we take this?

Lukasa (approved):
Yup, good spot.

PR #1766, opened 2021-02-24
Use the pre-succeeded void future when possible for async pipeline operations

glbrntt:
Motivation:

We added synchronous pipeline operations to allow the caller to save
allocations when they know they are already on the correct event loop.
However, we missed a trick! In some cases the caller cannot guarantee
they are on the correct event loop and must use an asynchronous method
instead. If that method returns a void future and is called on the event
loop, then we can perform the operation synchronously and return a
cached void future.

Modifications:

- Add API to `EventLoop` for creating a 'completed' future with a
  `Result` (similar to `EventLoopPromise.completeWith`)
- Add an equivalent for making completed void futures
- Use these when asynchronously adding handlers and the caller is
  already on the right event loop.

Result:

- Fewer allocations on the happiest of happy paths when adding handlers
  asynchronously to a pipeline.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -262,7 +262,7 @@ public protocol EventLoop: EventLoopGroup {
     /// Asserts that the current thread is the one tied to this `EventLoop`.
     /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
     func preconditionInEventLoop(file: StaticString, line: UInt)
-    
+
     /// Asserts that the current thread is _not_ the one tied to this `EventLoop`.
     /// Otherwise, the process will be abnormally terminated as per the semantics of `preconditionFailure(_:file:line:)`.
     func preconditionNotInEventLoop(file: StaticString, line: UInt)
@@ -372,7 +372,7 @@ extension TimeAmount: AdditiveArithmetic {
     public static func + (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds + rhs.nanoseconds)
     }
-    
+
     public static func +=(lhs: inout TimeAmount, rhs: TimeAmount) {
         lhs = lhs + rhs
     }
@@ -380,7 +380,7 @@ extension TimeAmount: AdditiveArithmetic {
     public static func - (lhs: TimeAmount, rhs: TimeAmount) -> TimeAmount {
         return TimeAmount(lhs.nanoseconds - rhs.nanoseconds)
     }
-    
+
     public static func -=(lhs: inout TimeAmount, rhs: TimeAmount) {
         lhs = lhs - rhs
     }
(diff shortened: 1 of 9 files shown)


Lukasa (approved):
I'm happy with this, I'd like to let @weissi review the API additions as well.

Sources/NIO/EventLoop.swift
```diff
@@ -608,6 +608,35 @@ extension EventLoop {
         }
     }
 
+    /// Creates and returns a new `EventLoopFuture` that is marked as succeeded or failed with the value held by `result`.
+    ///
+    /// - Parameters:
+    ///   - result: The value that is used by the `EventLoopFuture`
+    /// - Returns: A completed `EventLoopFuture`.
+    @inlinable
+    public func makeCompletedFuture<Success>(_ result: Result<Success, Error>, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
```
weissi, line 617: these don't need the `file`/`line`, don't know why `makeFailed/Succeeded` want them. The `file`/`line` is only interesting for promises so we can tell you where you created an unfulfilled one.

Sources/NIO/EventLoop.swift
```diff
@@ -608,6 +608,35 @@ extension EventLoop {
...
+            return self.makeFailedFuture(error, file: file, line: line)
+        }
+    }
+
+    /// Creates and returns a new `EventLoopFuture` that is marked as succeeded or failed with the value held by `result`.
+    ///
+    /// - Parameters:
+    ///   - result: The value that is used by the `EventLoopFuture`
+    /// - Returns: A completed `EventLoopFuture`.
+    public func makeCompletedVoidFuture(_ result: Result<Void, Error>, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Void> {
```
weissi, line 631: same here

weissi (changes requested):
This is very cool! We just have some extra parameters that (I think) we don't need :)

Sources/NIO/EventLoop.swift
```diff
@@ -608,6 +608,35 @@ extension EventLoop {
...
+            return self.makeFailedFuture(error)
+        }
+    }
+
+    /// Creates and returns a new `EventLoopFuture` that is marked as succeeded or failed with the value held by `result`.
+    ///
+    /// - Parameters:
+    ///   - result: The value that is used by the `EventLoopFuture`
+    /// - Returns: A completed `EventLoopFuture`.
+    public func makeCompletedVoidFuture(_ result: Result<Void, Error>) -> EventLoopFuture<Void> {
```
weissi, line 631: So technically, we don't need these two methods. We can just use the fully generic one and do:

```
if Success.self == Void.self {
    // code for makeCompletedVoidFuture here
} else {
   // code for makeCompletedFuture here
}
```

and because of the `@inlinable` we get the same perf and will have a nicer API. WDYT @Lukasa / @glbrntt ?

Lukasa: SGTM

glbrntt: Don't even need to do that I think, `makeCompletedFuture` does it for us :)

weissi (approved):
yay! thank you George!

PR #1771, opened 2021-03-01
Make stackdiff-dtrace.py a little more tolerant

glbrntt:
Motivation:

Some of the output from 'malloc-aggregation.d' trips up the
'stackdiff-dtrace.py' script -- especially when used with the output
from recent 5.4 builds. The regex for numbers includes hex which trips
us up when we try to convert that match to an int. Since we only use
this pattern for matching allocation counts, where we don't expect hex,
we can just modify the pattern.

Also the input parsing is a little tricky to follow without context so I
tidied it up little.

Modifications:

- relax the number regex (we don't expect counts to be in hex)
- drop the stack regex (it wasn't used)
- aggregate the current stack in a list (instead of a string)
- filter out non-stack lines (which were erroneously used to key the
  first stack!)
- a few comments

Result:

stackdiff-dtrace.py doesn't blow up on output from newer builds

### DIFF
--- a/dev/stackdiff-dtrace.py
+++ b/dev/stackdiff-dtrace.py
@@ -15,28 +15,71 @@
 
 import sys
 import re
+import collections
 
-num_regex = "^ +([0-9a-fx]+)$"
-stack_regex = "^ +(.*)$"
+num_regex = "^ +([0-9]+)$"
 
 def put_in_dict(path):
-    dictionary = {}
+    # Our input looks something like:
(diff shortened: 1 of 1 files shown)


weissi (approved):
Yippie, thank you!!

weissi:

glbrntt:

weissi:
CI broken, I'm going to merge this still.

PR #1775, opened 2021-03-08
Decrease overhead of internal size table used in AdaptiveRecvByteBufferAllocator

hassila:
### Motivation:
Decrease overhead of internal size table used in AdaptiveRecvByteBufferAllocator.
It currently has a fairly large number of entries that reasonably never will
be used in real life. Just a minor polish.

### Modifications:

Change the maximum size slot allocated to 1 << 32 for the buffer and
doesn't use unsigned wrap around as a stop condition anymore for the setup.

### Result:

Save roughly 256 bytes, easier to read code.


Lukasa (approved):
Sounds good, thanks for the patch!

Lukasa:

PR #1776, opened 2021-03-09
Make HTTP Parser import implementationOnly

Lukasa:
Motivation:

Non-implementation-only imports in Swift have an annoying tendency to
put C header files into all downstream build paths. The result is that
leaf packages that include any C header files run a very high risk of
colliding with declarations in the inner header files. In this case,
we've had users bump into issues with the HTTP_STATUS_* enum cases
already.

Modifications:

- Make CNIOHTTPParser import implementationOnly where possible.

Result:

Fewer risks of colliding symbols.


glbrntt (approved):

weissi (approved):
nice nice, thanks

PR #1777, opened 2021-03-09
More relevant email address for reporting SECURITY issues

glbrntt:
Motivation:

@tomerd made this suggestion but it was missed in the previous PR.

Modifications:

- Use 'sswg-security-reports' as the recipients are more relevant to
  this project


weissi (approved):
ta!

PR #1782, opened 2021-03-16
Support multiple shared files in the allocation counter test framework

glbrntt:
Motivation:

The allocation counter test framework only supports a single shared
file which is quite limiting, especially if you rely on generated code.

Modifications:

- The '-s' option can now be passed multiple times
- The symlink name for any of the passed files will be the basename of
  the provided file rather than 'shared.swift'

Result:

Allocation counter test framework can use more than one shared file.

### DIFF
--- a/IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
+++ b/IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
@@ -112,7 +112,7 @@ EOF
 }
 
 # <target> <template> <swiftpm_pkg_root> <swiftpm_pkg_name> <hooked_function_module> <bootstrap_module>
-# <shared file> <extra_dependencies_file> <modules...> -- <test_files...>
+# <extra_dependencies_file> <shared files...> -- <modules...> -- <test_files...>
 function build_package() {
     local target=$1
     local template=$2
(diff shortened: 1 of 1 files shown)


IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
```diff
@@ -176,9 +182,10 @@ function build_package() {
 EOF
         ln -s "$f" "Sources/Test_$module/file.swift"
         ln -s "../../scaffolding.swift" "Sources/Test_$module/"
-        if [[ -n "$shared_file" ]]; then
-            ln -s "$shared_file" "Sources/Test_$module/shared.swift"
-        fi
+        for shared_file in "${shared_files[@]}"; do
```
weissi, line 185: I think this will break if there are _no_ shared files. Or did you test that?

```suggestion
        for shared_file in "${shared_files[@]+"${shared_files[@]}"}"; do
```

weissi (changes requested):
Looks mostly great but I think there's one bug regarding bash's stupid handling of arrays

IntegrationTests/allocation-counter-tests-framework/run-allocation-counter.sh
glbrntt: Totally right it was busted if there were no shared files, thanks!

weissi (approved):

PR #1784, opened 2021-03-18
Make ChannelHandler removal cheaper.

Lukasa:
Motivation:

Any version of ChannelHandler removal that does not have a
ChannelHandlerContext already in hand is currently excessively
expensive. This is because it allocates a promise and a callback for
finding the context, despite already having a promise in hand for users
to complete.

We can remove a pair of allocations here by jumping to the event loop
directly and then running our operations synchronously.

Modifications:

- Rewrite removeHandler(name:promise:) and removeHandler(_:promise:) to
  jump directly to the event loops and then work synchronously.

Result:

Cheaper code

### DIFF
--- a/docker/docker-compose.2004.54.yaml
+++ b/docker/docker-compose.2004.54.yaml
@@ -21,8 +21,8 @@ services:
       - MAX_ALLOCS_ALLOWED_1000_addHandlers=47050
       - MAX_ALLOCS_ALLOWED_1000_addHandlers_sync=40050
       - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlercontext=9035
-      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlername=18035  # improvement from 5.3 which was 19035
-      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlertype=18035  # improvement from 5.3 which was 19035
+      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlername=9035
+      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlertype=9035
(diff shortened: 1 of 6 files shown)


Lukasa:
I'm waiting to see if the alloc counter tests are affected by this. If they aren't, I'll write one.

weissi (approved):
<3

docker/docker-compose.2004.54.yaml
```diff
@@ -21,8 +21,8 @@ services:
       - MAX_ALLOCS_ALLOWED_1000_addHandlers=47050
       - MAX_ALLOCS_ALLOWED_1000_addHandlers_sync=40050
       - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlercontext=9035
-      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlername=18035  # improvement from 5.3 which was 19035
-      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlertype=18035  # improvement from 5.3 which was 19035
+      - MAX_ALLOCS_ALLOWED_1000_addRemoveHandlers_handlername=9035  # improvement from 5.3 which was 19035
```
weissi, line 24: @Lukasa this comment isn't right anymore, 5.3 seems to be the same now?

Lukasa:
For future readers: yes, this patch saves 10 allocations from the use-case where we do not already have a ChannelHandlerContext, more than 50% of the allocations on that operation.

docker/docker-compose.2004.54.yaml
Lukasa: Yup, fixed.

weissi:
great stuff!

weissi:
50% win, super cool, AHC will thank you :)

PeterAdams-A (approved):
Looks good to me.

PR #1786, opened 2021-03-18
Sometimes @escaping is a lie.

Lukasa:
Motivation:

_contextSync and contextForPredicate0 both accept their closure as
@escaping. _contextSync does so presumably only becuase
contextForPredicate0 does. However, this @escaping directive is
unnecessary: the closure does not escape, and never could.

In general this doesn't matter as these methods are internal, and called
from within the library. However, _contextSync is @usableFromInline, and
is called on several @inlinable call paths, including:

- `ChannelPipeline.handler(type:)`
- `ChannelPipeline.context(handlerType:)`
- `ChannelPipeline.SynchronousOperations.context(handlerType:)`
- `ChannelPipeline.SynchronousOperations.handler(type:)`

All of these presumably allocate more than they need to because there's
an optimisation boundary preventing user code from observing that the
closure does not actually allocate.

Modifications:

Delete `@escaping`.

Result:

Presumably faster code.


weissi (approved):
nice :)

PR #1787, opened 2021-03-21
Better error type for inactive thread pools

Davidde94:
Running work on an inactive `NIOThreadPool` now throws an error `NIOThreadPoolError.ThreadPoolInactive`.

### Motivation:

Previously it threw `ChannelError.ioOnClosedChannel`, which was very misleading.

### Modifications:

Define the new error type and add a test to `NonBlockFileIO`

### Result:

A more descriptive error is thrown.

### DIFF
--- a/Tests/NIOTests/NonBlockingFileIOTest.swift
+++ b/Tests/NIOTests/NonBlockingFileIOTest.swift
@@ -919,20 +919,20 @@ class NonBlockingFileIOTest: XCTestCase {
     func testThrowsErrorOnUnstartedPool() throws {
         
         try withTemporaryFile(content: "hello, world") { fileHandle, path in
+            let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+            defer {
+                XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
+            }
             
             let expectation = XCTestExpectation(description: "Opened file")
             let threadPool = NIOThreadPool(numberOfThreads: 1)
             let fileIO = NonBlockingFileIO(threadPool: threadPool)
-            let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
             fileIO.openFile(path: path, eventLoop: eventLoopGroup.next()).whenFailure { (error) in
                 XCTAssertTrue(error is NIOThreadPoolError.ThreadPoolInactive)
                 expectation.fulfill()
             }
             
-            XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
             self.wait(for: [expectation], timeout: 1.0)
-            
         }
-    
     }
 }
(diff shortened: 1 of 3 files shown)


Tests/NIOTests/NonBlockingFileIOTest.swift
```diff
@@ -915,4 +915,22 @@ class NonBlockingFileIOTest: XCTestCase {
...
+        let expectation = XCTestExpectation(description: "Opened file")
+        let threadPool = NIOThreadPool(numberOfThreads: 1)
+        let fileIO = NonBlockingFileIO(threadPool: threadPool)
+        let eventLoop = MultiThreadedEventLoopGroup(numberOfThreads: 1).next()
+        fileIO.openFile(path: fileURL.path, eventLoop: eventLoop).whenFailure { (error) in
+            XCTAssertTrue(error is NIOThreadPoolError.ThreadPoolInactive)
+            expectation.fulfill()
+        }
+        
+        eventLoop.shutdownGracefully {_ in }
```
Lukasa, line 933: We should keep the event loop group around and use `syncShutdownGracefully` on it.

Tests/NIOTests/NonBlockingFileIOTest.swift
```diff
@@ -915,4 +915,22 @@ class NonBlockingFileIOTest: XCTestCase {
             XCTAssertEqual(numberOfChunks, numberOfCalls.load())
         })
     }
+    
+    func testThrowsErrorOnUnstartedPool() {
+        
+        let fileURL = URL(fileURLWithPath: "/tmp/nonblockingfileiotest.txt")
```
Lukasa, line 921: Can we use the `withTemporaryFile` helper here? It avoids test runs colliding with each other.

Lukasa (changes requested):
The code looks great, just a few notes on the test.

Tests/NIOTests/NonBlockingFileIOTest.swift
```diff
@@ -915,4 +915,24 @@ class NonBlockingFileIOTest: XCTestCase {
...
+            let expectation = XCTestExpectation(description: "Opened file")
+            let threadPool = NIOThreadPool(numberOfThreads: 1)
+            let fileIO = NonBlockingFileIO(threadPool: threadPool)
+            let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+            fileIO.openFile(path: path, eventLoop: eventLoopGroup.next()).whenFailure { (error) in
+                XCTAssertTrue(error is NIOThreadPoolError.ThreadPoolInactive)
+                expectation.fulfill()
+            }
+            
+            XCTAssertNoThrow(try eventLoopGroup.syncShutdownGracefully())
```
Lukasa, line 932: It would be better to move this to a `defer` block so that it's clear that this shutdown isn't material to the logic of the test itself.

Lukasa (approved):
Move cleanup to the defer block and I'm happy.

PR #1792, opened 2021-04-06
CircularBuffer: use vendored swift-collections/Deque instead

weissi:
Motivation:

Less (SwiftNIO maintained) code is better and Swift Collection's got an
excellent Deque implementation that NIO should just use.

Modification:

Use Swift Collection's Deque to back CircularBuffer instead rolling our
own (which is certainly much slower).

Result:

- Less (NIO maintained) code
- faster


weissi:
CC @Lukasa  / @lorenty

Sources/NIO/VendoredDeque/Deque+CustomReflectable.swift
```diff
@@ -0,0 +1,23 @@
+/* Changes for SwiftNIO
+   - renamed NIODeque to NIONIODeque (to prevent future clashes)
+   - made (NIO)NIODeque internal (not public)
```
Lukasa, line 3: Something here is prefixing NIO again.

ktoso:
Oh, awesome :-)

weissi:
> Oh, awesome :-)

Except that it doesn’t work 😂😂. https://forums.swift.org/t/rfc-deque-should-it-have-an-opaque-index/47203

PR #1788, opened 2021-03-26
Initial implementation of liburing support for SwiftNIO (io_uring)

hassila:
Added initial implementation of liburing as discussed in #1761.

### Motivation:

As outlined in https://github.com/apple/swift-nio/issues/1761, io_uring is a new async I/O facility on Linux that should have less overhead than epoll.

This draft PR includes a first stab at adding this to SwiftNIO so that a few of the 
remaining failed tests can be discussed in context of code / implementation.

There [practical details documented](https://github.com/apple/swift-nio/blob/1f0a0ccf29b13d754d49d8d96670f5095b188fb5/docs/io_uring.md) as part of the PR.

### Modifications:

- Separated Selector into concrete subclasses KqueueSelector, EpollSelector and UringSelector.
- Choose type of Selector in MultiThreadedEventLoopGroup - current implementation will try to make use of liburing if the library was compiled with it and it is available on the deployment machine (and has at least kernel version 5.12, which isn't enough - currently a custom kernel is needed to use the multishot poll feature.  
- Added shims for liburing for platforms that doesn't have it (pulled in two headers from liburing to allow compilation on epoll-only platforms - they are not used by the real implementation.)
- Manually disabled a few tests that doesn't work yet.
- Disabled one assert that trips during normal usage likely due to async nature of poll updates, for discussion
- Added shared kernel sqpoll ring support (can be run with normal user privs in 5.13)
- Support for both single shot polls (should work all the way back to 5.1 kernels, needs testing) and multishot streaming polls and modifications for polls (scheduled due in 5.13) for slightly better performance (and better impedance match to SwiftNIO usage)
- Added extensive debug logs (should likely be removed when bringup and testing is complete)
- Performance measurements and tuning remains - should aim for robustness first. 

- Also, there are several aspects to discuss:
- - io_uring is edge triggered (requiring us to do unnecessary reregistrations of polling as we are not doing exhaustive reads)
- - we are not using io_uring for read/write operations (as discussed in the original issue)
- - ...

### Result:

Basic liburing support is in place.

Issue #1761, opened 2021-02-23
[Linux] evaluate io_uring() as alternative async I/o mechanism 


hassila:
Stumbled across another (new) async I/o interface that seems worthwhile considering for best possible performance on Linux. See some background here:

https://lwn.net/Articles/810414/
https://kernel.dk/io_uring.pdf

And one simple test vs epoll with performance numbers:
https://github.com/frevib/io_uring-echo-server/blob/io-uring-op-provide-buffers/benchmarks/benchmarks.md

Is this something of interest?

hassila:
Here are some results from netty who adopted it:
https://github.com/netty/netty/issues/10622#issuecomment-701241587

PeterAdams-A:
Thanks for the heads up @hassila.  io_uring has been on our radar for a while.  Unfortunately it's yet to reach the top of our list as we are busy with many other things.  We agree that it is likely to lead to improved performance on linux.  

If you feel so inclined we'd welcome any code contributions in this direction.

hassila:
Thanks @PeterAdams-A - I could have a look, it seems Ubuntu 20.04 has a fresh enough kernel that supports it (even if performance of io_uring seems to have improved quite a bit further with more recent ones than 5.4 that Ubuntu uses, but it should be usable for bringup). 

Just a few basic initial questions (I'm sure I'd need some handholding on best integration with existing code-base):

- Looking at current #if usage, it seems that it is done on a platform rather than a feature basis - I'm used to (from previous life) to use e.g. ifdefs that checks for e.g. kqueue/epoll rather than OS. In this case, I don't think a PR should replace the current epoll implementation for multiple reasons (primarily support for older kernels on e.g. Amazon Linux 2 etc that lacks io_uring, but also eases of comparing performance between epoll/io_uring). What would be a best practice for supporting two implementations for a single platform and to be able to select between them?
- The inventor of io_uring also provides a convenience library  (https://github.com/axboe/liburing) that removes a significant chunk of boilerplate - is this something that would be acceptable as a dependency / part of implementation, or would a "clean" use of the lower-level primitives (but with possibly harder-to-read code) be preferred for a successful PR?
- Would any of the existing tests be viewed as usable for validating performance, or how would you propose validation is done in an acceptable manner?

I'm sure there'll be additional questions, but just some stuff that came to mind when looking at it.

Lukasa:
> What would be a best practice for supporting two implementations for a single platform and to be able to select between them?

[...]

PR #1793, opened 2021-04-06
Add custom cleanup hook to allow PipeChannel to properly deregister during halfclose

hassila:
### Motivation:
Currently PipeChannel doesn't properly deregister from the selector during half-closes.
This means that some backend implementations will keep a reference on the kernel side
which means that the POLLHUP will not be delivered for the read end of a pipe in
e.g. NIOTests.StreamChannelTest/testHalfCloseOwnOutput - this has the test failing
in #1761 / #1788.

### Modification:

Add a custom cleanup hook in BaseStreamSocketChannel in close0 to allow PipeChannel
to properly deregister.

### Result:

The final reference to the fd is dropped resulting in a POLLHUP being delivered
as needed - making NIOTests.StreamChannelTest/testHalfCloseOwnOutput to pass
for io_uring.

### DIFF
--- a/Sources/NIO/BaseStreamSocketChannel.swift
+++ b/Sources/NIO/BaseStreamSocketChannel.swift
@@ -75,14 +75,14 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
     // Hook for customizable socket shutdown processing for subclasses, e.g. PipeChannel
     func shutdownSocket(mode: CloseMode) throws {
         switch mode {
-            case .output:
-                try self.socket.shutdown(how: .WR)
-                self.outputShutdown = true
-            case .input:
-                try socket.shutdown(how: .RD)
-                self.inputShutdown = true
-            case .all:
-                break
+        case .output:
+            try self.socket.shutdown(how: .WR)
+            self.outputShutdown = true
+        case .input:
+            try socket.shutdown(how: .RD)
+            self.inputShutdown = true
+        case .all:
+            break
         }
     }
 
--- a/Sources/NIO/BaseStreamSocketChannel.swift
+++ b/Sources/NIO/BaseStreamSocketChannel.swift
@@ -72,6 +72,20 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
         }
     }
 
+    // Hook for customizable socket shutdown processing for subclasses, e.g. PipeChannel
+    func shutdownSocket(mode: CloseMode) throws {
+        switch mode {
+            case .output:
+                try self.socket.shutdown(how: .WR)
+                self.outputShutdown = true
+            case .input:
+                try socket.shutdown(how: .RD)
+                self.inputShutdown = true
+            case .all:
+                break
+        }
+    }
+
     // MARK: BaseSocketChannel's must override API that cannot be further refined by subclasses
     // This is `Channel` API so must be thread-safe.
     final override public var isWritable: Bool {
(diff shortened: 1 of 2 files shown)


Sources/NIO/BaseStreamSocketChannel.swift
```diff
@@ -161,6 +164,7 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                     promise?.fail(ChannelError.outputClosed)
                     return
                 }
+                _close0Cleanup(mode:mode)
```
Lukasa, line 167: Some minor style nits here: we'd normally want this to be:

```suggestion
                self._close0Cleanup(mode: mode)
```

The same applies to the two other instances below.

Lukasa (commented):
Thanks for this patch!

My top-level question is whether we need a new hook here, or whether we can simply `override func close0` for `PipeChannel`. It doesn't seem like it'd be a big deal to unregister as part of that work for the appropriate modes. Have I missed something?

hassila:
> Thanks for this patch!
> 
> My top-level question is whether we need a new hook here, or whether we can simply `override func close0` for `PipeChannel`. It doesn't seem like it'd be a big deal to unregister as part of that work for the appropriate modes. Have I missed something?

Well, I started on that path first, but assuming fundamentally that the rest of the work in close0 is needed, it ended up with a lot of code duplication. The actual deregistration is sandwiched in, so it was not possible to just to the deregistration and call super, so it felt like a less error prone change (if some of the reset of the close0 implementation would be changed in the future, it would need to be done in two places if duplicated, with the hook it should be fine...) - thus the hook.

Sources/NIO/BaseStreamSocketChannel.swift
hassila: I've fixed the callsites, thanks.

hassila:
Oh, and as the dereg couldn't be done at the beginning/end, code duplication was also problematic if we wanted to call super. And as super.close0 also calls super, it's a bit messy - so the hook was the cleanest solution I could see without some major surgery.

Lukasa:
Fair enough! In that case, I wonder if the more general solution is to factor out the call to `socket.shutdown` into a more general customisation point. That will let us model socket shutdown in a more complex way for pipes. How does that sound?

hassila:
Thanks, sounds very reasonable - I'll have a look tomorrow.

hassila:
Please have a look at the updated version in https://github.com/apple/swift-nio/pull/1793/commits/a19c45aac3201783fc39930bf3c63589e48b999d

Sources/NIO/BaseStreamSocketChannel.swift
```diff
@@ -72,6 +72,20 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
         }
     }
 
+    // Hook for customizable socket shutdown processing for subclasses, e.g. PipeChannel
+    func shutdownSocket(mode: CloseMode) throws {
+        switch mode {
+            case .output:
```
Lukasa, line 78: Nit: NIO indentation here would be:

```
switch mode {
case .output:
```

Sources/NIO/PipeChannel.swift
```diff
@@ -92,6 +92,18 @@ final class PipeChannel: BaseStreamSocketChannel<PipePair> {
         try! self.selectableEventLoop.deregister(channel: self, mode: .output)
         try! self.pipePair.outputFD.close()
     }
+
+    override func shutdownSocket(mode: CloseMode) throws {
+        switch mode {
+            case .input:
```
Lukasa, line 98: Same indentation nit here.

Lukasa (commented):
A couple of nits regarding code formatting, but otherwise this looks good.

Sources/NIO/BaseStreamSocketChannel.swift
hassila: Ok (but ouch for my personal taste :-) - fixed both and pushed...

Lukasa (approved):
Great, I like this a lot. @weissi?

weissi (approved):
Awesome, thank you! That looks great to me

PR #1794, opened 2021-04-07
Add 'promise' versions of EventLoopFuture.{and,when}All{succeed,complete}

glbrntt:
Motivation:

Most NIO APIs have two versions: one accepting a promise (often
optional) and another which returns a future. Many applications and
frameworks built on top of NIO also use this pattern. To maximise
composability we should try to provide both versions where possible.

Modifications:

Add versions of `andAllSucceed`, `andAllComplete`, `whenAllSucceed`
and `whenAllComplete` which accept an `EventLoopPromise` in place of
an `EventLoop` and defer the implementation of the future versions to
the promise versions.

Result:

- Better composability.
- Fewer allocations if the caller already has a promise.

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -1072,8 +1072,24 @@ extension EventLoopFuture {
     ///     - futures: An array of homogenous `EventLoopFutures`s to wait for.
     ///     - on: The `EventLoop` on which the new `EventLoopFuture` callbacks will execute on.
     /// - Returns: A new `EventLoopFuture` that waits for the other futures to succeed.
+    @inlinable
     public static func andAllSucceed(_ futures: [EventLoopFuture<Value>], on eventLoop: EventLoop) -> EventLoopFuture<Void> {
         let promise = eventLoop.makePromise(of: Void.self)
+        EventLoopFuture.andAllSucceed(futures, promise: promise)
+        return promise.futureResult
+    }
+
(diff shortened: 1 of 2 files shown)


weissi (commented):
This looks great and these functions are already implicitly tested (by the old ones) but I'd still argue that we should have at least one test case that uses the API. Just that we have direct test callers into everything that's API. If you think that's unnecessary I'm probably happy to get it in without direct tests too.

glbrntt:
> but I'd still argue that we should have at least one test case that uses the API. Just that we have direct test callers into everything that's API.

What's the benefit?

I couldn't see any given they are already covered via tests for the other API but happy to add them if you have a reason to.

weissi:
> What's the benefit?

Two reasons really:
-  it's easier to spot if we change API accidentally (should be covered by the API breakage checker)
- if (for any reason) we reimplement the others with another implementation, we'd be left without tests for the methods you just introduced.

glbrntt:
> > What's the benefit?
> 
> Two reasons really:
> 
> * it's easier to spot if we change API accidentally (should be covered by the API breakage checker)
> * if (for any reason) we reimplement the others with another implementation, we'd be left without tests for the methods you just introduced.

I don't buy into the first one for the reason you pointed out: the API breakage checker for us.

Second seems reasonable enough though, I'll add some tests. It'd be neat if we could check that in CI, i.e. determine when lines previously covered by tests stop being covered.

weissi (approved):
lgtm, thank you!

PR #1795, opened 2021-04-08
Split Selector into separate files per backend.

hassila:
### Motivation

Currently all backends are platform-conditional in the `Selector` code, making it
harder to read and difficult to extend for future new backend implementations.
By splitting the `Selector` into a generic part and per-backend platform implementation
it is easier to add additional backends in the future.

### Modifications
Split Selector into `SelectorGeneric`, `SelectorEpoll` and `SelectorKqueue`.
Added a new defaultSelectorFactory that picks the proper platform-dependent
backend.
Fixed tests.

### Result
- Easier to read code
- Makes it easier to add new backends


hassila:
I've removed the malloc checks for 5.0/5.1, but as I force-pushed the branch it does not seem to trigger a retest - is there any way except a dummy commit?

Sources/NIO/SelectorEpoll.swift
```diff
@@ -0,0 +1,290 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5: given that this code has mostly been moved, I think this needs to be 2017-2021

Sources/NIO/SelectorEpoll.swift
```diff
@@ -0,0 +1,290 @@
...
+            thing.formUnion(.output)
+        }
+        if selectorEventSet.contains(.readEOF) {
+            thing.formUnion(.readHangup)
+        }
+        self = thing
+    }
+}
+
+#if os(Linux) || os(Android)
```
weissi, line 56: can't we now just do one huge `#if os(Linux) || os(Android)` around the whole file?

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,333 @@
...
+
+import NIOConcurrencyHelpers
+
+internal enum SelectorLifecycleState {
+    case open
+    case closing
+    case closed
+}
+
+internal extension Optional {
```
weissi, line 23: please move all the access controls away from `extension` to the actual declarations. In this case `internal` is default anyway so you can just remove it.

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,333 @@
...
+            return try withUnsafePointer(to: &this) { x in
+                try body(x)
+            }
+        } else {
+            return try body(nil)
+        }
+    }
+}
+
+internal extension timespec {
```
weissi, line 35: same here, `internal` to the `func`, or remove `internal` entirely.

Sources/NIO/SelectorKqueue.swift
```diff
@@ -0,0 +1,356 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5: simliar note here, I think 2017-2021 is correct

Sources/NIO/SelectorKqueue.swift
```diff
@@ -0,0 +1,356 @@
...
+            }
+            try body(UnsafeMutableBufferPointer(start: keventBuffer, count: index))
+        }
+    }
+    #endif
+}
+
+/* this is deliberately not thread-safe, only the wakeup() function may be called unprotectedly */
+final internal class KqueueSelector<R: Registration>: Selector<R> {
+    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
```
weissi, line 95: here as well, can we make this one huge `#if os...` around the whole file?

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,333 @@
...
+    }
+    
+    deinit {
+        assert(self.registrations.count == 0, "left-over registrations: \(self.registrations)")
+        assert(self.lifecycleState == .closed, "Selector \(self.lifecycleState) (expected .closed) on deinit")
+        assert(self.selectorFD == -1, "self.selectorFD == \(self.selectorFD) on Selector deinit, forgot close?")
+    }
+    
+    // internal selectorFactory to choose variant of Selector to use.
+    static func defaultSelectorFactory<R>() throws -> NIO.Selector<R> {
```
weissi, line 134: as an alternative to the sub-classing we could use

```
#if os(Linux) || os(Android)
internal typealias Selector=EpollSelector
#if os(.......)
internal typealias Selector=KqueueSelector
#else
#error("sorry, your OS is unsupported, we need kqueue or epoll support")
#endif
```

and add the default functions using `extension Selector`. The only duplication would be the stored properties that we'd then add to each Selector individually.

@Lukasa WDYT: sub-classes as proposed here or `typealias Selector=MyOSSelector` and default functions as `extension`. With sub-classes we need to be careful if we want to avoid virtual calls (should be possible).

weissi: if we did sub-classes, then we should probably (depending on `#if os` etc return the concrete type of the selector here, not `NIO.Selector<R>`)

docker/docker-compose.1804.50.yaml
```diff
@@ -17,39 +17,6 @@ services:
 
   test:
     image: swift-nio:18.04-5.0
-    environment:
-      - MAX_ALLOCS_ALLOWED_1000_addHandlers=47050
```
weissi, line 21: these alloc changes (only in 5.0 & 5.1 which I don't actually care about) are probably the result of the subclassing I guess?

weissi (changes requested):
Thank you, this is awesome, wanted to do this for a long time.

This is pretty much ready to go, need to fix the copyright dates and decide if we're happy with the sub classes or not.

weissi:
> I've removed the malloc checks for 5.0/5.1, but as I force-pushed the branch it does not seem to trigger a retest - is there any way except a dummy commit? (I wanted to keep this as a single commit for nicer integration)

force push will re-trigger. Maybe it was a bit slow? But I see it currently is up to date. (manual retries can be done with a comment: `@swift-nio-bot test this please`)

docker/docker-compose.1804.50.yaml
hassila: 5.0 only I think, 5.1 could have stayed, it actually passed there, but as you didn't care about it, I thought it might go too.

Sources/NIO/SelectorEpoll.swift
hassila: Yep, should do it like that.

Sources/NIO/SelectorGeneric.swift
hassila: As the subclasses are "final" which should remove the dynamic dispatch according to https://developer.apple.com/swift/blog/?id=27

Sources/NIO/SelectorEpoll.swift
hassila: Fixed in e19e685.

Sources/NIO/SelectorEpoll.swift
hassila: Fixed in e19e685.

Sources/NIO/SelectorGeneric.swift
hassila: (didn't actually touch that, but fixed in e19e685)

Sources/NIO/SelectorKqueue.swift
hassila: Fixed in e19e685.

Sources/NIO/SelectorKqueue.swift
hassila: Fixed in e19e685.

docker/docker-compose.1804.50.yaml
hassila: It seems that there is something with the subclasses that pushed the 5.0 to generate slightly more mallocs in some cases, but not in 5.1 and later. Perhaps some missing optimisation in 5.0.

Sources/NIO/SelectorGeneric.swift
hassila: Fixed in e19e685.

Sources/NIO/SelectorGeneric.swift
hassila: Also, the reason for subclassing is to be able to at runtime provide a different type (such that I can return i.e. UringSelector... But fall back on epoll if it can't initialise.) - that does not work with the typaalias as far as I understand, as #if is at compile time only?

hassila: > if we did sub-classes, then we should probably (depending on `#if os` etc return the concrete type of the selector here, not `NIO.Selector<R>`)

I think it is required for `SelectorTest.swift` where we do `let selector : NIO.Selector<TestRegistration> = try NIO.Selector<TestRegistration>.defaultSelectorFactory()`.

weissi: I think you misunderstood that Swift blog post a little. It says that the compiler doesn't need to worry about subclasses it can't see with WMO (and that is true). So if you have

```
class Foo {
   func f() { ... }
}

@inline(never)
func getFoo() -> Foo {
   return Foo()
}
```

and then you do `getFoo().f()`, then the compiler can put a direct call to `Foo.f` there (it doesn't need to worry about a potential subclass `Bar: Foo`.

In our case however, we do have actual subclasses. In this case the de-virtualisation only works if the compiler can fully inline (and therefore see through) the method that selects the `Selector`. This may be possible to get right but it's hard and fragile and we have no good way of testing this. As a demonstration, consider this tiny program:

test-wmo-devirt.swift:
```swift
import Foo

doit()

// MODULE: Foo

class Base {
    func f() {
        print("hio")
    }
}

class Sub: Base {
    override func f() {
        print("X")
    }
}

public func doit() {
    let b = select()
    b.f()
}

@inline(never)
func select() -> Base {
    return Sub()
}
```

if you compile this (you can run it through `swift-nio/dev/make-single-file-spm test-wmo-devirt.swift build`, you'll get

```
$ otool -tV /tmp/test_package_ZG8iPz/.build/release/TestApp | swift demangle | grep -A10 ^Foo.doit
Foo.doit() -> ():
0000000100003ed0	pushq	%rbp
0000000100003ed1	movq	%rsp, %rbp
0000000100003ed4	pushq	%r13
0000000100003ed6	pushq	%rax
0000000100003ed7	callq	Foo.select() -> Foo.Base
0000000100003edc	movq	%rax, %r13
0000000100003edf	movq	(%rax), %rax
0000000100003ee2	callq	*0x50(%rax)   <--- virtual call
0000000100003ee5	movq	%r13, %rdi
0000000100003ee8	addq	$0x8, %rsp
```

see the virtual call up there?

Said that, I don't think a virtual call will actually kill us but it does introduce some overhead that we could avoid if we wanted to.

Regarding epoll/uring at runtime: Yes, that's true if we did a runtime switch, then we do need the dynamism. The question is if we actually want to switch at runtime of if we should just have users select io_uring with `swift build -Xswiftc -DSWIFT_NIO_ENABLE_IO_URING` or so...

docker/docker-compose.1804.50.yaml
weissi: yeah, that's fine by me although I'd like to consciously make the decision that we're giving up some performance here. We should probably have a look at the code because not only do we have subclasses, we also have a generic subclass which can break specialisation which could be a real issue.

Sources/NIO/SelectorGeneric.swift
weissi: actually, specialisation will also be an issue probably, let me write some more code to check.

hassila: It was my understanding that SPM in release mode did WMO and then it should see it?

hassila: But you did not mark the classes as `final` in your test?

hassila: (one reason why I liked the runtime switch is that it makes it easier to compare performance)

hassila: The specific point from the blog post was 

> Use final when you know that a declaration does not need to be overridden.
> The final keyword is a restriction on a class, method, or property that indicates that the declaration cannot be overridden. This allows the compiler to safely elide dynamic dispatch indirection. For instance, in the following point and velocity will be accessed directly through a load from the object’s stored property and updatePoint() will be called via a direct function call.

hassila: But you are right, tried the sample with a final and got the same result as you...

weissi: Yes, specialisation won't work because Swift never generates specialised vtables. See this massive difference:

This program:

```swift
import Foo

doit()

// MODULE: Foo
import Darwin

class Base<I: FixedWidthInteger> {
    @inline(never)
    func f(_ l: I, _ r: I) {
        _ = fcntl(CInt(0xff), CInt(0xdd), CInt(l &+ r))
    }
}

class Sub<I: FixedWidthInteger>: Base<I> {
    @inline(never)
    override func f(_ l: I, _ r: I) {
        _ = fcntl(CInt(0xff), CInt(0xdd), CInt(l &+ r))
    }
}

public func doit() {
    let b = selectBest()
    b.f(100, 200)
}

func selectBest() -> Base<CInt> {
    return Sub<CInt>()
}
```

generates:

```
$ otool -tV /tmp/test_package_xGrpAe/.build/release/TestApp | swift demangle | grep -A10 -e ^Foo.doit -e 'of generic specialization <Swift.Int32> of Foo.Sub.f'
Foo.doit() -> ():
0000000100003d70	pushq	%rbp
0000000100003d71	movq	%rsp, %rbp
0000000100003d74	movl	$0x64, %edi
0000000100003d79	movl	$0xc8, %esi
0000000100003d7e	popq	%rbp
0000000100003d7f	jmp	function signature specialization <Arg[2] = Dead> of generic specialization <Swift.Int32> of Foo.Sub.f(A, A) -> ()
0000000100003d84	nopw	%cs:(%rax,%rax)
0000000100003d8e	nop
function signature specialization <Arg[2] = Dead> of generic specialization <Swift.Int32> of Foo.Sub.f(A, A) -> ():
0000000100003d90	pushq	%rbp
0000000100003d91	movq	%rsp, %rbp
0000000100003d94	leal	(%rdi,%rsi), %edx
0000000100003d97	movl	$0xff, %edi
0000000100003d9c	movl	$0xdd, %esi
0000000100003da1	popq	%rbp
0000000100003da2	jmp	0x100007164                     ## symbol stub for: Darwin.fcntl(Swift.Int32, Swift.Int32, Swift.Int32) -> Swift.Int32
0000000100003da7	nopw	(%rax,%rax)
type metadata instantiation function for Foo.Base:
0000000100003db0	jmp	0x1000072c6                     ## symbol stub for: _swift_allocateGenericClassMetadata
```

but if you stick an `@inline(never)` in front of `func selectBest`, you'll get this horror:

```
Foo.doit() -> ():
0000000100003bc0	pushq	%rbp
0000000100003bc1	movq	%rsp, %rbp
0000000100003bc4	pushq	%r13
0000000100003bc6	subq	$0x18, %rsp
0000000100003bca	callq	Foo.selectBest() -> Foo.Base<Swift.Int32>
0000000100003bcf	movq	%rax, %r13
0000000100003bd2	movl	$0x64, -0x18(%rbp)
0000000100003bd9	movl	$0xc8, -0x10(%rbp)
0000000100003be0	movq	(%rax), %rax
0000000100003be3	leaq	-0x18(%rbp), %rdi
0000000100003be7	leaq	-0x10(%rbp), %rsi
0000000100003beb	callq	*0x60(%rax)
0000000100003bee	movq	%r13, %rdi
0000000100003bf1	addq	$0x18, %rsp
0000000100003bf5	popq	%r13
0000000100003bf7	popq	%rbp
0000000100003bf8	jmp	0x10000721e                     ## symbol stub for: _swift_release


Foo.Sub.f(A, A) -> ():
0000000100003450	pushq	%rbp
0000000100003451	movq	%rsp, %rbp
0000000100003454	pushq	%r15
0000000100003456	pushq	%r14
0000000100003458	pushq	%r13
000000010000345a	pushq	%r12
000000010000345c	pushq	%rbx
000000010000345d	subq	$0xa8, %rsp
0000000100003464	movq	%rsi, -0x90(%rbp)
000000010000346b	movq	%rdi, -0x40(%rbp)
000000010000346f	movq	(%r13), %rax
0000000100003473	movq	0x70(%rax), %r12
0000000100003477	movq	0x78(%rax), %rax
000000010000347b	movq	%rax, -0x88(%rbp)
0000000100003482	movq	0x8(%rax), %rbx
0000000100003486	movq	0x18(%rbx), %rax
000000010000348a	movq	0x10(%rax), %rsi
000000010000348e	movq	0x4b8b(%rip), %rcx              ## literal pool symbol address: protocol requirements base descriptor for Swift.ExpressibleByIntegerLiteral
0000000100003495	movq	0x4b6c(%rip), %r8               ## literal pool symbol address: associated type descriptor for Swift.ExpressibleByIntegerLiteral.IntegerLiteralType
000000010000349c	xorl	%edi, %edi
000000010000349e	movq	%rsi, -0x58(%rbp)
00000001000034a2	movq	%r12, %rdx
00000001000034a5	callq	0x1000071f4                     ## symbol stub for: _swift_getAssociatedTypeWitness
00000001000034aa	movq	%rax, -0x78(%rbp)
00000001000034ae	movq	-0x8(%rax), %rax
00000001000034b2	movq	0x40(%rax), %rax
00000001000034b6	addq	$0xf, %rax
00000001000034ba	andq	$-0x10, %rax
00000001000034be	movq	%rsp, %rcx
00000001000034c1	subq	%rax, %rcx
00000001000034c4	movq	%rcx, -0x70(%rbp)
00000001000034c8	movq	%rcx, %rsp
00000001000034cb	movq	-0x8(%r12), %rax
00000001000034d0	movq	%rax, -0x30(%rbp)
00000001000034d4	movq	0x40(%rax), %rax
00000001000034d8	movq	%rsp, %rcx
00000001000034db	addq	$0xf, %rax
00000001000034df	andq	$-0x10, %rax
00000001000034e3	subq	%rax, %rcx
00000001000034e6	movq	%rcx, -0x98(%rbp)
00000001000034ed	movq	%rcx, %rsp
00000001000034f0	movq	%rsp, %rcx
00000001000034f3	subq	%rax, %rcx
00000001000034f6	movq	%rcx, -0xc0(%rbp)
00000001000034fd	movq	%rcx, %rsp
0000000100003500	movq	%rsp, %rcx
0000000100003503	subq	%rax, %rcx
0000000100003506	movq	%rcx, -0x60(%rbp)
000000010000350a	movq	%rcx, %rsp
000000010000350d	movq	%rsp, %rcx
0000000100003510	subq	%rax, %rcx
0000000100003513	movq	%rcx, -0x68(%rbp)
0000000100003517	movq	%rcx, %rsp
000000010000351a	movq	%rsp, %rcx
000000010000351d	subq	%rax, %rcx
0000000100003520	movq	%rcx, -0xc8(%rbp)
0000000100003527	movq	%rcx, %rsp
000000010000352a	movq	%rsp, %rcx
000000010000352d	subq	%rax, %rcx
0000000100003530	movq	%rcx, -0x80(%rbp)
0000000100003534	movq	%rcx, %rsp
0000000100003537	movq	%rsp, %rcx
000000010000353a	subq	%rax, %rcx
000000010000353d	movq	%rcx, -0xa0(%rbp)
0000000100003544	movq	%rcx, %rsp
0000000100003547	movq	%rsp, %rcx
000000010000354a	subq	%rax, %rcx
000000010000354d	movq	%rcx, -0x50(%rbp)
0000000100003551	movq	%rcx, %rsp
0000000100003554	movq	%rsp, %r15
0000000100003557	subq	%rax, %r15
000000010000355a	movq	%r15, %rsp
000000010000355d	movq	%rsp, %r14
0000000100003560	subq	%rax, %r14
0000000100003563	movq	%r14, %rsp
0000000100003566	movq	%r14, %rdi
0000000100003569	movq	-0x90(%rbp), %rsi
0000000100003570	movq	-0x40(%rbp), %r13
0000000100003574	movq	%r12, %rdx
0000000100003577	movq	-0x88(%rbp), %rcx
000000010000357e	callq	0x10000706e                     ## symbol stub for: dispatch thunk of Swift.FixedWidthInteger.addingReportingOverflow(A) -> (partialValue: A, overflow: Swift.Bool)
0000000100003583	movq	%r12, %r13
0000000100003586	movq	%r12, %rdi
0000000100003589	movq	%rbx, -0x38(%rbp)
000000010000358d	movq	%rbx, %rsi
0000000100003590	callq	0x100007068                     ## symbol stub for: dispatch thunk of static Swift.BinaryInteger.isSigned.getter : Swift.Bool
0000000100003595	movl	%eax, %ebx
0000000100003597	movq	-0x30(%rbp), %rax
000000010000359b	movq	0x10(%rax), %rax
000000010000359f	movq	%r15, %rdi
00000001000035a2	movq	%r14, -0x48(%rbp)
00000001000035a6	movq	%r14, %rsi
00000001000035a9	movq	%r12, %rdx
00000001000035ac	movq	%rax, -0x40(%rbp)
00000001000035b0	callq	*%rax
00000001000035b2	testb	$0x1, %bl
00000001000035b5	je	0x10000384c
00000001000035bb	movq	%r15, %r13
00000001000035be	movq	%r12, %rdi
00000001000035c1	movq	-0x38(%rbp), %rsi
[...]
```

weissi: @hassila `final` doesn't matter for non-public things (because Swift can infer it with WMO, unless you do `@testable import`). The key thing here is that Swift _can't_ devirtualise this if it's legitimately dynamic. Well I guess it could scan the code and see that we only ever instantiate `Sub()` and never `Base()` and then Swift could pretend that `Base` doesn't exist at all. But this optimisation doesn't exist today.

But honestly, the virtual call is the smaller of the issues, the real issue is that Swift (right now, it could in theory) never generates vtables of that contain specialised functions. So with the pattern above in my test program (which I think fairly closely mirrors `Selector` with sub classes) we'll never get specialisation anymore and I think that's actually an issue. So we'd need to code the program in a way that the optimiser can see through it and fully devirtualise it. I'm pretty sure that's possible but annoying and brittle (because we can't easily test it).

And yes: SwiftPM always uses WMO, that blog post is written for Xcode where you can switch that on/off.

hassila: Ok, thanks a lot for clarification - starting to learn the lesson that one should be careful with generics and double-check actual codegen (wonder if Im the only one who misunderstood that article)... :-)

I'll look at using extensions then instead (not sure why the type alias would be needed at all though, except possibly for extra clarity, even if the implementations now are in dedicated files? )

hassila: @weissi could you please just clarify this comment:

> and add the default functions using `extension Selector`. The only duplication would be the stored properties that we'd then add to each Selector individually.

my understanding is that extensions doesn't allow for stored properties, so do you propose to put all the properties in the Selector class directly or am I misunderstanding?

weissi: > I'll look at using extensions then instead (not sure why the type alias would be needed at all though, except possibly for extra clarity, even if the implementations now are in dedicated files? )

I think the `typealias` will help to keep the `#if os` switch in one place and also to make the `extension Selector` work.


> my understanding is that extensions doesn't allow for stored properties, so do you propose to put all the properties in the Selector class directly or am I misunderstanding?

Yeah, I think we have to c&p the properties which I think should be the only [duplication](https://github.com/apple/swift-nio/pull/1795#discussion_r609596641).

weissi: It's really unfortunate we don't have an `abstract class` in Swift which we could use to tell the optimiser that these can't ever be initialised. I filed [SR-14458](https://bugs.swift.org/browse/SR-14458) because previously there was only an internal radar for that.

Lukasa: We can also conform the selector implementations to `SelectorProtocol` and implement the code in `extension SelectorProtocol`. This would let us constrain exactly what a `Selector` requires. We never have to actually _use_ the protocol: it exists just to provide compiler-aided copy-paste, like we do in HTTP/2.

hassila: > We can also conform the selector implementations to `SelectorProtocol` and implement the code in `extension SelectorProtocol`. This would let us constrain exactly what a `Selector` requires. We never have to actually _use_ the protocol: it exists just to provide compiler-aided copy-paste, like we do in HTTP/2.

Super, because that was actually what I was doing already :-) Will hope to let you have a look soon, some issues with HookedSelector as it actually subclasses, which creates some trouble.

weissi: Great point! That also follows what we already do with `protocol _BSDSocketProtocol`

hassila:
Ok, new try in eaf83ab without subclasses.

docker/docker-compose.1804.50.yaml
hassila: @weissi do you want me to add the malloc checks back for 5.0/5.1?

Sources/NIO/SelectorEpoll.swift
```diff
@@ -0,0 +1,248 @@
...
+            selectorEventSet.formUnion(.readEOF)
+        }
+        if epollEvent.events & Epoll.EPOLLHUP != 0 || epollEvent.events & Epoll.EPOLLERR != 0 {
+            selectorEventSet.formUnion(.reset)
+        }
+        self = selectorEventSet
+    }
+}
+
+extension Selector : _SelectorBackendProtocol {
```
weissi, line 96:
```suggestion
extension Selector: _SelectorBackendProtocol {
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    /// Apply the given `SelectorStrategy` and execute `body` once it's complete (which may produce `SelectorEvent`s to handle).
+    ///
+    /// - parameters:
+    ///     - strategy: The `SelectorStrategy` to apply
+    ///     - body: The function to execute for each `SelectorEvent` that was produced.
+    func whenReady(strategy: SelectorStrategy, _ body: (SelectorEvent<R>) throws -> Void) throws -> Void {
+        try _whenReady(strategy: strategy, body)
```
weissi, line 266:
```suggestion
        try self._whenReady(strategy: strategy, body)
```

NIO has mandatory `self.`

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+            throw IOError(errnoCode: EBADF, reason: "can't close selector as it's \(self.lifecycleState).")
+        }
+        try self._close()
+        self.lifecycleState = .closed
+        self.registrations.removeAll()
+    }
+
+    /* attention, this may (will!) be called from outside the event loop, ie. can't access mutable shared state (such as `self.open`) */
+    func wakeup() throws {
+        try _wakeup()
```
weissi, line 284:
```suggestion
        try self._wakeup()
```

Sources/NIO/SelectorGeneric.swift
hassila: I think is (hopefully) resolved with the latest commit, resolving this as the main point is addressed (new comments on the new commit if something weird).

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+    #endif
+}()
+
+/// This protocol defines the methods that are expected to be found on
+/// `Selector`. While defined as a protocol there is no expectation that any
+/// object other than `Selector` will implement this protocol: instead, this
+/// protocol acts as a reference for what new supported selector backends
+/// must implement.
+protocol _SelectorBackendProtocol {
+    associatedtype R : Registration
```
weissi, line 98:
```suggestion
    associatedtype R: Registration
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -0,0 +1,248 @@
...
+        timerev.data.fd = self.timerFD
+        try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: self.timerFD, event: &timerev)
+    }
+
+    func _deinit() {
+        assert(self.eventFD == -1, "self.eventFD == \(self.eventFD) on Epoll _deinit, forgot close?")
+        assert(self.timerFD == -1, "self.timerFD == \(self.timerFD) on Epoll _deinit, forgot close?")
+    }
+
+    func _register<S: Selectable>(selectable : S, fd: Int, interested: SelectorEventSet) throws {
```
weissi, line 121:
```suggestion
    func _register<S: Selectable>(selectable: S, fd: Int, interested: SelectorEventSet) throws {
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -0,0 +1,248 @@
...
+        var ev = Epoll.epoll_event()
+        ev.events = interested.epollEventSet
+        ev.data.fd = Int32(fd)
+
+        try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: ev.data.fd, event: &ev)
+    }
+
+    func _reregister<S: Selectable>(selectable : S, fd: Int, oldInterested: SelectorEventSet, newInterested: SelectorEventSet) throws {
```
weissi, line 129:
```suggestion
    func _reregister<S: Selectable>(selectable: S, fd: Int, oldInterested: SelectorEventSet, newInterested: SelectorEventSet) throws {
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+    func register<S: Selectable>(selectable: S, interested: SelectorEventSet, makeRegistration: (SelectorEventSet) -> R) throws {
+        assert(self.myThread == NIOThread.current)
+        assert(interested.contains(.reset))
+        guard self.lifecycleState == .open else {
+            throw IOError(errnoCode: EBADF, reason: "can't register on selector as it's \(self.lifecycleState).")
+        }
+
+        try selectable.withUnsafeHandle { fd in
+            assert(registrations[Int(fd)] == nil)
+            try self._register(selectable : selectable, fd: Int(fd), interested: interested)
```
weissi, line 215:
```suggestion
            try self._register(selectable: selectable, fd: Int(fd), interested: interested)
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+    ///     - interested: The `SelectorEventSet` in which we are interested and want to be notified about.
+    func reregister<S: Selectable>(selectable: S, interested: SelectorEventSet) throws {
+        assert(self.myThread == NIOThread.current)
+        guard self.lifecycleState == .open else {
+            throw IOError(errnoCode: EBADF, reason: "can't re-register on selector as it's \(self.lifecycleState).")
+        }
+        assert(interested.contains(.reset), "must register for at least .reset but tried registering for \(interested)")
+        try selectable.withUnsafeHandle { fd in
+            var reg = registrations[Int(fd)]!
+            try self._reregister(selectable : selectable, fd: Int(fd), oldInterested: reg.interested, newInterested: interested)
```
weissi, line 233:
```suggestion
            try self._reregister(selectable: selectable, fd: Int(fd), oldInterested: reg.interested, newInterested: interested)
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+            reg.interested = interested
+            registrations[Int(fd)] = reg
```
weissi, line 235:
```suggestion
            self.registrations[Int(fd)] = reg
```

Sources/NIO/SelectorKqueue.swift
```diff
@@ -0,0 +1,323 @@
...
+                    index += 1
+                }
+            }
+            try body(UnsafeMutableBufferPointer(start: keventBuffer, count: index))
+        }
+    }
+}
+
+/* this is deliberately not thread-safe, only the wakeup() function may be called unprotectedly */
+extension Selector : _SelectorBackendProtocol {
```
weissi, line 94:
```suggestion
extension Selector: _SelectorBackendProtocol {
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+/// This protocol defines the methods that are expected to be found on
+/// `Selector`. While defined as a protocol there is no expectation that any
+/// object other than `Selector` will implement this protocol: instead, this
+/// protocol acts as a reference for what new supported selector backends
+/// must implement.
+protocol _SelectorBackendProtocol {
+    associatedtype R : Registration
+    func _init() throws
+    func _deinit()
+    func _register<S: Selectable>(selectable: S, fd: Int, interested: SelectorEventSet) throws
```
weissi, line 101: typically `_functionName` means that a function has a higher access control level than it should have (for example `public func _foo` means "this should really be internal but can't be for some reason").

In this case I'm not sure if they should be `_`'d. I think we should probably call them `registerImpl` or `register0`, the `functionName0` being a Netty-ism...

Said that, I can live with `_functionName` too. @Lukasa wdyt?

Sources/NIO/SelectorGeneric.swift
```diff
@@ -0,0 +1,376 @@
...
+}()
+
+/// This protocol defines the methods that are expected to be found on
+/// `Selector`. While defined as a protocol there is no expectation that any
+/// object other than `Selector` will implement this protocol: instead, this
+/// protocol acts as a reference for what new supported selector backends
+/// must implement.
+protocol _SelectorBackendProtocol {
+    associatedtype R : Registration
+    func _init() throws
```
weissi, line 99: I'd probably give `_init` and `_deinit` better names for what they're meant to do. Maybe `initialiseState` and `validateAssertionsOnDeinit` or something? Dunno, not super important, I'll leave this up to you.

weissi (commented):
This is great! Left a few (mostly formatting) comments, most of my comments you can just apply the suggested change (removes some spaces usually) :)

docker/docker-compose.1804.50.yaml
weissi: I don't mind having them or not. Said that, I think allocation changes can highlight performance changes even if they only change for some Swift versions. Given that we expect precisely 0 perf change with this PR, I think verifying that the old ones still work could be useful. But I'm happy without them too.

Sources/NIO/SelectorGeneric.swift
hassila: Sure, updating.

Sources/NIO/SelectorGeneric.swift
hassila: Just let me know what you prefer - probably would go for the Netty-ism in that case if you don't like`_Name` as it is used in other places of the code (I've tried to decipher the `name0` as never encountered that, so thanks for pointing out origin).

weissi: `_Name` is only used for "is public, should be internal" or "is internal, should be private (usually to allow @usableFromInline)" in NIO so far. Most other NIO code uses the `name0`. So I too think that we should follow this

hassila:
Ok, applied all suggested changes and merged into a single commit for nicer merge.

weissi (approved):
Thank you very much, LGTM!

Sources/NIO/SelectorGeneric.swift
hassila: Done in latest.

docker/docker-compose.1804.50.yaml
hassila: They’re restored now.

weissi:
@Lukasa are you okay with this?

Lukasa (approved):
Yup, seems good to me.

PR #1796, opened 2021-04-09
Update documentation related to allocation troubleshooting on Linux and the ability to run unit tests in parallel

hassila:
### Motivation:

It's not obvious how to do analysis of allocation regressions on Linux, so some documentation would be helpful.
The run time for the unit tests can be quite long (a couple of minutes), can be cut down significantly if running tests in parallel, adding a hint about that.

### Modification:

Updated documentation.

### Result:

Easier to troubleshoot mallocs on Linux and faster running of unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,3 +353,10 @@ dnf install swift-lang /usr/bin/nc /usr/bin/lsof /usr/bin/shasum
 [repo-nio-ssl]: https://github.com/apple/swift-nio-ssl
 [repo-nio-transport-services]: https://github.com/apple/swift-nio-transport-services
 [repo-nio-ssh]: https://github.com/apple/swift-nio-ssh
+
+### Speeding up testing
+It's possible to run the test suite in parallel, it can save significant time if you have a larger multi-core machine, just add `--parallel` when running the tests. This can speed up the run time of the test suite with 30x or more.
+
+```
+swift test --parallel
+```
--- a/docs/debugging-allocations.md
+++ b/docs/debugging-allocations.md
@@ -311,3 +311,125 @@ before: 10000, after: 20000
 ```
 
 Now we see there's another stacktrace in the `AFTER` section which has no corresponding stacktrace in `BEFORE`. From the stack we can see it's originating from `doRequests(group:number:)`. In this instance we were working on options applied in this function so it appears we have added allocations.  We have also increased the number of allocations which are reported in the different numbers section where stack traces have been paired but with different numbers of allocations.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -353,3 +353,9 @@ dnf install swift-lang /usr/bin/nc /usr/bin/lsof /usr/bin/shasum
 [repo-nio-ssl]: https://github.com/apple/swift-nio-ssl
 [repo-nio-transport-services]: https://github.com/apple/swift-nio-transport-services
 [repo-nio-ssh]: https://github.com/apple/swift-nio-ssh
+
+### Speeding up testing
+It's possible to run the test suite in parallel, it can save significant time if you have a larger multi-core machine, just add ```--parallel``` when running the tests. This can speed up the run time of the test suite with 30x or more.
```
glbrntt, line 358:
```suggestion
It's possible to run the test suite in parallel, it can save significant time if you have a larger multi-core machine, just add `--parallel` when running the tests. This can speed up the run time of the test suite with 30x or more.
```

README.md
```diff
@@ -353,3 +353,9 @@ dnf install swift-lang /usr/bin/nc /usr/bin/lsof /usr/bin/shasum
...
+
+```swift test --parallel
+```
```
glbrntt, line 361:
````suggestion
```
swift test --parallel
```
````

File: docs/debugging-allocations.md [hunk 3]
```diff
@@ -311,3 +311,206 @@ before: 10000, after: 20000
...
+
+```
+heaptrack_print -T -d heaptrack.test_1000_autoReadGetAndSet.84341.gz heaptrack.test_1000_autoReadGetAndSet.84372.gz | swift demangle
+```
+`-T` gives us the temporary allocations (as it in this case was not a leak, but a transient alloaction - if you have leaks remove `-T`).
+
+The output can be quite long, but in this case as we look for transient allocations, scroll down to:
+
+```
+MOST TEMPORARY ALLOCATIONS
```
glbrntt, line 390: I'd recommend truncating some of the output like in the other examples, it's quite a wall of text otherwise!

glbrntt (commented):
This is really helpful, thank you! I left a couple of suggestions inline.

docs/debugging-allocations.md
hassila: Truncated and updated according to suggestions in b674ac0 now.

glbrntt (approved):
Looks great, thank you!

weissi (commented):
awesome, thank you!

PR #1797, opened 2021-04-09
Fix bug in eventfd(initval: Int32, flags: Int32)

hassila:
### Motivation:

There is bug in `eventfd(initval: Int32, flags: Int32)` where it doesn't actually pass on the arguments to `CNIOLinux.eventfd()` disregarding the intent of the caller.

### Modification:

Start using the arguments passed in, fixed type to be unsigned as eventfd expects on Linux.

### Result:

`eventfd(initval: Int32, flags: Int32)` now works as expected.


Lukasa:
Nice fix! Is there a good way for us to unit test this so we can't revert the change by accident?

hassila:
Not sure, fundamentally what one would like is the ability for the compiler to generate a warning for unused arguments in a function - that could happen anywhere really. Thinking of it, I'd probably want to have such a warning by default and have to explicitly have a `unused` annotation on an argument (e.g. if overriding a function and not actually needing one of the arguments) - it would give analogous help like the `override` keyword and help avoid such bugs in the first place.

hassila:
Opened this discussion at Swift forums:

https://forums.swift.org/t/add-warnings-for-unused-function-arguments-by-default/47271

weissi (approved):
Nice one, thank you! This will be a bit hard to test I guess so I'm happy without test in this case.

hassila:
Thanks. For the record the only call site is here:

https://github.com/apple/swift-nio/blob/22b58cebee26680b2601ae1310b6516599ec1f22/Sources/NIO/Selector.swift#L337

Davidde94:
@Lukasa doesn't changing the flags argument from `Int32` to `UInt32` change the API?

glbrntt:
> @Lukasa / @weissi doesn't changing the flags argument from `Int32` to `UInt32` change the API?

@Davidde94 it's `public` on an `internal` type, I have a PR lined up to fix a bunch of these but wanted to wait for this to go in first

PR #1798, opened 2021-04-09
Create `EmbeddedChannel` with multiple handlers

Davidde94:
Add an initialiser to `EmbeddedChannel` that adds multiple `ChannelHandler`s

### Motivation:

I have several use-cases in my tests where it's necessary to have an `EmbeddedChannel` with multiple `ChannelHandler`s.

### Modifications:

Add an initialiser that takes an array of `ChannelHandler`s, and adds them to the `ChannelPipeline`.

### Result:

Previously:
```swift
let embeddedChannel = EmbeddedChannel(handler: ...)
_ = try! embeddedChannel.pipeline.addHandlers([...]).wait()
```

Now:
```swift
let embeddedChannel = EmbeddedChannel(handlers: ...)
```

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -618,17 +618,9 @@ public final class EmbeddedChannel: Channel {
     /// - parameters:
     ///     - handler: The `ChannelHandler` to add to the `ChannelPipeline` before register or `nil` if none should be added.
     ///     - loop: The `EmbeddedEventLoop` to use.
-    public init(handler: ChannelHandler? = nil, loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
-        self.embeddedEventLoop = loop
-        self._pipeline = ChannelPipeline(channel: self)
-
-        if let handler = handler {
-            // This will be propagated via fireErrorCaught
-            _ = try? _pipeline.addHandler(handler).wait()
-        }
-
-        // This will never throw...
-        try! register().wait()
+    public convenience init(handler: ChannelHandler? = nil, loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+        let handlers = handler == nil ? [] : [handler!]
+        self.init(handlers: handlers, loop: loop)
     }
     
     /// Create a new instance.
@@ -636,7 +628,7 @@ public final class EmbeddedChannel: Channel {
     /// During creation it will automatically also register itself on the `EmbeddedEventLoop`.
     ///
     /// - parameters:
-    ///     - handlesr: The `ChannelHandler`s to add to the `ChannelPipeline` before register or `nil` if none should be added.
+    ///     - handlers: The `ChannelHandler`s to add to the `ChannelPipeline` before register.
     ///     - loop: The `EmbeddedEventLoop` to use.
     public init(handlers: [ChannelHandler], loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
         self.embeddedEventLoop = loop
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -630,6 +630,23 @@ public final class EmbeddedChannel: Channel {
         // This will never throw...
         try! register().wait()
     }
+    
+    /// Create a new instance.
+    ///
+    /// During creation it will automatically also register itself on the `EmbeddedEventLoop`.
+    ///
+    /// - parameters:
+    ///     - handlesr: The `ChannelHandler`s to add to the `ChannelPipeline` before register or `nil` if none should be added.
+    ///     - loop: The `EmbeddedEventLoop` to use.
+    public init(handlers: [ChannelHandler], loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+        self.embeddedEventLoop = loop
+        self._pipeline = ChannelPipeline(channel: self)
+
+        _ = try? _pipeline.addHandlers(handlers).wait()
+
+        // This will never throw...
+        try! register().wait()
+    }
 
     /// - see: `Channel.setOption`
     @inlinable
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -634,7 +634,7 @@ public final class EmbeddedChannel: Channel {
         self.embeddedEventLoop = loop
         self._pipeline = ChannelPipeline(channel: self)
 
-        try! _pipeline.syncOperations.addHandlers(handlers)
+        try! self._pipeline.syncOperations.addHandlers(handlers)
 
         // This will never throw...
         try! register().wait()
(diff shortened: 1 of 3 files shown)


Sources/NIO/Embedded.swift
```diff
@@ -630,6 +630,23 @@ public final class EmbeddedChannel: Channel {
         // This will never throw...
         try! register().wait()
     }
+    
+    /// Create a new instance.
+    ///
+    /// During creation it will automatically also register itself on the `EmbeddedEventLoop`.
+    ///
+    /// - parameters:
+    ///     - handlesr: The `ChannelHandler`s to add to the `ChannelPipeline` before register or `nil` if none should be added.
```
glbrntt, line 639:
```suggestion
    ///     - handlers: The `ChannelHandler`s to add to the `ChannelPipeline` before register.
```

Sources/NIO/Embedded.swift
```diff
@@ -630,6 +630,23 @@ public final class EmbeddedChannel: Channel {
...
+    ///     - loop: The `EmbeddedEventLoop` to use.
+    public init(handlers: [ChannelHandler], loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+        self.embeddedEventLoop = loop
+        self._pipeline = ChannelPipeline(channel: self)
+
+        _ = try? _pipeline.addHandlers(handlers).wait()
+
+        // This will never throw...
+        try! register().wait()
+    }
```
Lukasa, line 649: Can we reimplement the existing `init` in terms of this one? It'll make it easier to understand and make sure that they do the same things.

Also mind adding a unit test for this API?

glbrntt (changes requested):
I think we should add at least one tests for this!

Sources/NIO/Embedded.swift
Davidde94: Sure to both.

I tried adding a unit test already, but found that `ChannelPipeline` doesn't expose easy functionality for getting the handlers. I can come up with something though.

Lukasa: `.pipeline.handler` works fine.

Lukasa: In this case the only guarantee of the API is that the handlers you pass through the initializer are present in the pipeline, so you can just try to get them out and confirm they're there.

Davidde94: Surely we should guarantee the order though right?

Davidde94: I have something that does check the order, it's not massively pretty

Lukasa: Yeah, guaranteeing the order is a bit awkward, but the easiest way is probably to have `InboundIn == InboundOut == Array[ChannelHandler]` and then just do `fireChannelRead(Array<ChannelHandler>())`. Then, each handler can `append` itself to the data and fire it on. Should do the job.

Sources/NIO/Embedded.swift
weissi, line 639:
```suggestion
    ///     - handlers: The `ChannelHandler`s to add to the `ChannelPipeline` before register or `nil` if none should be added.
```

Sources/NIO/Embedded.swift
```diff
@@ -630,6 +630,23 @@ public final class EmbeddedChannel: Channel {
...
+    
+    /// Create a new instance.
+    ///
+    /// During creation it will automatically also register itself on the `EmbeddedEventLoop`.
+    ///
+    /// - parameters:
+    ///     - handlesr: The `ChannelHandler`s to add to the `ChannelPipeline` before register or `nil` if none should be added.
+    ///     - loop: The `EmbeddedEventLoop` to use.
+    public init(handlers: [ChannelHandler], loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+        self.embeddedEventLoop = loop
```
weissi, line 642: could we have the only `init(handler: ChannelHandler...)` one call this version? There should only be one `init` which does the actual work.

weissi (commented):
totally happy taking this, left some comments + others have already commented

Sources/NIO/Embedded.swift
```diff
@@ -618,14 +618,23 @@ public final class EmbeddedChannel: Channel {
     /// - parameters:
     ///     - handler: The `ChannelHandler` to add to the `ChannelPipeline` before register or `nil` if none should be added.
     ///     - loop: The `EmbeddedEventLoop` to use.
-    public init(handler: ChannelHandler? = nil, loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+    public convenience init(handler: ChannelHandler? = nil, loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
+        let handlers = handler == nil ? [] : [handler!]
```
Lukasa, line 622: Might be nicer to avoid the `!`:

```suggestion
        let handlers = handler.map { [$0] } ?? []
```

Lukasa (approved):
One nit and then I'm happy.

glbrntt (approved):
Handy addition, thanks David! Just need the CI to turn green now 🙂

Sources/NIO/Embedded.swift
```diff
@@ -618,13 +618,24 @@ public final class EmbeddedChannel: Channel {
...
+    ///     - handlers: The `ChannelHandler`s to add to the `ChannelPipeline` before register.
+    ///     - loop: The `EmbeddedEventLoop` to use.
+    public init(handlers: [ChannelHandler], loop: EmbeddedEventLoop = EmbeddedEventLoop()) {
         self.embeddedEventLoop = loop
         self._pipeline = ChannelPipeline(channel: self)
 
-        if let handler = handler {
-            // This will be propagated via fireErrorCaught
-            _ = try? _pipeline.addHandler(handler).wait()
+        if !handlers.isEmpty {
```
weissi, line 637:
```suggestion
        // `addHandlers` cannot fail here, the Channel is definitely not closed.
        try! self._pipeline.addHandlers(handlers).wait()
```

Sources/NIO/Embedded.swift
```diff
@@ -618,13 +618,24 @@ public final class EmbeddedChannel: Channel {
...
+            _ = try? _pipeline.addHandlers(handlers).wait()
         }
```
weissi, line 639:
```suggestion
        // `addHandlers` cannot fail here, the Channel is definitely not closed.
        try! self._pipeline.addHandlers(handlers).wait()
```

weissi: - the `if` is superfluous
- shouldn't we also use the `.syncOperations` here? @glbrntt

Davidde94: Its to avoid allocations

weissi: @Davidde94 hence `.syncOperations`. But also: where do these allocs come from?

glbrntt: @weissi the allocation is because `EmbeddedEventLoop` doesn't have a cached void future: `addHandlers`. The `if` returns us to the same state as before: only allocate if we actually add a handler. 

But yeah, sync operations is best here.

Sources/NIO/Embedded.swift
```diff
@@ -634,9 +634,7 @@ public final class EmbeddedChannel: Channel {
         self.embeddedEventLoop = loop
         self._pipeline = ChannelPipeline(channel: self)
 
-        if !handlers.isEmpty {
-            _ = try? _pipeline.addHandlers(handlers).wait()
-        }
+        try! _pipeline.syncOperations.addHandlers(handlers)
```
weissi, line 637:
```suggestion
        try! self._pipeline.syncOperations.addHandlers(handlers)
```

missing `self.`

weissi (approved):
Very nice! Thank you

Issue #381, opened 2018-05-02
remove the `deregistrationsHappened` workaround from Selector


weissi:
we should remove the `deregistrationsHappened` workaround from `Selector` and replace it with something better.

PR #1799, opened 2021-04-09
Add sequenceIdentifier to Registrations to handle outdated events

hassila:
### Motivation:

Currently the selector implementation just discards all pending events from the kernel when a deregistration
has occurred during the event loop tick. This has been needed to not fail the test testWeDoNotDeliverEventsForPreviouslyClosedChannels.

A cleanup case has been around for a while at: https://github.com/apple/swift-nio/issues/381

And when discussing this in https://github.com/apple/swift-nio/pull/1788 (to solve https://github.com/apple/swift-nio/issues/1761)

We came to the conclusion that the solution with tagging registrations on the backend could not suit only io_uring, but also solve the issue robustly for kqueue and epoll.

### Modification:

Added sequenceIdentifier to the Registration and add support for it throughout.
Mask in sequence identifier into user data in both kqueue and epoll.
In kqueue, we just store the sequence identifier in the udata field (to support 32-bit properly).
In epoll, where we know we have 64-bits, we mask it in together with fd.
Removed all references to the old workaround from #381.

### Result:

New solution that doesn't need to discard all events when deregistrations are done during the event loop tick.

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -58,9 +58,15 @@ import CNIOWindows
 #endif
 
 /// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
+/// `sequenceIdentifier` is used by the event notification backends (kqueue, epoll, ...)
+/// to mark events to allow for filtering of received return values to not be delivered to a
+/// new `Registration` instance that receives the same file descriptor. Ok if it wraps.
+/// Needed for i.e. testWeDoNotDeliverEventsForPreviouslyClosedChannels to succeed.
+typealias RegistrationSequenceIdentifier = UInt32
 protocol Registration {
     /// The `SelectorEventSet` in which the `Registration` is interested.
     var interested: SelectorEventSet { get set }
+    var sequenceIdentifier: RegistrationSequenceIdentifier { get set }
 }
 
 protocol SockAddrProtocol {
--- a/Sources/NIO/SelectorEpoll.swift
+++ b/Sources/NIO/SelectorEpoll.swift
@@ -103,13 +103,13 @@ extension Selector: _SelectorBackendProtocol {
 
         var ev = Epoll.epoll_event()
         ev.events = SelectorEventSet.read.epollEventSet
-        ev.data.fd = self.eventFD
+        ev.data.u64 = UInt64(self.eventFD)
 
         try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: self.eventFD, event: &ev)
 
         var timerev = Epoll.epoll_event()
         timerev.events = Epoll.EPOLLIN | Epoll.EPOLLERR | Epoll.EPOLLRDHUP
-        timerev.data.fd = self.timerFD
+        timerev.data.u64 = UInt64(self.timerFD)
         try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: self.timerFD, event: &timerev)
     }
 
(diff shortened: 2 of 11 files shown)


Sources/NIO/SelectorEpoll.swift
```diff
@@ -118,23 +118,23 @@ extension Selector: _SelectorBackendProtocol {
         assert(self.timerFD == -1, "self.timerFD == \(self.timerFD) in deinitAssertions0, forgot close?")
     }
 
-    func register0<S: Selectable>(selectable: S, fd: Int, interested: SelectorEventSet) throws {
+    func register0<S: Selectable>(selectable: S, fd: Int, interested: SelectorEventSet, sequenceIdentifier: RegistrationSequenceIdentifier) throws {
         var ev = Epoll.epoll_event()
         ev.events = interested.epollEventSet
-        ev.data.fd = Int32(fd)
+        ev.data.u64 = UInt64(UInt64(sequenceIdentifier) << 32 + UInt64(fd)) // we put fd in lower 4 bytes to allow eventfd & timerfd to work as is
```
Lukasa, line 124: Nit: the outer `UInt64` constructor isn't necessary.
```suggestion
        ev.data.u64 = UInt64(sequenceIdentifier) << 32 + UInt64(fd) // we put fd in lower 4 bytes to allow eventfd & timerfd to work as is
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -118,23 +118,23 @@ extension Selector: _SelectorBackendProtocol {
...
 
-        try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: ev.data.fd, event: &ev)
+        try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: Int32(fd), event: &ev)
```
Lukasa, line 126: Nit: Avoid `Int32` when what we mean is `CInt`.

```suggestion
        try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: CInt(fd), event: &ev)
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -118,23 +118,23 @@ extension Selector: _SelectorBackendProtocol {
...
     }
 
-    func reregister0<S: Selectable>(selectable: S, fd: Int, oldInterested: SelectorEventSet, newInterested: SelectorEventSet) throws {
+    func reregister0<S: Selectable>(selectable: S, fd: Int, oldInterested: SelectorEventSet, newInterested: SelectorEventSet, sequenceIdentifier: RegistrationSequenceIdentifier) throws {
         var ev = Epoll.epoll_event()
         ev.events = newInterested.epollEventSet
-        ev.data.fd = Int32(fd)
+        ev.data.u64 = UInt64(UInt64(sequenceIdentifier) << 32 + UInt64(fd))
```
Lukasa, line 132:
```suggestion
        ev.data.u64 = UInt64(sequenceIdentifier) << 32 + UInt64(fd)
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -171,12 +171,11 @@ extension Selector: _SelectorBackendProtocol {
...
 
-        // start with no deregistrations happened
-        self.deregistrationsHappened = false
-        // temporary workaround to stop us delivering outdated events; possibly set in `deregister`
-        for i in 0..<ready where !self.deregistrationsHappened {
+        for i in 0..<ready {
             let ev = events[i]
-            switch ev.data.fd {
+            let fd = Int32(UInt64(ev.data.u64 & 0x00000000FFFFFFFF))
+            let eventSequenceIdentifier = UInt32(UInt64(ev.data.u64 >> 32))
```
Lukasa, line 177: We don't need the intermediary `UInt64` constructor here:

```suggestion
            let eventSequenceIdentifier = UInt32(ev.data.u64 >> 32)
```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -171,12 +171,11 @@ extension Selector: _SelectorBackendProtocol {
...
         }
 
-        // start with no deregistrations happened
-        self.deregistrationsHappened = false
-        // temporary workaround to stop us delivering outdated events; possibly set in `deregister`
-        for i in 0..<ready where !self.deregistrationsHappened {
+        for i in 0..<ready {
             let ev = events[i]
-            switch ev.data.fd {
+            let fd = Int32(UInt64(ev.data.u64 & 0x00000000FFFFFFFF))
```
Lukasa, line 176: We don't need the UInt64 constructor and we can safely just use `truncatingIfNeeded` here. Similarly, we should use `CInt` instead of `Int32`.

```suggestion
            let fd = CInt(truncatingIfNeeded: ev.data.u64)
```

Sources/NIO/SelectorGeneric.swift
```diff
@@ -121,18 +121,15 @@ protocol _SelectorBackendProtocol {
 /* this is deliberately not thread-safe, only the wakeup() function may be called unprotectedly */
 internal class Selector<R: Registration>  {
     var lifecycleState: SelectorLifecycleState
-
-    // temporary workaround to stop us delivering outdated events; read in `whenReady`, set in `deregister`
-    var deregistrationsHappened: Bool = false
     var registrations = [Int: R]()
+    var sequenceIdentifier : RegistrationSequenceIdentifier = 1
```
Lukasa, line 125: Pedantic nit, but NIO style:

```suggestion
    var sequenceIdentifier: RegistrationSequenceIdentifier = 1
```

Sources/NIO/BaseSocket.swift
```diff
@@ -58,9 +58,15 @@ import CNIOWindows
 #endif
 
 /// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
+/// `sequenceIdentifier` is used by the event notification backends (kqueue, epoll, ...)
+/// to mark events to allow for filtering of received return values to not be delivered to a
+/// new `Registration` instance that receives the same file descriptor. Ok if it wraps.
+/// Needed for i.e. testWeDoNotDeliverEventsForPreviouslyClosedChannels to succeed.
+typealias RegistrationSequenceIdentifier = UInt32
```
Lukasa, line 65: Hmm, this typealias is causing me some confusion I think, because we're a bit inconsistent with how we use it. IMO we should either promote this to an actual type (`struct` wrapping a `UInt32`) or drop the typealias. We could aim to be consistent, but autocorrect has an awkward tendency to fail to tell you to use the typealias but instead to use the underlying type.

Sources/NIO/SelectorGeneric.swift
```diff
@@ -121,18 +121,15 @@ protocol _SelectorBackendProtocol {
...
 
-    let externalSelectorFDLock = Lock()
-
+    let myThread: NIOThread
     // The rules for `self.selectorFD`, `self.eventFD`, and `self.timerFD`:
     // reads: `self.externalSelectorFDLock` OR access from the EventLoop thread
     // writes: `self.externalSelectorFDLock` AND access from the EventLoop thread
+    let externalSelectorFDLock = Lock()
```
Lukasa, line 131: These things got moved around a bit: can they be put back? Helps keep the `git blame` a bit more valuable.

weissi (approved):
Thank you very much, looks great!

Sources/NIO/SelectorKqueue.swift
```diff
@@ -202,24 +204,24 @@ extension Selector: _SelectorBackendProtocol {
...
         }
 
-        // start with no deregistrations happened
-        self.deregistrationsHappened = false
-        // temporary workaround to stop us delivering outdated events; possibly set in `deregister`
-        for i in 0..<ready where !self.deregistrationsHappened {
+        for i in 0..<ready {
             let ev = events[i]
             let filter = Int32(ev.filter)
+            let eventSequenceIdentifier = UInt32(Int(bitPattern: ev.udata))
```
Lukasa, line 215: We can skip the intermediate `Int`.

```suggestion
            let eventSequenceIdentifier = UInt32(bitPattern: ev.udata)
```

PR #1801, opened 2021-04-13
refactor registration IDs

weissi:
Motivation:

Instead of manual shifting / masking we can write the whole registration
ID code normally in Swift.

Modifications:

Refactored RegistrationID to use Swift instead of shifting / masking.

Result:

Nicer code.

### DIFF
--- a/Sources/NIO/BaseSocket.swift
+++ b/Sources/NIO/BaseSocket.swift
@@ -57,16 +57,10 @@ import struct WinSDK.socklen_t
 import CNIOWindows
 #endif
 
-/// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
-/// `sequenceIdentifier` is used by the event notification backends (kqueue, epoll, ...)
-/// to mark events to allow for filtering of received return values to not be delivered to a
-/// new `Registration` instance that receives the same file descriptor. Ok if it wraps.
-/// Needed for i.e. testWeDoNotDeliverEventsForPreviouslyClosedChannels to succeed.
-typealias RegistrationSequenceIdentifier = UInt32
 protocol Registration {
     /// The `SelectorEventSet` in which the `Registration` is interested.
     var interested: SelectorEventSet { get set }
-    var sequenceIdentifier: RegistrationSequenceIdentifier { get set }
+    var registrationID: SelectorRegistrationID { get set }
 }
 
 protocol SockAddrProtocol {
--- a/Sources/NIO/SelectorEpoll.swift
+++ b/Sources/NIO/SelectorEpoll.swift
@@ -93,6 +93,33 @@ extension SelectorEventSet {
     }
 }
 
+// EPollUserData supports (un)packing into an `UInt64` because epoll has a user info field that we can attach which is
+// up to 64 bits wide. We're using all of those 64 bits, 32 for a "registration ID" and 32 for the file descriptor.
+@usableFromInline struct EPollUserData {
+    @usableFromInline var registrationID: SelectorRegistrationID
+    @usableFromInline var fileDescriptor: CInt
+
+    @inlinable init(registrationID: SelectorRegistrationID, fileDescriptor: CInt) {
+        assert(MemoryLayout<UInt64>.size == MemoryLayout<EPollUserData>.size)
+        self.registrationID = registrationID
+        self.fileDescriptor = fileDescriptor
+    }
+
+    @inlinable init(rawValue: UInt64) {
+        let unpacked = IntegerBitPacking.unpackUInt32CInt(rawValue)
+        self = .init(registrationID: SelectorRegistrationID(rawValue: unpacked.0), fileDescriptor: unpacked.1)
+    }
+}
+
+extension UInt64 {
+    @inlinable
+    init(_ epollUserData: EPollUserData) {
+        let fd = epollUserData.fileDescriptor
+        assert(fd >= 0, "\(fd) is not a valid file descriptor")
+        self = IntegerBitPacking.packUInt32CInt(epollUserData.registrationID.rawValue, fd)
+    }
+}
+
 extension Selector: _SelectorBackendProtocol {
     func initialiseState0() throws {
         self.selectorFD = try Epoll.epoll_create(size: 128)
@@ -103,13 +130,15 @@ extension Selector: _SelectorBackendProtocol {
 
         var ev = Epoll.epoll_event()
         ev.events = SelectorEventSet.read.epollEventSet
-        ev.data.u64 = UInt64(self.eventFD)
+        ev.data.u64 = UInt64(EPollUserData(registrationID: .initialRegistrationID,
+                                           fileDescriptor: self.eventFD))
 
         try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: self.eventFD, event: &ev)
 
         var timerev = Epoll.epoll_event()
         timerev.events = Epoll.EPOLLIN | Epoll.EPOLLERR | Epoll.EPOLLRDHUP
-        timerev.data.u64 = UInt64(self.timerFD)
+        timerev.data.u64 = UInt64(EPollUserData(registrationID: .initialRegistrationID,
+                                                fileDescriptor: self.timerFD))
         try Epoll.epoll_ctl(epfd: self.selectorFD, op: Epoll.EPOLL_CTL_ADD, fd: self.timerFD, event: &timerev)
     }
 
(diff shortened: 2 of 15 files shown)


weissi:
CC @hassila

hassila:
Much nicer!

Sources/NIO/BaseSocket.swift
```diff
@@ -58,15 +58,39 @@ import CNIOWindows
 #endif
 
 /// A Registration on a `Selector`, which is interested in an `SelectorEventSet`.
-/// `sequenceIdentifier` is used by the event notification backends (kqueue, epoll, ...)
+/// `registrationID` is used by the event notification backends (kqueue, epoll, ...)
 /// to mark events to allow for filtering of received return values to not be delivered to a
 /// new `Registration` instance that receives the same file descriptor. Ok if it wraps.
 /// Needed for i.e. testWeDoNotDeliverEventsForPreviouslyClosedChannels to succeed.
-typealias RegistrationSequenceIdentifier = UInt32
+@usableFromInline struct SelectorRegistrationID: Equatable {
```
Lukasa, line 65: Nit: may as well be `Hashable`. Will save me from wondering why it isn't later on down the road.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
...
+        return self._backing.fileDescriptor
+    }
+
+    @inlinable init(registrationID: SelectorRegistrationID, fileDescriptor: CInt) {
+        assert(MemoryLayout<UInt64>.size == MemoryLayout<EPollUserData>.size)
+        self._backing = (registrationID, fileDescriptor)
+    }
+
+    @inlinable init(rawValue: UInt64) {
+        self._backing = unsafeBitCast(rawValue, to: type(of: self._backing))
```
Lukasa, line 114: I don't love the `unsafeBitCast` here or below. We don't need it: the optimiser is quite capable of handling extracting the relevant fields to the appropriate layout. This seems to fall into a place where we're doing something unsafe unnecessarily, and it'll set off alarm bells every time I see it in the future.

weissi: Hmm, `unsafeBitCast` gives us a guarantee that nothing else does: That `UInt64` and the backing has the _same size_. That's why I used it because I think it's actually more safe than doing some shifting/masking manually here. But yeah, it does have "Unsafe" in the name ;). Happy to add a comment though

weissi: Or of course refactor to not use `unsafeBitCast` but IMHO this is kinda the right choice here?

Lukasa: I don't agree that only `unsafeBitCast` can give us that guarantee. You wrote another line of code that can give that guarantee in the other `init`. 

But even if it were true I don't think that guarantee would be enough to justify the downsides. It makes this code much more dangerous to change in future, and that makes me very nervous.

weissi: @Lukasa I missed a "at compile time" . uBC does it at compile time whereas `assert` only does it at runtime (in debug). And I need to manually shift/mask.

Lukasa: uBC doesn't do it at compile time. Try out the following program:

```swift
func main() {
    let x = UInt64(5)
    let y = unsafeBitCast(x, to: UInt32.self)
    let a = UInt32(5)
    let b = unsafeBitCast(a, to: UInt64.self)
    print(y)
    print(b)
}

main()
```

This compiles just fine for me. It crashes at runtime, of course. As for "in debug", we can use `precondition` and get it in release mode too.

As for "you need to manually shift and mask": yes, you do. However, SSWG guidelines don't allow "I have to write more code" as a reason to deploy unsafe operations.

weissi: Okay, I was convinced that `error: Can't unsafeBitCast between types of different sizes` is compile time but seems like you're right! Will replace in this case. We should file a bug for this though, will do later.

Sources/NIO/BaseSocket.swift
```diff
@@ -58,15 +58,39 @@ import CNIOWindows
...
+    @usableFromInline var _id: UInt32
```
glbrntt, line 66: Do we need `_id` here? Doesn't `@usableFromInline internal private(set) var rawValue: UInt32` achieve the same?

weissi: @glbrntt no, this doesn't work, we do write to it

Sources/NIO/SelectorGeneric.swift
```diff
@@ -383,3 +385,37 @@ enum SelectorStrategy {
...
+/// Needed for i.e. testWeDoNotDeliverEventsForPreviouslyClosedChannels to succeed.
+@usableFromInline struct SelectorRegistrationID: Hashable {
+    @usableFromInline var _rawValue: UInt32
+
+    @inlinable var rawValue: UInt32 {
+        return self._rawValue
+    }
+
+    @inlinable static var initialRegistrationID: SelectorRegistrationID {
+        return SelectorRegistrationID(rawValue: .max)
```
weissi, line 402: @Lukasa I changed this to use `UInt32.max` so we're sure we run the overflow codepath always.

weissi: CC @hassila too, just in case you get reg ids that look weird :)

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
     }
 }
 
+@usableFromInline struct EPollUserData {
+    // We're backing this by a tuple to get standard C layout
```
Lukasa, line 97: This comment is now a lie.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
...
+
+
+    @inlinable init(registrationID: SelectorRegistrationID, fileDescriptor: CInt) {
+        assert(MemoryLayout<UInt64>.size == MemoryLayout<EPollUserData>.size)
+        self.registrationID = registrationID
+        self.fileDescriptor = fileDescriptor
+    }
+
+    @inlinable init(rawValue: UInt64) {
+        let registrationID = UInt32(truncatingIfNeeded: (rawValue & (UInt64(UInt32.max) << 32)) >> 32)
```
Lukasa, line 109: We don't need the mask here: shifting right by 32 is the same as masking off the low 32 bits and then shifting by 32.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
...
+        let fileDescriptor = CInt(truncatingIfNeeded: rawValue & UInt64(UInt32.max))
```
Lukasa, line 110: Relatedly, we don't need the mask here: `truncatingIfNeeded` applies it automatically.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
...
+        self = .init(registrationID: SelectorRegistrationID(rawValue: registrationID),
+                     fileDescriptor: fileDescriptor)
+    }
+}
+
+extension UInt64 {
+    @inlinable
+    init(_ epollUserData: EPollUserData) {
+        self = UInt64((UInt64(epollUserData.registrationID.rawValue) << 32) |
+                        (UInt64(UInt32(bitPattern: epollUserData.fileDescriptor))))
```
Lukasa, line 121: I know we discussed what should happen if we _receive_ `-1` from the kernel, but should it be possible for us to _pass_ `-1` to it? I think I'd rather the answer was no, so I kinda think that instead of `UInt32(bitPattern:)` here we should just promote to `UInt64` and crash if this is negative.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
     }
 }
 
+@usableFromInline struct EPollUserData {
+    // We're backing this by a tuple to get standard C layout
+    @usableFromInline var registrationID: SelectorRegistrationID
+    @usableFromInline var fileDescriptor: CInt
+
+
+    @inlinable init(registrationID: SelectorRegistrationID, fileDescriptor: CInt) {
```
Lukasa, line 102: Bit of a mild nit here, but all call sites of this appear to have `fileDescriptor` as an `Int` and so have to manually force it down to `CInt`. Should this init instead take `Int` and do the shrink here?

Lukasa (changes requested):

Sources/NIO/SelectorEpoll.swift
hassila: Actually, I thought about file descriptors representation in general. Shouldn’t it be a type alias used consistently (which i believe should be CInt to allow for negative/invalid fd). It seems like an overkill to use Int, no?

Lukasa: I don’t think the memory usage of `Int` in these spaces is an issue: either way it consumes a register, and in these cases it’s largely being passed around as an argument. However, I do agree that it would be nice to apply a type to it. In this case that might be a sensible follow-up proposal.

hassila: Ok. I was not concerned about the memory usage/register usage, but rather to remove the need to shrink it when doing systemcalls. Basically all backends except a CInt, so it would remove the need to convert it ever.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
     }
 }
 
+@usableFromInline struct EPollUserData {
+    // We're backing this by a tuple to get standard C layout
+    @usableFromInline var registrationID: SelectorRegistrationID
+    @usableFromInline var fileDescriptor: CInt
+
+
```
glbrntt, line 101:
```suggestion

```

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
     }
 }
 
+@usableFromInline struct EPollUserData {
```
glbrntt, line 96: FWIW without the context of the rest of the PR I don't think it's immediately obvious why we're packing/unpacking from a `UInt64`, a brief comment might be useful.

Sources/NIO/SelectorEpoll.swift
```diff
@@ -93,6 +93,35 @@ extension SelectorEventSet {
...
+        let fileDescriptor = CInt(truncatingIfNeeded: rawValue & UInt64(UInt32.max))
+
+        self = .init(registrationID: SelectorRegistrationID(rawValue: registrationID),
+                     fileDescriptor: fileDescriptor)
+    }
+}
+
+extension UInt64 {
+    @inlinable
+    init(_ epollUserData: EPollUserData) {
```
glbrntt, line 119: nit: I think having the `epollUserData` label here would actually be helpful

weissi: The [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/) say

>  In initializers that perform value preserving type conversions, omit the first argument label, e.g. Int64(someUInt32)

Sources/NIO/SelectorEpoll.swift
weissi: fixed, thanks!

Sources/NIO/SelectorEpoll.swift
weissi: removed, now uses the generic bitpacking

Sources/NIO/SelectorEpoll.swift
weissi: thanks, fixed.

Sources/NIO/SelectorEpoll.swift
weissi: @Lukasa how strong are you on this one. I agree with @hassila that all these backends can only handle `CInt`. So if anything, we should refactor the existing code to do `CInt` instead. I can do this in this PR or another one (which I think should be preferred). WDYT?

Lukasa: Another one is fine.

Sources/NIO/SelectorEpoll.swift
weissi: done

Sources/NIO/SelectorEpoll.swift
weissi: awesome, thanks --> https://github.com/apple/swift-nio/pull/1815

Sources/NIO/SelectorEpoll.swift
weissi: replaced by generic version, code is gone

glbrntt (approved):
LGTM

Lukasa (approved):

PR #1804, opened 2021-04-14
Added second implementation of liburing as discussed in #1761.

hassila:
Motivation:

As outlined in #1761, io_uring is a new async I/O facility on Linux that should have less overhead than epoll.

This commit includes a second stab at adding this to SwiftNIO with a much leaner approach than previous #1788. No more automatic detection and dynamic loading with dlopen/dlsym. We just require command line defines to pull in liburing and link against it directly instead. (thus all shims can be dropped, all external headers can be dropped, all code for dynamic loading / resolving can be dropped, ...)

Modifications:

Added Uring Selector implementation.

Added minimal liburing support shims.

Disabled one assert that trips during normal usage likely due to async nature of poll updates.

Support for both single shot polls (should work all the way back to 5.1 kernels, needs testing) and multishot streaming polls and modifications for polls (due in 5.13) for slightly better performance (and better impedance match to SwiftNIO usage)

Added extensive debug logs which can be enabled with -D compiler flags.

Adjusted tests.

Added documentation.

Result:

Basic liburing support is in place.


hassila:

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+    if (_check_capabilities() != 0) {
+        return -1;
+    }
+    return 0;
+}
+
+// We should look at cpu affinity and preferably bind the SQPOLL thread to a specific core
+// and support cgroup / cpu sets on Linux to isolate it - then we should set IORING_SETUP_SQ_AFF here.
+
+static pthread_mutex_t global_ring_mutex = PTHREAD_MUTEX_INITIALIZER;
```
hassila, line 84: Use pthread_once() instead of mutex.

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+#include <pthread.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <signal.h>
+#include <signal.h>
+#include <sys/poll.h>
+
+// Definitons from syscall.c in liburing, we'll just do a syscall that will fail
```
weissi, line 37: do we need this in this PR? If people don't have io_uring, then they shouldn't enable `SWIFTNIO_LIBURING`?

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+// Support functions for liburing
+
+// Check if this is needed, copied from shim.c to avoid possible problems due to:
+// Xcode's Archive builds with Xcode's Package support struggle with empty .c files
+// (https://bugs.swift.org/browse/SR-12939).
+void CNIOLinux_i_do_nothing_just_working_around_a_darwin_toolchain_bug2(void) {}
+
+#ifdef __linux__
+
+#ifdef SWIFTNIO_LIBURING
```
weissi, line 24: this is a super nit but I'd name it `SWIFTNIO_USE_IO_URING` or so? That we're (currently) using liburing is an impl detail in my mind. But again, this isn't important.

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+#if SWIFTNIO_LIBURING
+#if os(Linux)
+
+import CNIOLinux
+
+// we stuff the event type into the user data for the sqe together with
+// the fd and sequenceIdentifier to match the events without needing any
+// memory allocations or references.
+
+internal enum CqeEventType : Int {
```
weissi, line 24:
```suggestion
internal enum CQEEventType: Int {
```

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+    case poll = 1, pollModify, pollDelete // start with 1 to not get zero bit patterns for stdin
```
weissi, line 25: is the 0 bit pattern a problem?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+    var pollMask : UInt32
+    var sequenceIdentifier : UInt32
+    init () {
+        self.fd = -1
+        self.pollMask = 0
+        self.sequenceIdentifier = 0
+    }
+}
+
+struct fdEventKey: Hashable {
```
weissi, line 54: is this still necessary after #1801 ?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+    }
+}
+
+final internal class Uring {
+    internal static let POLLIN: CUnsignedInt = numericCast(CNIOLinux.POLLIN)
+    internal static let POLLOUT: CUnsignedInt = numericCast(CNIOLinux.POLLOUT)
+    internal static let POLLERR: CUnsignedInt = numericCast(CNIOLinux.POLLERR)
+    internal static let POLLRDHUP: CUnsignedInt = CNIOLinux_POLLRDHUP() // numericCast(CNIOLinux.POLLRDHUP) 
+    internal static let POLLHUP: CUnsignedInt = numericCast(CNIOLinux.POLLHUP)
+    internal static let POLLCANCEL: CUnsignedInt = 0xF0000000 // Poll cancelled, need to reregister for singleshot polls
```
weissi, line 70: can't we pull the same trick as with `POLLRDHUP`?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+
+    private var ring = io_uring()
+    private let ringEntries: CUnsignedInt = 8192
+    private let cqeMaxCount : UInt32 = 8192 // this is the max chunk of CQE we take.
+
+    var cqes : UnsafeMutablePointer<UnsafeMutablePointer<io_uring_cqe>?>
+    var fdEvents = [fdEventKey : UInt32]() // fd, sequence_identifier : merged event_poll_return
+    var emptyCqe = io_uring_cqe()
+
+    internal static let initializedUring: Bool = {
```
weissi, line 80: do we need this? We're always compatible if we have `SWIFTNIO_URING`, right?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+        CNIOLinux_check_compatibility() == 0
+    }()
+
+    internal func fd() -> Int32 {
+       return ring.ring_fd
+    }
+
+    func _dumpCqes(_ header:String, count: Int = 1)
+    {
```
weissi, line 89: NIO uses 1tbs style where the opening `{` go on the `func` line (in C & Swift)

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+        #if SWIFTNIO_LIBURING_DEBUG_DUMP_CQE
+        func _debugPrintCQE(_ s : String) {
```
weissi, line 91: generally, no spaces before `:`

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+        pthread_mutex_lock(&global_ring_mutex);
+        if (!global_ring.ring_fd)
+        {
+            // setup a small global ring whose fd we can reference to use shared kernel sqpoll thread for all rings
+            // we don't particulariy care if we fail, then no SQPOLL shared ring
+            (void) io_uring_queue_init(4, &global_ring, IORING_SETUP_SQPOLL); // should support IORING_SETUP_SQ_AFF
+        }
+        pthread_mutex_unlock(&global_ring_mutex);
+
+        if (global_ring.ring_fd) // use shared kernel thread if it exists, otherwise fallthrough to normal setup
```
weissi, line 100: don't think this is correct. This assumes that `0` isn't a valid fd which isn't true (you can close `stdin`).

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+        return -1;
+    }
+    return 0;
+}
+
+// We should look at cpu affinity and preferably bind the SQPOLL thread to a specific core
+// and support cgroup / cpu sets on Linux to isolate it - then we should set IORING_SETUP_SQ_AFF here.
+
+static pthread_mutex_t global_ring_mutex = PTHREAD_MUTEX_INITIALIZER;
+static struct io_uring global_ring; // shared small ring to be able to reuse SQPOLL kernel thread
```
weissi, line 85: let's 0 initialise this explicitly with `= {0};`

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+    unsigned flags)
+{
+    if (flags & IORING_SETUP_SQPOLL)
+    {
+        pthread_mutex_lock(&global_ring_mutex);
+        if (!global_ring.ring_fd)
+        {
+            // setup a small global ring whose fd we can reference to use shared kernel sqpoll thread for all rings
+            // we don't particulariy care if we fail, then no SQPOLL shared ring
+            (void) io_uring_queue_init(4, &global_ring, IORING_SETUP_SQPOLL); // should support IORING_SETUP_SQ_AFF
```
weissi, line 96: I think we should check the return code here instead of relying on `fd != 0` later.

Sources/CNIOLinux/liburing_shims.c
```diff
@@ -0,0 +1,125 @@
...
+
+// We should look at cpu affinity and preferably bind the SQPOLL thread to a specific core
+// and support cgroup / cpu sets on Linux to isolate it - then we should set IORING_SETUP_SQ_AFF here.
+
+static pthread_mutex_t global_ring_mutex = PTHREAD_MUTEX_INITIALIZER;
+static struct io_uring global_ring; // shared small ring to be able to reuse SQPOLL kernel thread
+int CNIOLinux_io_uring_queue_init(unsigned entries, struct io_uring *ring,
+    unsigned flags)
+{
+    if (flags & IORING_SETUP_SQPOLL)
```
weissi, line 89: what's the current state w.r.t. SQPOLL requiring `root`?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+        // IORING_SETUP_SQPOLL will be fundamentally useful in 5.13, as we no longer need elevated privileges as previously
+        // and CNIOLinux.io_uring_queue_init will setup a shared uring instance which allows us to use a single
+        // kernel sqpoll thread that is shared amongst all loops. NB, each process will have it's own polling thread
+        // if SQPOLL is enabled, so for hosts running multiple processes using it, some care should be taken to not overload.
+        #if SWIFTNIO_LIBURING_SQPOLL_ENABLED
+        let _sqpollEnabled : UInt32 = 1
+        #else
+        let _sqpollEnabled : UInt32 = 0
+        #endif
+        if (CNIOLinux_io_uring_queue_init(ringEntries, &ring, _sqpollEnabled ) != 0)
```
weissi, line 135:
```suggestion
        if (CNIOLinux_io_uring_queue_init(ringEntries, &ring, _sqpollEnabled) != 0)
```

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+         {
+             throw UringError.uringSetupFailure
+         }
+
+        _debugPrint("io_uring_queue_init \(self.ring.ring_fd)")
+     }
+
+    internal func io_uring_queue_exit() {
+        _debugPrint("io_uring_queue_exit \(self.ring.ring_fd)")
+        CNIOLinux.io_uring_queue_exit(&ring)
```
weissi, line 145: can all these `io_uring_*` calls not return `EINTR`?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+        #else
+        return false
+        #endif
+    }
+
+    // Adopting some retry code from queue.c from liburing with slight
+    // modifications - we never want to have to handle retries of
+    // SQE allocation in all places it could possibly occur.
+    // If the SQ ring is full, we may need to submit IO first
+    internal func get_sqe() -> UnsafeMutablePointer<io_uring_sqe>?
```
weissi, line 160:
```suggestion
    internal func getSQE() -> UnsafeMutablePointer<io_uring_sqe>?
```

apart from that, I think we should make this function a 

```
func withSEQ<R>(_ body: (UnsafeMutablePointer<io_uring_sqe>?) throws -> R) rethrows -> R
```

because otherwise we may need to worry about the lifetime of the pointer.

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+    {
+        // io_uring_submit can fail here due to backpressure from kernel for not reaping CQE:s.
+        //
+        // I think we should consider handling that as a fatalError, as fundamentally the ring size is too small
+        // compared to the amount of events the user tries to push through in a single eventloop tick.
+        //
+        // This is mostly a problem for synthetic tests that e.g. do a huge amount of registration modifications.
+        //
+        // This is a slight design issue with SwiftNIO in general that should be discussed.
```
weissi, line 169: Sounds good. Let's discuss this.

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+        //
+        while true {
+            if let sqe = CNIOLinux.io_uring_get_sqe(&ring) {
+                return sqe
+            }
+            self.io_uring_flush()
```
weissi, line 175: is there a guarantee that this works and eventually gets space? Because we're not consuming CQE here, right?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+
+    // Ok, this was a bummer - turns out that flushing multiple SQE:s
+    // can fail midflight and this will actually happen for real when e.g. a socket
+    // has gone down and we are re-registering polls this means we will silently lose any
+    // entries after the failed fd. Ouch. Proper approach is to use io_uring_sq_ready() in a loop.
+    // See: https://github.com/axboe/liburing/issues/309
+    internal func io_uring_flush() {         // When using SQPOLL this is basically a NOP
+        var waitingSubmissions : UInt32 = 0
+        var submissionCount = 0
+        var retval : Int32
```
weissi, line 188:
```suggestion
        var retval: CInt
```

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,509 @@
...
+            // time around. Given the async nature, this is fine, we will not
+            // lose any submissions. We could possibly still get stuck
+            // trying to get new SQE if the actual SQE queue is full, but
+            // that would be due to user error in usage IMHO and we should fatalError there.
+            case -EAGAIN:
+                _debugPrint("io_uring_flush io_uring_submit -EBUSY/-EAGAIN waitingSubmissions[\(waitingSubmissions)] submissionCount[\(submissionCount)]. Breaking out and resubmitting later (whenReady() end).")
+                break loop
+            // -ENOMEM when there is not enough memory to do internal allocations on the kernel side.
+            // Right nog we just loop with a sleep trying to buy time, but could also possibly fatalError here.
+            // See: https://github.com/axboe/liburing/issues/309
```
weissi, line 214: isn't https://github.com/axboe/liburing/issues/310 supposed to fix this? What's the state there?

Sources/NIO/SelectorUring.swift
```diff
@@ -0,0 +1,353 @@
...
+        _debugPrint("UringSelector up and running fd [\(self.selectorFD)] wakeups on event_fd [\(self.eventFD)]")
+    }
+
+    func deinitAssertions0() {
+        assert(self.eventFD == -1, "self.eventFD == \(self.eventFD) on deinitAssertions0 deinit, forgot close?")
+    }
+
+    func register0<S: Selectable>(selectable: S, fd: Int, interested: SelectorEventSet, sequenceIdentifier: RegistrationSequenceIdentifier) throws {
+        _debugPrint("register interested \(interested) uringEventSet [\(interested.uringEventSet)] sequenceIdentifier[\(sequenceIdentifier)]")
+        deferredReregistrationsPending = true
```
weissi, line 134: do we still need this? If yes, it's missing a `self.`

Sources/NIO/SelectorUring.swift
```diff
@@ -0,0 +1,353 @@
...
+                                       submitNow:!deferReregistrations)
+    }
+
+    private func _shouldRefreshPollForEvent(selectorEvent:SelectorEventSet) -> Bool {
+        if selectorEvent.contains(.read) {
+            // as we don't do exhaustive reads, we need to prod the kernel for
+            // new events, would be even better if we knew if we had read all there is
+            return true
+        }
+        // FIXME: Question for SwiftNIO team: Need to verify that SwiftNIO writes until blocking (exhaustive writing)
```
weissi, line 184: Yes, at the moment, NIO (with `BaseSocketChannel`) will always write until we get `EWOULDBLOCK` (or of course nothing more to write / error / ...).

Sources/NIO/SelectorUring.swift
```diff
@@ -0,0 +1,353 @@
...
+        _debugPrint("deregister interested \(selectable) reg.interested.uringEventSet [\(oldInterested.uringEventSet)]")
+
+        deferredReregistrationsPending = true
+        ring.io_uring_prep_poll_remove(fd: Int32(fd),
+                                       pollMask: oldInterested.uringEventSet,
+                                       sequenceIdentifier: sequenceIdentifier,
+                                       submitNow:!deferReregistrations)
+    }
+
+    private func _shouldRefreshPollForEvent(selectorEvent:SelectorEventSet) -> Bool {
```
weissi, line 178:
```suggestion
    private func shouldRefreshPollForEvent(selectorEvent: SelectorEventSet) -> Bool {
```

weissi (changes requested):
Awesome, this is much more compact now, thank you! I left a few comments and I think there are a few things to discuss around the number of submissions , edge/level triggered etc

Sources/CNIOLinux/liburing_shims.c
hassila: You mean the check for the syscall etc? I guess not strictly needed, just defensive - can remove it. I guess at this point, anyone trying it out will be a brave soul anyway and know what they are doing...

Sources/NIO/LinuxUring.swift
hassila: It was just to be able to have an assert for zero bit pattern (to sanity check), there was naturally a zero bit pattern when running with stdin on one of the integration tests. It could be put to zero and remove the assert now.

Sources/NIO/LinuxUring.swift
hassila: I will need to rework after #1801 yep.

Sources/NIO/LinuxUring.swift
hassila: Nope, because POLLCANCEL is invented by me to be able to signal up to the UringSelector code that a re-registration is needed - it does not exist in any system header. It was to be space efficient and not to tuck on extra data. The alternative is to extend the uring event type, would be cleaner I suppose.

Sources/NIO/LinuxUring.swift
hassila: Yeah, we can go that route, np.

Sources/CNIOLinux/liburing_shims.c
hassila: Ouch, you are right of course. Will rework to pthread_once.

Sources/CNIOLinux/liburing_shims.c
hassila: Yep. Actually, even `= {};` is correct btw (struct members that are not specified in the assignment will be initialised as static variables would be in C, which is zero...)

Sources/NIO/LinuxUring.swift
hassila: Sure, will fix. Old habits are hard to break, still do it by reflex sometimes, sorry.

Sources/CNIOLinux/liburing_shims.c
hassila: It is [planned to be lifted in 5.13](https://git.kernel.dk/cgit/linux-block/commit/?h=for-5.13/io_uring&id=548d819d1eed7b6bf86d36c8de2fbc54b69db571) which is a little bit out, 5.12 is not shipped yet...

weissi: @Lukasa WDYT, should we still ship SQPOLL support despite straight away the fact that even with the latest-greatest it needs `root`? Don't really want to encourage people to run NIO programs as root. But of course we can just take the code as is and runtime detect. But I'd assume nobody to actually be running `SQPOLL` (because `root`) so we've more code that's less frequently run so more permutations?

Sources/CNIOLinux/liburing_shims.c
weissi: @Lukasa WDYT? I'd personally think start with the least amount of code and add in once we have people screaming rather than being overly defensive.

weissi: (but I'm cool with syscall testing too)

Sources/CNIOLinux/liburing_shims.c
hassila: In current kernels, if non-privileged, it will just work as usual and disregard the sqpoll flag IIRC.

weissi: Yes, that's my understanding too. What I was trying to say is given that (hopefully) nobody runs this as root, nobody will exercise the `SQPOLL` code. So maybe we should have the initial version without `SQPOLL` support at all and at that later (once kernels support it).

Sources/CNIOLinux/liburing_shims.c
hassila: Ok, removing it.

Sources/CNIOLinux/liburing_shims.c
hassila: Yes, changing.

Sources/NIO/LinuxUring.swift
hassila: Ok.

Sources/NIO/LinuxUring.swift
hassila: (although easier to read the debug output, so even while the assert is now removed, keeping event type starting from 1 for easier reading of debug output)

Sources/NIO/LinuxUring.swift
hassila: I've paired it down to return a bool up to the selector instead, just used for debugging purposes at this stage really.

Sources/NIO/LinuxUring.swift
hassila: Removed.

Sources/CNIOLinux/liburing_shims.c
Lukasa: I'm inclined to drop it and add it back later once we think folks can actually exercise the code.

Sources/CNIOLinux/liburing_shims.c
hassila: Ok, removed this code based on other feedback.

Sources/CNIOLinux/liburing_shims.c
hassila: Code removed.

Sources/CNIOLinux/liburing_shims.c
hassila: Code removed.

Sources/CNIOLinux/liburing_shims.c
hassila: Ok,, code removed.

Sources/NIO/LinuxUring.swift
hassila: Removed.

Sources/NIO/LinuxUring.swift
hassila: There are fundamentally three io_uring system calls that can result from using the liburing functions - io_uring_setup, io_uring_register and io_uring_enter - only the last can return EINTR. That only occur according to documentation in “Can happen while waiting for events with IORING_ENTER_GETEVENTS” which is basically fine (that is when we are blocked waiting for events, so we would just return no events and run the next event loop tick).

hassila: (most of the io_uring helper functions just set up fields and prepare for calling one of those syscalls in a nicer way, except for queue setup/exit that also `mmap()`, but that can't EINTR AFAIU and would be a bug in the liburing library if it did.)

Sources/NIO/LinuxUring.swift
hassila: No, there is no guarantee - se comment above, I could see us doing a few retries and then fatalError() if failing. This is tied to the overall discussion above about the dynamics of generating outbound events / reaping events - if having a "bad producer" generating infinite events, we will eventually get back pressure from the kernel and not being able to send more. It's fundamentally an issue for the synthetic tests that do a lot of registration changes in a loop - should not be an issue for well behaved normal channel handlers I think. I think one possible solution would be to just keep track of all pending registration changes that needs to be sent to the kernel and merge it, and just do the actual re-registration to the kernel when we enter whenReady0 the next time. I've done that in the io_uring selector except for the merging (I postpone the flushing of registration changes until the next whenReady0 call, which helps alleviate this for real-world scenarios, but an aggressive enough client will eventually run out of SQE:s and block there.)

Sources/NIO/LinuxUring.swift
hassila: No, https://github.com/axboe/liburing/issues/310 is about being able to do updates to existing polls (especially multishot polls), the kernel could still presumably fail with internal memory allocations (however unlikely, it was just pointed out by one of the kernel developers as a possible failure state when [discussing the appropriate approach](https://github.com/axboe/liburing/issues/309).

Sources/NIO/SelectorUring.swift
hassila: Yes, related to comment above about deferring actual flushing of registration modifications until entering `whenReady0`. Could probably discuss if this would be useful behaviour overall for other selector backends (for example, for `kqueue`, this could decrease the number of syscalls significantly if multiple registration changes are done during a single event loop tick, as all of them would be coalesced/postpostponed until entering `whenReady0` the next time).

hassila:
Thanks for the reviews @weissi and @Lukasa, I've updated everything according to PR feedback, waiting for #1801 and #1807 to drop and will do the final adjustments after that unless you have anything more?

Couple of open discussions too above also that we might want to finalise (may result in new issues possibly) and/or if you want to close if happy with response.

weissi:
@hassila apologies this is taking so long. I wanted to get https://github.com/apple/swift-nio/pull/1801 first in which requires https://github.com/apple/swift-nio/pull/1807 but that showed a problem in the compiler which lead to https://bugs.swift.org/browse/SR-14516 and then https://github.com/apple/swift-nio/pull/1814 as a workaround (which hopefully works).

hassila:
@weissi , no worries, I am following those sagas :-) Will sort it out when you are ready.

Sources/NIO/LinuxUring.swift
hassila: Rechecked context, we do still need to merge events internally in the backend (having worked in #1801 now).

Sources/NIO/SelectorUring.swift
hassila: Added self.

hassila:
Ok, I have updated the implementation now to adopt for the changes made in #1801, #1807 and #1815 and adjusted for previous PR feedback, so I think its ready for another round now. 

The build queue for PR request validation seems to be stuck so hope there are no issues with older Swift versions... The build queue at https://ci.swiftserver.group/view/nio/ is 86 deep right now...

Sources/NIO/BaseSocketChannel.swift
```diff
@@ -1095,7 +1095,9 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
 
             return readStreamState
         }
+        #if !SWIFTNIO_USE_IO_URING
```
Lukasa, line 1098: So I've lost track of where we are with this: @hassila do you feel confident this is not masking a bug?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+}
+
+extension UInt64 {
+    init(_ uringUserData: URingUserData) {
+        let fd = uringUserData.fileDescriptor
+        let eventType = uringUserData.eventType.rawValue
+        assert(fd >= 0, "\(fd) is not a valid file descriptor")
+        assert(eventType >= 0, "\(eventType) is not a valid eventType")
+
+        self = IntegerBitPacking.packUInt32UInt16UInt8(UInt32(fd),
```
Lukasa, line 72: Are we expecting to tolerate negative FDs here?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+    init () {
+        self.fd = -1
+        self.pollMask = 0
+        self.registrationID = 0
+        self.pollCancelled = false
+    }
+}
+
+// This is the key we use for merging events in our internal hashtable
+struct fdEventKey: Hashable {
```
Lukasa, line 93: Nit: this doesn't match Swift naming conventions. Should be `FDEventKey`.

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+
+    private var ring = io_uring()
+    private let ringEntries: CUnsignedInt = 8192
+    private let cqeMaxCount : UInt32 = 8192 // this is the max chunk of CQE we take.
+
+    var cqes : UnsafeMutablePointer<UnsafeMutablePointer<io_uring_cqe>?>
+    var fdEvents = [fdEventKey : UInt32]() // fd, sequence_identifier : merged event_poll_return
+    var emptyCqe = io_uring_cqe()
+
+    internal func fd() -> Int32 {
```
Lukasa, line 120: Any reason this isn't a computed property?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+
+        _debugPrint("io_uring_queue_init \(self.ring.ring_fd)")
+     }
+
+    internal func io_uring_queue_exit() {
+        _debugPrint("io_uring_queue_exit \(self.ring.ring_fd)")
+        CNIOLinux.io_uring_queue_exit(&ring)
+    }
+
+    static internal func io_uring_use_multishot_poll() -> Bool {
```
Lukasa, line 169: Any reason this isn't a computed property?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+            // We can get -EAGAIN if the CQE queue is full and we get back pressure from
+            // the kernel to start processing CQE:s. If we break here with unsubmitted
+            // SQE:s, they will stay pending on the user-level side and be flushed
+            // to the kernel after we had the opportunity to reap more CQE:s
+            // In practice it will be at the end of whenReady the next
+            // time around. Given the async nature, this is fine, we will not
+            // lose any submissions. We could possibly still get stuck
+            // trying to get new SQE if the actual SQE queue is full, but
+            // that would be due to user error in usage IMHO and we should fatalError there.
+            case -EAGAIN:
```
Lukasa, line 229: Rather than use `fallthrough` can we use `case -EBUSY, -EAGAIN:`?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+                _debugPrint("io_uring_flush io_uring_submit -EBUSY/-EAGAIN waitingSubmissions[\(waitingSubmissions)] submissionCount[\(submissionCount)]. Breaking out and resubmitting later (whenReady() end).")
+                break loop
+            // -ENOMEM when there is not enough memory to do internal allocations on the kernel side.
+            // Right nog we just loop with a sleep trying to buy time, but could also possibly fatalError here.
+            // See: https://github.com/axboe/liburing/issues/309
+            case -ENOMEM:
+                usleep(1_000_000) // let's not busy loop to give the kernel some time to recover if possible
```
Lukasa, line 236: This sleep is _very_ long. Surely we don't need to sleep for a whole second?

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,533 @@
...
+                if let current = fdEvents[fdEventKey(uringUserData.fileDescriptor, uringUserData.registrationID)] {
+                    fdEvents[fdEventKey(uringUserData.fileDescriptor, uringUserData.registrationID)] = current | pollError
+                } else {
+                    fdEvents[fdEventKey(uringUserData.fileDescriptor, uringUserData.registrationID)] = pollError
+                }
+                break
+            case -EINVAL:
+                _debugPrint("Failed poll with -EINVAL for cqeIndex[\(cqeIndex)]")
+                break
+            case -EBADF:
```
Lukasa, line 365: Is there any case when these errors should not be fatal to our process? That is, is there any circumstance where they don't reflect programmer error?

Sources/NIO/SelectorUring.swift
```diff
@@ -0,0 +1,359 @@
...
+///  - `readHangup` corresponds to `POLLRDHUP`
+///  - `input` corresponds to `POLLIN`
+///  - `output` corresponds to `POLLOUT`
+///  - `error` corresponds to `POLLERR`
+private struct UringFilterSet: OptionSet, Equatable {
+    typealias RawValue = UInt8
+
+    let rawValue: RawValue
+
+    static let _none = UringFilterSet([])
```
Lukasa, line 30: Can I propose using `rawValue` directly here? Swift sometimes generates weird code around the array-literal constructions.

Sources/NIO/SelectorUring.swift
```diff
@@ -0,0 +1,359 @@
...
+        print("S [\(NIOThread.current)] " + s())
+        #endif
+    }
+
+    func initialiseState0() throws {
+        try ring.io_uring_queue_init()
+        self.selectorFD = ring.fd()
+
+        // eventfd are always singleshot and re-register each time around
+        // as certain use cases of nio seems to generate tons of wakeups
```
Lukasa, line 109: Yup, this "tons of wakeups" is not what we want to do at the moment, but it's where we're at. We should investigate getting rid of them. In the fullness of time we'll be at risk of some redundant wakeups because we'll want to move to an MPSC queue, but right now when we have a big dumb lock we can absolutely avoid spurious wakes.

Lukasa (commented):
Cool, looking really good so far! I've left a few notes for now.

Sources/NIO/LinuxUring.swift
hassila: Not really, it would be due to a higher-level up error (thus the assert, copied from the Epoll implementation). Would you prefer a fatalError there?

Sources/NIO/LinuxUring.swift
hassila: Fixed.

Sources/NIO/LinuxUring.swift
hassila: Nope, just reflects that I've not Swiftified myself enough yet (WIP) ;-) Thanks, changing.

Sources/NIO/LinuxUring.swift
hassila: Same, changing.

Sources/NIO/LinuxUring.swift
hassila: Updated.

Sources/NIO/LinuxUring.swift
hassila: Changed it to 10ms.

Sources/NIO/LinuxUring.swift
hassila: Agree that EINVAL should be fatal, EBADF could happen possibly for a close race (would _probably_ be sloppy coding from user, but we would handle it robustly and seems a bit harsh to terminate then). I've removed EINVAL and changed unexpected errors to fatalError instead as it would be a severe implementation error. EAGAIN/EALREADY can happen during certain modification races and would be ok I think, so those are still handled separately.

Sources/NIO/SelectorUring.swift
hassila: Sure. Changing identical usage in epoll implementation too then.

Sources/NIO/SelectorUring.swift
hassila: Yeah, wasn't perhaps the most accurate term, updated the comment. MPSC seems like an interesting direction.

Sources/NIO/BaseSocketChannel.swift
hassila: Yes, I think that the root problem here is that the io_uring backend does not have the implicit synchronisation between modifications to the poll mask and the actual returned events on the completion queue that kqueue and epoll has. For kqueue and epoll, there is an implicit synchronisation point such that after a modification of the poll mask has been issued, the next call to reap events will be sure to not include events which does not match the new poll mask. Specifically for this assert, it means that we will be guaranteed to never receive a POLLIN notification unless there are bytes available to read.

For a fully asynchronous backend like io_uring, there are no such implicit synchronisation point, so after we have submitted the asynchronous event to change the poll mask, we may still reap pending asynchronous replies for the old poll mask, and thus receive a POLLIN even though we have modified the mask visavi the kernel.

The only way to avoid that race, would be to use heavy handed synchronisation primitives like IOSQE_IO_DRAIN (basically flushing **all** pending requests and wait for a fake event result to sync up) - see some discussions at (LWN for fsync)[https://lwn.net/Articles/785690/]. It would be awful from a performance POV to get that guarantee, so would definitely not prefer to assert() this (as the assert simply is not valid for the full asynchronous backend case without implicit synchronisation points).

Sources/NIO/SelectorUring.swift
hassila: There was a reason for the epoll take on it:

```
/home/ubuntu/swiftnio/swift-nio/Sources/NIO/SelectorUring.swift:30:16: warning: static property '_none' produces an empty option set
    static let _none = UringFilterSet(rawValue: 0)
               ^
/home/ubuntu/swiftnio/swift-nio/Sources/NIO/SelectorUring.swift:30:16: note: use [] to silence this warning
    static let _none = UringFilterSet(rawValue: 0)
               ^                     ~~~~~~~~~~~~~
                                     ([])
```

hassila: (so reverted to the [] pattern to squelch warnings)

hassila:
> Cool, looking really good so far! I've left a few notes for now.

Thanks, I've addressed them in 983d0a0 - couple of things unresolved if you want to read the replies (not sure what is considered best practice / being courteous in when closing review comments - should they be closed as addressed by updates, or by the original reviewer? Im closing those that are "obviously fixed" and leaving the other open for your review).

Sources/NIO/BaseSocketChannel.swift
Lukasa: So long as we don't think this is covering up a bug I'm fine with that. Can we add a code comment with this rationale though?

Sources/NIO/LinuxUring.swift
Lukasa: Oh, hey, I missed the assert. Given the assert is here, let's use `UInt32(truncatingIfNeeded:)` which should elide the check for unrepresentable values.

Sources/NIO/LinuxUring.swift
Lukasa: Right now NIO does not tolerate EBADF from its syscalls under any circumstances: the kind of sloppy coding it implies is highly dangerous and will almost certainly crash later on in the process.

Lukasa:
I don't really mind either way: I expand the closed comments on busy reviews like this because of the odds I'll miss something anyway.

Sources/NIO/BaseSocketChannel.swift
hassila: Sure.

Sources/NIO/LinuxUring.swift
hassila: The reason I wrote _probably_, was that I had a vague feeling there was a valid reason why this happened - and there is, I just removed the handling to refresh my memory and reran the test suite...

Fundamentally its related to the discussion with the assert above - as we can have a stream of pending requests that are pending towards the kernel - and in this case, the main test thread will perform a close (and we have basically prepared to send the deregistration to the kernel, so a close() system call will be performed, which races with our old pending requests (once again, completely async backend bites us, no implicit synchronisation point for the deregistration). So would need to be robust for EBADF for the close race - it would similarly to the assert either require extremely heavy handed synchronisation when changing registrations visavi the kernel, or simply that we are robust when it happens. So in this particular case, its not really sloppy coding that is the root issue, but rather the true asynchronous nature of the io_uring requests.

```
Test Case 'BootstrapTest.testPreConnectedSocketSetsChannelOptionsBeforeChannelInitializer' started at 2021-04-27 13:17:01.157
...
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_prep_poll_add fileDescriptor[13] pollMask[24] bitpatternAsPointer[Optional(0x000100000000000d)] submitNow[false] multishot[true]
S [NIOThread(actualName = NIO-ELT-42-#0)] Re-register old SelectorEventSet(rawValue: 1) new SelectorEventSet(rawValue: 3) uringEventSet [24] reg.uringEventSet [8216]
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_poll_update fileDescriptor[13] oldPollmask[24] newPollmask[8216]  userBitpatternAsPointer[Optional(0x000200000000000d)]
S [NIOThread(actualName = NIO-ELT-42-#0)] register interested SelectorEventSet(rawValue: 1) uringEventSet [24] registrationID[SelectorRegistrationID(_rawValue: 1)]
...
S [NIOThread(actualName = NIO-ELT-42-#0)] deregister interested BaseSocket { fd=13 } reg.interested.uringEventSet [8216]
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_prep_poll_remove fileDescriptor[13] pollMask[8216] bitpatternAsPointer[Optional(0x000100000000000d)] 
...
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_flush io_uring_submit needed [1] submission(s), submitted [1] SQE:s out of [7] possible
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_flush io_uring_submit needed [2] submission(s), submitted [6] SQE:s out of [6] possible
L [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_peek_batch_cqe found [12] events
Q [NIOThread(actualName = NIO-ELT-42-#0)] io_uring_peek_batch_cqe CQE:s [0x00007fe988000da0] - ring flags are [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 0 = fd[11] eventType[poll] registrationID[0] res [1] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 1 = fd[12] eventType[pollModify] registrationID[65535] res [0] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 2 = fd[13] eventType[poll] registrationID[0] res [-9] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 3 = fd[13] eventType[pollModify] registrationID[0] res [-2] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 4 = fd[14] eventType[poll] registrationID[1] res [8192] flags [2]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 5 = fd[14] eventType[pollModify] registrationID[1] res [0] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 6 = fd[14] eventType[poll] registrationID[1] res [8192] flags [2]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 7 = fd[14] eventType[pollModify] registrationID[1] res [0] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 8 = fd[14] eventType[poll] registrationID[1] res [8193] flags [2]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 9 = fd[13] eventType[pollDelete] registrationID[0] res [-2] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 10 = fd[12] eventType[poll] registrationID[65535] res [-125] flags [0]
Q [NIOThread(actualName = NIO-ELT-42-#0)] 11 = fd[12] eventType[pollDelete] registrationID[65535] res [0] flags [0]
Fatal error: Failed poll with unexpected error (-9 for cqeIndex[2]: file NIO/LinuxUring.swift, line 360
Current stack trace:
```

Here we can see how all three requests are sent together in this particular circumstance, but are clearly processed out of order and

Sources/NIO/LinuxUring.swift
hassila: Fixed.

Sources/NIO/LinuxUring.swift
weissi: @hassila if there are no ordering guarantees, how are we then guaranteeing that we don't get stuck for example? What if a `pollModify` unregisters say from `.read` meant for `fd=123 (registrationID=1)` but gets processed for `fd=123 (registrationID=2)`?

hassila: @weissi the registration id is part of the user data that is used to match the request on the kernel side, so we can't modify polls for other registrations.

weissi: @hassila okay, so we're saying the _only_ thing affected here are requests for fds that are already closed? And I suppose the kernel will _also_ send us `EBADF` for trying to modify `fd=123/regID=1` when in fact we're registered with `fd=123/regID=2`?

hassila: @weissi also, in this case, the major issue is that the `close()` can race with our pending requests. Later on when moving to a true io_uring implementation, the point will be moot as well, as we would then schedule `read()` directly and not do any modifications (just cancellations of reads, where we can choose to throw away any racy result returned if semantically correct to do so if a user closes a channel while there is inbound data arriving).

hassila: > @hassila okay, so we're saying the _only_ thing affected here are requests for fds that are already closed? And I suppose the kernel will _also_ send us `EBADF` for trying to modify `fd=123/regID=1` when in fact we're registered with `fd=123/regID=2`?

@weissi It would give us an ENOENT for that case, but an error - yes.

weissi: Thanks for the clarifications. This all makes sense. I think in this particular case accepting `EBADF` is then unfortunately reasonable. I think we should add a good comment for that. The key thing obviously is that there is no synchronisation barrier with the kernel so we have no idea when the kernel will have apply our submission requests.
@Lukasa wdyt?

hassila: I'm adding some comments now.

Lukasa: I'm in principle fine with this, but I'd like to be confident that iouring absolutely does take user data into account when identifying a poll request.

I can't find a specific citation that implies that the IORING_OP_POLL_ADD or IORING_OP_POLL_REMOVE events do anything in particular with the user data other than return it. That doesn't mean they don't, only that I haven't yet found any evidence that they do.

The thing that worries me is the question of when the kernel resolves the file descriptor into the file description. I am entirely confident that, once the kernel has done that resolution, closing the userland fd will not cause us any difficulty. What gives me pause is the possibility that we close one fd and then obtain it from a new open, registering it on a new uring, all before the first uring processes the SQE referring to the old (now closed-and-replaced) file descriptor. What happens here?

My gut instinct is that io_uring does not use the user_data to distinguish these requests and so can return a spurious poll completion. Does that sound right?

hassila: AFAIU, looking [at the implementation](https://git.kernel.dk/cgit/linux-block/diff/fs/io_uring.c?h=for-5.13%2Fio_uring&id=9ba5fac8cf3b607652397f863dc229bbc8c3cbc1&context=40&ignorews=0&dt=0)
`io_poll_find()` is used to find the request for both poll update/cancel SQE:s and there we have the check `if (sqe_addr != req->user_data)` matching on user_data. Poll add doesn't need to do more than return it. 

> What gives me pause is the possibility that we close one fd and then obtain it from a new open, registering it on a new uring, all before the first uring processes the SQE referring to the old (now closed-and-replaced) file descriptor. What happens here?

I'll see if I can ask @isilence or @axboe to comment on that specific aspect.

isilence: > AFAIU, looking [at the implementation](https://git.kernel.dk/cgit/linux-block/diff/fs/io_uring.c?h=for-5.13%2Fio_uring&id=9ba5fac8cf3b607652397f863dc229bbc8c3cbc1&context=40&ignorews=0&dt=0)
> `io_poll_find()` is used to find the request for both poll update/cancel SQE:s and there we have the check `if (sqe_addr != req->user_data)` matching on user_data. Poll add doesn't need to do more than return it.

Right, it only does matching against it. Also, modify/remove don't take an fd, so matching uses only that passed data.

> > What gives me pause is the possibility that we close one fd and then obtain it from a new open, registering it on a new uring, all before the first uring processes the SQE referring to the old (now closed-and-replaced) file descriptor. What happens here?

fds are resolved during submission, after that the kernel operates with referenced `struct file *`, so no update of userspace file tables can affect it. So, not non-sqpoll rings it's stable right after `io_uring_submit()` and so the corresponding enter syscall returns.

sqpoll rings may lag in that sense because the submission is done by async process that should wake up and reap SQEs. But SQ can be waited upon if needed. IIRC there was a helper in liburing.

isilence: > > AFAIU, looking [at the implementation](https://git.kernel.dk/cgit/linux-block/diff/fs/io_uring.c?h=for-5.13%2Fio_uring&id=9ba5fac8cf3b607652397f863dc229bbc8c3cbc1&context=40&ignorews=0&dt=0)
> > `io_poll_find()` is used to find the request for both poll update/cancel SQE:s and there we have the check `if (sqe_addr != req->user_data)` matching on user_data. Poll add doesn't need to do more than return it.
> 
> Right, it only does matching against it. Also, modify/remove don't take an fd, so matching uses only that passed in user_data/sqe_addr.
> 
> > > What gives me pause is the possibility that we close one fd and then obtain it from a new open, registering it on a new uring, all before the first uring processes the SQE referring to the old (now closed-and-replaced) file descriptor. What happens here?
> 
> fds are resolved during submission, after that the kernel operates with referenced `struct file *`, so no update of userspace file tables can affect it. So, not non-sqpoll rings it's stable right after `io_uring_submit()` and so the corresponding enter syscall returns.
> 
> sqpoll rings may lag in that sense because the submission is done by async process that should wake up and reap SQEs. But SQ can be waited upon if needed. IIRC there was a helper in liburing.

hassila: Thank you very much @isilence for clarification. We currently don't support SQPOLL rings for Swift NIO so given that the kernel operates with the [`struct file *`](https://www.oreilly.com/library/view/linux-device-drivers/0596000081/ch03s04.html) after the submit, we should not get updates just because someone else happened to get the same fd in userspace after close/open. Must admit it is always interesting to straddle and peek into kernel land ;-)

Lukasa: Great, ok. In that case I think I'm happy with the code as-is, so long as we comment about this (and ideally reference this thread).

Lukasa: Thanks for chiming in @isilence, much appreciated.

isilence: fwiw, `struct file` may be better known for some as "open file description" or OFD

hassila: Ok, I've updated the comment and put in a reference to this discussion too.

Sources/NIO/LinuxUring.swift
hassila: I think (as a separate issue) it might be worth considering introducing some back pressure mechanism for registration modifications (that really only would kick in for degenerate cases), or alternatively (what I probably would prefer) is that we'd start tracking the registration modifications and merge/coalesce all of them that are done during a single event loop tick. Then we'd have an upper bound of `numfds` (channels) of outbound registration change events which would not fill up e.g. the outbound SQE queue and make less system calls for such cases.  Then again, it is mostly an issue for tests, but it would be nice to handle degenerate behaviours too robustly. I don't know what you think? Should probably move on to a separate issue anyway.

hassila: @weissi @Lukasa - opened https://github.com/apple/swift-nio/issues/1832 for that, hope it is clear enough.

hassila:
Updated for #1820 as well now. Not sure if there is anything truly outstanding now?

weissi:
@Lukasa / @hassila I think we can get this in now. Thank you so much @hassila .

If you wouldn't mind, could you fix these lines?

    git grep ' : ' Sources/*.swift | grep -v ' ? '

The usual NIO codestyle is to use `variable: Type` (ie no space before the `:` if it's a type annotation / extension decl)

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,534 @@
...
+        assert(eventType >= 0, "\(eventType) is not a valid eventType")
+
+        self = IntegerBitPacking.packUInt32UInt16UInt8(UInt32(truncatingIfNeeded: fd),
+                                                       uringUserData.registrationID,
+                                                       eventType)
+    }
+}
+
+// These are the events returned up to the selector
+internal struct UringEvent {
```
weissi, line 79: nit: elsewhere you use `URing` (not `Uring` like here). I think we should be consistent and use `URing`

Sources/NIO/LinuxUring.swift
```diff
@@ -0,0 +1,534 @@
...
+    var registrationID : UInt16 // we just have the truncated lower 16 bits of the registrationID
+
+    init(_ f: CInt, _ s : UInt16) {
+        self.fileDescriptor = f
+        self.registrationID = s
+    }
+}
+
+final internal class Uring {
+    internal static let POLLIN: CUnsignedInt = numericCast(CNIOLinux.POLLIN)
```
weissi, line 104: nits: usually I'd suggest to name these in a Swift way (ie. `pollIn`). Also `URing` vs `Uring`.

hassila: The names were modelled after how its done in epoll-uring - I think it's nice if the style is consistent? (also for someone used to working with poll interfaces, they are immediately recognisable for what they are) - I have updated to URing though.

Sources/NIO/LinuxUring.swift
hassila: Fixed.

hassila:
> @Lukasa / @hassila I think we can get this in now. Thank you so much @hassila .
> 
> If you wouldn't mind, could you fix these lines?
> 
> ```
> git grep ' : ' Sources/*.swift | grep -v ' ? '
> ```
> 
> The usual NIO codestyle is to use `variable: Type` (ie no space before the `:` if it's a type annotation / extension decl)

Yeah, sorry about that - fixed all. Would really love to have a standard formatting when saving the file in Xcode that adheres to formatting standards, it is great work for a computer...

weissi:
> Yeah, sorry about that - fixed all. Would really love to have a standard formatting when saving the file in Xcode that adheres to formatting standards, it is great work for a computer...

Agreed. AFAIK there aren't any tools in Swift that reformat only lines that have been edited anyway and we're kinda reluctant to reformat the whole project. Because likely there'll be another reformat once the "official swift styleguide & formatting tool" it out :|

weissi (approved):
Thank you so much @hassila . I think we should merge this!

Lukasa (approved):
Great, let's merge this. This will give it some time to bake and free up @hassila and other interested parties to play around with it more easily in-tree. It'll also set us up to do round 2 of this work, which will be to try a whole new event loop implementation.

hassila:
@Lukasa  agree, the new event loop will be the most interesting aspect, then we should see some real performance and scalability gains. Thanks to both you and @weissi for some patience sorting this out...

Lukasa:
No, thank you! Reviewing these patches is always easier than actually writing them. You did the hard work, we just helped guide it into landing. Please feel free to continue to reach out if you're doing future work, especially if you're planning to tackle a new event loop, as we can try to provide some guidance on how you might design it.

weissi:
> @Lukasa agree, the new event loop will be the most interesting aspect, then we should see some real performance and scalability gains. Thanks to both you and @weissi for some patience sorting this out...

Much like Cory says, you did all the hard work and you kept the patience with us. This is fantastic!!!

PR #1806, opened 2021-04-15
support qemu: qemu returns ENOPROTOOPT for unknown socket opts

weissi:
Motivation:

When running SwiftNIO under Qemu, then one test case fails because it
expects `ENOTSOCK` but Qemu returns `ENOPROTOOPT`. This is a Qemu bug.

Modification:

Accept `ENOTSOCK` or `ENOPROTOOPT`.

Result:

Test suite works under Qemu.

### DIFF
--- a/Tests/NIOTests/SystemTest.swift
+++ b/Tests/NIOTests/SystemTest.swift
@@ -36,10 +36,9 @@ class SystemTest: XCTestCase {
                 }
                 XCTFail("success even though the call was invalid")
             } catch let e as IOError {
-                XCTAssertEqual(ENOTSOCK, e.errnoCode)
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

weissi:
Qemu is damn good!

```
$ time docker run -it --rm --platform linux/amd64 -v "$PWD:$PWD" -w "$PWD" -e QEMU_CPU=max swift:5.3 swift test
[10/10] Linking swift-nioPackageTests.xctest
Test Suite 'All tests' started at 2021-04-15 10:04:45.700
[...]
Test Suite 'All tests' passed at 2021-04-15 10:09:16.200
	 Executed 1314 tests, with 0 failures (0 unexpected) in 270.299 (270.299) seconds

real	5m3.851s
user	0m0.235s
sys	0m0.151s
```

PR #1810, opened 2021-04-19
fix compilation on Ubuntu 14.04 and other old Linuxes

weissi:
Motivation:

Glibc (in Sep 2014) changed the first parameter of `eventfd` from `int`
to `unsigned int` (which we assume since recently).

Modifications:

Make the code work with `int` and `unsigned int`.

Result:

Compiles again on 14.04.


Lukasa (approved):

weissi:
note, the change came in #1797

PR #1813, opened 2021-04-21
Fix crash: Create ByteBuffer Slice from ByteBuffer Slice after 16MB

fabianfett:
Fixes an issue that could lead to an assert in Debug mode and to undefined behavior in Release mode.

Motivation:

- When reading a new `ByteBuffer` slice from an originating `ByteBuffer` slice where the new `ByteBuffer` slice starts more than 16MiB into the underlying `ByteBuffer` storage, the new `ByteBuffer` slice will get a new 0 based storage.
- When creating the storage for the new slice we have an issue in where storage based indices were used which should have been slice based indices.

Modifications:

- Using the correct indices, when creating a `ByteBuffer` slice which is more than 16MiB into its originating storage.

Result:

- We can create `ByteBuffer` slices from slices after the 16MiB threshold.
- Fixes an [issue that was raised in PostgresNIO](https://github.com/vapor/postgres-nio/issues/148)

### DIFF
--- a/Sources/NIO/ByteBuffer-core.swift
+++ b/Sources/NIO/ByteBuffer-core.swift
@@ -640,8 +640,8 @@ public struct ByteBuffer {
             // the slice's begin is past the maximum supported slice begin value (16 MiB) so the only option we have
             // is copy the slice into a fresh buffer. The slice begin will then be at index 0.
             var new = self
-            new._moveWriterIndex(to: sliceStartIndex + length)
-            new._moveReaderIndex(to: sliceStartIndex)
+            new._moveWriterIndex(to: index + length)
+            new._moveReaderIndex(to: index)
             new._copyStorageAndRebase(capacity: length, resetIndices: true)
             return new
         }
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -1725,7 +1725,8 @@ class ByteBufferTest: XCTestCase {
         
         let finalSliceLength = 1
         // let's create a new buffer that uses all but one byte
-        XCTAssertEqual(remainingSlice.readBytes(length: remainingSliceLength - finalSliceLength), [UInt8](repeating: 2, count: remainingSliceLength - finalSliceLength))
+        XCTAssertEqual(remainingSlice.readBytes(length: remainingSliceLength - finalSliceLength),
+                       Array<UInt8>(repeating: 2, count: remainingSliceLength - finalSliceLength))
         
         // there should only be one byte left.
         XCTAssertEqual(remainingSlice.readableBytes, finalSliceLength)
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -1711,7 +1711,7 @@ class ByteBufferTest: XCTestCase {
         
         // create a buffer with 16MiB + 1 byte
         let inputBufferLength = 16 * 1024 * 1024 + 1
-        var inputBuffer = ByteBuffer.Allocator().buffer(capacity: inputBufferLength)
+        var inputBuffer = ByteBufferAllocator().buffer(capacity: inputBufferLength)
         inputBuffer.writeRepeatingByte(1, count: 8)
         inputBuffer.writeRepeatingByte(2, count: inputBufferLength - 9)
         inputBuffer.writeRepeatingByte(3, count: 1)
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -1704,6 +1704,41 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(0xcc, slice.getInteger(at: 0, as: UInt8.self))
         XCTAssertEqual(0xdd, slice.getInteger(at: slice.writerIndex - 1, as: UInt8.self))
     }
+    
+    func testSliceOnSliceAfterHitting16MBMark() {
+        // This test ensures that a slice will get a new backing storage if its start is more than
+        // 16MiB after the originating backing storage.
+        
+        // create a buffer with 16MiB + 1 byte
+        let inputBufferLength = 16 * 1024 * 1024 + 1
+        var inputBuffer = ByteBuffer.Allocator().buffer(capacity: inputBufferLength)
+        inputBuffer.writeRepeatingByte(1, count: 8)
+        inputBuffer.writeRepeatingByte(2, count: inputBufferLength - 9)
+        inputBuffer.writeRepeatingByte(3, count: 1)
+        // read a small slice from the inputBuffer, to create an offset of eight bytes
+        XCTAssertEqual(inputBuffer.readInteger(as: UInt64.self), 0x0101010101010101)
+        
+        // read the remaining bytes into a new slice (this will have a length of 16MiB - 7Bbytes)
+        let remainingSliceLength = inputBufferLength - 8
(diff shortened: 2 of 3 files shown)


Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,37 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(0xcc, slice.getInteger(at: 0, as: UInt8.self))
         XCTAssertEqual(0xdd, slice.getInteger(at: slice.writerIndex - 1, as: UInt8.self))
     }
+    
+    func testSliceOnSliceAfterHitting16MBMark() {
+        // This test ensures that a slice will get a new backing storage if its start is more than
+        // 16MiB after the originating backing storage.
+        
+        // create a buffer with (16MiB - 1byte) + 2 bytes = 16MiB + 1 byte
+        let inputBufferLength = Int(_UInt24.max) + 2
```
weissi, line 1713: Just hardcode the value here. The `_` means that `_UInt24` is meant to be private. So `let inputBufferLength =  1 << 24 + 1` or similar.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,37 @@ class ByteBufferTest: XCTestCase {
...
+        var inputBuffer = ByteBuffer(repeating: 1, count: inputBufferLength)
```
weissi, line 1714: Let's prepare the buffer with a different byte pattern so that we can prove at the end that we read the correct one. For example:
- 8 `1`s (which will be read in the `readInteger`)
- 16 MB - 6 (?) `2`s (which will be read in the `readSlice`)
- 1 `3` (which will be the leftovers aka the "final slice")

Lukasa (changes requested):
While this patch is great, it does a bit too much reaching into the ByteBuffer internals for my comfort. Can we rewrite it to avoid using UInt24 and avoid touching the _slice directly?

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,37 @@ class ByteBufferTest: XCTestCase {
...
+        // read a small slice from the inputBuffer, to create an offset of eight bytes
+        XCTAssertNotNil(inputBuffer.readInteger(as: UInt64.self))
+        
+        // read the remaining bytes into a new slice (this will have a length of 16MiB - 7Bbytes)
+        let remainingSliceLength = inputBufferLength - MemoryLayout<UInt64>.size
```
weissi, line 1719: shouldn't this be equal to `readableBytes - 1`? We could assert that

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,37 @@ class ByteBufferTest: XCTestCase {
...
+        let finalSliceLength = 1
+        // let's create a new buffer that uses all but one byte
+        XCTAssertNotNil(remainingSlice.readSlice(length: remainingSliceLength - finalSliceLength))
+        
+        // there should only be one byte left.
+        XCTAssertEqual(remainingSlice.readableBytes, finalSliceLength)
+        
+        // the upper bound is exactly one byte above the 16MiB threshold. For this reason a slice
+        // of the last byte, will need to get a new backing storage.
+        XCTAssertEqual(remainingSlice._slice.upperBound, UInt32(_UInt24.max) + 2)
```
weissi, line 1731: please don't access `_things` here, they're meant to be private, just not private because inlining

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,37 @@ class ByteBufferTest: XCTestCase {
...
+        let finalSlice = remainingSlice.readSlice(length: finalSliceLength)
+        XCTAssertEqual(finalSlice?.storageCapacity, 1)
+        XCTAssertEqual(finalSlice?._slice.lowerBound, 0)
+
```
weissi, line 1735: we should additionally assert
- that the slices/values read have the correct content (`1`s, `2`s, and `3`s or so as suggested above)
- that the underlying storage has indeed changed. There's a (test only) property called `buffer.storageIdentifier` or so which stays the same if nothing got reallocated/copied and changes if it has

weissi (changes requested):
Thank you so much @fabianfett , this is really fantastic work!!

Sources/NIO/ByteBuffer-core.swift
```diff
@@ -640,8 +640,8 @@ public struct ByteBuffer {
             // the slice's begin is past the maximum supported slice begin value (16 MiB) so the only option we have
```
weissi, line 640: There's almost no correct way to use the `sliceStartIndex` variable. I wonder if it would be nicer to eliminate it and put it straight into the condition. Ie. 

    guard self._slice.lowerBound + index <= ByteBuffer.Slice.maxSupportedLowerBound else

fabianfett: It is also used after the guard condition in L651.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,41 @@ class ByteBufferTest: XCTestCase {
...
         XCTAssertEqual(0xdd, slice.getInteger(at: slice.writerIndex - 1, as: UInt8.self))
     }
+    
+    func testSliceOnSliceAfterHitting16MBMark() {
+        // This test ensures that a slice will get a new backing storage if its start is more than
+        // 16MiB after the originating backing storage.
+        
+        // create a buffer with 16MiB + 1 byte
+        let inputBufferLength = 16 * 1024 * 1024 + 1
+        var inputBuffer = ByteBuffer.Allocator().buffer(capacity: inputBufferLength)
```
Lukasa, line 1714:
```suggestion
        var inputBuffer = ByteBufferAllocator().buffer(capacity: inputBufferLength)
```

Lukasa (commented):
One minor cleanup and then I think this is good.

Lukasa (approved):
LGTM, nice patch!

Sources/NIO/ByteBuffer-core.swift
weissi: fair

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -1704,6 +1704,41 @@ class ByteBufferTest: XCTestCase {
...
+        XCTAssertEqual(inputBuffer.readInteger(as: UInt64.self), 0x0101010101010101)
+        
+        // read the remaining bytes into a new slice (this will have a length of 16MiB - 7Bbytes)
+        let remainingSliceLength = inputBufferLength - 8
+        XCTAssertEqual(inputBuffer.readableBytes, remainingSliceLength)
+        var remainingSlice = inputBuffer.readSlice(length: remainingSliceLength)!
+        
+        let finalSliceLength = 1
+        // let's create a new buffer that uses all but one byte
+        XCTAssertEqual(remainingSlice.readBytes(length: remainingSliceLength - finalSliceLength), [UInt8](repeating: 2, count: remainingSliceLength - finalSliceLength))
```
weissi, line 1728:
```suggestion
        XCTAssertEqual(remainingSlice.readBytes(length: remainingSliceLength - finalSliceLength),
                       Array<UInt8>(repeating: 2, count: remainingSliceLength - finalSliceLength))
```

weissi (approved):
Awesome, I left one suggestion which you may want to apply. We prefer lines <= 120 (IIRC) and `Array<UInt8>(...)` over `[UInt8](...)` for the constructors.

PR #1815, opened 2021-04-22
Selectors: make file descriptors CInts everywhere

weissi:
Motivation:

We and the OSes only support `CInt` as file descriptors. For some reason
we had interfaces that took `Int`s as fds which requires frequent
casting around. This is unnecessary.

Modifications:

- make fds `CInt` (internally)
- clean up some `Int32` as `CInt` use

Result:

cleaner code, less casting


Lukasa:
Merged the parent

weissi:
thanks @Lukasa , updated this one

Lukasa (approved):

PR #1816, opened 2021-04-23
Less allocs during `flatMapErrorThrowing` and `flatMapThrowing`

fabianfett:
Less allocs = Faster code

### Motivation:

- We alloc quite a lot with our implementations of `flatMapThrowing` and `flatMapErrorThrowing`.
- While we don't use Futures a lot in NIO itself a lot of our users, depend quite a bit on them. Let's make their code faster.

### Modifications:

- Create a `Promise` and use `_whenComplete` directly instead of going through another `flatMap` method

### Result:

- In my testing I see a reduction of 3 allocs per invocation 🎉

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -506,13 +506,21 @@ extension EventLoopFuture {
     public func flatMapThrowing<NewValue>(file: StaticString = #file,
                                 line: UInt = #line,
                                 _ callback: @escaping (Value) throws -> NewValue) -> EventLoopFuture<NewValue> {
-        return self.flatMap(file: file, line: line) { (value: Value) -> EventLoopFuture<NewValue> in
-            do {
-                return EventLoopFuture<NewValue>(eventLoop: self.eventLoop, value: try callback(value), file: file, line: line)
-            } catch {
-                return EventLoopFuture<NewValue>(eventLoop: self.eventLoop, error: error, file: file, line: line)
+        let next = EventLoopPromise<NewValue>(eventLoop: eventLoop, file: file, line: line)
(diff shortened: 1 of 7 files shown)


Lukasa (approved):
Cool, LGTM. Do we have allocation counter tests for these? If not, it'd be nice to have them.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -506,13 +506,22 @@ extension EventLoopFuture {
...
-                return EventLoopFuture<NewValue>(eventLoop: self.eventLoop, error: error, file: file, line: line)
+        let next = EventLoopPromise<NewValue>(eventLoop: eventLoop, file: file, line: line)
+        self._whenComplete {
+            switch self._value! {
+            case .success(let t):
+                do {
+                    let r = try callback(t)
+                    return next._setValue(value: .success(r))
+                }
+                catch {
```
glbrntt, line 517: nit

```suggestion
                } catch {
```

Sources/NIO/EventLoopFuture.swift
```diff
@@ -531,13 +540,22 @@ extension EventLoopFuture {
...
+        self._whenComplete {
+            switch self._value! {
+            case .success(let t):
+                return next._setValue(value: .success(t))
+            case .failure(let e):
+                do {
+                    let r = try callback(e)
+                    return next._setValue(value: .success(r))
+                }
+                catch {
```
glbrntt, line 553:
```suggestion
                } catch {
```

glbrntt (approved):
Nice!

weissi (approved):
🎉 thank you!

Issue #1697, opened 2020-11-30
Avoid allocating when changing EventLoopFuture to Void type.


PeterAdams-A:
### Expected behavior
No Allocation

### Actual behavior
Allocation

### Steps to reproduce

Code like the following allocates

`return streamPromise.futureResult.map { _ in }`

The purpose of this code is to convert and EventLoopFuture<Something> into an EventLoopFuture<Void>

### If possible, minimal yet complete reproducer code (or URL to code)

`return streamPromise.futureResult.map { _ in }`

### SwiftNIO version/commit hash

2.20

### Swift & OS version (output of `swift --version && uname -a`)

5.3

PR #1818, opened 2021-04-26
Reducing allocations in future chains

fabianfett:
Less allocs = Faster code. Follow up to #1816

### Motivation:

- We alloc quite a lot with in implementations of EventLoopFuture chaining methods.
- While we don't use Futures a lot in NIO itself a lot of our users, depend quite a bit on them. Let's make their code faster.

### Modifications:

Create a `Promise` and use `_whenComplete` directly instead of going through another `flatMap`/`map` methods

### Result:

- Less allocations when using EventLoopFuture chains.
- We now have done what we is possible for: https://github.com/apple/swift-nio/issues/1697

### DIFF
--- a/Sources/NIO/EventLoopFuture.swift
+++ b/Sources/NIO/EventLoopFuture.swift
@@ -588,10 +588,13 @@ extension EventLoopFuture {
             self.whenSuccess(callback as! (Value) -> Void)
             return self as! EventLoopFuture<NewValue>
         } else {
-            return self.flatMap(file: file, line: line) { return EventLoopFuture<NewValue>(eventLoop: self.eventLoop, value: callback($0), file: file, line: line) }
+            let next = EventLoopPromise<NewValue>(eventLoop: eventLoop, file: file, line: line)
+            self._whenComplete {
+                return next._setValue(value: self._value!.map(callback))
+            }
+            return next.futureResult
         }
     }
-
 
     /// When the current `EventLoopFuture<Value>` is in an error state, run the provided callback, which
     /// may recover from the error by returning an `EventLoopFuture<NewValue>`. The callback is intended to potentially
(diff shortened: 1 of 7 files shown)


Sources/NIO/EventLoopFuture.swift
```diff
@@ -667,9 +682,16 @@ extension EventLoopFuture {
     /// - returns: A future that will receive the recovered value.
     @inlinable
     public func recover(file: StaticString = #file, line: UInt = #line, _ callback: @escaping (Error) -> Value) -> EventLoopFuture<Value> {
-        return self.flatMapError(file: file, line: line) {
-            return EventLoopFuture<Value>(eventLoop: self.eventLoop, value: callback($0), file: file, line: line)
+        let next = EventLoopPromise<Value>(eventLoop: eventLoop, file: file, line: line)
+        self._whenComplete {
+            switch self._value! {
```
glbrntt, line 687: Similar note to above, can we use `self._value!.mapError` here?

glbrntt (commented):
Nice allocation savings @fabianfett! Just one suggestion to tidy things up a little.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -588,11 +588,19 @@ extension EventLoopFuture {
...
         } else {
-            return self.flatMap(file: file, line: line) { return EventLoopFuture<NewValue>(eventLoop: self.eventLoop, value: callback($0), file: file, line: line) }
+            let next = EventLoopPromise<NewValue>(eventLoop: eventLoop, file: file, line: line)
+            self._whenComplete {
+                switch self._value! {
+                case .success(let t):
+                    return next._setValue(value: .success(callback(t)))
+                case .failure(let e):
+                    return next._setValue(value: .failure(e))
+                }
```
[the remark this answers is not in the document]
fabianfett: Great suggestion. Applied!

Sources/NIO/EventLoopFuture.swift
fabianfett: `mapError` sadly doesn't work here. Because it maps `Error` to a new `Error`. But in this case we want to map `Error` -> `Value`.

glbrntt: Oh yes, my mistake!

glbrntt (approved):
LGTM, thanks!

Sources/NIO/EventLoopFuture.swift
```diff
@@ -643,14 +646,21 @@ extension EventLoopFuture {
...
             case .success(let value):
-                return self.eventLoop.makeSucceededFuture(value, file: file, line: line)
-            case .failure(let error):
-                return self.eventLoop.makeFailedFuture(error, file: file, line: line)
+                switch body(value) {
+                case .success(let newValue):
+                    return next._setValue(value: .success(newValue))
+                case .failure(let error):
+                    return next._setValue(value: .failure(error))
+                }
```
Lukasa, line 658: The `switch` here is unnecessary: this can just be `return next._setValue(value: body(value))`

Lukasa (changes requested):
One minor cleanup needed here.

weissi (approved):
Thank you! My only comment is what @Lukasa already said, so once that's done I'm good to go!

Sources/NIO/EventLoopFuture.swift
fabianfett: This btw explains why we also had a switch in the code we replace:

```swift
        return self.flatMap(file: file, line: line) { value in
            switch body(value) {
            case .success(let value):
                return self.eventLoop.makeSucceededFuture(value, file: file, line: line)
            case .failure(let error):
                return self.eventLoop.makeFailedFuture(error, file: file, line: line)
            }
        }
```

Lukasa: Christ what a pain. Fair enough.

Lukasa (approved):

PR #1819, opened 2021-04-26
Implement reserveCapacity for CircularBuffer.

Lukasa:
Motivation:

While spelunking through the code I noticed we called
CircularBuffer.reserveCapacity, but never actually implemented it. This
means it never did anything. It should do something.

Modifications:

- Implemented reserveCapacity.

Result:

reserveCapacity does something now.

### DIFF
--- a/Sources/NIO/CircularBuffer.swift
+++ b/Sources/NIO/CircularBuffer.swift
@@ -322,22 +322,59 @@ extension CircularBuffer {
     }
 
     /// Double the capacity of the buffer and adjust the headIdx and tailIdx.
+    ///
+    /// Must only be called when buffer is full.
     @inlinable
     internal mutating func _doubleCapacity() {
+        // Double the storage. This can't use _resizeAndFlatten because the buffer is
+        // full at this stage. That's ok: we have some optimised code paths for this use-case.
+        let newCapacity = self.capacity << 1
+        assert(self.headBackingIndex == self.tailBackingIndex)
+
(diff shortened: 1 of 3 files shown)


weissi (approved):
Thanks! LGTM

Sources/NIO/CircularBuffer.swift
```diff
@@ -682,6 +719,20 @@ extension CircularBuffer: RangeReplaceableCollection {
         }
         return self[self.startIndex]
     }
+
+    /// Prepares the `CircularBuffer` to store the specified number of elements.
+    @inlinable
+    public mutating func reserveCapacity(_ capacity: Int) {
```
glbrntt, line 725: Should we make it clearer that we'll reserve enough capacity for _at least_ the specified number of elements and use `minimumCapacity` rather than `capacity` for the label?

glbrntt (commented):
Looks great, one minor question.

Sources/NIO/CircularBuffer.swift
Lukasa: Strictly the label is `_`. We can’t change that because the label is a part of the protocol requirement. We can change the inner name to `minimumCapacity`, but I’m not sure if that helps at all.

glbrntt: The inner label shows up in autocompletion which I find helpful. We could also just mention in the docs.

glbrntt (approved):

Issue #1811, opened 2021-04-20
WebSocket continuation (Fragmentation)


dnadoba:
Hi everyone,

Does SwiftNIO include something similar to [`WebSocketFrameAggregator`](https://github.com/netty/netty/blob/e5951d46fc89db507ba7d2968d2ede26378f0b04/codec-http/src/main/java/io/netty/handler/codec/http/websocketx/WebSocketFrameAggregator.java) from Netty?
It buffers incoming WebSocketFrames with the continuation opcode.

I didn't found anything in the swift-nio repository.
[Vapor has implemented buffering to handle the continuation opcode](https://github.com/vapor/websocket-kit/blob/bc3c30d6bb7a970b36ae933646750c4cc6109356/Sources/WebSocketKit/WebSocket.swift#L226-L234), but does not enforce a maximum buffer size.

Thanks
David

Lukasa:
Nope, SwiftNIO does not provide such a thing. If you wanted to implement it we'd be more than happy to accept the patch.

dnadoba:
Thank you for your quick response! I will implement it and open a PR.

PR #1823, opened 2021-04-28
WebSocket Frame Aggregator

dnadoba:
### Motivation:
We need a WebSocket fragment reassembly in RSocket. Solves #1811 

### Modifications:
add `NIOWebSocketFrameAggregator`
add tests for `NIOWebSocketFrameAggregator`

### Result:
`NIOWebSocketFrameAggregator` can now be added into a channel pipeline after a `WebSocketFrameDecoder` to aggregate fragmented WebSocket frames.


Lukasa:

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,138 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2019 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5: Nit: can we update the copyright date here?

```suggestion
// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```

PeterAdams-A:

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,138 @@
...
+
+
+/// `NIOWebSocketFrameAggregator` buffers inbound fragmented `WebSocketFrame`'s and aggregates them into a single `WebSocketFrame`.
+/// It guarantees that a `WebSocketFrame` with an `opcode` of `.continuation` is never forwarded.
+/// Frames which are not fragmented are just forwarded without any processing.
+/// Fragmented frames are unmasked, concatenated and forwarded as a new `WebSocketFrame` which is either a `.binary` or `.text` frame.
+/// `extensionData`, `rsv1`, `rsv2` and `rsv3` are lost if a frame is fragmented because they cannot be concatenated.
+/// - Note: `.ping`, `.pong`, `.closeConnection`frames are forwarded during frame aggregation
+public final class NIOWebSocketFrameAggregator: ChannelInboundHandler {
+    enum Error: Swift.Error {
```
Lukasa, line 24: This error should be `public`.

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,138 @@
...
+    
+    public func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let frame = unwrapInboundIn(data)
+        do {
+            switch frame.opcode {
+            case .continuation:
+                guard let firstFrameOpcode = bufferedFrames.first?.opcode else {
+                    throw Error.didReceiveFragmentBeforeReceivingTextOrBinaryFrame
+                }
+                try bufferFrame(frame)
```
Lukasa, line 70: A mild nit here and throughout this function: NIO prefers to use the `self` keyword explicitly where possible.

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,138 @@
...
+        for frame in bufferedFrames {
+            var unmaskedData = frame.unmaskedData
+            dataBuffer.writeBuffer(&unmaskedData)
+        }
+        
+        return WebSocketFrame(fin: true, opcode: opcode, data: dataBuffer)
+    }
+    
+    private func clearBuffer() {
+        bufferedFrames = []
```
Lukasa, line 136: To reduce allocator traffic it might be best to preserve the existing allocated capacity where we can: maybe use `bufferedFrames.removeAll(keepingCapacity: true)`?

Lukasa (changes requested):
Nice, this is a really great patch! I don't have any substantive feedback on the implementation itself, I have just one or two little notes and a stylistic note or two. If you clean those up I think we can get this merged.

Lukasa:
Also, as a note:

1. The 5.0 and 5.1 tests are failing because you're overriding an xctest method that doesn't exist in those versions. You'll need to find a way to avoid using `tearDownWithError`.
2. The soundness checks are failing because you haven't run `scripts/generate_linux_tests.rb`. Mind running that?

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,138 @@
...
+//===----------------------------------------------------------------------===//
+import NIO
+
+
+/// `NIOWebSocketFrameAggregator` buffers inbound fragmented `WebSocketFrame`'s and aggregates them into a single `WebSocketFrame`.
+/// It guarantees that a `WebSocketFrame` with an `opcode` of `.continuation` is never forwarded.
+/// Frames which are not fragmented are just forwarded without any processing.
+/// Fragmented frames are unmasked, concatenated and forwarded as a new `WebSocketFrame` which is either a `.binary` or `.text` frame.
+/// `extensionData`, `rsv1`, `rsv2` and `rsv3` are lost if a frame is fragmented because they cannot be concatenated.
+/// - Note: `.ping`, `.pong`, `.closeConnection`frames are forwarded during frame aggregation
```
fabianfett, line 22:
```suggestion
/// - Note: `.ping`, `.pong`, `.closeConnection` frames are forwarded during frame aggregation
```

weissi:

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
dnadoba: Done. 
`accumulatedFrameSize` is currently  also not cached and it just iterates over all buffered frames to get the current accumulated size. We should probably also cache the sum to improve performance. WDYT?

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
dnadoba: Done. Do you have a documented code style guide?

Tests/NIOWebSocketTests/NIOWebSocketFrameAggregatorTests.swift
```diff
@@ -17,53 +17,35 @@ import NIO
 import NIOWebSocket
 
 final class NIOWebSocketFrameAggregatorTests: XCTestCase {
-    var channel: EmbeddedChannel!
+    let channel = EmbeddedChannel(
```
weissi, line 20: This is actually a problem for weird `XCTestCase` reasons. `EmbeddedChannel` wants you to call `finish` when you're done with it. For some weird reasons however each `XCTestCase` gets initialised _twice_ (and your code only runs on one of those). Hence, if you have a `let something: ...` it'll be initialised but you can't finish its lifecycle.

I'd recommend going for `var channel: EmbeddedChannel!`, configuring it in `setUp` and tearing it down in `tearDown`.

@glbrntt I got this right with the `XCTestCase`, right?

glbrntt: Yes that sounds right but IIRC this only happens if you select a single test to run in Xcode, running all tests is fine.

Tests/NIOWebSocketTests/NIOWebSocketFrameAggregatorTests.swift
```diff
@@ -0,0 +1,156 @@
...
+    }
+    
+    override func tearDown() {
+        XCTAssertEqual(try self.channel.finish().isClean, true)
+    }
+    
+    func testEmptyButFinalFrameIsForwardedEvenIfMinNonFinalFragmentSizeIsGreaterThanZero() throws {
+        let frame = WebSocketFrame(fin: true, opcode: .binary, data: ByteBuffer())
+        try channel.writeInbound(frame)
+        XCTAssertEqual(try channel.readInbound(as: WebSocketFrame.self), frame)
```
fabianfett, line 39:
```suggestion
    func testEmptyButFinalFrameIsForwardedEvenIfMinNonFinalFragmentSizeIsGreaterThanZero() {
        let frame = WebSocketFrame(fin: true, opcode: .binary, data: ByteBuffer())
        XCTAssertNoThrow(try channel.writeInbound(frame))
        XCTAssertEqual(try channel.readInbound(as: WebSocketFrame.self), frame)
    }
```

One nit that I've just seen. All added tests potentially throw. Those throwing tests sadly have unclear error messages in case they actually throw, which is unfortunate in CI setups. Would you mind wrapping all `try`s in an `XCTAssertNoThrow`?

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
Lukasa: I'm afraid we don't. Our style is fairly lenient most of the time. One day Swift will decide on a common style and then we'll take a big jump over to that but until then we're kinda stuck doing this.

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
Lukasa: Yup, I think that's a good idea.

Tests/NIOWebSocketTests/NIOWebSocketFrameAggregatorTests.swift
dnadoba: Thanks, didn't know that! But does EmbeddedChannel requires explicit shutdown even if if was never used? I thought that  only `MultithreadedEventLoopGroup` does.

Tests/NIOWebSocketTests/NIOWebSocketFrameAggregatorTests.swift
dnadoba: Done. Hower, Xcode shows me the line which throws e.g.:
Screenshot 2021-04-29 at 11 31 59
Is this just a limitation of CLI or Linux?

Lukasa: I believe the answer is "both".

dnadoba:
Anything left to do or I can improve?

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,136 @@
...
+        self.accumulatedFrameSize += frame.length
+        
+        guard self.accumulatedFrameSize <= self.maxAccumulatedFrameSize else {
+            throw Error.accumulatedFrameSizeIsTooLarge
+        }
+    }
+    
+    private func aggregateFrames(opcode: WebSocketOpcode) -> WebSocketFrame {
+        var dataBuffer = ByteBuffer()
+        dataBuffer.reserveCapacity(minimumWritableBytes: self.accumulatedFrameSize)
```
fabianfett, line 122: Since we are in the Channel pipeline, do we maybe want to use the channel's byteBufferAllocator for this?

Lukasa: We do indeed, good catch @fabianfett.

Lukasa: In principle a minor optimisation available to us is to use the buffer from the first frame to accumulate the rest, saving a copy. That also avoids the need to deal with this issue.

Lukasa (commented):
I think once we clean up the ByteBuffer allocation we're good.

Sources/NIOWebSocket/NIOWebSocketFrameAggregator.swift
```diff
@@ -0,0 +1,142 @@
...
+            switch frame.opcode {
+            case .continuation:
+                guard let firstFrame = self.bufferedFrames.first else {
+                    throw Error.didReceiveFragmentBeforeReceivingTextOrBinaryFrame
+                }
+                try self.bufferFrame(frame)
+                
+                guard frame.fin else { break }
+                // final frame received
+                let aggregatedFrame = firstFrame.aggregated(
```
Lukasa, line 69: Hmm, sorry, now that I see this I think this will trigger a copy-on-write.

Fixing the CoW will be very fiddly so it might just be best to go back to the previous design but use the `context.channel.allocator` to get the ByteBuffer. Sorry for wasting your time there!

dnadoba: Don't worry :)

Lukasa (approved):
Great, I think this looks really good.

PR #1826, opened 2021-04-28
Stop measuring alloc counts on 5.0

Lukasa:
Motivation:

5.0 is several years old now, and while we still support it there is
beginning to be some conflict with performance optimizations for newer
Swift versions. We should stop counting the allocation counts on this
version now: we no longer care as much about regressions.

Modifications:

- Stop setting alloc counter limits on 5.0
- Remove the script that tries to add them back.

Result:

No longer measure allocs on 5.0

### DIFF
--- a/dev/update-alloc-limits-to-last-completed-ci-build
+++ b/dev/update-alloc-limits-to-last-completed-ci-build
@@ -6,7 +6,7 @@ set -o pipefail
 here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 tmpdir=$(mktemp -d /tmp/.last-build_XXXXXX)
 
-for f in 50 51 52 53 54 -nightly; do
+for f in 51 52 53 54 -nightly; do
     echo "swift$f"
     url="https://ci.swiftserver.group/job/swift-nio2-swift$f-prb/lastCompletedBuild/consoleFull"
     echo "$url"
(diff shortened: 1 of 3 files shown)


Davidde94 (approved):

weissi (approved):
🎉

glbrntt (commented):
Don't we need to not run the tests in CI for 5.0? Won't the tests just run and fail because no limits are known?

glbrntt:
> Don't we need to not run the tests in CI for 5.0? Won't the tests just run and fail because no limits are known?

Oh, 5.0 has just got green, apparently not.

glbrntt (approved):

weissi:
@Lukasa to George's point: should we also edit the docker file to skip the perf tests? something like `-f tests_0[012356789]` (note the missing `04` which IIRC are the perf tests)?

glbrntt:
> @Lukasa to George's point: should we also edit the docker file to skip the perf tests? something like `IntegrationTests/run-tests.sh -f tests_0[012356789]` (note the missing `04` which IIRC are the perf tests)?

@weissi further question: should the tests fail if there are no limits, or is that by design?

Lukasa:
Dockerfile edited

weissi:
> @weissi further question: should the tests fail if there are no limits, or is that by design?

that's by design. We can't expect everybody to have limits configured (we emit warnings with limit suggestions though). But we can expect everybody to be able to run the tests.

weissi (approved):
thanks!

weissi:
@Lukasa 

```
13:29:58 FAILURE: assertion '95001' <= '95000' failed
```

in 5.3. This is a result of a too-tight limit and slight instability. In the 5.3 file, mind upping that limit from 95000 to  95050 ?

PR #1833, opened 2021-04-28
bring back accidental commented benchmark

weissi:
Motivation:

In #1733 I accidentally commented out a perf benchmark (who knows why).

Modification:

Bring it back.

Result:

Benchmark is back.


Lukasa (approved):
Uh...whoops.

PR #1839, opened 2021-04-30
ByteBuffer: one fewer allocs to go to Data

weissi:
Motivation:

When getting a `Data` from a `ByteBuffer` we currently allocate twice
(`__DataStorage`) and the closure for `Data.Deallocator`.

Modifications:

We can optimise that by making the closure capture exactly one
`AnyObject` which is already a reference counted object. The compiler
realises that (on Linux) and saves us an alloc.

Thanks @lukasa for the suggestion here: https://github.com/apple/swift-nio/pull/1836#issuecomment-830044155

Result:

Fewer allocs.

### DIFF
--- a/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
+++ b/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2024 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


IntegrationTests/tests_04_performance/test_01_resources/test_bytebuffer_lots_of_rw.swift
```diff
@@ -25,37 +26,39 @@ func run(identifier: String) {
         @inline(never)
         func doWrites(buffer: inout ByteBuffer) {
             /* these ones are zero allocations */
-            // buffer.writeBytes(foundationData) // see SR-7542
```
Lukasa, line 28: Is this gone on purpose?

IntegrationTests/tests_04_performance/test_01_resources/test_bytebuffer_lots_of_rw.swift
```diff
@@ -25,37 +26,39 @@ func run(identifier: String) {
...
             buffer.writeBytes([0x41])
             buffer.writeBytes("A".utf8)
             buffer.writeString("A")
             buffer.writeStaticString("A")
             buffer.writeInteger(0x41, as: UInt8.self)
-
-            /* those down here should be one allocation each (on Linux) */
-            buffer.writeBytes(dispatchData) // see https://bugs.swift.org/browse/SR-9597
```
Lukasa, line 36: Similarly, is this gone on purpose?

Lukasa (changes requested):

Lukasa (approved):

weissi:
We should resurrect this

PR #1840, opened 2021-05-01
dev/lldb-smoker: LLDB smoke test using NIO

weissi:
Motivation:

We often hit a bunch of problems with LLDB, especially on Linux. So
maybe it would be cool if we could smoke test LLDB using SwiftNIO.

Modifications:

Add script `dev/lldb-smoker` which does the following things:

- takes 2000 random picks amongst all the files in the package
- for each of those files, pick a random line number
- write an LLDB script which sets break points for each of the file/line
  pairs (one-shot breakpoints)
- runs the program
- for each breakpoint hit, run `frame variable` to list all the
  variables

Result:

LLDB smoke test

### DIFF
--- a/dev/lldb-smoker
+++ b/dev/lldb-smoker
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+set -eu
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+number_of_breaks=2000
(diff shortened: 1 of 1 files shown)


weissi:
CC @tomerd / @aprantl / @fredriss

Lukasa (approved):

PR #1850, opened 2021-05-06
Rewrite RecvByteBufferAllocator.

### LINKED ISSUE #1848: Change adaptive allocator size table to work on platforms with 32-bit Int
`AdaptiveRecvByteBufferAllocator.sizeTable` is calculated as follows:

https://github.com/apple/swift-nio/blob/c51907f551c6b2e4366b5ccbe8c0509b0b061026/Sources/NIO/RecvByteBufferAllocator.swift#L55-L71

On platforms where `Int` is 32-bits (such as watchOS), the loop starting on line 65 will either trap, as it attempts to stuff a `UInt32.max` into an `Int`, or it will trap, when we shift a 1 into the sign bit and attempt to coerce a negative `Int` into a `UInt32`. Either way, the loop does not function as intended.

Lukasa:
Motivation:

Platforms with 32-bit integers continue to exist. On those platforms,
the way we calculate the size table for AdaptiveRecvByteBufferAllocator
will trap, as we attempt to compare an Int to UInt32.max as a loop
termination condition.

While I was amending this code, I noticed that
AdaptiveRecvByteBufferAllocator had a number of very strange behaviours,
and was arguably excessively complex. To that end, this patch
constitutes a substantial rewrite of the allocator. To understand what
it does, we need to describe the previous allocator algorithm.

The goal of AdaptiveRecvByteBufferAllocator is to attempt to dynamically
track the throughput of a TCP connection and to minimise the memory
usage required to get it to achieve maximum throughput, within
user-defined constraints. To that end, it implements a fairly simple
resizing algorithm.

At a high level, the algorithm is as follows. The allocator keeps track
of the size of the buffer it is offering the user. When the user reports
how much of that buffer they actually used, the allocator determines
whether better throughput could be achieved with larger allocation
sizes, or whether throughput would not be harmed with smaller allocation
sizes. It then adjusts accordingly.

When does the allocator believe beetter throughput could be achieved
with larger allocation sizes? When the allocation was entirely filled.
If a network recv() entirely fills the buffer, that means there was
likely more data to read that could not be added to the buffer. Improved
throughput would be gained by using larger buffers, as fewer system
calls would be necessary.

Similarly, the allocator believes it could shrink the buffer size
without harming throughput when the recv() uses less memory than the
next buffer size down. However, the algorithm attempts to detect the
possibility that we have "read until the end". The reason this is
relevant is that after a read completes we will often serve other work
on the loop for a while, causing data to pile up. As a result, we don't
want to shrink the buffer if the average read size would be higher, just
because one read happened to be short.

The original implementation's algorithm was based on a bucketed
scheme. For allocation sizes below 512 bytes, the allocation buckets
moved 16 bytes at a time. In principle this gave the allocator 16 byte
granularity. For allocation sizes above 512 bytes, the allocation
buckets would double: 512, 1024, 2048, 4096, and so on, up to
UInt32.max. This, incidentally, was the source of the crash on 32-bit
systems.

Unfortunately, this scheme had a few failings.

Firstly, the implementation complexity was fairly high. The bounds
provided by the user needed to be turned into bucket indices,
necessitating an awkward and complex binary search of the bucket array.
The bucket array itself needed to be generated, forcing a dispatch_once
to guard it and dirtying memory.

Secondly, the scheme was weighted very heavily toward growing memory and
not giving it back. Whenever the buffer was filled and a new size up
wanted to be chosen, the allocator would jump _4_ size buckets. As the
default initial size was 2048 bytes, and the default maximum was 65kB,
the first read of 2048 bytes would cause the allocator to immediately
jump up to allocating 32kB of memory for the next read: a huge leap! It
would then only release memory after two short reads, at which point it
would drop back only 1 bucket, meaning that there was a very aggressive
sawtooth pattern favouring higher memory usage. This pattern favours
benchmarks, where high-throughput localhost connections will naturally
want larger buffer sizes, but it is unnecessarily aggressive for
real-world networks.

Thirdly, the scheme had a bug that would mean it did not require two
_consecutive_ short reads to shrink the buffer, just two short reads
between an increase. This means that it had a tendency not to stabilise:
two "read to the end"s in two different event loop ticks would cause
the buffer to shrink, even if there had been 10 almost-full-reads
in between.

Fourthly, the system would occasionally report that the Channel should
reallocate the buffer because it was larger, when in fact it wasn't:
we'd hit the max, and were never going to get larger. This forced
high-throughput Channels to excessively allocate, albeit only in systems
that customized this allocator (and basically no-one does).

Fifthly, the bucketing system was defeated by ByteBuffer. While having
16-byte granularity was nice in principle, ByteBuffer only allows
power-of-two allocation sizes. This meant that, at the low end, there
were several wasted buckets that were effectively identical. Between 256
bytes and 512 bytes there were 15 redundant buckets!

This last point is the main reason to justify a rewrite. The complexity
of the scheme was in principle justified by having fine, granular
control of allocation sizes. Given that those don't exist, there is no
reason not to simply resort to power-of-two sizing. Once we do that, we
can replace the complex bucketing system by simple shift-and-multiply
math. The effect of this is to produce smaller code (we save about 30%
of the instructions, even as we add some extra safety preconditions)
with no additional branching, and no need to load from a random table in
memory. This avoids the need to keep that table in cache, reducing cache
pressure in the hot read loop. Additionally, we can drop the index into
the table, which lets us save some per-Channel memory, further reducing
cache pressure.

Additionally, constructing one of these is now vastly cheaper. That
matters less (it's not really hot-code), but it does improve Channel
creation time somewhat.

Modifications:

- Remove the size table.
- Round all values to powers of two.
- Implement new "previous power of two" function.
- Cap the allocation size at the largest power of 2 representable in a
  32-bit Int.
- Add more tests.

Result:

The result is less code, simpler code, and faster code. No trapping on
32-bit integer platforms.

Resolves #1848

### DIFF
--- a/Sources/NIO/RecvByteBufferAllocator.swift
+++ b/Sources/NIO/RecvByteBufferAllocator.swift
@@ -48,37 +48,14 @@ public struct FixedSizeRecvByteBufferAllocator: RecvByteBufferAllocator {
 
 /// `RecvByteBufferAllocator` which will gracefully increment or decrement the buffer size on the feedback that was recorded.
 public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
-
-    private static let indexIncrement = 4
-    private static let indexDecrement = 1
-
-    private static let sizeTable: [Int] = {
-        var sizeTable: [Int] = []
-
-        var i: Int = 16
-        while i < 512 {
-            sizeTable.append(i)
-            i += 16
-        }
-
-        i = 512
-        while i < UInt32.max { // 1 << 32 max buffer size
-            sizeTable.append(i)
-            i <<= 1
-        }
-
-        return sizeTable
-    }()
-
-    private let minIndex: Int
-    private let maxIndex: Int
     public let minimum: Int
     public let maximum: Int
     public let initial: Int
 
-    private var index: Int
     private var nextReceiveBufferSize: Int
     private var decreaseNow: Bool
+
+    private static let maximumAllocationSize = 1 << 31
 
     public init() {
         self.init(minimum: 64, initial: 2048, maximum: 65536)
@@ -86,81 +63,53 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
 
     public init(minimum: Int, initial: Int, maximum: Int) {
         precondition(minimum >= 0, "minimum: \(minimum)")
-        precondition(initial > minimum, "initial: \(initial)")
-        precondition(maximum > initial, "maximum: \(maximum)")
+        precondition(initial >= minimum, "initial: \(initial)")
+        precondition(maximum >= initial, "maximum: \(maximum)")
 
-        let minIndex = AdaptiveRecvByteBufferAllocator.sizeTableIndex(minimum)
-        if AdaptiveRecvByteBufferAllocator.sizeTable[minIndex] < minimum {
-            self.minIndex = minIndex + 1
-        } else {
-            self.minIndex = minIndex
-        }
+        // We need to round all of these numbers to a power of 2. Initial will be rounded down,
+        // minimum down, and maximum up.
+        self.minimum = min(minimum, AdaptiveRecvByteBufferAllocator.maximumAllocationSize).previousPowerOf2()
+        self.initial = min(initial, AdaptiveRecvByteBufferAllocator.maximumAllocationSize).previousPowerOf2()
+        self.maximum = min(maximum, AdaptiveRecvByteBufferAllocator.maximumAllocationSize).nextPowerOf2()
 
-        let maxIndex = AdaptiveRecvByteBufferAllocator.sizeTableIndex(maximum)
-        if AdaptiveRecvByteBufferAllocator.sizeTable[maxIndex] > maximum {
-            self.maxIndex = maxIndex - 1
-        } else {
-            self.maxIndex = maxIndex
-        }
-
-        self.initial = initial
-        self.minimum = minimum
-        self.maximum = maximum
-
-        self.index = AdaptiveRecvByteBufferAllocator.sizeTableIndex(initial)
-        self.nextReceiveBufferSize = AdaptiveRecvByteBufferAllocator.sizeTable[index]
+        self.nextReceiveBufferSize = self.initial
         self.decreaseNow = false
     }
 
-    private static func sizeTableIndex(_ size: Int) -> Int {
-        var low: Int = 0
-        var high: Int = sizeTable.count - 1
-
-        repeat {
-            if high < low {
-                return low
-            }
-            if high == low {
-                return high
-            }
-
-            // It's important to put (...) around as >> is executed before + in Swift while in Java it's the other way around (doh!)
-            let mid = (low + high) >> 1
-
-            let a = sizeTable[mid]
-            let b = sizeTable[mid + 1]
-
-            if size > b {
-                low = mid + 1
-            } else if size < a {
-                high = mid - 1
-            } else if size == a {
-                return mid
-            } else {
-                return mid + 1
-            }
-        } while true
-    }
-
     public func buffer(allocator: ByteBufferAllocator) -> ByteBuffer {
-        return allocator.buffer(capacity: nextReceiveBufferSize)
+        return allocator.buffer(capacity: self.nextReceiveBufferSize)
     }
 
     public mutating func record(actualReadBytes: Int) -> Bool {
-        if actualReadBytes <= AdaptiveRecvByteBufferAllocator.sizeTable[max(0, index - AdaptiveRecvByteBufferAllocator.indexDecrement - 1)] {
-            if decreaseNow {
-                index = max(index - AdaptiveRecvByteBufferAllocator.indexDecrement, minIndex)
-                nextReceiveBufferSize = AdaptiveRecvByteBufferAllocator.sizeTable[index]
-                decreaseNow = false
-            } else {
-                decreaseNow = true
-            }
-        } else if actualReadBytes >= nextReceiveBufferSize {
-            index = min(index + AdaptiveRecvByteBufferAllocator.indexIncrement, maxIndex)
-            nextReceiveBufferSize = AdaptiveRecvByteBufferAllocator.sizeTable[index]
-            decreaseNow = false
+        precondition(self.nextReceiveBufferSize % 2 == 0)
+        precondition(self.nextReceiveBufferSize >= self.minimum)
+        precondition(self.nextReceiveBufferSize <= self.maximum)
 
+        var mayGrow = false
+
+        // This right shift is safe: nextReceiveBufferSize can never be negative, so this will stop at 0.
+        let lowerBound = self.nextReceiveBufferSize &>> 1
+
+        // Here we need to be careful with 32-bit systems: if maximum is too large then any shift or multiply will overflow, which
+        // we don't want. Instead we check, and clamp to this current value if we overflow.
+        let upperBoundCandidate = Int(truncatingIfNeeded: Int64(self.nextReceiveBufferSize) &<< 1)
+        let upperBound = upperBoundCandidate == 0 ? self.nextReceiveBufferSize : upperBoundCandidate
+
+        if actualReadBytes <= lowerBound && lowerBound >= self.minimum {
+            if self.decreaseNow {
+                self.nextReceiveBufferSize = lowerBound
+                self.decreaseNow = false
+            } else {
+                self.decreaseNow = true
+            }
+        } else if actualReadBytes >= self.nextReceiveBufferSize && upperBound <= self.maximum {
+            self.nextReceiveBufferSize = upperBound
+            self.decreaseNow = false
+            mayGrow = true
+        } else {
+            self.decreaseNow = false
         }
-        return actualReadBytes >= nextReceiveBufferSize
+
+        return mayGrow
     }
 }
--- a/Sources/NIO/ByteBuffer-int.swift
+++ b/Sources/NIO/ByteBuffer-int.swift
@@ -110,6 +110,16 @@ extension FixedWidthInteger {
         }
         return 1 << (Self.bitWidth - (self - 1).leadingZeroBitCount)
     }
+
+    /// Returns the previous power of 2, or self if it already is.
+    @inlinable
+    func previousPowerOf2() -> Self {
+        guard self != 0 else {
+            return 0
+        }
+
+        return 1 << ((Self.bitWidth - 1) - self.leadingZeroBitCount)
+    }
 }
 
 extension UInt32 {
--- a/Tests/NIOTests/RecvByteBufAllocatorTest+XCTest.swift
+++ b/Tests/NIOTests/RecvByteBufAllocatorTest+XCTest.swift
@@ -29,6 +29,9 @@ extension AdaptiveRecvByteBufferAllocatorTest {
       return [
                 ("testAdaptive", testAdaptive),
                 ("testFixed", testFixed),
+                ("testMaxAllocSizeIsIntMax", testMaxAllocSizeIsIntMax),
+                ("testAdaptiveRoundsValues", testAdaptiveRoundsValues),
+                ("testSettingMinimumAboveMaxAllowed", testSettingMinimumAboveMaxAllowed),
            ]
    }
 }
--- a/Tests/NIOTests/RecvByteBufAllocatorTest.swift
+++ b/Tests/NIOTests/RecvByteBufAllocatorTest.swift
@@ -15,47 +15,85 @@
 import XCTest
 import NIO
 
-public final class AdaptiveRecvByteBufferAllocatorTest : XCTestCase {
+final class AdaptiveRecvByteBufferAllocatorTest: XCTestCase {
     private let allocator = ByteBufferAllocator()
-    private var adaptive = AdaptiveRecvByteBufferAllocator(minimum: 64, initial: 1024, maximum: 16 * 1024)
-    private var fixed = FixedSizeRecvByteBufferAllocator(capacity: 1024)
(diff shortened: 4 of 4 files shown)


Sources/NIO/RecvByteBufferAllocator.swift
```diff
@@ -89,78 +66,50 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
         precondition(initial > minimum, "initial: \(initial)")
         precondition(maximum > initial, "maximum: \(maximum)")
```
glbrntt, line 67: Why can't `initial` and `maximum` be equal to `initial`?

Sources/NIO/RecvByteBufferAllocator.swift
```diff
@@ -89,78 +66,50 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
...
-        } else {
-            self.maxIndex = maxIndex
-        }
-
-        self.initial = initial
-        self.minimum = minimum
-        self.maximum = maximum
+        // We need to round all of these numbers to a power of 2. Initial will be rounded down,
+        // minimum down, and maximum up.
+        self.minimum = minimum.previousPowerOf2()
```
glbrntt, line 71: Do we need to `min` this with the max allocation size here as well? Looks like we could end up with `minimum > initial` if minimum and initial were sufficiently large

Lukasa:

glbrntt (commented):
Nice change; one question but looks good otherwise.

Lukasa:
While we're here, this generates lovely straight-line code:

```
NIOHTTP1Client`AdaptiveRecvByteBufferAllocator.record(actualReadBytes:):
NIOHTTP1Client[0xcb6b0] <+0>:   pushq  %r13
NIOHTTP1Client[0xcb6b2] <+2>:   movq   0x18(%r13), %rax
NIOHTTP1Client[0xcb6b6] <+6>:   testb  $0x1, %al
NIOHTTP1Client[0xcb6b8] <+8>:   jne    0xcb728                   ; <+120> [inlined] Swift runtime failure: precondition failure at RecvByteBufferAllocator.swift:85
NIOHTTP1Client[0xcb6ba] <+10>:  movq   (%r13), %rdx
NIOHTTP1Client[0xcb6be] <+14>:  cmpq   %rdx, %rax
NIOHTTP1Client[0xcb6c1] <+17>:  jl     0xcb728                   ; <+120> [inlined] Swift runtime failure: precondition failure at RecvByteBufferAllocator.swift:85
NIOHTTP1Client[0xcb6c3] <+19>:  movq   0x8(%r13), %r8
NIOHTTP1Client[0xcb6c7] <+23>:  cmpq   %rax, %r8
NIOHTTP1Client[0xcb6ca] <+26>:  jl     0xcb72a                   ; <+122> [inlined] Swift runtime failure: precondition failure at RecvByteBufferAllocator.swift:86
NIOHTTP1Client[0xcb6cc] <+28>:  movq   %rax, %rcx
NIOHTTP1Client[0xcb6cf] <+31>:  sarq   %rcx
NIOHTTP1Client[0xcb6d2] <+34>:  movq   %rax, %rsi
NIOHTTP1Client[0xcb6d5] <+37>:  addq   %rax, %rsi
NIOHTTP1Client[0xcb6d8] <+40>:  cmovoq %rax, %rsi
NIOHTTP1Client[0xcb6dc] <+44>:  cmpq   %rdi, %rcx
NIOHTTP1Client[0xcb6df] <+47>:  jl     0xcb6f9                   ; <+73> at RecvByteBufferAllocator.swift:105:35
NIOHTTP1Client[0xcb6e1] <+49>:  cmpq   %rdx, %rcx
NIOHTTP1Client[0xcb6e4] <+52>:  jl     0xcb6f9                   ; <+73> at RecvByteBufferAllocator.swift:105:35
NIOHTTP1Client[0xcb6e6] <+54>:  leaq   0x20(%r13), %rsi
NIOHTTP1Client[0xcb6ea] <+58>:  cmpb   $0x1, 0x20(%r13)
NIOHTTP1Client[0xcb6ef] <+63>:  jne    0xcb71f                   ; <+111> at RecvByteBufferAllocator.swift
NIOHTTP1Client[0xcb6f1] <+65>:  movq   %rcx, 0x18(%r13)
NIOHTTP1Client[0xcb6f5] <+69>:  xorl   %edx, %edx
NIOHTTP1Client[0xcb6f7] <+71>:  jmp    0xcb721                   ; <+113> at RecvByteBufferAllocator.swift
NIOHTTP1Client[0xcb6f9] <+73>:  cmpq   %rdi, %rax
NIOHTTP1Client[0xcb6fc] <+76>:  jg     0xcb714                   ; <+100> at RecvByteBufferAllocator.swift:110:30
NIOHTTP1Client[0xcb6fe] <+78>:  cmpq   %rsi, %r8
NIOHTTP1Client[0xcb701] <+81>:  jl     0xcb714                   ; <+100> at RecvByteBufferAllocator.swift:110:30
NIOHTTP1Client[0xcb703] <+83>:  movq   %rsi, 0x18(%r13)
NIOHTTP1Client[0xcb707] <+87>:  addq   $0x20, %r13
NIOHTTP1Client[0xcb70b] <+91>:  movb   $0x1, %al
NIOHTTP1Client[0xcb70d] <+93>:  xorl   %edx, %edx
NIOHTTP1Client[0xcb70f] <+95>:  movq   %r13, %rsi
NIOHTTP1Client[0xcb712] <+98>:  jmp    0xcb723                   ; <+115> at RecvByteBufferAllocator.swift
NIOHTTP1Client[0xcb714] <+100>: addq   $0x20, %r13
NIOHTTP1Client[0xcb718] <+104>: xorl   %edx, %edx
NIOHTTP1Client[0xcb71a] <+106>: movq   %r13, %rsi
NIOHTTP1Client[0xcb71d] <+109>: jmp    0xcb721                   ; <+113> at RecvByteBufferAllocator.swift
NIOHTTP1Client[0xcb71f] <+111>: movb   $0x1, %dl
NIOHTTP1Client[0xcb721] <+113>: xorl   %eax, %eax
NIOHTTP1Client[0xcb723] <+115>: movb   %dl, (%rsi)
NIOHTTP1Client[0xcb725] <+117>: popq   %r13
NIOHTTP1Client[0xcb727] <+119>: retq   
NIOHTTP1Client[0xcb728] <+120>: ud2    
NIOHTTP1Client[0xcb72a] <+122>: ud2   
```

Sources/NIO/RecvByteBufferAllocator.swift
```diff
@@ -89,78 +66,50 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
...
-            } else if size == a {
-                return mid
-            } else {
-                return mid + 1
-            }
-        } while true
-    }
+        // Here we need to be careful with 32-bit systems: if maximum is Int32.max then any shift or multiply will overflow, which
+        // we don't want. Instead we check, and clamp to this current value if we overflow.
+        let (upperBoundCandidate, upperBoundOverflow) = self.nextReceiveBufferSize.multipliedReportingOverflow(by: 2)
```
weissi, line 95: Why don't we make all these calculations on a type that's always big enough like `UInt32`? Then we can just shift and if the result is 0, then we overflowed?

Ie. `(UInt32(1) << 31) << 1 == 0`

Alternatively, we can just use `Int64` and also always shift and then check if `(self.nextReceiveBufferSize << 1) & Int64(UInt32.max) == 0` which then also would mean "too big".

weissi (commented):
Thank you! Definitely an improvement but at least one comment isn't quite right

Sources/NIO/RecvByteBufferAllocator.swift
Lukasa: No particular reason, but I wasn't planning to lift that constraint today.

Sources/NIO/RecvByteBufferAllocator.swift
Lukasa: Yes, we do. Good catch.

Sources/NIO/RecvByteBufferAllocator.swift
```diff
@@ -89,78 +66,50 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
...
-            } else if size < a {
-                high = mid - 1
-            } else if size == a {
-                return mid
-            } else {
-                return mid + 1
-            }
-        } while true
-    }
+        // Here we need to be careful with 32-bit systems: if maximum is Int32.max then any shift or multiply will overflow, which
```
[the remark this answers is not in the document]
Lukasa: Yeah, this comment is from an older version of this code.

Sources/NIO/RecvByteBufferAllocator.swift
Lukasa: We absolutely can do. The reason I shied away from it is there is exactly one operation that's at any risk of overflowing, this one, and we can very easily clamp it instead. That seemed to me to be much simpler than hoisting into Int64 space and then doing math and clamping back down on the way out, with the fringe benefit that the compiler flags register already has the information we want, i.e. this operation degrades to `add, cmov`.

weissi: Hmm, what I don't like about it is that it essentially runs different code on 32 and 64 bit platforms. The 64bit ones will never overflow.

Regarding the instruction selection:
- this is not a hot code path, we're going to allocate anyway
- I don't think (at least on 64bit) we'd generate worse code

But I don't mind really.

Lukasa: > this is not a hot code path, we're going to allocate anyway

This is not quite true. This function is called once per `recv`, and this calculation is done to decide if we'll allocate. In the steady state, we never allocate, but we still always do this math.

I agree though that it runs different code on 32-and-64-bit platforms. That's somewhat unavoidable: at the edges of this type we need to work in `Int`-space, and so _somewhere_ we need to transform the input values from `Int` to whatever, and then transform back on the way out. Those operations are also different code on different platforms. That is an unavoidable fact of this API.

You're absolutely right that the code is no worse on Int64, that wasn't quite what I was saying: what I was saying is that this solution is no less optimal than doing the math in Int64. (Side note: onlimited 32-bit systems, none of which Swift supports, the original implementation may generate better code than doing the math in Int64-space if those systems have a limited number of 64-bit registers or no 64-bit instructions. This is not a factor in my decision making because to a first approximation we support none of those systems, but I want to bring it up for completeness.)

FWIW, here's the rewritten version, I'll merge which ever one you want:

```swift
        let upperBoundCandidate = Int(truncatingIfNeeded: Int64(self.nextReceiveBufferSize) &<< 1)
        let upperBound = upperBoundCandidate == 0 ? self.nextReceiveBufferSize : upperBoundCandidate
```

weissi: > This is not quite true. This function is called once per `recv`, and this calculation is done to decide if we'll allocate. In the steady state, we never allocate, but we still always do this math.

fair enough, but `recv` means going to the kernel, which is even more expensive.


> I agree though that it runs different code on 32-and-64-bit platforms. That's somewhat unavoidable: at the edges of this type we need to work in `Int`-space, and so _somewhere_ we need to transform the input values from `Int` to whatever, and then transform back on the way out. Those operations are also different code on different platforms. That is an unavoidable fact of this API.

But we can do the math on a type where it _always_ works. And then transform the result into an `Int`. If we're over 2GB on 32bit, then yes this crashes but it also requires the _user_ to have configured an upper maximum of over 2GB.

If we want to be nice, we can just put an assertion where we take the users values and do `assert(userMax < Int.max)`, then they immediately get a crash when they configure the bogus value.


> You're absolutely right that the code is no worse on Int64, that wasn't quite what I was saying: what I was saying is that this solution is no less optimal than doing the math in Int64. (Side note: onlimited 32-bit systems, none of which Swift supports, the original implementation may generate better code than doing the math in Int64-space if those systems have a limited number of 64-bit registers or no 64-bit instructions. This is not a factor in my decision making because to a first approximation we support none of those systems, but I want to bring it up for completeness.)

Right, yeah, I completely ignore 32bit devices when it comes to perf. ByteBuffer for example is really bad on 32bit platforms because it won't fit in the existential...

> FWIW, here's the rewritten version, I'll merge which ever one you want:
> 
> ```swift
>         let upperBoundCandidate = Int(truncatingIfNeeded: Int64(self.nextReceiveBufferSize) &<< 1)
>         let upperBound = upperBoundCandidate == 0 ? self.nextReceiveBufferSize : upperBoundCandidate
> ```

I think that one's substantially nicer I think?

Lukasa: > But we can do the math on a type where it always works. And then transform the result into an Int. If we're over 2GB on 32bit, then yes this crashes but it also requires the user to have configured an upper maximum of over 2GB.

Again, I agree, but I think that looks substantially _worse_ because we have 4 inputs to this function that all need to be duplicated into Int64 land before we compute on them, and two that need to be turned back into `Int` land on the way out. At a certain point we're drowning in integer conversions, each one of which has to be evaluated for correctness in 32-bit mode. That's why my proposed change was to modify exactly one line to do the dance: it reduces the cognitive pressure on the rest of the function. And, of course, any slip of the finger to use the original variable names screws you over again. :wink:

> I think that one's substantially nicer I think?

I think it is strictly worse, but you're reviewing.

Lukasa: I already pushed the change.

glbrntt (approved):

weissi (approved):
Awesome, looks good to me!

Sources/NIO/RecvByteBufferAllocator.swift
weissi: @Lukasa I'm okay with either version, my approval won't change whichever one you end up picking.

PR #1852, opened 2021-05-07
dev/scripts: add missing license headers

weissi:
Motivation:

We need license headers but the soundness script couldn't check if the
scripts don't have a `.sh` extension.

Modification:

- Add missing license headers.
- Better license checking (search for shell shebangs too)

Result:

Proper license attribution.

### DIFF
--- a/scripts/soundness.sh
+++ b/scripts/soundness.sh
@@ -147,10 +147,18 @@ EOF
 
   (
     cd "$here/.."
-    find . \
-      \( \! -path './.build/*' -a \
-      \( "${matching_files[@]}" \) -a \
-      \( \! \( "${exceptions[@]}" \) \) \) | while read line; do
+    {
+        find . \
+            \( \! -path './.build/*' -a \
+            \( "${matching_files[@]}" \) -a \
+            \( \! \( "${exceptions[@]}" \) \) \)
+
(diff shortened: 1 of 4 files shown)


glbrntt (commented):
Is it worth extending the license checker to look in `dev` and `scripts` for files without extensions and assume they are bash?

weissi:
@glbrntt I totally thought I'd get away with this :P . Fixed now :)

scripts/soundness.sh
```diff
@@ -147,10 +147,18 @@ EOF
...
-      \( \! \( "${exceptions[@]}" \) \) \) | while read line; do
+    {
+        find . \
+            \( \! -path './.build/*' -a \
+            \( "${matching_files[@]}" \) -a \
+            \( \! \( "${exceptions[@]}" \) \) \)
+
+        if [[ "$language" = bash ]]; then
+            # add everything with a shell shebang too
+            git grep --full-name -l '#!/bin/bash'
```
weissi, line 158: @glbrntt that's the fix. Additionally, we now look for shebangs.

glbrntt (approved):
Looks good!

Lukasa (approved):

PR #1853, opened 2021-05-07
Tests should tolerate cpusets.

Lukasa:
Motivation:

Our current CPU pinning tests do not expect to be pinned to cores that
do not include core 0. That's unnecessary.

Modifications:

- Expect to pin to the first core in our cpuset.

Result:

Tests run inside cpusets.


weissi (approved):
Fab, looks good! Thank you

Davidde94 (approved):
Awesome!

PR #1855, opened 2021-05-24
Random WebSocket Request Key

dnadoba:
Add a method to generate a random WebSocket Request Key as defined in [RFC6455](https://tools.ietf.org/html/rfc6455#section-4.1).
### Motivation:

At least [RSocket](https://github.com/rsocket/rsocket-swift/blob/d9c087660b6f29d2eaf0404265a66608d009ea5e/Sources/RSocketWSTransport/WSTransport.swift#L25-L35) and [Vapor](https://github.com/vapor/websocket-kit/blob/a2d26b3de8b3be292f3208d1c74024f76ac503da/Sources/WebSocketKit/WebSocketClient.swift#L72-L77) have implemented a method to generate a random request key. 
Since generating a random request key is always required for a WebSocket client, this should be included in SwiftINO.
As a bonus this implementation does not depend on `Foundation` because we can use the internal base64 encoder already included in SwiftNIO.

### Modifications:

- add static methods `randomRequestKey(using:)` and `randomRequestKey()` on `NIOWebSocketClientUpgrader`
- use `randomRequestKey()` as the default `requestKey` for `NIOWebSocketClientUpgrader`
- document `NIOWebSocketClientUpgrader` initializer

### Result:

All users of `NIOWebSocketClientUpgrader` no longer need to implement a custom random request key method.


Tests/NIOWebSocketTests/NIOWebSocketClientUpgraderTests.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.nextRandomNumberIndex = nextRandomNumberIndex
+    }
+    mutating func next() -> UInt64 {
+        defer { nextRandomNumberIndex += 1 }
+        return numbers[nextRandomNumberIndex % numbers.count]
+    }
+}
+
+final class NIOWebSocketClientUpgraderTests: XCTestCase {
+    func testRandomRequestKey() {
```
Lukasa, line 34: Can we add a test of the system random number generator entry point as well?

Lukasa (changes requested):
One minor note here and I think we'll be good.

Tests/NIOWebSocketTests/NIOWebSocketClientUpgraderTests.swift
dnadoba: Good point. Done ✅

Sources/NIOWebSocket/NIOWebSocketClientUpgrader.swift
```diff
@@ -92,3 +97,25 @@ public final class NIOWebSocketClientUpgrader: NIOHTTPClientProtocolUpgrader {
         }
     }
 }
+
+extension NIOWebSocketClientUpgrader {
+    /// Generates a random WebSocket Request Key by generating 16 bytes randomly and encoding them as a base64 string as defined in RFC6455 https://tools.ietf.org/html/rfc6455#section-4.1
+    /// - Parameter generator: the `RandomNumberGenerator` used as a the source of randomness
+    /// - Returns: base64 encoded request key
+    public static func randomRequestKey<Generator>(
```
Lukasa, line 105: Can we make this `@inlinable`? That will help with the perf cost of the generic.

Lukasa (commented):
Great, I just missed the `@inlinable` annotation. If we add that we're good.

Sources/NIOWebSocket/NIOWebSocketClientUpgrader.swift
dnadoba: Done ✅

Sources/NIOWebSocket/Base64.swift
```diff
@@ -18,6 +18,7 @@
 extension String {
 
   /// Base64 encode a collection of UInt8 to a string, without the use of Foundation.
+  @usableFromInline
```
Lukasa, line 21: Oh sorry, can we make this `@inlinable` too?

dnadoba: Done ✅ but does it need to be `@inlinable` to be specialized? We call this method in `randomRequestKey(using:)` and the generic type (`ByteBufferView`) is known inside the module.

Lukasa: Yes it does.

The mental model to apply for when `@inlinable` will be needed is not whether the call-site _as written_ is inside the module, but whether the call-site _as compiled_ is inside the module. In this instance, `randomRequestKey(using:)` is `@inlinable`. When compiled and specialized, the effective call site of this code is inside the specialized code, generated by the caller. As a result the _effective_ call-site is in user code, not inside the module.

In generic situations like this you tend to have to add inlinables the whole way down the stack.

Lukasa (approved):

PR #1858, opened 2021-05-25
Add alloc counter benchmark for result erasing maps.

Lukasa:
Motivation:

Peter thinks that result-erasing maps should not allocate, and we have
special code paths in the code to try to make Void -> Void maps not
allocate. Sadly, both code paths currently do allocate.

Per our rules for not trying to make optimizations without data, we
should start measuing these closures so we can make optimizations.

Modifications:

- Added an alloc couter test for result-erasing maps.

Result:

Alloc counter test suitable for any fix of #1697.

### DIFF
--- a/IntegrationTests/tests_04_performance/test_01_resources/test_future_erase_result.swift
+++ b/IntegrationTests/tests_04_performance/test_01_resources/test_future_erase_result.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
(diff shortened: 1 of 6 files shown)


IntegrationTests/tests_04_performance/test_01_resources/test_future_erase_result.swift
```diff
@@ -0,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2019 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5:
```suggestion
// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```

glbrntt (approved):
Looks good once limits are added.

PR #1820, opened 2021-04-27
Reduce selector wakeups.

Lukasa:
Motivation:

Right now all tasks scheduled onto an event loop from a different loop
wake up the underlying selector by writing to an eventfd. This is a
fairly inefficient way to handle things: it incurs a lot of syscall
traffic unnecessarily.

Given that we currently protect the pending tasks queue with a big dumb
lock, we can safely keep track of whether we're going to dequeue this
task. If we are, we don't need to wake the selector.

Modifications:

- Keep track of whether we're de-queueing tasks or not.
- Arrange to wake the selector only once.

Result:

Cheaper task enqueueing on hot loops.

cc @hassila, who will be affected by this patch

PR #1860, opened 2021-05-27
Make Swift 5.2 the minimum requirement

Davidde94:
Make Swift 5.2 the minimum requirement, dropping support for Swift 5.0 and 5.1.

### Motivation:

Whenever we have problems, Swift 5.0 and 5.1 seem to be the culprits. Dropping support for these very old versions will require less maintenance and free up our time to work on new features.

### Modifications:

- Set the tools version in Package.swift to 5.2
- Remove CI configurations for 5.0 and 5.1
- Update the various readmes to reflect that this change will be rolled out in NIO 2.30.0

### Result:

Swift 5.2 is the minimum version of Swift required to use NIO.


docs/migration-guide-NIO1-to-NIO2.md
```diff
@@ -4,6 +4,8 @@ This migration guide is our recommendation to migrate from NIO 1 to NIO 2. None
 
 This repository also contains a fairly complete list of public [API changes](https://github.com/apple/swift-nio/blob/main/docs/public-api-changes-NIO1-to-NIO2.md) from NIO 1 to NIO 2.
 
+### Note: As of NIO 2.30.0 the minimum version of Swift required is 5.3.
```
0xTim, line 7: Typo?

```suggestion
### Note: As of NIO 2.30.0 the minimum version of Swift required is 5.2.
```

Davidde94: Oops, thanks!

dev/make-single-file-spm
```diff
@@ -51,7 +51,7 @@ function setup_swiftpm_package() {
 
     cd "$destination"
     cat > Package.swift <<"EOF"
-// swift-tools-version:5.1
+// swift-tools-version:5.2
```
Lukasa, line 54: Did you validate this script works? I suspect it now produces a broken SPM Package manifest.

Lukasa (changes requested):
This will need to be coordinated with @tomerd to turn off the relevant builders for SwiftNIO.

dev/make-single-file-spm
weissi: we also don't need this change. You can still depend on packages that have newer requirements with older manifests.

Davidde94: Good point

Davidde94:
@Lukasa yep! He's in the review list and I'm sending an email

glbrntt (dismissed):
I think a few other things need updating; `scripts/build_podspecs.sh` and `CONTRIBUTING.md`.

Also I think we should be able to get rid of the generated tests now and use `--enable-test-discovery` when it's necessary. Might be easier to do that in a follow up.

weissi:
> Also I think we should be able to get rid of the generated tests now and use `--enable-test-discovery` when it's necessary. Might be easier to do that in a follow up.

Sadly not: https://bugs.swift.org/browse/SR-12008

Davidde94:
@glbrntt we can't get rid of linux tests right now

glbrntt:
> > Also I think we should be able to get rid of the generated tests now and use `--enable-test-discovery` when it's necessary. Might be easier to do that in a follow up.
> 
> Sadly not: https://bugs.swift.org/browse/SR-12008

euuuughhhhh 😡😡😡

Lukasa (approved):

adam-fowler:
Is the setting of the Package minimum requirements something that should be coordinated with some of the key serverside swift package owners. I doubt any of them want to restrict what version of SwiftNIO their clients can use. Some obvious packages would be AHC, Soto, MongoSwift and RediStack. I think all the Vapor packages have a minimum requirement of 5.2 already.

adam-fowler:
> Is the setting of the Package minimum requirements something that should be coordinated with some of the key serverside swift package owners. I doubt any of them want to restrict what version of SwiftNIO their clients can use. Some obvious packages would be AHC, Soto, MongoSwift and RediStack. I think all the Vapor packages have a minimum requirement of 5.2 already.

After talking to @fabianfett  offline it appears this isn't an issue. If someone is compiling with Swift 5.2 or later they will pick up later versions of SwiftNIO even though a library in the middle says its minimum requirement is Swift 5.1. Unless that library indicates it wants version 2.30 or later of NIO.

finestructure:
In case you haven't seen the Twitter thread, here's a list of packages in the Swift Package Index that depend on NIO showing their Swift version compatibility as we compute it: https://www.icloud.com/numbers/049dkCvCOid9P8S0UxtCGE3iA#NIO-dependency-check

([Twitter thread](https://twitter.com/_sa_s/status/1398286782807035910?s=21))

Note that of the 101 packages (of 298) that are _not_ 5.2+ compatible quite a few are probably only flagged as incompatible because they use OpenSSL and we don't have it installed on our builders. I.e. there's a good chance the impact is smaller than that.

fabianfett (approved):
🚀🎉

README.md
```diff
@@ -20,6 +20,8 @@ Repository | NIO 2 (Swift 5+) | NIO 1 (Swift 4+)
 [https://github.com/apple/swift-nio-transport-services][repo-nio-transport-services] <br> first-class support for macOS, iOS, tvOS, and watchOS | `from: "1.0.0"` | `from: "0.1.0"`
 [https://github.com/apple/swift-nio-ssh][repo-nio-ssh] <br> SSH support | `.upToNextMinor(from: "0.2.0")` | _n/a_
 
+NIO 2.29.0 and older support Swift 5.0+.
```
PeterAdams-A, line 23: Wondering if there is a better way of representing this.  5.0 would work if they did `from: 2.0.0`

PeterAdams-A (approved):
OK, one comment but I don't have a concrete suggestion that's better.

README.md
Lukasa: 5.0 would work if they did `from`: anything 2.29.0 and earlier. I think this is probably sufficient.

Davidde94:

yim-lee:

yim-lee:

helje5:
> Whenever we have problems, Swift 5.0 and 5.1 seem to be the culprits. 
In what ways?

helje5:
A few issues raised in the Slack:

1. https://github.com/apple/swift-nio/pull/1647 - just affects building Windows w/ <5.2, no other platforms affected. Windows affected neither, just needs to use 5.2 there
2. https://github.com/apple/swift-nio/pull/1820 - perf only, could use different compile paths for the two
3. https://github.com/apple/swift-nio/pull/1823 - just a test setup affected, can be worked around

PR #1862, opened 2021-05-29
Improving performance of base64 encoding by about 10%

mzaks:
This change improves the performance of base64 encode by about 10%

### Motivation:

I was curious if I could make the code a bit faster.

### Modifications:

The main modification is to use `String(cString:)` constructor, which is the fastest way to insatiate a Swift `String` from a sequence of bytes. I also checked against `String(unsafeUninitializedCapacity capacity:, initializingUTF8With initializer:)` which performs comparably fast, but is only available since Swift version 5.3.

The change of String instantiation provides a speedup of about 10%.

Additionally, I changed the type of the `alphabet` to be a `StaticString` instead of `[UInt8]`. This change provided additional speedup of about 1-2%.

The two modification are unrelated and can be broken down into 2 separate PRs if desired.

### Result:

It is a non breaking change focusing purely on performance.

### DIFF
--- a/Sources/NIOWebSocket/Base64.swift
+++ b/Sources/NIOWebSocket/Base64.swift
@@ -41,58 +41,65 @@ internal struct Base64 {
     let base64StringLength = ((bytes.count + 2) / 3) * 4
     let alphabet = Base64.encodeBase64
     
-    // temporary buffer we use to store the output as zero terminated C string
-    let outputBuffer = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: base64StringLength + 1)
-    defer {
-      outputBuffer.deallocate()
+    return String(customUnsafeUninitializedCapacity: base64StringLength) { backingStorage in
+      var input = bytes.makeIterator()
+      var offset = 0
+      while let firstByte = input.next() {
+        let secondByte = input.next()
+        let thirdByte = input.next()
+        
+        backingStorage[offset] = Base64.encode(alphabet: alphabet, firstByte: firstByte)
+        backingStorage[offset + 1] = Base64.encode(alphabet: alphabet, firstByte: firstByte, secondByte: secondByte)
+        backingStorage[offset + 2] = Base64.encode(alphabet: alphabet, secondByte: secondByte, thirdByte: thirdByte)
+        backingStorage[offset + 3] = Base64.encode(alphabet: alphabet, thirdByte: thirdByte)
+        offset += 4
+      }
+      return offset
     }
-    
-    var input = bytes.makeIterator()
-    var offset = 0
-  
-    while let firstByte = input.next() {
-      let secondByte = input.next()
-      let thirdByte = input.next()
-      
-      outputBuffer[offset] = Base64.encode(alphabet: alphabet, firstByte: firstByte)
-      outputBuffer[offset + 1] = Base64.encode(alphabet: alphabet, firstByte: firstByte, secondByte: secondByte)
-      outputBuffer[offset + 2] = Base64.encode(alphabet: alphabet, secondByte: secondByte, thirdByte: thirdByte)
-      outputBuffer[offset + 3] = Base64.encode(alphabet: alphabet, thirdByte: thirdByte)
-      offset += 4
-    }
-    
-    // last char has to be a `0` in a zero terminated C string
-    outputBuffer[offset] = 0
-
-    // it is safe to force unwrap the baseAddress since we know that the output buffer is not empty
-    return String(cString: outputBuffer.baseAddress!)
   }
   
   // MARK: Internal
   
   // The base64 unicode table.
   @usableFromInline
-  static let encodeBase64: StaticString = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+  static let encodeBase64: [UInt8] = [
+    UInt8(ascii: "A"), UInt8(ascii: "B"), UInt8(ascii: "C"), UInt8(ascii: "D"),
+    UInt8(ascii: "E"), UInt8(ascii: "F"), UInt8(ascii: "G"), UInt8(ascii: "H"),
+    UInt8(ascii: "I"), UInt8(ascii: "J"), UInt8(ascii: "K"), UInt8(ascii: "L"),
+    UInt8(ascii: "M"), UInt8(ascii: "N"), UInt8(ascii: "O"), UInt8(ascii: "P"),
+    UInt8(ascii: "Q"), UInt8(ascii: "R"), UInt8(ascii: "S"), UInt8(ascii: "T"),
+    UInt8(ascii: "U"), UInt8(ascii: "V"), UInt8(ascii: "W"), UInt8(ascii: "X"),
+    UInt8(ascii: "Y"), UInt8(ascii: "Z"), UInt8(ascii: "a"), UInt8(ascii: "b"),
+    UInt8(ascii: "c"), UInt8(ascii: "d"), UInt8(ascii: "e"), UInt8(ascii: "f"),
+    UInt8(ascii: "g"), UInt8(ascii: "h"), UInt8(ascii: "i"), UInt8(ascii: "j"),
+    UInt8(ascii: "k"), UInt8(ascii: "l"), UInt8(ascii: "m"), UInt8(ascii: "n"),
+    UInt8(ascii: "o"), UInt8(ascii: "p"), UInt8(ascii: "q"), UInt8(ascii: "r"),
+    UInt8(ascii: "s"), UInt8(ascii: "t"), UInt8(ascii: "u"), UInt8(ascii: "v"),
+    UInt8(ascii: "w"), UInt8(ascii: "x"), UInt8(ascii: "y"), UInt8(ascii: "z"),
+    UInt8(ascii: "0"), UInt8(ascii: "1"), UInt8(ascii: "2"), UInt8(ascii: "3"),
+    UInt8(ascii: "4"), UInt8(ascii: "5"), UInt8(ascii: "6"), UInt8(ascii: "7"),
+    UInt8(ascii: "8"), UInt8(ascii: "9"), UInt8(ascii: "+"), UInt8(ascii: "/"),
+  ]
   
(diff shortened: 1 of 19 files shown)


Sources/NIOWebSocket/Base64.swift
```diff
@@ -33,80 +33,66 @@ internal struct Base64 {
...
     let alphabet = Base64.encodeBase64
     
-    // NOTE: Once SE-263 lands we should replace this implementation with one
-    //       that makes use of the non copy initializer. For more information
-    //       please see:
-    //       https://github.com/apple/swift-evolution/blob/master/proposals/0263-string-uninitialized-initializer.md
-    var outputBytes = [UInt8]()
-    outputBytes.reserveCapacity(newCapacity)
+    // temporary buffer we use to store the output as zero terminated C string
+    let outputBuffer = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: base64StringLength + 1)
```
Lukasa, line 45: I’m fairly disinclined to use an `UnsafeMutableBufferPointer` here. However, I would be more inclined to use `String.init(unsafeUninitializedCapacity:initializingWith:)`. We will need to provide a backport of that initializer, but one already exists in `swift-nio-http2` which we can bring over here: https://github.com/apple/swift-nio-http2/blob/1ef8bac96a4bbbc870dcf3c9825df56982188620/Sources/NIOHPACK/HuffmanCoding.swift#L247-L289

Sources/NIOWebSocket/Base64.swift
```diff
@@ -33,80 +33,66 @@ internal struct Base64 {
...
-    UInt8(ascii: "g"), UInt8(ascii: "h"), UInt8(ascii: "i"), UInt8(ascii: "j"),
-    UInt8(ascii: "k"), UInt8(ascii: "l"), UInt8(ascii: "m"), UInt8(ascii: "n"),
-    UInt8(ascii: "o"), UInt8(ascii: "p"), UInt8(ascii: "q"), UInt8(ascii: "r"),
-    UInt8(ascii: "s"), UInt8(ascii: "t"), UInt8(ascii: "u"), UInt8(ascii: "v"),
-    UInt8(ascii: "w"), UInt8(ascii: "x"), UInt8(ascii: "y"), UInt8(ascii: "z"),
-    UInt8(ascii: "0"), UInt8(ascii: "1"), UInt8(ascii: "2"), UInt8(ascii: "3"),
-    UInt8(ascii: "4"), UInt8(ascii: "5"), UInt8(ascii: "6"), UInt8(ascii: "7"),
-    UInt8(ascii: "8"), UInt8(ascii: "9"), UInt8(ascii: "+"), UInt8(ascii: "/"),
-  ]
+  static let encodeBase64: StaticString = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
```
Lukasa, line 75: I’m _highly_ suspicious of this change being helpful. There are very few data types in the standard library cheaper than `Array`. There _is_ one advantage, which is that `String` can be embedded directly in the binary where `Array<UInt8>` tends to fail to be embedded: https://bugs.swift.org/browse/SR-14102. 

Regardless, I’d like to pull this change out to a separate one so we can get an idea of exactly how meaningful it is.

Lukasa (changes requested):
Thanks for this patch! In general I’m happy to take this: however, we don’t currently have a base64 benchmark, so it’s hard for us to get a feel for whether these changes are paying for themselves.

Are you open to adding a benchmark? We can benchmark `WebSocketClientHandler.randomRequestKey` which is split between the costs of RNG and Base64, but should be enough to get some numbers at least. The example benchmarks are in the `NIOPerformanceTester` target.

mzaks:
> Thanks for this patch! In general I’m happy to take this: however, we don’t currently have a base64 benchmark, so it’s hard for us to get a feel for whether these changes are paying for themselves.
> 
> Are you open to adding a benchmark? We can benchmark `WebSocketClientHandler.randomRequestKey` which is split between the costs of RNG and Base64, but should be enough to get some numbers at least. The example benchmarks are in the `NIOPerformanceTester` target.

Yes sure. I will add benchmark to this PR, just not sure if I will be able to do it today.

Sources/NIOWebSocket/Base64.swift
mzaks: This being a `StaticString` it is definitely embedded, but I agree that it is better to put it in a separate PR.
I will remove this change from this PR.

Sources/NIOWebSocket/Base64.swift
mzaks: My benchmarks showed that the cString initialiser is still a tiny bit faster and it does not need a back port.
But I guess, I should introduce a bench mark in this PR first. 
This will make discussing and comparing strategies much better.

mzaks:
@Lukasa please find the performance test in following PR https://github.com/apple/swift-nio/pull/1863
This way I think it is easier to track changes in performance characteristics.

glbrntt (commented):
@mzaks FWIW this is vendored from https://github.com/swift-extras/swift-extras-base64, they may also appreciate this patch!

mzaks:
> @mzaks FWIW this is vendored from https://github.com/swift-extras/swift-extras-base64, they may also appreciate this patch!

They already use `String(unsafeUninitializedCapacity capacity:, initializingUTF8With initializer:)` but they do not use `String(cString:)` as fallback. Implementing `String(cString:)` as fallback will be a slightly bigger refactoring, if someone form the project is interested I can happily do it, but I am not sure if they see it as necessary.

glbrntt:
> > @mzaks FWIW this is vendored from https://github.com/swift-extras/swift-extras-base64, they may also appreciate this patch!
> 
> They already use `String(unsafeUninitializedCapacity capacity:, initializingUTF8With initializer:)` but they do not use `String(cString:)` as fallback. Implementing `String(cString:)` as fallback will be a slightly bigger refactoring, if someone form the project is interested I can happily do it, but I am not sure if they see it as necessary.

Ah fair enough, I didn't realise that.

Lukasa:
Cool, should we rebase this onto main now?

mzaks:
> Cool, should we rebase this onto main now?

@Lukasa yes please :)

Lukasa:
@mzaks It's probably easier for you to do the rebase, if you're comfortable. If not, let me know and I can do it.

mzaks:
@Lukasa there are no merge conflicts now, but to be honest I am not sure how I should rebase it. So help would be very appreciated

Lukasa:
A further rebase shouldn't be necessary, so long as it'll merge cleanly we'll squash it down.

Sources/NIOWebSocket/Base64.swift
```diff
@@ -33,36 +33,29 @@ internal struct Base64 {
   static func encode<Buffer: Collection>(bytes: Buffer)
     -> String where Buffer.Element == UInt8
   {
+    guard bytes.isEmpty == false else {
```
Lukasa, line 36: Rather than an explicit comparison to `false` can this just be `guard !bytes.isEmpty`?

mzaks:
@Lukasa ok, I see there are some issues on Jenkins, I will have a look.

Sources/NIOWebSocket/Base64.swift
```diff
@@ -128,3 +121,51 @@ internal struct Base64 {
...
+    // The buffer will store zero terminated C string
+    let buffer = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: capacity + 1)
+    defer {
+        buffer.deallocate()
+    }
+
+    let initializedCount = try initializer(buffer)
+    precondition(initializedCount <= capacity, "Overran buffer in initializer!")
+    // add zero termination
+    buffer[capacity] = 0
```
Lukasa, line 141: We need to null-terminate at `initializedCount`, not at `capacity`, or we risk using uninitialised memory here.

mzaks:
@Lukasa seems like all tests are passing. Is there a place I can see the result of the performance test? I am curious :)

Lukasa:
Right here!

Lukasa:

Davidde94:

Lukasa:
Hmm, those numbers are kinda all over the shop. That suggests a certain stability issue, maybe caused by us triggering two perf runs in one go.

mzaks:
@Lukasa yeah it is a strange result. Specifically as there are significant differences in places which we didn't touch at all.

Lukasa:
Right now we can consider that result bogus until proven otherwise. :grin:

Lukasa:
Got it.

Hey @tomerd/@yim-lee, can we change the docker image used by the NIO2 Performance PRB builder to be 2004.54 instead of 1604.52? We aren't getting accurate numbers for comparison I'm afraid.

Davidde94:
Worth running again (but not in parallel)? 😆

Lukasa:
It's not a parallel problem: as my comment above suggests, we're running against different Swift versions in the PR build than in the main build. Hence my request to Yim and Tom.

mzaks:
For comparison, when I run the NIOPerformanceTester in Release mode on my "Mac mini (M1, 2020)" I get following numbers for old build:
```
measuring: generate_10k_random_request_keys: 0.006438834, 0.008278834, 0.006697167, 0.007452042, 0.006186875, 0.006924625, 0.007962458, 0.007085584, 0.00637725, 0.006505917
```
and new one:
```
measuring: generate_10k_random_request_keys: 0.005487083, 0.005380458, 0.00822125, 0.0054175, 0.005368958, 0.005361917, 0.005341875, 0.005332584, 0.005336, 0.005467417
```

Lukasa:
That looks more likely to me.

Lukasa:

Lukasa:
Heyyyy that looks more likely! 

The TL;DR here is that we're in the noise per our builder. I'm not _overly_ surprised by that: I asked you to push some things out of the change, _and_ the change can be dominated by other costs (e.g. depending on how things go this can be swallowed up by the RNG costs or even the costs of allocating Strings).

mzaks:
@Lukasa so you think the change is still meaningful?

Lukasa:

Lukasa (approved):
Yes, I think this patch remains worthwhile.

PR #1863, opened 2021-05-30
Add performance test for web socket client random request key

mzaks:
Adding performance test for random request key creation

### Motivation:

Adding this performance test in order to compare performance improvements of base64 encode method

### Modifications:

Adding one performance test

### Result:

Performance test can be executed and the performance characteristics of base64 encode method can be measured and tracked

### DIFF
--- a/Sources/NIOPerformanceTester/main.swift
+++ b/Sources/NIOPerformanceTester/main.swift
@@ -16,6 +16,7 @@ import NIO
 import NIOHTTP1
 import NIOFoundationCompat
 import Dispatch
+import NIOWebSocket
 
 // MARK: Test Harness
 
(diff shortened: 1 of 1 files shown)


Lukasa:

Lukasa (approved):
Great, this looks fantastic. I'll run the benchmarks to confirm that everything works well but I can't see an issue.

Sources/NIOPerformanceTester/main.swift
```diff
@@ -770,3 +771,8 @@ try measureAndPrint(desc: "byte_buffer_view_iterator_1mb", benchmark: ByteBuffer
 
 try measureAndPrint(desc: "byte_to_message_decoder_decode_many_small",
                     benchmark: ByteToMessageDecoderDecodeManySmallsBenchmark(iterations: 1_000, bufferSize: 16384))
+
+measureAndPrint(desc: "generate_10k_random_request_keys") {
+    let keys = (0 ..< 10_000).map { _ in NIOWebSocketClientUpgrader.randomRequestKey() }
```
Lukasa, line 776: Oh, actually, I can: instead of using `map` let's use something else that introduces less allocator overhead. Perhaps `reduce(into: 0, { $0 &+= NIOWebSocketClientUpgrader.randomRequestKey().count })`. That should prevent the optimiser from killing the benchmark, but won't force the allocator to do extra work.

Lukasa (changes requested):

Sources/NIOPerformanceTester/main.swift
mzaks: Very good point!
Will change in a few minutes

Lukasa (approved):

Lukasa:

Lukasa:
Shiny, LGTM.

Issue #1867, opened 2021-06-03
Make CNIO* `extern` values `const`.


Lukasa:
CNIODarwin and CNIOLinux have got some extern values that should be made `const`, as we don't intend to allow them to ever change.

PR #1868, opened 2021-06-03
Made the extern to const in CNIODarwin

BastianInuk:
Externs in CNIODarwin are now const
### Motivation:

Externs should be immutable as per #1867.

### Modifications:

Externs in CNIODarwin got added the const keyword.

### Result:

Externs are now immutable

### DIFF
--- a/Sources/CNIODarwin/include/CNIODarwin.h
+++ b/Sources/CNIODarwin/include/CNIODarwin.h
@@ -31,11 +31,11 @@ typedef struct {
     unsigned int msg_len;
 } CNIODarwin_mmsghdr;
 
-extern int CNIODarwin_IPTOS_ECN_NOTECT;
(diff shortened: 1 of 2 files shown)


Lukasa:

Davidde94 (approved):

Lukasa:
I have unlinked #1867 because this only _partially_ fixes that issue.

Lukasa (approved):

BastianInuk:
I don't have a linux install at hand, so I can't swiftly also run tests on that 😅

Davidde94:
@BastianInuk if you install docker and `cd` into the repo then you can use this docker command:

```
docker run --rm -it -v "$PWD":"$PWD" -w "$PWD" swift:5.0 swift test
```

PR #1869, opened 2021-06-03
Made the externs const in CNIOLinux

BastianInuk:
Externs in CNIOLinux are now const

Motivation:

Externs should be immutable as per #1867.

Modifications:

Externs in CNIOLinux got added the const keyword.

Result:

Externs are now immutable


Lukasa (approved):
Thanks @BastianInuk!

PR #1877, opened 2021-06-10
Fix tests for 32-bit platforms, tested for Android armv7

buttaface:
Motivation:

Get the tests passing on 32-bit platforms again.

Modifications:

- Change the way _UInt56.max is initialized.
- Set the proper max capacity for AdaptiveRecvByteBufferAllocator and change the test to match it.
- Add a cmsghdrExample for Android armv7.

Result:

All the same tests pass on Android armv7 as Android AArch64.

Now that I have a working Android armv7 5.4 toolchain again, termux/termux-packages#6941, I noticed that the armv7 tests had regressed since February. These changes got me to parity with AArch64, ie only `BootstrapTest/testClientBindWorksOnSocketsBoundToEitherIPv4OrIPv6Only` fails now, will look into that later.

### DIFF
--- a/Sources/NIO/IntegerTypes.swift
+++ b/Sources/NIO/IntegerTypes.swift
@@ -75,7 +75,8 @@ struct _UInt56 {
 
     static let bitWidth: Int = 56
 
-    static let max: _UInt56 = .init((1 << 56) - 1)
+    private static let initializeUInt64 : UInt64 = (1 << 56) - 1
+    static let max: _UInt56 = .init(initializeUInt64)
     static let min: _UInt56 = .init(0)
 }
 
--- a/Sources/NIO/RecvByteBufferAllocator.swift
+++ b/Sources/NIO/RecvByteBufferAllocator.swift
@@ -55,7 +55,7 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
     private var nextReceiveBufferSize: Int
     private var decreaseNow: Bool
 
-    private static let maximumAllocationSize = 1 << 31
+    private static let maximumAllocationSize = 1 << 30
 
     public init() {
         self.init(minimum: 64, initial: 2048, maximum: 65536)
@@ -93,7 +93,7 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
         // Here we need to be careful with 32-bit systems: if maximum is too large then any shift or multiply will overflow, which
         // we don't want. Instead we check, and clamp to this current value if we overflow.
         let upperBoundCandidate = Int(truncatingIfNeeded: Int64(self.nextReceiveBufferSize) &<< 1)
-        let upperBound = upperBoundCandidate == 0 ? self.nextReceiveBufferSize : upperBoundCandidate
+        let upperBound = upperBoundCandidate <= 0 ? self.nextReceiveBufferSize : upperBoundCandidate
 
         if actualReadBytes <= lowerBound && lowerBound >= self.minimum {
             if self.decreaseNow {
@@ -102,7 +102,8 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
             } else {
                 self.decreaseNow = true
             }
-        } else if actualReadBytes >= self.nextReceiveBufferSize && upperBound <= self.maximum {
+        } else if actualReadBytes >= self.nextReceiveBufferSize && upperBound <= self.maximum &&
+                  self.nextReceiveBufferSize != upperBound {
             self.nextReceiveBufferSize = upperBound
             self.decreaseNow = false
             mayGrow = true
--- a/Tests/NIOTests/RecvByteBufAllocatorTest.swift
+++ b/Tests/NIOTests/RecvByteBufAllocatorTest.swift
@@ -117,8 +117,8 @@ final class AdaptiveRecvByteBufferAllocatorTest: XCTestCase {
         }
 
         let buffer = self.adaptive.buffer(allocator: self.allocator)
-        XCTAssertEqual(buffer.capacity, 1 << 31)
-        XCTAssertEqual(self.adaptive.maximum, 1 << 31)
+        XCTAssertEqual(buffer.capacity, 1 << 30)
+        XCTAssertEqual(self.adaptive.maximum, 1 << 30)
         XCTAssertEqual(self.adaptive.minimum, 0)
     }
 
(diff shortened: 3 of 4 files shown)


Sources/NIO/IntegerTypes.swift
```diff
@@ -75,7 +75,8 @@ struct _UInt56 {
 
     static let bitWidth: Int = 56
 
-    static let max: _UInt56 = .init((1 << 56) - 1)
+    private static let initializeUInt64 : UInt64 = (1 << 56) - 1
+    static let max: _UInt56 = .init(initializeUInt64)
```
buttaface, line 79: I think the problem here is that this numeric literal is always typed as `Int32` on 32-bit platforms, even when I set this with `.init( UInt64 ( (1 << 56) - 1 ) )`, which then overflows the `Int32` and causes a segfault. Defining a separate `UInt64` variable works around this, but perhaps there's a better way to do it?

Sources/NIO/RecvByteBufferAllocator.swift
```diff
@@ -55,7 +55,7 @@ public struct AdaptiveRecvByteBufferAllocator: RecvByteBufferAllocator {
     private var nextReceiveBufferSize: Int
     private var decreaseNow: Bool
 
-    private static let maximumAllocationSize = 1 << 31
+    private static let maximumAllocationSize = 1 << 30
```
buttaface, line 58: The maximum power of 2 that can be stored in an `Int32` is 2^30 not 2^31: `1 << 31` actually sets the sign bit and causes segfaults on Android armv7.

Sources/NIO/IntegerTypes.swift
Lukasa: I think the fact that your proposed workaround didn't work seems like a bug, and that should be filed. This is an appropriate fix, given that your original attempt didn't work.

Tests/NIOTests/RecvByteBufAllocatorTest.swift
```diff
@@ -113,12 +113,12 @@ final class AdaptiveRecvByteBufferAllocatorTest: XCTestCase {
         self.adaptive = AdaptiveRecvByteBufferAllocator(minimum: 0, initial: .max / 2, maximum: .max)
         var mayGrow = true
         while mayGrow {
-            mayGrow = self.adaptive.record(actualReadBytes: .max)
+            mayGrow = self.adaptive.record(actualReadBytes: .max / 2 )
```
Lukasa, line 116: Why was this change necessary?

Tests/NIOTests/SystemTest.swift
```diff
@@ -62,6 +62,21 @@ class SystemTest: XCTestCase {
...
+                                                  0x0D, 0x00, 0x00, 0x00, // Length 13 including header
+                                                  0x00, 0x00, 0x00, 0x00, // IPPROTO_IP
+                                                  0x01, 0x00, 0x00, 0x00, // IP_TOS
+                                                  0x01, 0x00, 0x00, 0x00] // ECT-1 (1 byte)
+    private static let cmsghdr_secondStartPosition = 16
+    private static let cmsghdr_firstDataStart = 12
+    private static let cmsghdr_firstDataCount = 4
+    private static let cmsghdr_secondDataCount = 1
+    private static let cmsghdr_firstType = IP_PKTINFO
+    private static let cmsghdr_secondType = IP_TOS
```
buttaface, line 79: I simply modified the shorter Apple example above to match the linux example below and this worked, didn't actually look at the relevant networking headers or streams.

Sources/NIO/RecvByteBufferAllocator.swift
Davidde94: I don't follow - where is the missing bit?

Davidde94: Scratch that, got it

Tests/NIOTests/RecvByteBufAllocatorTest.swift
buttaface: With this set to `Int32.max`, [this line keeps doubling `nextReceiveBufferSize` until it overflows on 32-bit armv7](https://github.com/apple/swift-nio/blob/main/Sources/NIO/RecvByteBufferAllocator.swift#L95), then [this precondition trips because it thinks`nextReceiveBufferSize` is negative and therefore less than `minimum`](https://github.com/apple/swift-nio/blob/main/Sources/NIO/RecvByteBufferAllocator.swift#L85).

This test change is just working around that overflow, should probably fix that to not overflow on 32-bit platforms instead.

Lukasa: That precondition shouldn't trip. Instead, the line underneath should be rewritten to be ` upperBound <= 0 ?` instead. That will generate the correct logic, which is that we'll clamp at the maximum size.

buttaface: Will do.

buttaface: Made that change to check if `upperBoundCandidate` is negative, but I still had to make this change to the test. Otherwise, the `nextReceiveBufferSize` correctly maxes out at 2^30 and this test loop becomes an infinite loop: I guess because it can never reach `Int32.max`.

Lukasa: Oh, line 105 of `RecvByteBufferAllocator` needs an extra check: it needs to check that `self.nextReceiveBufferSize != upperBound`. Otherwise this isn't growing at all.

Tests/NIOTests/RecvByteBufAllocatorTest.swift
```diff
@@ -136,8 +136,8 @@ final class AdaptiveRecvByteBufferAllocatorTest: XCTestCase {
         }
 
         let adaptive = AdaptiveRecvByteBufferAllocator(minimum: targetValue, initial: targetValue + 1, maximum: targetValue + 2)
-        XCTAssertEqual(adaptive.minimum, 1 << 31)
-        XCTAssertEqual(adaptive.maximum, 1 << 31)
-        XCTAssertEqual(adaptive.initial, 1 << 31)
+        XCTAssertEqual(adaptive.minimum, 1 << 30)
+        XCTAssertEqual(adaptive.maximum, 1 << 30)
+        XCTAssertEqual(adaptive.initial, 1 << 30)
```
buttaface, line 141: I had initially missed these tests that aren't run on armv7, got them after testing this pull on Android AArch64 too.

Lukasa (approved):
Thanks for this cleanup! ✨

Lukasa:

PR #1879, opened 2021-06-14
Add async implementation of EventLoopGroup.shutdownGracefully to _NIOConcurrency

### LINKED ISSUE #1878: Add an async implementation of EventLoopGroup.shutdownGracefully to _NIOConcurrency
EventLoopGroups sometimes need to be shutdown gracefully. This operation does not return a future (as it has no loop to execute on) but instead uses a callback. We can provide a nice async wrapper of this fairly easily by using `withCheckedThrowingContinuation`.

simonjbeaumont:
Adds a new extension to `EventLoopGroup` in `_NIOConcurrency` to provide an asynchronous `shutdownGracefully()`.

### Motivation

Shutting down an event loop was special because it used a completion handler and executed on a `DispatchQueue` because there would be no event loop to use an `EventLoopFuture`.

Now we have the new Swift concurrency features, we can provide an `async` `shutdownGracefully`.

Fixes #1878.

### Modifications

Adds `EventLoopGroup.shutdownGracefully() async` and some code to use it in `NIOAsyncAwaitDemo`.

There's no obvious way to ask an ELG if it has been shutdown but I made the following local-only changes to verify that this performs as expected:

```diff
diff --git a/Sources/NIO/EventLoop.swift b/Sources/NIO/EventLoop.swift
index d18a33dd..972b3b05 100644
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -840 +840 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
-    private enum RunState {
+    public enum RunState {
@@ -852 +852 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
-    private var runState: RunState = .running
+    public var runState: RunState = .running
diff --git a/Sources/NIOAsyncAwaitDemo/main.swift b/Sources/NIOAsyncAwaitDemo/main.swift
index 97a9c0c3..f8b29ace 100644
--- a/Sources/NIOAsyncAwaitDemo/main.swift
+++ b/Sources/NIOAsyncAwaitDemo/main.swift
@@ -60,0 +61 @@ func main() async {
+        print("elg state: \(group.runState)")  // .running
@@ -61,0 +63,7 @@ func main() async {
+        print("elg state: \(group.runState)")  // .closed(nil)
+
+        /// This causes the following error:
+        /// ERROR: Cannot schedule tasks on an EventLoop that has already shut down. This will be upgraded to a forced crash in future SwiftNIO versions.
+        group.next().execute {
+            print("This will not execute.")
+        }
```

### DIFF
--- a/Sources/NIOAsyncAwaitDemo/main.swift
+++ b/Sources/NIOAsyncAwaitDemo/main.swift
@@ -22,13 +22,8 @@ import Dispatch
 
 import _Concurrency
 
-let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-defer {
-    try! group.syncShutdownGracefully()
-}
-
 @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
-func makeHTTPChannel(host: String, port: Int) async throws -> AsyncChannelIO<HTTPRequestHead, NIOHTTPClientResponseFull> {
+func makeHTTPChannel(host: String, port: Int, group: EventLoopGroup) async throws -> AsyncChannelIO<HTTPRequestHead, NIOHTTPClientResponseFull> {
     let channel = try await ClientBootstrap(group: group).connect(host: host, port: port).get()
     try await channel.pipeline.addHTTPClientHandlers().get()
     try await channel.pipeline.addHandler(NIOHTTPClientResponseAggregator(maxContentLength: 1_000_000))
@@ -39,7 +34,9 @@ func makeHTTPChannel(host: String, port: Int) async throws -> AsyncChannelIO<HTT
 @available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *)
 func main() async {
     do {
-        let channel = try await makeHTTPChannel(host: "httpbin.org", port: 80)
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+
+        let channel = try await makeHTTPChannel(host: "httpbin.org", port: 80, group: group)
         print("OK, connected to \(channel)")
 
         print("Sending request 1", terminator: "")
--- a/Sources/NIOAsyncAwaitDemo/main.swift
+++ b/Sources/NIOAsyncAwaitDemo/main.swift
@@ -57,6 +57,9 @@ func main() async {
         print(", response:", String(buffer: response2.body ?? ByteBuffer()))
 
         try await channel.close()
+
+        try await group.shutdownGracefully()
+
         print("all, done")
     } catch {
         print("ERROR: \(error)")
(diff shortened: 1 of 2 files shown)


Sources/NIOAsyncAwaitDemo/main.swift
```diff
@@ -57,6 +57,9 @@ func main() async {
         print(", response:", String(buffer: response2.body ?? ByteBuffer()))
 
         try await channel.close()
+
+        try await group.shutdownGracefully()
```
Lukasa, line 61: This patch isn't quite right as we now shut the loop down twice. Can you move the existing event loop construction out of the top-level scope and into this async function instead?

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
```diff
@@ -38,6 +38,22 @@ extension EventLoopFuture {
...
+    /// Shuts down the event loop gracefully.
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func shutdownGracefully() async throws {
+        return try await withCheckedThrowingContinuation { cont in
+            self.shutdownGracefully { error in
+                if let error = error {
+                    cont.resume(throwing: error)
+                }
+                cont.resume()
```
Lukasa, line 51: Continuations must be resumed _exactly once_: this will fail in the event we hit the error path, as we'll resume twice.

Lukasa (changes requested):
Thanks, looks good! Just two small notes in the diff.

Sources/NIOAsyncAwaitDemo/main.swift
simonjbeaumont: Yes, I noticed this.

I wasn't able able to use the `defer` shutdown with the `async`, and because the comments said it was safe to call shutdown twice, I thought the safest thing to do was to leave the top-level shutdown.

I'll move the group lifecycle into the `async` function, though as it's probably the right thing to do.

Sources/_NIOConcurrency/AsyncAwaitSupport.swift
simonjbeaumont: Of course, oops. I forgot the early-return/switch/else here. Total fail 😅.

Sources/NIOAsyncAwaitDemo/main.swift
Lukasa: Yeah, you can't `await` in a `defer`. In this case, a `do...catch` is probably easier.

Lukasa:

Sources/NIOAsyncAwaitDemo/main.swift
simonjbeaumont: Lemme add the shutdown to the `catch`.

Lukasa (approved):

PR #1885, opened 2021-06-28
[EXAMPLE] Use a newer swift-format

Davidde94:
We use a very old swift format. Now that we've updated to 5.2, it might be a good idea to update swift format too.

We can spend years bike-shedding on the exact things we want. So far we have selected the following rules:
- [None, everything is default]


Lukasa:
NIO does not run a formatter: we're continuing to wait until a formal Swift project formatter is adopted, at which time we'll make a single change to meet its requirements. This is in service of reducing churn.

Davidde94:
Let's have a team discussion first - I'm going to reduce the change size

Lukasa:
We have new formatting logic. Thanks for this @Davidde94!

Issue #1870, opened 2021-06-03
Improve ByteBuffer/ByteBufferView equability


Davidde94:
There are some instances, particularly in tests, where it would be really useful to provide a clean API for comparing a `ByteBuffer` and an expected array of bytes.

Current method 
```swift
XCTAssert(buffer.readableBytesView == ByteBufferView([0x05, 0x00]))
XCTAssert(buffer == ByteBuffer(bytes: [0x05, 0x00]))
```

I think we should provide a shorthand convenience, e.g.
```swift
XCTAssert(buffer == [0x05, 0x00])
XCTAssert(buffer.readableBytesView == [0x05, 0x00])
```

Lukasa:
In this instance, I think it would be substantially more useful to conform `ByteBufferView` to `ExpressibleByArrayLiteral`, which would almost completely solve the problem.

Patil2099:
@Davidde94  I would love to take this up.

Davidde94:
@Patil2099 go for it! Reach out if you need any assistance

mcomisso:
Hi, I wanted to give a look around and find a easy first issue, and noticed this one. 
I don't want to steal anybody's task or time as this was already assigned, but I've added this change on a feature branch in my swift-nio fork to play with the project.

@Patil2099 please let me know if you are still working on it and/or want to compare solutions, the branch is [this](https://github.com/mcomisso/swift-nio/tree/feature/ByteBufferView-ExpressibleByArrayLiteral).

Thanks!

Davidde94:
@Patil2099 @mcomisso any progress? We'll happily take a look at PRs :)

mcomisso:
I've opened a very small PR with the changes, hope you don't mind @Patil2099!

Lukasa:
Resolved by #1886.

PR #1886, opened 2021-06-28
Adds Equatable and ExpressibleByArrayLiteral conformance to ByteBufferView

mcomisso:
Hi!

This PR aims to close issue #1870 , adding `ExpressibleByArrayLiteral` and `Equatable` to `ByteBufferView`. 

This allows writing expressions such as:

```swift
(buffer.readableBytesView == [0x00, 0x01, 0x02])
```

### DIFF
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -191,7 +191,17 @@ extension ByteBuffer {
 extension ByteBufferView: Equatable {
     /// required by `Equatable`
     public static func == (lhs: ByteBufferView, rhs: ByteBufferView) -> Bool {
-        lhs._buffer == rhs._buffer
+
+        guard lhs._range.count == rhs._range.count else {
+            return false 
+        }
+
+        let leftBufferSlice = lhs._buffer.getSlice(at: lhs._range.startIndex, length: lhs._range.count)
+        let rightBufferSlice = rhs._buffer.getSlice(at: rhs._range.startIndex, length: rhs._range.count)
+        
+        return leftBufferSlice == rightBufferSlice
+    }
+}
     }
 }
 
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -202,6 +202,11 @@ extension ByteBufferView: Equatable {
         return leftBufferSlice == rightBufferSlice
     }
 }
+
+extension ByteBufferView: Hashable {
+    /// required by `Hashable`
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(_buffer.getSlice(at: _range.startIndex, length: _range.count))
     }
 }
 
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -209,7 +209,7 @@ extension ByteBufferView: Hashable {
     /// required by `Hashable`
     public func hash(into hasher: inout Hasher) {
         // A well-formed ByteBufferView can never have a range that is out-of-bounds of the backing ByteBuffer.
-        // As a result, this getSlice call can never fail, and we'd like to know it if they do.
+        // As a result, this getSlice call can never fail, and we'd like to know it if it does.
         hasher.combine(self._buffer.getSlice(at: self._range.startIndex, length: self._range.count)!)
     }
 }
--- a/Sources/NIO/ByteBuffer-views.swift
+++ b/Sources/NIO/ByteBuffer-views.swift
@@ -187,3 +187,17 @@ extension ByteBuffer {
         self = view._buffer.getSlice(at: view.startIndex, length: view.count)!
     }
 }
+
+extension ByteBufferView: Equatable {
+    /// required by `Equatable`
+    public static func == (lhs: ByteBufferView, rhs: ByteBufferView) -> Bool {
+        lhs._buffer == rhs._buffer
+    }
+}
+
+extension ByteBufferView: ExpressibleByArrayLiteral {
+    /// required by `ExpressibleByArrayLiteral`
+    public init(arrayLiteral elements: Element...) {
+        self.init(elements)
+    }
+}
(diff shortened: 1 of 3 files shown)


Sources/NIO/ByteBuffer-views.swift
```diff
@@ -187,3 +187,17 @@ extension ByteBuffer {
         self = view._buffer.getSlice(at: view.startIndex, length: view.count)!
     }
 }
+
+extension ByteBufferView: Equatable {
+    /// required by `Equatable`
+    public static func == (lhs: ByteBufferView, rhs: ByteBufferView) -> Bool {
+        lhs._buffer == rhs._buffer
```
Lukasa, line 194: I don't think this implementation is right, but to explain why we need to consider what it must mean for two ByteBufferViews to be equal.

Conceptually the equality constraint for BBV should be the same as Array: two BBVs are equal if their contents are equal. This patch doesn't achieve that because it disregards the `range`s of each BBV.

We likely need to slice out the relevant ranges from each backing buffer and then compare those for equality instead.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -187,3 +187,17 @@ extension ByteBuffer {
         self = view._buffer.getSlice(at: view.startIndex, length: view.count)!
     }
 }
+
+extension ByteBufferView: Equatable {
```
Lukasa, line 191: Let's add `Hashable` as well: it's rare to have something that can usefully be `Equatable` but not `Hashable`.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -3052,3 +3052,14 @@ extension ByteBufferTest {
...
 }
+
+// MARK: - ExpressibleByArrayLiteral init
+extension ByteBufferTest {
+    
+    func testCreateBufferFromArray() {
+        let bufferView: ByteBufferView = [0x00, 0x01, 0x02]
+        let buffer = ByteBuffer(ByteBufferView(bufferView))
+        
+        XCTAssert(buffer.readableBytesView == [0x00, 0x01, 0x02])
```
Lukasa, line 3063: I'd like to see substantially more testing for equatability here. We should aim to have some tests that prove non-equality works appropriately, that work for slices of BBVs, and so-on.

Lukasa (changes requested):
Thanks for this patch! I think the core of it is really sound, but I've left some notes in the diff for ways this can be further improved.

Sources/NIO/ByteBuffer-views.swift
mcomisso: Thanks for the review! I see why this couldn't work, I missed out something pretty foundamental.

I'll likely have some time later today to make the suggested changes and push.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -187,3 +187,32 @@ extension ByteBuffer {
...
+    /// required by `Equatable`
+    public static func == (lhs: ByteBufferView, rhs: ByteBufferView) -> Bool {
+
+        guard lhs._range.count == rhs._range.count else {
+            return false 
+        }
+
+        let leftBufferSlice = lhs._buffer.getSlice(at: lhs._range.startIndex, length: lhs._range.count)
+        let rightBufferSlice = rhs._buffer.getSlice(at: rhs._range.startIndex, length: rhs._range.count)
```
Lukasa, line 200: Great, can we force-unwrap these slices? We need to add a comment to explain why we're doing it, which can say:

> A well-formed ByteBufferView can never have a range that is out-of-bounds of the backing ByteBuffer. As a result, these `getSlice` calls can never fail, and we'd like to know it if they do.

Sources/NIO/ByteBuffer-views.swift
```diff
@@ -187,3 +187,32 @@ extension ByteBuffer {
...
+        
+        return leftBufferSlice == rightBufferSlice
+    }
+}
+
+extension ByteBufferView: Hashable {
+    /// required by `Hashable`
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(_buffer.getSlice(at: _range.startIndex, length: _range.count))
```
Lukasa, line 209: Similarly, we should force-unwrap the call to `getSlice` here, with the same commented reasoning as above.

Also, a minor nit: SwiftNIO style is to explicitly use the `self` keyword to access member fields and functions.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -3052,3 +3052,45 @@ extension ByteBufferTest {
...
+    }
+
+    func testInvalidBufferEqualityWithDifferentContent() {
+        var buffer = self.allocator.buffer(capacity: 4)
+        buffer.writeString("AAAA")
+        
+        let view = ByteBufferView(buffer: buffer, range: 0..<4)
+        let comparisonBuffer: ByteBufferView = [0x41, 0x41, 0x00, 0x00]
+
+        XCTAssertFalse(view == comparisonBuffer)
```
Lukasa, line 3093: Some minor cleanups in your assertions: can you use `XCTAssertEqual` and `XCTAssertNotEqual`? We get substantially better test output from those.

Lukasa (changes requested):
Cool, this is much closer. An extra requirement here, but it would be nice to have a test for hashability as well. This can build on your equatable tests by stuffing everything into a Set and confirming what's left behind.

Sources/NIO/ByteBuffer-views.swift
mcomisso: Thanks for the comment, I've added it as you wrote it as I couldn't come up with anything clearer/briefer than this.

Lukasa (approved):
Great, this LGTM. Thanks @mcomisso!

Lukasa:
@swift-nio-bot please test

Lukasa:

mcomisso:
Thanks for all the guidance, @Lukasa!

Sources/NIO/ByteBuffer-views.swift
Davidde94: +1

Davidde94 (approved):
Great patch, thanks!

PR #1887, opened 2021-06-30
GetaddrinfoResolver: don't block EventLoops

weissi:
Motivation:

GetaddrinfoResolver is blocking the calling thread for anything that
uses a `ClientBootstrap`. Often, `ClientBootstrap`s are used on
EventLoops which obviously creates a problem because you mustn't block
EventLoops.

Modifications:

- Offload the `getaddrinfo` call to a `DispatchQueue`.
- Keep the semantics:
  * only do one concurrent `getaddrinfo` call per SelectableEventLoop
  * prevent thread explosion on `SelectableEventLoop`

Result:

Less EL blocking.


Lukasa (changes requested):
Can you update a couple of the code comments and the license block? The code comment at the top of the file discusses blocking an event loop thread (which this no longer does, though it may block a per-event-loop-thread queue), as does the `init` comment.

Otherwise this seems fine to me.

weissi:
> Can you update a couple of the code comments and the license block? The code comment at the top of the file discusses blocking an event loop thread (which this no longer does, though it may block a per-event-loop-thread queue), as does the `init` comment.
> 
> Otherwise this seems fine to me.

Good catch, thank you! Done.

Lukasa (approved):

Lukasa (approved):

PR #1888, opened 2021-07-01
Feature: Datagram Packet Info

Austinpayne:
Add support for inspecting ingress UDP packet info.

### Motivation:

Currently there is no method to inspect the destination address or interface of an incoming UDP packet. This functionality is useful for certain applications such as an mDNS resolver that should return different IPv6 Link Local addresses depending on the ingress interface. Both Linux and MacOS have the ability to introspect this information using the `IP*_RECVPKTINFO` socket option and the `in*_pktinfo` control message received via `recvmsg`, per RFC 3542 and 2292.

See also:
- https://datatracker.ietf.org/doc/html/rfc3542.html
- https://datatracker.ietf.org/doc/html/rfc2292
- https://github.com/apple/swift-nio/pull/1596

### Modifications:

- Add a  new channel option, `ReceivePacketInfo`, that delivers packet info on `recvmsg`
- Add `NIOPacketInfo` to `AddressedEnvelope.Metadata` for access to packet info in `ChannelHandler`s
- Parse `in*_pktinfo` control messages into `NIOPacketInfo` using existing `ControlMessageParser`

### Result:

Users will now be able to expose packet info (destination address and ingress interface index) using the `ReceivePacketInfo` channel option. For example:

```swift
let server = try! DatagramBootstrap(group: group)
        .channelOption(ChannelOptions.receivePacketInfo, value: true)
            return channel.pipeline.addHandler(MyHandler())
        }
        .bind(host: "::", port: 9999).wait()
...

final class MyHandler: ChannelInboundHandler {
    typealias InboundIn = AddressedEnvelope<ByteBuffer>
    ...

    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        let envelope = self.unwrapInboundIn(data)
        let packetInfo = envelope.metadata?.packetInfo
        print("Destination address: \(packetInfo?.destinationAddress)"
        print("Interface: \(packetInfo?.interfaceIndex)")
        ...
    }
}
```


Austinpayne:
cc: @PeterAdams-A

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -38,9 +38,11 @@ public struct AddressedEnvelope<DataType> {
     public struct Metadata: Hashable {
         /// Details of any congestion state.
         public var ecnState: NIOExplicitCongestionNotificationState
+        public let packetInfo: NIOPacketInfo?
```
Lukasa, line 41: Can you make this `var` instead of `let`?

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -38,9 +38,11 @@ public struct AddressedEnvelope<DataType> {
...
         
-        public init(ecnState: NIOExplicitCongestionNotificationState) {
+        public init(ecnState: NIOExplicitCongestionNotificationState, packetInfo: NIOPacketInfo? = nil) {
```
Lukasa, line 43: Unfortunately we can't add new elements to this initializer: it's not backwards compatible. Can you add a new initializer that takes two arguments instead of one? No need to default the second argument.

Sources/NIO/AddressedEnvelope.swift
```diff
@@ -62,3 +64,8 @@ public enum NIOExplicitCongestionNotificationState: Hashable {
     /// Congestion Experienced.
     case congestionExperienced
 }
+
+public struct NIOPacketInfo: Hashable {
+    public let destinationAddress: SocketAddress
+    public let interfaceIndex: Int
```
Lukasa, line 70: Can these both be made `var`, and can we add a `public init` to this type?

Sources/NIO/BSDSocketAPI.swift
```diff
@@ -308,6 +308,9 @@ extension NIOBSDSocket.Option {
     /// IPv4 and IPv6.
     static let ip_recv_tos: NIOBSDSocket.Option =
             NIOBSDSocket.Option(rawValue: IP_RECVTOS)
+
+    static let ip_recv_pktinfo: NIOBSDSocket.Option =
```
Lukasa, line 312: Can you add a doc comment to this in both places? It's only for us, but I'd like to address what it's for and note why it's `private` (namely, that we want users to use the `ChannelOption`).

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
     }
     
     mutating func receiveMessage(_ controlMessage: UnsafeControlMessage) {
-        if controlMessage.level == IPPROTO_IP && controlMessage.type == ControlMessageParser.ipv4TosType {
+        if controlMessage.level == IPPROTO_IP {
+            receiveIPv4Message(controlMessage)
+        } else if controlMessage.level == IPPROTO_IPV6 {
+            receiveIPv6Message(controlMessage)
```
Lukasa, line 179: NIO style is to use the explicit `self` keyword when accessing member functions or properties.

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
+        }
+    }
+
+    mutating func receiveIPv4Message(_ controlMessage: UnsafeControlMessage) {
```
Lukasa, line 183: This should probably be `private`.

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
+                addr.sin_family = sa_family_t(NIOBSDSocket.AddressFamily.inet.rawValue)
+                addr.sin_port = in_port_t(0).bigEndian
+                addr.sin_addr = info.pointee.ipi_addr
+                self.packetInfo = NIOPacketInfo(destinationAddress: SocketAddress(addr, host: ""), interfaceIndex: Int(info.pointee.ipi_ifindex))
+            }
+
+        }
+    }
+
+    mutating func receiveIPv6Message(_ controlMessage: UnsafeControlMessage) {
```
Lukasa, line 207: This should probably be `private`.

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
+        if controlMessage.type == IPV6_TCLASS {
             if let data = controlMessage.data {
                 let readValue = ControlMessageParser._readCInt(data: data)
                 self.ecnValue = .init(receivedValue: readValue)
             }
+        } else if controlMessage.type == Posix.IPV6_PKTINFO {
+            if let data = controlMessage.data {
+                guard let info = data.bindMemory(to: in6_pktinfo.self).baseAddress else {
```
Lukasa, line 215: This presumably doesn't work on Linux, as we need to use the `CNIOLinux` variant of this structure, right?

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
             if let data = controlMessage.data {
                 assert(data.count == 1)
                 precondition(data.count >= 1)
                 let readValue = CInt(data[0])
                 self.ecnValue = .init(receivedValue: readValue)
             }
-        } else if controlMessage.level == IPPROTO_IPV6 && controlMessage.type == IPV6_TCLASS {
+        } else if controlMessage.type == Posix.IP_PKTINFO {
+            if let data = controlMessage.data {
+                guard let info = data.bindMemory(to: in_pktinfo.self).baseAddress else {
```
Lukasa, line 193: Instead of using `bindMemory` here and below, can we use `load`?

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
+                    return
+                }
+
+                var addr = sockaddr_in()
+                addr.sin_family = sa_family_t(NIOBSDSocket.AddressFamily.inet.rawValue)
+                addr.sin_port = in_port_t(0).bigEndian
```
Lukasa, line 199: No real need to do an endianness swap for port `0`.

Sources/NIO/ControlMessage.swift
```diff
@@ -172,18 +173,57 @@ struct ControlMessageParser {
...
             }
+        } else if controlMessage.type == Posix.IPV6_PKTINFO {
+            if let data = controlMessage.data {
+                guard let info = data.bindMemory(to: in6_pktinfo.self).baseAddress else {
+                    return
+                }
+
+                var addr = sockaddr_in6()
+                addr.sin6_family = sa_family_t(NIOBSDSocket.AddressFamily.inet6.rawValue)
+                addr.sin6_port = in_port_t(0).bigEndian
```
Lukasa, line 221: No real need to do an endianness swap for port 0.

Sources/NIO/ControlMessage.swift
Lukasa: Can we do `load` here as well?

Sources/NIO/SocketChannel.swift
```diff
@@ -451,6 +452,24 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
                 // Explicit congestion notification is only supported for IP
                 throw ChannelError.operationUnsupported
             }
+        case _ as ChannelOptions.Types.ReceivePacketInfo:
+            let valueAsInt: Int32 = value as! Bool ? 1 : 0
```
Lukasa, line 456:
```suggestion
            let valueAsInt: CInt = value as! Bool ? 1 : 0
```

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -791,4 +791,29 @@ final class DatagramChannelTests: XCTestCase {
         XCTAssertNoThrow(try handler.becameUnwritable.futureResult.wait())
         XCTAssertNoThrow(try handler.becameWritable.futureResult.wait())
     }
+
+    func testSetGetPktInfoOption() {
```
Lukasa, line 795: Can we also add a test that validates that the packet info metadata actually appears and is correct?

Lukasa (changes requested):
Thanks for this patch, this generally looks really good! I don't have substantive notes on the patch itself, just a few minor tweaks here and there, and a request for a little more testing. With that said, if we can get all of that cleaned up I think this is good to go!

Sources/NIO/ControlMessage.swift
Austinpayne: It does work on Linux. I added a `typealias` to the `CNIOLinux` variant in `System.swift`: https://github.com/apple/swift-nio/pull/1888/files#diff-3f800fb8f45dc37abc935011731dc02fe9bb2eb245b7f943be8a8214af930bc5R29 I'm open to other ways to solve this but this struct is behind a macro that appears to have the same issue as `_GNU_SOURCE`: https://bugs.swift.org/browse/SR-6772

Austinpayne:
Thanks for the excellent review @Lukasa! I've made the changes you suggested. Please let me know if there is anything else.

Sources/NIO/ControlMessage.swift
Lukasa: Great, that's what I missed. The `typealias` is fine.

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -667,8 +667,13 @@ final class DatagramChannelTests: XCTestCase {
         } ())
     }
     
-    private func testEcnReceive(address: String, vectorRead: Bool, vectorSend: Bool) {
+    private func testEcnReceive(address: String, vectorRead: Bool, vectorSend: Bool, receivePacketInfo: Bool = false) {
```
Lukasa, line 670: Can I suggest renaming this method if we're going to overload it?

Lukasa:

Tests/NIOTests/DatagramChannelTests.swift
```diff
@@ -667,8 +667,13 @@ final class DatagramChannelTests: XCTestCase {
...
         XCTAssertNoThrow(try {
+            // Fake sending packet to self on the loopback interface if requested
+            let fakeDestinationAddress = try SocketAddress(ipAddress: address, port: 0)
+            let fakeIngressIfaceIndex = 1
```
Lukasa, line 674: I think this risks breaking in CI: as the write path doesn't do anything, and as these are real channels, we actually need to know the real interface index. You can get this out of [`System.enumerateDevices`](https://apple.github.io/swift-nio/docs/current/NIO/Enums/System.html#/s:3NIO6SystemO16enumerateDevicesSayAA16NIONetworkDeviceVGyKFZ), which will let you search for the right one. A simple `filter` should find the interface we're using.

Lukasa (changes requested):
Cool, only some tweaks in the tests now.

Austinpayne:
Thanks again @Lukasa. I also updated the Linux tests, which I forgot to do in the previous commits.

Lukasa (approved):

Sources/NIO/System.swift
```diff
@@ -204,6 +205,10 @@ internal enum Posix {
     static let IPTOS_ECN_ECT0: CInt = CNIODarwin_IPTOS_ECN_ECT0
     static let IPTOS_ECN_ECT1: CInt = CNIODarwin_IPTOS_ECN_ECT1
     static let IPTOS_ECN_CE: CInt = CNIODarwin_IPTOS_ECN_CE
+    static let IP_RECVPKTINFO: CInt = CNIODarwin.IP_RECVPKTINFO
```
JacobMao, line 208: Hi @Lukasa , @Austinpayne :

Sorry for adding comment in this merged PR.

Background:
I am using `Bazel` to build `NIOPosix` as static framework, but I got error - `"module 'CNIODarwin' has no member named 'IP_RECVPKTINFO'"`.
I didn't find where CNIODarwin exposes `IP_RECVPKTINFO`, and if I change to `Darwin.IP_RECVPKTINFO`, building is ok.
so why did you use `CNIODarwin.IP_RECVPKTINFO` at here?

Thanks so much!

Lukasa: `CNIODarwin` should expose it by way of an include of a system header file. Can you clarify what system you're building on, with what SDK and so on?

JacobMao: Thanks for your reply.

I am an iOS developer, so I build it on mac os with iOS sdk 15, and `Command Line Tools` is `Xcode 13.1 (13A1030d)`

BTW, if I use Xcode to create a new project and add `swift-nio` in `Swift Package` way, everything is ok. But my company project is using `Bazel` to build these frameworks

Lukasa: This seems like it's a Bazel issue then. It may be worth following up with the maintainers of the Swift builders for Bazel to find out what they're doing differently than Xcode.

JacobMao: got it.

thanks for your comment

PR #1889, opened 2021-07-05
ByteBuffer: unconditionally use zero-copy for JSON Decoding

### LINKED ISSUE #1511: [Performance] Decoding Decodables from ByteBuffer is not zero-copy.
### Expected behavior

The `ByteBuffer`'s `getJSONDecodable` ought to be zero-copy.

### Actual behavior

The `ByteBuffer`'s `getJSONDecodable` doesn't use zero-copy flags to `getData`.

The `getData`'s `byteTransferStrategy:` option at https://github.com/apple/swift-nio/blob/eb140159386d19fa2c9439af24286982a52fb0b3/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift#L109 could have been set to `.automatic`, that copies everything less than 256kb.

### Steps to reproduce

Observe https://github.com/apple/swift-nio/blob/eb140159386d19fa2c9439af24286982a52fb0b3/Sources/NIOFoundationCompat/Codable%2BByteBuffer.swift#L31

pejato:
Changes ByteBuffer decoding behavior to always use zero-copy byte transfer strategy.

### Motivation:

When decoding decodables from Byte Buffer, it is possible that we copy bytes (for small buffers). Per discussion in #1511 , it is better to unconditionally never copy.

### Modifications:

Changes `ByteBuffer.getJSONDecodable` to always use `.noCopy` `ByteTransferStrategy`

### Result:

Fewer unnecessary copies of small `ByteBuffer`'s. Resolves #1511

### DIFF
--- a/Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
+++ b/Sources/NIOFoundationCompat/Codable+ByteBuffer.swift
@@ -28,7 +28,7 @@ extension ByteBuffer {
     public func getJSONDecodable<T: Decodable>(_ type: T.Type,
                                                decoder: JSONDecoder = JSONDecoder(),
                                                at index: Int, length: Int) throws -> T? {
-        guard let data = self.getData(at: index, length: length) else {
+        guard let data = self.getData(at: index, length: length, byteTransferStrategy: .noCopy) else {
             return nil
         }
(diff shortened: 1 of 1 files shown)


Lukasa:

Lukasa (approved):
Thanks, this LGTM!

Issue #1892, opened 2021-07-09
This repo no longer builds with the latest official trunk snapshots for linux: an upstream stdlib change probably broke building NIO


buttaface:
### Expected behavior
`swift build` builds NIO.

### Actual behavior
Build fails with this error:
```
> ../src/swift-DEVELOPMENT-SNAPSHOT-2021-07-07-a-centos8/usr/bin/swift build
/home/butta/swift-nio/Sources/_NIO1APIShims/NIO1APIShims.swift:49:1: error: type 'StaticString' does not conform to protocol 'Collection'
extension StaticString: Collection {
^
/home/butta/swift-nio/Sources/_NIO1APIShims/NIO1APIShims.swift:49:1: error: unavailable subscript 'subscript(_:)' was used to satisfy a requirement of protocol 'Collection'
extension StaticString: Collection {
^
Swift.Collection:3:12: note: 'subscript(_:)' declared here
    public subscript(bounds: Range<Self.Index>) -> Self.SubSequence { get }
           ^
Swift.Collection:12:5: note: requirement 'subscript(_:)' declared here
    subscript(bounds: Range<Self.Index>) -> Self.SubSequence { get }
    ^
```
Most likely related to apple/swift#38161?
### SwiftNIO version/commit hash

latest trunk ffa6a87

### Swift & OS version (output of `swift --version && uname -a`)

That is one of the latest official trunk snapshot builds, July 7, running on Fedora Core 33. I can also reproduce with the July 6 trunk source snapshot toolchain built natively on Android AArch64, which is where I first hit this.

Lukasa:
I think this will need a `bugs.swift.org` report. Are you open to filing one?

buttaface:
Will do it now.

buttaface:
Filed SR-14905.

glessard:
This error is triggered because the type's Collection conformance relied on a bad `Range<Index>` subscript. Calling that subscript in your code would have resulted in an accidental infinite recursion [SR-14848](https://bugs.swift.org/browse/SR-14848). The best mitigation in this case is to add the now-missing subscript.

Amusingly this build failure is in a long-deprecated conformance.

glessard:
I'm testing the following fix:
```swift
    @available(*, deprecated, message: "don't use the StaticString: Collection extension please")
    public subscript(bounds: Range<Int>) -> SubSequence {
        precondition(0 <= bounds.lowerBound &&
[...]

PR #1893, opened 2021-07-09
add a missing `subscript(bounds:)` implementation

glessard:
An older version of the standard library supplied a bogus implementation by error.  
Calling *that* would have been a crashing bug.

### Motivation:

Fix https://github.com/apple/swift-nio/issues/1892,  
which was triggered by https://github.com/apple/swift/pull/38161,  
which is a fix for https://bugs.swift.org/browse/SR-14848.

### Modifications:

Add the missing `subscript(bounds:)` implementation. It should always have been there.

### Result:

Package should build with the July 8 2021 snapshot of Swift.

### DIFF
--- a/Sources/_NIO1APIShims/NIO1APIShims.swift
+++ b/Sources/_NIO1APIShims/NIO1APIShims.swift
@@ -67,6 +67,15 @@ extension StaticString: Collection {
         precondition(position < self.utf8CodeUnitCount, "index \(position) out of bounds")
         return self.utf8Start.advanced(by: position).pointee
     }
+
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Thanks for the improvement!

Lukasa:

glessard:
An improvement that is deprecated-on-arrival. 🤷🏽

glessard:
This also triggered a failure in the compatibility suite (sorry for failing to notice). Could you update after merge?

Lukasa:
I believe that the Swift team are in charge of updating that.

PR #1895, opened 2021-07-12
Extract basic data structures to NIOCore.

Lukasa:
Motivation:

Currently the "core" NIO module (currently called NIO) contains two
fairly distinct kinds of things. The first is the core abstractions and
data structures: the things that all NIO programs must necessarily deal
with. This includes our base abstraction protocols (e.g. `EventLoop`,
`Channel`, and `ChannelHandler`), as well as our base data types (e.g.
`ByteBuffer`, `SocketAddress`, `EventLoopFuture`).

The second thing the core NIO module contains is the default POSIX
sockets event-loop and channels. These are the mainline production loops
for NIO programs, and are supported in the various BSD-like OSes to
varying extents. These can be thought of as a baseline implementation of
the core abstraction protocols that form the first part of NIO.

These two fairly distinct use-cases have no particular reason to be
glued together, and in fact in gluing them together we have done a bit
of a disservice to our adopters. The biggest issue is that it has become
impossible to write a NIO program or library that does not bring along
the POSIX layer of NIO. This has made porting NIO to other platforms
difficult (Windows has been painful, and Webassembly is coming down the
road as well), and has also bloated the size of iOS/macOS applications
that use NIO via NIOTransportServices, as all of these applications must
bring along an event loop and several Channels they will not use.

To that end, we plan to begin a scope of work to split the core NIO
module in two. Specifically, we intend to extract the baseline protocols
and data types into a new module, called `NIOCore`, which will be able
to be the baseline NIO module.

The end goal is that libraries that implement protocols on top of
SwiftNIO should be able to depend solely on `NIOCore`, not `NIO`. This
will allow NIO applications to bring along only the event loops they
want or need, without needing to pay for the POSIX sockets
implementation. This should also make porting easier, as there will be
no pressure to bring the existing `NIO` module to new platforms if it
does not fit well.

Modifications:

This is the first in a series of patches, which extracts some of the
low-hanging fruit. In particular, it extracts:

- `ByteBuffer` (and associated types)
- `RecvByteBufferAllocator`
- `CircularBuffer`

This also adds in the backwards compatibility shim: the `NIO` package
performs an exported import of `NIOCore`. This will allow this change to
avoid breaking API.

However, to make a clean break we intend not to release a new NIO minor
version until we have completed our extraction. If we fail to do this it
won't be the end of the world, we can release subsequent minor versions
that continue to extract new types. However, as much as possible should
go at once.

This patch also duplicates a few methods and files. At the end of the
patch series we need to reconcile this duplication. In many cases the
duplication will no longer be necessary, but in some cases we may have
necessary duplicates.

Result:

The first step along the road to better platform citizenship is taken.

cc @0xTim @gwynne @tachyonics @compnerd who are all likely to care about this work in one way or another


Lukasa:
The API breakage checker is naturally not thrilled about this outcome, but in my testing the API breakage is not actually real: it just doesn't know how to deal with `@_exported` imports.

PeterAdams-A:
Seems reasonable - does it change performance at all?

Lukasa:

Lukasa:
So far at least it doesn't regress allocs: let's see about perf.

Lukasa:
The TL;DR here is: no major performance impact. The things that got _faster_ seem silly.

Tests/NIOTests/ByteBufferTest.swift
```diff
@@ -14,7 +14,7 @@
 
 import struct Foundation.Data
 import XCTest
-@testable import NIO
+@testable import NIOCore
```
glbrntt, line 17: Should this file (and others) move to a `NIOCoreTests`?

glbrntt (commented):
> However, to make a clean break we intend not to release a new NIO minor
> version until we have completed our extraction. If we fail to do this it
> won't be the end of the world, we can release subsequent minor versions
> that continue to extract new types. However, as much as possible should
> go at once.

Is it worth merging these patches into a new branch so we can keep them together?

Tests/NIOTests/ByteBufferTest.swift
Lukasa: Some can move easily (like this one), but plenty cannot because they have awkward cross-dependencies. My plan for now is to ignore the test files and let them blur together, and then to break them apart later on.

Lukasa:
> Is it worth merging these patches into a new branch so we can keep them together?

I'm not sure. The biggest annoyance there is that we don't get good CI on them (the CI runs against PRs on the `main` branch only), so reviewing them is harder.

glbrntt:
> > Is it worth merging these patches into a new branch so we can keep them together?
> 
> I'm not sure. The biggest annoyance there is that we don't get good CI on them (the CI runs against PRs on the `main` branch only), so reviewing them is harder.

Good point, the lack of CI is a bit awkward.

glbrntt (approved):

PR #1897, opened 2021-07-13
remove NIO1 API shims

weissi:
Motivation:

The lastest NIO versions require Swift 5.2+ to compile. Given that
hopefully nobody ever created a NIO1 application for Swift 5.2+, I'd say
it's about time to remove the NIO1 API shims (and related docs).

Modification:

- remove the NIO1 API shims
- remove the migration docs

Result:

Less code.


Lukasa (approved):
Pour one out for our fallen comrades.

PR #1898, opened 2021-07-14
Extract EventLoop protocols and EventLoopFuture to Core

Lukasa:
Motivation:

The next step on extracting the base abstractions is to pull out
EventLoop, EventLoopGroup and related types, and also EventLoopPromise
and EventLoopFuture. These types are fundamental to the API.

Modifications:

- Extract EventLoop, EventLoopGroup, and related types.
- Extract EventLoopPromise and EventLoopFuture.
- Add new @testable imports.
- Split MultiThreadedEventLoopGroup into new file, appropriately named.

Result:

Better separation.


Sources/NIOCore/EventLoop.swift
```diff
@@ -777,13 +753,14 @@ extension EventLoopGroup {
...
 
     public func syncShutdownGracefully() throws {
-        if let eventLoop = MultiThreadedEventLoopGroup.currentEventLoop {
-            preconditionFailure("""
-            BUG DETECTED: syncShutdownGracefully() must not be called when on an EventLoop.
-            Calling syncShutdownGracefully() on any EventLoop can lead to deadlocks.
-            Current eventLoop: \(eventLoop)
-            """)
-        }
+        // FIXME: Add new API surface to reimplement this.
```
glbrntt, line 756: Are the fixes imminent or do we need to create issues for them?

Lukasa: The fixes should be part of this patch series, yeah, as otherwise we're removing functionality in many of these cases.

Lukasa: #1899 #1900

fabianfett: I guess we can remove this FIXME's now.

Lukasa: Nope, this _particular_ FIXME is actually still open. On me to address.

Sources/NIO/MultiThreadedEventLoopGroup.swift
```diff
@@ -0,0 +1,345 @@
...
+        // If the `runState` was not `running` when `shutdownGracefully` was called,
+        // the shutdown has already been initiated and we have to return here.
+        guard wasRunning else {
+            return
+        }
+
+        var result: Result<Void, Error> = .success(())
+
+        for loop in self.eventLoops {
+            g.enter()
```
Davidde94, line 231: I know this is old code, but it feels odd that `g` is declared half a mile above

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -179,7 +179,11 @@ Further information:
         self.tasksCopy.reserveCapacity(4096)
         self.canBeShutdownIndividually = canBeShutdownIndividually
         // note: We are creating a reference cycle here that we'll break when shutting the SelectableEventLoop down.
-        self._succeededVoidFuture = EventLoopFuture(eventLoop: self, value: (), file: "n/a", line: 0)
+        // note: We have to create the promise and complete it because otherwise we'll hit a loop in `makeSucceededFuture`. This is
```
Davidde94, line 182: Is this worth filing an issue about?

Davidde94 (approved):

Sources/NIO/MultiThreadedEventLoopGroup.swift
Lukasa: Yeah, it's not great. For now I'm proposing we leave it.

Sources/NIO/SelectableEventLoop.swift
Lukasa: I don't think so: this is an unfortunate wrinkle in the way the fast-path for completed void futures works. It's kinda unavoidable now that we've split them apart, and there isn't a good way to fix this.

Sources/NIO/SelectableEventLoop.swift
```diff
@@ -616,7 +620,12 @@ Further information:
...
         guard self.inEventLoop, let voidFuture = self._succeededVoidFuture else {
-            return EventLoopFuture(eventLoop: self, value: (), file: "n/a", line: 0)
+            // We have to create the promise and complete it because otherwise we'll hit a loop in `makeSucceededFuture`. This is
+            // fairly dumb, but it's the only option we have. This one can only happen after the loop is shut down,
+            // which is an acceptable compromise.
+            let voidPromise = self.makePromise(of: Void.self)
+            voidPromise.succeed(())
+            return voidPromise.futureResult
         }
         return voidFuture
```
fabianfett, line 630: Why, do we need to check, if we are on the EL, the method is called on?

Lukasa: `_succeededVoidFuture` creates a reference cycle which can leak an event loop (it's stored on `self` and holds a reference to its owning `EventLoop`, which is `self`). As a result, it is `nil`'d when we exit the event loop. To avoid needing a lock to read it, we only use this fast path if we're on the loop, where we can safely read it synchronously.

fabianfett: Ah! No the comment makes a little more sense, but I think we should update the code comment to talk about a reference cycle instead of a loop. Also this seems to be out of date:

> This one can only happen after the loop is shut down, which is an acceptable compromise.

Lukasa: There is a reference to the loop in the assignment to the variable in `init`.

fabianfett (approved):
🚢

glbrntt (approved):

PR #1899, opened 2021-07-14
Enable all EventLoops to track promise creation.

Lukasa:
Motivation:

EventLoopFuture is capable of tracking where it was allocated from in
debug mode, to enable high-fidelity tracking of leaked promises.
Currently this behaviour is limited to SelectableEventLoop, but there is
no good reason to tie it so specifically to that loop.

Modifications:

- Extend `EventLoop` to allow all event loops to track promise
  allocation and destruction.
- Add helpers to `EmbeddedEventLoop` to let it participate.
- Add new type to express a future ID.

Result:

Wider support for diagnostics.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -283,6 +283,30 @@ public protocol EventLoop: EventLoopGroup {
     /// This method should not be called by users directly, it should only be implemented by `EventLoop` implementers that
     /// need to customise the behaviour.
     func _preconditionSafeToWait(file: StaticString, line: UInt)
+
+    /// Debug hook: track a promise creation and its location.
+    ///
+    /// This debug hook is called by EventLoopFutures and EventLoopPromises when they are created, and tracks the location
+    /// of their creation. It combines with `_promiseCompleted` to provide high-fidelity diagnostics for debugging leaked
+    /// promises.
+    ///
+    /// In release mode, this function will never be called.
+    ///
+    /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
+    /// however, both of them should be implemented.
+    func _promiseCreated(futureIdentifier: _NIOEventLoopFutureIdentifier, file: StaticString, line: UInt)
+
+    /// Debug hook: complete a specific promise and return its creation location.
+    ///
+    /// This debug hook is called by EventLoopFutures and EventLoopPromises when they are deinited, and removes the data from
+    /// the promise tracking map and, if available, provides that data as its return value. It combines with `_promiseCreated`
+    /// to provide high-fidelity diagnostics for debugging leaked promises.
+    ///
+    /// In release mode, this function will never be called.
+    ///
+    /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
+    /// however, both of them should be implemented.
+    func _promiseCompleted(futureIdentifier: _NIOEventLoopFutureIdentifier) -> (file: StaticString, line: UInt)?
 }
 
 extension EventLoop {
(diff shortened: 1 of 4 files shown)


Lukasa:
I believe the API breakage checker is wrong about the API break (the new protocol requirements are defaulted), but please validate that carefully.

Sources/NIO/EventLoop.swift
```diff
@@ -273,13 +273,47 @@ public protocol EventLoop: EventLoopGroup {
     /// Contrary to `makeSucceededFuture`, `makeSucceededVoidFuture` is a customization point for `EventLoop`s which
     /// allows `EventLoop`s to cache a pre-succeded `Void` future to prevent superfluous allocations.
     func makeSucceededVoidFuture() -> EventLoopFuture<Void>
+
+    /// Debug hook: track a promise creation and its location.
```
glbrntt, line 277: Should we just extract these two functions into another internal-underscored `protocol` and have `EventLoop` implementations optionally support it? It'd be less visible to someone holding an `EventLoop` and we don't have to add the default implementations.

Sources/NIO/EventLoopFuture.swift
```diff
@@ -1574,3 +1571,27 @@ extension EventLoopFuture {
...
+/// tracking data structure.
+public struct _NIOEventLoopFutureIdentifier: Hashable {
+    private var opaqueID: UInt
+
+    @usableFromInline
+    internal init<T>(_ future: EventLoopFuture<T>) {
+        self.opaqueID = _NIOEventLoopFutureIdentifier.obfuscatePointerValue(future: future)
+    }
+
+
```
glbrntt, line 1589:
```suggestion

```

Sources/NIO/EventLoop.swift
```diff
@@ -273,13 +273,47 @@ public protocol EventLoop: EventLoopGroup {
...
+    ///
+    /// This debug hook is called by EventLoopFutures and EventLoopPromises when they are created, and tracks the location
+    /// of their creation. It combines with `_promiseCompleted` to provide high-fidelity diagnostics for debugging leaked
+    /// promises.
+    ///
+    /// In release mode, this function will never be called.
+    ///
+    /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
+    /// however, both of them should be implemented.
+    func _promiseCreated(futureIdentifier: _NIOEventLoopFutureIdentifier, file: StaticString, line: UInt)
```
glbrntt, line 287: Maybe a silly question, but why does this need to be framed in terms of a future identifier rather than just a future and having the `EventLoop` implementation deal with the identity of that future?

Sources/NIO/EventLoop.swift
Lukasa: I think I'd rather not do it: things that can be safely defaulted without any damage should be just fine. Users are highly unlikely to call this method directly, and even if they did it's not like anything bad would happen. Underscored methods don't appear in the autocomplete unless you deliberately attempt to find them, so they don't impose any particular cognitive overhead on users. They also save us from needing runtime type checks to validate a given Event Loop's conformance.

Sources/NIO/EventLoop.swift
Lukasa: The short reason is that I wanted to persist the requirement that led to `SelectableEventLoop` using the somewhat odd `obfuscatePointerValue` function. If we pass the future to the callback directly, users will likely naively store this in a dictionary using `ObjectIdentifier`. This will have the effect of causing the Xcode memory graph debugger to think that the future is being retained by the dictionary. It won't be: object identifiers don't retain objects, but the memory graph debugger doesn't know that.

This way we can arrange for an identifier that is stably defined in terms of the object identifier but is not confusable for a pointer to a given object. While we're here, it also discourages implementers from being "clever" and using this to observe promise creation and destruction for any reason other than this one.

Sources/NIO/EventLoop.swift
glbrntt: True, it just feels slightly off to me to add internal debug hooks here. It's not a big deal though.

glbrntt (approved):

Sources/NIO/EventLoop.swift
Lukasa: I mean, they're not entirely internal: we're happy for third-party implementers of event loops to implement them if they want to. For example, NIOTS will.

fabianfett (approved):

PR #1900, opened 2021-07-14
Avoid EventLoopFuture.wait knowing about specific implementations.

Lukasa:
Motivation:

As we break apart NIO, it's a bit awkward that EventLoopFuture.wait
knows about two specific EventLoop implementations. It does this in
order to detect deadlocks, so we should try to preserve the capability
as best as possible.

Modifications:

- Add EventLoop._preconditionSafeToWait, which can reproduce much of the
  logic currently in EventLoopFuture.wait().
- Provide implementations for SelectableEventLoop and EmbeddedEventLoop.

Result:

Partial reproduction of the current behaviour of EventLoopFuture.wait().
Unfortunately this cannot completely reimplement this behaviour, as we
cannot safely call MultiThreadedEventLoopGroup.currentEventLoop for all
event loops. This is as close as we can safely get without breaking the
abstraction boundary, sadly.

### DIFF
--- a/Sources/NIO/EventLoop.swift
+++ b/Sources/NIO/EventLoop.swift
@@ -279,6 +279,9 @@ public protocol EventLoop: EventLoopGroup {
     /// This method is a debugging hook that can be used to override the behaviour of `EventLoopFuture.wait()` when called.
     /// By default this simply becomes `preconditionNotInEventLoop`, but some `EventLoop`s are capable of more exhaustive
     /// checking and can validate that the wait is not occuring on an entire `EventLoopGroup`, or even more broadly.
+    ///
+    /// This method should not be called by users directly, it should only be implemented by `EventLoop` implementers that
+    /// need to customise the behaviour.
     func _preconditionSafeToWait(file: StaticString, line: UInt)
 }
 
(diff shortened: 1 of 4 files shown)


Sources/NIO/EventLoop.swift
```diff
@@ -273,13 +273,24 @@ public protocol EventLoop: EventLoopGroup {
     /// Contrary to `makeSucceededFuture`, `makeSucceededVoidFuture` is a customization point for `EventLoop`s which
     /// allows `EventLoop`s to cache a pre-succeded `Void` future to prevent superfluous allocations.
     func makeSucceededVoidFuture() -> EventLoopFuture<Void>
+
+    /// Must crash if it is not safe to call `wait()` on an `EventLoopFuture`.
+    ///
+    /// This method is a debugging hook that can be used to override the behaviour of `EventLoopFuture.wait()` when called.
+    /// By default this simply becomes `preconditionNotInEventLoop`, but some `EventLoop`s are capable of more exhaustive
+    /// checking and can validate that the wait is not occuring on an entire `EventLoopGroup`, or even more broadly.
+    func _preconditionSafeToWait(file: StaticString, line: UInt)
```
fabianfett, line 282: Why do we want to underscore this method?

Lukasa: Because we don't require implementing it and we don't want anyone to call it directly.

glbrntt: Can we document that we don't want anyone to call it directly?

Lukasa: Done.

glbrntt (approved):

fabianfett (approved):

PR #1902, opened 2021-07-19
ByteBuffer.writeLengthPrefixed

dnadoba:
A convenience method to write a message and prepend its length in front of it.

### Motivation:
Many protocols use a length field to split a byte stream into multiple separate messages. `swift-nio-extras` supports these kind of protocols through `LengthFieldPrepender` and friends. However, sometimes a channel handler is not the right tool, especially if the a message itself contains multiple dynamic length fields. One example is [RSocket Routing Metadata Extensions](https://github.com/rsocket/rsocket/blob/master/Extensions/Routing.md). 

A `ByteBuffer` is probably a better tool to create this message. This can be tricky to get right if strings are part of the message. A naive solution like the following has two issues:
```swift
func add(tag: String, to buffer: inout ByteBuffer) {
    let length = tag.count
    buffer.writeInteger(UInt8(length))
    buffer.writeString(tag)
}
```
The `count` property in Swift is different than in most other languages. It does **not** return the number of UTF-8 code points and can result in incorrect behavior if multi-byte characters are used (e.g. Emojis). Furthermore, the size of the string could be larger than `UInt8` can represent.
### Modifications:

This PR introduces a new method on `ByteBuffer` called `writeLengthPrefix(endianness:as:writeMessage:)`. It solves both problems mentioned above. It first reserves space for the length prefix and delegates the actually writing of the message to `writeMessage`. Afterwards, it counts how many bytes were written during the call to `writeMessage` and then sets it on the previously reserved space in front of the actually message.

### Result:
We can now write arbitrary content to a buffer and prefix it with the number of bytes written.
```swift
func add(tag: String, to buffer: inout ByteBuffer) {
    buffer.writeLengthPrefix(as: UInt8.self) { buffer in
        buffer.writeString(tag)
    }
}
```

### Alternatives
We could also name the method `prependLengthToMessage(endianness:as:writeMessage:)` but this would be against the general rule that all methods which increment the `writtenBytes` property should be prefixed with `write`.


Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,55 @@
...
+
+extension ByteBuffer {
+    /// Prefixes a message written by `writeMessage` with the number of bytes written as an `Integer`.
+    /// - Throws: If the number of bytes written during `writeMessage` can not be exactly represented as the given `Integer` i.e. if the number of bytes written is greater than `Integer.max`
+    /// - Returns: Number of total bytes written
+    @discardableResult
+    @inlinable
+    public mutating func writeLengthPrefix<Integer: FixedWidthInteger>(
+        endianness: Endianness = .big,
+        as integer: Integer.Type = Integer.self,
```
glbrntt, line 34: I don't think we want the default here, the caller should explicitly provide the length type.

Lukasa (changes requested):

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,55 @@
...
+        message writeMessage: (inout ByteBuffer) throws -> ()
+    ) throws -> Int {
```
glbrntt, line 36: nit: the normal NIO style is to always use explicit `self`, we're missing a few in the body of this.

Tests/NIOTests/ByteBufferLengthPrefixTests+XCTest.swift
```diff
@@ -0,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 0 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5: Something's gone wrong here with the year... was it generated like this?

glbrntt (changes requested):

Tests/NIOTests/ByteBufferLengthPrefixTests+XCTest.swift
Lukasa: #1903.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+extension ByteBuffer {
+    /// Prefixes a message written by `writeMessage` with the number of bytes written as an `Integer`.
+    /// - Throws: If the number of bytes written during `writeMessage` can not be exactly represented as the given `Integer` i.e. if the number of bytes written is greater than `Integer.max`
+    /// - Returns: Number of total bytes written
+    @discardableResult
+    @inlinable
+    public mutating func writeLengthPrefix<Integer: FixedWidthInteger>(
```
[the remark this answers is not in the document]
dnadoba: `set` method added and renamed `write` method to `writeLengthPrefixed`.

I'm not sure about the `get`/`read` methods. They are fairly trivial to implement and there are no pitfalls that I am aware of at least. However, if you think they add value I can still add them.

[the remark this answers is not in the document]
Lukasa: I think the pitfall is that it's easy to not enforce the read-length: that is, to issue a read and not police that a) the entire body is consumed, and b) that nothing is overconsumed.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,55 @@
...
+        as integer: Integer.Type = Integer.self,
+        message writeMessage: (inout ByteBuffer) throws -> ()
+    ) throws -> Int {
+        let lengthPrefixIndex = writerIndex
+        // Write a zero as a placeholder which will later be overwritten by the actual number of bytes written
+        var totalBytesWritten = writeInteger(.zero, endianness: endianness, as: Integer.self)
+        
+        let messageStartIndex = writerIndex
+        try writeMessage(&self)
+        let messageEndIndex = writerIndex
```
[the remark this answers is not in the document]
dnadoba: I have changed the `writeMessage` function to return a number of bytes written for now. The `set` version of this method actually requires it, otherwise we do not know how many bytes were actually written if only `set` methods are used during `writeMessage`.

However, I find it more convenient to not return the number of bytes written because it can become rather difficult to provide this number if you use any kind of control flow during `writeMessage`. I would like that `setLengthPrefixed` requires a number of bytes written to be returned but `writeLengthPrefixed ` should not require it and compute the bytes written through the indices. WDYT?

As an alternative, we could also require the `writeMessage` function of `writeLengthPrefixed` to return the number of bytes written and add an assert/precondition that this number is actually correct but we should definitely document this behavior.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: Agree, changed 👍

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: Done ✅

Tests/NIOTests/ByteBufferLengthPrefixTests+XCTest.swift
dnadoba: Good Catch, thanks. Updated to 2021

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
Lukasa: I think I'd rather keep them the same. In principle it shouldn't be too challenging to do this with control flow: a number of NIO projects including NIOSSH effectively include this function and manage their control flow fairly straightforwardly. The pattern tends to be:

```swift
buffer.writeLengthPrefixed(as: UInt32.self) { buffer in
    var written = 0
    written += buffer.writeString(self.someString)
    if self.foo {
       written += buffer.writeInteger(self.bar)
   else {
       written += buffer.writeInteger(self.baz)
    }
    return written
}

```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,64 @@
...
+        
+        return try self.setLengthPrefixed(at: lengthPrefixIndex, endianness: endianness, as: Integer.self, message: writeMessage)
+    }
+    
+    /// Prefixes a message with the number returned by `setMessage` as an `Integer`.
+    /// - Throws: If the number of bytes returned by `setMessage` can not be exactly represented as the given `Integer` i.e. if the number is greater than `Integer.max`
+    /// - Returns: Number of total bytes written i.e. size of the integer plus the number returned by `setMessage`
+    @discardableResult
+    @inlinable
+    public mutating func setLengthPrefixed<Integer: FixedWidthInteger>(
```
glbrntt, line 49: The semantics of this don't seem quite right to me, imagine:

```swift
var buffer = ByteBuffer()
buffer.writeBytes([0, 0, 0, 0])

buffer.setLengthPrefixed(at: self.readerIndex, as: UInt8.self) { buffer in
  buffer.writeBytes([1, 2, 3, 4])
}
```

`buffer` would be `[4, 0, 0, 0, 1, 2, 3, 4]` which would surprise people, I think.

I wonder if `setMessage` should also be given the index after the length bytes, i.e. the index where it should start setting bytes?

Lukasa: Oh good note: I seem to recall having tripped over this elsewhere. I think the `set` variant is probably not sensible on length-prefixed writes for exactly this reason. Sorry about this suggestion @dnadoba, I'm wondering if we should just revert to only having a `write` version.

glbrntt: Yeah, even if given an index to use in `setMessage` it's still pretty easy to get wrong.

dnadoba: Agree, I will remove the `set` method again.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
dnadoba: Okay, but I think a `get` method would **not** make sense because we could not police that or can we? However, I will add a `read` method.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
dnadoba: Okay, let's keep it this way.

Just out of curiosity, do we get any benefit compared to the "automatic" calculation ?

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// - Throws: if `readMessage` returns a non-nil value and does not read exactly the number of bytes passed to it.
+    /// - Returns: `nil` if the length prefix could not be read, 
+    ///            the length prefix is negative or
+    ///            the buffer does not contain enough bytes to read a message of this length.
+    ///            Otherwise the result of `readMessage`.
+    @inlinable
+    public mutating func readLengthPrefixed<Integer, Result>(
+        endianness: Endianness = .big,
+        as integer: Integer.Type,
+        readMessage: (Integer, inout ByteBuffer) throws -> Result?
```
Lukasa, line 80: Ok, we need to be very careful with this API, I think.

Concretely, under what circumstances should `readMessage` return `nil`? We can only call the `readMessage` function if we have an appropriate number of bytes. `ByteBuffer` traditionally uses `nil` to signal "not enough bytes to parse", but if we call `readMessage` we got as many bytes as we're ever gonna get.

I think if `readMessage` returns `nil` we need to `throw`. This helps make the `Result?` optional unambiguous: it returns `nil` only if there weren't enough bytes for the full body to be present.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+    ) throws -> Result? where Integer: FixedWidthInteger {
+        guard let messageLength = self.readInteger(endianness: endianness, as: Integer.self) else {
+            return nil
+        }
+        guard messageLength >= 0,
+              messageLength <= readableBytes else {
```
Lukasa, line 86:
```suggestion
              messageLength <= self.readableBytes else {
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+            return nil
+        }
+        
+        let readStartIndex = self.readerIndex
+        guard let result = try readMessage(messageLength, &self) else {
```
Lukasa, line 91: I think if we should `slice` ourselves here and pass the slice into the `readMessage` function. This will avoid the need to dance with indices, and will also let the various `readX` functions behave sensibly in the body of `readMessage`.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
Lukasa: Yeah, better composition. When the inner function body returns `Void` then you can't write:

```swift
buffer.writeLengthPrefixed(as: UInt32.self) { $0.writeInteger(someInteger) }
```

With this version, you can. It also saves us a bunch of work, computationally. I'm more than happy for us to _police_ the bounds though: we can add an `assert` statement that enforces that the amount the user claims they wrote is actually what they wrote.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+        endianness: Endianness = .big,
+        as integer: Integer.Type,
+        readMessage: (Integer, inout ByteBuffer) throws -> Result?
+    ) throws -> Result? where Integer: FixedWidthInteger {
+        guard let messageLength = self.readInteger(endianness: endianness, as: Integer.self) else {
+            return nil
+        }
+        guard messageLength >= 0,
+              messageLength <= readableBytes else {
+            return nil
```
glbrntt, line 87: By returning here we can leave the buffer in an odd state: we've read the message length but don't have enough bytes to read the message. If we subsequently append more bytes and call this function again we'll have lost the length from the buffer.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
dnadoba: You can actually write it like in your example because all write methods are marked with `@discardableResult`.

It should also only be two memory reads and one subtraction compared to the manual calculations. 
Maybe my intuitions are wrong but I thought that this should not be noticeable because whatever `writeMessage` does is probably be more expensive.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
glbrntt: +1 on passing in a slice. I think a separate `readLengthPrefixedSlice` would actually be a useful function to have (it could then be used as the base implementation for this function).

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+    ///     - writeMessage:
+    /// - Throws: If the number of bytes written during `writeMessage` can not be exactly represented as the given `Integer` i.e. if the number of bytes written is greater than `Integer.max`
+    /// - Returns: Number of total bytes written
+    @discardableResult
+    @inlinable
+    public mutating func writeLengthPrefixed<Integer>(
+        endianness: Endianness = .big,
+        as integer: Integer.Type,
+        writeMessage: (inout ByteBuffer) throws -> Int
+    ) throws -> Int  where Integer: FixedWidthInteger {
```
glbrntt, line 44:
```suggestion
    ) throws -> Int where Integer: FixedWidthInteger {
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,107 @@
...
+        public static let readLessBytesThanSpecifiedInLengthPrefix: LengthPrefixError = .init(baseError: .readLessBytesThanSpecifiedInLengthPrefix)
+    }
+}
+
+extension ByteBuffer {
+    /// Prefixes a message written by `writeMessage` with the number of bytes written as an `Integer`.
+    /// - Parameters:
+    ///     - endianness: The endianness of the length prefix `Integer` in this `ByteBuffer` (defaults to big endian).
+    ///     - integer: the desired `Integer` type used to write the length prefix
+    ///     - writeMessage:
```
glbrntt, line 35: nit: missing docs here

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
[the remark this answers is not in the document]
Lukasa: Your intuitions are right, it's not that it's computationally expensive, it's that it's cognitively expensive. The load is simply more complex than allowing the method to say how many bytes it wrote.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: Changed to use a `slice`. 
@glbrntt do you mean something like this:
```swift
mutating func readLengthPrefixedSlice<Integer>(
    endianness: Endianness = .big,
    as integer: Integer.Type
) -> ByteBuffer? where Integer: FixedWidthInteger {
    readLengthPrefixed(endianness: endianness, as: Integer.self) { _, slice in slice }
}
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: I have changed the implementation to only move the `readerIndex` forward if everything was read successful.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
glbrntt: More or less -- just the other way around:

```swift
public mutating func readLengthPrefixed<Integer, Result>(
    endianness: Endianness = .big,
    as integer: Integer.Type,
    readMessage: (Int, inout ByteBuffer) throws -> Result?
) rethrows -> Result? where Integer: FixedWidthInteger {
    var slice = self.readLengthPrefixedSlice(...)
    return try readMessage(slice.readableBytes, &slice)
}
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+              var messageBuffer = self.getSlice(at: readerIndex + prefixSize, length: messageLength)
+        else {
+            return nil
+        }
+        
+        guard let result = try readMessage(messageLength, &messageBuffer) else {
+            return nil
+        }
+        
+        _moveReaderIndex(forwardBy: prefixSize + messageLength)
```
glbrntt, line 101:
```suggestion
        self._moveReaderIndex(forwardBy: prefixSize + messageLength)
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+    ///            Otherwise the result of `readMessage`.
+    @inlinable
+    public mutating func readLengthPrefixed<Integer, Result>(
+        endianness: Endianness = .big,
+        as integer: Integer.Type,
+        readMessage: (Int, inout ByteBuffer) throws -> Result?
+    ) rethrows -> Result? where Integer: FixedWidthInteger {
+        let prefixSize = MemoryLayout<Integer>.size
+        
+        guard let lengthPrefix = self.getInteger(at: readerIndex, endianness: endianness, as: Integer.self),
```
glbrntt, line 90:
```suggestion
        guard let lengthPrefix = self.getInteger(at: self.readerIndex, endianness: endianness, as: Integer.self),
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+              let messageLength = Int(exactly: lengthPrefix),
+              var messageBuffer = self.getSlice(at: readerIndex + prefixSize, length: messageLength)
```
glbrntt, line 92:
```suggestion
              var messageBuffer = self.getSlice(at: self.readerIndex + prefixSize, length: messageLength)
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+        else {
+            return nil
+        }
+        
+        guard let result = try readMessage(messageLength, &messageBuffer) else {
```
glbrntt, line 97: Do we want to pass in the message length? It can be recovered from the slice if necessary.

glbrntt: Also wondering if we should drop the `inout` when passing in the slice, I think it makes it clearer that "this is your slice, do what you want with it".

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: I have changed the implementation and it now no longer throws at all. Reading too much is not possible because we pass `readMessage` a `slice` now. Not consuming all bytes is still possible and does not throw. However, I think we want to actually allow this because protocols like [RTP use padding bytes at the end of a message which should be ignored](https://datatracker.ietf.org/doc/html/rfc3550#page-37).

I think it is okay to return nil if `readMessage` returns nil. This mirrors the behavior of `readString(length:)` which also returns nil if decoding fails. Otherwise the following example would throw if String decoding fails:
```swift
buffer.readLengthPrefixed(as: UInt8.self) { length, buffer in 
    buffer.readString(length)
}
```
This would be surprising at least for me.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
glbrntt, line 101: We need to the reader index forward if `readMessage` throws or returns `nil` (since we read enough bytes from the buffer).

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+        as integer: Integer.Type,
+        writeMessage: (inout ByteBuffer) throws -> Int
+    ) throws -> Int where Integer: FixedWidthInteger {
+        var totalBytesWritten = 0
+        
+        let lengthPrefixIndex = self.writerIndex
+        // Write a zero as a placeholder which will later be overwritten by the actual number of bytes written
+        totalBytesWritten += self.writeInteger(.zero, endianness: endianness, as: Integer.self)
+        
+        let startWriterIndex = writerIndex
```
glbrntt, line 47:
```suggestion
        let startWriterIndex = self.writerIndex
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,105 @@
...
+        let messageLength = try writeMessage(&self)
+        let endWriterIndex = writerIndex
```
glbrntt, line 49:
```suggestion
        let endWriterIndex = self.writerIndex
```

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
Lukasa: `readString(length:)` does not return `nil` if decoding fails because decoding cannot fail. Specifically, we use [`String(decoding:as:)`](https://developer.apple.com/documentation/swift/string/2907004-init) which is non-failable. This is not the same as [`String(bytes:encoding:)`](https://developer.apple.com/documentation/swift/string/3126734-init) which _is_ failable.

There should be no `read` method in NIO that returns `nil` in a circumstance other than "there were not enough readable bytes to satisfy the operation", and if there is that's a bug.

dnadoba: Ah I see! Thanks for taking the time to explain it to me :)

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: The "Problem" is that you would need to make your own local copy to actually read something from that slice.
This would no longer work:
```swift
buffer.readLengthPrefixed(as: UInt8.self) { buffer in 
    buffer.readString(buffer.readableBytes)
}
```
You would need to write the following:
```swift
buffer.readLengthPrefixed(as: UInt8.self) { buffer in 
    var buffer = buffer
    return buffer.readString(buffer.readableBytes)
}
```

I'm also thinking about removing the `readLengthPrefixed` method and just keeping `readLengthPrefixedSlice` because you could just write something like this:
```swift
guard var slice = buffer.readLengthPrefixedSlice(as: UInt8.self),
           let message = slice.readString(slice.readableBytes) 
else {
    throw MyCustomError()
}
```
At least in the case of RSocket we probably want to throw a custom error anyway. WDYT?

Lukasa: I think I'd be fine with that.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: Done ✅

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
dnadoba: Added ✅

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -0,0 +1,144 @@
...
+    ) throws -> Result? where Integer: FixedWidthInteger {
+        guard let buffer = self.readLengthPrefixedSlice(endianness: endianness, as: Integer.self) else {
+            return nil
+        }
+        guard let result = try readMessage(buffer) else {
+            throw LengthPrefixError.messageCouldNotBeReadSuccessfully
+        }
+        return result
+    }
+    /// Reads an `Integer` from `self` and reads a slice of that length from `self`
```
glbrntt, line 99:
```suggestion
    }
    
    /// Reads an `Integer` from `self` and reads a slice of that length from `self`
```

Tests/NIOTests/ByteBufferLengthPrefixTests.swift
```diff
@@ -0,0 +1,230 @@
...
+        XCTAssertEqual(
+            try buffer.readLengthPrefixed(endianness: .little, as: UInt16.self) { buffer in
+                var buffer = buffer
+                return buffer.readString(length: buffer.readableBytes)
+            },
+            String(repeating: "A", count: 256)
+        )
+    }
+    func testReadMessageWithMaliciousLength() {
+        buffer.writeInteger(UInt64(Int.max) + 1)
```
glbrntt, line 178: Can we have a similar test with a negative length?

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
glbrntt: Same here.

glbrntt (commented):
I left a formatting suggestion and I think we need one more test. Otherwise this looks great. Up to you whether you want to leave in `readLengthPrefixed` or not.

dnadoba:
I added a space between method declarations, added a test for a negative length prefix and updated the documentation.
I decided to keep the `readLengthPrefixed` method to be more in line with the `writeLengthPrefixed` method.

Lukasa (approved):

glbrntt (approved):
Looks great, thanks @dnadoba!

Lukasa:
@dnadoba Can I recommend you build RSocket with NIO `main`? I believe you'll find that this patch has introduced an ambiguity with your downstream code.

dnadoba:
Yes it has. `ByteBuffer.LengthPrefixError` was also defined by RSocketCore. Thank you! Fixed by https://github.com/rsocket/rsocket-swift/pull/57

PR #1903, opened 2021-07-20
Provide a fallback date for new files.

Lukasa:
Motivation:

The test generation script automatically fills in the dates for the
autogenerated test files. Unfortunately, it doesn't tolerate a brand-new
file very well and puts in year zero. That's not right.

Modifications:

Default to the current year if no years are present.

Result:

Better default dates.

### DIFF
--- a/scripts/generate_linux_tests.rb
+++ b/scripts/generate_linux_tests.rb
@@ -19,6 +19,7 @@
 #
 #   Created by Tony Stone on 5/4/16.
 #
+require 'date'
 require 'getoptlong'
 require 'fileutils'
 require 'pathname'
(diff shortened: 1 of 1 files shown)


glbrntt (approved):

scripts/generate_linux_tests.rb
```diff
@@ -36,6 +37,11 @@ def extractCopyright(log)
     Regexp.last_match.offset(0).first
   end
 
+  # If there are no years, assume this year.
+  if indices.count == 0
+    return "#{Date.today.year}"
+  end
```
agnosticdev, line 43: Great catch!

PR #1904, opened 2021-07-20
Avoid syncShutdownGracefully knowing about MTELG.

Lukasa:
Motivation:

As part of the ongoing split of NIO into two components, we need to
decouple EventLoopGroup from MultiThreadedEventLoopGroup. This is
largely straightforward, but EventLoopGroup.syncShutdownGracefully uses
a hook on MTELG to try to provide debugging help.

That hook cannot stand, so we must break it. Unfortunately this slightly
reduces our ability to detect possible blocks of a SelectableEventLoop,
but in return we get to increase our ability to detect blocks of other
event loop groups.

Modifications:

- Added a debugging hook to detect possible blocking of event loops.

Result:

Less coupling.

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIO/Embedded.swift
+++ b/Sources/NIO/Embedded.swift
@@ -208,6 +208,11 @@ public final class EmbeddedEventLoop: EventLoop {
         return self._promiseCreationStore.removeValue(forKey: futureIdentifier)!
     }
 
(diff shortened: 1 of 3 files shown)


fabianfett (approved):
LGTM

glbrntt (approved):
LGTM

Issue #1891, opened 2021-07-06
NIOFoundationCompat+ByteBuffer: Support JSONSerialization directly


weissi:
In `NIOFoundationCompat` we already have a bunch of useful convenience for `ByteBuffer` + `JSONEncoder`/`JSONDecoder`. We should have the same for `JSONSerialization`. For example:

    try JSONSerialization.jsonObject(with: Data(buffer: byteBuffer), ...)

is tedious, we could directly have

    try JSONSerialization.jsonObject(buffer: byteBuffer, ...)

agnosticdev:
Wanted to get your thoughts on the following:
```swift

extension ByteBuffer {
    
    /// Attempts to derive a Foundation object from a ByteBuffer and return it  as `T`.
    ///
    /// - parameters:
    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
    @inlinable
    public func object<T>(_ type: T.Type) throws -> T? {
        guard let bytes = self.getData(at: 0, length: self.readableBytes),
              let t = try JSONSerialization.jsonObject(with: bytes, options: .mutableContainers) as? T else {
            return nil
        }
        return t
    }
    
    /// Attempts to derive a Foundation object from a ByteBuffer that is transformed into a InputStream and returned  as `T`.
    ///
    /// - parameters:
    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
    @inlinable
    public func objectFromStream<T>(_ type: T.Type) throws -> T? {
        guard let bytes = self.getData(at: 0, length: self.readableBytes) else {
            return nil
        }
        let stream = InputStream(data: bytes)
        stream.open()
        guard let t = try JSONSerialization.jsonObject(with: stream, options: .mutableContainers) as? T else {
            return nil
        }
        stream.close()
        return t
    }
}
```
With the usage looking something like the following:
```swift

let array = ["String1", "String2", "String3"]
let dictionary = ["key1": "val1", "key2": "val2", "key3": "val3"]

var dataArray = try JSONSerialization.data(withJSONObject: self.array, options: .prettyPrinted)
var dataDictionary = try JSONSerialization.data(withJSONObject: self.dictionary, options: .prettyPrinted)

[...]

PR #1906, opened 2021-07-20
issue-1891: API Addition for JSONSerialization in NIOFoundationCompat

agnosticdev:
Motivation:

Issue for #1891 to add JSONSerialization in NIOFoundationCompat.

Modifications:

Adds:
JSONSerialization+ByteBuffer.swift
JSONSerialization+ByteBufferTest.swift
JSONSerialization+ByteBufferTest+XCTest.swift

Result:

Support for transforming byteBuffer data to a Foundation object.


Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIO
+import Foundation
+
+extension JSONSerialization {
+    
+    /// Attempts to derive a Foundation object from a ByteBuffer and return it  as `T`.
```
Lukasa, line 20:
```suggestion
    /// Attempts to derive a Foundation object from a ByteBuffer and return it as `T`.
```

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,33 @@
...
+    ///
+    /// - parameters:
+    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
```
Lukasa, line 23:
```suggestion
    ///    - type: The Foundation type that is attempting to be derived from the ByteBuffer.
```

Tests/NIOFoundationCompatTests/JSONSerialization+ByteBufferTest.swift
```diff
@@ -0,0 +1,52 @@
...
+        var dataDictionary = Data()
+        
+        XCTAssertTrue(JSONSerialization.isValidJSONObject(array), "Array object cannot be converted to JSON")
+        XCTAssertTrue(JSONSerialization.isValidJSONObject(dictionary), "Dictionary object cannot be converted to JSON")
+        
+        XCTAssertNoThrow(dataArray = try JSONSerialization.data(withJSONObject: array, options: .prettyPrinted))
+        XCTAssertNoThrow(dataDictionary = try JSONSerialization.data(withJSONObject: dictionary, options: .prettyPrinted))
+        
+        let arrayDispatchData = dataArray.withUnsafeBytes { DispatchData(bytes: $0) }
+        let dictDispatchData = dataDictionary.withUnsafeBytes { DispatchData(bytes: $0) }
```
Lukasa, line 37: Is there any reason to involve `DispatchData` in this? Seems like we could use the `Data` directly.

agnosticdev: Good call.  Replacing with `ByteBuffer(data: Data(dataArray))` does get rid of `DispatchData`.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
agnosticdev: Fixed in lastest commit.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
agnosticdev: Fixed in latest commit.

agnosticdev:
Looks like I triggered the API Breakage failure with:

```
14:07:50 Build step 'Execute shell' marked build as failure
14:07:50 $ ssh-agent -k
14:07:51 echo Agent pid 23516 killed;
14:07:51 [ssh-agent] Stopped.
14:07:51 [Set GitHub commit status (universal)] ERROR on repos [] (sha:28c1188) with context:swift-nio2-api-breakage-prb
```
It doesn't looks like there is anything I need to do here, but just wanted to ask?

agnosticdev:
@Lukasa feedback has been addressed in the latest commit.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,33 @@
...
+    
+    /// Attempts to derive a Foundation object from a ByteBuffer and return it as `T`.
+    ///
+    /// - parameters:
+    ///    - type: The Foundation type that is attempting to be derived from the ByteBuffer.
+    ///    - buffer: The ByteBuffer being used to derive the Foundation type.
+    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
+    @inlinable
+    public static func jsonObject<T>(_ type: T.Type, buffer: ByteBuffer) throws -> T? {
+        guard let t = try JSONSerialization.jsonObject(with: Data(buffer: buffer), options: .mutableContainers) as? T else {
```
Lukasa, line 28: We probably shouldn't choose `.mutableContainers` ourselves: can we accept the `options` as an argument on our function?

Tests/NIOFoundationCompatTests/JSONSerialization+ByteBufferTest.swift
```diff
@@ -0,0 +1,48 @@
...
+        var dataDictionary = Data()
+        
+        XCTAssertTrue(JSONSerialization.isValidJSONObject(array), "Array object cannot be converted to JSON")
+        XCTAssertTrue(JSONSerialization.isValidJSONObject(dictionary), "Dictionary object cannot be converted to JSON")
+        
+        XCTAssertNoThrow(dataArray = try JSONSerialization.data(withJSONObject: array, options: .prettyPrinted))
+        XCTAssertNoThrow(dataDictionary = try JSONSerialization.data(withJSONObject: dictionary, options: .prettyPrinted))
+        
+        let arrayByteBuffer = ByteBuffer(data: Data(dataArray))
+        let dictByteBuffer = ByteBuffer(data: Data(dataDictionary))
```
Lukasa, line 37: Nit: The extra calls to `Data()` are superfluous, `dataArray` and `dataDictionary` are already `Data` objects.

agnosticdev: Agreed.  This is removed in the latest commit.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
agnosticdev: Yeah, I agree that as soon as we make the decision to use `.mutableContainers` there will be a desire to use `.mutableLeaves` so this would be best served as an argument on the function.  Note that my test case did not include `.fragmentsAllowed` because I did have issues finding this option on Linux. macOS seemed to find it fine.

agnosticdev:
@Lukasa feedback has been addressed in the latest commit.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,34 @@
...
+    
+    /// Attempts to derive a Foundation object from a ByteBuffer and return it as `T`.
+    ///
+    /// - parameters:
+    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
+    ///    - buffer: The ByteBuffer being used to derive the Foundation type.
+    ///    - readingOption: The reading option used when the parser derives the Foundation type from a ByteBuffer.
+    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
+    @inlinable
+    public static func jsonObject<T>(_ type: T.Type, buffer: ByteBuffer, readingOption: ReadingOptions) throws -> T? {
```
Lukasa, line 28: We should probably emulate the equivalent signature as closely as possible, so we should probably call our options argument `options`.

agnosticdev: Done in the latest commit.

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,34 @@
...
+    /// - parameters:
+    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
+    ///    - buffer: The ByteBuffer being used to derive the Foundation type.
+    ///    - options: The reading option used when the parser derives the Foundation type from a ByteBuffer.
+    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
+    @inlinable
+    public static func jsonObject<T>(_ type: T.Type, buffer: ByteBuffer, options: ReadingOptions) throws -> T? {
+        guard let t = try JSONSerialization.jsonObject(with: Data(buffer: buffer), options: options) as? T else {
+            return nil
+        }
```
weissi, line 31:
```suggestion
        return try JSONSerialization.jsonObject(with: Data(buffer: buffer), options: options)
            return nil
        }
```

Sources/NIOFoundationCompat/JSONSerialization+ByteBuffer.swift
```diff
@@ -0,0 +1,34 @@
...
+    
+    /// Attempts to derive a Foundation object from a ByteBuffer and return it as `T`.
+    ///
+    /// - parameters:
+    ///    - type: The Founation type that is attempting to be derived from the ByteBuffer.
+    ///    - buffer: The ByteBuffer being used to derive the Foundation type.
+    ///    - options: The reading option used when the parser derives the Foundation type from a ByteBuffer.
+    /// - returns: The Foundation value if successful or `nil` if there was an issue creating the Foundation type.
+    @inlinable
+    public static func jsonObject<T>(_ type: T.Type, buffer: ByteBuffer, options: ReadingOptions) throws -> T? {
```
weissi, line 28: I think we should exactly match the Foundation signature (modulo `ByteBuffer` instead of `Data`), wdyt @Lukasa ?

```
class func jsonObject(with data: Data, options opt: JSONSerialization.ReadingOptions = []) throws -> Any
```

is Foundation's

Lukasa: Probably, yeah.

agnosticdev: So ditch the generic entirely and just work with `Any` then?
```swift
public static func jsonObject(with buffer: ByteBuffer,
                              options opt: JSONSerialization.ReadingOptions = []) throws -> Any {
    return try JSONSerialization.jsonObject(with: Data(buffer: buffer), options: opt)
}
```

agnosticdev: It looked like this is what you wanted so I pushed a commit up ditching the generic and using `Any`.

agnosticdev:
@Lukasa looks like I broke the build for Swift 5.5.  Let me know if there is something I need to do here.  I can't see anything in Jenkins log that looks like an API fix needs to be made.

agnosticdev:
I'll pull in the main and see if that works out the issue..

Lukasa:
It doesn't look like this breakage has anything to do with you: it was a failure of the syscall perf test and so that's inherently transient.

agnosticdev:
Oh, thanks.  It looks like you just pulled in the main.

agnosticdev:
@Lukasa anything else I should take a look at here?

Lukasa (approved):
LGTM, thanks!

Lukasa:
Ah, one check, can you run the test generation script for me please?

agnosticdev:
Regenerated the tests and added `JSONSerialization+ByteBufferTest+XCTest.swift`.  I seen [this](https://github.com/apple/swift-nio/commit/f9b8f15512ac96089b975952597b61ee8bd8be99) come through today, is there something going on with the test generation script that is causing the soundness check to fail?

Lukasa:
Not any more, now it's just that your test file hasn't been added to the Linux tests file.

agnosticdev:
Added `LinuxMain.swift`.  Hopefully that cleans things up a bit.

Lukasa (approved):

PR #1907, opened 2021-07-21
Move SocketAddress and NIOBSDSocket to NIOCore

Lukasa:
Motivation:

This is the next stage of our move of common interchange objects to
NIOCore and out of the NIO module. This time around we need
SocketAddress, which is part of the Channel API. Sadly, SocketAddress is
not as clean as some of the other ports, because it leaks a number of
POSIX-y concepts. This also forces us to bring along NIOBSDSocket, which
ideally we would not move, but we foolishly exposed as API on
SocketAddress.

Modifications:

- Move NIOBSDSocket API components into NIOCore.
- Split out the internal abstractions for NIOBSDSocket and leave those
  in NIO.
- Move SocketAddress into NIOCore.

Result:

SocketAddress will be sitting in NIOCore.

### DIFF
--- a/Sources/NIO/Interfaces.swift
+++ b/Sources/NIO/Interfaces.swift
@@ -254,7 +254,7 @@ extension NIONetworkDevice {
                                                 dst: $0.baseAddress!,
                                                 size: INET_ADDRSTRLEN)
                 }
-                return SocketAddress(mask, host: mask.addressDescription())
+                return SocketAddress(mask)
             }
             let v6mask: (UINT8) -> SocketAddress? = { _ in
                 var buffer: [CChar] =
@@ -266,7 +266,7 @@ extension NIONetworkDevice {
                                                 dst: $0.baseAddress!,
                                                 size: INET6_ADDRSTRLEN)
                 }
-                return SocketAddress(mask, host: mask.addressDescription())
+                return SocketAddress(mask)
             }
 
             switch pAddress.pointee.Address.lpSockaddr.pointee.sa_family {
@@ -430,7 +430,7 @@ public final class NIONetworkInterface {
                                             dst: $0.baseAddress!,
                                             size: INET_ADDRSTRLEN)
             }
-            return SocketAddress(mask, host: mask.addressDescription())
+            return SocketAddress(mask)
         }
         let v6mask: (UINT8) -> SocketAddress? = { _ in
             var buffer: [CChar] =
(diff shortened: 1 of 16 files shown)


Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -0,0 +1,370 @@
...
+#if !os(Windows)
+extension NIOBSDSocket.Option {
+    /// Indicate when to generate timestamps.
+    public static let so_timestamp: NIOBSDSocket.Option =
+            NIOBSDSocket.Option(rawValue: SO_TIMESTAMP)
+}
+#endif
+
+extension NIOBSDSocket {
+    // Sadly this was defined on BSDSocket, and we need it for SocketAddress.
```
fabianfett, line 334: Do we maybe want to create a ticket already for NIO 3.0? And link it here?

Sources/NIO/Interfaces.swift
```diff
@@ -254,7 +254,7 @@ extension NIONetworkDevice {
                                                 dst: $0.baseAddress!,
                                                 size: INET_ADDRSTRLEN)
                 }
-                return SocketAddress(mask, host: mask.addressDescription())
+                return SocketAddress(mask)
```
Davidde94, line 257: I may have missed something, why do we no longer care about the `host`?

Sources/NIOCore/Utilities.swift
```diff
@@ -22,3 +22,11 @@ internal func debugOnly(_ body: () -> Void) {
     // FIXME: duplicated with NIO.
     assert({ body(); return true }())
 }
+
+/// Allows to "box" another value.
+final class Box<T> {
+    // FIXME: Duplicated with NIO.
```
Davidde94, line 28: Can't we make this public so that NIO gets it too, and avoid the duplication?

Davidde94: Or use spi if we don't want it to be public?

Sources/NIO/Interfaces.swift
Lukasa: `addressDescription` is `internal` to NIOCore, and we added a new initializer to `SocketAddress` that defaults to this behaviour, so the call is identical.

Sources/NIOCore/Utilities.swift
Lukasa: Yes, in principle we can. I'd like to defer that API design question to a later PR for now, as this type is exceedingly trivial, but we do have a number of `FIXME`s around duplicated code that needs to be addressed. I think it should be case-by-case though, as some things will benefit from just being separate (e.g. `Box` is the kind of thing that we maybe just want to keep entirely separate).

Davidde94 (approved):

fabianfett (approved):
I tried my very best to go over this PR with the knowledge I have. Looks good from my side. Only caveat: I think we should document which parts we would like to change, once we go for a major break.

PR #1908, opened 2021-07-22
Add more synchronous pipeline operations.

Lukasa:
Motivation:

NIO provides some scope for performing "synchronous" ChannelPipeline
operations. These are only safe to perform from the event loop to which
the pipeline is bound, and they avoid the need to check what thread
context you're running on in release mode.

Several of the built-in channels currently skip the public, thread-safe
API to ChannelPipeline in favour of directly calling into the
synchronous pipeline interface. This is going to cause us some trouble
when we pull ChannelPipeline out into NIOCore.

To that end, we should add the rest of the ChannelPipeline API surface
area to the SyncOperations structure and convert ourselves over to using
them.

Modifications:

- Implemented the pipeline operations on SyncOperations
- Added a test
- Swapped our channels to use the new APIs.

Result:

Better abstraction between ChannelPipeline and the Channels

### DIFF
--- a/Sources/NIO/ChannelPipeline.swift
+++ b/Sources/NIO/ChannelPipeline.swift
@@ -1206,7 +1206,7 @@ extension ChannelPipeline {
             self._pipeline.flush0()
         }
 
-        /// Fires `flush` from the tail to the head.
+        /// Fires `read` from the tail to the head.
         ///
         /// This method should typically only be called by `Channel` implementations directly.
         public func read() {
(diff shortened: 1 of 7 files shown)


fabianfett (approved):
🚀

Sources/NIO/ChannelPipeline.swift
```diff
@@ -1117,6 +1117,150 @@ extension ChannelPipeline {
...
+
+        /// Fires `flush` from the tail to the head.
+        ///
+        /// This method should typically only be called by `Channel` implementations directly.
+        public func flush() {
+            self.eventLoop.assertInEventLoop()
+            self._pipeline.flush0()
+        }
+
+        /// Fires `flush` from the tail to the head.
```
glbrntt, line 1209:
```suggestion
        /// Fires `read` from the tail to the head.
```

glbrntt (approved):
Looks good aside from typo

PR #1920, opened 2021-07-27
Move Channel abstractions to NIOCore.

Lukasa:
Motivation:

The most important API surface area in NIO are the Channel abstractions.
These are shared in all NIO programs, and are also used by several
projects to implement their I/O abstraction. There are several moving
parts to this abstraction, all of which are moving:

- Channel itself
- ChannelPipeline
- ChannelHandler

As these all move, they force several other pieces of API to move with
them. Most notably they force us to move NIOAny, which also forces us to
move FileHandle and FileRegion. That also forces us to bring over part
of our syscall abstraction. This duplication is acceptable due to its
minimal surface area, but it is definitely a flaw in our abstraction
design that we had to do that at all.

We also need to move the channel option abstraction, AddressedEnvelope,
and the DeadChannel.

Modifications:

- Moved a bunch of the Channel abstraction over.
- Moved Channel-associated types.

Result:

Channel will be part of NIOCore.


glbrntt (approved):
Looks good as far as I can tell.

PR #1924, opened 2021-08-03
Copy Android constants needed by Channel over to NIOCore too

buttaface:
Motivation:

Unbreak Android build from NIOCore not having these constants defined.

Modifications:

Copy the same constants that Channel was using in NIO over.

Result:

Android compiles again and the same tests pass.

This is a consequence of #1920, which [broke my daily Android CI](https://github.com/buttaface/swift-android-sdk/runs/3229068619?check_suite_focus=true#step:13:21).


Lukasa (approved):
LGTM.

PR #1925, opened 2021-08-04
add missing setters to complete implementation of MutableCollection

glessard:
An older version of the standard library supplied a bogus implementation by error.
Calling *that* would have been a crashing bug.

### Motivation:

Fix the compatibility suite,
which was triggered by https://github.com/apple/swift/pull/38509,
this is a fix for https://bugs.swift.org/browse/SR-14850 and https://bugs.swift.org/browse/SR-14848

### Modifications:

Add missing `subscript(bounds:)` setter implementations. They should always have been there.
Note that these minimal implementations simply `fatalError`, but the previous behaviour was also a runtime crash.

### Result:

Package should build with the August 3 2021 snapshot of Swift.


Lukasa (changes requested):
Please let me know if you're not up for actually making fixes and I'll try to expedite doing it myself.

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -263,13 +263,18 @@ extension CircularBuffer: RandomAccessCollection {
...
-        newRing.headBackingIndex = bounds.lowerBound.backingIndex
-        newRing.tailBackingIndex = bounds.upperBound.backingIndex
-        return newRing
+            var newRing = self
+            newRing.headBackingIndex = bounds.lowerBound.backingIndex
+            newRing.tailBackingIndex = bounds.upperBound.backingIndex
+            return newRing
+        }
+        set {
+            fatalError()
```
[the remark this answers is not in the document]
glessard: Good thought. I was still pre-coffee.

glessard:
I added a minimal non-crashing implementation for `MarkedCircularBuffer`, but there are 2 problems with it.
- Its `Index` doesn't allow a `Range<Index>` to have a `count` property,
- That makes it impossible to assert that the bounds of the setter have the same length as the newValue.
- Of course maybe `MarkedCircularBuffer` should also be an RRC, which would make it possible to just call the underlying `CircularBuffer`'s setter directly.

Sources/NIO/MarkedCircularBuffer.swift
```diff
@@ -161,6 +161,13 @@ extension MarkedCircularBuffer: Collection, MutableCollection {
         get {
             return self._buffer[bounds]
         }
+        set {
+            var index = bounds.lowerBound
+            for newElement in newValue where index < bounds.upperBound {
```
Lukasa, line 166: As you're using `formIndex(after:)` this can just be a `while` loop, which terminates when `index == bounds.upperBound`.

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -261,15 +253,30 @@ extension CircularBuffer: RandomAccessCollection {
...
+            var newRing = self
+            newRing.headBackingIndex = bounds.lowerBound.backingIndex
+            newRing.tailBackingIndex = bounds.upperBound.backingIndex
+            return newRing
+        }
+        set {
+            precondition(self.distance(from: self.startIndex, to: bounds.lowerBound) >= 0)
+            precondition(self.distance(from: bounds.upperBound, to: self.endIndex) >= 0)
+
+            replaceSubrange(bounds, with: newValue)
```
Lukasa, line 278: Minor nit: NIO style prefers the explicit use of the `self` keyword.

Tests/NIOTests/CircularBufferTests.swift
```diff
@@ -325,6 +325,23 @@ class CircularBufferTests: XCTestCase {
...
+
+        let index = ring.firstIndex(of: 1)!
+        let originalCount = ring.count
+        XCTAssertEqual(ring[index..<ring.endIndex].count, 2)
+        ring.replaceSubrange(index..<ring.endIndex, with: Array(10..<20))
+        XCTAssertTrue(ring.testOnly_verifyInvariantsForNonSlices())
+        XCTAssertEqual(originalCount + 8, ring.count)
+        XCTAssertEqual(4, ring.first)
+        XCTAssertEqual(19, ring.last)
+    }
```
Lukasa, line 343: This test doesn't actually assign through the subscript.

Lukasa (changes requested):
Thanks, good first step!

Regarding `MarkedCircularBuffer`, it's not an RRC on purpose: the notion of the mark doesn't express itself well when it comes to resizing. As a result, I think the implementation you've written is basically write, except for the note I left in the diff.

Sources/NIO/MarkedCircularBuffer.swift
glessard: What I wanted to do was use `zip`, but the type of `MarkedCircularBuffer.Index` wouldn't allow that. The loop needs two iterators either way, so using it seems to me that using a `while` would just make both iterations explicit. Thoughts?

glessard: After peeking, it looks like what you're saying is that `==` would be faster than `<`.

Sources/NIO/MarkedCircularBuffer.swift
```diff
@@ -161,6 +161,15 @@ extension MarkedCircularBuffer: Collection, MutableCollection {
...
             return self._buffer[bounds]
         }
+        set {
+            var index = bounds.lowerBound
+            var iterator = newValue.makeIterator()
+            while let newElement = iterator.next(), index != bounds.upperBound {
+                self._buffer[index] = newElement
+                formIndex(after: &index)
+            }
+            assert(iterator.next() == nil && index == bounds.upperBound)
```
Lukasa, line 171: Let's make this a `precondition`.

Lukasa (approved):
So long as the tests look good, this should be fine by me.

glessard:
To unblock the compatibility suite, it would help if 2.30.1 were tagged after this. It looks like it might require a branch and a cherry-pick, so I'll squash this to a single commit.

PR #1927, opened 2021-08-05
Move the Codecs to NIOCore

Lukasa:
Motivation:

Our basic Channel Handlers don't need to be in NIO anymore: they aren't
realistically tied to that code. So we can move them to NIOCore.

Modifications:

- Move Codecs to NIOCore
- Move SingleStepByteToMessageDecoder to NIOCore.

Result:

Even more stuff in NIOCore


PeterAdams-A (commented):
Should we move the tests too?

Lukasa:
Thus far we haven't moved any of the tests at all. I'm planning to do a bunch of that cleanup once I think all the main moves are done and we can see where everything lands.

PeterAdams-A (approved):
ok

glbrntt (approved):

PR #1928, opened 2021-08-05
Add appropriate Sendable conformances.

Lukasa:
Motivation:

As part of the adoption of 5.5 there is now some limited amount of
runtime enforcement of thread-safety in the form of the Sendable marker
protocol. A number of NIO's types are safe to share across threads, so
to smooth the path to wider adoption of Sendable we should adopt this
marker protocol where we can.

Modifications:

- Added Sendable to many, many types.

Result:

Users can use NIO in concurrency-aware contexts more easily.

### DIFF
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -1570,3 +1570,16 @@ public struct _NIOEventLoopFutureIdentifier: Hashable {
         return UInt(bitPattern: ObjectIdentifier(future)) ^ 0xbf15ca5d
     }
 }
+
+#if compiler(>=5.5)
+// EventLoopPromise is a reference type, but by its very nature is Sendable.
+extension EventLoopPromise: Sendable { }
+
+// EventLoopFuture is a reference type, but it is Sendable. However, we enforce
+// that by way of the guarantees of the EventLoop protocol, so the compiler cannot
+// check it.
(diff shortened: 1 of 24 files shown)


Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1570,3 +1570,16 @@ public struct _NIOEventLoopFutureIdentifier: Hashable {
...
 }
+
+#if compiler(>=5.5)
+// EventLoopPromise is a reference type, but by its very nature is Sendable.
+extension EventLoopPromise: Sendable { }
+
+// EventLoopFuture is a reference type, but it is Sendable. However, we enforce
+// that by way of the guarantees of the EventLoop protocol, so the compiler cannot
+// check it.
+extension EventLoopFuture: @unchecked Sendable { }
```
fabianfett, line 1581: I think we should only make the EventLoopFuture sendable, IF the value is sendable as well. Users might have classes that they pass through Futures and those must be thread safe. Declaring the Future sendable by default might send the wrong impression.

Lukasa: The `Future` is _always_ sendable: you cannot mutate it directly and all callbacks execute on the same thread.

adam-fowler:
Should those compiler checks also have `canImport(_Concurrency)`

Lukasa:
Yes.

FranzBusch:
This PR can probably be closed now

PR #1930, opened 2021-08-05
Split and move SocketOptionProvider

Lukasa:
Motivation:

SocketOptionProvider is a general interface for setting socket options
that are "wider" than a single platform Int. There is nothing inherent
in NIO to require the definition of this protocol to be there, so we
can move it to NIOCore.

Modifications:

- Move the definition of SocketOptionProvider to NIOCore.

Result:

SocketOptionProvider is generally available.


Sources/NIO/SocketOptionProvider.swift
```diff
@@ -12,272 +12,6 @@
...
-        /// - returns: An `EventLoopFuture` containing the value of the socket option, or
-        ///     any error that occurred while retrieving the socket option.
-        public func getTCPConnectionInfo() -> EventLoopFuture<tcp_connection_info> {
-            return self.unsafeGetSocketOption(level: .tcp, name: .tcp_connection_info)
-        }
-    #endif
-}
-
-
 extension BaseSocketChannel: SocketOptionProvider {
```
glbrntt, line 15: Should we rename this file `BaseSocketChannel+SocketOptionProvider.swift` or  just move this extension to `BaseSocketChannel.swift`?

glbrntt (approved):

PR #1931, opened 2021-08-06
Move MulticastChannel and NIONetworkDevice to NIOCore

Lukasa:
Motivation:

MultcastChannel is a general abstraction for expressing multicast
capabilities on a given Channel. This abstraction doesn't have any
particularly tight tie to the POSIX layer, so it belongs in NIOCore.

This is also expressed in terms of NIONetworkDevice, so we need to move
that over. That also encourages us to bring over
System.enumerateDevices, and given that System.coreCount is also fairly
general-purpose we may as well bring it along too.

Modifications:

- Move MulticastChannel to NIOCore
- Move NIONetworkDevice to NIOCore
- Move System to NIOCore

Result:

More general-purpose abstractions in NIOCore.


glbrntt (approved):
LGTM

PR #1933, opened 2021-08-09
Move EmbeddedChannel to its own library.

Lukasa:
Motivation:

EmbeddedChannel is an important testing tool, and we want to use it
without needing to bring along the POSIX layer. They are not tightly
coupled. However, it also doesn't belong naturally in NIOCore, so we
should probably put it in its own place.

Modifications:

- Moved EmbeddedChannel and EmbeddedEventLoop to NIOEmbedded.
- Moved the tests to NIOEmbeddedTests
- Duplicated some test helpers

Result:

Easy to use EmbeddedChannel without the POSIX layer.


Tests/NIOEmbeddedTests/EmbeddedChannelTest.swift
```diff
@@ -13,7 +13,60 @@
 //===----------------------------------------------------------------------===//
 
 import XCTest
-@testable import NIO
+import NIOCore
+@testable import NIOEmbedded
+
+class ChannelLifecycleHandler: ChannelInboundHandler {
```
fabianfett, line 19: Is that used anywhere?

Lukasa: Yup, in the tests within this file.

Package.swift
```diff
@@ -19,13 +19,15 @@ var targets: [PackageDescription.Target] = [
     .target(name: "NIOCore",
             dependencies: ["NIOConcurrencyHelpers", "CNIOLinux"]),
     .target(name: "_NIODataStructures"),
+    .target(name: "NIOEmbedded", dependencies: ["NIOCore", "_NIODataStructures"]),
```
PeterAdams-A, line 22: Don't like the name.

This is all about Testing really?  Embedded also don't give much scope for adding things here.  Also makes me think of embedded systems.

Tests/NIOEmbeddedTests/EmbeddedChannelTest.swift
fabianfett: Found it! Thanks! Where did this code live before? Is this now duplicated?

Lukasa: Yup, it lived in NIO’s channel tests. I don’t think we’re likely to bother deduplicating it though, it’s fairly trivial.

Package.swift
Lukasa: That’s true, but I don’t want us to push too hard on testing: while this is its primary use, there are reasons you might want to use it in production code.

Maybe `NIOSingleThreaded`?

glbrntt: I don't mind the `NIOEmbedded` name.

I wonder if explaining the various "main" NIO modules (`NIO`, `NIOCore`, `NIOEmbedded`, `NIOFoundationCompat` etc) in the README would be useful addition.

glbrntt (approved):

Package.swift
PeterAdams-A: What's the real reason why it doesn't fit anywhere else?  Is regular EL still in NIO[Posix]?

Lukasa: Define "regular EL"?

Lukasa: The `EventLoop` protocol is in NIOCore. The SelectableEventLoop is in NIO, and will stay there as it's inherently POSIX-specific.

Lukasa: Discussed in a stand-up meeting, agreed to go ahead.

PR #1935, opened 2021-08-12
Clean up dependencies and imports.

Lukasa:
Motivation:

As we've largely completed our move to split out our core abstractions,
we now have an opportunity to clean up our dependencies and imports. We
should arrange for everything to only import NIO if it actually needs
it, and to correctly express dependencies on NIOCore and NIOEmbedded
where they exist.

We aren't yet splitting out tests that only test functionality in
NIOCore, that will follow in a separate patch.

Modifications:

- Fixed up imports
- Made sure our protocols only require NIOCore.

Result:

Better expression of dependencies.

### Motivation:

### Modifications:

### Result:


Sources/NIOWebSocket/WebSocketErrorCodes.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2032 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5:
```suggestion
// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```

glbrntt (approved):

Sources/NIOWebSocket/WebSocketErrorCodes.swift
Lukasa: Wait, I didn't write this in the future?

glbrntt (approved):

fabianfett (approved):

PR #1936, opened 2021-08-13
Move NIO to NIOPosix, make NIO a shell.

Lukasa:
Motivation:

The remaining NIO code really conceptually belongs in a module called
NIOPosix, and NIOCore should really be called NIO. We can't really do
that last step, but we can prepare by pushing the bulk of the remaining
code into a module called NIOPosix.

Modifications:

- Move NIO to NIOPosix
- Make NIO an umbrella module.

Result:

NIOPosix exists.


Tests/NIOPosixTests/NIOAnyDebugTest.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) YEARS Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5: ha, nice catch

glbrntt (approved):

PeterAdams-A (approved):
Seems reasonable - if you can get CI working again.

Package.swift
```diff
@@ -41,57 +44,59 @@ var targets: [PackageDescription.Target] = [
     .target(name: "NIOHTTP1",
             dependencies: ["NIO", "NIOCore", "NIOConcurrencyHelpers", "CNIOHTTPParser"]),
     .target(name: "NIOEchoServer",
-            dependencies: ["NIO", "NIOCore", "NIOConcurrencyHelpers"]),
+            dependencies: ["NIOPosix", "NIOCore", "NIOConcurrencyHelpers"]),
```
fabianfett, line 47: We still need `NIO` here, for reasons you are way to much aware off :)

fabianfett (approved):

PR #1937, opened 2021-08-16
Re-add NIO as a dependency of our library targets.

Lukasa:
Motivation:

In #1935 I removed NIO as a dependency of a number of our library
targets. This unfortunately led to some downstream breakage in modules
that were (incorrectly) assuming they could import NIO without
expressing a dependency on it in their Package.swift, e.g.
swift-server/swift-aws-lambda-runtime#218.

While a forums thread
(https://forums.swift.org/t/semantic-versioning-should-removing-a-dependency-be-a-semver-major/51179)
is ongoing to discuss the implications of this, we should re-add the
dependency to undo the breakage against main. I've validated this
locally: merely having the dependency is enough, we don't have to use
it.

Modifications:

- Re-add NIO as a dependency to:
    - _NIOConcurrency
    - NIOFoundationCompat
    - NIOHTTP1
    - NIOTLS
    - NIOWebSocket

Result:

Broken downstreams can build again.


fabianfett (approved):

Lukasa:

PR #1939, opened 2021-08-16
Add AsyncSequence helpers

fabianfett:
This PR does not have any tests yet. It is here, to demonstrate some `AsyncSequence` and `ByteBuffer` helpers that we might want to add to make our consumers lifes' easier.

### Motivation:

In the [[RFC] for an async/await API in AsyncHTTPClient](https://github.com/swift-server/async-http-client/pull/406#issuecomment-899365012) @adam-fowler asked, if there could be a helper method to transform an AsyncSequence of `ByteBuffer`s into a single `ByteBuffer`. Following our strict rule of *implementing conformances only to types you we own* we must add those convenience helpers to SwiftNIO directly.

### Modifications:

- Add a convenience helper `func consume(maxBytes: Int) async throws -> ByteBuffer` that consumes an AsyncSequence of `ByteBuffer`s into a single `ByteBuffer`
- Add a convenience helper `func bytes()` that translates an `AsyncSequence` of `ByteBuffer`s into an AsyncSequence of Bytes (`UInt8`)


Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+        
+        @usableFromInline
+        enum State {
+            case hasBuffer(ByteBuffer, Upstream.AsyncIterator)
+            case askForMore(Upstream.AsyncIterator)
+            case finished
+            case modifying
+        }
+        
+        @usableFromInline
```
Lukasa, line 34: Prefer `inlinable` here.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+        @inlinable
+        public mutating func next() async throws -> Element? {
+            switch self.state {
+            case .askForMore(var upstream):
+                self.state = .modifying
+                
+                while true {
+                    switch try await upstream.next() {
+                    case .some(let nextBuffer) where nextBuffer.readableBytes == 0:
+                        break
```
Lukasa, line 48: Please avoid `break` in `switch` statements within loops: it can be very hard to work out what this is doing. I suggest either using a labelled loop and then a labelled `continue` statement, or just writing `()` in the body.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+        @usableFromInline
+        init(_ upstream: Upstream.AsyncIterator) {
+            self.state = .askForMore(upstream)
+        }
+        
+        @inlinable
+        public mutating func next() async throws -> Element? {
+            switch self.state {
+            case .askForMore(var upstream):
+                self.state = .modifying
```
Lukasa, line 43: Can we try to prove out whether 5.5 still has the CoW issue that required the use of `modifying` in the first place? I think it should be easy enough to prove, and if it doesn't I'd rather we just removed it.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+                        break
+                        
+                    case .some(var nextBuffer):
+                        assert(nextBuffer.readableBytes > 0)
+                        let result = nextBuffer.readInteger(as: UInt8.self)
+                        if nextBuffer.readableBytes > 0 {
+                            self.state = .hasBuffer(nextBuffer, upstream)
+                        } else {
+                            self.state = .askForMore(upstream)
+                        }
```
Lukasa, line 57: This code is repeated twice: can we factor it out? We can probably use an `init` on `State` that contains this logic.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+    
+    @inlinable
+    public func makeAsyncIterator() -> Iterator {
+        Iterator(self.upstream.makeAsyncIterator())
+    }
+    
+    @usableFromInline
+    let upstream: Upstream
+    
+    /*private but*/ @usableFromInline init(_ upstream: Upstream) {
```
Lukasa, line 95: Prefer `inlinable`.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+        self.upstream = upstream
+    }
+}
+
+@usableFromInline
+struct TooManyBytesError: Error {
```
Lukasa, line 101: Must be `public`.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+    @usableFromInline
+    init() {}
```
Lukasa, line 103: Should be `public`.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
fabianfett: Tested this in a small side repo. Sadly we still need the `modifying` state.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+        public mutating func next() async throws -> Element? {
+            try await self.state.next()
+        }
+    }
+    
+}
+
+public struct TooManyBytesError: Error {
+    public init() {}
+}
```
fabianfett, line 104: @Lukasa should we prefix with `NIO`?

Lukasa: Yup.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,141 @@
...
+                case .hasBuffer(var buffer, let upstream):
+                    assert(buffer.readableBytes > 0)
+                    let result = buffer.readInteger(as: UInt8.self)
+                    self = .init(buffer: buffer, upstream: upstream)
+                    return result
+                    
+                case .finished:
+                    return nil
+                }
+            }
```
fabianfett, line 86: @Lukasa moved into enum, to get CoW fast path!

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+#if swift(>=5.5)
+import struct NIO.ByteBuffer
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public struct NIOByteBufferToUInt8AsyncSequence<Upstream: AsyncSequence>: AsyncSequence where Upstream.Element == ByteBuffer {
+    public typealias Element = UInt8
+    public typealias AsyncIterator = Iterator
+    
+    @usableFromInline
+    let upstream: Upstream
```
Lukasa, line 24: Nit: this may as well be `var`. It doesn't _really_ matter but it can't hurt.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+    public func toBytes() -> NIOByteBufferToUInt8AsyncSequence<Self> {
+        NIOByteBufferToUInt8AsyncSequence(self)
+    }
+    
+    /// Consume an ``Swift/AsyncSequence`` of ``NIO/ByteBuffer``s into a single `ByteBuffer`.
+    ///
+    /// - Parameter maxBytes: The maximum number of bytes that the result ByteBuffer is allowed to have.
+    /// - Returns: A ``NIO/ByteBuffer`` that holds all the bytes of the AsyncSequence
+    @inlinable
+    public func consume(maxBytes: Int) async throws -> ByteBuffer? {
```
Lukasa, line 114: Hmm, is returning `nil` here a good idea? I'm a bit inclined to want to return an empty buffer.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+        
+        @inlinable
+        public mutating func next() async throws -> Element? {
+            switch self.state {
+            case .askForMore(var upstream):
+                while true {
+                    switch try await upstream.next() {
+                    case .some(let nextBuffer) where nextBuffer.readableBytes == 0:
+                        // we received an empty buffer. for this reason, let's continue and get the
+                        // next buffer fro, the sequence
```
glbrntt, line 68:
```suggestion
                        // next buffer from the sequence
```

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+                        continue
+                        
+                    case .some(var nextBuffer):
+                        assert(nextBuffer.readableBytes > 0)
+                        let result = nextBuffer.readInteger(as: UInt8.self)
```
glbrntt, line 73: Should this be banged? We assert that there must readable bytes so reading one mustn't fail

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.5)
+import struct NIO.ByteBuffer
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+public struct NIOByteBufferToUInt8AsyncSequence<Upstream: AsyncSequence>: AsyncSequence where Upstream.Element == ByteBuffer {
```
glbrntt, line 19: I don't love the name. `URLSession` uses `AsyncBytes` (i.e. `URLSession.AsyncBytes`) which I think is quite clear.  How about `ByteBuffer.AsyncBytes`?

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+    /// Transform an AsyncSequence of ByteBuffers into an AsyncSequence of single bytes.
+    @inlinable
+    public func toBytes() -> NIOByteBufferToUInt8AsyncSequence<Self> {
+        NIOByteBufferToUInt8AsyncSequence(self)
+    }
+    
+    /// Consume an ``Swift/AsyncSequence`` of ``NIO/ByteBuffer``s into a single `ByteBuffer`.
+    ///
+    /// - Parameter maxBytes: The maximum number of bytes that the result ByteBuffer is allowed to have.
+    /// - Returns: A ``NIO/ByteBuffer`` that holds all the bytes of the AsyncSequence
```
glbrntt, line 112: Not necessarily _all_ the bytes

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
glbrntt: `consume` feels to me like we're actually "reading" the bytes, collect might be more appropriate.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
```diff
@@ -0,0 +1,136 @@
...
+
+public struct NIOTooManyBytesError: Error {
+    public init() {}
+}
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+extension AsyncSequence where Element == ByteBuffer {
+    /// Transform an AsyncSequence of ByteBuffers into an AsyncSequence of single bytes.
+    @inlinable
+    public func toBytes() -> NIOByteBufferToUInt8AsyncSequence<Self> {
```
glbrntt, line 105: Should this be a computed property? 

```swift
for try await byte in body.bytes {
  // ...
}
```

vs.

```swift
for try await byte in body.toBytes() {
  // ...
}
```

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
glbrntt: I wonder if an additional lower-level variant accepting an `inout ByteBuffer` would also be useful? Callers could for example, reserve capacity or collect from  multiple sequences into the same buffer without an intermediate copy.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
karwa: Is this not basically an async version of FlattenSequence/.joined() from the standard library, specialised for ByteBuffer?

So... ByteBuffer.FlattenSequence, perhaps?

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
karwa: joined() seems like a better name for this, to mirror both FlattenSequence and the String concatenation helper from the standard library.

Sources/_NIOConcurrency/AsyncSequenceSupport.swift
karwa: This seems like it would be more useful as a general "collect bytes from this stream of buffers" API on ByteBuffer itself.

Which means I think there might be benefit to moving it, and loosening the constraints to allow for async sequences of any buffer type (Data, Array<UInt8>, etc).

PR #1941, opened 2021-08-17
Update NIOAsyncAwaitDemo to use correct @available checks (post WWDC '21)

fabianfett:
### Motivation

`import _Concurrency` was needed to enable Swift 5.5 new async/await support. Gladly this is no longer needed.

### Changes

- Remove `import _Concurrency` statements

### Result

- Cleaner imports
- static-stdlib linking works on 5.5 (it does not if we explicitly import _Concurrency)

### DIFF
--- a/Sources/NIOAsyncAwaitDemo/main.swift
+++ b/Sources/NIOAsyncAwaitDemo/main.swift
@@ -18,7 +18,6 @@ import NIOHTTP1
 import Dispatch
 
 #if compiler(>=5.5)
-import _Concurrency
 
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 func makeHTTPChannel(host: String, port: Int, group: EventLoopGroup) async throws -> AsyncChannelIO<HTTPRequestHead, NIOHTTPClientResponseFull> {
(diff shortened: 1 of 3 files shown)


Lukasa (approved):

Sources/NIOAsyncAwaitDemo/main.swift
```diff
@@ -81,9 +79,5 @@ if #available(macOS 9999, iOS 9999, watchOS 9999, tvOS 9999, *) {
 }
 dg.wait()
 #else
-print("ERROR: This demo only works with async/await enabled (NIO.System.hasAsyncAwaitSupport = \(NIO.System.hasAsyncAwaitSupport))")
-print("Try:   swift run -Xswiftc -Xfrontend -Xswiftc -enable-experimental-concurrency NIOAsyncAwaitDemo")
-#endif
-#else
-print("ERROR: Concurrency only supported on Swift > 5.4.")
+print("ERROR: Concurrency only supported on Swift > 5.5.")
```
glbrntt, line 82:
```suggestion
print("ERROR: Concurrency only supported on Swift >= 5.5.")
```

glbrntt (approved):

PR #1942, opened 2021-08-18
Add compiler setting for __APPLE_USE_RFC_3542 for CNIODarwin

glbrntt:
Motivation:

In #1888 a C compiler setting was added to Package.swift for CNIODarwin.
This isn't reflected in the associated CNIODarwin podspec so the pod
fails to build.

Modifications:

Modify the build_podspecs.sh script to add the missing compiler setting
for CNIODarwin.

Result:

CNIODarwin pod builds.

### DIFF
--- a/scripts/build_podspecs.sh
+++ b/scripts/build_podspecs.sh
@@ -83,6 +83,11 @@ for target in "${targets[@]}"; do
     libraries="s.libraries = 'z'"
   fi
 
+  compiler_flags=""
+  if [[ "$target" == "CNIODarwin" ]]; then
+    compiler_flags="s.compiler_flags = '-D__APPLE_USE_RFC_3542=1'"
+  fi
+
   cat > "${tmpfile}/${target}.podspec" <<- EOF
 Pod::Spec.new do |s|
   s.name = '$target'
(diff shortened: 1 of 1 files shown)


scripts/build_podspecs.sh
```diff
@@ -83,6 +83,11 @@ for target in "${targets[@]}"; do
     libraries="s.libraries = 'z'"
   fi
 
+  compiler_flags=""
+  if [[ "$target" == "CNIODarwin" ]]; then
+    compiler_flags="s.compiler_flags = '-D__APPLE_USE_RFC_3542=1'"
```
weissi, line 88: @glbrntt if we set this unconditionally, why don't we just `#define __APPLE_USE_RFC_3542 1` before the `include` of the relevant header (that makes use of this)?

glbrntt: Ah good point! I went for parity with the Package.swift but this is better, thanks for the suggestion.

weissi: Then we can probably also just take it out of `Package.swift`, no?

glbrntt: That's right, yes.

Unfortunately I can't get this to build this way -- perhaps there's a reason it was defined in Package.swift to begin with. Will have another look later.

Lukasa:
@glbrntt Where are we with this?

glbrntt:
> @glbrntt Where are we with this?

I couldn't get this working with weissi's suggestion. I suggest we review it as it is.

Lukasa (approved):

PR #1946, opened 2021-08-25
Use default SubSequence for StaticString

baarde:
### Motivation:

The default `SubSequence` for `Collection` is `Slice<Self>`. When using `ArraySlice<UInt8>`, we should provide an implementation of `subscript(_:)` for `Range<Int>`, but we don't. Therefore:
* The latest version of the compiler refuses to compile.
* Older versions use `subscript(_:)` for `R: RangeExpression` instead, which causes the app to crash when trying to slice `StaticString` due to infinite recursion.

### Modifications:

`StaticString.SubSequence` is now `Slice<StaticString>`.

### Result:

The code compiles with the latest version of the compiler and slicing a `StaticString` doesn't cause the app to crash.

### DIFF
--- a/Sources/NIO/ContiguousCollection.swift
+++ b/Sources/NIO/ContiguousCollection.swift
@@ -20,7 +20,6 @@ public protocol ContiguousCollection: Collection {
 
 extension StaticString: Collection {
     public typealias Element = UInt8
-    public typealias SubSequence = ArraySlice<UInt8>
     
     public typealias _Index = Int
     
--- a/Tests/NIOTests/ByteBufferTest.swift
+++ b/Tests/NIOTests/ByteBufferTest.swift
@@ -1036,7 +1036,10 @@ class ByteBufferTest: XCTestCase {
             buf.write(bytes: ptr.dropFirst(0)) + buf.write(bytes: ptr.dropFirst(4 /* drop all of them */))
         }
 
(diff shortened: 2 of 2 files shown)


baarde:
This is a PR for SwiftNIO 1.14.

I currently use SwiftNIO 1 in a project and cannot update to SwiftNIO 2 at the moment. The project no longer compiles with the latest version of the compiler (Swift version 5.5 (swiftlang-1300.0.29.102 clang-1300.0.28.1)).

I know nobody likes maintaining old branches (I certainly don't). But I think I am probably not the other developer out there to be stuck with SwiftNIO 1.

fabianfett:
Hi @baarde, 

NIO2 was released more than two and a half years ago. The Swift 5.5 compiler is currently in pre-release. While we appreciate the reasons for your patch, we must state that this change would be source breaking. Admittedly it would allow SwiftNIO compilation with Swift 5.5, but nonetheless existing code, that depends on SwiftNIO1, might not be able to compile after applying your patch.

We think those are your best options forward:
- [update to SwiftNIO 2](https://github.com/apple/swift-nio/blob/2.29.0/docs/migration-guide-NIO1-to-NIO2.md)
- use an older Swift toolchain
- depend on your own fork of SwiftNIO1

If you have further questions regarding updating to SwiftNIO2, we are happy to help. We help in the Swift forums, here in GitHub issues or in the swift-server slack.

PR #1950, opened 2021-09-02
Fix doc generation and jazzy version

glbrntt:
From: https://github.com/apple/swift-nio-extras/pull/141

motivation: fix ci

changes:
* only install ruby and jazzy on focal since jazzy is not supported on older versions of ubuntu
* fix doc generations script adjusting it to latest source-kitten syntax


tomerd:

tomerd (approved):

Lukasa (approved):

PR #1951, opened 2021-09-06
Add safer abstraction for kqueue filter set.

Lukasa:
Motivation:

Our kqueue filter set differ logic attempted to stack allocate some
kevent objects and then treat them as a fixed size array by creating a
tuple and then grabbing a pointer to it. That's fairly gratuitously
unsafe, and we should shrink the amount of code that can touch unsafe
stuff.

Modifications:

- Defined KeventTriple, a value type that encapsulates a fixed-size
  array and a length.
- Rewrote calculateKQueueFilterSetChanges to use the new type.

Result:

Safer code in the NIO core

### DIFF
--- a/Sources/NIOPosix/SelectorKqueue.swift
+++ b/Sources/NIOPosix/SelectorKqueue.swift
@@ -67,29 +67,24 @@ extension KQueueEventFilterSet {
                                         registrationID: SelectorRegistrationID,
                                          _ body: (UnsafeMutableBufferPointer<kevent>) throws -> Void) rethrows {
         // we only use three filters (EVFILT_READ, EVFILT_WRITE and EVFILT_EXCEPT) so the number of changes would be 3.
-        var keventsHopefullyOnStack = (kevent(), kevent(), kevent())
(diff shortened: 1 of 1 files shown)


Sources/NIOPosix/SelectorKqueue.swift
```diff
@@ -332,4 +327,62 @@ extension kevent {
...
+            // well before we overflow.
+            self.initialized &+= 1
+        }
+
+        self[self.initialized].setEvent(fileDescriptor: fd, filter: filter, flags: flags, registrationID: registrationID)
+    }
+
+    mutating func withUnsafeBufferPointer(_ body: (UnsafeMutableBufferPointer<kevent>) throws -> Void) rethrows {
+        try withUnsafeMutablePointer(to: &self.kevents) { keventPtr in
+            // Pointer to a homogeneous tupe of a given type is also implicitly bound to the element type, so
```
glbrntt, line 379:
```suggestion
            // Pointer to a homogeneous tuple of a given type is also implicitly bound to the element type, so
```

glbrntt (approved):
Looks good

PR #1952, opened 2021-09-13
Improve performance of `HTTPHeaders.subscript(canonicalForm:)`

glbrntt:
Motivation:

When getting the canonical form of header values any ascii whitespace
is stripped from the values. The current implementation does this by
doing equality checks on `Character` which is quite expensive.

Modifications:

- Update the `Substring.trimWhitespace()` function to trim on a UTF8
  view
- Add performance tests.

Result:

Retrieving the canonical form of header values is cheaper, significantly
so when values contain whitespace.

On my machine:
- http_headers_canonical_form: 0.16
- http_headers_canonical_form_trimming_whitespace (without patch): 0.47
- http_headers_canonical_form_trimming_whitespace (with patch): 0.24

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -540,28 +540,37 @@ extension HTTPHeaders: RandomAccessCollection {
     }
 }
 
-extension Character {
+extension UTF8.CodeUnit {
     var isASCIIWhitespace: Bool {
-        return self == " " || self == "\t" || self == "\r" || self == "\n" || self == "\r\n"
+        switch self {
+        case UInt8(ascii: " "),
+             UInt8(ascii: "\t"),
+             UInt8(ascii: "\r"),
+             UInt8(ascii: "\n"):
+          return true
+
+        default:
+          return false
+        }
     }
 }
 
 extension String {
     func trimASCIIWhitespace() -> Substring {
-        return self.dropFirst(0).trimWhitespace()
+        return Substring(self).trimWhitespace()
     }
 }
 
-private extension Substring {
-    func trimWhitespace() -> Substring {
-        var me = self
-        while me.first?.isASCIIWhitespace == .some(true) {
-            me = me.dropFirst()
+extension Substring {
+    fileprivate func trimWhitespace() -> Substring {
+        guard let firstNonWhitespace = self.utf8.firstIndex(where: { !$0.isASCIIWhitespace }) else {
+          // The whole substring is ASCII whitespace.
+          return Substring()
         }
-        while me.last?.isASCIIWhitespace == .some(true) {
-            me = me.dropLast()
-        }
-        return me
+
+        // There must be at least one non-ascii whitespace character, so banging here is safe.
+        let lastNonWhitespace = self.utf8.lastIndex(where: { !$0.isASCIIWhitespace })!
+        return Substring(self.utf8[firstNonWhitespace...lastNonWhitespace])
     }
 }
 


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -540,28 +540,37 @@ extension HTTPHeaders: RandomAccessCollection {
...
+extension Substring {
+    fileprivate func trimWhitespace() -> Substring {
+        guard let firstNonWhitespace = self.utf8.firstIndex(where: { !$0.isASCIIWhitespace }) else {
+          // The whole substring is ASCII whitespace.
+          return Substring()
         }
-        return me
+
+        // There must be at least one non-ascii whitespace character, so banging here is safe.
+        let lastNonWhitespace = self.utf8.lastIndex(where: { !$0.isASCIIWhitespace })!
```
Lukasa, line 572: I don't believe this force-unwrap is safe: we can in principle decode an empty string, and while that's not really valid in our specific use-case we should probably have some defense in depth against that bug making its way into lower areas of the code, or with other areas.

Davidde94 (approved):
LGTM.

Lukasa (changes requested):
Thanks for this! I've left a note in the diff.

Can I also ask you to add an allocation counter test for this? Perhaps in a separate PR so we get a nice before-and-after? Re-using either or both of the perf tests would be fine.

Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -540,28 +540,37 @@ extension HTTPHeaders: RandomAccessCollection {
...
+        switch self {
+        case UInt8(ascii: " "),
+             UInt8(ascii: "\t"),
+             UInt8(ascii: "\r"),
+             UInt8(ascii: "\n"):
+          return true
+
+        default:
+          return false
+        }
```
Davidde94, line 554: Apparently two are missing:
- Vertical tab `\v`
- Form feed `\f`

Lukasa: I'm disinclined to make this more complex. We're only really trying to catch the [`OWS` rule from the HTTP -semantics draft](https://www.ietf.org/archive/id/draft-ietf-httpbis-semantics-19.html#name-whitespace), which strictly forbids newline characters, and I don't think we really expect to see them, so arguably we could even remove the `\r` and `\n` cases. VTAB and form-feed should also be disregarded by this loop for the same reason.

Sources/NIOHTTP1/HTTPTypes.swift
glbrntt: I don't see how this is unsafe given `firstNonWhitespace` is known to not be `nil` at this point: we're checking for the same predicate but from the other direction?

Lukasa: Oh apologies, you're right. I missed that we were duplicating the check.

Sources/NIOHTTP1/HTTPTypes.swift
Davidde94: Sounds fine to me tbh, although it says `SHOULD` which I'm sure someone somewhere interprets as "not necessary"

Lukasa: Sure, you can put a VTAB in there: the mistake would be for us to strip it, not for us to fail to. If your VTAB makes it to the app the app is naturally going to flip its lid: what we shouldn't do is treat it as though it was accidentally equivalent to SP or HTAB.

glbrntt:
> Can I also ask you to add an allocation counter test for this? Perhaps in a separate PR so we get a nice before-and-after? Re-using either or both of the perf tests would be fine.

Do you expect a difference or is this just for completeness? (I ask because there is no difference)

Sources/NIOHTTP1/HTTPTypes.swift
glbrntt: Should we follow up with a PR to remove `\r` and `\n`?

Lukasa:
Completeness. We should actually add a test case that has a long header field name (more than 15 characters long both before and after the slice), because I had to do a lot of code reading to convince myself that allocating a `Substring` from a `Substring.UTF8View` would not allocate new backing storage.

glbrntt:

glbrntt:
Relevant bits:

name | current | previous | winner | diff
-|-|-|-|-
http_headers_canonical_form | 0.088066669 | 0.105399777 | current | -16%
http_headers_canonical_form_trimming_whitespace | 0.166675255 | 0.45974008 | current | -63%
http_headers_canonical_form_trimming_whitespace_from_short_string | 0.152118162 | 0.294293338 | current | -48%
http_headers_canonical_form_trimming_whitespace_from_long_string | 0.233780961 | 0.367409353 | current | -36%

Lukasa (approved):

PR #1953, opened 2021-09-13
Add performance and allocation tests for canonical form headers

glbrntt:
Motivation:

To justify performance changes we need to measure the code being
changed. We believe that `HTTPHeaders.subscript(canonicalForm:)` is a
little slow.

Modifications:

- Add allocation and performance tests for fetching header values in
  their canonical form

Results:

More benchmarks!


Lukasa (approved):
Shiny, LGTM. This also validates my expectation, which is that the initial code did not allocate when stripping the whitespace.

PR #1954, opened 2021-09-13
Don't count CR or LF as whitespace when trimming canonical header values

glbrntt:
Motivation:

Follow up to https://github.com/apple/swift-nio/pull/1952#discussion_r707351015

The OWS rule from the HTTP semantics draft only considers 'SP' and
'HTAB' to be whitespace. We also (unnecessarily) consider CR and LF to
be whitespace.

Modifications:

- Remove CR and LF from the characters we consider to be whitespace
- Update tests

Result:

We no longer consider CR or LF to be whitespace when trimming
whitespace when producing the canonical form of header values.


Davidde94 (approved):

Lukasa:

Lukasa (approved):
LGTM

Lukasa:
Not surprised there isn't much of a change here. :wink:

PR #1959, opened 2021-09-17
[NIOSingleStepByteToMessageProcessor] Inline process methods

fabianfett:
We like speed. To make `NIOSingleStepByteToMessageProcessor` even faster, we should mark the process method as `@inlinable`.

NIOSingleStepByteToMessageProcessor overhead is reduced:

Previous:
without_inline

Inlined:
with_inline

The important bit is the relative performance improvement between `withNonCowBuffer` and the actual decode invocation.

### Result:

Faster `NIOSingleStepByteToMessageProcessor`.

### DIFF
--- a/Sources/NIOCore/SingleStepByteToMessageDecoder.swift
+++ b/Sources/NIOCore/SingleStepByteToMessageDecoder.swift
@@ -205,13 +205,14 @@ public final class NIOSingleStepByteToMessageProcessor<Decoder: NIOSingleStepByt
     /// - parameters:
     ///     - decoder: The `NIOSingleStepByteToMessageDecoder` to decode the bytes into message.
     ///     - maximumBufferSize: The maximum number of bytes to aggregate in-memory.
+    @inlinable
     public init(_ decoder: Decoder, maximumBufferSize: Int? = nil) {
         self.decoder = decoder
         self.maximumBufferSize = maximumBufferSize
     }
 
-    @usableFromInline
-    func append(_ buffer: ByteBuffer) {
+    @inlinable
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/SingleStepByteToMessageDecoder.swift
```diff
@@ -230,7 +236,8 @@ public final class NIOSingleStepByteToMessageProcessor<Decoder: NIOSingleStepByt
         return result
     }
 
-    private func decodeLoop(decodeMode: DecodeMode, seenEOF: Bool = false, _ messageReceiver: (Decoder.InboundOut) throws -> Void) throws {
+    @inlinable
+    func decodeLoop(decodeMode: DecodeMode, seenEOF: Bool = false, _ messageReceiver: (Decoder.InboundOut) throws -> Void) throws {
```
Lukasa, line 240: Can you prefix this with an underscore so we know it was supposed to be private?

Sources/NIOCore/SingleStepByteToMessageDecoder.swift
```diff
@@ -206,7 +210,8 @@ public final class NIOSingleStepByteToMessageProcessor<Decoder: NIOSingleStepByt
         self.maximumBufferSize = maximumBufferSize
     }
 
-    private func append(_ buffer: ByteBuffer) {
+    @usableFromInline
+    func append(_ buffer: ByteBuffer) {
```
Lukasa, line 214: This may as well be `inlinable` too, as should `init`.

Lukasa (approved):

PR #1962, opened 2021-09-20
Update workarounds list with our recent ones

Lukasa:


glbrntt (approved):
We should probably review these periodically to see if we can reverse any of them.

glbrntt:
Merging over nightly failures (missing sendable conformance)

PR #1956, opened 2021-09-15
Work around Xcode 13 GM SDK issues.

Lukasa:
Motivation:

Xcode 13 GM shipped with a Swift overlay for libsystem in macOS that
marked free's first argument as non-nullable. This leads to an awkward
breakage for us, because we're trying to hold a reference to free as a
function pointer, and to do that we had an explicit type annotation.

We'd like to keep NIO compiling in Xcode 13 GM.

Modifications:

- Defined the free function as a thunk that passes through to the
  underlying OS free call, but takes its first argument as non-nullable.

Result:

Should compile on the Xcode 13 GM again.

Resolves #1955.

PR #1964, opened 2021-09-21
NIO1 Work around Xcode 13 GM SDK issues

baarde:
### Motivation:

This is a backport of #1956.

Xcode 13 GM shipped with a Swift overlay for libsystem in macOS that marked free's first argument as non-nullable. This leads to an awkward breakage for us, because we're trying to hold a reference to free as a function pointer, and to do that we had an explicit type annotation.

We'd like to keep NIO compiling in Xcode 13 GM.

### Modifications:

- Defined the free function as a thunk that passes through to the underlying OS free call, but takes its first argument as non-nullable.

### Result:

Should compile on the Xcode 13 GM again.


Lukasa:

Lukasa (approved):
Thanks for this!

PR #1966, opened 2021-09-22
Only include concurrency features on new OSes

Lukasa:
Motivation:

Swift 5.5 provides concurrency support on almost all platforms where it
is available. However, the Xcode 13 GM currently provides a macOS 11 SDK
with Swift 5.5, and the concurrency features are not available in that
SDK. As a result, NIO's concurrency features cause compile errors when
building the concurrency library on macOS using Xcode 13 GM.

Modifications:

- Only build the concurrency features when the concurrency library is
  present.

Result:

We can build NIO using Xcode 13 GM.

### DIFF
--- a/Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
+++ b/Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
@@ -15,7 +15,7 @@
 import NIOCore
 import NIOHTTP1
 
-#if compiler(>=5.5)
+#if compiler(>=5.5) && canImport(_Concurrency)
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 struct AsyncChannelIO<Request, Response> {
     let channel: Channel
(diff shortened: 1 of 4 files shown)


fabianfett (approved):
LGTM. Maybe change title to: Only use concurrency on new OSes where available?

Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
```diff
@@ -15,7 +15,7 @@
 import NIOCore
 import NIOHTTP1
 
-#if compiler(>=5.5)
+#if compiler(>=5.5) && canImport(_Concurrency)
```
weissi, line 18: Lgtm but we could also try `#if compiler(>=5.5) && $AsyncAwait`. That’s what the original version of this code had, no idea when we lost that. Tbh I don’t know if `$AsyncAwait` or `canImport(_Concurrency)` is more correct here. CC @dgregor

DougGregor: The `#if compiler(>=5.5) && canImport(_Concurrency)` is correct, because you need both the 5.5 compiler (all released versions and betas support async/await) and the _Concurrency module.

PR #1969, opened 2021-09-23
Move AsyncAwait helpers into NIOCore

fabianfett:
### Motivation

With the release of Swift 5.5 we should move our AsyncAwait helpers into our stable API.

### Changes

- Moved the `AsyncAwaitSupport.swift` to `NIOCore`
- Moved the `NIOCore.System.hasAsyncAwaitSupport` into the `AsyncAwaitSupport.swift` file

### Open questions

- Do we want to move those extensions next to their objects?

### Result

Async/await API is declared stable.


Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -215,3 +213,15 @@ extension ChannelPipeline {
...
 #endif
+
+#if compiler(>=5.5)
+fileprivate let hasAsyncAwait = true
+#else
+fileprivate let hasAsyncAwait = false
+#endif
+
+extension NIOCore.System {
+    public static var hasAsyncAwaitSupport: Bool {
```
Lukasa, line 224: Do we need to keep this?

Lukasa:
Note that nightly is failing due to missing sendable conformances. We need to add this, but it's not blocking us from moving forward for now.

Sources/NIOCore/AsyncAwaitSupport.swift
fabianfett: Removed

Lukasa (approved):

fabianfett:
`swift:nightly-main` still falling because of `Sendable` requirements.

PR #1977, opened 2021-10-15
Remove not on eventLoop precondition for NIOPipeBootstrap

jabwd:
Remove not on eventLoop precondition for NIOPipeBootstrap

### Motivation:

Currently you cannot create a NIOPipeBootstrap when you are on
an event loop due to a precondition check. This check seeks to prevent
consumers of the API from passing in a file descriptor thats referencing
a file on disk or on the network.
fstat could potentially block the current event loop especially if the file descriptor is referencing a file over the network.
This check however prevents certain use cases of SwiftNIO, where the downsides
of a consumer of this api blocking their own event loop do not weigh in against preventing
an entire use case from SwiftNIO.

### Modifications:

Removed the precondition

### Result:

After this change it will be possible to feed file descriptors into the bootstrap which potentially block the current event loop.
A potential (portable) replacement for fstat still has to be found in order to solve this problem completely.


Lukasa:

Lukasa:
Do you mind running `./scripts/generate_linux_tests.rb` and committing the resulting changes?

jabwd:
My bad, forgot about running that! Changes have been pushed 👍

Lukasa:

PR #1984, opened 2021-11-08
[HTTPDecoder] Decode informational http response head correctly

fabianfett:
We should support HTTP requests that are answered with an informational response first, which is followed up by the actual response. See issue: https://github.com/swift-server/async-http-client/issues/461

### Motivation:

- Currently requests that lead to a 1xx response are handled incorrectly

### Modifications:

- Add an `InformalResponseStrategy` enum that allows the user to specify whether http 1xx responses shall be forwarded or dropped. 
- Implement the necessary cases in `HTTPDecoder` and `BetterHTTPParser`

### Result:

We will be able to fix https://github.com/swift-server/async-http-client/issues/461

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -298,8 +298,7 @@ private class BetterHTTPParser {
                 return .error(HPE_UNKNOWN)
             }
             
-            if (HTTPResponseStatus.continue.code <= statusCode && statusCode < HTTPResponseStatus.ok.code)
-                && statusCode != HTTPResponseStatus.switchingProtocols.code {
+            if 100 <= statusCode && statusCode < 200 && statusCode != 101 {
                 // if the response status is in the range of 100..<200 but not 101 we don't want to
                 // pop the request method. The actual request head is expected with the next HTTP
                 // head.
@@ -482,10 +481,10 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
     // the actual state
     private let parser: BetterHTTPParser
     private let leftOverBytesStrategy: RemoveAfterUpgradeStrategy
-    private let informalResponseStrategy: InformalResponseStrategy
+    private let informationalResponseStrategy: NIOInformationalResponseStrategy
     private let kind: HTTPDecoderKind
     private var stopParsing = false // set on upgrade or HTTP version error
-    private var lastHeaderWasInformal = false
+    private var lastResponseHeaderWasInformational = false
 
     /// Creates a new instance of `HTTPDecoder`.
     ///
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -473,15 +482,36 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
...
     /// - parameters:
     ///     - leftOverBytesStrategy: The strategy to use when removing the decoder from the pipeline and an upgrade was,
     ///                              detected. Note that this does not affect what happens on EOF.
-    public init(leftOverBytesStrategy: RemoveAfterUpgradeStrategy = .dropBytes) {
+    public convenience init(leftOverBytesStrategy: RemoveAfterUpgradeStrategy = .dropBytes) {
+        self.init(leftOverBytesStrategy: leftOverBytesStrategy, informalResponseStrategy: .drop)
+    }
+    
+    /// Strategy to use when a HTTPDecoder receives an informal HTTP response (1xx except 101)
+    public enum InformalResponseStrategy {
```
Lukasa, line 500: Nit: these are not _informal_ responses, they are _informational_ responses. All the terminology in this PR should be changed appropriately.

Lukasa: Similarly, we should probably use an extensible enum for our API design here (i.e. `enum` wrapped in `struct` with `static let`s.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -293,16 +293,25 @@ private class BetterHTTPParser {
...
                 self.richerError = NIOHTTPDecoderError.unsolicitedResponse
                 return .error(HPE_UNKNOWN)
             }
-
-            if method == .HEAD || method == .CONNECT {
-                skipBody = true
-            } else if statusCode / 100 == 1 ||  // 1XX codes
-                statusCode == 204 || statusCode == 304 {
+            
+            if (HTTPResponseStatus.continue.code <= statusCode && statusCode < HTTPResponseStatus.ok.code)
```
Lukasa, line 301: I think it's fine to treat these codes as numbers here.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -473,15 +482,36 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
     // the actual state
     private let parser: BetterHTTPParser
     private let leftOverBytesStrategy: RemoveAfterUpgradeStrategy
+    private let informalResponseStrategy: InformalResponseStrategy
     private let kind: HTTPDecoderKind
     private var stopParsing = false // set on upgrade or HTTP version error
+    private var lastHeaderWasInformal = false
```
Lukasa, line 488: We don't really mean "header" here, we mean "response".

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -561,7 +592,23 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
...
+        case .response where (100..<200).contains(statusCode) && statusCode != 101:
+            self.lastHeaderWasInformal = true
+            
+            switch self.informalResponseStrategy {
+            case .forward:
+                let resHead: HTTPResponseHead = HTTPResponseHead(version: .init(major: versionMajor, minor: versionMinor),
+                                                                 status: .init(statusCode: statusCode),
+                                                                 headers: HTTPHeaders(self.headers,
+                                                                                      keepAliveState: keepAliveState))
+                message = NIOAny(HTTPClientResponsePart.head(resHead))
```
Lukasa, line 605: Can I propose we wrap the common code in this case and the case below into a single helper?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -473,15 +482,36 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
...
-    public init(leftOverBytesStrategy: RemoveAfterUpgradeStrategy = .dropBytes) {
+    public convenience init(leftOverBytesStrategy: RemoveAfterUpgradeStrategy = .dropBytes) {
+        self.init(leftOverBytesStrategy: leftOverBytesStrategy, informalResponseStrategy: .drop)
+    }
+    
+    /// Strategy to use when a HTTPDecoder receives an informal HTTP response (1xx except 101)
+    public enum InformalResponseStrategy {
+        /// Drop the informal response and only forward the "real" response
+        case drop
+        /// Forward the informal response and then forward the "real" response
```
Lukasa, line 503: We should clarify what this looks like (namely, multiple `.head` messages for a single `.end`).

Tests/NIOHTTP1Tests/HTTPDecoderTest.swift
```diff
@@ -792,6 +792,60 @@ class HTTPDecoderTest: XCTestCase {
         XCTAssertEqual(["channelReadComplete", "write", "flush", "channelRead", "errorCaught"], eventCounter.allTriggeredEvents())
         XCTAssertNoThrow(XCTAssertTrue(try channel.finish().isClean))
     }
+    
+    func testForwardContinueThanResponse() {
```
Lukasa, line 796: Minor nit: 

```suggestion
    func testForwardContinueThenResponse() {
```

Lukasa (approved):
Really nice. One minor nit, otherwise I’m happy.

PR #1985, opened 2021-11-09
Allow HTTP Server to send multiple informational heads before actual response head

fabianfett:
To make testing in AHC easier we should support informational responses on the server side as well. This lands the changes already proposed in #1330 and and #1422.

### Motivation:

Server developers want to issue informational responses.

### Result:

Proper server support for informational responses.


t089:
Thanks @fabianfett :)

Lukasa (approved):
Nice patch, LGTM.

PR #1987, opened 2021-11-18
ByteBuffer: provide multi read/write int methods

weissi:
Motivation:

Many network protocols (especially for example NFS) have quite a number
of integer values next to each other. In NIO, you'd normally parse/write
them with multiple read/writeInteger calls.

Unfortunately, that's a bit wasteful because we're checking the bounds
as well as the CoW state every time.

Modifications:

- Provide read/writeMultipleIntegers for up to 10 FixedWidthIntegers.
- Benchmarks

Result:

Faster code. For 10 UInt32s, this is a 5x performance win on my machine,
see benchmarks.


weissi:
The generated code (10 UInt32s) looks pretty awesome, lots of vector stuff haha.

Screenshot 2021-11-18 at 4 55 39 pm

weissi:
@swift-nio-bot perf test this please

weissi:
@Lukasa we should probably make this `@_alwaysEmitIntoClient` to not incur huge code sizes because of the (unnecessary) generic impls.

Sources/NIOCore/ByteBuffer-multi-int.swift
```diff
@@ -0,0 +1,1053 @@
...
+            assert(offset <= ptr.count)
+            return offset
+        }
+    }
+
+}
+
+
+/* --- AUTO-GENERATION SCRIPT FOR THIS FILE ---
+#!/bin/bash
```
glbrntt, line 904: As much as a I love script which generates itself, can move this to its own file?

Sources/NIOCore/ByteBuffer-multi-int.swift
```diff
@@ -0,0 +1,1053 @@
...
+            (basePtr + offset).copyMemory(from: &v9, byteCount: MemoryLayout<T9>.size)
+            offset = offset &+ MemoryLayout<T9>.size
+            assert(offset <= ptr.count)
+            return offset
+        }
+    }
+
+    @inlinable
+    public mutating func readMultipleIntegers<T1: FixedWidthInteger, T2: FixedWidthInteger, T3: FixedWidthInteger, T4: FixedWidthInteger, T5: FixedWidthInteger, T6: FixedWidthInteger, T7: FixedWidthInteger, T8: FixedWidthInteger, T9: FixedWidthInteger, T10: FixedWidthInteger>(endianness: Endianness = .big, as: (T1, T2, T3, T4, T5, T6, T7, T8, T9, T10).Type = (T1, T2, T3, T4, T5, T6, T7, T8, T9, T10).self) -> (T1, T2, T3, T4, T5, T6, T7, T8, T9, T10)? {
+        let bytesRequired = Int(MemoryLayout<T1>.size) &+ Int(MemoryLayout<T2>.size) &+ Int(MemoryLayout<T3>.size) &+ Int(MemoryLayout<T4>.size) &+ Int(MemoryLayout<T5>.size) &+ Int(MemoryLayout<T6>.size) &+ Int(MemoryLayout<T7>.size) &+ Int(MemoryLayout<T8>.size) &+ Int(MemoryLayout<T9>.size) &+ Int(MemoryLayout<T10>.size)
```
glbrntt, line 764: The compiler is unhappy with how long it takes to type check this, can we split this up and re-use the sizes too?

Sources/NIOPerformanceTester/ByteBufferWriteMultipleBenchmarks.swift
```diff
@@ -0,0 +1,72 @@
+import NIOCore
```
glbrntt, line 1: License header?

glbrntt (commented):
Left a couple of nits but otherwise looks good and I think this is a useful change, thanks jw!

Sources/NIOCore/ByteBuffer-multi-int.swift
weissi: I can do that too, where would you like it to live?

Sources/NIOCore/ByteBuffer-multi-int.swift
weissi: I think it’ll work with regular +s and given that this is compile-time evaluated we don’t incur overhead for overflow checks

Sources/NIOPerformanceTester/ByteBufferWriteMultipleBenchmarks.swift
weissi: Ta!

Sources/NIOCore/ByteBuffer-multi-int.swift
glbrntt: `scripts` should be fine

Sources/NIOCore/ByteBuffer-multi-int.swift
dnadoba: I think the compiler can type check it faster if each addition is a separate expressions e.g.:
```
var bytesRequired = 0
bytesRequired &+= Int(MemoryLayout<T1>.size)
bytesRequired &+= Int(MemoryLayout<T2>.size)
bytesRequired &+= Int(MemoryLayout<T3>.size)
...
```
This should result in linear time complexity for type checking because the compiler can find the right overload for just one `&+=` without looking at the other expressions.

Wrapping all calls to `.size` in another `Int` should also not be needed here because [the return type of `MemoryLayout<T*>.size` is already `Int`](https://developer.apple.com/documentation/swift/memorylayout/2432227-size). After that we can simplify it even a bit further:
```
var bytesRequired = 0
bytesRequired &+= MemoryLayout<T1>.size
bytesRequired &+= MemoryLayout<T2>.size
bytesRequired &+= MemoryLayout<T3>.size
...
```

weissi:

weissi:

Sources/NIOCore/ByteBuffer-multi-int.swift
```diff
@@ -0,0 +1,2393 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// NOTE: THIS FILE IS AUTO-GENERATED BY THE SCRIPT FOUND AT THE END OF THE FILE
+
+extension ByteBuffer {
+    @inlinable
+    @_alwaysEmitIntoClient
```
weissi, line 19: I added the `@_alwaysEmitIntoClient` now because otherwise we'll add A LOT of binary size for the unused generic versions of this. We only want to have the specialised ones.

weissi:
here's the perf proof:

Screenshot 2021-11-19 at 11 03 16 am

fabianfett:
Nice, this will unlock some nice perf gains in postgres :)

Sources/NIOCore/ByteBuffer-multi-int.swift
weissi: @glbrntt done, it's its own script now.

dev/generate-bytebuffer-multi-int.sh
```diff
@@ -0,0 +1,167 @@
...
+    done
+    echo
+    echo "        return self.writeWithUnsafeMutableBytes(minimumWritableBytes: spaceNeeded) { ptr -> Int in"
+    echo "            var offset = 0"
+    echo "            let basePtr = ptr.baseAddress! // safe: pointer is non zero length"
+    for n in $(seq 1 "$how_many"); do
+        echo "            (basePtr + offset).copyMemory(from: &v$n, byteCount: MemoryLayout<T$n>.size)"
+        echo "            offset = offset &+ MemoryLayout<T$n>.size"
+    done
+    echo "            assert(offset <= ptr.count)"
```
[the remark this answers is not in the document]
weissi: @Lukasa thanks, I now added `assert`s after each `offset` increment. Is this what you were looking for?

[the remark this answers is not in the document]
Lukasa: Oh, I'm sorry, I should have been clearer. We know how big this has to be, because we have a `minimumWritableBytes` argument above. We should assert that the pointer has that much space before we start writing to it.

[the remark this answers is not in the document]
weissi: okay, sounds good! will do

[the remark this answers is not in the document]
weissi: @Lukasa done: I now asserted that there's enough space. And additionally that `offset` at the end matches `bytesRequired`/`spaceNeeded`. Those should line up exactly.

Lukasa (approved):

dnadoba (approved):
LGTM. A good example for what we might use [TypeSequences aka Variadic Generics](https://forums.swift.org/t/pitching-the-start-of-variadic-generics/51467) if we get the same performance out of it. Maybe worth posting this use case on the pitch thread.

weissi:
> LGTM. A good example for what we might use [TypeSequences aka Variadic Generics](https://forums.swift.org/t/pitching-the-start-of-variadic-generics/51467) if we get the same performance out of it. Maybe worth posting this use case on the pitch thread.

We could, however the variadics would only really if the types are all the same. Ie `(UInt32, UInt32,  UInt32, ...)` and not mixed like supported by this patch (eg. `(UInt32, UInt16, UInt64, ...)`)

weissi:
> @weissi Not sure I can follow you here, the pitch for TypeSequences contains this code example:
> 
> ```swift
> func debugPrint<T...>(_ items: T...) 

Ah yes, I missed the ellipsis in the generic arguments. That's cool if this would be supported, very nice!

Issue #1991, opened 2021-11-23
MessageToByteHandler conformance to RemovableHandler


AlexisQapa:
### Expected behavior
MessageToByteHandler should conform to RemovableHandler

### Actual behavior
MessageToByteHandler doesn't conform to RemovableHandler

Is there a reason for MessageToByteHandler to not conform to RemovableHandler ? 

### SwiftNIO version/commit hash
SwiftNIO 2.34.0

### Swift & OS version (output of `swift --version && uname -a`)
swift-driver version: 1.26.9 Apple Swift version 5.5.1 (swiftlang-1300.0.31.4 clang-1300.0.29.6)

Lukasa:
I agree, it should! There's no reason that support hasn't been added, it just wasn't high up on anyone's priority list. We'd happily accept a patch that adds the conformance.

AlexisQapa:
I'll try do add it then. Is there any gotcha I should be aware of ? ByteToMessagehander conformance seams quite complicated but some other conformances I read don't do anything.

Lukasa:
I don't think there is any particular gotcha _except_ that we should allow implementers to intercept removal and do something with it.

AlexisQapa:
Implementers ? MessageToByteHandler is a final class. Do you mean to send event to decoders ? ByteToMessageHandler doesn't seem to notify it's encoder while being removed.

Lukasa:
ByteToMessageHandler _does_ notify. Here's the conformance:

https://github.com/apple/swift-nio/blob/a58c36c55ce7b244a1b5dcec8c451ab06df3dc8c/Sources/NIOCore/Codec.swift#L720-L741

Note that this calls `processLeftovers`:

https://github.com/apple/swift-nio/blob/a58c36c55ce7b244a1b5dcec8c451ab06df3dc8c/Sources/NIOCore/Codec.swift#L543-L560

This calls `decodeLoop` with `decodeMode` set to `.last`, which will call `decodeLast` on the `ByteToMessageDecoder`:

https://github.com/apple/swift-nio/blob/a58c36c55ce7b244a1b5dcec8c451ab06df3dc8c/Sources/NIOCore/Codec.swift#L569-L603

We need a similar pattern: we need a new method on `MessageToByteEncoder` that `MessageToByteHandler` will call when it is removed from the pipeline.

PR #1993, opened 2021-11-23
Add RemovableChannelHandler conformance to MessageToByteHandler

AlexisQapa:
Add RemovableChannelHandler conformance to MessageToByteHandler as a follow up for #1991

### Motivation:

RemovableChannelHandler conformance was missing MessageToByteHandler while ByteToMessageHandler did conform to it.

### Modifications:

Add RemovableChannelHandler conformance to ByteToMessageHandler
Use the same pattern as for ByteToMessage with hooks for encoders to act on removal.
Added hooks have default implementation.
```swift
func encodeLast(data: OutboundIn, out: inout ByteBuffer) throws
mutating func encoderRemoved(context: ChannelHandlerContext)
mutating func encoderAdded(context: ChannelHandlerContext)
```

### Result:

MessageToByteHandler now conform to RemovableChannelHandler.

### DIFF
--- a/Sources/NIOCore/Codec.swift
+++ b/Sources/NIOCore/Codec.swift
@@ -63,7 +63,7 @@ extension ByteToMessageDecoderError {
 /// ### Implementing ByteToMessageDecoder
 ///
 /// A type that implements `ByteToMessageDecoder` may implement two methods: decode and decodeLast. Implementations
-/// must implement decode: if they do not implement decodeLast, a default implementation will be used that 
+/// must implement decode: if they do not implement decodeLast, a default implementation will be used that
 /// simply calls decode.
 ///
 /// `decode` is the main decoding method, and is the one that will be called most often. `decode` is invoked
@@ -238,7 +238,7 @@ extension ByteToMessageDecoder {
     public func wrapInboundOut(_ value: InboundOut) -> NIOAny {
         return NIOAny(value)
     }
-    
+
     public mutating func decodeLast(context: ChannelHandlerContext, buffer: inout ByteBuffer, seenEOF: Bool) throws  -> DecodingState {
         while try self.decode(context: context, buffer: &buffer) == .continue {}
         return .needMoreData
@@ -703,20 +703,7 @@ extension ByteToMessageHandler: ChannelOutboundHandler, _ChannelOutboundHandler 
     }
 }
 
-/// A protocol for straightforward encoders which encode custom messages to `ByteBuffer`s.
-/// To add a `MessageToByteEncoder` to a `ChannelPipeline`, use
-/// `channel.pipeline.addHandler(MessageToByteHandler(myEncoder)`.
-public protocol MessageToByteEncoder {
-    associatedtype OutboundIn
-
-    /// Called once there is data to encode.
-    ///
-    /// - parameters:
-    ///     - data: The data to encode into a `ByteBuffer`.
-    ///     - out: The `ByteBuffer` into which we want to encode.
-    func encode(data: OutboundIn, out: inout ByteBuffer) throws
-}
-
+// MARK: ByteToMessageHandler: RemovableChannelHandler
 extension ByteToMessageHandler: RemovableChannelHandler {
     public func removeHandler(context: ChannelHandlerContext, removalToken: ChannelHandlerContext.RemovalToken) {
         precondition(self.removalState == .notBeingRemoved)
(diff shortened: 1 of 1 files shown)


AlexisQapa:
I must say I'm not sure about my implementation as I'm quite new to the code base. I'll look into tests when I can thought any advice about how to would be appreciated. Should this be a draft PR ?

Sources/NIOCore/Codec.swift
```diff
@@ -773,15 +844,27 @@ public final class MessageToByteHandler<Encoder: MessageToByteEncoder>: ChannelO
...
+        self.encoder.encoderAdded(context: context)
     }
 
     public func handlerRemoved(context: ChannelHandlerContext) {
+        self.removalState = .handlerRemovedCalled
+        if !self.state.isDone {
+            self.state = .done
+        }
+
         self.state = .done
```
AlexisQapa, line 864: Yeah I think its time to rest

Sources/NIOCore/Codec.swift
```diff
@@ -740,12 +727,77 @@ extension ByteToMessageHandler: RemovableChannelHandler {
...
+    ///
+    /// - parameters:
+    ///     - context: The `ChannelHandlerContext` which this `ByteToMessageDecoder` belongs to.
+    mutating func encoderRemoved(context: ChannelHandlerContext)
+
+    /// Called when this `MessageToByteEncoder` is added to the `ChannelPipeline`.
+    ///
+    /// - parameters:
+    ///     - context: The `ChannelHandlerContext` which this `MessageToByteEncoder` belongs to.
+    mutating func encoderAdded(context: ChannelHandlerContext)
```
Lukasa, line 760: I don't think either of these methods need the `ChannelHandlerContext`: we don't currently expose this to `MessageToByteEncoder`s anyway.

Sources/NIOCore/Codec.swift
```diff
@@ -740,12 +727,77 @@ extension ByteToMessageHandler: RemovableChannelHandler {
...
+        /// removal is triggered by Channel teardown.
+        case removalStarted
+
+        /// The user-triggered removal is complete. This state will not be entered if the removal is triggered by
+        /// Channel teardown.
+        case removalCompleted
+
+        /// This handler has been removed from the pipeline.
+        case handlerRemovedCalled
+    }
```
Lukasa, line 799: I don't think we need this complexity.

The complexity in `ByteToMessageEncoder` is present because of the fact that the encoder gets access to the `ChannelHandlerContext`, and so can trigger arbitrary re-entrant calls. `MessageToByteEncoder` does not allow this: you can only write into the `ByteBuffer`. So in this case, we can skip defending against re-entrancy.

As a result, we can remove almost all of this: removal can be done essentially immediately, with only the need to notify the encoder and give it an opportunity to write a final message. Does this make sense?

AlexisQapa: Yeah that's mostly what I thought as there is not pending buffer. Will update ASAP.

AlexisQapa:
I could turn the RemovalState to a Bool but I'm unsure if it then needs a precondition in handlerAdded.

Lukasa:
You shouldn't need the precondition in handlerAdded.

Sources/NIOCore/Codec.swift
```diff
@@ -812,3 +857,21 @@ extension MessageToByteHandler {
...
+            assert(!self.state.isOperational, "illegal state: \(self.state)")
+            if self.isBeingRemoved {
+                self.isBeingRemoved = false
+            } else {
+                assertionFailure("illegal removal state: \(self.isBeingRemoved)")
+            }
+            // this is necessary as it'll complete the promise.
+            context.leavePipeline(removalToken: removalToken)
+        }
+    }
```
Lukasa, line 876: I think we can simplify this still further.

`isBeingRemoved` and the `eventLoop.execute` call both exist to try to unwind the call stack and avoid Law of Exclusivity violations. We won't have to do that here: because we don't make it possible to perform outcalls from the MessageToByteEncoder it's not likely that they will happen.

As a result, we can avoid the dance with setting `isBeingRemoved` and then calling `encodeLast` from `write`. Instead, we can simply drop the `data:` argument from `encodeLast` and synchronously call it here, sending any data, before synchronously calling `leavePipeline`.

Does that make sense?

AlexisQapa: Should encodeLast still get passed the buffer and/or be throwing ?
Since decodeLast will be called without data should it's default implementation do nothing ?

Lukasa: Yes, it should be passed the buffer and be throwing.

Yes, the default implementation should do nothing.

AlexisQapa: Hi, I'm not sure if you saw my last commit and/or if there anything you want me to change. 
As recommended here and on the forum I refactored my code to use state instead of removing my handlers but if you are still interested into merging this I could do any requested changes.

Lukasa: Sorry for the delay, I was on vacation.

Sources/NIOCore/Codec.swift
```diff
@@ -760,10 +786,19 @@ public final class MessageToByteHandler<Encoder: MessageToByteEncoder>: ChannelO
...
         }
+
+        var isOperational: Bool {
+            switch self {
+            case .operational:
+                return true
+            case .notInChannelYet, .error, .done:
+                return false
+            }
+        }
```
Lukasa, line 797: I don't think we need this any longer, we don't use it.

Sources/NIOCore/Codec.swift
```diff
@@ -812,3 +849,18 @@ extension MessageToByteHandler {
         }
     }
 }
+
+// MARK: ByteToMessageHandler: RemovableChannelHandler
+extension MessageToByteHandler: RemovableChannelHandler {
+    public func removeHandler(context: ChannelHandlerContext, removalToken: ChannelHandlerContext.RemovalToken) {
+        do {
```
Lukasa, line 856: I think we need the state checking that we do in `func write` here.

Sources/NIOCore/Codec.swift
```diff
@@ -853,14 +844,25 @@ extension MessageToByteHandler {
...
-        do {
-            self.buffer!.clear()
-            try self.encoder.encodeLast(out: &self.buffer!)
-            context.write(self.wrapOutboundOut(self.buffer!), promise: nil)
-        } catch {
-            self.state = .error(error)
+        switch self.state {
+        case .notInChannelYet:
+            preconditionFailure("MessageToByteHandler.removeHandler called before it was added to a Channel")
+        case .error(let error):
```
AlexisQapa, line 850: I'm unsure about this, should I still remove the handler ?

Lukasa: Yes.

Lukasa (commented):
Great, this looks really good! One final note: can you add a few unit tests that test the paths through the removal flow here? That way we won't accidentally regress this behaviour.

PR #1994, opened 2021-11-27
benchmarks: lock performance for 1, 2, 4, 8 threads wanting lock

weissi:
Motivation:

To judge the cost of `PTHREAD_MUTEX_ERRORCHECK` as well as
`os_unfair_lock` vs `pthread_mutex_t` it's useful to have a few simple
benchmarks.

Modification:

Add lock benchmarks.

Result:

Better data.

PR #1995, opened 2021-11-27
disable PTHREAD_MUTEX_ERRORCHECK in release mode

weissi:
Motivation:

`pthread_mutex_t` locks can do advanced error checking with
`PTHREAD_MUTEX_ERRORCHECK`, the cost of which was unknown. With #1994 I
could now measure it.

The differences from my local machine (M1 MBP) for uncontended locks
are:

- on Linux: 10% faster without `PTHREAD_MUTEX_ERRORCHECK`
- on macOS: 25% faster without `PTHREAD_MUTEX_ERRORCHECK`!

Modification:

Do the extended error checking only in debug mode. Note that we still
`precondition` on all pthread return values (which is basically free).

Result:

Faster code.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -43,7 +43,9 @@ public final class Lock {
 #else
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr)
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

Lukasa:

PR #1998, opened 2021-12-02
remove/deprecate the file:line: parameters from flatMap and friends

### LINKED ISSUE #1997: remove/deprecate the `file:line:` parameters from `flatMap` and friends
For NIO's 'promise leak detector' we added `file:line:` labels to `makePromise` and there it makes sense. A user might create a promise and then never fulfil it, bad. With the `file:line:` arguments we can give good diagnostics.

However, we (probably that @weissi again) also added it to `flatMap` and friends where it doesn't make sense at all.

Sure, EventLoopFuture's implementation may create a promise in the implementation of `flatMap` but this promise is never leaked unless the previous future is never fulfilled (or NIO has a terrible bug). Suffice to say that in a future chain, it's never a `flatMap` etc which is responsible for leaking the first promise...

weissi:
Motivation:

For NIO's 'promise leak detector' we added file:line: labels to
makePromise and there it makes sense. A user might create a promise and
then never fulfil it, bad. With the file:line: arguments we can give
good diagnostics.

However, we (probably that @weissi again) also added it to flatMap and
friends where it doesn't make sense at all.

Sure, EventLoopFuture's implementation may create a promise in the
implementation of flatMap but this promise is never leaked unless the
previous future is never fulfilled (or NIO has a terrible bug). Suffice
to say that in a future chain, it's never a flatMap etc which is
responsible for leaking the first promise...

Explain here the context, and why you're making that change.
What is the problem you're trying to solve.

Modifications:

Remove all unnecessary `file:line:` parameters whilst keeping the public
API intact.

Result:

- More sensible code.
- fixes #1997

### DIFF
--- a/Sources/NIOCore/EventLoop+Deprecated.swift
+++ b/Sources/NIOCore/EventLoop+Deprecated.swift
@@ -1,0 +1,27 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension EventLoop {
+    @inlinable
+    @available(*, deprecated, message: "Please don't pass file:line:, there's no point.")
+    public func makeFailedFuture<T>(_ error: Error, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<T> {
+        return self.makeFailedFuture(error)
+    }
+
+    @inlinable
+    @available(*, deprecated, message: "Please don't pass file:line:, there's no point.")
+    public func makeSucceededFuture<Success>(_ value: Success, file: StaticString = #file, line: UInt = #line) -> EventLoopFuture<Success> {
+        return self.makeSucceededFuture(value)
+    }
+}
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -156,6 +156,13 @@ public struct EventLoopPromise<Value> {
     /// `EventLoopPromise` is completed.
     public let futureResult: EventLoopFuture<Value>
 
+    @inlinable
+    internal static func makeUnleakablePromise(eventLoop: EventLoop, line: UInt = #line) -> EventLoopPromise<Value> {
+        return EventLoopPromise<Value>(eventLoop: eventLoop,
+                                       file: "BUG in SwiftNIO (please report), unleakable promise leaked.",
+                                       line: line)
+    }
+
     /// General initializer
     ///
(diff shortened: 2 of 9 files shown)


Sources/NIOCore/EventLoop+Deprecated.swift
```diff
@@ -0,0 +1,29 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension EventLoop {
+    @inlinable
+    @_disfavoredOverload
```
Lukasa, line 17: I don't think `_disfavoredOverload` should be necessary: the Swift ambiguity rules should prefer the "more specific" implementations, which don't have `file:line:` defaulted.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -386,31 +386,33 @@ public final class EventLoopFuture<Value> {
...
+    internal convenience init(eventLoop: EventLoop, value: Value) {
+        self.init(_eventLoop: eventLoop, value: .success(value))
     }
 
     /// A EventLoopFuture<Value> that has already failed
     @inlinable
-    internal convenience init(eventLoop: EventLoop, error: Error, file: StaticString, line: UInt) {
-        self.init(_eventLoop: eventLoop, value: .failure(error), file: file, line: line)
+    internal convenience init(eventLoop: EventLoop, error: Error) {
+        self.init(_eventLoop: eventLoop, value: .failure(error))
```
Lukasa, line 415: Can I propose that instead of using a `convenience init` here we make these proper `init`s? That will reduce the temptation to use the init that doesn't do file and line storage, as you either have a value up front or you don't.

weissi:
thanks @Lukasa all done

Lukasa (approved):

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -163,7 +163,7 @@ public struct EventLoopPromise<Value> {
     ///     - file: The file this promise was allocated in, for debugging purposes.
     ///     - line: The line this promise was allocated on, for debugging purposes.
     @inlinable
-    internal init(eventLoop: EventLoop, file: StaticString, line: UInt) {
+    internal init(eventLoop: EventLoop, file: StaticString = "cannot leak or bug in NIO", line: UInt = #line) {
```
fabianfett, line 166: I assume you set this string, so that you don't always need to pass the same string, when creating a Promise. I assume it is only used in `makePromise` from here on. I think we should add a code comment here, explaining the string and when it is important to overwrite.

fabianfett (approved):
Very cool! This will make code completion much nicer! Would you mind adding one comment explaining the magic string?

Sources/NIOCore/EventLoopFuture.swift
weissi: @fabianfett PTAL, I made the code cleaner I think. @Lukasa too

Lukasa (approved):

fabianfett (approved):
🥳 I like this even more! Thanks!

PR #2002, opened 2021-12-03
Don't tolerate EBADF on close.

### LINKED ISSUE #2001: EBADF relaxation too much
@Lukasa / @John-Connolly I think #1999 relaxed the `EBADF` precondition too aggressively. We should still be checking `EBADF` on `close` on iOS. Because if we get one of these "defuncted" sockets, then `accept` etc will fail will `EBADF` but the `close` should still work.

I did speak to the kernel team and yes, it turns out that `EBADF` is expected here (but still needs a `close`) and whilst that's arguably the wrong errno returned by the kernel (because it's not a bad file descriptor) this cannot be fixed anymore because it would break binary compatibility. It wasn't clear why `EBADF` was chosen to start with  but here we are... So yeah, on `iOS`, `tvOS` and probably `watchOS`, we should allow `EBADF` for anything but `close`. The good thing is that we'll then still catch actual `EBADF`s on iOS, just slightly later.

Lukasa:
Motivation:

It's like that #1999 was a little wide, and we should have forbidden
EBDAF on close.

Modifications:

Still forbid EBADF on close.

Result:

EBADF still preconditions on close.
Fixes #2001.

### DIFF
--- a/Sources/NIOPosix/System.swift
+++ b/Sources/NIOPosix/System.swift
@@ -129,7 +129,7 @@ private func isUnacceptableErrno(_ code: Int32) -> Bool {
     // is valid but the accepted one is not. The right solution here is to perform a check for
     // SO_ISDEFUNCT when we see this happen, but we haven't yet invested the time to do that.
     // In the meantime, we just tolerate EBADF on iOS.
-    #if os(iOS)
+    #if os(iOS) || os(watchOS) || os(tvOS)
     switch code {
     case EFAULT:
         return true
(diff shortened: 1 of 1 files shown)


weissi (approved):

weissi:
Thank you @Lukasa

weissi:
@Lukasa/@john-connolly turns out there is a `SO_ISDEFUNCT` `getsockopt` which we use to query if the socket is actually `EBADF` or just defunct. So we could change the code on iOS to something to check `SO_ISDEFUNCT` on iOS if we receive an `EBADF`.

Lukasa:
I added a comment to the check that mention this. However, I don't think it's that simple: in the case of accept we'd have to check EBADF on the _listening_ socket, not the one that hit the error. Actually doing this makes the code quite hard, I think, and we get most of the improvement from this version of the patch, so I consider it a future enhancement.

weissi:
@Lukasa works for me!

weissi (approved):
thank you!

PR #2003, opened 2021-12-03
provide EventLoopGroup.any() which is sticky to the current EventLoop

weissi:
Motivation:

A lot of libraries that use SwiftNIO don't allow/require the user to
specify what `EventLoop`s a certain function runs on. Something like

```
myHTTPClient.get("https://example.com) -> EventLoopFuture<...>
```

Internally `MyHTTPClient` has not much choice but using the
`EventLoopGroup.next()` method to obtain an `EventLoop` on which to
create the `EventLoopFuture` (that is returned).

This all works fine but unfortunately it usually forces a thread switch
which is most of the time avoidable if we're already running on an
`EventLoop`.

Modifications:

Provide an `EventLoopGroup.any()` method which can be used like so:

```swift
func get(_ url: String) -> EventLoopFuture<Response> {
    let promise = self.group.any().makePromise(of: Response.self)

    [...]

    return promise.futureResult
}
```

`EventLoopGroup.any()` very much works like `EventLoopGroup.next()`
except that it tries -- if possible -- to return the _current_
`EventLoop`.

Note that this means that `any()` is _not_ the right solution if you
want to load balance. Likely, everything will now stay on the same
`EventLoop`.

Result:

Fewer thread switches.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -678,6 +678,13 @@ extension EventLoop {
         return self
     }
 
+    /// An `EventLoop` forms a singular `EventLoopGroup`, returning itself as 'any' `EventLoop`.
+    ///
+    /// - returns: Itself, because an `EventLoop` forms a singular `EventLoopGroup`.
+    public func any() -> EventLoop {
+        return self
+    }
+
     /// Close this `EventLoop`.
     public func close() throws {
         // Do nothing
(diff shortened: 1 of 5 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -777,12 +784,39 @@ extension EventLoop {
 
 /// Provides an endless stream of `EventLoop`s to use.
 public protocol EventLoopGroup: AnyObject {
-    /// Returns the next `EventLoop` to use.
+    /// Returns the next `EventLoop` to use, this is useful for load balacing.
```
Lukasa, line 787:
```suggestion
    /// Returns the next `EventLoop` to use, this is useful for load balancing.
```

Lukasa (changes requested):

Lukasa (approved):

weissi:
API breakage checker fails for the known limitation that it doesn't recognise new protocol requirements that do however come with default implementations:

```
11:45:49 /* Protocol Requirement Change */
11:45:49 Func EventLoopGroup.any() has been added as a protocol requirement
```

Please do have a good look however that I didn't mess up.

dnadoba (approved):
Awesome! This will allow AsyncHTTPClient to automatically prefer the current EventLoop if it needs to start a new connection by only changing one line:
https://github.com/swift-server/async-http-client/blob/5ce7377a8adf3226ee937f8c5aa48543445d58b8/Sources/AsyncHTTPClient/HTTPClient.swift#L516

Lukasa (approved):

fabianfett (approved):
This is very, very cool!

Issue #2000, opened 2021-12-02
Change Swift Concurrency Availability to macOS 10.15, iOS 13, tvOS 13, and watchOS 6


PSchmiedmayer:
### Expected behavior
With the support for Swift Concurrency coming to macOS 10.15, iOS 13, tvOS 13, and watchOS 6 with Xcode 13.2, the Swift Concurrency features should be available when compiling for these targets using Xcode 13.2.

### Actual behavior
The Swift Concurrency features are marked with `@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)` and can therefore not be used on macOS 10.15, iOS 13, tvOS 13, and watchOS 6.

### Steps to reproduce

1. Change the source code of the NIOAsyncAwaitDemo on Xcode 13.2 and change the `@available` annotations to macOS 10.15, iOS 13, tvOS 13, and watchOS 6.
2. The Swift compiler throws compiler errors due to the `@available` in the Swift NIO targets.

### Solution
A simple solution would be to change the `@available` annotations to `@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)` as demonstrated at https://github.com/PSchmiedmayer/swift-nio.
Unfortunately, this breaks builds on the latest stable Xcode version (13.1) due to compiler errors disallowing Swift Concurrency features previous to macOS 12, which was the state when Xcode 13.1 was released.
This behavior is demonstrated by the GitHub Actions that I have added to https://github.com/PSchmiedmayer/swift-nio: https://github.com/PSchmiedmayer/swift-nio/actions/runs/1532925179 (latest == Xcode 13.2, latest_stable == Xcode 13.1)

I am not sure how this could be easily solved. I would be happy to receive any suggestions to successfully create a PR to change the Swift Concurrency availability to macOS 10.15, iOS 13, tvOS 13, and watchOS 6 with Xcode 13.2.

### SwiftNIO version/commit hash

[d08e8ff938bcfa339f47677716d21f531fa493c3](https://github.com/apple/swift-nio/commit/d08e8ff938bcfa339f47677716d21f531fa493c3)

### Swift & OS version (output of `swift --version && uname -a`)

swift-driver version: 1.26.21 Apple Swift version 5.5.2 (swiftlang-1300.0.47.2 clang-1300.0.29.30)
Target: x86_64-apple-macosx12.0
[...]

PR #2004, opened 2021-12-03
Adds Swift Concurrency Availability to macOS 10.15, iOS 13, tvOS 13, and watchOS 6

PSchmiedmayer:
Adds Swift Concurrency Availability to macOS 10.15, iOS 13, tvOS 13, and watchOS 6

### Motivation:

This PR is related to #2000.
With the support for Swift Concurrency coming to macOS 10.15, iOS 13, tvOS 13, and watchOS 6 with Xcode 13.2, the Swift Concurrency features should be available when compiling for these targets using Xcode 13.2.

### Modifications:

Change the @available annotations to macOS 10.15, iOS 13, tvOS 13, and watchOS 6 and introduce a temporary code duplication as discussed in #2000.
Changes the source code of the NIOAsyncAwaitDemo to only work on Swift 5.2.2 (Xcode 13.2) on macOS 10.15 and newer and all other non-Darwin platforms.

### Result:

The Swift Concurrency features are marked with `@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)` and can therefore be used on macOS 10.15, iOS 13, tvOS 13, and watchOS 6.


Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
```diff
@@ -15,8 +15,8 @@
 import NIOCore
 import NIOHTTP1
 
-#if compiler(>=5.5) && canImport(_Concurrency)
-@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+#if canImport(_Concurrency) && compiler(>=5.5.2) || (!os(macOS) && compiler(>=5.5))
```
Lukasa, line 18: The `os` guards here are insufficient to cover the use-case (we'd need to check for all Apple platforms), but I don't think it matters. On non-Apple platforms the two function implementations are totally fine, so we can just drop the `||` here.

Sources/NIOAsyncAwaitDemo/main.swift
```diff
@@ -16,9 +16,9 @@ import NIOPosix
 import NIOHTTP1
 import Dispatch
 
-#if compiler(>=5.5) && canImport(_Concurrency)
+#if canImport(_Concurrency) && compiler(>=5.5.2) || (!os(macOS) && compiler(>=5.5))
```
Lukasa, line 19: Similarly, we can drop the `||` here.

Sources/NIOCore/AsyncAwaitSupport+OldXcodes.swift
```diff
@@ -0,0 +1,215 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if os(macOS) && compiler(>=5.5) && !compiler(>=5.5.2)
```
Lukasa, line 15: And we can drop the `os` guard here.

Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
PSchmiedmayer: Good point, I have updated the PR to incorporate your feedback 👍

Lukasa:

Lukasa (approved):

Lukasa:
Thanks for the patch!

PSchmiedmayer:
Happy to help! Thanks for your fast response and your input @Lukasa & @weissi!

PR #2006, opened 2021-12-09
Enable io_uring builds with newer liburing

Lukasa:
Motivation:

Newer copies of liburing have changed header imports, making some
of our implicit includes fail, and have also changed the types
of some methods in a way that would only produce warnings in C but
which produces errors in Swift. Additionally, liburing broke when
we did the module split, but we didn't notice.

Modifications:

- Import NIOCore for the Uring Swift code.
- Import poll.h in CNIOLinux
- Support arguments that are either pointers or UInt64s.

Results:

liburing code compiles again

### DIFF
--- a/Sources/CNIOLinux/include/CNIOLinux.h
+++ b/Sources/CNIOLinux/include/CNIOLinux.h
@@ -15,6 +15,7 @@
 #define C_NIO_LINUX_H
 
 #ifdef __linux__
+#include <poll.h>
 #include <sys/epoll.h>
 #include <sys/eventfd.h>
 #include <sys/timerfd.h>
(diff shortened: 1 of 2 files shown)


Lukasa:
cc @hassila

hassila:
Thanks @Lukasa - been busy times unfortunately, but hope we can pick up discussions about how to move io_uring forward to the next level (including memory ownership discussions and in light of support for Concurrency in NIO) if you / the team have bandwidth early next year perhaps?

fabianfett (approved):
Awesome `#include <pool.h>` is much smarter than what I did locally.

Lukasa:
@hassila For sure, we're happy to sync up whenever. I mostly wanted to let you review this given that you wrote the original code! :grin:

hassila:
> @hassila For sure, we're happy to sync up whenever. I mostly wanted to let you review this given that you wrote the original code! 😁

Thanks, LGTM!

PR #2009, opened 2021-12-13
Add baseline performance and allocation tests for scheduleTask and execute

FranzBusch:
### Motivation:

In issue https://github.com/apple/swift-nio/issues/1316, we see a large number of allocations to happen when scheduling tasks. This can definitely be optimized. This PR adds a number of baseline allocation and performance tests for both `scheduleTask` and `execute`. In the next PRs, I am going to try a few optimizations to reduce the number of allocations.

### Modifications:

Added baseline performance and allocation tests for `scheduleTask` and `execute`

Issue #1541, opened 2020-06-02
Ordering of tasks scheduled with the same deadline on `SelectableEventLoop` is undefined


glbrntt:
If two tasks are scheduled on an event loop with the same deadline then the order in which they fire is not guaranteed. This is unlikely to happen in the real world but it is possible. On `EmbeddedEventLoop` it is far more likely to happen as time is controlled. 

We should update `SelectableEventLoop` such that scheduled tasks are ordered by their deadline and ties are broken on the order in which they scheduled. `EmbeddedEventLoop` has already had this treatment.

See also: https://github.com/apple/swift-nio/pull/1540

weissi:
We have to fix this. In the old days where we had only _in N nanoseconds_ API, this was probably okay(ish) because it's impossible to get a clash in the real world (because you'd need to get to equal values back from `DispatchTime.now()` in a row. Even in a tight loop, that's not possible with the time resolution of today's CPUs.

Just as a quick test:

```swift
import Dispatch

func foo() {
    var minDiff: UInt64 = .max

    var tLast: UInt64 = 0
    for _ in 0 ..< 100_000_000 {
        let t = DispatchTime.now().uptimeNanoseconds

        let diff = t - tLast
        if diff < minDiff {
            minDiff = diff
        }

        tLast = t
    }

    print(minDiff)
}

foo()
```

```
$ swiftc -O test.swift && echo ok && ./test
ok
21
```

so the closest I could get these two is 21 which leaves us enough head room because in the real world, we'd allocate a bunch and add things to the `PriorityQueue` etc so it'll be much more than 21 ns difference.

What _is_ a real problem however is the deadline based API. If someone did something like:

```swift
let now = NIODeadline.now()
myEL.preconditionOnEL()

for f in 0..<100 {
    myEL.schedule(deadline: now) {
        print(f)
    }
}
```

then it's reasonable to assume it'll print `0, 1, 2, ..., 99` and not some random order. Today however, this is not true because our `PriorityQueue` isn't a FIFO if two elements have the same priority (where priority is the deadline here).

Lukasa:
I noodled over this a bit this weekend in between doing other things. I went through the following set of ideas, from worst to best. All of them rely on the idea that however we enqueue tasks will be linearizable: that is, if there’s a happens-before edge between the submission of two tasks then they must be de-queued by the loop in that order.

1. Use a 64-bit counter, monotonically incremented as tasks are put into the priority queue.

[...]

PR #2010, opened 2021-12-13
Convert `ScheduledTask` to a struct to reduce allocations for scheduling

FranzBusch:
### Motivation:

In my previous PR https://github.com/apple/swift-nio/pull/2009, I added baseline performance and allocation tests around `scheduleTask` and `execute`. After analysing, the various allocations that happen when scheduling a task there were only a few that could be optimized away potentially.

### Modifications:

This PR converts the `ScheduledTask` class to a struct which will reduce the number of allocations for scheduling tasks by 1. The only thing that needs to be worked around when converting to a struct is giving it an identity so that we can implement `Equatable` conformance properly. I explored two options. First, using an `ObjectIdentifier` passed to the init. Second, using an atomic counter per EventLoop. I went with the latter since the former requires an additional allocation in the case of calling `execute`

### Result:

`scheduleTask` and `execute` require one less allocation

### DIFF
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -16,12 +16,15 @@ import Dispatch
 import _NIODataStructures
 import NIOCore
 
-private final class EmbeddedScheduledTask {
+
+private struct EmbeddedScheduledTask {
+    let id: UInt64
     let task: () -> Void
     let readyTime: NIODeadline
     let insertOrder: UInt64
 
-    init(readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void) {
+    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void) {
+        self.id = id
         self.readyTime = readyTime
         self.insertOrder = insertOrder
         self.task = task
@@ -38,7 +41,7 @@ extension EmbeddedScheduledTask: Comparable {
     }
 
     static func == (lhs: EmbeddedScheduledTask, rhs: EmbeddedScheduledTask) -> Bool {
-        return lhs === rhs
+        return lhs.id == rhs.id
     }
 }
 
@@ -63,6 +66,7 @@ public final class EmbeddedEventLoop: EventLoop {
     /// The current "time" for this event loop. This is an amount in nanoseconds.
     /* private but tests */ internal var _now: NIODeadline = .uptimeNanoseconds(0)
 
+    private var scheduledTaskCounter: UInt64 = 0
     private var scheduledTasks = PriorityQueue<EmbeddedScheduledTask>()
 
     /// Keep track of where promises are allocated to ensure we can identify their source if they leak.
@@ -92,16 +96,18 @@ public final class EmbeddedEventLoop: EventLoop {
     @discardableResult
     public func scheduleTask<T>(deadline: NIODeadline, _ task: @escaping () throws -> T) -> Scheduled<T> {
         let promise: EventLoopPromise<T> = makePromise()
-        let task = EmbeddedScheduledTask(readyTime: deadline, insertOrder: self.nextTaskNumber()) {
+        self.scheduledTaskCounter += 1
+        let task = EmbeddedScheduledTask(id: self.scheduledTaskCounter, readyTime: deadline, insertOrder: self.nextTaskNumber(), task: {
             do {
                 promise.succeed(try task())
             } catch let err {
                 promise.fail(err)
             }
-        }
+        })
 
+        let taskId = task.id
         let scheduled = Scheduled(promise: promise, cancellationTask: {
-            self.scheduledTasks.remove(task)
+            self.scheduledTasks.removeFirst { $0.id == taskId }
         })
         scheduledTasks.push(task)
         return scheduled
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -63,6 +63,8 @@ internal final class SelectableEventLoop: EventLoop {
     // This may only be read/written while holding the _tasksLock.
     internal var _pendingTaskPop = false
     @usableFromInline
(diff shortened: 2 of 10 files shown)


Sources/NIOEmbedded/Embedded.swift
```diff
@@ -15,13 +15,18 @@
 import Dispatch
 import _NIODataStructures
 import NIOCore
+import NIOConcurrencyHelpers
 
-private final class EmbeddedScheduledTask {
+internal let atomicCounter = NIOAtomic.makeAtomic(value: 0)
```
Lukasa, line 20: Can we explicitly type this as 64-bit, and make it unsigned? This will give us the opportunity to solve #1541 using the same mechanism (though I don't propose doing so as part of this PR).

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -92,16 +97,17 @@ public final class EmbeddedEventLoop: EventLoop {
...
             } catch let err {
                 promise.fail(err)
             }
-        }
+        })
 
+        let taskId = task.id
         let scheduled = Scheduled(promise: promise, cancellationTask: {
-            self.scheduledTasks.remove(task)
+            self.scheduledTasks.removeAll { $0.id == taskId }
```
Lukasa, line 110: We don't need `removeAll`, we only need a `removeFirst` function: scanning the entire array once we've found the one we want to remove is unnecessary.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -271,9 +273,10 @@ Further information:
             promise.fail(error)
         }, deadline)
 
+        let taskId = task.id
         let scheduled = Scheduled(promise: promise, cancellationTask: {
             self._tasksLock.withLockVoid {
-                self._scheduledTasks.remove(task)
+                self._scheduledTasks.removeAll(where: { $0.id == taskId })
```
Lukasa, line 279: Same note here about `removeAll`.

Sources/_NIODataStructures/Heap.swift
```diff
@@ -115,6 +115,21 @@ internal struct Heap<Element: Comparable> {
             return false
         }
     }
+    
+    @inlinable
+    internal mutating func removeAll(where shouldBeRemoved: (Element) throws -> Bool) rethrows {
```
Lukasa, line 120: Hmm, why do you think we need this?

Sources/_NIODataStructures/Heap.swift
```diff
@@ -115,6 +115,21 @@ internal struct Heap<Element: Comparable> {
...
+        guard self.storage.count > 0 else {
+            return
+        }
+        
+        for (index, element) in self.storage.enumerated().reversed() {
```
Lukasa, line 125: Be very cautious here: `enumerated` does _not_ give you indices, it gives you a counter. Additionally, `EnumeratedSequence` is not a `BidirectionalCollection`, so I suspect this function call allocates, as `reversed` must iterate the entire sequence in order to produce the backwards version.

Lukasa (changes requested):

Sources/_NIODataStructures/Heap.swift
FranzBusch: I used `enumerated ` and `reversed` intentionally here since I am modifying the collection while iterating over it. However, your comment above is completely right that we only need a `removeFirst` or a `firstIndex(where)`. So I can remove this code anyhow and will update the PR 👍

FranzBusch:

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -15,13 +15,18 @@
 import Dispatch
 import _NIODataStructures
 import NIOCore
+import NIOConcurrencyHelpers
```
glbrntt, line 18: I don't think `NIOEmbedded` depended on `NIOConcurrencyHelpers` prior to this, could you add the dependency explicitly to the target in `Package.swift`?

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -15,13 +15,18 @@
...
 
-private final class EmbeddedScheduledTask {
+internal let atomicCounter = NIOAtomic.makeAtomic(value: UInt64(0))
```
glbrntt, line 20: Can this have a more explicit name?

Sources/NIOEmbedded/Embedded.swift
glbrntt: Equally, embedded scheduled task now has an ID and an insertion order.

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -63,6 +67,7 @@ public final class EmbeddedEventLoop: EventLoop {
     /// The current "time" for this event loop. This is an amount in nanoseconds.
     /* private but tests */ internal var _now: NIODeadline = .uptimeNanoseconds(0)
 
+    private var scheduledTaskCounter = NIOAtomic.makeAtomic(value: UInt64(0))
```
glbrntt, line 70: It's a shame that this now regresses a bunch of allocation counting tests that use `EmbeddedEventLoop` (and will no doubt regress allocation counting tests in other swift-nio-* packages). I wonder if we should make the counter for embedded loops global (or static) to reduce the noise of this one.

glbrntt: Actually `EmbeddedEventLoop` isn't thread safe, can we just use a `UInt64`?

Lukasa: We can, and probably should.

FranzBusch:

dnadoba:

Sources/NIOEmbedded/Embedded.swift
glbrntt: Hate to be a pain but we now no longer need this import! Looks good otherwise.

Lukasa (approved):
Really nice change, well done!

PR #2011, opened 2021-12-14
Call the `cancellationTask` of `Scheduled` directly

FranzBusch:
### Motivation:

In my previous PR https://github.com/apple/swift-nio/pull/2010, I was able to decrease the allocations for both `scheduleTask` and `execute` by 1 already. Gladly, there are no more allocations left to remove from `execute` now; however, `scheduleTask` still provides a couple of allocations that we can try to get rid of.

### Modifications:

This PR removes two allocations inside `Scheduled` where we were using the passed in `EventLoopPromise` to call the `cancellationTask` once the `EventLoopFuture` of the promise fails. This requires two allocations inside `whenFailure` and inside `_whenComplete`. However, since we are passing the `cancellationTask` to `Scheduled` anyhow and `Scheduled` is also the one that is failing the promise from the `cancel()` method. We can just go ahead and store the `cancellationTask` inside `Scheduled` and call it from the `cancel()` method directly instead of going through the future.

Importantly, here is that the `cancellationTask` is not allowed to retain the `ScheduledTask.task` otherwise we would change the semantics and retain the `ScheduledTask.task` longer than necessary. My previous PR https://github.com/apple/swift-nio/pull/2010, already implemented the work to get rid of the retain from the `cancellationTask` closure. So we are good to go ahead and store the `cancellationTask` inside `Scheduled` now

### Result:

`scheduleTask` requires two fewer allocations

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -21,18 +21,12 @@ import Dispatch
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
     /* private but usableFromInline */ @usableFromInline let _promise: EventLoopPromise<T>
+    /* private but usableFromInline */ @usableFromInline let _cancellationTask: (() -> Void)
 
     @inlinable
     public init(promise: EventLoopPromise<T>, cancellationTask: @escaping () -> Void) {
         self._promise = promise
-        promise.futureResult.whenFailure { error in
-            guard let err = error as? EventLoopError else {
-                return
-            }
-            if err == .cancelled {
-                cancellationTask()
-            }
-        }
+        self._cancellationTask = cancellationTask
     }
 
     /// Try to cancel the execution of the scheduled task.
@@ -42,6 +36,7 @@ public struct Scheduled<T> {
     @inlinable
     public func cancel() {
         self._promise.fail(EventLoopError.cancelled)
+        self._cancellationTask()
     }
 
     /// Returns the `EventLoopFuture` which will be notified once the execution of the scheduled task completes.
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -20,14 +20,20 @@ import NIOCore
 private struct EmbeddedScheduledTask {
     let id: UInt64
     let task: () -> Void
+    let failFn: (Error) -> ()
     let readyTime: NIODeadline
     let insertOrder: UInt64
 
-    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void) {
+    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void, _ failFn: @escaping (Error) -> ()) {
         self.id = id
         self.readyTime = readyTime
         self.insertOrder = insertOrder
         self.task = task
+        self.failFn = failFn
+    }
+
+    func fail(_ error: Error) {
+        self.failFn(error)
     }
 }
 
@@ -103,7 +109,7 @@ public final class EmbeddedEventLoop: EventLoop {
             } catch let err {
                 promise.fail(err)
             }
-        })
+        }, promise.fail)
 
         let taskId = task.id
         let scheduled = Scheduled(promise: promise, cancellationTask: {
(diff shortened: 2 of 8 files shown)


FranzBusch:

FranzBusch:

FranzBusch:

Sources/NIOCore/EventLoop.swift
```diff
@@ -21,18 +21,13 @@ import Dispatch
 /// will be notified once the execution is complete.
 public struct Scheduled<T> {
     /* private but usableFromInline */ @usableFromInline let _promise: EventLoopPromise<T>
+    
+    @usableFromInline let cancellationTask: (() -> Void)
```
glbrntt, line 25: Presumably this would be `private` if it wasn't `@usableFromInline`. By convention in these cases we prefix the name with a `_` so this should be `_cancellationTask`.

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -183,7 +189,9 @@ public final class EmbeddedEventLoop: EventLoop {
         // Just drop all the remaining scheduled tasks. Despite having run all the tasks that were
         // scheduled when we entered the method this may still contain tasks as running the tasks
         // may have enqueued more tasks.
-        while self.scheduledTasks.pop() != nil {}
+        while self.scheduledTasks.peek() != nil {
+            self.scheduledTasks.pop()?.fail(EventLoopError.shutdown)
```
glbrntt, line 193: Why not `while let task = self.scheduledTasks.pop() { ... }`?

glbrntt (commented):
Looks great modulo a couple of nits.

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -183,7 +189,9 @@ public final class EmbeddedEventLoop: EventLoop {
         // Just drop all the remaining scheduled tasks. Despite having run all the tasks that were
         // scheduled when we entered the method this may still contain tasks as running the tasks
         // may have enqueued more tasks.
-        while self.scheduledTasks.pop() != nil {}
+        while let task = self.scheduledTasks.pop() {
+            task.fail(EventLoopError.shutdown)
+        }
```
Lukasa, line 194: I _suspect_ this is a behavioural change. Can you write a test that encodes this?

FranzBusch: @Lukasa  So there is a test that is executing this `testFinishWithRecursivelyScheduledTasks` and without this change the test is actually crashing because the future is leaking. Which I am still not 100% sure why it did work before. Should I write an additional test that checks if the error is actually the `EventLoopError.shutdown`?

Lukasa: Yes please!

FranzBusch: I found a different test that was already testing the `EmbeddedEventLoop` specifically when draining the remaining tasks. I just added an assertion that makes sure the last task is failed with the proper error. Is that fine with you?

Lukasa (approved):

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -20,14 +20,20 @@ import NIOCore
 private struct EmbeddedScheduledTask {
     let id: UInt64
     let task: () -> Void
+    let failFn: (Error) ->()
```
glbrntt, line 23:
```suggestion
    let failFn: (Error) -> ()
```

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -20,14 +20,20 @@ import NIOCore
...
     let readyTime: NIODeadline
     let insertOrder: UInt64
 
-    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void) {
+    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void, _ failFn: @escaping (Error) ->()) {
```
glbrntt, line 27:
```suggestion
    init(id: UInt64, readyTime: NIODeadline, insertOrder: UInt64, task: @escaping () -> Void, _ failFn: @escaping (Error) -> ()) {
```

glbrntt (approved):

PR #2013, opened 2021-12-16
update docs generation script to match latest source-kitten

tomerd:
motivation: fix broken doc generation script

changes:
* pass the --spm flag now required by source-kitten
* update location of source-kitten source to make it more robust
* update list of module to generate docs for to match the structural changes in nio


glbrntt (approved):
Thanks Tomer!

tomerd:
@glbrntt may be worth going over the the NIO repos and making sure their doc generation script is updated the same way.

it can be tested locally with: `docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2004.54.yaml run --rm shell -clx scripts/generate_docs.sh`

glbrntt:
Thanks -- I updated the other NIO repos

PR #2014, opened 2021-12-19
Lock: use PTHREAD_MUTEX_ADAPTIVE_NP

weissi:
Motivation:

For consistent, short critical sections, [`PTHREAD_MUTEX_ADAPTIVE_NP`](https://stackoverflow.com/questions/19863734/what-is-pthread-mutex-adaptive-np) might be beneficial (under contention).

Modification:

Use `PTHREAD_MUTEX_ADAPTIVE_NP` on Linux.

Result:

Could be a perf advantage (let's see the benchmarks).

### DIFF
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -43,6 +43,9 @@ public final class Lock {
 #else
         var attr = pthread_mutexattr_t()
         pthread_mutexattr_init(&attr)
+#if os(Linux)
+        pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ADAPTIVE_NP))
+#endif
         debugOnly {
             pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK))
         }


weissi:
@swift-nio-bot perf test this please

weissi:

weissi:
lock_1_thread_10M_ops | 0.178784859 | 0.159197135 | previous | 12%
-- | -- | -- | -- | --
lock_2_threads_10M_ops | 0.378932489 | 0.79521017 | current | -52%
lock_4_threads_10M_ops | 0.373160936 | 0.870138925 | current | -57%
lock_8_threads_10M_ops | 0.354516861 | 1.065500287 | current | -66%

Pretty phenomenal improvements under contention and a 12% slowdown without contention. This result is entirely expected because:

- the adaptive lock needs to measure the length (in time) of the critical section. That adds some overhead.
- if contended however, we save a boat load of `futex` syscalls.

I reckon this isn't really worth it for SwiftNIO because we're hoping for mostly no contention because the `EventLoop`s themselves are usually used as the synchronisation for most code. At the moment the `EventLoop`s internally use locks for the task queue but most well-architected NIO applications should have very little cross-EL traffic which is to say that the EL's task queue lock should be pretty much uncontended.

@Lukasa / @glbrntt / @fabianfett / @dnadoba WDYT?

glbrntt:
Interesting, TIL about PTHREAD_MUTEX_ADAPTIVE_NP.

> I reckon this isn't really worth it for SwiftNIO because we're hoping for mostly no contention because the `EventLoop`s themselves are usually used as the synchronisation for most code. At the moment the `EventLoop`s internally use locks for the task queue but most well-architected NIO applications should have very little cross-EL traffic which is to say that the EL's task queue lock should be pretty much uncontended.

I think that's a fair assessment.

IIUC the benchmark might also be flattering for PTHREAD_MUTEX_ADAPTIVE_NP in the contended tests because the benchmark is so predictable. The other relevant point is that:

> This adaptive approach is specialized, in the sense that it will not work for all locks in all programs, so it is packaged as a special mutex type.

Using it for NIOs 'general purpose' lock doesn't seem wise but maybe there are opportunities to use it in specific places in NIO? No doubt you'd have more insight there than I would!

Lukasa:
I'm inclined to say that if we were going to expose this we should expose it as a configuration on the standard lock, rather than simply converting all of them over to it. The most promising lock to use for this would be the event loop task lock, which is the most likely to be contended and by far the hottest. However, I'd be very reluctant to make that change: while it improves our synthetic benchmarks I doubt it'll produce meaningful change in real programs, where that lock has simply never been that hot.

The better thing to do is to continue our attempts to drop older Swift versions so that we can adopt swift-atomics and start trying to implement a proper MPSC queue.

PR #2015, opened 2021-12-20
Add canImport(_Concurrency) for Xcode 13.0

fabianfett:
Xcode 13.0 lacked concurrency support for macOS. We need to make sure we only use _Concurrency features, if the platform supports them.

### Modifications:

- Add `canImport(_Concurrency)` for Xcode 13.0 in `AsyncAwaitSupport+OldXcodes.swift`

### Result:

Users can compile for macOS with Xcode 13.0.


Jomy10 (approved):
This indeed fixes the issue

glbrntt (approved):

Issue #1239, opened 2019-11-13
switch to SwiftPM's `swift package experimental-api-diff`


weissi:
Ideally, the `check_no_api_breakages.sh` script should use `swift package experimental-api-diff 2.0.0` that is present in _very new_ Swift nightlies to do its job. Then it will no longer need to enumerate the modules etc.

CC @aciidb0mb3r

PR #2019, opened 2021-12-23
Use SPM breaking changes checker

FranzBusch:
### Motivation:

SPM has built in functionality to check the API of modules against a target git treeish. We can use this to simplify our `check_no_api_breakages.sh` script. Closes https://github.com/apple/swift-nio/issues/1239

### Modifications:

This PR, exchanges the direct calls to Swift's API checker with the new SPM `diagnose-api-breaking-changes` tool. This allows us to get rid of the manual module parsing, build invocations and result comparisons.

### Result:

We are now using SPMs `diagnose-api-breaking-changes` to check for breaking changes.

### DIFF
--- a/scripts/check_no_api_breakages.sh
+++ b/scripts/check_no_api_breakages.sh
@@ -15,35 +15,6 @@
 
 set -eu
 
-# repodir
-function all_modules() {
-    local repodir="$1"
-    (
-    set -eu
-    cd "$repodir"
-    swift package dump-package | jq '.products |
-                                     map(select(.type | has("library") )) |
-                                     map(.name) | .[]' | tr -d '"'
-    )
-}
-
-# repodir tag output
-function build_and_do() {
-    local repodir=$1
-    local tag=$2
-    local output=$3
-
-    (
-    cd "$repodir"
-    git checkout -q "$tag"
-    swift build
-    while read -r module; do
-        swift api-digester -sdk "$sdk" -dump-sdk -module "$module" \
-            -o "$output/$module.json" -I "$repodir/.build/debug"
-    done < <(all_modules "$repodir")
-    )
-}
-
 function usage() {
     echo >&2 "Usage: $0 REPO-GITHUB-URL NEW-VERSION OLD-VERSIONS..."
     echo >&2
@@ -63,12 +34,6 @@ if [[ $# -lt 3 ]]; then
     exit 1
 fi
 
-sdk=/
-if [[ "$(uname -s)" == Darwin ]]; then
-    sdk=$(xcrun --show-sdk-path)
-fi
-
-hash jq 2> /dev/null || { echo >&2 "ERROR: jq must be installed"; exit 1; }
 tmpdir=$(mktemp -d /tmp/.check-api_XXXXXX)
 repo_url=$1
 new_tag=$2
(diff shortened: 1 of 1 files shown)


FranzBusch:
@swift-nio-bot please test

FranzBusch:

FranzBusch:

FranzBusch:

scripts/check_no_api_breakages.sh
```diff
@@ -15,111 +15,5 @@
...
-    rm -rf "$tmpdir/api-new" "$tmpdir/api-old"
-done
-
-if [[ "$errors" == 0 ]]; then
-    echo "OK, all seems good"
-fi
-echo done
-exit "$errors"
+git fetch
+swift package diagnose-api-breaking-changes refs/remotes/origin/main
```
weissi, line 19: this isn't ideal as it will only work if `origin` is actually the upstream remote.

FranzBusch: I know this is not ideal. I really just want to use just `main` but that does not work. I guess the CI is doing a shallow clone and `main` is not fetched in that case. I am still trying some things out to find a good config that works both remotely and locally.

FranzBusch:
Here are two example runs one that failed and one that succeeded:
https://ci.swiftserver.group/job/swift-nio2-api-breakage-prb/2579/console
https://ci.swiftserver.group/job/swift-nio2-api-breakage-prb/2580/console

FranzBusch:
@weissi @Lukasa since the `diagnose-api-breaking-changes` is only available in the nightly toolchains I would suggest us to move the api breakage pipeline to the `docker-compose.2004.main.yaml` docker setup. I also tried to use the experimental tool that is shipped with 5.5 but I ran into several errors that have been fixed on main already. What do you think about using the other docker setup for the api breaking checks?

scripts/check_no_api_breakages.sh
weissi: Maybe @tomerd can help here. I’m pretty sure that CI would somehow be able to refer to the base branch for the PR. Technically `main` is also not correct (but correct enough probably) because you want to compare to the target branch your PR goes into.

FranzBusch: With my latest changes it actually works and I just refactored your script to call out to SPM instead. I does a separate clone still. If we just want to simply invoke SPM with `main` then we need some changes on the CI probably.

glbrntt:
> @weissi @Lukasa since the `diagnose-api-breaking-changes` is only available in the nightly toolchains I would suggest us to move the api breakage pipeline to the `docker-compose.2004.main.yaml` docker setup. I also tried to use the experimental tool that is shipped with 5.5 but I ran into several errors that have been fixed on main already. What do you think about using the other docker setup for the api breaking checks?

I'd rather stick to a released version to avoid dealing with potential problems associated with nightly toolchains (compiler crashes etc.) and revisit this when the tool is no longer experimental.

FranzBusch:
@glbrntt I spoke with @fabianfett yesterday about this as well and we came to the same conclusion. Let's put this PR on hold until the tool lands in a released Swift version and then we can revisit it.

scripts/check_no_api_breakages.sh
```diff
@@ -77,49 +42,14 @@ shift 2
...
-    rm -rf "$tmpdir/api-new" "$tmpdir/api-old"
+    swift package diagnose-api-breaking-changes "$old_tag"
 done
 
-if [[ "$errors" == 0 ]]; then
-    echo "OK, all seems good"
-fi
 echo done
-exit "$errors"
+exit 0
```
weissi, line 55: that line is superfluous but also doesn't hurt ;)

weissi (approved):
This is fantastic, thank you!

PR #2020, opened 2021-12-27
Guarantee ordering of `scheduleTask` with same deadline

FranzBusch:
### Motivation:

When scheduling tasks with the same deadline the current order of execution is undefined. Fixes https://github.com/apple/swift-nio/issues/1541

### Modifications:

In my previous PR https://github.com/apple/swift-nio/pull/2010, I added an internal id to every `ScheduledTask` to give them an identity for cancellation purposes. In this PR, I am now using the same id to also ensure that the execution of tasks with the same deadline is the same as the order they were scheduled in.

### Result:

`ScheduledTask`s are now executed in their scheduled order when they have the same deadline.

### DIFF
--- a/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
+++ b/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
@@ -339,6 +339,11 @@ extension MultiThreadedEventLoopGroup: CustomStringConvertible {
 
 @usableFromInline
 internal struct ScheduledTask {
+    /// The id of the scheduled task.
+    ///
+    /// - Important: This id has two purposes. First, it is used to give this struct an identity so that we can implement ``Equatable``
+    ///     Second, it is used to give the tasks an order which we use to execute them.
+    ///     This means, the ids need to be unique for a given ``SelectableEventLoop`` and they need to be in ascending order.
     @usableFromInline
     let id: UInt64
     let task: () -> Void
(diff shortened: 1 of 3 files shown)


dnadoba (approved):
LGTM

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -376,7 +376,11 @@ extension ScheduledTask: CustomStringConvertible {
 extension ScheduledTask: Comparable {
     @usableFromInline
     static func < (lhs: ScheduledTask, rhs: ScheduledTask) -> Bool {
-        return lhs._readyTime < rhs._readyTime
+        if lhs._readyTime == rhs._readyTime {
```
fabianfett, line 379: Please add a comment to `id` in line 343, explaining that the `id` must be unique within an EventLoop and increasing for new scheduled tasks. The reason for this being that we use the id and the readyTime to determine ordering.

Tests/NIOPosixTests/EventLoopTest.swift
```diff
@@ -184,6 +184,26 @@ public final class EventLoopTest : XCTestCase {
         XCTAssertEqual(error as? EventLoopError, .cancelled)
     }
 
+    public func testScheduledTasksAreOrdered() throws {
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
Lukasa, line 188: Can we have a version of this test for `EmbeddedEventLoop` too?

Lukasa (changes requested):

Tests/NIOPosixTests/EventLoopTest.swift
FranzBusch: There is a test for the `EmbeddedEventLoop` already that tests the ordering `testScheduledTasksInOrder`. It looks slightly different but has the same effect. Is that one enough or do you want me to add the exact same test as well?

Lukasa: That should be fine.

Lukasa (approved):

fabianfett (approved):

Issue #1339, opened 2020-01-14
we should assert/precondition that read doesn't return EINVAL


weissi:
`read` returning `EINVAL` is a bug in NIO and we should assert/precondition that it doesn't happen similar to what we do for `EBADF`/`EFAULT` on all syscalls. See #1337 for an example where that happens.

Bouke:
In another project, I'm wrapping all posix calls like this: https://github.com/Bouke/HAP/blob/a494b40ab0c84b729e4a618a132879dcc68265f9/Sources/COperatingSystem/libc.swift#L7-L11. Do you have something similar in mind?

Lukasa:
We do. We have a wrapper like that as well, but it turns some errors into crashes, on the grounds that they should never happen. Johannes is proposing adding EINVAL to that list of errors.

weissi:
@Bouke just a drive-by comment: errno numbers differ across operating systems, your snippet looks like they’re hardcoded. If you have taken care of this in a different way please ignore

Bouke:
@weissi I didn’t know that, thanks! In my code I’m mostly interested in success or failure anyway. Do you know of a cross os way to create strict posix error (enum)?

weissi:
@Bouke hmm, what are you using the enum for? I’d just work with raw values (check NIO’s `IOError` type for example). The number of errno codes between platforms will also vary.

And regarding success or failure: in posix, there are some error codes that you definitely need to handle specially, for example `EINTR`.

Also be very very careful using `errno` in Swift, there is no guaranteed correct way to use errno in Swift because of ARC. Bugs.swift.org is down so I can’t find the bug number right now. But the back story is that ARC can run code at any point in time and the code run by ARC might change the `errno` value. Therefore you must somehow make sure that between the syscall and the place where you read `errno` there’s definitely no `deinit`s run by ARC. Using carefully placed inline(never) and inline(__always), NIO does the best we can do today in our `enum Posix`

PR #2022, opened 2021-12-30
Precondition that read doesn't return EINVAL

FranzBusch:
### Motivation:

`read` should never return `EINVAL`. If it does return it though it is a bug inside NIO and we should precondition for it similar to what we do for `EBADF` & `EFAULT`. Closes https://github.com/apple/swift-nio/issues/1339

### Modifications:

This PR adds a new precondition for `EINVAL` and adds test that exercise this precondition.

### Result:

We are now preconditioning against `read` returning `EINVAL`.


IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
```diff
@@ -61,7 +63,7 @@ EOF
 make_package
 
 for mode in debug release; do
-    for error in EFAULT EBADF; do
+    for error in EFAULT EBADF, EINVAL; do
```
weissi, line 66: The comma isn’t correct here

FranzBusch, line 66:
```suggestion
    for error in EFAULT EBADF EINVAL; do
```

FranzBusch:
@swift-nio-bot test please

FranzBusch:

FranzBusch:
@weissi Do you have any idea why these tests are failing? They are also failing for me locally on main and I cannot figure out why they are returning the wrong value now.

IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
```diff
@@ -47,10 +47,12 @@ if pipeErr != 0 {
 }
 let makeEBADFHappen =  CommandLine.arguments.dropFirst().first == .some("EBADF")
 let makeEFAULTHappen = CommandLine.arguments.dropFirst().first == .some("EFAULT")
-var whatevs: UInt8 = 123
+let makeEINVALHappen = CommandLine.arguments.dropFirst().first == .some("EINVAL")
+var whatevs: UInt8 = makeEINVALHappen ? 3021 : 123
```
Lukasa, line 51: This does not produce EINVAL on Posix.write. From the manual page:

```
[EINVAL]           The pointer associated with fildes is negative.
```

`3021` is not negative.

Lukasa: It is, however, larger than `UInt8` can hold, so the compile fails. This is why the tests are failing.

IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
```diff
@@ -47,10 +47,12 @@ if pipeErr != 0 {
 }
 let makeEBADFHappen =  CommandLine.arguments.dropFirst().first == .some("EBADF")
 let makeEFAULTHappen = CommandLine.arguments.dropFirst().first == .some("EFAULT")
-var whatevs: UInt8 = 123
+let makeEINVALHappen = CommandLine.arguments.dropFirst().first == .some("EINVAL")
+var whatevs: UInt8 = makeEINVALHappen ? -1 : 123
```
Lukasa, line 51: This still won't work. `EINVAL` is triggered by the _pointer_ being negative: this is the `pointee`, and so unaffected. Also, this still won't compile. :smile:

Tests/NIOPosixTests/EventLoopTest+XCTest.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2022 Apple Inc. and the SwiftNIO project authors
```
FranzBusch, line 5: This is currently breaking on the CI and was caused by my previous PR where the CI did not detect this necessary change. I am just going to include it here to fix the CI.

Lukasa (changes requested):
This change is good, but it's a bit too narrow and a bit too broad.

The original issue only asked to forbid `EINVAL` on `read`. This change adds it to all syscalls in NIOCore, but does not affect those in NIOPosix, which is where the relevant `read` is actually made. As a result, it does not resolve the original issue.

Additionally, the change as written affects _all_ system calls. However, some EINVALs we actually want to tolerate. As an example, see #1030. As a result, we likely need a more complex solution that only adds EINVAL policing to some specific system calls, not all of them.

IntegrationTests/tests_02_syscall_wrappers/test_03_unacceptable_read_errnos.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 6:
```suggestion
## Copyright (c) 2017-2022 Apple Inc. and the SwiftNIO project authors
```

Sources/NIOPosix/System.swift
```diff
@@ -411,14 +386,14 @@ internal enum Posix {
 
     @inline(never)
     internal static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<ssize_t> {
-        return try syscall(blocking: true) {
+        return try syscall(blocking: true, unacceptableErrnoCodes: defaultUnacceptableErrnoCodes + [EINVAL]) {
```
Lukasa, line 389: Unfortunately I don't think this approach is sufficient for us.

The biggest issue I have here is that this causes a runtime Array allocation on all codepaths through this function. I suspect the performance impact of this is fairly substantial.

Unfortunately, the only really effective solution at the moment is to duplicate the `isUnacceptableErrno` function with a new body. This is what we did for `close`, and sadly is what we have to do here as well.

Lukasa (changes requested):

Sources/NIOPosix/System.swift
Lukasa: Ah, yup, in fact the allocation counter tests failed.

Lukasa:

Lukasa:
Yup, that report above is not right: those are clearly the results from @weissi's previous PR.

Lukasa:
@tomerd Seems like the NIO performance builds aren't working again.

Sources/NIOPosix/System.swift
```diff
@@ -195,19 +220,32 @@ internal func syscall<T: FixedWidthInteger>(blocking: Bool,
...
+
+/*
+ * Sorry, we really try hard to not use underscored attributes. In this case
+ * however we seem to break the inlining threshold which makes a system call
+ * take twice the time, ie. we need this exception.
+ */
 @inline(__always)
-internal func wrapErrorIsNullReturnCall<T>(where function: String = #function, _ body: () throws -> T?) throws -> T {
+@discardableResult
+internal func sysReadCall<T: FixedWidthInteger>(where function: String = #function,
```
Lukasa, line 231: Hmm, I don't love this: it's too specific to `read` (same with those above). Maybe `syscallForbiddingEINVAL`?

Lukasa (approved):
LGTM, nice one!

PR #2038, opened 2022-02-03
Add `collect(into:maxBytes:)` to `AsyncSequence`

dnadoba:
Adds two methods `collect(into:maxBytes:)` and `collect(maxBytes:)` on `AsyncSequence where Element: RandomAccessCollection, Element.Element == UInt8` to safely accumulate fragments into one `ByteBuffer`.

### Motivation:
New async/await streaming APIs use an `AsyncSequence` of `ByteBuffer`s to incrementally forward new data to the API consumer. However, sometimes they are not interested in each fragment and just want the complete result. [One such new API is `AsyncHTTPClient`s new `HTTPClientResponse.Body`](https://github.com/swift-server/async-http-client/blob/14c95bf6c94e04b652ef44c99dc0b21eed1a9c1b/Sources/AsyncHTTPClient/AsyncAwait/HTTPClientResponse.swift#L50-L68). If a user wants to decode JSON they need the complete result before they can pass it to `JSONDecoder`.

### Modifications:
- Adds two methods `collect(into:maxBytes:)` and `collect(maxBytes:)` on `AsyncSequence where Element: RandomAccessCollection, Element.Element == UInt8`
- Adds two convenience methods `collect(into:maxBytes:)` and `collect(maxBytes:)` on `AsyncSequence where Element == ByteBuffer`
- Add `XCTest+AsyncAwait.swift` from `AsyncHTTPClient` (which has most of it from `grpc-swift`) to be able to test async/await APIs on Linux

### Result:

Users can safely convert an `AsyncSequence` of Bags of Bytes into a single `ByteBuffer`.


Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -218,4 +218,77 @@ extension ChannelPipeline {
...
+    
+    /// Consumes an ``Swift/AsyncSequence`` of ``Swift/RandomAccessCollection``s into a single ``NIO/ByteBuffer``.
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes this method is allowed to write into `accumulationBuffer`
+    /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+    @inlinable
+    public func collect(
+        maxBytes: Int
+    ) async throws -> ByteBuffer {
+        var accumulationBuffer = ByteBuffer()
```
fabianfett, line 258: I think, we should allocate some space here. Otherwise we realloc a lot in the beginning.

dnadoba: But how much? 
For the case where `AsyncSequence.Element == ByteBuffer` I have just added an optimisation to use the first element as the `accumulationBuffer`. Do you think this would be enough?

glbrntt: Historically we have always suggested that users use an allocator provided by the channel to create buffers. Of course we now have a handful of `init`s for creating `ByteBuffer`s without an allocator which is certainly useful for testing etc. 

Is it our position now that we're happy for folks (and indeed us) to create `ByteBuffer`s without an allocator?

Lukasa: I'm inclined to keep this API, but to have it take a `ByteBufferAllocator`. That way we preserve the implication that we may one day do something sensible with this type.

Lukasa: And I do agree with `@fabianfett` that we should pre-allocate some space here. Maybe 512 bytes? Just to skip a few doublings.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -218,4 +218,102 @@ extension ChannelPipeline {
...
+    ///   - accumulationBuffer: buffer to write all the elements of `self` into
+    ///   - maxBytes: The maximum number of bytes this method is allowed to write into `accumulationBuffer`
+    /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+    /// Note that previous elements of `self` might be already write to `accumulationBuffer`.
+    @inlinable
+    public func collect(
+        into accumulationBuffer: inout ByteBuffer,
+        maxBytes: Int
+    ) async throws {
+        try await self.map(\.readableBytesView).collect(into: &accumulationBuffer, maxBytes: maxBytes)
```
Lukasa, line 279: I think this is unnecessarily slow: going through the generic `writeBytes` implementation for `ByteBufferView` will be substantially less performant than using `writeImmutableBuffer` for the ByteBuffers directly.

Sources/NIOCore/AsyncAwaitSupport.swift
fabianfett: `512` sounds good to me. I would also be okay with `1024`. My feeling is generally that swift is so memory efficient, that we can gladly trade a little more memory consumption for better performance.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -218,4 +218,114 @@ extension ChannelPipeline {
...
+        var iterator = self.makeAsyncIterator()
+        guard var head = try await iterator.next() else {
+            return ByteBuffer()
+        }
+        guard head.readableBytes <= maxBytes else {
+            throw NIOTooManyBytesError()
+        }
+        
+        let tail = AsyncSequenceFromIterator(iterator)
+        try await tail.collect(into: &head, maxBytes: maxBytes - head.readableBytes)
```
glbrntt, line 310: In before Cory ;)

```suggestion
        try await tail.collect(into: &head, maxBytes: maxBytes &- head.readableBytes)
```

Tests/NIOCoreTests/XCTest+AsyncAwait.swift
```diff
@@ -0,0 +1,106 @@
+//===----------------------------------------------------------------------===//
```
glbrntt, line 1: License headers are like Pokemon; gotta catch 'em all.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -265,25 +266,36 @@ extension AsyncSequence where Element: RandomAccessCollection, Element.Element =
...
-    /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+    /// - Throws: ``NIOTooManyBytesError`` if the the sequence contains more than `maxBytes`.
     /// Note that previous elements of `self` might be already write to `accumulationBuffer`.
     @inlinable
     public func collect(
         into accumulationBuffer: inout ByteBuffer,
         maxBytes: Int
     ) async throws {
-        try await self.map(\.readableBytesView).collect(into: &accumulationBuffer, maxBytes: maxBytes)
+        var bytesRead = 0
```
fabianfett, line 280: should we precondition on maxBytes being negative?

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -218,4 +218,114 @@ extension ChannelPipeline {
...
+    @usableFromInline typealias Element = AsyncIterator.Element
+    
+    @usableFromInline var iterator: AsyncIterator
+    
+    @inlinable init(_ iterator: AsyncIterator) {
+        self.iterator = iterator
+    }
+    
+    @inlinable func makeAsyncIterator() -> AsyncIterator {
+        iterator
```
glbrntt, line 327:
```suggestion
        self.iterator
```

glbrntt (approved):
Looks good modulo a couple of nits. I think we should @fabianfett's suggestion of the precondition though.

Tests/NIOCoreTests/XCTest+AsyncAwait.swift
fabianfett: 😍 🐲  🤣

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -218,4 +218,120 @@ extension ChannelPipeline {
...
+    
+    /// Accumulates an ``Swift/AsyncSequence`` of ``Swift/RandomAccessCollection``s into a single ``NIO/ByteBuffer``.
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes this method is allowed to accumulate
+    ///   - allocator: Allocator used for allocating the result `ByteBuffer`
+    /// - Throws: `NIOTooManyBytesError` if the the sequence contains more than `maxBytes`.
+    @inlinable
+    public func collect(
+        maxBytes: Int,
+        using allocator: ByteBufferAllocator
```
fabianfett, line 257: The argument order feels weird to me here (as it is different to all other function calls, where `maxBytes` is the last argument). This is increased through the naming. IIRC plain argument names should come more to the end, whereas argument names that form a sentence should be more at the beginning.

dnadoba: Yeah, naming is hard...
I did come up with another idea:
```
func collect(upTo maxBytes: Int, into accumulationBuffer: inout ByteBuffer)
func collect(upTo maxBytes: Int, using allocator: ByteBufferAllocator) -> ByteBuffer
// for the ByteBuffer specific one without allocator
func collect(upTo maxBytes: Int) -> ByteBuffer
```
WDYT?

fabianfett: I think I prefer the maxBytes at the end. But that's just my personal opinion. @Lukasa @glbrntt wdyt?

glbrntt: I prefer David's suggestion with `upTo maxBytes` first, it reads better.

dnadoba: renamed to `upTo maxBytes` and moved it to the front.

Sources/NIOCore/AsyncAwaitSupport.swift
dnadoba: increased to `1024`

fabianfett (approved):
Wow. Discussing new API is fun, but in the end it turned out awesome. Congrats to a great API addition.

Lukasa (approved):

PR #2039, opened 2022-02-03
Improve the performance of copying BBV

Lukasa:
Motivation:

When turning a BBV into a Collection, several of the collection
algorithms can do better if you implement the semi-private
`_copyContents` hook. We can easily implement it, so we should. While
I'm here we should clean up some other bits.

Modifications:

- Implement _copyContents.
- Implement a faster `.count` for `ByteBufferView`.
- Make `ByteBuffer.capacity` `@inlinable` to reduce ARC traffic

Result:

Copying a BBV into an Array is faster.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -35,13 +35,15 @@ extension _ByteBufferSlice: Equatable {}
 /// fits within 24 bits, otherwise the behaviour is undefined.
 @usableFromInline
 struct _ByteBufferSlice {
-    @usableFromInline var upperBound: ByteBuffer._Index
-    @usableFromInline var _begin: _UInt24
+    @usableFromInline private(set) var upperBound: ByteBuffer._Index
+    @usableFromInline private(set) var _begin: _UInt24
     @inlinable var lowerBound: ByteBuffer._Index {
         return UInt32(self._begin)
     }
     @inlinable var count: Int {
-        return Int(self.upperBound - self.lowerBound)
+        // Safe: the only constructors that set this enforce that upperBound > lowerBound, so
+        // this cannot underflow.
+        return Int(self.upperBound &- self.lowerBound)
     }
     init() {
         self._begin = .init(0)
@@ -453,6 +455,7 @@ public struct ByteBuffer {
 
     /// The current capacity of the storage of this `ByteBuffer`, this is not constant and does _not_ signify the number
     /// of bytes that have been written to this `ByteBuffer`.
+    @inlinable
     public var capacity: Int {
         return self._slice.count
     }
--- a/Sources/NIOCore/ByteBuffer-views.swift
+++ b/Sources/NIOCore/ByteBuffer-views.swift
@@ -61,6 +61,13 @@ public struct ByteBufferView: RandomAccessCollection {
     }
 
     @inlinable
+    public var count: Int {
+        // Unchecked is safe here: Range enforces that upperBound is strictly greater than
+        // lower bound, and we guarantee that _range.lowerBound >= 0.
+        return self._range.upperBound &- self._range.lowerBound
+    }
+
+    @inlinable
     public subscript(position: Index) -> UInt8 {
         get {
             guard position >= self._range.lowerBound && position < self._range.upperBound else {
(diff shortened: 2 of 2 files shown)


Lukasa:

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -60,6 +60,13 @@ public struct ByteBufferView: RandomAccessCollection {
         return i + 1
     }
 
+    @inlinable
+    public var count: Int {
+        // Unchecked is safe here: Range enforces that upperBound is strictly greater than
+        // lower bound. Annoyingly, Range itself does not do this.
```
dnadoba, line 66: is this worth a bug report or even directly a patch?

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -106,6 +113,23 @@ public struct ByteBufferView: RandomAccessCollection {
...
+        precondition(ptr.count >= self.count)
+
+        let bytesToWrite = self.count
+
+        let endIndex = self.withContiguousStorageIfAvailable { ourBytes in
+            ptr.initialize(from: ourBytes).1
+        }
+        precondition(endIndex == bytesToWrite)
+
+        let iterator = self[self.endIndex...self.endIndex].makeIterator()
```
dnadoba, line 130: what is this iterator suppose to be? Now it is just empty

Sources/NIOCore/ByteBuffer-views.swift
Lukasa: Yes, probably.

Sources/NIOCore/ByteBuffer-views.swift
dnadoba: ah no, it is not empty. `self.endIndex...self.endIndex` will result in a range with just `self.endIndex` in it. But this seems wrong...

Sources/NIOCore/ByteBuffer-views.swift
glbrntt: Surely it's not safe for Range to do that? e.g. `-1 ... Int.max`?

Lukasa: Yup good catch, that will trap.

Sources/NIOCore/ByteBuffer-views.swift
Lukasa: Nope, it's right. [This](https://github.com/apple/swift/blob/ee7e04822c556b481e28dabc4274e54d4ab4f0ae/stdlib/public/core/Sequence.swift#L360-L385) is the customisation point we're implementing. From it we need to return an iterator on _our_ type which contains any elements we couldn't copy into the buffer. Because we can always copy the elements into the buffer, we return the empty iterator.

dnadoba: Wouldn't it then be `self.endIndex..<self.endIndex` instead of `self.endIndex...self.endIndex`, in other words a `Range` instead of a `ClosedRange` to create a real empty range? Otherwise the iterator will try to iterate over one element but will fail to do say because it is outside the bounds of `self`.

dnadoba: The test currently crash too. If I change it to a `Range` they no longer crash and pass.

tomerd:

tomerd:

tomerd:

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -35,13 +35,15 @@ extension _ByteBufferSlice: Equatable {}
 /// fits within 24 bits, otherwise the behaviour is undefined.
 @usableFromInline
 struct _ByteBufferSlice {
-    @usableFromInline var upperBound: ByteBuffer._Index
-    @usableFromInline var _begin: _UInt24
+    @usableFromInline private(set) var upperBound: ByteBuffer._Index
+    @usableFromInline private(set) var _begin: _UInt24
```
fabianfett, line 39: 👍 Love that this is possible and we should use it more often.

fabianfett (commented):
@dnadoba and @glbrntt have already done a great job here. Nothing to add from my side.

Sources/NIOCore/ByteBuffer-core.swift
Lukasa: This is also a surprising performance win in some cases (though not here, I think). In particular it means that the compiler can exhaustively audit all code that is able to `set` this stored property, and so can often elide runtime exclusivity checks.

Sources/NIOCore/ByteBuffer-views.swift
Lukasa: Yeah good catch, I constructed this incorrectly.

Sources/NIOCore/ByteBuffer-views.swift
dnadoba: I would then suggest to remove `Annoyingly, Range itself does not do this.` and add that we guarantee `range.lowerBound >= 0`.

dnadoba (approved):
One small nit, otherwise LGTM. Looks like it can be a huge performance improvement! 🏎

Lukasa:
-91%, yes, I'd call that significant.

tomerd:
nice one

PR #2041, opened 2022-02-06
Allow NIOEchoClient to receive fragmented echo responses.

SeJV:
Allow NIOEchoClient to receive fragmented echo responses.

### Motivation:

While sending a message to NIOEchoServer and receiving the echoed response, it is not guaranteed that the sent message arrives inside one `channelRead()` call. This would fail in the current implementation.  

### Modifications:

On `channelRead` the data will be written to a buffer until all `sendBytes` got received.  

### Result:

The updated NIOEchoClient can now buffer split `InboundIn` data and prints the result, when all `sendBytes` got received.   

#### Tests: 
Tests for executables require `swift-tools-version 5.5`. To recreate the tests of the new `NIOEchoClient` receive `message` argument from constructor instead of the `readLine()`.  Additionally `context.writeAndFlush()` the received buffer instead of printing it. 

```swift 
let expectedMessage = "Expected to read"
let expectedBuffer = ByteBuffer(string: expectedMessage)

let handler = EmbeddedChannel(handler: EchoHandler(message: expectedMessage))
try handler.connect(to: .init(ipAddress: "127.0.0.1", port: 80), promise: nil)
// empty outbound buffer
_ = try handler.readOutbound(as: ByteBuffer.self)

// Test case: EchoHandler receives split buffer
try handler.writeInbound(ByteBuffer(string: String(expectedMessage.prefix(5))))
try handler.writeInbound(ByteBuffer(string: String(expectedMessage.dropFirst(5))))
XCTAssertEqual(try handler.readOutbound(as: ByteBuffer.self), expectedBuffer)
```


Lukasa (approved):
Nice change, thanks!

Lukasa:

PR #2043, opened 2022-02-08
Fix the double encoding issue in encodeAsByteBuffer

stepan-ulyanin:
Fixes the double encoding issue with writing the bytes directly into the byte buffer in `encodeAsByteBuffer`.

### Motivation:
`encodeAsByteBuffer` encodes a value into bytes and then, during the `buffer.writeJSONEncodable(_, encoder:)` call, encodes the value into bytes again.

### Modifications:
The call to `buffer.writeJSONEncodable(_, encoder:)` is replaced with writing the bytes directly into the byte buffer.

### Result:
No more double encoding the value into bytes.


Lukasa (approved):
Nice one, LGTM!

Lukasa:

Lukasa (approved):

fabianfett (approved):
Great catch!

PR #2045, opened 2022-02-09
Add utilties for reading and writing UUIDs

glbrntt:
Motivation:

UUIDs are often sent over the wire but writing and reading their bytes
to/from a buffer is a bit of a pain: the obvious solution is grabbing
the underlying 16-tuple and writing each byte manually and doing the reverse
to read. There are also unsafe options but this is reasonably
straightforward to do (if slightly fiddly) by packing the bytes into two
`UInt64`s.

Modifications:

- Add utilties to 'NIOFoundationCompat' for reading/writing and
  getting/setting a UUID on a `ByteBuffer`.

Result:

Easier to write/read UUIDs to/from a buffer.


Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
```diff
@@ -265,6 +265,101 @@ extension ByteBuffer {
...
+    /// Get a `UUID` from the 16 bytes starting at `index`. This will not change the reader index.
+    /// If there are less than 16 bytes starting at `index` then `nil` will be returned.
+    ///
+    /// - Parameters:
+    ///   - index: The starting index of the bytes of interest into the `ByteBuffer`.
+    /// - Returns: A `UUID` value containing the bytes of interest or `nil` if the selected bytes
+    ///            are not readable or there were not enough bytes.
+    public func getUUID(at index: Int) -> UUID? {
+        guard let chunk1 = self.getInteger(at: index, as: UInt64.self),
+              let chunk2 = self.getInteger(at: index + 8, as: UInt64.self) else {
```
fabianfett, line 280: I think we should use the new multiread here... :)

glbrntt: There's no `getMultipleIntegers` unfortunately. I considered reading a slice and then reading integers from that but then we're doing even more checking.

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
```diff
@@ -265,6 +265,101 @@ extension ByteBuffer {
...
+            UInt8(truncatingIfNeeded: chunk1),
+            UInt8(truncatingIfNeeded: chunk2 >> 56),
+            UInt8(truncatingIfNeeded: chunk2 >> 48),
+            UInt8(truncatingIfNeeded: chunk2 >> 40),
+            UInt8(truncatingIfNeeded: chunk2 >> 32),
+            UInt8(truncatingIfNeeded: chunk2 >> 24),
+            UInt8(truncatingIfNeeded: chunk2 >> 16),
+            UInt8(truncatingIfNeeded: chunk2 >> 8),
+            UInt8(truncatingIfNeeded: chunk2)
+        )
```
fabianfett, line 301: As we have to copy exactly 16bytes here, I think I would go for using `unsafe` APIs.

fabianfett (commented):
Some thoughts.

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
weissi: want to add it?

glbrntt: Yeah, we should add get and set. I just don't have the time right now 😔 (I only opened this PR because I already had the changes locally somewhere else and thought it would be generally useful.)

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
glbrntt: My gut feeling is that this codepath isn't common enough to warrant dropping into unsafe land and the current approach shouldn't perform terribly.

Lukasa (approved):

Sources/NIOFoundationCompat/ByteBuffer-foundation.swift
Lukasa: [It totally does](https://swift.godbolt.org/z/MvPTY4efv):

```
        mov     rdx, rsi
        mov     rax, rdi
        bswap   rdx
        bswap   rax
        ret
```

PR #2046, opened 2022-02-16
[SelectableEventLoop] SR-15872; Save allocations when appending task

fabianfett:
We should probably wait for #2047, before we merge this one.

### Motivation

- Workaround for [SR-15872](https://bugs.swift.org/browse/SR-15872). We don't want to occur an allocate for every task that shall be allocated.

### Changes

- Use `[ScheduledTask]` instead of `[() -> Void]`

### Result

- Fewer allocations = faster code

### Alternatives

We could create a struct that really only wraps the closure.

### DIFF
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -66,7 +66,10 @@ internal final class SelectableEventLoop: EventLoop {
     internal var scheduledTaskCounter = NIOAtomic.makeAtomic(value: UInt64(0))
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
-    private var tasksCopy = ContiguousArray<() -> Void>()
+
+    // We only need the ScheduledTask's task closure. However, an `Array<() -> Void>` allocates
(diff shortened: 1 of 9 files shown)


Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -66,7 +66,10 @@ internal final class SelectableEventLoop: EventLoop {
     internal var scheduledTaskCounter = NIOAtomic.makeAtomic(value: UInt64(0))
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
-    private var tasksCopy = ContiguousArray<() -> Void>()
+
+    // We only need the ScheduledTask's task closure. However, an `Array<() -> Void>` allocates
+    // for every appended closure. https://bugs.swift.org/browse/SR-15872
+    private var tasksCopy = ContiguousArray<ScheduledTask>()
```
glbrntt, line 72: Can you add this PR to https://github.com/apple/swift-nio/blob/main/docs/workarounds.md?

FranzBusch (approved):
Nice improvements! LGTM.

PR #2047, opened 2022-02-16
Add an allocation test for execute which needs to hop threads.

fabianfett:
We have a bunch of allocation tests around scheduling tasks. However we have calls execute on an `EventLoop` that we are not executing on.

### Motivation:

- Know how expensive a call to `execute` is, if we are of the `EventLoop`.

### Modifications:

- Add an allocation test for execute which needs to hop threads.

### Result:

The more allocation knowledge, the better to optimize.


glbrntt (commented):
LGTM but we need to add the allocation limits for this test

fabianfett:
@glbrntt we are also missing `MAX_ALLOCS_ALLOWED_1000_copying_bytebufferview_to_array=1000` do you mind if I add those as well?

glbrntt (approved):

PR #2049, opened 2022-02-17
ELF: provide flatMapWithEventLoop and friends (#1996)

### LINKED ISSUE #1996: consider adding a new suite of functions named `EventLoopFuture.flatMapWithEventLoop` or so
Sometimes in NIO programs you're stuck in this situation:

```swift
doSomething().flatMap { result in
    guard iLikeTheShape(of: result) else {
        struct IDontLikeIt: Error {}
        //     vvvvvvvvv---- where do I get that from?
        return eventLoop.makeFailedFuture(IDontLikeIt())
    }
    return doMore()
}
```

that's all okay, but where do we get `eventLoop` from? In many cases, it's passed in but that's not always the case, I would like a suite of new functions (everything that allows to return `-> EventLoopFuture<...>`) which are named like the existing ones with an additional `WithEventLoop` at the end. Such that I can write

```swift
doSomething().flatMapWithEventLoop { result, eventLoop /* <- thank you */ in
    guard iLikeTheShape(of: result) else {
        struct IDontLikeIt: Error {}
        return eventLoop.makeFailedFuture(IDontLikeIt())
    }
    return doMore()
}
```

The implementation would be trivial:

```swift
extension EventLoopFuture {
    @inlinable
    public func flatMapWithEventLoop<NewValue>(file: StaticString = #file, line: UInt = #line, _ callback: @escaping (Value, EventLoop) -> EventLoopFuture<NewValue>) -> EventLoopFuture<NewValue> {
        let eventLoop = self.eventLoop
        return self.flatMap(file: file, line: line) {
            callback($0, eventLoop)
        }
}
```

weissi:
### Motivation:

If you have to call an API that requires an EventLoop, it's super annoying to just continuing to use the current one. Consider this

```swift
serviceA.call().flatMap { result in
   httpClient.get(eventLoop: /* what do I put here? */)
}
```

Today, you have to write

```
let future = serviceA.call()
return future.flatMap { result in
   httpClient.get(eventLoop: future.eventLoop)
}
```

which is awkward.

### Modifications:

Provide `flatMapWithEventLoop`, `flatMapErrorWithEventLoop` and `foldWithEventLoop` the counterparts to `flatMap`, `flatMapError` and `fold` but also provide the `EventLoop`. This is the full set of combinators that return ELFs from the callback.

So you can just write

```swift
serviceA.call().flatMapWithEventLoop { result, loop in
   httpClient.get(eventLoop: loop)
}
```
### Result:

- nicer code
- fixes #1996

### DIFF
--- a/Sources/NIOCore/EventLoopFuture+WithEventLoop.swift
+++ b/Sources/NIOCore/EventLoopFuture+WithEventLoop.swift
@@ -1,0 +1,139 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension EventLoopFuture {
+    /// When the current `EventLoopFuture<Value>` is fulfilled, run the provided callback,
+    /// which will provide a new `EventLoopFuture` alongside the `EventLoop` associated with this future.
+    ///
+    /// This allows you to dynamically dispatch new asynchronous tasks as phases in a
+    /// longer series of processing steps. Note that you can use the results of the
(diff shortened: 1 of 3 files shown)


Lukasa (approved):

fabianfett (approved):
LGTM.

PR #2050, opened 2022-02-21
Make all ByteBuffer methods inlinable

Lukasa:
Motivation:

We've held off on doing this for a while on the theory that function
calls are cheap, and we're happy to make them where there is no
particular need to specialize. Unfortunately, while this is true in
general, Swift is capable of optimizing ARC and exclusivity checking
when it can better understand what a method actually does.

Additionally, we'd hoped that cross-module-optimization would be a
useful addition here. Sadly, right now this still hasn't rolled out as a
default, and even if it had, it tends to be limited to smaller methods
and generic functions, all of which we've already annotated.

Modifications:

- Add @inlinable to essentially everything.

Result:

Better codegen around ByteBuffer.


Lukasa:

glbrntt (approved):

Lukasa:
Took at look at the codegen for the only benchmark that got slower. I think we need a few judicious @inline(never)s on functions that are always slow-paths, such as the reallocation functions. I'll play around.

Lukasa:

Issue #1650, opened 2020-09-24
Provide IPAddress type


Lukasa:
If you want to work with a complete socket address type including port, we have a helper in `SocketAddress`. However, the moment you want to talk only about an IP address we force you to hold an `in_addr` or `in_addr6` structure: hardly the friendliest versions of these data types.

We should provide a helpful wrapper `IPAddress` type that can be converted into those types as needed. This would be a much friendlier way to work with IP addresses. This is a good issue for those who are familiar with either IP addresses or Swift.

shekhar-rajak:
Hi, I would like to work on this issue. I think we want a structure inherent from `SocketAddress `  to only work on IP address with name `IPAddress `

Lukasa:
I don't think we need to involve `SocketAddress`: an independent type called `NIOIPAddress` is probably sufficient.

shekhar-rajak:
Thanks I will work on it.

shekhar-rajak:
I was exploring the codebase, here is my findings : 

- [`init(ipAddress: String, port: Int)`](https://github.com/apple/swift-nio/blob/main/Sources/NIO/SocketAddresses.swift#L301) is the constructor that takes IP in string & port in Int type. So we want a new type named `NIOIPAddress`  to have the same behaviour to check ip4 and ip6 with other attributes related to IP address
- This new type `NIOIPAddress` should be class I think. 

I want to understand some of the test cases and requirements for this new type NIOIPAddress so that I can work on it as per the unit tests.

Lukasa:
Yes, we absolutely want an `init(string: String)` constructor. I think we need a few others:

- `init<Bytes: Collection>(packedBytes bytes: Bytes) where Element == UInt8`
- `init(posixIPv4Address: in_addr)`
- `init(posixIPv6Address: in6_addr)`

I don't think `NIOIPAddress` should be a `class` though, I think it should be an `enum` with two cases, each of which should be `struct`s. It certainly wants to have value semantics.

For our test cases, we need to confirm that it can accept the wide range of string formats for IP addresses:

- IPv4 dotted decimal (`a.b.c.d`)
- IPv6 segmented (hexadecimal separated by colons) (see https://datatracker.ietf.org/doc/html/rfc4291.html#section-2.2)
- IPv6 segmented with scope ID (see https://datatracker.ietf.org/doc/html/rfc4007.html#section-11)

[...]

PR #2054, opened 2022-02-23
Provide IPAddress wrapper type

SeJV:
Provide IPAddress type. 

### Motivation:

As described in #1650, we currently need to hold a libc `in_addr` or `in6_addr` to represent an IP address. Similar to `SocketAddress`, there should be an `IPAddress` helper to simplify the handling of IP addresses. 

### Modifications:

Implements `IPAddress` and `IPAddressTest`. The `IPAddress` allows initialisations from strings(IPv4 and IPv6), from libc `in_addr` and `in6_addr`, from byte collection and from a tuple of 4 (IPv4) or 16 (IPv6) bytes. 

They are stored as a tuple of `UInt8` bytes and can be read as string, `in_addr`, `in6_addr` and byte collection.  

### Result:

With this handy `IPAddress` enum for ip addresses, code that does only need to work with the IP address can now be written more expressive and is user-friendly than relying on `in_addr` and `in6_addr` for that.

### DIFF
--- a/Sources/NIOCore/IPAddresses.swift
+++ b/Sources/NIOCore/IPAddresses.swift
@@ -18,6 +18,7 @@ import CNIOLinux
 import CoreFoundation
 
 
+/// `Error` that may be thrown if we fail to create a `IPAddress`
 public enum IPAddressError: Error {
     /// Given string input is not supported IP Address
     case failedToParseIPString(String)
@@ -28,13 +29,8 @@ public enum IPAddressError: Error {
 public typealias IPv4BytesTuple = (UInt8, UInt8, UInt8, UInt8)
 public typealias IPv6BytesTuple = (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8)
 
-public struct IPv4Bytes: Collection {
-    public typealias Index = Int
-    public typealias Element = UInt8
-    
-    public var startIndex: Index { 0 }
-    public var endIndex: Index { 4 }
-    
+/// Represent the bytes for an IPv4Address.
+public struct IPv4Bytes {
     private let ipv4BytesTuple: IPv4BytesTuple
     
     public var bytes: IPv4BytesTuple {
@@ -44,6 +40,14 @@ public struct IPv4Bytes: Collection {
     init(_ bytes: IPv4BytesTuple) {
         self.ipv4BytesTuple = bytes
     }
+}
+
+extension IPv4Bytes: Collection {
+    public typealias Index = Int
+    public typealias Element = UInt8
+    
+    public var startIndex: Index { 0 }
+    public var endIndex: Index { 4 }
     
     public subscript(position: Index) -> Element {
         get {
@@ -62,13 +66,8 @@ public struct IPv4Bytes: Collection {
     }
 }
 
-public struct IPv6Bytes: Collection {
-    public typealias Index = Int
-    public typealias Element = UInt8
-    
-    public var startIndex: Index { 0 }
-    public var endIndex: Index { 16 }
-    
+/// Represent the bytes for an IPv6Address.
+public struct IPv6Bytes {
     private let ipv6BytesTuple: IPv6BytesTuple
     
     public var bytes: IPv6BytesTuple {
@@ -78,6 +77,14 @@ public struct IPv6Bytes: Collection {
     init(_ bytes: IPv6BytesTuple) {
         self.ipv6BytesTuple = bytes
     }
+}
+
+extension IPv6Bytes: Collection {
+    public typealias Index = Int
+    public typealias Element = UInt8
+    
+    public var startIndex: Index { 0 }
+    public var endIndex: Index { 16 }
     
     public subscript(position: Index) -> Element {
         get {
@@ -109,6 +116,7 @@ public struct IPv6Bytes: Collection {
 }
 
 extension UInt8 {
+    /// Convenience function especially useful for representing IPv6 bytes as readable string.
     func toHex() -> Character {
         switch self {
         case 0: return "0"
@@ -132,39 +140,52 @@ extension UInt8 {
     }
 }
 
-/// Represent a IP address
-public enum IPAddress: CustomStringConvertible {
+/// Represent a single `IPAddress`.
+public enum IPAddress {
     public typealias Element = UInt8
     
     /// A single IPv4 address for `IPAddress`.
     public struct IPv4Address {
-        /// The libc ip address for an IPv4 address.
+        /// The bytes storing the address of the IPv4 address.
         let address: IPv4Bytes
         
-        /// Get the IP address as a string
+        /// Get the `IPv4Address` as a string.
         var ipAddress: String {
             self.address.lazy.map({String($0)}).joined(separator: ".")
         }
         
+        /// Get the libc address for an IPv4 address.
         var posix: in_addr {
             get {
                 return in_addr.init(s_addr: UInt32(self.address.bytes.0) << 24
-                                          + UInt32(self.address.bytes.1) << 16
-                                          + UInt32(self.address.bytes.2) << 8
-                                          + UInt32(self.address.bytes.3))
+                                    + UInt32(self.address.bytes.1) << 16
+                                    + UInt32(self.address.bytes.2) << 8
+                                    + UInt32(self.address.bytes.3))
             }
         }
         
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - address: Bytes that hold the IPv4 address.
         fileprivate init(address: IPv4Bytes) {
             self.address = address
         }
         
-        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Element == UInt8 {
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        fileprivate init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8, Bytes.Index == Int {
             self = .init(address: .init((
                 bytes[0], bytes[1], bytes[2], bytes[3]
             )))
         }
         
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - string: String representation of an IPv4 address.
         fileprivate init(string: String) throws {
             var bytes: [UInt8] = [0,0,0,0]
             var idx: Int = 0
@@ -190,10 +211,10 @@ public enum IPAddress: CustomStringConvertible {
     
     /// A single IPv6 address for `IPAddress`
     public struct IPv6Address {
-        /// The libc ip address for an IPv6 address.
+        /// The bytes storing the address of the IPv6 address.
         let address: IPv6Bytes
         
-        /// Get the IP address as a string
+        /// Get the `IPv6Address` as a string.
         var ipAddress: String {
             stride(from: 0, to: 15, by: 2).lazy.map({ idx in
                 let hexValues = [
@@ -216,22 +237,35 @@ public enum IPAddress: CustomStringConvertible {
             }).joined(separator: ":")
         }
         
+        /// Get the libc address for an IPv6 address.
         var posix: in6_addr {
             get {
                 return in6_addr.init(__u6_addr: .init(__u6_addr8: self.address.bytes))
             }
         }
         
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - address: Bytes that hold the IPv6 address.
         fileprivate init(address: IPv6Bytes) {
             self.address = address
         }
         
-        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Element == UInt8 {
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        fileprivate init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8, Bytes.Index == Int {
             self = .init(address: .init((
                 bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7], bytes[8], bytes[9], bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
             )))
         }
         
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - string: String representation of an IPv6 address.
         fileprivate init(string: String) throws {
             var idx = 0
             var ipv6Bytes: [UInt16] = [0,0,0,0,0,0,0,0]
@@ -242,7 +276,7 @@ public enum IPAddress: CustomStringConvertible {
                 if char == ":" {
                     if isLastCharSeparator {
                         if shortenerIndex != nil {
-                            // Two shortener are not allowed
+                            // Two shortener are not allowed.
                             throw IPAddressError.failedToParseIPString(string)
                         }
                         shortenerIndex = idx
@@ -263,17 +297,17 @@ public enum IPAddress: CustomStringConvertible {
             }
             
             if let shortenerIndex = shortenerIndex {
-                // For all i after shortenerIndex move to i + (7 - idx)
+                // For all i after shortenerIndex move to i + (7 - idx).
                 let shiftBy = ipv6Bytes.count - 1 - idx
                 
-                // Go from last index backwards to the shortener position
+                // Go from last index backwards to the shortener position.
                 for i in (shortenerIndex+1...idx).reversed() {
                     ipv6Bytes[i + shiftBy] = ipv6Bytes[i]
                     ipv6Bytes[i] = 0
                 }
             } else {
                 if (idx != ipv6Bytes.count - 1) {
-                    // if IPv6 wasn't shortened, every segment should be set explicitly
+                    // if IPv6 wasn't shortened, every segment should be set explicitly.
                     throw IPAddressError.failedToParseIPString(string)
                 }
             }
@@ -298,7 +332,7 @@ public enum IPAddress: CustomStringConvertible {
         }
     }
     
-    /// Get the IP address as a string
+    /// Get the `IPAddress` as a string.
     public var ipAddress: String {
         switch self {
         case .v4(let addr):
@@ -319,7 +353,7 @@ public enum IPAddress: CustomStringConvertible {
     }
     
     /// Creates a new `IPAddress` for the given string.
-    /// "d.d.d.d" with decimal values for IPv4 and "h:h:h:h:h:h:h:h" with hexadecimal values for IPv6. Hybrid versions for IPv6 are not (yet) supported.
+    /// "d.d.d.d" with decimal values for IPv4 and "h:h:h:h:h:h:h:h" with hexadecimal values for IPv6. Also allowing for shortened IPv6 representation with "::". Hybrid versions for IPv6 are not (yet) supported.
     ///
     /// - parameters:
     ///     - string: String representation of IPv4 or IPv6 Address
@@ -333,13 +367,12 @@ public enum IPAddress: CustomStringConvertible {
         }
     }
     
-    /// Creates a new `IPAddress` for the given bytes and optional zone.
+    /// Creates a new `IPAddress` for the given bytes.
     ///
     /// - parameters:
     ///     - bytes: Either 4 or 16 bytes representing the IPAddress value.
     /// - returns: The `IPAddress` for the given string or `nil` if the string representation is not supported.
-    // TODO: update to init<Bytes: Collection>(packedBytes bytes: Bytes) where Element == UInt8
-    public init<Bytes: Collection>(packedBytes bytes: Bytes) where Element == UInt8 {
+    public init<Bytes: Collection>(packedBytes bytes: Bytes) throws where Bytes.Element == UInt8, Bytes.Index == Int {
         switch bytes.count {
         case 4: self = .v4(.init(packedBytes: bytes))
         case 16: self = .v6(.init(packedBytes: bytes))
@@ -348,6 +381,10 @@ public enum IPAddress: CustomStringConvertible {
         }
     }
     
+    /// Creates a new `IPAddress` for the given libc IPv4 address.
+    ///
+    /// - parameters:
+    ///     - posixIPv4Address: libc ipv4 address.
     public init(posixIPv4Address: in_addr) {
         let uint8Bitmask: UInt32 = 0x000000FF
         
@@ -361,19 +398,23 @@ public enum IPAddress: CustomStringConvertible {
         self = .v4(.init(address: uint8AddressBytes))
     }
     
+    /// Creates a new `IPAddress` for the given libc IPv6 address.
+    ///
+    /// - parameters:
+    ///     - posixIPv6Address: libc ipv6 address.
     public init(posixIPv6Address: in6_addr) {
         self = .v6(.init(address: .init(posixIPv6Address.__u6_addr.__u6_addr8)))
     }
 }
 
-/// We define an extension on `IPv4Bytes` that gives it an elementwise equatable conformance
+/// We define an extension on `IPv4Bytes` that gives it an equatable conformance.
 extension IPv4Bytes: Equatable {
     public static func == (lhs: IPv4Bytes, rhs: IPv4Bytes) -> Bool {
         return lhs.bytes == rhs.bytes
     }
 }
 
-/// We define an extension on `IPv6Bytes` that gives it an elementwise equatable conformance
+/// We define an extension on `IPv6Bytes` that gives it an element wise equatable conformance.
 extension IPv6Bytes: Equatable {
     public static func == (lhs: IPv6Bytes, rhs: IPv6Bytes) -> Bool {
         return zip(lhs, rhs).allSatisfy {$0 == $1}
@@ -385,35 +426,22 @@ extension IPAddress.IPv6Address: Equatable {}
 extension IPAddress: Equatable {}
 
 
+/// We define an extension on `IPv4Bytes` that combines each byte to the hasher.
 extension IPv4Bytes: Hashable {
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(self.bytes.0)
-        hasher.combine(self.bytes.1)
-        hasher.combine(self.bytes.2)
-        hasher.combine(self.bytes.3)
+        self.forEach { hasher.combine($0) }
     }
 }
+
+/// We define an extension on `IPv6Bytes` that combines each byte to the hasher.
 extension IPv6Bytes: Hashable {
     public func hash(into hasher: inout Hasher) {
-        hasher.combine(self.bytes.0)
-        hasher.combine(self.bytes.1)
-        hasher.combine(self.bytes.2)
-        hasher.combine(self.bytes.3)
-        hasher.combine(self.bytes.4)
-        hasher.combine(self.bytes.5)
-        hasher.combine(self.bytes.6)
-        hasher.combine(self.bytes.7)
-        hasher.combine(self.bytes.8)
-        hasher.combine(self.bytes.9)
-        hasher.combine(self.bytes.10)
-        hasher.combine(self.bytes.11)
-        hasher.combine(self.bytes.12)
-        hasher.combine(self.bytes.13)
-        hasher.combine(self.bytes.14)
-        hasher.combine(self.bytes.15)
+        self.forEach { hasher.combine($0) }
     }
 }
 
 extension IPAddress.IPv4Address: Hashable {}
 extension IPAddress.IPv6Address: Hashable {}
 extension IPAddress: Hashable {}
+
+extension IPAddress: CustomStringConvertible {}
--- a/Sources/NIOCore/IPAddresses.swift
+++ b/Sources/NIOCore/IPAddresses.swift
@@ -72,7 +72,16 @@ public enum IPAddress: CustomStringConvertible {
             addressString = "\(addr.address)"
             type = "IPv4"
         case .v6(let addr):
-            let hexRepresentation = Mirror(reflecting: addr.address).children.map {String(format: "%02X", $0.value as! UInt8)}
+            let hexRepresentation: [UInt16] = [
+                addr.address.0 | addr.address.1 << 8,
+                addr.address.2 | addr.address.3 << 8,
+                addr.address.4 | addr.address.5 << 8,
+                addr.address.6 | addr.address.7 << 8,
+                addr.address.8 | addr.address.9 << 8,
+                addr.address.10 | addr.address.11 << 8,
+                addr.address.12 | addr.address.13 << 8,
+                addr.address.14 | addr.address.15 << 8
+            ].map {String(format: "%02X", $0)}
             addressString = "\(hexRepresentation.joined(separator: ":"))"
             
             if let zone = addr.zone {
@@ -99,12 +108,9 @@ public enum IPAddress: CustomStringConvertible {
     
     public init(bytes: [UInt8]) {
         if bytes.count == 16 {
-            var byteTuple: IPv6Bytes = (0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0)
-            withUnsafeMutablePointer(to: &byteTuple) { dst -> () in
-                memcpy(dst, bytes, 16)
-                return
-            }
-            self = .v6(.init(address: byteTuple))
+            self = .v6(.init(address: (
+                bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7], bytes[8], bytes[9], bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
+            )))
         } else {
             // TODO: throw exception
             self = .v4(.init(address: (0,0,0,0)))
--- a/Sources/NIOCore/IPAddresses.swift
+++ b/Sources/NIOCore/IPAddresses.swift
@@ -1,0 +1,124 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Darwin
+import CNIOLinux
+import CoreFoundation
+
+public typealias IPv4Bytes = (UInt8, UInt8, UInt8, UInt8)
+public typealias IPv6Bytes = (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8)
+
+
+/// Represent a IP address
+public enum IPAddress: CustomStringConvertible {
+    
+    /// A single IPv4 address for `IPAddress`.
+    public struct IPv4Address {
+        /// The libc ip address for an IPv4 address.
+        /// 8b.8b.8b.8b => 32b Int
+        private let _storage: IPv4Bytes
+        
+        public var address: IPv4Bytes {
+            return _storage
+        }
+    
+        fileprivate init(address: IPv4Bytes) {
+            self._storage = address
+        }
+    }
+    
+    /// A single IPv6 address for `IPAddress`
+    public struct IPv6Address {
+        /// The libc ip address for an IPv6 address.
+        private let _storage: Box<(address: IPv6Bytes, zone: String?)>
+        
+        public var address: IPv6Bytes {
+            return self._storage.value.address
+        }
+        
+        public var zone: String? {
+            return self._storage.value.zone
+        }
+    
+        fileprivate init(address: IPv6Bytes, zone: String?=nil) {
+            self._storage = Box((address: address, zone: zone))
+        }
+    }
+        
+    /// An IPv4 `IPAddress`.
+    case v4(IPv4Address)
+
+    /// An IPv6 `IPAddress`.
+    case v6(IPv6Address)
+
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+    /// Given string input is not supported IP Address
+    case bytesArrayHasWrongLength(Int)
+}
+
+public typealias IPv4BytesTuple = (UInt8, UInt8, UInt8, UInt8)
+public typealias IPv6BytesTuple = (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8)
+
+/// Represent the bytes for an `IPv4Address`.
+public struct IPv4Bytes {
+    private let ipv4BytesTuple: IPv4BytesTuple
```
Lukasa, line 34: For both this and the `IPv6Bytes` struct, this should be `var` and should probably be `public`. We can even make it the `bytes` accessor and simply remove that. Additionally, the `init` should be `public` for both types.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        self.ipv6BytesTuple = bytes
+    }
+}
+
+/// Represent a single `IPAddress`.
+public enum IPAddress {
+    /// A single IPv4 address for `IPAddress`.
+    public struct IPv4Address {
+        /// The bytes storing the address of the IPv4 address.
+        let address: IPv4Bytes
```
Lukasa, line 63: As a nit, this should be `var`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            self.address.lazy.map({String($0)}).joined(separator: ".")
+        }
+        
+        /// Get the libc address for an IPv4 address.
+        var posix: in_addr {
+            get {
+                return in_addr.init(s_addr: UInt32(self.address.bytes.0) << 24
+                                    + UInt32(self.address.bytes.1) << 16
+                                    + UInt32(self.address.bytes.2) << 8
+                                    + UInt32(self.address.bytes.3))
```
Lukasa, line 76: Can we move this `UInt32` math into a `UInt32.init` function? It's a bit unfortunate to bury it here.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+
+/// Represent a single `IPAddress`.
+public enum IPAddress {
+    /// A single IPv4 address for `IPAddress`.
+    public struct IPv4Address {
+        /// The bytes storing the address of the IPv4 address.
+        let address: IPv4Bytes
+        
+        /// Get the `IPv4Address` as a string.
+        var ipAddress: String {
```
Lukasa, line 66: This should probably be moved into a `var description` on a `CustomStringConvertible` conformance.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        ///   - address: Bytes that hold the IPv4 address.
+        fileprivate init(address: IPv4Bytes) {
+            self.address = address
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        fileprivate init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8, Bytes.Index == Int {
```
Lukasa, line 92: There's no reason to constrain this to having an `Int` index: we just want the first 4, and we can get them through generic programming. This should also be `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+                                    + UInt32(self.address.bytes.2) << 8
+                                    + UInt32(self.address.bytes.3))
+            }
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - address: Bytes that hold the IPv4 address.
+        fileprivate init(address: IPv4Bytes) {
```
Lukasa, line 84: This should be `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            self = .init(address: .init((
+                bytes[0], bytes[1], bytes[2], bytes[3]
+            )))
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - string: String representation of an IPv4 address.
+        fileprivate init(string: String) throws {
```
Lukasa, line 102: This should be `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            var bytes: [UInt8] = [0,0,0,0]
+            var idx: Int = 0
+            
+            for char in string {
```
Lukasa, line 106: Let's not compute on the `String` as a collection: it's slow, and it only gives us features we don't want (supporting non-ASCII characters). We can use the `UTF8View` (`string.utf8`) and process the characters as bytes directly.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+                throw IPAddressError.failedToParseIPString(string)
+            }
+            self = .init(packedBytes: bytes)
+        }
+    }
+    
+    /// A single IPv6 address for `IPAddress`
+    public struct IPv6Address {
+        /// The bytes storing the address of the IPv6 address.
+        let address: IPv6Bytes
```
Lukasa, line 128: Should be `var` and `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+                if hexValues[0] + hexValues[1] + hexValues[2] == 0 {
+                    removeLeadingZeros = 3
+                } else if hexValues[0] + hexValues[1] == 0 {
+                    removeLeadingZeros = 2
+                } else if hexValues[0] == 0 {
+                    removeLeadingZeros = 1
+                }
+                
+                return String(hexValues[removeLeadingZeros...].lazy.map {$0.toHex()})
+            }).joined(separator: ":")
```
Lukasa, line 150: This doesn't correctly handle runs of leading zeros in the middle, where it will emit those as `0:0:0:0` instead of as `::`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        }
+        
+        /// Get the libc address for an IPv6 address.
+        var posix: in6_addr {
```
Lukasa, line 154: No reason for this not to be `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            get {
+                return in6_addr.init(__u6_addr: .init(__u6_addr8: self.address.bytes))
+            }
+        }
+        
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - address: Bytes that hold the IPv6 address.
+        fileprivate init(address: IPv6Bytes) {
```
Lukasa, line 164: This can be `public` too.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            self.address = address
+        }
+        
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        fileprivate init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8, Bytes.Index == Int {
```
Lukasa, line 172: Same note here: this should be `public` and not rely on having an `Int` index.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            self = .init(address: .init((
+                bytes[0], bytes[1], bytes[2], bytes[3], bytes[4], bytes[5], bytes[6], bytes[7], bytes[8], bytes[9], bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
+            )))
+        }
+        
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - string: String representation of an IPv6 address.
+        fileprivate init(string: String) throws {
```
Lukasa, line 182: This should also be `public`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            var idx = 0
+            var ipv6Bytes: [UInt16] = [0,0,0,0,0,0,0,0]
+            var isLastCharSeparator: Bool = false
+            var shortenerIndex: Int?
+            
+            for char in string {
```
Lukasa, line 188: Same note here, this should compute against the UTF8 view.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        switch self {
+        case .v4(_):
+            return "[IPv4]\(self.ipAddress)"
+        case .v6(_):
+            return "[IPv6]\(self.ipAddress)"
+        }
+    }
+    
+    /// Get the `IPAddress` as a string.
+    public var ipAddress: String {
```
Lukasa, line 249: Nit: this should probably be `ipAddressString`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+            try self = .v6(.init(string: string))
+        }
+    }
+    
+    /// Creates a new `IPAddress` for the given bytes.
+    ///
+    /// - parameters:
+    ///     - bytes: Either 4 or 16 bytes representing the IPAddress value.
+    /// - returns: The `IPAddress` for the given string or `nil` if the string representation is not supported.
+    public init<Bytes: Collection>(packedBytes bytes: Bytes) throws where Bytes.Element == UInt8, Bytes.Index == Int {
```
Lukasa, line 288: Same note here: remote the `Index` constraint.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+    /// - parameters:
+    ///     - posixIPv4Address: libc ipv4 address.
+    public init(posixIPv4Address: in_addr) {
+        let uint8Bitmask: UInt32 = 0x000000FF
+        
+        let uint8AddressBytes: IPv4Bytes = .init((
+            UInt8((posixIPv4Address.s_addr >> 24) & uint8Bitmask),
+            UInt8((posixIPv4Address.s_addr >> 16) & uint8Bitmask),
+            UInt8((posixIPv4Address.s_addr >> 8) & uint8Bitmask),
+            UInt8(posixIPv4Address.s_addr & uint8Bitmask)
```
Lukasa, line 308: Nit: we don't need the bit mask here, we can use `UInt8(truncatingIfNeeded:)` instead.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+    /// - parameters:
+    ///     - posixIPv6Address: libc ipv6 address.
+    public init(posixIPv6Address: in6_addr) {
+        self = .v6(.init(address: .init(posixIPv6Address.__u6_addr.__u6_addr8)))
+    }
+}
+
+extension UInt8 {
+    /// Convenience function especially useful for representing IPv6 bytes as readable string.
+    func toHex() -> Character {
```
Lukasa, line 325: Transforming to `Character` is almost never the right move: this costs us a lot of performance. We can and should construct strings from UTF8 bytes, so this should map to the ASCII encoding for a given hex character using `UInt8(ascii: "0")` and friends.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        case 13: return "D"
+        case 14: return "E"
+        case 15: return "F"
+        default: preconditionFailure()
+        }
+    }
+}
+
+/// We define an extension on `IPv4Bytes` that gives it a collection conformance.
+extension IPv4Bytes: Collection {
```
Lukasa, line 349: This should conform to `MutableCollection` (by adding a `set` to our subscript) and `RandomAccessCollection`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+    
+    public func index(after: Index) -> Index {
+        return after + 1
+    }
+}
+
+/// We define an extension on `IPv6Bytes` that gives it a collection conformance.
+extension IPv6Bytes: Collection {
```
Lukasa, line 374: Same here, this should conform to `MutableCollection` and `RandomAccessCollection`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,447 @@
...
+        }
+    }
+    
+    public func index(after: Index) -> Index {
+        return after + 1
+    }
+}
+
+/// We define an extension on `IPv4Bytes` that gives it an equatable conformance.
+extension IPv4Bytes: Equatable {
```
Lukasa, line 411: We can use a synthesised conformance to `Equatable` and `Hashable`, we do not need to override these.

Lukasa (commented):
Thanks for this! I've left a first round of notes in the diff. There'll be some further notes on the algorithmical stuff in a later round of feedback, but I wanted to start by dealing with some of the code layout and style questions first.

Lukasa:

Sources/NIOCore/IPAddresses.swift
SeJV: For Tuple those synthesised conformance of `Equatable` and `Hashable` is not (yet?) available.

SeJV:
Hey, thanks for the notes! 
I resolved them and also updated the string generation algorithm for IPv6 Address.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+        ///   - address: Bytes that hold the IPv4 address.
+        public init(address: IPv4Bytes) {
+            self.address = address
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8 {
```
Lukasa, line 86: Can we make this `@inlinable`?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+            
+            self = .init(address: ipv4Bytes)
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - string: String representation of an IPv4 address.
+        public init(string: String) throws {
+            var bytes: [UInt8] = [0,0,0,0]
```
Lukasa, line 101: We could probably do this directly on top of `IPv4Bytes` to save the array allocation.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+                    let number = char - UInt8(ascii: "0")
+                    if idx > 3 || bytes[idx] > 25 || (255 - number < (bytes[idx] * 10)) {
+                        throw IPAddressError.failedToParseIPString(string)
+                    }
+                    bytes[idx] = bytes[idx] * 10 + number
+                } else {
+                    throw IPAddressError.failedToParseIPString(string)
+                }
+            }
+            if idx != 3 {
```
Lukasa, line 117: I don't think this algorithm is quite right: it would allow `...` as a valid IPv4 address, which it should not. Can we add some test cases that hit these edge cases and confirm that we're doing the appropriate thing?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+            for (idx, value) in bucketValues.enumerated() {
+                switch value {
+                case 0: consecutiveCounter += 1
+                default: consecutiveCounter = 0
+                }
+                consecutiveZeros[idx] = consecutiveCounter
+            }
+            let mostConsecutiveZeros = consecutiveZeros.max()
+            
+            if let mostConsecutiveZeros = mostConsecutiveZeros {
```
Lukasa, line 149: When will this ever be `nil`?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+            // find most consecutive zeros and remove them
+            var consecutiveZeros = [0, 0, 0, 0, 0, 0, 0, 0]
+            var consecutiveCounter = 0
+            for (idx, value) in bucketValues.enumerated() {
+                switch value {
+                case 0: consecutiveCounter += 1
+                default: consecutiveCounter = 0
+                }
+                consecutiveZeros[idx] = consecutiveCounter
+            }
```
Lukasa, line 146: Do we need the Array? Presumably we need to keep track of the last best candidate, and then our current state, and if at any point our current state ends up better than the last best candidate we can overwrite it, right?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+        }
+        
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8 {
+            var ipv6Bytes = IPv6Bytes((0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0))
+            
+            for (idx, elt) in bytes.prefix(16).enumerated() {
```
Lukasa, line 196: Rather than use a `prefix` here we should `throw` if the collection does not contain exactly 16 elements.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+        }
+        
+        /// Creates a new `IPv4Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8 {
+            var ipv4Bytes = IPv4Bytes((0,0,0,0))
+            
+            for (idx, elt) in bytes.prefix(4).enumerated() {
```
Lukasa, line 89: Rather than use `prefix` here we should throw if the collection does not contain exactly 4 elements.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+        ///   - address: Bytes that hold the IPv6 address.
+        public init(address: IPv6Bytes) {
+            self.address = address
+        }
+        
+        /// Creates a new `IPv6Address`.
+        ///
+        /// - parameters:
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        public init<Bytes: Collection>(packedBytes bytes: Bytes) where Bytes.Element == UInt8 {
```
Lukasa, line 193: Can we make this `@inlinable`?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+        ///   - string: String representation of an IPv4 address.
+        public init(string: String) throws {
+            var bytes: [UInt8] = [0,0,0,0]
+            var idx: Int = 0
+            
+            for char in string.utf8 {
+                if char == UInt8(ascii: ".") {
+                    idx += 1
+                } else if UInt8(ascii: "0")...UInt8(ascii: "9") ~= char {
+                    let number = char - UInt8(ascii: "0")
```
Lukasa, line 108: Should we use `UInt8(asciiValue:)` here?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+            try self = .v6(.init(string: string))
+        }
+    }
+    
+    /// Creates a new `IPAddress` for the given bytes.
+    ///
+    /// - parameters:
+    ///     - bytes: Either 4 or 16 bytes representing the IPAddress value.
+    /// - returns: The `IPAddress` for the given string or `nil` if the string representation is not supported.
+    public init<Bytes: Collection>(packedBytes bytes: Bytes) throws where Bytes.Element == UInt8 {
```
Lukasa, line 313: Can we make this `@inlinable` as well?

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,577 @@
...
+            asciiCompressed = asciiHex[1...]
+        case 4096...65535:
+            asciiCompressed = asciiHex[...]
+        default: preconditionFailure()
+        }
+        
+        // append null termination for cString
+        asciiCompressed.append(0)
+        
+        self = asciiCompressed.withUnsafeBufferPointer { bytes -> String in
```
Lukasa, line 440: We don't need this, we can use `String(decoding:as:)` instead.

Lukasa (commented):
Sorry for the delay there, but this is coming along really nicely! I've left a few more notes in the diff.

Sources/NIOCore/IPAddresses.swift
SeJV: `UInt8(asciiValue:)` gives the UInt8 value from a hex character that is represented in ascii. 
Here we want the ascii number for a given character, which goes the other direction. 

I can see the the confusion as the naming is not perfect. I changed the `UInt8(asciiValue:)` to `UInt8(hexAsciiValue:)` and added a line of comment. But if you have suggestions to make it more clear with the naming Im happy to update it.

SeJV:
Hey, I resolved the latest notes 🙂

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+//===----------------------------------------------------------------------===//
+
+
+import Darwin
+import CNIOLinux
+import CoreFoundation
+
+
+/// `Error` that may be thrown if we fail to create a `IPAddress`
+public enum IPAddressError: Error {
```
Lukasa, line 22: New top-level types need to be prefixed with `NIO`. As there are quite a few new top-level types in this proposal, we may want to nest them all within `public enum IPAddress` (which will have to become `public enum NIOIPAddress`), as that avoids us needing to namespace them all.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+        ///   - packedBytes: Collection of UInt8 that holds the address.
+        @inlinable
+        public init<Bytes: Collection>(packedBytes bytes: Bytes) throws where Bytes.Element == UInt8 {
+            if bytes.count != 4 {
+                throw IPAddressError.bytesArrayHasWrongLength(bytes.count)
+            }
+            
+            var ipv4Bytes = IPv4Bytes((0,0,0,0))
+            
+            for (idx, elt) in bytes.prefix(4).enumerated() {
```
Lukasa, line 94: We can remove `prefix(4)` here: it doesn't achieve anything with the check above.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+                }
+            }
+            
+            if mostConsecutiveZeros > 0 {
+                let startCompression = mostConsecutiveZerosIndex - mostConsecutiveZeros + 1
+                
+                switch (startCompression, mostConsecutiveZerosIndex) {
+                case (0, 7): bucketStrings.replaceSubrange(startCompression...mostConsecutiveZerosIndex, with: ["::"])
+                case (0, _): bucketStrings.replaceSubrange(startCompression...mostConsecutiveZerosIndex, with: [":"])
+                case (_, 7): bucketStrings.replaceSubrange(startCompression...mostConsecutiveZerosIndex, with: [":"])
```
Lukasa, line 173: Minor style nit, but we can fold these two cases together.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+        case UInt8(ascii: "6"): self = 6
+        case UInt8(ascii: "7"): self = 7
+        case UInt8(ascii: "8"): self = 8
+        case UInt8(ascii: "9"): self = 9
+        case UInt8(ascii: "A"): self = 10
+        case UInt8(ascii: "B"): self = 11
+        case UInt8(ascii: "C"): self = 12
+        case UInt8(ascii: "D"): self = 13
+        case UInt8(ascii: "E"): self = 14
+        case UInt8(ascii: "F"): self = 15
```
Lukasa, line 382: We probably need to support lower-case here too.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+
+extension UInt16 {
+    /// Creates an integer from the given `UInt8` tuple.
+    init(uint8Tuple: (UInt8, UInt8)) {
+        self = UInt16(uint8Tuple.0) << 8
+             + UInt16(uint8Tuple.1)
+    }
+    
+    /// Interprets value as hex and creates ascii `UInt8`.
+    func asciiValue() -> UInt8 {
```
Lukasa, line 406: This can just be `var`.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+            case 0: self.bytes.0 = value
+            case 1: self.bytes.1 = value
+            case 2: self.bytes.2 = value
+            case 3: self.bytes.3 = value
+            default: preconditionFailure()
+            }
+        }
+    }
+    
+    public func index(after: Index) -> Index {
```
Lukasa, line 489: I don't think we need this.

Sources/NIOCore/IPAddresses.swift
```diff
@@ -0,0 +1,586 @@
...
+            case 12: self.bytes.12 = value
+            case 13: self.bytes.13 = value
+            case 14: self.bytes.14 = value
+            case 15: self.bytes.15 = value
+            default: preconditionFailure()
+            }
+        }
+    }
+    
+    public func index(after: Index) -> Index {
```
Lukasa, line 547: I don't think we need this.

Lukasa (commented):
Cool this is really coming along. I've left a few more notes in the diff, but things are really just stylistic now.

SeJV:
Hey, I resolved the newest review round 😄

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+
+/// Represent a single `NIOIPAddress`.
+public enum NIOIPAddress: CustomStringConvertible {
+    /// `Error` that may be thrown if we fail to create a `NIOIPAddress`
+    public enum IPAddressError: Error {
+        /// Given string input is not supported IP Address
+        case failedToParseIPString(String)
+        /// Given string input is not supported IP Address
+        case bytesArrayHasWrongLength(Int)
+    }
```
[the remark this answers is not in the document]
Lukasa: I think it's reasonable to provide errors: they can always be converted to `nil` when users don't care about the information, but they can't get that information back from a `nil` later.

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+    
+    /// A single IPv4 address for `NIOIPAddress`.
+    public struct IPv4Address: CustomStringConvertible {
+        /// The bytes storing the address of the IPv4 address.
+        public var address: IPv4Bytes
+        
+        /// Get the `IPv4Address` as a string.
+        public var ipAddressString: String {
+            self.address.lazy.map({String($0)}).joined(separator: ".")
```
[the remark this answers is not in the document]
Lukasa: For now efficiency concerns are not my biggest concern. I'd rather land this with an API shape we're happy with and iterate on its performance than tie up this PR in review forever.

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+        }
+        
+        /// A human-readable description of this `IPv4Address`. Mostly useful for logging.
+        public var description: String {
+            return "[IPv4]\(self.ipAddressString)"
```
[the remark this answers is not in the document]
Lukasa: It does, but this is not marked `LosslessStringConvertible`.

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+
+extension UInt16 {
+    /// Creates an integer from the given `UInt8` tuple.
+    init(uint8Tuple: (UInt8, UInt8)) {
+        self = UInt16(uint8Tuple.0) << 8
+             + UInt16(uint8Tuple.1)
+    }
+    
+    /// Interprets value as hex and creates ascii `UInt8`.
+    var asciiValue: UInt8 {
```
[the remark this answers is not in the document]
Lukasa: There's absolutely no reason to use unsafe code for this.

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    /// Represent the bytes for an `IPv6Address.
+    public struct IPv6Bytes {
+        public var bytes: IPv6BytesTuple
+        
+        public init(_ bytes: IPv6BytesTuple) {
+            self.bytes = bytes
+        }
+    }
```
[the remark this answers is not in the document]
Lukasa: These types have a collection conformance hung off them, while the IP addresses themselves do not. I think that's a reasonable API design choice.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: It does cost, though, both in binary size for the extra type and its conformances and in code-paths which cannot be merged because they have to throw different errors (one of which containing an ARC-managed String).

In my own parser, I've seen reasonable gains by specialising errors away via generics, and if I were writing it again, I wouldn't include them at all. How is anybody going to handle the error that the bytes array had the wrong length, if they got it wrong the first time? Nobody guesses when it comes to the bytes length - either it's correct or something irrecoverably bad happened.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: Is it the _right_ API design choice, though? Especially when you consider not just the type, but the extra conformances as well.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: But is that the right design for this API? Is the `[IPv4]` prefix actually meaningful?

If it is for debugging purposes only, perhaps `CustomDebugStringConvertible` conformance and a `debugDescription` is appropriate. Personally, I think the addresses are visually distinct enough that the prefix can go and nobody will be confused, and that re-parsability is a much more desirable feature. But that's just MHO, of course.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: Sure. Just pointing it out.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: The `ASCII(_unsafeUnchecked: ...)` is an internal WebURL API which is not "unsafe" as I think you're thinking of "unsafe".

It's called "unsafe" to draw attention to the fact that it does not validate that values are <128. There are reasons (beyond the scope of this PR) why it could be problematic for such values to exist, and why additional precautions are necessary when using that API, to ensure they are never created (e.g. creating a value with a byte taken from a string literal, as here, is safe and doesn't need a run-time check).

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
Lukasa: Broadly my instinct is yes. Tuples are profoundly trash: their inability to expose a collection interface makes working with them a chore, their inability to be named or evolved makes them a trap for API evolution, and using them prominently in the API usually invites creating an alternative representation at some point down the road.

In the case of the extra type I consider that a plus, not a minus: nominal types are good for all the reasons tuples are bad. As for the conformances, I also don't consider conformances to be inherently a bad thing. Yes, they have a binary size impact, but for server applications that impact is substantially less important than it is for client use-cases, and in the case of this specific type the conformances are justified by the utility they bring (for example, allowing users to compute on the bytes of the tuple easily).

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
Lukasa: I'm open to removing them here, as we do add them in `SocketAddress` anyway.

Sources/NIOCore/NIOIPAddresses.swift
```diff
@@ -0,0 +1,583 @@
...
+            self.address.lazy.map({String($0)}).joined(separator: ".")
+        }
+        
+        /// A human-readable description of this `IPv4Address`. Mostly useful for logging.
+        public var description: String {
+            return "[IPv4]\(self.ipAddressString)"
+        }
+        
+        /// Get the libc address for an IPv4 address.
+        public var posix: in_addr {
```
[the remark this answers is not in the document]
Lukasa: Sadly I think for now this belongs here: we don't support a platform where `in_addr` is not a real type. `SocketAddress` already baked this into the system.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
Lukasa: Granted, that makes sense. I agree this can be faster, but again I'd like to land this PR first and iterate the performance later.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: Yeah that's what I thought, too - but then I figured all callers would need to handle the failure either way, and actually, the more common thing is for people to parse hostnames first as an IP address and treat it as a domain as a fallback. So they'll never even see the detailed error. And if there is no fallback, the caller will need to handle it in some other way, perhaps by throwing their own "invalid IP address <somestring>" or "invalid hostname <somestring>". But it's their choice and adding an error type doesn't actually add anything.

It's analogous to `Int(String)`. It doesn't bother throwing detailed errors; it isn't worth it, and there's not realistically anything you can do to handle it, so it's better to be a failible initialiser.

That's not to mention that it's a public enum, so it couldn't evolve to be any more detailed than the current 2 cases.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: If Swift had fixed-size arrays (i.e. tuples that conform to collection), that would be much better. In the case of IP addresses, I don't think evolution is really a problem though - and it is actually the role of the IP address type to be that wrapper over the tuple which gives its pieces meaning. If there is a compelling reason for a collection-of-bytes API, IMO it should be exposed as a 'view' of the IP address.

Another issue to be aware of is that storing these as tuples of bytes may cause alignment problems and other code-gen difficulties. I'm thinking about changing WebURL's implementation to store a `UInt32` and `(UInt64, UInt64)` to guarantee that operations such as equality and comparison lower to simple integer operations rather than expensive byte-wise comparison (it is surprisingly difficult to do that).

Switching between stored and computed properties is API compatible so this is just something to be aware of, although the required changes would be significant so may be worth doing it now.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
Lukasa: I don't disagree that those use-cases exist: I'm just leaning much more strongly towards having more diagnostic information and letting users throw it away, not less. The trade off between failable initialisers and throwing ones is a tight space, and I definitely think this is in a "gut feel" situation. If we find it's totally not worth it, we can always deprecate and replace the API.

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
Lukasa: > In the case of IP addresses, I don't think evolution is really a problem though

API evolution is basically always a problem. It's very hard to know when a language feature is going to come along that you'd like to adopt.

> and it is actually the role of the IP address type to be that wrapper over the tuple which gives its pieces meaning. If there is a compelling reason for a collection-of-bytes API, IMO it should be exposed as a 'view' of the IP address.

I agree, and IMO that's the shape of the current API. The name of the accessor to the view is `.address`, which is bad, but if it were `.bytes` then I think the API would be exactly what you're looking for.

> WebURL exposes the bytes using the tuple elements, and it works well IMO. I'm not sure full Collection conformance is totally necessary beyond this, but maybe you think it is?

I do: a bunch of APIs are spelled in terms of Collection and related protocols, and it's nice to be able to pass the type directly in. Tuples either require wrapping or unsafe code: it's nicer for us to write the few lines of code required to do the wrapping once. In this instance we aren't forbidding access to the tuple: we're just prioritising access as a `Collection`.

As to the layout: yes, I agree, we could swap to using other representations (and my original discussion of the issue proposed that representation), but we can straightforwardly change it later so I'm not too fussed right now.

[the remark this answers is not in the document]
karwa: `.bytes` view sounds good!

Sources/NIOCore/NIOIPAddresses.swift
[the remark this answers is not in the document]
karwa: It's totally up to you and I'm not arguing with it, but I just wanted to add that this:

> I'm just leaning much more strongly towards having more diagnostic information and letting users throw it away, not less

Is what I used to think as well. But it turns out, at scale across complex parsers with tight performance requirements, there can be a significant cost to that because it means the compiler effectively can't merge loops and perform other optimisations; it's tightly bound to your code pretty-much as you wrote it.

Then when you add the additional types, conformances and whatnot, the awkwardness in building an interface that can evolve... it totally turned me in the other direction. Add errors where you really expect they might meaningfully be handled or used in user-facing diagnostics, or you need some clearer indication at the call-site of a failable operation. That's my approach these days.

PR #2055, opened 2022-03-01
Rebind less load more

Lukasa:
Reduce the amount of rebinding we do with sockaddr types.

### Motivation:

[SR-15912](https://bugs.swift.org/browse/SR-15912) recently caused a bunch of issues with compilation of NIO in the nightlies. While these were arguably Swift bugs, we're also skating pretty close to the edge by using `withMemoryRebound` here: it's often safer for us to load the structures, rather than use random pointers.

### Modifications:

- Replace a bunch of rebindings with loads

### Result:

Fewer calls with `withMemoryRebound`.

### DIFF
--- a/Sources/NIOPosix/BaseSocket.swift
+++ b/Sources/NIOPosix/BaseSocket.swift
@@ -82,9 +82,9 @@ extension sockaddr_storage {
     /// Converts the `socketaddr_storage` to a `sockaddr_in`.
     ///
     /// This will crash if `ss_family` != AF_INET!
-    mutating func convert() -> sockaddr_in {
+    func convert() -> sockaddr_in {
         precondition(self.ss_family == NIOBSDSocket.AddressFamily.inet.rawValue)
-        return withUnsafeMutableBytes(of: &self) {
+        return withUnsafeBytes(of: self) {
             $0.load(as: sockaddr_in.self)
         }
     }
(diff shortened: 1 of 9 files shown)


Sources/NIOPosix/BaseSocket.swift
```diff
@@ -84,10 +84,8 @@ extension sockaddr_storage {
     /// This will crash if `ss_family` != AF_INET!
     mutating func convert() -> sockaddr_in {
         precondition(self.ss_family == NIOBSDSocket.AddressFamily.inet.rawValue)
-        return withUnsafePointer(to: &self) {
-            $0.withMemoryRebound(to: sockaddr_in.self, capacity: 1) {
-                $0.pointee
-            }
+        return withUnsafeMutableBytes(of: &self) {
```
glbrntt, line 87: I think this can be `withUnsafeBytes(of:)` (and elsewhere)

Lukasa: This func is mutating because on older Swift releases we didn't have a version of `withUnsafeBytes` that did _not_ take an `inout` as an argument. It's possible we dropped all the Swift releases where that's true, let me check.

Lukasa: Seems to be fine, I've updated the PR with removal of the mutatings.

glbrntt (approved):
Nice one!

Lukasa:

glessard (approved):
I like your improvements to my drive-by fix.

Desdu1184 (approved):

Lukasa:

Issue #1827, opened 2021-04-28
CircularBuffer: implement Collection/Sequence special customisation points to increase perf


weissi:
Much like #1385, `CircularBuffer` should implement some more `Collection`/`Sequence` customisation points to get better performance.
Generally, this will look something like

```
public func _customTHEFUNCTION(ARGUMENTS) {
        return try self._buffer.withVeryUnsafeBytes { ptr in
            let ptr = UnsafeRawBufferPointer(start: ptr.baseAddress!.advanced(by: self._range.lowerBound), count: self._range.count))
            return ptr.THEFUNCTION(ARGUMENTS)
        }
}
```

The `Collection` customisation points we should implement are

- [x] (possibly?) [`_customIndexOfEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L523) using `ptr.firstIndex(of:)`
- [x] (possibly?) [`_customLastIndexOfEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L535) using `ptr.lastIndex(of:)`
- [x] (possibly?) all three variants of [`_failEarlyRangeCheck`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Collection.swift#L646) by just checking the `self._range`
- [x] (possibly?) [`_customContainsEquatableElement`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L352) using `ptr.contains()`
- [x] [`_copyToContiguousArray`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L358)
- [x] [`_copyContents`](https://github.com/apple/swift/blob/3423ee107418e635182f768f3d6160a212eed846/stdlib/public/core/Sequence.swift#L362)

We also don't need to implement all of those in one go, this can be done in multiple PRs.

Some notes:
- we should back the gains up by benchmarks
- this requires basically no SwiftNIO knowledge but intermediate Swift knowledge about how protocols/Collection/Sequence

weissi:
[...]

PR #2059, opened 2022-03-07
Improve the performance of copying CircularBuffer

simonjbeaumont:
### Motivation:

#1827 suggests we implement some hooks on Collection and Sequence to get better performance for CircularBuffer to increase performance. #2058 added the benchmarks so we can see the improvement from this change.

### Modifications:

Implement `Sequence._copyContents` for `CircularBuffer` in terms of its underlying `ContiguousArray` to avoid going through the extra layer of indices.

### Result:

Copying `CircularBuffer` should be faster.

### DIFF
--- a/Sources/NIOCore/CircularBuffer.swift
+++ b/Sources/NIOCore/CircularBuffer.swift
@@ -247,7 +247,7 @@ extension CircularBuffer: Collection, MutableCollection {
             }
         }
 
-        return (self[endIndex..<self.endIndex].makeIterator(), self.count)
+        return (self[self.endIndex..<self.endIndex].makeIterator(), self.count)
     }
 }
 
--- a/Sources/NIOCore/CircularBuffer.swift
+++ b/Sources/NIOCore/CircularBuffer.swift
@@ -233,18 +233,23 @@ extension CircularBuffer: Collection, MutableCollection {
             return (self.makeIterator(), buffer.startIndex)
         }
 
-        let indexRanges: [Range<Int>]
-        if self.tailBackingIndex >= self.headBackingIndex {
-            indexRanges = [self.headBackingIndex..<self.tailBackingIndex]
-        } else {
-            indexRanges = [ self.headBackingIndex..<self._buffer.endIndex, 0..<self.tailBackingIndex ]
-        }
-
-        for indexRange in indexRanges {
-            for index in indexRange {
+        @inline(__always)
+        func copy(
+            range: Range<Int>,
+            from src: ContiguousArray<Element?>,
+            to ptr: inout UnsafeMutablePointer<Element>
+        ) {
+            for index in range {
                 ptr.initialize(to: self._buffer[index]!)
                 ptr += 1
             }
+        }
+
+        if self.tailBackingIndex >= self.headBackingIndex {
+            copy(range: self.headBackingIndex..<self.tailBackingIndex, from: self._buffer, to: &ptr)
+        } else {
+            copy(range: self.headBackingIndex..<self._buffer.endIndex, from: self._buffer, to: &ptr)
+            copy(range: 0..<self.tailBackingIndex, from: self._buffer, to: &ptr)
         }
 
         return (self[self.endIndex..<self.endIndex].makeIterator(), self.count)
--- a/Sources/NIOCore/CircularBuffer.swift
+++ b/Sources/NIOCore/CircularBuffer.swift
@@ -224,7 +224,7 @@ extension CircularBuffer: Collection, MutableCollection {
     }
 
     @inlinable
-    public __consuming func _copyContents(
+    public func _copyContents(
         initializing buffer: UnsafeMutableBufferPointer<Element>
     ) -> (Iterator,UnsafeMutableBufferPointer<Element>.Index) {
         precondition(buffer.count >= self.count)
--- a/Sources/NIOCore/CircularBuffer.swift
+++ b/Sources/NIOCore/CircularBuffer.swift
@@ -226,7 +226,7 @@ extension CircularBuffer: Collection, MutableCollection {
     @inlinable
     public func _copyContents(
         initializing buffer: UnsafeMutableBufferPointer<Element>
-    ) -> (Iterator,UnsafeMutableBufferPointer<Element>.Index) {
+    ) -> (Iterator, UnsafeMutableBufferPointer<Element>.Index) {
         precondition(buffer.count >= self.count)
 
         guard var ptr = buffer.baseAddress else {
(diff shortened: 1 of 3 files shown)


Lukasa:

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -222,6 +222,33 @@ extension CircularBuffer: Collection, MutableCollection {
             return end.backingIndex &- start.backingIndex
         }
     }
+
+    @inlinable
+    public __consuming func _copyContents(
```
Lukasa, line 227: Let's drop the `__consuming` annotation here, as it's non-stable.

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -222,6 +222,33 @@ extension CircularBuffer: Collection, MutableCollection {
...
+        initializing buffer: UnsafeMutableBufferPointer<Element>
+    ) -> (Iterator,UnsafeMutableBufferPointer<Element>.Index) {
```
Lukasa, line 229: Nit: 
```suggestion
    ) -> (Iterator, UnsafeMutableBufferPointer<Element>.Index) {
```

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -222,6 +222,33 @@ extension CircularBuffer: Collection, MutableCollection {
...
+        precondition(buffer.count >= self.count)
+
+        guard var ptr = buffer.baseAddress else {
+            return (self.makeIterator(), buffer.startIndex)
+        }
+
+        let indexRanges: [Range<Int>]
```
Lukasa, line 236: Hmm, we need to confirm that this doesn't allocate, but I forgot to get you to set limits. One sec.

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -222,6 +222,33 @@ extension CircularBuffer: Collection, MutableCollection {
...
+        if self.tailBackingIndex >= self.headBackingIndex {
+            indexRanges = [self.headBackingIndex..<self.tailBackingIndex]
+        } else {
+            indexRanges = [ self.headBackingIndex..<self._buffer.count, 0..<self.tailBackingIndex ]
+        }
+
+        for indexRange in indexRanges {
+            for index in indexRange {
+                ptr.initialize(to: self._buffer[index]!)
+                ptr += 1
```
Lukasa, line 246: This can be unchecked (`&+= 1`) as it cannot possibly overflow.

Sources/NIOCore/CircularBuffer.swift
```diff
@@ -222,6 +222,33 @@ extension CircularBuffer: Collection, MutableCollection {
...
+
+        guard var ptr = buffer.baseAddress else {
+            return (self.makeIterator(), buffer.startIndex)
+        }
+
+        let indexRanges: [Range<Int>]
+        if self.tailBackingIndex >= self.headBackingIndex {
+            indexRanges = [self.headBackingIndex..<self.tailBackingIndex]
+        } else {
+            indexRanges = [ self.headBackingIndex..<self._buffer.count, 0..<self.tailBackingIndex ]
```
Lukasa, line 240: We should prefer `self._buffer.endIndex`.

Sources/NIOCore/CircularBuffer.swift
simonjbeaumont: I don't think this infix operator can be applied to a LHS of pointer:

```
Binary operator '&+=' cannot be applied to operands of type 'UnsafeMutablePointer<Element>' and '_'
```

Sources/NIOCore/CircularBuffer.swift
simonjbeaumont: Done.

Sources/NIOCore/CircularBuffer.swift
simonjbeaumont: Done.

Sources/NIOCore/CircularBuffer.swift
simonjbeaumont: Done.

simonjbeaumont (commented):
@Lukasa Thanks for the review. Should I run the alloc tests locally or wait for the CI to do it's thang?

Sources/NIOCore/CircularBuffer.swift
dnadoba: TIL: [`+=` calls](https://github.com/apple/swift/blob/6c829e02a80269e2e2941ae907b901cb4265972c/stdlib/public/core/Pointer.swift#L381-L384) down [to `advance(by)`](https://github.com/apple/swift/blob/6c829e02a80269e2e2941ae907b901cb4265972c/stdlib/public/core/Pointer.swift#L288-L291) which uses `Builtin.gep_Word` to increment the pointer. No idea where we could look up the implementation of `Builtin.gep_Word` but [the assembly code shows that we already do unchecked arithmetic](https://godbolt.org/z/MoM8qjo6r).

Lukasa: Yeah I forgot this was a pointer: pointers don't need unchecked math.

Lukasa:
@simonjbeaumont Looks like Swift was not able to optimise the Array you allocated in the implementation. As we will only ever store one or two ranges, can I propose using some `var`s to remove the `Array`?

simonjbeaumont:

simonjbeaumont:
Hm... this seems to have made things worse... I mean, it's still better for this test than `main` but it's not as much better than the original state of this PR and also seems to have caused some significant regressions in other benchmarks that weren't there before we tried to reduce allocations.

Lukasa:
The regressions _were_ there in the other benchmarks. I suspect that many of these benchmarks are not running often enough so they're below the noise floor.

Lukasa:
I think there are two options here, but the best one is to probably make a PR that raises the runtime of these various benchmarks to around the hundreds-of-ms runtime. That is a larger runtime that makes it a bit easier for us to observe the performance changes. Is that something you have the bandwidth to do? If not I can try to get to it.

simonjbeaumont:

Lukasa:
Unrelated I suspect, that number has been a bit aggressively low for a while. Let's just re-run, it should pass.

Lukasa:

simonjbeaumont:

Lukasa (approved):
In general this looks good to me. We could get this to go a little faster by dropping down to raw pointers ourselves and using unsafeBitcasts and things, but I don't think it's really worth it.

PR #2062, opened 2022-03-08
Remove unnecessary extra swift file.

Lukasa:
The nightly builders have started automatically adding a struct with
@main to packages generated using swift package init. This is
interfering with our syscallwrapper tests, so we should delete that
file.


FranzBusch (approved):
LGTM!

PR #2067, opened 2022-03-21
[WIP] Channel sequence, and an example of async/await structure

Lukasa:
This PR contains a work-in-progress general purpose channel-to-`AsyncSequence` transformer, as well as an example app that is primarily intended to guide users in how to structure their Swift-on-Server apps with structured concurrency.

This is up for early feedback now. We're missing tests for the async sequence transformation, and we have to write code that can appropriately handle cancellation of the sequence consumer, but the general shape of this is about right. If people have opinions or thoughts I'd like to hear them. Start at the new example app for a lengthy description of intended structure.

Still to be done:

- [ ] Error propagation
- [ ] Handle users dropping the consumption end of the `AsyncSequence`: this should lead to automatic teardown.
- [ ] Provide helpers for scheduling deadlines and cancelling `Task`s
- [ ] Expand the example to allow users to cancel the main `Task` on `SIGUSR1` to demonstrate graceful shutdown.
- [ ] Expand the example to allow users to cancel just the listener on `SIGUSR2` to demonstrate how you might quiesce a server without using NIO's handlers
- [ ] Add a stream writer that can consume an `AsyncSequence`, rather than just relying on `write`, so that users have a fully back-to-back system
- [ ] Consider how we might handle user events in this pattern


FranzBusch (commented):
Thanks, @Lukasa for creating this. I think this is very valuable for the community to see. What I would be interested in is information around two things:
1. Where exactly thread hops are happening here and how custom executors could fix this.
2. How this would integrate with `ServiceLifecycle`

Sources/NIOAsyncAwaitServerExample/Server.swift
```diff
@@ -0,0 +1,175 @@
...
+                    taskGroup.addTask {
+                        await Self.handleChildChannel(childChannel)
+                    }
+                }
+            }
+        } onCancel: {
+            // This is a nice property of NIO: we can do this without actually
+            // awaiting it.
+            serverChannel.close(promise: nil)
+        }
```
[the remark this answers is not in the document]
ktoso: What do you mean by "async cancellation"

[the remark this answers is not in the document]
ktoso: Arguably the reason people want to await in `onCancel` is because cancel work in an actor; and the ACTUAL missing feature here is `send`, i.e. "notify that actor that <something> but no need to await" that's the semantics we need for such cancellation handlers IMHO, and we'll keep hitting that wall until we get "don't await" (i.e. `send`)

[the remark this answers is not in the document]
FranzBusch: Ah, that makes sense. Thanks for elaborating! What about errors that can happen during cancellation, should they just be dropped?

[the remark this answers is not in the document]
ktoso: You'd need some "back channel"; generally there's too much complexity in this and in reality... what would you even do when cancelling something failed? Cancelling is always racy and best effort to begin with, so it's not a reliable mechanism you can build upon for "did it definitely cancel or not" etc. 

This is similar to why we decided that reactive streams have no channel for cancellation -- all you can to is send completion/error _forward_ but if the cancel even had an effect or not, you're never 100% sure unless you exactly know what error it should cause. I.e. the cancellation after all propagates to children in our system, so the child task may return/complete and you don't really know as the outsider (NIO) if that just completed quickly, or BECAUSE of the cancellation that some child task checked for.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,302 @@
...
+
+        @inlinable public func next() async throws -> Element? {
+            switch await self._handler.loadEvent() {
+            case .channelRead(let message, let promise):
+                promise?.succeed(())
+                return message
+            case .eof:
+                return nil
+            }
+        }
```
FranzBusch, line 272: I know we are wrapping the calls to this to a `withTaskCancellationHandler(handler:operation:)` above but shouldn't we check `isCancelled` here as well. In the case users get it wrong?

Sources/NIOCore/Channel.swift
```diff
@@ -154,6 +161,28 @@ extension Channel {
...
+        let handler = NIOAsyncChannelAdapterHandler<InboundIn>(config: config)
+
+        if self.eventLoop.inEventLoop {
+            try self.pipeline.syncOperations.addHandler(handler)
+        } else {
+            try await self.pipeline.addHandler(handler)
+        }
+
+        return NIOInboundChannelStream(handler)
+    }
```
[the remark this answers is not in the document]
Lukasa: Yeah, this assumes that the end of your pipeline only has one chunk of data. I agree, we should document the limitation.

Sources/NIOAsyncAwaitServerExample/main.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+// This example program implements a basic TCP echo server, with the business logic in
+// async/await code. The reason to keep the logic so simple is to allow us to focus on the
+// structure of the program, including the way it manages tasks.
+
+#if canImport(_Concurrency) && compiler(>=5.5.2)
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+func main() async {
+    let server = Server(host: "localhost", port: 8765)
+    try! await server.run()
```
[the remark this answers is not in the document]
Lukasa: I definitely think we need to do it: I don't want to do it here though, probably in another example.

The specific reason I'm reluctant to do it here is that the _straightforward_ case is a server that owns everything and is purely reactive. That's easy to handle. For long-running background tasks we need a different abstraction, and that adds meaningfully more complexity to the situation. So I think I'd like to produce a different example that makes that pattern clearer.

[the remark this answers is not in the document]
FranzBusch: Great. Happy to help here as well!

Sources/NIOAsyncAwaitServerExample/Server.swift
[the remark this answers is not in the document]
simonjbeaumont: This is something that I've ended up wrestling with. Often getting stuck with not being able to call actor (async) functions from within `deinit` or the Task `cancellationHandler`. I think @ktoso comments are sensible, there's only so much you can do, short of spinning off an unstructured `Task` within one of these contexts, which I think undermines the idea of structured concurrency that we're aiming for here?

Sources/NIOAsyncAwaitServerExample/main.swift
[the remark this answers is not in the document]
simonjbeaumont: I'm very interested in the stateful server case too, which is much of what's been causing me issues. It's nice to see this be built up in layers though. 👍

Sources/NIOCore/Channel.swift
```diff
@@ -154,6 +161,28 @@ extension Channel {
...
 
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+extension Channel {
+    /// Default implementation: construct a stream with the given API surface, not modifying the configuration.
+    ///
+    /// Almost all Channels will use the default implementation. A few Channels may require specialized behaviour:
+    /// those Channels can override and modify this implementation as they need.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @inlinable
+    public func makeAsyncStream<InboundIn>(of type: InboundIn.Type, config: NIOInboundChannelStreamConfig) async throws -> NIOInboundChannelStream<InboundIn> {
```
simonjbeaumont, line 172: What removes these handlers once the stream is no longer in use? 

If a user calls...

```swift
for _ in 1...1000 {
    _ = try await channel.makeAsyncStream(...)
}
```

...won't this suffer from the lapsed listener problem? I.e. won't we have 1000 handlers that no one cares about any more?

As we've discussed before, one ugly solution to this was to wrap the `AsyncSequence` returned in a reference type and try and use its lifecycle to deregister the listener, which kind of worked but wasn't great.

Have I missed something here, maybe lapsed listeners are reaped somehow in this proposal that I overlooked?

Lukasa: Nope, you haven't missed anything here: for now we'd need to use `deinit` on the `AsyncSequence`. It's in my list of things to tackle before we merge this.

Sources/NIOAsyncAwaitServerExample/Server.swift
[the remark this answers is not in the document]
ktoso: Yeah, what Cory said -- we're missing language features IMHO, there should be a `send actor.cancelPlease()` that _does not_ await, and could have a slightly lighter task, but that also would be spawning a new task to kick off the work.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,302 @@
...
+        case .success:
+            context.read()
+        case .failure:
+            context.close(promise: nil)
+        }
+    }
+
+    @inlinable
+    func _receiveNewEvent(_ event: InboundMessage) {
+        // Sadly, we need to do some lock shenanigans here.
```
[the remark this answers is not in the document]
Lukasa: Yeah, you're right, this is from an older version of the code where I was doing some stuff I simply could not guarantee was safe. I'll clean these up.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,302 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+
+import NIOConcurrencyHelpers
+import Darwin
```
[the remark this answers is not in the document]
Lukasa: Nope.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,296 @@
...
+            result.1?.succeed(())
+            return result.0
+        } else {
+            return await self._loadEventSlowPath()
+        }
+    }
+
+    @inlinable
+    func _loadEventSlowPath() async -> InboundMessage {
+        return await withCheckedContinuation { (continuation: CheckedContinuation<InboundMessage, Never>) in
```
artemredkin, line 179: how would you recommend handling errors? Why not use `CheckedContinuation<InboundMessage, Error>` and `resume(throwing:)` in the `errorCaught`?

artemredkin, line 179: how would you recommend handling errors? Why not use `CheckedContinuation<InboundMessage, Error>` and `resume(throwing:)` in the `errorCaught`?

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,296 @@
...
+                    return nil
+                }
+            case .bufferingWithPendingRead(var buffer, let promise):
+                // No matter what happens here, we need to complete this promise.
+                if buffer.count > 0 {
+                    let result = buffer.removeFirst()
+                    self._state = .bufferingWithoutPendingRead(buffer)
+                    return (result, promise)
+                } else {
+                    return nil
```
artemredkin, line 162: it looks like we are not completing this promise if buffer is empty

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    @inlinable
+    func _delayedRead(result: Result<Void, Error>, context: ChannelHandlerContext) {
+        self._lock.withLockVoid {
+            switch self._state {
+            case .bufferingWithPendingRead(let buffer, _):
+                self._state = .bufferingWithoutPendingRead(buffer)
+            case .bufferingWithoutPendingRead, .waitingForBuffer:
+                preconditionFailure()
```
artemredkin, line 109: I think there is a bug here, if buffer size is small and async reader is fast, we might end up in a situation where `_delayedRead` promise will be completed at an unpredictable time, when state can be `. bufferingWithoutPendingRead` (I'm hitting this constantly with a reader that just accumulates bytes count)

Sources/NIOCore/AsyncChannelAdapter.swift
Lukasa: Yup, we should do exactly that. This is very much WIP.

Sources/NIOCore/AsyncChannelAdapter.swift
Lukasa: Yes agreed.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,514 @@
...
+
+    public struct AsyncIterator: AsyncIteratorProtocol {
+        @usableFromInline var _iterator: Producer.AsyncIterator
+
+        @inlinable
+        init(_ iterator: Producer.AsyncIterator) {
+            self._iterator = iterator
+        }
+
+        @inlinable public func next() async throws -> Element? {
```
FranzBusch, line 490: Just a drive by we should make this `mutating` before we forget it. I missed to make the `NIOAsyncSequenceProducer.next()` mutating which was mistake but we can correct it here

Sources/NIOCore/AsyncChannel.swift
```diff
@@ -0,0 +1,102 @@
...
+/// logic should use ``NIOAsyncChannel`` to consume and produce data to the network.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public final class NIOAsyncChannel<InboundIn: Sendable, OutboundOut: Sendable>: Sendable {
+    /// The underlying channel being wrapped by this ``NIOAsyncChannel``.
+    public let channel: Channel
+
+    public let inboundStream: NIOInboundChannelStream<InboundIn>
+
+    @usableFromInline
+    let outboundWriter: NIOAsyncChannelWriterHandler<OutboundOut>.Writer
```
FranzBusch, line 45: I have experimented around with this a bit and I think we should make this a `public let` and re-abstract the concrete type. This allows higher level APIs to re-wrap just the writer instead of the whole `NIOAsyncChannel`.

fumoboy007:
Thanks for the hard work. Really excited for this addition that will help bridge SwiftNIO and Swift Concurrency with back pressure!

adam-fowler:
Been playing with this over the weekend and it's really nice building servers using swift concurrency. Built a simple HTTP server with not too much trouble. The main issues I came across were currently there is no access to channel events and propagating channel errors to my NIOAsyncChannel handler.

FranzBusch:
> Been playing with this over the weekend and it's really nice building servers using swift concurrency. Built a simple HTTP server with not too much trouble. The main issues I came across were currently there is no access to channel events and propagating channel errors to my NIOAsyncChannel handler.

Thanks for testing it! We are still in the midst of figuring things out. I just talked @Lukasa about the events this morning. We think that in general events should be a concept of the channel handler pipeline; however, we probably want to treat the `ChannelShouldQuiesceEvent` in a special way because that makes sense to expose to users of the `NIOAsyncChannel`. 
The other events can be handled by adding a custom handler which exposes an `AsyncSequence` where you can subscribe to the event.

I know that you have looked at the `IdleStateHandler`. One way we could make this work is by having that handler expose the events as a separate `AsyncSequence`. We still haven't fully made up our mind here and are happy to hear more opinions.

adam-fowler:
> I know that you have looked at the IdleStateHandler. One way we could make this work is by having that handler expose the events as a separate AsyncSequence. We still haven't fully made up our mind here and are happy to hear more opinions.

If you are going to do that, why don't you provide a separate `AsyncSequence` with all the events being passed down the pipeline. This could be used to access quiesce events as well.

adam-fowler:
I can see how there would be some ordering issues. 

>  It's frankly not clear to me that user events should leave the channel at all: they seem to generally be channel pipeline concerns

Well that is the question where should code be placed, in the code handling the NIOAsyncChannel or in a ChannelHandler. I guess I could add an additional channel handler that just dealt with user events.

Sources/NIOCore/AsyncChannelAdapter.swift
```diff
@@ -0,0 +1,514 @@
...
+    @inlinable
+    func channelInactive(context: ChannelHandlerContext) {
+        self._completeStream(context: context)
+        context.fireChannelInactive()
+    }
+
+    @inlinable
+    func errorCaught(context: ChannelHandlerContext, error: Error) {
+        self._completeStream(with: error, context: context)
+        context.fireErrorCaught(error)
```
adam-fowler, line 163: Should we have an option here to not propagate the error further down the pipeline so the code handling the `NIOAsyncChannel.inboundStream` can deal with the error instead. Currently it is propagated to the next channel handler `NIOAsyncChannelWriterHandler` which closes the write side of the channel.

Lukasa: The writer handler should probably just not choke on that error.

adam-fowler: That's an option as well

FranzBusch:
I think we can close this PR now since we have landed all of this functionality in NIO and ServiceLifecycle

Lukasa:
We have NIOAsyncChannel.

PR #2069, opened 2022-03-23
Add `Hashable` conformance to `HTTPResponseStatus`

FranzBusch:
# Motivation
`HTTPResponseStatus` is currently not conforming to `Hashable`.

# Modification
Add `Hashable` conformance to `HTTPResponseStatus`

# Result
`HTTPResponseStatus` is `Hashable`

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -1281,18 +1281,9 @@ public enum HTTPResponseStatus {
     }
 }
 
-extension HTTPResponseStatus: Equatable {
-    public static func ==(lhs: HTTPResponseStatus, rhs: HTTPResponseStatus) -> Bool {
-        switch (lhs, rhs) {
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
If we can rely on the synthesised conformance, can we remove the custom `Equatable` implementation? I'm always nervous having two very different implementations.

FranzBusch:
@Lukasa Agreed. I had it removed first but reverted it. Here it is again.

Lukasa (approved):

Lukasa:
5.3 is the only real failure here, and it's transient: the limit is a bit low.

Lukasa:

FranzBusch:
5.7 and nightly hit an allocation regression which is unrelated to this PR. The breaking change tool failed since it cannot see the auto synthesized `==` implementation.

@Lukasa If this looks good to you, feel free to merge over the failures.

Lukasa:
Yup, will do.

PR #2063, opened 2022-03-09
Increase runtime of performance tests to O(10 ms) to increase SNR

simonjbeaumont:
### Motivation:

The current running time of some of the performance benchmarks is O(1 ms) which can lead to low signal in the results. It would be better if these tests ran for O(10 ms).

### Modifications:

* Increase the iteration counts in all the tests within `NIOPerformanceTester` to aim for O(10 ms) measurements with the current code.
* Exception: `execute_100k_tasks` which is still too quick but the CI cannot handle 1M tasks. 

### Result:

* All benchmarks bodies should now take O(100 ms).  

### Notes:

The scaling factors were determined by looking at the `current` column from a recent report on #2059:

```sh 
$ python3 scaling_factors.py perf-output
Loading file: perf-output...
Loaded 62 results.
Scaling tests so that runtime is O(10 ms)

name                                                               current_s    O(current_s)    O(desired_s)  scale_by
write_http_headers                                                 0.004221364  1E-3            1E-2          1E1
http_headers_canonical_form                                        0.085858433  1E-2            1E-2          1E0
http_headers_canonical_form_trimming_whitespace                    0.162971487  1E-1            1E-2          1E-1
http_headers_canonical_form_trimming_whitespace_from_short_string  0.148830606  1E-1            1E-2          1E-1
http_headers_canonical_form_trimming_whitespace_from_long_string   0.229662477  1E-1            1E-2          1E-1
bytebuffer_write_12MB_short_string_literals                        0.151619987  1E-1            1E-2          1E-1
bytebuffer_write_12MB_short_calculated_strings                     0.297283243  1E-1            1E-2          1E-1
bytebuffer_write_12MB_medium_string_literals                       0.058404425  1E-2            1E-2          1E0
bytebuffer_write_12MB_medium_calculated_strings                    0.15610171   1E-1            1E-2          1E-1
bytebuffer_write_12MB_large_calculated_strings                     0.142773805  1E-1            1E-2          1E-1
bytebuffer_lots_of_rw                                              0.433748776  1E-1            1E-2          1E-1
bytebuffer_write_http_response_ascii_only_as_string                0.039572042  1E-2            1E-2          1E0
bytebuffer_write_http_response_ascii_only_as_staticstring          0.030672688  1E-2            1E-2          1E0
bytebuffer_write_http_response_some_nonascii_as_string             0.039276667  1E-2            1E-2          1E0
[...]

PR #2072, opened 2022-03-25
Use unbuffered IO for stdout in NIOPerformanceTester

simonjbeaumont:
### Motivation:

While updating `NIOPerformanceTester` in #2063 a test was crashing but it was hard to see which test was running at the time of the crash due to output buffering which meant we had to tediously binary search for the test that was causing the CI crash.

### Modifications:

Switch to using unbuffered IO for standard output within `NIOPerformanceTester`.

### Result:

Output from `NIOPerformanceTester` will be immediately flushed so that, in the event of a crash in CI, it will be clear which test was running.


simonjbeaumont:
//cc @tomerd @Lukasa

simonjbeaumont:

simonjbeaumont:
Note, I didn't run the perf test because I think this change should improve performance, just to check it doesn't break the harness itself.

simonjbeaumont:
Let me test that this _actually_ helps by causing the same crash as we had in #2063. Standby...

simonjbeaumont:

simonjbeaumont:
Here's the CI run with the same crash as we had in #2063: https://ci.swiftserver.group/job/swift-nio2-performance-prb/131/

From the logs, we can now see exactly which test it was that failed, the `schedule_1m_tasks`:

```
11:43:30 measuring: schedule_1m_tasks: Build step 'Execute shell' marked build as failure
```

I'll now revert that commit so it can be squashed away in the merge.

tomerd:
seems reasonable to me 👍

simonjbeaumont:

Sources/NIOPerformanceTester/main.swift
```diff
@@ -19,6 +19,9 @@ import NIOFoundationCompat
 import Dispatch
 import NIOWebSocket
 
+// Use unbuffered stdout to help detect exactly which test was running in the event of a crash.
+setbuf(stdout, nil)
+
```
adam-fowler, line 24: Not sure this is relevant in this case, but macOS outputs test results to stderr

simonjbeaumont: Hmm. This isn’t an XCTest thing; this is just a Swift binary with a main.swift using `print`. Are you sure that goes to stderr on macOS?

adam-fowler: Ah sorry. Should have realised that.

simonjbeaumont: That’s good to know for future, thanks!

Lukasa (approved):

PR #2075, opened 2022-04-11
Remove invalid math in ByteBuffer slice slow path

Lukasa:
Motivation:

ByteBuffer uses a synthetic UInt24 as the lower bound of the slice of
memory that it can access. This necessitates a slow path when we create
slices of ByteBuffers that would force the lower bound to be larger than
a UInt24.

This slow path has a bunch of math within it, and it turns out that some
of it is invalid. This means in rare cases slicing a ByteBuffer can
cause a crash. That's not great: the code should only crash if we broke
an internal invariant, not because we did some invalid math.

In particular, the relevant check is calculating the upper bound of the
storage to be copied. This was:

```
min(storageRebaseAmount + _toCapacity(self._slice.count), self._slice.upperBound, storageRebaseAmount + capacity)
```

In this case, `storageRebaseAmount` is either `self._slice.lowerBound` or
`self._slice.lowerBound + self._readerIndex`: that is,
`storageRebaseAmount >= self._slice.lowerBound`. Given that
`self._slice.count == self._slice.upperBound - self._slice.lowerBound`,
we know that `storageRebaseAmount + self._slice.count` is always
strictly greater than or equal to `self._slice.upperBound`. Calculating
this is unnecessary. Unfortunately, it can also overflow: if
`self._slice.upperBound + self._readerIndex` is greater than UInt32 then
this will crash.

Note that the other math, `storageRebaseAmount + capacity`, has been
kept. `capacity` is always computed by us and this should only ever
overflow if we are increasing the size of the buffer past UInt32.max,
which we cannot handle anyway.

Modifications:

- Restructured and heavily commented _copyStorageAndRebase.
- Removed the redundant crashing check.

Result:

Less trapping!

### DIFF
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -361,10 +361,9 @@ public struct ByteBuffer {
         let newSlice = storageRebaseAmount ..< storageUpperBound
         self._storage = self._storage.reallocSlice(newSlice, capacity: capacity)
 
-        // Step 5: Fixup the indices. These can never underflow: `indexRebaseAmount` can only ever be 0 or `self._readerIndex`, so we either do nothing
-        // or we move the indices backwards by a valid range, as `self._writerIndex` is always >= `self._readerIndex`.
-        self._moveReaderIndex(to: self._readerIndex &- indexRebaseAmount)
-        self._moveWriterIndex(to: self._writerIndex &- indexRebaseAmount)
+        // Step 5: Fixup the indices. These should never trap, but we're going to leave them checked because this method is fiddly.
+        self._moveReaderIndex(to: self._readerIndex - indexRebaseAmount)
+        self._moveWriterIndex(to: self._writerIndex - indexRebaseAmount)
 
         // Step 6: As we've reallocated the buffer, we can now use the entire new buffer as our slice.
         self._slice = self._storage.fullSlice
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -318,12 +318,55 @@ public struct ByteBuffer {
     @inlinable
     @inline(never)
     mutating func _copyStorageAndRebase(capacity: _Capacity, resetIndices: Bool = false) {
+        // This math has to be very careful, because we already know that in some call paths _readerIndex exceeds 1 << 24, and lots of this math
+        // is in UInt32 space. It's not hard for us to trip some of these conditions. As a result, I've heavily commented this
+        // fairly heavily to explain the math.
+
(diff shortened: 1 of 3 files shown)


Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -318,12 +318,55 @@ public struct ByteBuffer {
...
+
+        // Step 4: Allocate the new buffer and copy the slice of bytes we want to keep.
+        let newSlice = storageRebaseAmount ..< storageUpperBound
         self._storage = self._storage.reallocSlice(newSlice, capacity: capacity)
-        self._moveReaderIndex(to: self._readerIndex - indexRebaseAmount)
-        self._moveWriterIndex(to: self._writerIndex - indexRebaseAmount)
+
+        // Step 5: Fixup the indices. These can never underflow: `indexRebaseAmount` can only ever be 0 or `self._readerIndex`, so we either do nothing
+        // or we move the indices backwards by a valid range, as `self._writerIndex` is always >= `self._readerIndex`.
+        self._moveReaderIndex(to: self._readerIndex &- indexRebaseAmount)
```
weissi, line 366: Thanks for rewriting, commenting & fixing this! The only question I have is why we use unchecked math here. This function is the slow path and never inlined. So I don't think there's any harm caused by always doing checked math (even if it's clearly safe do use the unchecked math in _this version_ of the code.)

The reason I'm mildly worried is because of potential future changes, they might cargo cult the unchecked math even in a place where it doesn't matter and so a new vulnerability or smth may slip in.

Feel free to ignore :)

Lukasa: Yeah, I'm definitely on the fence about that. Let's see what the others think.

Tests/NIOPosixTests/ByteBufferTest.swift
```diff
@@ -1746,6 +1746,30 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(0xcc, slice.getInteger(at: 0, as: UInt8.self))
         XCTAssertEqual(0xdd, slice.getInteger(at: slice.writerIndex - 1, as: UInt8.self))
     }
+
+    func testSliceOfMassiveBufferWithAdvancedReaderIndexIsOk() throws {
+        // We only want to run this test on 64-bit systems: 32-bit systems can't allocate buffers
+        // large enough to run this test safely.
+        guard MemoryLayout<Int>.size >= 8 else {
+            return
```
FranzBusch, line 1754: Nit: Can we use `XCTSkip` here instead of just returning
```suggestion
            throw XCTSkip("This test is only supported on 64-bit systems")
```

FranzBusch (approved):
LGTM! Thanks for the changes and the thorough comments.

PR #2076, opened 2022-04-12
Fix doc comment formatting in `SingleStepByteToMessageDecoder.swift`

MaxDesiatov:
### Motivation:

Rendering of this doc comment is currently broken, here's a screenshot from DocC:

Screenshot 2022-04-12 at 12 53 17

`decode` in the comment is currently missing a starting backquote, which is fixed in this PR.


Lukasa:

Issue #2073, opened 2022-03-27
Can't build SwiftNIOPosix in Release (using Pods as a packet manager)


madmag77:
### Expected behavior
I expect I can build in release mode my app that uses SwiftNIO pod.
This is my pod (https://github.com/madmag77/STUNClient/blob/master/StunClient.podspec) that uses SwiftNIO as a dependency.

### Actual behavior
This is a sample app (https://github.com/madmag77/STUNClient/tree/master/iOSExampleWithPods) that can't be built in release mode when uses SwiftNIO pod version 2.38.

Output:
```
Ld /Users/artemgoncharov/Downloads/STUNClient/iOSExampleWithPods/DerivedData/STUNClient/Build/Intermediates.noindex/Pods.build/Release-iphoneos/SwiftNIOPosix.build/Objects-normal/arm64/Binary/NIOPosix normal arm64 (in target 'SwiftNIOPosix' from project 'Pods')
    cd /Users/artemgoncharov/Downloads/STUNClient/iOSExampleWithPods/Pods
[...]

PR #1719, opened 2021-01-12
Remove CNIOAtomics and CNIOSHA1 as dependency

kimdv:
_[One line description of your change]_

### Motivation:

Remove the two mentioned dependencies and resolves #1705 

### Modifications:

Removed two dendencies

### Result:

Two less dependencies in NIO

PR #2082, opened 2022-04-19
Script to list transitive dependencies

glbrntt:
Motivation:

Cocoapods appears to have a different idea of the dependency graph to
SPM which has led to a handful of build failures (e.g.
https://github.com/apple/swift-nio/issues/2073).

This appears to have originated when we dropped the explicit dependency
on `CNIOAtomics` from `NIO` (https://github.com/apple/swift-nio/pull/1719).

We can work around this by listing all transitive dependencies as
requirements in the generated podspecs.

Modifications:

- Add a script to list transitive dependencies for a given module.
- Fix a bug in build_podspecs.sh which did a `pod repo update` even if
  the pods were not being uploaded.
- Update build_podspecs.sh to use transitive dependencies.

Result:

Podspecs should include all transitive dependencies.


glbrntt:
FWIW I tested the script on HTTP/2 and GRPC since they include package dependencies. I also looked over the generated podspecs.

scripts/build_podspecs.sh
```diff
@@ -75,7 +75,7 @@ for target in "${targets[@]}"; do
 
   while read -r raw_dependency; do
     dependencies+=( "${newline}  s.dependency '$raw_dependency', s.version.to_s" )
-  done < <("${here}/list_topsorted_dependencies.sh" -d "${target#Swift}" | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
+  done < <("${here}/list_transitive_dependencies.py" "${target#Swift}" | grep -v "NIOPriorityQueue" | sed 's/^NIO/SwiftNIO/')
```
weissi, line 78: should we just drop the `.sh`/`.py` suffix? It doesn't really help and causes issues in situations like this...

Lukasa: Is it causing an issue? We're calling a different script entirely, not the same one rewritten in a different language.

scripts/list_transitive_dependencies.py
```diff
@@ -0,0 +1,122 @@
...
+                        dependency_name, self._packages[package_name], dependencies
+                    )
+
+    def _ensure_package_is_cached(self, package, package_name):
+        if package_name in self._packages:
+            return
+
+        # Find the package dependency with the right name.
+        for package_dependency in package["dependencies"]:
+            if "sourceControl" not in package_dependency:
```
Lukasa, line 92: Should we throw an exception in this case rather than ignore it?

Lukasa (approved):
Basically LGTM except for a couple of notes

scripts/build_podspecs.sh
glbrntt: Generally I also have a slightly preference to not having the prefix for scripts _but_ the extension is used by the soundness script when validating the license header so it does serve a small purpose here.

scripts/list_transitive_dependencies.py
glbrntt: Yeah, we can probably just remove this check entirely and hard fail if there's no "sourceControl" key.

scripts/build_podspecs.sh
weissi: Oh shoot, that's probably also why I named it that in the first place. Not sure if that's a good reason but at least there is a reason :). So yes, consider my point here moot.

PR #2083, opened 2022-04-20
Define AsyncEmbeddedEventLoop

Lukasa:
Motivation

The rise of Swift concurrency has meant that a number of our APIs need
to be recontextualised as async/await capable. While generally this is a
straightforward task, any time those APIs were tested using
EmbeddedChannel we have a testing issue. Swift Concurrency requires the
use of its own cooperative thread pool, which is completely incapable of
safely interoperating with EmbeddedChannel and EmbeddedEventLoop. This
is becuase those two types "embed" into the current thread and are not
thread-safe, but our concurrency-focused APIs want to enable users to
use them from any Task.

To that end we need to develop new types that serve the needs of
EmbeddedChannel and EmbeddedEventLoop (control over I/O and task
scheduling) while remaining fully thread-safe. This is the first of a
series of patches that adds this functionality, starting with the
AsyncEmbeddedEventLoop.

Modifications

- Define AsyncEmbeddedEventLoop

Result

A required building block for AsyncEmbeddedChannel exists.

cc @0xTim @gwynne


Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,358 @@
...
+        await self.advanceTime(to: self.now + increment)
+    }
+
+    /// Unwrap a future result from this event loop.
+    ///
+    /// This replaces `EventLoopFuture.get()` for use with `AsyncEmbeddedEventLoop`. This is necessary because attaching
+    /// a callback to an `EventLoopFuture` (which is what `EventLoopFuture.get` does) requires scheduling a work item onto
+    /// the event loop, and running that work item requires spinning the loop. This is a non-trivial dance to get right due
+    /// to timing issues, so we need to
+    public func awaitFuture<ResultType>(_ future: EventLoopFuture<ResultType>) async throws -> ResultType {
```
FranzBusch: I understand why we need all of this but it is slightly more than just a replacement for `.get()` right? It also spins the run loop which means it is a combination of the previous `eventLoop.run() + future.get()`.

What about slightly renaming this method to indicate that it runs the event loop. `runAndAwaitFuture()`?

Sources/NIOEmbedded/ActorEmbeddedChannel.swift
```diff
@@ -0,0 +1,26 @@
...
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+/// `ActorEmbeddedChannel` provides a thread-safe, Swift-concurrency-aware implementation of
+/// NIO's `Channel` protocol that is focused on deterministic testing.
+///
+/// Importantly, this `Channel` allows users to inject I/O and to control the progress of the
+/// run loop. This ensures that operations only proceed in an appropriate order.
+
+
+
+#endif
```
FranzBusch, line 26: @Lukasa This is basically empty. Should this be part of this PR?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+/// necessary to ensure that the event loop is not making progress.
+///
+/// Time is controllable on an `AsyncEmbeddedEventLoop`. It begins at `NIODeadline.uptimeNanoseconds(0)`
+/// and may be advanced by a fixed amount by using `advanceTime(by:)`, or advanced to a point in
+/// time with `advanceTime(to:)`.
+///
+/// If users wish to perform multiple tasks at once on an `AsyncEmbeddedEventLoop`, it is recommended that they
+/// use `establishContext` to perform the operations.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public final class AsyncEmbeddedEventLoop: EventLoop, @unchecked Sendable {
```
glbrntt, line 43: Generally speaking I think we should treat `@unchcked Sendable` as 'unsafe': we should comment why it's okay in this instance.

glbrntt: Should this be `NIOAsyncEmbeddedEventLoop`?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+    private let scheduledTaskCounter = NIOAtomic<UInt64>.makeAtomic(value: 0)
+    private var scheduledTasks = PriorityQueue<EmbeddedScheduledTask>()
+
+    /// Keep track of where promises are allocated to ensure we can identify their source if they leak.
+    private let _promiseCreationStore = PromiseCreationStore()
+
+    // The number of the next task to be created. We track the order so that when we execute tasks
+    // scheduled at the same time, we may do so in the order in which they were submitted for
+    // execution.
+    private var taskNumber = UInt64(0)
```
glbrntt, line 73: Should this only be accessed from `queue`?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+    public func advanceTime(by increment: TimeAmount) async {
+        await self.advanceTime(to: self.now + increment)
+    }
+
+    /// Unwrap a future result from this event loop.
+    ///
+    /// This replaces `EventLoopFuture.get()` for use with `AsyncEmbeddedEventLoop`. This is necessary because attaching
+    /// a callback to an `EventLoopFuture` (which is what `EventLoopFuture.get` does) requires scheduling a work item onto
+    /// the event loop, and running that work item requires spinning the loop. This is a non-trivial dance to get right due
+    /// to timing issues, so we need to
```
glbrntt, line 178: The suspense is killing me!

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+                let newTime = max(deadline, self.now)
+
+                while let nextTask = self.scheduledTasks.peek() {
+                    guard nextTask.readyTime <= newTime else {
+                        break
+                    }
+
+                    // Now we want to grab all tasks that are ready to execute at the same
+                    // time as the first.
+                    var tasks = Array<EmbeddedScheduledTask>()
```
glbrntt, line 225: Minor: should we use a circular buffer and move it outside of the while-loop to avoid allocating an array for each distinct `readyTime`?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+        }
+    }
+
+    internal func drainScheduledTasksByRunningAllCurrentlyScheduledTasks() {
+        var currentlyScheduledTasks = self.scheduledTasks
+        while let nextTask = currentlyScheduledTasks.pop() {
+            self._now.store(nextTask.readyTime.uptimeNanoseconds)
+            nextTask.task()
+        }
+        // Just fail all the remaining scheduled tasks. Despite having run all the tasks that were
```
glbrntt, line 268: IIRC MTELG allows some number of tasks to be executed if they were scheduled as a result of another task running. Are we okay diverging from that behaviour?

glbrntt (commented):
I haven't gone over the tests yet, but dropping my comments so far.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+        await self.advanceTime(to: self.now + increment)
+    }
+
+    /// Unwrap a future result from this event loop.
+    ///
+    /// This replaces `EventLoopFuture.get()` for use with `AsyncEmbeddedEventLoop`. This is necessary because attaching
+    /// a callback to an `EventLoopFuture` (which is what `EventLoopFuture.get` does) requires scheduling a work item onto
+    /// the event loop, and running that work item requires spinning the loop. This is a non-trivial dance to get right due
+    /// to timing issues, so we need to
+    public func awaitFuture<ResultType: Sendable>(_ future: EventLoopFuture<ResultType>) async throws -> ResultType {
```
dnadoba, line 179: I don't quite like the name because on the call site you will write `await` twice e.g.
```swift
let result = try await el.awaitFuture(someFuture)
```
I'm think about moving this into an extension on `EventLoopFuture` like this:
```swift
extension EventLoopFuture {
    func get(on eventLoop: AsyncEmbeddedEventLoop) async throws -> Value {
        ...
    }
}
let result = try await someFuture.get(on: el)
```
This way it would also show up in autocomplete when someone tries to call the "wrong" `.get()`.

fabianfett: > I don't quite like the name because on the call site you will write await twice

I agree.

> This way it would also show up in autocomplete when someone tries to call the "wrong" .get().

However there is no way for the user to know which is the right and which is the wrong `get()`. Why would I sometimes supply the EventLoop, if it is optional.

Maybe:

```swift
let result = try await el.runUntil(someFuture)
```

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+        return try await withThrowingTaskGroup(of: ResultType.self, returning: ResultType.self) { group in
+            group.addTask {
+                try await future.get()
+            }
+
+            group.addTask {
+                while true {
+                    await self.run()
+                    try Task.checkCancellation()
+                    try await Task.sleep(nanoseconds: 100_000)
```
fabianfett, line 191: Out of interest: Why do we use `sleep` here and not `yield`?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+                }
+            }
+
+            do {
+                // This force-unwrap is safe: there are only two tasks, one will either return or hang, and the
+                // one that returns cannot be nil, only error. We then cancel the other Task, but we'll never
+                // ask for its result.
+                let result = try await group.next()!
```
fabianfett, line 199: I think we need some kind of timeout here. What happens if the future never completes, because of a programming error? Should we error after five seconds?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,357 @@
...
+
+    /// - see: `EventLoop.shutdownGracefully`
+    public func shutdownGracefully(queue: DispatchQueue, _ callback: @escaping (Error?) -> Void) {
+        self.queue.async {
+            self._shutdownGracefully()
+            queue.async {
+                callback(nil)
+            }
+        }
+    }
```
fabianfett, line 289: Take it or leave it.
```suggestion
    public func shutdownGracefully(queue callbackQueue: DispatchQueue, _ callback: @escaping (Error?) -> Void) {
        self.queue.async {
            self._shutdownGracefully()
            callbackQueue.async {
                callback(nil)
            }
        }
    }
```

Sources/NIOEmbedded/ActorEmbeddedChannel.swift
Lukasa: Nope!

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,358 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2022 Apple Inc. and the SwiftNIO project authors
```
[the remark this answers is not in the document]
Lukasa: Yes: this is derived from `EmbeddedChannel` via copy-paste.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,358 @@
...
+/// Unlike more complex `EventLoop`s, such as `SelectableEventLoop`, the `AsyncEmbeddedEventLoop`
+/// has no proper eventing mechanism. Instead, reads and writes are fully controlled by the
+/// entity that instantiates the `AsyncEmbeddedEventLoop`. This property makes `AsyncEmbeddedEventLoop`
+/// of limited use for many application purposes, but highly valuable for testing and other
+/// kinds of mocking. Unlike `EmbeddedEventLoop`, `AsyncEmbeddedEventLoop` is fully thread-safe and
+/// safe to use from within a Swift concurrency context.
+///
+/// Unlike `EmbeddedEventLoop`, `AsyncEmbeddedEventLoop` does require that user tests appropriately
+/// enforce thread safety. Used carefully it is possible to safely operate the event loop without
+/// explicit synchronization, but it is recommended to use `establishContext` in any case where it's
```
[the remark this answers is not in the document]
Lukasa: Yup!

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,358 @@
...
+    /// A dispatch specific that we use to determine whether we are on the queue for this
+    /// "event loop".
+    private static let inQueueKey = DispatchSpecificKey<ObjectIdentifier>()
+
+    // Our scheduledTaskCounter needs to be an atomic because we're going to access it from
+    // arbitrary threads. This is required by the EventLoop protocol and cannot be avoided.
+    // Specifically, Scheduled<T> creation requires us to be able to define the cancellation
+    // operation, so the task ID has to be created early.
+    private let scheduledTaskCounter = NIOAtomic<UInt64>.makeAtomic(value: 0)
+    private var scheduledTasks = PriorityQueue<EmbeddedScheduledTask>()
```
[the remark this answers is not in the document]
Lukasa: Yes, I think so. PRs welcome. 😁

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Hmm, I don't love `runUntil` either. `unwrap`?

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Great question. I'm not confident `yield` is sufficient, but we could.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: In general I think I agree, _except_ that we can't really do that and appropriately clean up our `Task`s. The problem is that `EventLoopFuture.get` isn't cancellable. This is a separate issue: #2087

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: For now, I think yes. This is consistent with `EmbeddedEventLoop`.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Leave it: this is part of the `EventLoop` protocol.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Yes, updated.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: I've added a test for this that is skipped until we deal with #2087.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
fabianfett: I don't think this would be a problem though, we just name the parameter differently. But I'm fine with leaving it.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,405 @@
...
+        await self.advanceTime(to: self.now + increment)
+    }
+
+    /// Unwrap a future result from this event loop.
+    ///
+    /// This replaces `EventLoopFuture.get()` for use with `AsyncEmbeddedEventLoop`. This is necessary because attaching
+    /// a callback to an `EventLoopFuture` (which is what `EventLoopFuture.get` does) requires scheduling a work item onto
+    /// the event loop, and running that work item requires spinning the loop. This is a non-trivial dance to get right due
+    /// to timing issues, so this function provides a helper to ensure that this will work.
+    public func awaitFuture<ResultType: Sendable>(_ future: EventLoopFuture<ResultType>, timeout: TimeAmount) async throws -> ResultType {
```
fabianfett, line 202: We had a naming bikeshed here:

@dnadoba raised that `awaitFuture` will require us to write:
```swift
let foo = try await el.awaitFuture(future)
```
which has `await` twice and likely `future` twice, which is not nice to read.

---

@fabianfett proposed: `runUntil`, which IMO explicitly states the side-effects this method has.

@Lukasa didn't like `runUntil` and proposed `unwrap`.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,405 @@
...
+///     loop.execute { secondTask() }
+///     loop.execute { thirdTask() }
+/// }
+/// ```
+///
+/// There is a tricky requirement around waiting for `EventLoopFuture`s when working with this
+/// event loop. Simply calling `.wait()` from the test thread will never complete. This is because
+/// `wait` calls `loop.execute` under the hood, and that callback cannot execute without calling
+/// `loop.run()`. As a result, if you need to await an `EventLoopFuture` created on this loop you
+/// should use `awaitFuture`.
```
fabianfett, line 57: Do we want to add a callback to the `EventLoop` protocol, to check if an EventLoop supports awaiting a future without explicitly running? If it doesn't support it, we could crash in `wait()`.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    /// - see: `EventLoop.scheduleTask(in:_:)`
+    @discardableResult
+    public func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T> {
+        return self.scheduleTask(deadline: self.now + `in`, task)
+    }
+
+    /// On an `EmbeddedEventLoop`, `execute` will simply use `scheduleTask` with a deadline of _now_. This means that
+    /// `task` will be run the next time you call `EmbeddedEventLoop.run`.
```
fabianfett, line 175: Code comment should be tailored to `AsyncEmbeddedEventLoop`.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,405 @@
...
+                self._now.store(newTime.uptimeNanoseconds)
+
+                continuation.resume()
+            }
+        }
+    }
+
+    /// Executes the given function in the context of this event loop. This is useful when it's necessary to be confident that an operation
+    /// is "blocking" the event loop. As long as you are executing, nothing else can execute in this loop.
+    public func executeInContext<ReturnType: Sendable>(_ task: @escaping @Sendable () throws -> ReturnType) async throws -> ReturnType {
```
fabianfett, line 282: I think we need more docu for this. IIUC this method is mainly to ensure that you can schedule multiple tasks at once, without side-effects getting into the order. However you execute before potentially scheduled work. 
In most cases, you don't want to actually run something here. However execute might indicate that.

By actually "blocking" the EL, you easily run into deadlocks here. Anyhow, it took me quite a moment to wrap my head around all of this. I guess more documentation will help here.

fabianfett (commented):
Looks very good. This is mostly requests for better docs and a bit of naming bikeshed.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: As discussed on the call I think the answer is "no": for now I don't think the diagnostic information we gather is sufficiently useful. We can always add this later if it turns out to be a problem.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Return type needs to be `Sendable` because we'll construct it in one thread and access it from another. Swift compiler will moan if we don't. :grin:

Tests/NIOEmbeddedTests/AsyncEmbeddedEventLoopTests.swift
```diff
@@ -0,0 +1,691 @@
...
+                XCTAssertEqual(callbackCount.load(), 1)
+            }
+            await loop.advanceTime(by: .nanoseconds(1))
+            try await loop.executeInContext {
+                XCTAssertEqual(callbackCount.load(), 1)
+            }
+            await loop.advanceTime(by: .hours(1))
+            try await loop.executeInContext {
+                XCTAssertEqual(callbackCount.load(), 1)
+            }
```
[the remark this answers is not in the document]
Lukasa: Nah, this is mainly here to enforce a cross-thread barrier that settles all the atomics. It's not _really_ necessary to be honest, so we could remove it.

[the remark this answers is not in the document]
Lukasa: Oh sorry, you mean the "jump forward by 1 hour". Yeah, that demonstrates that we don't call it again. It's a better check than attempting to observe the reference count.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: Added some documentation

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: I definitely don't like most of these names.

FranzBusch: Reading through these suggestions I like `advanceTime(until:)` the most. IMO it is important that we convey that the loop is running when using this method.

FranzBusch (commented):
This looks very good now! I think we just need to converge on the naming of one or two methods.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: One issue with this name is that this method does not advance time. It just runs the loop repeatedly. We wouldn't know what time to advance the loop to.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
```diff
@@ -0,0 +1,422 @@
...
+                // cancel the entire TaskGroup. Depending on how things get scheduled it is possible for that
+                // cancellation to cancel the runner task above _before_ it spins again, meaning that the
+                // promise failure never actually happens (as it has to establish event loop context). So
+                // before we cancel this we make sure that we get a chance to spin the loop.
+                await self.run()
+                throw NIOAsyncEmbeddedEventLoopError.timeoutAwaitingFuture
+            }
+
+            do {
+                // This force-unwrap is safe: there are only two tasks, one will either return or hang, and the
```
glbrntt, line 236: This comment needs an update now that there are three tasks

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
glbrntt: Since we should use this in favour of `EventLoopFuture.get()` what about just calling it `get`?

glbrntt (approved):
LGTM. Only sticking point is the naming of `awaitFuture`.

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
dnadoba: Maybe the solution is to just compose `run` and `get` together like this:
```swift
try await eventLoop.run { 
    let foo = try await eventLoop.get()
    // do something with foo
}
// or 
let foo = try await eventLoop.run { try await eventLoop.get() }
// do something with foo
```
we could also call it `runUntil`. `runWhile` or something similar but the basic idea is that user needs to compose them and there is no special function that does both at the same time.
WDYT?

dnadoba (approved):
After landing on a name for `awaitFuture` this looks good to me 👍

Sources/NIOEmbedded/AsyncEmbeddedEventLoop.swift
Lukasa: The downside here is that I think this is less obvious than a single function. It relies on the user understanding _why_ a regular `wait` will hang, and being able to observe that this behaviour works. I think the `func run(_:) async` that this implies is also a very weird function, as it's essentially a "run until the other task completes". I can't think of another situation that requires this kind of pattern, so I'm inclined to lean away from it.

Lukasa:
The consensus has been to keep `awaitFuture`, so that's what we're gonna do.

PR #2084, opened 2022-04-20
Add initial support for connected datagram sockets

simonjbeaumont:
### Motivation:

From the man page for `connect(2)`:
>     The parameter socket is a socket.  If it is of type SOCK_DGRAM, this call specifies the peer
>     with which the socket is to be associated; this address is that to which datagrams are to be
>     sent, and the only address from which datagrams are to be received.

This presents an opportunity for some performance optimisations for applications that want to use UDP with a fixed peer, but is currently prohibited in NIO.

This PR adds some preliminary support for connected datagram sockets and allow further work to optimise that code path.

### Modifications:

* Allow the calling of `sendmsg(2)` with `NULL` for `msg_name` by changing the `Socket` types.
* Keep track of whether the `DatagramChannel` is connected in `PendingDatagramWritesManager` and, if it is connected, use `NULL` for `msg_name` in the headers for scalar and vector writes.
* Allow connect to be called after bind in `BaseSocketChannel`.
* Allow `connectSocket(to:)` to be called on `DatagramChannel`.
* Rename `DatagramBootstrap.bind0(makeChannel:registerAndBind:)` to `withNewChannel(makeChannel:bringup:)` to account for the fact that it isn't only used for `bind`.
* Add set of `DatagramBootstrap.connect(...)` APIs.
* Remove DatagramChannelTests.testConnectionFails because that doesn't fail anymore.
* Add ConnectedDatagramChannelTests, inheriting from DatagramChannelTests but connecting both channels together.
* Use connected-mode UDP in NIOUDPEchoClient.

### Result:

* Can now call `Channel.connect(to:)` when using datagram sockets.
* Some minor performance improvements in `PendingDatagramWritesManager` because the address is no longer used (or copied).
* Opportunities for future performance enhancements.
* All tests pass on Darwin.

### Considerations:

* Should we add support for `disconnectx(2)`?
* To support `connect` after `bind` the state machine in `BaseSocketChannel` was made more lenient. Is this acceptable? 
* Is the `NIOUDPEchoClient` considered API? If not, can we rework the argument parsing so we can include a `--connect` flag?
* Need to make sure we're testing the codepaths on Windows (different socket API) and Linux (which uses `sendmmsg` for vector writes).

### DIFF
--- a/Sources/NIOPosix/BaseSocketChannel.swift
+++ b/Sources/NIOPosix/BaseSocketChannel.swift
@@ -140,12 +140,18 @@ private struct SocketChannelLifecycleManager {
                 pipeline.syncOperations.fireChannelUnregistered()
             }
 
+        // TODO: This is to support connect after bind for connected-mode datagram sockets. Might need a better solution for this.
+        // origin: .activated
+        case (.activated, .activate):
+            return { promise, pipeline in
+                promise?.succeed(())
+            }
+
         // bad transitions
         case (.fresh, .activate),                  // should go through .registered first
              (.preRegistered, .activate),       // need to first be fully registered
              (.preRegistered, .beginRegistration), // already registered
              (.fullyRegistered, .beginRegistration),  // already registered
-             (.activated, .activate),              // already activated
              (.activated, .beginRegistration),        // already fully registered (and activated)
              (.activated, .finishRegistration),         // already fully registered (and activated)
              (.fullyRegistered, .finishRegistration),   // already fully registered
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -914,6 +914,74 @@ public final class DatagramBootstrap {
         }
     }
 
+    /// Connect the `DatagramChannel` to `host` and `port`.
+    ///
+    /// - parameters:
+    ///     - host: The host to connect to.
+    ///     - port: The port to connect to.
+    ///
+    /// - NOTE: This function is specifically for connected-mode UDP and does not call `bind`.
+    public func connect(host: String, port: Int) -> EventLoopFuture<Channel> {
+        return connect0 {
+            return try SocketAddress.makeAddressResolvingHost(host, port: port)
+        }
+    }
+
+    /// Connect the `DatagramChannel` to `address`.
+    ///
+    /// - parameters:
+    ///     - address: The `SocketAddress` to connect to.
+    ///
+    /// - NOTE: This function is specifically for connected-mode UDP and does not call `bind`.
+    public func connect(to address: SocketAddress) -> EventLoopFuture<Channel> {
+        return connect0 { address }
+    }
+
+    /// Connect the `DatagramChannel` to a UNIX Domain Socket.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The path of the UNIX Domain Socket to connect to. `path` must not exist, it will be created by the system.
+    public func connect(unixDomainSocketPath: String) -> EventLoopFuture<Channel> {
+        return bind0 {
+            return try SocketAddress(unixDomainSocketPath: unixDomainSocketPath)
+        }
+    }
+
+    /// Connect the `DatagramChannel` to a UNIX Domain Socket.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The path of the UNIX Domain Socket to connect to. `path` must not exist, it will be created by the system.
+    ///     - cleanupExistingSocketFile: Whether to cleanup an existing socket file at `path`.
+    public func connect(unixDomainSocketPath: String, cleanupExistingSocketFile: Bool) -> EventLoopFuture<Channel> {
+        if cleanupExistingSocketFile {
+            do {
+                try BaseSocket.cleanupSocket(unixDomainSocketPath: unixDomainSocketPath)
+            } catch {
+                return group.next().makeFailedFuture(error)
+            }
+        }
+
+        return self.bind(unixDomainSocketPath: unixDomainSocketPath)
+    }
+
+    private func connect0(_ makeSocketAddress: () throws -> SocketAddress) -> EventLoopFuture<Channel> {
+        let address: SocketAddress
+        do {
+            address = try makeSocketAddress()
+        } catch {
+            return group.next().makeFailedFuture(error)
+        }
+        func makeChannel(_ eventLoop: SelectableEventLoop) throws -> DatagramChannel {
+            return try DatagramChannel(eventLoop: eventLoop,
+                                       protocolFamily: address.protocol)
+        }
+        return withNewChannel(makeChannel: makeChannel) { (eventLoop, channel) in
+            channel.register().flatMap {
+                channel.connect(to: address)
+            }
+        }
+    }
+
     private func withNewChannel(makeChannel: (_ eventLoop: SelectableEventLoop) throws -> DatagramChannel, _ bringup: @escaping (EventLoop, DatagramChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
         let eventLoop = self.group.next()
         let channelInitializer = self.channelInitializer ?? { _ in eventLoop.makeSucceededFuture(()) }
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -843,7 +843,7 @@ public final class DatagramBootstrap {
         func makeChannel(_ eventLoop: SelectableEventLoop) throws -> DatagramChannel {
             return try DatagramChannel(eventLoop: eventLoop, socket: socket)
         }
-        return bind0(makeChannel: makeChannel) { (eventLoop, channel) in
+        return withNewChannel(makeChannel: makeChannel) { (eventLoop, channel) in
             let promise = eventLoop.makePromise(of: Void.self)
             channel.registerAlreadyConfigured0(promise: promise)
             return promise.futureResult
@@ -907,14 +907,14 @@ public final class DatagramBootstrap {
             return try DatagramChannel(eventLoop: eventLoop,
                                        protocolFamily: address.protocol)
         }
-        return bind0(makeChannel: makeChannel) { (eventLoop, channel) in
+        return withNewChannel(makeChannel: makeChannel) { (eventLoop, channel) in
             channel.register().flatMap {
                 channel.bind(to: address)
             }
         }
     }
 
-    private func bind0(makeChannel: (_ eventLoop: SelectableEventLoop) throws -> DatagramChannel, _ registerAndBind: @escaping (EventLoop, DatagramChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
+    private func withNewChannel(makeChannel: (_ eventLoop: SelectableEventLoop) throws -> DatagramChannel, _ bringup: @escaping (EventLoop, DatagramChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
         let eventLoop = self.group.next()
         let channelInitializer = self.channelInitializer ?? { _ in eventLoop.makeSucceededFuture(()) }
         let channelOptions = self._channelOptions
@@ -932,7 +932,7 @@ public final class DatagramBootstrap {
                 channelInitializer(channel)
             }.flatMap {
                 eventLoop.assertInEventLoop()
-                return registerAndBind(eventLoop, channel)
+                return bringup(eventLoop, channel)
             }.map {
                 channel
             }.flatMapError { error in
--- a/Sources/NIOPosix/PendingDatagramWritesManager.swift
+++ b/Sources/NIOPosix/PendingDatagramWritesManager.swift
@@ -445,7 +445,7 @@ final class PendingDatagramWritesManager: PendingWritesManager {
     ///     - scalarWriteOperation: An operation that writes a single, contiguous array of bytes (usually `sendmsg`).
     ///     - vectorWriteOperation: An operation that writes multiple contiguous arrays of bytes (usually `sendmmsg`).
     /// - returns: The `WriteResult` and whether the `Channel` is now writable.
-    func triggerAppropriateWriteOperations(scalarWriteOperation: (UnsafeRawBufferPointer, UnsafePointer<sockaddr>, socklen_t, AddressedEnvelope<ByteBuffer>.Metadata?) throws -> IOResult<Int>,
+    func triggerAppropriateWriteOperations(scalarWriteOperation: (UnsafeRawBufferPointer, UnsafePointer<sockaddr>?, socklen_t, AddressedEnvelope<ByteBuffer>.Metadata?) throws -> IOResult<Int>,
                                            vectorWriteOperation: (UnsafeMutableBufferPointer<MMsgHdr>) throws -> IOResult<Int>) throws -> OverallWriteResult {
         return try self.triggerWriteOperations { writeMechanism in
             switch writeMechanism {
@@ -515,7 +515,7 @@ final class PendingDatagramWritesManager: PendingWritesManager {
     ///
     /// - parameters:
     ///     - scalarWriteOperation: An operation that writes a single, contiguous array of bytes (usually `sendto`).
-    private func triggerScalarBufferWrite(scalarWriteOperation: (UnsafeRawBufferPointer, UnsafePointer<sockaddr>, socklen_t, AddressedEnvelope<ByteBuffer>.Metadata?) throws -> IOResult<Int>) rethrows -> OneWriteOperationResult {
+    private func triggerScalarBufferWrite(scalarWriteOperation: (UnsafeRawBufferPointer, UnsafePointer<sockaddr>?, socklen_t, AddressedEnvelope<ByteBuffer>.Metadata?) throws -> IOResult<Int>) rethrows -> OneWriteOperationResult {
         assert(self.state.isFlushPending && self.isOpen && !self.state.isEmpty,
                "illegal state for scalar datagram write operation: flushPending: \(self.state.isFlushPending), isOpen: \(self.isOpen), empty: \(self.state.isEmpty)")
         let pending = self.state.nextWrite!
--- a/Sources/NIOPosix/PendingDatagramWritesManager.swift
+++ b/Sources/NIOPosix/PendingDatagramWritesManager.swift
@@ -140,6 +140,7 @@ private struct PendingDatagramWritesState {
     private var pendingWrites = MarkedCircularBuffer<PendingDatagramWrite>(initialCapacity: 16)
     private var chunks: Int = 0
     public private(set) var bytes: Int64 = 0
+    private(set) var isConnected: Bool = false
 
     public var nextWrite: PendingDatagramWrite? {
         return self.pendingWrites.first
@@ -192,6 +193,10 @@ private struct PendingDatagramWritesState {
     /// All writes before this checkpoint will eventually be written to the socket.
     public mutating func markFlushCheckpoint() {
         self.pendingWrites.mark()
+    }
+
+    mutating func markConnected() {
+        self.isConnected = true
     }
 
     /// Indicate that a write has happened, this may be a write of multiple outstanding writes (using for example `sendmmsg`).
@@ -400,6 +405,10 @@ final class PendingDatagramWritesManager: PendingWritesManager {
     /// Mark the flush checkpoint.
     func markFlushCheckpoint() {
         self.state.markFlushCheckpoint()
+    }
+
+    func markConnected() {
+        self.state.markConnected()
     }
 
     /// Is there a flush pending?
--- a/Sources/NIOPosix/PendingDatagramWritesManager.swift
+++ b/Sources/NIOPosix/PendingDatagramWritesManager.swift
@@ -99,14 +99,31 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
 
         p.data.withUnsafeReadableBytesWithStorageManagement { ptr, storageRef in
             storageRefs[c] = storageRef.retain()
-            let addressLen = p.copySocketAddress(addresses.baseAddress! + c)
+
+            /// From man page of `sendmsg(2)`:
+            ///
+            /// > The `msg_name` field is used on an unconnected socket to specify
+            /// > the target address for a datagram.  It points to a buffer
+            /// > containing the address; the `msg_namelen` field should be set to
+            /// > the size of the address.  For a connected socket, these fields
+            /// > should be specified as `NULL` and 0, respectively.
+            let address: UnsafeMutablePointer<sockaddr_storage>?
+            let addressLen: socklen_t
+            if pending.isConnected {
+                address = nil
+                addressLen = 0
+            } else {
+                address = addresses.baseAddress! + c
+                addressLen = p.copySocketAddress(address!)
+            }
+
             iovecs[c] = iovec(iov_base: UnsafeMutableRawPointer(mutating: ptr.baseAddress!), iov_len: numericCast(toWriteForThisBuffer))
 
             var controlBytes = UnsafeOutboundControlBytes(controlBytes: controlMessageStorage[c])
             controlBytes.appendExplicitCongestionState(metadata: p.metadata, protocolFamily: p.address.protocol)
             let controlMessageBytePointer = controlBytes.validControlBytes
 
-            let msg = msghdr(msg_name: addresses.baseAddress! + c,
+            let msg = msghdr(msg_name: address,
                              msg_namelen: addressLen,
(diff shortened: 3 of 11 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -907,14 +907,82 @@ public final class DatagramBootstrap {
...
+    public func connect(to address: SocketAddress) -> EventLoopFuture<Channel> {
+        return connect0 { address }
+    }
+
+    /// Connect the `DatagramChannel` to a UNIX Domain Socket.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The path of the UNIX Domain Socket to connect to. `path` must not exist, it will be created by the system.
+    public func connect(unixDomainSocketPath: String) -> EventLoopFuture<Channel> {
+        return bind0 {
```
Lukasa, line 945: Should this still call `bind0`? (This suggests a missing test)

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -907,14 +907,82 @@ public final class DatagramBootstrap {
...
+            return try SocketAddress(unixDomainSocketPath: unixDomainSocketPath)
+        }
+    }
+
+    /// Connect the `DatagramChannel` to a UNIX Domain Socket.
+    ///
+    /// - parameters:
+    ///     - unixDomainSocketPath: The path of the UNIX Domain Socket to connect to. `path` must not exist, it will be created by the system.
+    ///     - cleanupExistingSocketFile: Whether to cleanup an existing socket file at `path`.
+    public func connect(unixDomainSocketPath: String, cleanupExistingSocketFile: Bool) -> EventLoopFuture<Channel> {
```
Lukasa, line 955: I don't think `cleanupExistingSocketFile` makes sense on a `connect` call, so we don't need this path.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -515,14 +541,28 @@ final class PendingDatagramWritesManager: PendingWritesManager {
...
+    private func triggerScalarBufferWrite(scalarWriteOperation: (UnsafeRawBufferPointer, UnsafePointer<sockaddr>?, socklen_t, AddressedEnvelope<ByteBuffer>.Metadata?) throws -> IOResult<Int>) rethrows -> OneWriteOperationResult {
         assert(self.state.isFlushPending && self.isOpen && !self.state.isEmpty,
                "illegal state for scalar datagram write operation: flushPending: \(self.state.isFlushPending), isOpen: \(self.isOpen), empty: \(self.state.isEmpty)")
         let pending = self.state.nextWrite!
         do {
-            let writeResult = try pending.address.withSockAddr { (addrPtr, addrSize) in
-                try pending.data.withUnsafeReadableBytes {
-                    try scalarWriteOperation($0, addrPtr, socklen_t(addrSize), pending.metadata)
+            let writeResult: IOResult<Int>
+            if self.state.isConnected {
```
Lukasa, line 550: I think in this case we want to confirm that any write we're about to send has an address that matches the address we connected to.

Sources/NIOUDPEchoClient/main.swift
```diff
@@ -126,14 +126,23 @@ defer {
...
 
-let channel = try { () -> Channel in
-    switch connectTarget {
-    case .ip(let host, _, let listeningPort):
-        return try bootstrap.bind(host: host, port: listeningPort).wait()
-    case .unixDomainSocket(_, let listeningPath):
-        return try bootstrap.bind(unixDomainSocketPath: listeningPath).wait()
+
+let channel: Channel
+let connectedMode = true
```
Lukasa, line 131: Currently this is unconditionally `true`: did you want to add support for both?

Tests/NIOPosixTests/DatagramChannelTests.swift
```diff
@@ -912,3 +906,11 @@ final class DatagramChannelTests: XCTestCase {
         testEcnAndPacketInfoReceive(address: "::1", vectorRead: true, vectorSend: true, receivePacketInfo: true)
     }
 }
+
+final class ConnectedDatagramChannelTests: DatagramChannelTests {
```
Lukasa, line 910: I think this will stop working once we adopt test discovery on Linux, which strictly works structurally and cannot traverse into subclasses. Can we either use copy-paste or have two separate classes which both defer to a common implementation?

Sources/NIOPosix/BaseSocketChannel.swift
```diff
@@ -140,12 +140,18 @@ private struct SocketChannelLifecycleManager {
                 pipeline.syncOperations.fireChannelUnregistered()
             }
 
+        // TODO: This is to support connect after bind for connected-mode datagram sockets. Might need a better solution for this.
+        // origin: .activated
+        case (.activated, .activate):
```
Lukasa, line 145: Yeah we may want to rethink this a bit. I wonder if we can construct a system whereby only datagram channel allows this.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -907,14 +907,82 @@ public final class DatagramBootstrap {
...
     }
 
-    private func bind0(makeChannel: (_ eventLoop: SelectableEventLoop) throws -> DatagramChannel, _ registerAndBind: @escaping (EventLoop, DatagramChannel) -> EventLoopFuture<Void>) -> EventLoopFuture<Channel> {
+    /// Connect the `DatagramChannel` to `host` and `port`.
+    ///
+    /// - parameters:
+    ///     - host: The host to connect to.
+    ///     - port: The port to connect to.
+    ///
+    /// - NOTE: This function is specifically for connected-mode UDP and does not call `bind`.
```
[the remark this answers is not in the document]
simonjbeaumont: Ah, yes. These are leftover from before I allowed connect after bind. I'll get rid of them.

Addressed in af93fbea7.

Sources/NIOPosix/Bootstrap.swift
simonjbeaumont: Um, no. Copy-pasta error here. This should clearly be `connect0`. Addressed in af93fbea7.

Just checked and none of the existing `DatagramChannelTests` used anything other than `bind(host:port:)` and so the `ConnectedDatagramChannelTests` also don't use this path.

Sources/NIOPosix/PendingDatagramWritesManager.swift
simonjbeaumont: Oh, I was thinking that this would be an opportunity to get a small perf win. I realise there'll be a perf win within the call to `sendmsg`, but by not bothering with the `withSockAddr` here we avoid a copy. 

This comes down to a question about the layer of abstraction we want to provide here I suppose. Do we want to provide an API that checks you are doing the right thing or be more transparent?

I guess we are already abstracting `sendmsg` calls away because calling them with a `msg_name` will cause an error.

WDYT about checking that the address matches on the way _in_ to PDWM, in the call to `add` if it currently `isConnected`? This would make the actual writing out of data quicker?

Now we're talking about this, it's going to be even more complicated/costly in the vector write case to make this check for each of the message headers.

Do you think there's any milage in an API that, once you call `connect()` sends all datagrams to that address? This would be the most performant, but I wonder if it might mean we should be exposing different APIs for sending these messages (that don't take an `AddressedEnvelope`) and disallow the `AddressedEnvelope` calls when connected?

Sources/NIOUDPEchoClient/main.swift
simonjbeaumont: Yeah, that's what I meant by this in the considerations in the PR description:

> Is the NIOUDPEchoClient considered API? If not, can we rework the argument parsing so we can include a --connect flag?

If it's not considered a breaking change then I can add an option. Adding an option with the existing argument parsing strategy is gonna be pretty clunky.

Tests/NIOPosixTests/DatagramChannelTests.swift
simonjbeaumont: Oh, that's disappointing. I didn't realise that.

`grpc-swift` makes use of subclasses, in that every test inherits from `GRPCTestCase` and there is further subclassing for shared tests IIRC. It also makes use of test discovery on Linux, given it's `LinuxMain.swift` consists only of:

```swift
fatalError("Use --enable-test-discovery to run tests on Linux")
```

What am I missing that means we can't do the same in NIO?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -907,14 +907,82 @@ public final class DatagramBootstrap {
...
+            return try SocketAddress.makeAddressResolvingHost(host, port: port)
+        }
+    }
+
+    /// Connect the `DatagramChannel` to `address`.
+    ///
+    /// - parameters:
+    ///     - address: The `SocketAddress` to connect to.
+    ///
+    /// - NOTE: This function is specifically for connected-mode UDP and does not call `bind`.
```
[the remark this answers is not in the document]
simonjbeaumont: Addressed in af93fbea7.

Sources/NIOPosix/BaseSocketChannel.swift
simonjbeaumont: Yes, this was what I was thinking we might be wanting.

Additionally, we should consider whether we want to support calling `connect(2)` after `connect(2)` which seems to be a valid use case to re-associate a connected datagram socket with a new address.

Sources/NIOPosix/Bootstrap.swift
simonjbeaumont: Removed in af93fbea7.

Sources/NIOPosix/PendingDatagramWritesManager.swift
simonjbeaumont: Is a4ee0756d what you had in mind?

Lukasa: @simonjbeaumont Yeah the performance win is valuable, but I think we need to detect misuse of the API as well.

If we wanted to get the performance win I think we get it by allowing sending naked byte buffers instead of addressedenvelopes once we connect. If users still send addressed envelopes then we still have to check. I'm fine with us doing the check on the way into PDWM, but if we do that we also have to validate anything that's buffered when the `connect` call comes in.

Sources/NIOUDPEchoClient/main.swift
Lukasa: We can rewrite the argument parsing here.

Tests/NIOPosixTests/DatagramChannelTests.swift
Lukasa: We should double-check that all of grpc-swift's tests are actually running on Linux: my experiments suggest the answer is no.

Sources/NIOPosix/BaseSocketChannel.swift
Lukasa: Yeah we should probably allow that.

Tests/NIOPosixTests/DatagramChannelTests.swift
Lukasa: Looking into this:

- `GRPCTestCase` adds functionality but does not define tests, so that parent class is safe.
- `EchoTestCaseBase` adds functionality but does not define tests, so that parent class is safe.
- `FunctionalTestsInsecureTransport` defines tests, which is sketchy
- `GRPCInsecureInteroperabilityTests` defines tests, which is sketchy
- `ServerThrowingTests` defines tests, which is sketchy

Testing this locally I can confirm that the `FunctionalTestsInsecureTransport` subclasses `FunctionalTestsAnonymousClient` and `FunctionalTestsMutualAuthentication` do not run. The `GRPCInsecureInteroperabilityTests` subclass `GRPCSecureInteroperabilityTests` do not run. The `ServerThrowingTests` subclasses `ServerDelayedThrowingTests`, `ClientThrowingWhenServerReturningErrorTests`, `ServerErrorTransformingTests` do not run. This is a bug against grpc-swift.

Lukasa: Filed as grpc/grpc-swift#1392.

Sources/NIOPosix/PendingDatagramWritesManager.swift
simonjbeaumont: Hm, the presence of buffering makes this an interesting problem since it's possible the connected state of the socket can change while things are buffered. That's the case even with what's in this PR, but becomes even more apparent if we start supporting `disconnectx` and re-`connect`.

Sources/NIOPosix/BaseSocketChannel.swift
simonjbeaumont: This is now handled in ae1b8d409 with some tests.

Sources/NIOPosix/PendingDatagramWritesManager.swift
simonjbeaumont: OK so I reverted the commit that checks the writes when flushing and replaced it with ae1b8d409 which fails any buffered writes when connecting or reconnecting the datagram channel and validates the destination address on the way in, when the channel is connected.

Also added some tests.

Tests/NIOPosixTests/DatagramChannelTests.swift
simonjbeaumont: I've added tests for the connected support to the existing class for now. I'll resolve this thread and if it turns out that we want to have separate suite, I'll make sure I don't make the mistake of subclassing :)

Sources/NIOUDPEchoClient/main.swift
simonjbeaumont: Is depending on `swift-argument-parser` a no-go?

simonjbeaumont: Added commit:

```
* 4d5652cc6 NIOUDPEchoClient: Connect socket to remote only if --connect is used
```

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -672,6 +684,14 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
     override func bufferPendingWrite(data: NIOAny, promise: EventLoopPromise<Void>?) {
         let data = self.unwrapData(data, as: AddressedEnvelope<ByteBuffer>.self)
 
+        // If the datagram socket is connected then we check here that the remote is appropriate.
+        if let remoteAddress = self.remoteAddress {
+            guard data.remoteAddress == remoteAddress else {
+                promise?.fail(IOError(errnoCode: EISCONN, reason: "Socket was connected before flushing pending write."))
+                return
+            }
+        }
```
simonjbeaumont, line 693: @Lukasa I'm thinking that we might get even more perf win here once PDWM supports `ByteBuffer`. In this branch we could buffer only the buffer, without the envelope. Then we wouldn't even need to do the address stripping on the way out in PDWM flush?

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -526,12 +526,24 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
     }
 
     override func connectSocket(to address: SocketAddress) throws -> Bool {
-        // For now we don't support operating in connected mode for datagram channels.
-        throw ChannelError.operationUnsupported
+        // TODO: this could be a channel option to do other things instead here, e.g. fail the connect
+        if !self.pendingWrites.isEmpty {
```
simonjbeaumont, line 530: @Lukasa Do I need to consider `self.pendingWrites.isPendingFlush` here? I guess if that's `true` then it means that the API has been used correctly. Or do we still leave it to the user to ensure that the flush has taken place, by means of the associated future?

Sources/NIOPosix/SocketChannel.swift
simonjbeaumont: 👍 Is that something we want in the first PR or can we do this incrementally?

simonjbeaumont: OK, great. I added support for sending unaddressed `ByteBuffer` and have added tests showing how users should only use `AddressedEnvelope` with unconnected channels and `ByteBuffer` with connected channels.

simonjbeaumont:
@Lukasa I added the `ByteBuffer` based write APIs which take a faster path and added some more testing. I think this is ready for another pass when you have time.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -668,9 +680,44 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+            assert(self.isActive)
+            self.pipeline.syncOperations.fireChannelWritabilityChanged()
+        }
+    }
+
+    /// Buffer a write in preparation for a flush.
+    private func bufferPendingAddressedWrite(data: AddressedEnvelope<ByteBuffer>, promise: EventLoopPromise<Void>?) {
+        // It is only appropriate to use an AddressedEnvelope if the socket is not connected.
+        guard self.remoteAddress == nil else {
+            promise?.fail(IOError(errnoCode: EISCONN, reason: """
```
Lukasa, line 716: Here we should only error if the remote address doesn't match the address in the envelope, but otherwise it's totally fine.

Lukasa (commented):
Great stuff! I've skipped reviewing the tests for now while I have feedback in the product code to speed up our turnaround time.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -433,6 +463,27 @@ final class PendingDatagramWritesManager: PendingWritesManager {
...
+    ///     - promise: Optionally an `EventLoopPromise` that will get the write operation's result
+    /// - result: If the `Channel` is still writable after adding the write of `data`.
+    func add(data: ByteBuffer, promise: EventLoopPromise<Void>?) -> Bool {
+        assert(self.isOpen)
+        self.state.append(.init(data: data,
+                                promise: promise,
+                                address: nil,
+                                metadata: nil))
+
+        if self.state.bytes > waterMark.high && channelWritabilityFlag.compareAndExchange(expected: true, desired: false) {
```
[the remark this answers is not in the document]
simonjbeaumont: Done in 501cac07a.

simonjbeaumont (commented):
Thanks for the latest comments, @Lukasa!

I've refactored the common logic out of the two `PDWM.add(...)` functions.

I have some thoughts regarding the API semantics which I've added in https://github.com/apple/swift-nio/pull/2084/files#r858949312 that might influence the next steps.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -515,14 +566,31 @@ final class PendingDatagramWritesManager: PendingWritesManager {
...
                "illegal state for scalar datagram write operation: flushPending: \(self.state.isFlushPending), isOpen: \(self.isOpen), empty: \(self.state.isEmpty)")
         let pending = self.state.nextWrite!
         do {
-            let writeResult = try pending.address.withSockAddr { (addrPtr, addrSize) in
-                try pending.data.withUnsafeReadableBytes {
-                    try scalarWriteOperation($0, addrPtr, socklen_t(addrSize), pending.metadata)
+            let writeResult: IOResult<Int>
+
+            if let address = pending.address {
+                assert(!self.state.isConnected, "Pending write with address on connected socket.")
```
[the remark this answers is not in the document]
Lukasa: So the problem with reflecting that semantics is the issue of `AddressedEnvelope.Metadata`. This is still meaningful to apply even in connected mode, but it's not accessible to users anymore if they connect. I don't think that's an intended outcome. Supporting both and just fixing it up allows users to wish to to keep sending the metadata but still gives access to a fast path. Does that make sense?

I don't think this conflicts with late-binding the address though: the `DatagramChannel` just needs to communicate the metadata to the PDWM, at which point it can safely drop the address before storing into the `PendingDatagramWrite`. Does that make sense?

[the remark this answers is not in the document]
Lukasa: Also you're right about the vector case, my bad, I just missed it.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -526,12 +526,24 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
     }
 
     override func connectSocket(to address: SocketAddress) throws -> Bool {
-        // For now we don't support operating in connected mode for datagram channels.
-        throw ChannelError.operationUnsupported
+        // TODO: this could be a channel option to do other things instead here, e.g. fail the connect
+        if !self.pendingWrites.isEmpty {
+            self.pendingWrites.failAll(
+                error: IOError(
```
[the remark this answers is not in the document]
Lukasa: I think a NIO-specific error adds value because it helps us debug the exact cause of the problem. IMO, NIO shouldn't pretend that it isn't there when it is doing things that are unique and specific to it. In this case, we're making the decision to drop some data, and we should own that instead of pretending that the OS failed them.

Sources/NIOPosix/PendingDatagramWritesManager.swift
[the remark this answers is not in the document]
simonjbeaumont: Your statements make sense. I thought, however, we wanted to offer APIs in terms of `ByteBuffer` and if we offer such APIs we should guide/force users to use the appropriate one.

Right, the point about the metadata is interesting and I came across it when I was making changes to PDWM. The additional flags passed to the `sendmsg` are a function of not just the `AddressedEnvelope.metadada` but also the `AddressedEnvelope.address.protocol`. I played around with having this property lifted into the `PendingDatagramWrite` so that it can be used even when `AddressedEnvelope.address` is `nil`.

I think if we want to continue to support these additional bits of metadata along with the write, which I completely expect that we do, we have (at least) two options:

1. `DatagramChannel.writeToSocket(_:)` accepts just an `AddressedEnvelope` and the implementation (either channel, PDWM, or both) discards the address along the way. This is an implicit API which might be harder to reason about.
2. Expose a more  explicit API that takes `ByteBuffer` and expect users to use it when connected. This API would need some way of also passing in the metadata that is appropriate, i.e. the congestion control mechanism currently composed of `AddressedEnvelope.metadata` and `AddressedEnvelope.address.protocol`, without passing the the metadata that is inappropriate, i.e. the destination address.

Sources/NIOPosix/SocketChannel.swift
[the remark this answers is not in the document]
simonjbeaumont: That reasoning makes total sense; I'll work on a more explicit error for this case where buffered writes are failed.

Sources/NIOPosix/PendingDatagramWritesManager.swift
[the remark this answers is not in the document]
Lukasa: Nope, my preference is that `PendingDatagramWrite` has a `nil` address when we don't want it, and then just restructure how we construct the metadata to use the connected address from `DatagramChannel` to populate the `protocol` for the metadata early.

As to the broader question of forcing folks to use the "right" API, I think unless we're willing to change the write to be "ByteBuffer + optional metadata" (which I think is not great) then it's probably fine to just support `AddressedEnvelope` for now in cases where folks want to send ECN.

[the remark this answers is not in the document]
simonjbeaumont: OK, I relaxed support to allow for `AddressedEnvelope` in writes to a connected `DatagramChannel` in eb580d69c. It validates that the address in the envelope matches that of the connected socket.

Sources/NIOPosix/SocketChannel.swift
simonjbeaumont: Resolved by https://github.com/apple/swift-nio/pull/2084/commits/eb580d69ca04991d91b7324fee725ca25bce9af9.

simonjbeaumont:
@Lukasa LMK if you have any additional feedback on this one.

Sources/NIOPosix/BaseSocketChannel.swift
```diff
@@ -140,12 +140,18 @@ private struct SocketChannelLifecycleManager {
                 pipeline.syncOperations.fireChannelUnregistered()
             }
 
+        // TODO: This is to support connect after bind for connected-mode datagram sockets. Might need a better solution for this.
```
Lukasa, line 143: Does this TODO still apply?

simonjbeaumont: Yes, currently it is required. Without it, we get the following error.

```
NIOPosix/BaseSocketChannel.swift:166: Fatal error: illegal transition: state=activated, event=activate
2022-05-23 11:07:35.867416+0100 xctest[6470:11648282] NIOPosix/BaseSocketChannel.swift:166: Fatal error: illegal transition: state=activated, event=activate
```

I suppose, if we want to only allow this on `DatagramChannel` then we would need to change the implementation of `BaseSocketChannel.moveState(event:)` (currently private) to `fatalError("must be overridden")` and provide different implementations in `BaseStreamSocket` and `DatagramChannel`.

WDYT?

PR #2086, opened 2022-04-22
BaseSocket: Remove dead code in BaseSocket.bind(to:)

simonjbeaumont:
### Motivation:

There's a local function created in `BaseSocket.bind(to:)` which doesn't appear to be used and is probably vestigial following an historic refactor.

### Modifications:

Remove the unused local function.

### Result:

Less unused code.


simonjbeaumont:
@Lukasa Just a trivial PR that resulted from digging around the socket code.

Lukasa (approved):
Nice catch, thanks! :sparkle:

PR #2088, opened 2022-04-22
EventLoopFuture: Support cancellation in async get()

### LINKED ISSUE #2087: EventLoopFuture.get should support cancellation
Right now `EventLoopFuture.get` is not cancellable: if users cancel a `Task` blocked in that call, nothing happens. This is a bad behaviour, and we should fix it.

simonjbeaumont:
### Motivation:

`EventLoopFuture.get() async throws` does not support cancellation. If the task from which it is being called is cancelled it should unblock. 

### Modifications:

Add cancellation handler in `EventLoopFuture.get()` and set the future value to `CancellationError()` if the waiting task is cancelled.

### Result:

Cancelling the task waiting for a future will fail the future with `CancellationError()`.

Fixes #2087.

### DIFF
--- a/Sources/NIOCore/AsyncAwaitSupport.swift
+++ b/Sources/NIOCore/AsyncAwaitSupport.swift
@@ -28,13 +28,23 @@ extension EventLoopFuture {
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
     @inlinable
     public func get() async throws -> Value {
-        return try await withUnsafeThrowingContinuation { cont in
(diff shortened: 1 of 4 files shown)


Lukasa (changes requested):
Sadly, we cannot complete the future with cancellation just because one waiter is cancelled.  This will affect other waiters, which isn’t great. We’re gonna need another system. I gave it a shot in #2089, which isn’t gonna work either I think. And @dnadoba has raised some good questions in #2087.

PR #2090, opened 2022-04-28
Adopt `Sendable` in NIOConcurrencyHelpers

Davidde94:
Stop `Sendable` related warnings when using types defined inside `NIOConcurrencyHelpers`. All the code should actually already be thread safe, so no changes needed other than labelling some classes as either `Sendable` or `@unchecked Sendable`.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOAtomic.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOAtomic.swift
@@ -311,3 +311,9 @@ public final class NIOAtomic<T: NIOAtomicPrimitive> {
         }
     }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension NIOAtomic: Sendable {
+
+}
+#endif
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -286,3 +286,12 @@ public final class ConditionLock<T: Equatable> {
 internal func debugOnly(_ body: () -> Void) {
     assert({ body(); return true }())
 }
+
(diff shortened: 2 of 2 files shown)


Sources/NIOConcurrencyHelpers/NIOAtomic.swift
```diff
@@ -311,3 +311,9 @@ public final class NIOAtomic<T: NIOAtomicPrimitive> {
         }
     }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension NIOAtomic: Sendable {
+
+}
+#endif
```
glbrntt, line 319: This one can just be `extension NIOAtomic: NIOSendable` (`NIOSendable` is just a typealias to `Sendable` when it's available and `Any` when it isn't)

Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -286,3 +286,12 @@ public final class ConditionLock<T: Equatable> {
 internal func debugOnly(_ body: () -> Void) {
     assert({ body(); return true }())
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension Lock: Sendable {
```
glbrntt, line 291: Same for this one, just `NIOSendable`

Sources/NIOConcurrencyHelpers/NIOAtomic.swift
Davidde94: That will involve depending on `NIOCore` though, is that something we're ok doing?

glbrntt: Very good point. No, probably not.

glbrntt (approved):

PR #2093, opened 2022-05-03
Move tests of NIOCore types from NIOPosixTests to NIOCoreTests

### LINKED ISSUE #2016: Split `NIOPosix` tests up

simonjbeaumont:
### Motivation:

As described in #2016, there are many tests in `NIOPosixTests` for types that have since moved from `NIOPosix` to `NIOCore`. 

### Modifications:

Move the tests from `NIOPosixTests` that do not `import NIOPosix` into `NIOCoreTests`

### Result:

Tests for types defined in the `NIOCore` module are more likely to be located in `NIOCoreTests` test target.

Resolves #2016.


Lukasa (approved):
Nice one, solid change.

Lukasa:
You've got a few CI failures to fix.

simonjbeaumont:
> You've got a few CI failures to fix.

```
20:53:16 [196/272] Compiling NIOCoreTests LinuxTest+XCTest.swift
20:53:16 /code/Tests/NIOCoreTests/LinuxTest.swift:33:13: error: cannot find 'withTemporaryFile' in scope
20:53:16             withTemporaryFile(content: quota) { (_, quotaPath) -> Void in
```

... `LinuxTest`. Makes sense why this passed locally :) Will fix.

Lukasa (approved):

Lukasa:
Looks like some helper functions are missing.

Tests/NIOCoreTests/XCTest+Extensions.swift
```diff
@@ -0,0 +1,92 @@
...
+        XCTAssertNoThrow(try fileHandle.close())
+        XCTAssertEqual(0, unlink(path))
+    }
+    if let content = content {
+        let fileDesciptor = FileDescriptor(rawValue: fd)
+        try fileDesciptor.writeAll(content.utf8)
+        XCTAssertEqual(try fileDesciptor.seek(offset: 0, from: .start), 0)
+    }
+    return try body(fileHandle, path)
+}
```
simonjbeaumont, line 59: This was lifted from `NIOPosixTests/TestUtils.swift` and reimplemented to not use `Posix.pwrite`. However, in so doing, we pulled in `openTemporaryFile` (below) which still makes use of the direct `mkstemp`. This might not be OK. If it isn't we could move `LinuxTest` back to `NIOPosixTests` for now and revisit later.

Tests/NIOCoreTests/XCTest+Extensions.swift
```diff
@@ -0,0 +1,92 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIOCore
+import struct System.FileDescriptor
```
Lukasa, line 17: We don't currently have an expressed dependency on swift-system, so we can't use this as it stands.

simonjbeaumont: 😁 Of course...

simonjbeaumont:
@Lukasa I think this is all resolved now. The 5.7 and nightly pipelines are failing with some allocation counts. Presumably that's not related.

Lukasa:
Seems unlikely, yeah.

Tests/NIOCoreTests/XCTest+Extensions.swift
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func withTemporaryFile<T>(content: String? = nil, _ body: (NIOCore.NIOFileHandle, String) throws -> T) throws -> T {
+    let temporaryFilePath = "\(temporaryDirectory)/nio_\(UUID())"
+    FileManager.default.createFile(atPath: temporaryFilePath, contents: content?.data(using: .utf8))
+    defer {
+        XCTAssertNoThrow(try FileManager.default.removeItem(atPath: temporaryFilePath))
+    }
+
+    let fileHandle = try NIOFileHandle(path: temporaryFilePath, mode: .write)
```
Lukasa, line 52: This should probably be open `[.read, .write]`.

Lukasa (approved):

PR #2096, opened 2022-05-04
Adopt `Sendable` for `SocketAddress`, `AddressedEnvelope` and friends 

dnadoba:
Incremental progress on `Sendable` adoption.


Davidde94 (approved):

Sources/NIOCore/AddressedEnvelope.swift
```diff
@@ -62,8 +62,10 @@ extension AddressedEnvelope: Equatable where DataType: Equatable {}
 
 extension AddressedEnvelope: Hashable where DataType: Hashable {}
 
+extension AddressedEnvelope: NIOSendable where DataType: NIOSendable {}
```
Lukasa, line 65: This conformance isn't quite right: we don't need the `DataType` to conform to `NIOSendable`, we only need it to conform to `Sendable`. This will make this declaration more annoying to write, but we should still spell it correctly.

Lukasa (changes requested):

Sources/NIOCore/AddressedEnvelope.swift
Lukasa: Oh shoot so it is, good catch.

Lukasa (approved):
LGTM

PR #2098, opened 2022-05-06
Adopt `Sendable` for Types in `ChannelOptions` Namespace

dnadoba:

### DIFF
--- a/Sources/NIOCore/NIOSendable.swift
+++ b/Sources/NIOCore/NIOSendable.swift
@@ -17,3 +17,9 @@ public typealias NIOSendable = Swift.Sendable
 #else
 public typealias NIOSendable = Any
 #endif
+
+#if swift(>=5.6)
+@preconcurrency public protocol NIOPreconcurrencySendable: Sendable {}
+#else
+public protocol NIOPreconcurrencySendable {}
+#endif
--- a/Sources/NIOCore/ChannelOption.swift
+++ b/Sources/NIOCore/ChannelOption.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// A configuration option that can be set on a `Channel` to configure different behaviour.
-public protocol ChannelOption: Equatable {
+public protocol ChannelOption: Equatable, NIOPreconcurrencySendable {
     /// The type of the `ChannelOption`'s value.
     associatedtype Value
 }
@@ -386,6 +386,6 @@ extension ChannelOptions {
     }
 }
 
-#if swift(>=5.5) && canImport(_Concurrency)
+#if swift(>=5.6)
 extension ChannelOptions.Storage: @unchecked Sendable {}
 #endif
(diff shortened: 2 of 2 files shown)


dnadoba:
Rereading the code, I think we can't actually make `ChannelOptions.Storage` conform to `Sendable`. `ChannelOption` isn't `Sendable` and we can't add `Sendable` requirement because it would be a API breaking change. All of our types that conform to  `ChannelOption `conform to `Sendable` too but not necessarily custom user defined types.

Lukasa:
We can add Sendable to it by marking the protocol `@preconcurrency Sendable`. This is what has to be done for any protocol of this kind which we always expected to be `Sendable` (e.g. `Channel`, `EventLoop`).

dnadoba:
I have added a new type `NIOPreconcurrencySendable` which is defined as
```swift
#if swift(>=5.6)
@preconcurrency public protocol NIOPreconcurrencySendable: Sendable {}
#else
public protocol NIOPreconcurrencySendable {}
#endif
```
and similar to `NIOSendable`. This will make it easier to let protocols inherit from `Sendable` in a source stable way without duplicating the complete protocol definition, including documentation. This isn't a real problem for the `ChannelOption` protocol in this PR as it is only a couple of lines, but will be for others e.g. `Channel`.

`@preconcurrency` is only available in Swift 5.6 so I have made `ChannelOptions.Storage` only conform to Sendable in Swift 5.6 too.

Lukasa (approved):

Sources/NIOCore/NIOSendable.swift
```diff
@@ -17,3 +17,9 @@ public typealias NIOSendable = Swift.Sendable
 #else
 public typealias NIOSendable = Any
 #endif
+
+#if swift(>=5.6)
+@preconcurrency public protocol NIOPreconcurrencySendable: Sendable {}
+#else
+public protocol NIOPreconcurrencySendable {}
```
FranzBusch, line 24: I still believe these should be non `public`. They are just marker types for internal convenience. Adopters, still we see the correct types as `Sendable` and we are not exposing unnecessary protocols publicly.
```suggestion
@preconcurrency protocol NIOPreconcurrencySendable: Sendable {}
#else
public NIOPreconcurrencySendable {}
```

FranzBusch (changes requested):

Sources/NIOCore/NIOSendable.swift
Lukasa: Given that we've already shipped NIOSendable I think the symmetry argument shouldn't be overlooked here.

FranzBusch: I get that point, but should we expose more public API that is just an implementation detail because of symmetry reasons?

dnadoba: It needs to be public, otherwise public protocols (e.g. `ChannelOption` or `EventLoop`) can **not** inherit from it.

FranzBusch (approved):

Lukasa:
The API breakage checker is wrong: merging over it.

Issue #1957, opened 2021-09-15
Provide NonBlockingFileIO helpers that use Swift System


Lukasa:
As Swift System has gone 1.0 we can in principle take a dependency on the package. This might be a good opportunity to clean up NonBlockingFileIO abstraction to be rewritten in terms of Swift System's public types.

adam-fowler:
@Lukasa would this mean it can be moved into `NIOCore`

Lukasa:
In principle, yes.

simonjbeaumont:
@Lukasa I played around with this yesterday evening and had a question.

Are we thinking that we would preserve the current `NonBlockingFileIO` API surface as it currently is today, and just change the implementation detail under the hood, or would we want to deprecate the current API in favour of something new?

Assuming the need to preserve API, I found I needed to translate the `System.Errno` errors back into NIO's `NonBlockingFileIO.Error.descriptorSetToNonBlocking` and `IOError` types. After this, the various read and write APIs seem to satisfy the existing tests.

The potential gotcha here is the lack of an `ftruncate` equivalent in Swift System, unless I'm missing something?

Lukasa:
I think we'd ideally keep the existing API around for now, where we can, but we'd also want to add new API surface that consumes the types from swift-system as well. We can definitely then deprecate the older API in favour of the new one.

As for a missing `ftruncate`, that is something we should follow-up on with swift-system.

simonjbeaumont:
OK, I have a patch for Swift System to add a new API to Swift System for `ftruncate` functionality, I'll open a PR when I can.

simonjbeaumont:
> OK, I have a patch for Swift System to add a new API to Swift System for `ftruncate` functionality, I'll open a PR when I can.

https://github.com/apple/swift-system/pull/82

weissi:
@simonjbeaumont / @Lukasa what are the thoughts w.r.t. Swift System having an OS-dependent API?

Lukasa:
At the moment I think I'm most interested in going to the higher level API surface, but in general I don't think it should matter overmuch.

weissi:
@Lukasa are we still investing in NBFIO? If not, should we close?

Lukasa:
Agreed.

PR #2099, opened 2022-05-06
Use Swift System to implement NonBlockingFileIO

simonjbeaumont:
### Motivation:

#1957 suggests we make use of Swift System for the `NonBlockingFileIO` implementation which would also reduce its dependency on `NIOPosix`.
 
### Modifications:

* Add dependency on Swift System package.
* Replace implementation of `NonBlockingFileIO` with System's `FileDescriptor`-based implementation.

### To do:
- [x] Currently this PR depends on an unmerged PR in swift-system, this needs to be merged and then we need to depend on a released version.
- [ ] Decide if we want to expose the `withUnsafeSystemFileDescriptor` function (probably not initially?)


Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -500,3 +499,11 @@ public struct NonBlockingFileIO {
     }
 
 }
+
+extension NIOFileHandle {
+    public func withUnsafeSystemFileDescriptor<T>(_ body: (SystemPackage.FileDescriptor) throws -> T) throws -> T {
```
weissi, line 504: @compnerd/@milseman Does `swift-system` have `FileDescriptor` on Windows and all other platforms?

simonjbeaumont: It does have `FileDescriptor` on Windows but, now you mention it, we explicitly opted to _not_ include `resize()` on Windows. See https://github.com/apple/swift-system/pull/82.

compnerd: That is easily accomplished with `SetFilePointerEx` and `SetEndOfFile`; you will need `_get_osfhandle` to convert the file descriptor to the appropriate handle.

1proprogrammerchant (commented):
seems the commits are out of date to the branch

simonjbeaumont:
> seems the commits are out of date to the branch

Yeah, this one died on the vine. And the addition of `ftruncate` to swift-system for Windows got reverted because it broke the build and we don't have any CI on swift-system for Windows so I'm a bit hesitant to add it again.

Lukasa:
We’ve deprecated NonBlockingFileIO.

PR #2100, opened 2022-05-09
Adopt `Sendable` for `SocketOptionProvider`

dnadoba:


dnadoba:
API breakage failure is a false positive because of `NIOPreconcurrencySendable`:
```
18:03:18   💔 API breakage: protocol SocketOptionProvider has generic signature change from  to <Self : NIOCore.NIOPreconcurrencySendable>
18:03:18   💔 API breakage: protocol SocketOptionProvider has added inherited protocol NIOPreconcurrencySendable
18:03:18   💔 API breakage: protocol SocketOptionProvider has added inherited protocol Sendable
```

glbrntt (approved):

PR #2101, opened 2022-05-10
Adopt `Sendable` in `EventLoop.swift`

dnadoba:


Sources/NIOCore/EventLoop.swift
```diff
@@ -19,15 +19,28 @@ import Dispatch
 ///
 /// A `Scheduled` allows the user to either `cancel()` the execution of the scheduled task (if possible) or obtain a reference to the `EventLoopFuture` that
 /// will be notified once the execution is complete.
-public struct Scheduled<T> {
+public struct Scheduled<T>: NIOSendable {
```
Lukasa, line 22: Does this need to be `preconcurrency`? We can only be `Sendable` where the generic type is `Sendable`.

Sources/NIOCore/EventLoop.swift
```diff
@@ -220,7 +263,47 @@ public struct EventLoopIterator: Sequence, IteratorProtocol {
 public protocol EventLoop: EventLoopGroup {
     /// Returns `true` if the current `NIOThread` is the same as the `NIOThread` that is tied to this `EventLoop`. `false` otherwise.
     var inEventLoop: Bool { get }
+    
+    #if swift(>=5.6)
```
Lukasa, line 267: As a stylistic note, it might be better to put the `#if...#else` around every method. It'll make it easier to keep track of whether we've appropriately duplicated the methods.

Sources/NIOCore/EventLoop.swift
dnadoba: Good question and I think the question boils down to if `EventLoopPromise` should be unconditionality `Sendable` or not.

dnadoba: I have made it only conditionally `Sendable`  as `EventLoopPromise` should have been only conditionally `Sendable` as well.

Sources/NIOCore/EventLoop.swift
dnadoba: Done

Sources/NIOCore/EventLoop.swift
```diff
@@ -21,13 +21,26 @@ import Dispatch
...
     /* private but usableFromInline */ @usableFromInline let _promise: EventLoopPromise<T>
-    /* private but usableFromInline */ @usableFromInline let _cancellationTask: (() -> Void)
-
+    #if swift(>=5.6)
+    /* private but usableFromInline */ @usableFromInline let _cancellationTask: @Sendable () -> Void
+    #else
+    /* private but usableFromInline */ @usableFromInline let _cancellationTask: () -> Void
+    #endif
+    
+    #if swift(>=5.6)
```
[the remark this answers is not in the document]
dnadoba: Update all to 5.7

Sources/NIOCore/EventLoop.swift
```diff
@@ -54,15 +71,41 @@ public final class RepeatedTask {
...
+    private var task: Optional<@Sendable (RepeatedTask) -> EventLoopFuture<Void>>
+    #else
     private var task: Optional<(RepeatedTask) -> EventLoopFuture<Void>>
-
-    internal init(interval: TimeAmount, eventLoop: EventLoop, cancellationPromise: EventLoopPromise<Void>? = nil, task: @escaping (RepeatedTask) -> EventLoopFuture<Void>) {
+    #endif
+    
+    #if swift(>=5.6)
+    @preconcurrency
```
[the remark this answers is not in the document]
dnadoba: it isn't necessary here, removed.

Lukasa (approved):
Nice stuff, thanks @dnadoba.

Lukasa:
API breakage change is expected.

PR #2104, opened 2022-05-10
Adopt `Sendable` for `Channel`

dnadoba:

### DIFF
--- a/Sources/NIOCore/Channel.swift
+++ b/Sources/NIOCore/Channel.swift
@@ -102,7 +102,7 @@ public protocol ChannelCore: AnyObject {
 /// passed to or returned by the operations are used to retrieve the result of an operation after it has completed.
 ///
 /// A `Channel` owns its `ChannelPipeline` which handles all I/O events and requests associated with the `Channel`.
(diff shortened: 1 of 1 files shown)


Davidde94:
I don't think we want anything to do with `Channel` to be `Sendable` - everything should happen on the event loop.

glbrntt:
> I don't think we want anything to do with `Channel` to be `Sendable` - everything should happen on the event loop.

`Channel`s are thread-safe so this is fine.

Davidde94:
cc @Lukasa

Davidde94:
Also wouldn't this be a breaking change as it will require any implementation of `Channel` to conform?

Lukasa (approved):
`Channel`s are very much required to be `Sendable`: they are deliberately intended to be manipulated from multiple threads.

As to the fact that this will require conformances, thanks to the `@preconcurrency` annotation it will only require conformances once users start adopting concurrency. Anyone importing NIO who has _not_ adopted concurrency will not need to add a conformance.

Lukasa:
API breakage checker is mad about the new protocol.

fabianfett (approved):

PR #2107, opened 2022-05-11
Adopt `Sendable` for `EventLoopFuture` and `EventLoopPromise`

dnadoba:

We sadly need to duplicate a lot of code because `@preconcurrency` is only available in Swift 5.6+ and there is no other way to add this annotation in a source stable way.

### DIFF
--- a/Sources/NIOCore/AsyncAwaitSupport.swift
+++ b/Sources/NIOCore/AsyncAwaitSupport.swift
@@ -26,12 +26,12 @@ extension EventLoopFuture {
             self.whenComplete { result in
                 switch result {
                 case .success(let value):
-                    cont.resume(returning: value)
+                    cont.resume(returning: UnsafeTransfer(value))
                 case .failure(let error):
                     cont.resume(throwing: error)
                 }
             }
-        }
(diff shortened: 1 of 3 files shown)


dnadoba:

Lukasa (changes requested):
This diff has ended up being absolutely giant due to the need to duplicate code. Can we reduce duplication by slightly refactoring the code? I'm wondering whether the best thing to do is something like this:

```swift
#if swift(>=5.6)
internal typealias DoWhateverCallback = @Sendable (Value) -> Void

@preconcurrency
public func doWhatever(_ doWhateverCallback: @escaping @Sendable (Value) -> Void) {
    self._doWhatever(doWhateverCallback)
}
#else
internal typealias DoWhateverCallback = (Value) -> Void

public func doWhatever(_ doWhateverCallback: @escaping (Value) -> Void) {
    self._doWhatever(doWhateverCallback)
}
#endif

@inlinable
internal func _doWhatever(_ doWhateverCallback: @escaping DoWhateverCallback)
```

This way we'll have the appropriate type definition at the surface but call into a common implementation.

PR #2108, opened 2022-05-12
`ChannelOption.Value: Sendable`

dnadoba:
It looks like requiring `ChannelOption.Value` to conform to `Sendable` is a source breaking change e.g. our own `RecvAllocatorOption` no longer compiles as the `RecvByteBufferAllocator` protocol wasn't required to be `Sendable`.
To my knowledge, `@preconcurrency` can not be applied to associated types which would normally be the solution to this kind of problem.


Lukasa:
Agreed! This is done.

PR #2110, opened 2022-05-12
Deprecate unsafe methods for adding/removing non-`Sendable` `ChannelHandler`s

dnadoba:
### Motivation:

`ChannelHandler`s are not required to be thread-safe and we need to assume they can only be used on their `Channel`s `EventLoop`.

### Modifications:

- deprecate method which take non-`Sendable` `ChannelHandler`s
- add new methods which take `Sendable` `ChannelHandler`s

### Result:

Better thread-safety

### DIFF
--- a/Sources/NIOCore/AsyncAwaitSupport+OldXcodes.swift
+++ b/Sources/NIOCore/AsyncAwaitSupport+OldXcodes.swift
@@ -163,28 +163,6 @@ extension ChannelOutboundInvoker {
 
 extension ChannelPipeline {
     @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-    public func addHandler(_ handler: ChannelHandler,
-                           name: String? = nil,
-                           position: ChannelPipeline.Position = .last) async throws {
-        try await self.addHandler(handler, name: name, position: position).get()
-    }
-
-    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-    public func removeHandler(_ handler: RemovableChannelHandler) async throws {
-        try await self.removeHandler(handler).get()
-    }
-
-    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-    public func removeHandler(name: String) async throws {
-        try await self.removeHandler(name: name).get()
-    }
-
-    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
-    public func removeHandler(context: ChannelHandlerContext) async throws {
-        try await self.removeHandler(context: context).get()
-    }
-
-    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
     public func context(handler: ChannelHandler) async throws -> ChannelHandlerContext {
         return try await self.context(handler: handler).get()
     }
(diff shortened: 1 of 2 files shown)


Lukasa:
I'd like us to keep this open for a while so that we can discuss this with some other folks.

ktoso (commented):
I thought about this a little bit right now and checked any usages in projects I maintain... 

I think this looks right; since in practice handlers are not thread-safe by themselfes this checks out -- thanks for reaching out to check!

It definitely is a bit weird I guess; moving handlers into the pipeline would perhaps be cool, but we don't have that yet.

ktoso (approved):

weissi:
> I think this looks right; since in practice handlers are not thread-safe by themselfes this checks out

Yes and no. Almost all `ChannelHandler`s are indeed _not_ thread-safe, so at a first glance it makes sense to either deprecate them or require `Sendable` (which essentially is the same thing: they become unusable).

On the other hand however this deprecation doesn't make sense because you can trivially side-step it by using

```
try await channel.pipeline.addHandler(NonSendableHandler()).get()
```

instead of

```
try await channel.pipeline.addHandler(NonSendableHandler())
```

One other way, that I'm probably not recommending which would allow us to keep the same syntax is

```
func addHandler(_ handler: @Sendable @autoclosure () -> ChannelHandler) async throws
```

because that way you could write the always correct `try await channel.pipeline.addHandler(NonSendableHandler())` but not sometimes incorrect

```
let h = MyHandler()
try await channel.pipeline.addHandler(h)
```

In general, this requires design because NIO allows you to do many things safely (often by making sure two `Channel`s are on the same EL or smth) that cannot be expressed in `Sendable`.

Now the question becomes: Are we making the methods unavailable in Concurrency (which would likely lead users to just sidestep Concurrency by using the future API + `.get()`) or are we keeping them available in Concurrency allowing the user to do potentially racy stuff (as it is right now).

Lukasa:
> On the other hand however this deprecation doesn't make sense because you can trivially side-step it 

I get that limitation, but I think the awkward reality of the situation is that the side-step may also have to go away. Swift's `Sendable` semantics interact very awkwardly with the way we use `ChannelHandler`s here.

I think the `@autoclosure` idea isn't a bad one though.

weissi:
> > On the other hand however this deprecation doesn't make sense because you can trivially side-step it
> 
> I get that limitation, but I think the awkward reality of the situation is that the side-step may also have to go away.

Agreed. In the fulness of time, the NIO project could grow a better Concurrency <--> manually evented code bridge than the current EL(Future/Promise)<-->Concurrency bridge but that doesn't seem to be today (likely this will need to wait for custom executors, Swift 6 and a breaking release). So the question still is: Does it make sense to deprecate it _today_ where it's trivially bypasseable and kinda forces people down the wrong road...

>  Swift's `Sendable` semantics interact very awkwardly with the way we use `ChannelHandler`s here.

Agreed. It's not a common pattern anyway but it's obviously a thing you can do...

> I think the `@autoclosure` idea isn't a bad one though.

Yeah, dunno, it's an option that may buy us some time in this particular case....

dnadoba:
We could also add an async `modify(_:)` method which takes a closure which takes `SynchronousOperations` as an argument:
```swift
extension ChannelPipeline {
    public func modify<Result: Sendable>(
        _ callback: @Sendable (SynchronousOperations) throws -> Result
    ) async throws -> Result {
        //...
    }
}
```
which could safely be used to add `ChannelHandler`s as the closure is executed on the  `EventLoop` of the `Channel`/`ChannelPipeline`. 
`NIOAsycnAwaitDemo` would use it like this:
```swift
try await channel.pipeline.modify {
    try $0.addHandler(NIOHTTPClientResponseAggregator(maxContentLength: 1_000_000))
    try $0.addHandler(MakeFullRequestHandler())
}
```

dnadoba:
> One other thing (that I'm probably not recommending) which would allow us to keep the same syntax is
> 
> ```
> func addHandler(_ handler: @Sendable @autoclosure () -> ChannelHandler) async throws
> ```
> 
> because that way you could write the always correct `try await channel.pipeline.addHandler(NonSendableHandler())` but not the sometimes incorrect
> 
> ```
> let h = MyHandler()
> try await channel.pipeline.addHandler(h)
> ```
@weissi if used incorrectly we will actually get a warning because the closure is marked as `@Sendable`
```
Capture of 'h' with non-sendable type 'MyHandler' in a `@Sendable` closure
```

But how would we add such a method without breaking source stability?

PR #2113, opened 2022-05-13
Make `EventLoopFuture/Promise` conditionally `Sendable`

dnadoba:
and `.get()` only available if `Value` is `Sendable`:

```swift
extension EventLoopPromise: Sendable where Value: Sendable { }
extension EventLoopFuture: Sendable where Value: Sendable { }
extension EventLoopFuture where Value: Sendable {
    func get() async throws { ... }
}
```
Also fix warnings because of these changes.

Normally this is a source-breaking change. However, `Sendable` is special and it looks like it may be a API stable change.

### DIFF
--- a/docker/docker-compose.2004.57.yaml
+++ b/docker/docker-compose.2004.57.yaml
@@ -30,7 +30,7 @@ services:
       - MAX_ALLOCS_ALLOWED_1000_getHandlers_sync=35
       - MAX_ALLOCS_ALLOWED_1000_reqs_1_conn=30400
       - MAX_ALLOCS_ALLOWED_1000_tcpbootstraps=4050
-      - MAX_ALLOCS_ALLOWED_1000_tcpconnections=154050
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -98,7 +98,7 @@ extension Channel {
     /// Get the value of `option` for this `Channel`.
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
     @inlinable
-    public func getOption<Option: ChannelOption>(_ option: Option) async throws -> Option.Value {
+    public func getOption<Option: ChannelOption>(_ option: Option) async throws -> Option.Value where Option.Value: Sendable {
```
dnadoba, line 101: ChannelOption.Value should be `Sendable` but currently we can't do so in source stable way #2108

PR #2115, opened 2022-05-16
Introduce internal `UnsafeTransfer` type

dnadoba:
### Motivation:
Soon we will need to silence some `Sendable` warnings. [The recommended way is to use a wrapper type that is unconditionally `Sendable`.](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#adaptor-types-for-legacy-codebases)
### Modifications:
- add internal `UnsafeTransfer` type

### Result:

We can silence `Sendable` warnings and errors by using `UnsafeTransfer` to pass non-`Sendable` values around.

### DIFF
--- a/Sources/NIOCore/NIOSendable.swift
+++ b/Sources/NIOCore/NIOSendable.swift
@@ -23,3 +23,19 @@ public typealias NIOSendable = Any
 #else
 public protocol NIOPreconcurrencySendable {}
 #endif
+
+#if swift(>=5.5) && canImport(_Concurrency)
+@usableFromInline
+struct UnsafeTransfer<Wrapped> : @unchecked Sendable {
+    @usableFromInline
+    var wrappedValue: Wrapped
+    
+    @inlinable
+    init(_ wrappedValue: Wrapped) {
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/NIOSendable.swift
```diff
@@ -23,3 +23,19 @@ public typealias NIOSendable = Any
 #else
 public protocol NIOPreconcurrencySendable {}
 #endif
+
+#if swift(>=5.5) && canImport(_Concurrency)
+@usableFromInline
+struct UnsafeTransfer<Wrapped> : @unchecked Sendable {
```
glbrntt, line 29: nit:
```suggestion
struct UnsafeTransfer<Wrapped>: @unchecked Sendable {
```

glbrntt: Could you also add a small piece of documentation to the `struct` explaining what it's for?

glbrntt (commented):
Two tiny nits (sorry!)

glbrntt (approved):

PR #2080, opened 2022-04-19
Drop support for Swift 5.2 and 5.3

fabianfett:
As outlined in a [Swift forums post in November ’21](https://forums.swift.org/t/swiftnio-swift-version-support/53232), SwiftNIO will only support the latest non-patch Swift release and the 2 immediately prior non-patch versions.

In this commit we drop support for Swift 5.2 and 5.3. We update CI for Swift 5.4 to run on bionic instead of focal to ensure that we still test bionic.

PR #2117, opened 2022-05-16
NIOPerformanceTester: Add DeadlineNowBenchmark for NIODeadline.now()

simonjbeaumont:
### Motivation:

`NIODeadline.now()` is a function that is called many times in typical workloads. It is currently implemented in terms of `DispatchTime.now()`, which is provided by libdispatch, on all platforms. The underlying implementation has a lot of branches to offer a more flexible API than we make use of. There might be a chance to improve the performance by dropping down to a lower-level implementation within NIO itself but we need a baseline measurement of performance to assess any future proposed improvements.

### Modifications:

Add performance benchmark for `NIODeadline.now()`.

### Result:

`NIOPerformanceTester` has benchmark test for `NIODeadline.now()`.

### DIFF
--- a/Sources/NIOPerformanceTester/DeadlineNowBenchmark.swift
+++ b/Sources/NIOPerformanceTester/DeadlineNowBenchmark.swift
@@ -32,8 +32,7 @@ final class DeadlineNowBenchmark: Benchmark {
         for _ in 0..<self.iterations {
             let now = NIODeadline.now().uptimeNanoseconds
             counter &+= now
-
         }
-        return 1
+        return Int(truncatingIfNeeded: counter)
     }
 }
--- a/Sources/NIOPerformanceTester/DeadlineNowBenchmark.swift
+++ b/Sources/NIOPerformanceTester/DeadlineNowBenchmark.swift
@@ -1,0 +1,39 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+final class DeadlineNowBenchmark: Benchmark {
+    private let iterations: Int
+
+    init(iterations: Int) {
+        self.iterations = iterations
+    }
+
+    func setUp() throws {
+    }
+
+    func tearDown() {
+    }
+
+    func run() -> Int {
+        var counter: UInt64 = 0
+        for _ in 0..<self.iterations {
+            let now = NIODeadline.now().uptimeNanoseconds
+            counter &+= now
+
+        }
+        return 1
+    }
+}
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
LGTM. I'm interested to see what perf gains can be made from looking into this!

glbrntt:

simonjbeaumont:
@glbrntt @Lukasa the CI is failing with:

```
13:20:25 .FileNotFoundError: [Errno 2] No such file or directory: './docker/docker-compose.2004.54.yaml'
```

I noticed that this was removed recently, in #2080 (although, maybe not deliberately since the commit message is only about dropping 5.2 and 5.3).

I guess we either need this Docker Compose file for 5.4 back, or we need to update the CI job to use a newer one?

glbrntt:
```
13:20:24 + docker-compose -f docker/docker-compose.yaml -f docker/docker-compose.2004.54.yaml -p swift-nio2-performance-prb build --pull
13:20:25 .FileNotFoundError: [Errno 2] No such file or directory: './docker/docker-compose.2004.54.yaml'
13:20:25 Build step 'Execute shell' marked build as failure
```

CI is expecting `docker-compose.2004.54.yaml` but we recently moved Swift 5.4 CI to 18.04 (https://github.com/apple/swift-nio/pull/2080) so that file no longer exists which is why perf tests now fail.

I suspect we actually want to use Swift 5.6.

simonjbeaumont:
@glbrntt I'll open a PR for that separately.

simonjbeaumont:
> I suspect we actually want to use Swift 5.6.

Who can update the CI job to make use of 5.6 one?

In the meantime should we reinstate the 5.4 Bionic?

Lukasa:
@yim-lee can you update the performance jobs to use 5.6 instead of 5.4? Thankyou!

Sources/NIOPerformanceTester/DeadlineNowBenchmark.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    func run() -> Int {
+        var counter: UInt64 = 0
+        for _ in 0..<self.iterations {
+            let now = NIODeadline.now().uptimeNanoseconds
+            counter &+= now
+
+        }
+        return 1
```
Lukasa, line 37: Can I suggest we return `counter` here? That way the benchmarks will be less inclined to kill the code.

simonjbeaumont: Sure, I think I saw another test returning `1` here. Do I understand correctly that this return value has no meaningful use, but is just used to ensure that code doesn't get optimised away?

simonjbeaumont: Addressed in https://github.com/apple/swift-nio/pull/2117/commits/edffb7d202d004c86c900ebe1bc85c7930482857.

Lukasa: @simonjbeaumont Yes: the result is thrown away, but the call site doesn't know that.

simonjbeaumont:

simonjbeaumont:
> significant differences found

These are a result of the move to Swift 5.6 as the image for the perf test presumably since nothing in this PR will affect those benchmarks.

Lukasa (approved):

PR #2119, opened 2022-05-16
wip: Use clock_gettime for NIODeadline.now()

simonjbeaumont:
### Motivation:

`NIODeadline.now()` is a function that is called many times in typical workloads. It is currently implemented in terms of `DispatchTime.now()`, which is provided by `libdispatch`, on all platforms. The underlying implementation has a lot of branches to offer a more flexible API than we make use of and also involves a call to another library. The way in which NIO calls this function as part of `NIODeadline.now()` _always_ results in a simple call to `clock_gettime(3)`. Calling this directly might yield a performance benefit.

### Modifications:

* Replace call to `DispatchTime.now()` with call to `clock_gettime` in `NIODeadline.now()`.

### Result:

* About a 10% speedup (on Linux only for now).

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -14,7 +14,9 @@
 
 import NIOConcurrencyHelpers
 import Dispatch
+#if os(Linux)
 import CNIOLinux
+#endif // os(Linux)
 
 /// Returned once a task was scheduled on the `EventLoop` for later execution.
 ///
@@ -480,18 +482,16 @@ public struct NIODeadline: Equatable, Hashable, NIOSendable {
     }
 
 
-//#if os(Linux)
     @inline(__always)
     private static func timeNow() -> UInt64 {
-        if #available(macOS 10.12, *) {
-            var ts = timespec()
-            clock_gettime(CLOCK_MONOTONIC, &ts)
-            return UInt64(ts.tv_sec) &* 1_000_000_000 &+ UInt64(ts.tv_nsec)
-        } else {
-            return DispatchTime.now().uptimeNanoseconds
-        }
+#if os(Linux)
+        var ts = timespec()
+        clock_gettime(CLOCK_MONOTONIC, &ts)
+        return UInt64(ts.tv_sec) &* 1_000_000_000 &+ UInt64(ts.tv_nsec)
+#else // os(Linux)
+        return DispatchTime.now().uptimeNanoseconds
+#endif // os(Linux)
     }
-//#endif
 
     public static func now() -> NIODeadline {
         return NIODeadline.uptimeNanoseconds(timeNow())
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -482,11 +482,24 @@ public struct NIODeadline: Equatable, Hashable, NIOSendable {
     }
 
 
+    /// Getting the time is a very common operation so it warrants optimization.
+    ///
+    /// Prior to this function, NIO relied on `DispatchTime.now()`, on all platforms. In addition to
+    /// the overhead of making a library call, the underlying implementation has a lot of branches
+    /// because `libdispatch` supports many more usecases than we are making use of here.
+    ///
+    /// On Linux, `DispachTime.now()` _always_ results in a simple call to `clock_gettime(3)` and so
+    /// we make that call here, directly from NIO.
+    ///
+    /// - TODO: Investigate optimizing the call to `DispatchTime.now()` away on other platforms too.
     @inline(__always)
     private static func timeNow() -> UInt64 {
 #if os(Linux)
         var ts = timespec()
         clock_gettime(CLOCK_MONOTONIC, &ts)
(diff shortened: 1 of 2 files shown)


simonjbeaumont:

simonjbeaumont:
Well, it seems to have improved the `deadline_now_1M_times` by 12% and some other performance gains generally throughout the suite of benchmarks. However, it has resulted in a pretty significant regression in some of the locking benchmarks, but not all:

name | current | previous | winner | diff
-- | -- | -- | -- | --
lock_1_thread_1M_ops | 0.015752419 | 0.015964714 | current | -1%
lock_2_threads_1M_ops | 0.079739508 | 0.053677755 | previous | 48%
lock_4_threads_1M_ops | 0.086140819 | 0.053396918 | previous | 61%
lock_8_threads_1M_ops | 0.073295558 | 0.082103087 | current | -10%

Lukasa:
Hmm, I'd be interested in seeing those numbers reproduced, because I think those numbers have been fluctuating around quite a lot. In particular, those benchmarks shouldn't be touching this time code at all.

simonjbeaumont:
> Hmm, I'd be interested in seeing those numbers reproduced, because I think those numbers have been fluctuating around quite a lot.

I'll rerun and see how they get on. 

> In particular, those benchmarks shouldn't be touching this time code at all.

Right, it seems for `lock_1_thread_1M_ops` it only gets called about 100 times so shouldn't be the problem, I guess.

simonjbeaumont:
Wow, they really are noisy...

name | current | previous | winner | diff
-- | -- | -- | -- | --
lock_1_thread_1M_ops | 0.015758304 | 0.019061978 | current | -17%
lock_2_threads_1M_ops | 0.06912248 | 0.066318807 | previous | 4%
lock_4_threads_1M_ops | 0.081254568 | 0.087642399 | current | -7%
lock_8_threads_1M_ops | 0.057924588 | 0.022824741 | previous | 153%

I might open an empty PR to prove out that these are noisy independent of this PR.

simonjbeaumont:
So these locking benchmarks are very very noisy. Here's an empty PR to validate that claim: https://github.com/apple/swift-nio/pull/2120

simonjbeaumont:

Sources/NIOCore/EventLoop.swift
```diff
@@ -478,8 +479,22 @@ public struct NIODeadline: Equatable, Hashable, NIOSendable {
         self._uptimeNanoseconds = nanoseconds
     }
 
+
+//#if os(Linux)
```
simonjbeaumont, line 483: @Lukasa I had it like this for testing but from what I read in libdispatch this won't be appropriate on Darwin: https://github.com/apple/swift-corelibs-libdispatch/blob/34f383d34450d47dd5bdfdf675fcdaa0d0ec8031/src/shims/time.h#L126-L159.

How should we proceed here? Should we (a) just do this Linux only for now and continue to just call `DispatchTime.now()` on Darwin, or (b) do something different on Darwin, e.g. `mach_absolute_time()`?

Sources/NIOCore/EventLoop.swift
```diff
@@ -14,6 +14,9 @@
 
 import NIOConcurrencyHelpers
 import Dispatch
+#if os(Linux)
```
Lukasa, line 17: The `#if` guard should be unnecessary, `CNIOLinux` is just empty on Darwin.

Sources/NIOCore/EventLoop.swift
```diff
@@ -478,8 +481,20 @@ public struct NIODeadline: Equatable, Hashable, NIOSendable {
         self._uptimeNanoseconds = nanoseconds
     }
 
+
+    @inline(__always)
+    private static func timeNow() -> UInt64 {
+#if os(Linux)
+        var ts = timespec()
+        clock_gettime(CLOCK_MONOTONIC, &ts)
+        return UInt64(ts.tv_sec) &* 1_000_000_000 &+ UInt64(ts.tv_nsec)
```
Lukasa, line 490: We need comments added to explain why it's safe for this math to be unchecked. In this case this boils down to: "UInt64.max nanoseconds is more than 580 years, and the odds that this code will still be running 530 years from now is very, very low, so as a practical matter this will never overflow."

Lukasa (commented):
Generally looks good, a few quick notes.

Sources/NIOCore/EventLoop.swift
simonjbeaumont: Happy to remove, but looking around the repo all existing imports of `CNIOLinux` _are_ guarded:

```
$  git grep -B2 "import CNIOLinux"
```

Sources/NIOCore/EventLoop.swift
```diff
@@ -478,8 +481,20 @@ public struct NIODeadline: Equatable, Hashable, NIOSendable {
         self._uptimeNanoseconds = nanoseconds
     }
 
+
+    @inline(__always)
+    private static func timeNow() -> UInt64 {
```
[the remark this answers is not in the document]
simonjbeaumont: Added in af2e007e2.

Lukasa (approved):

PR #2132, opened 2022-05-30
NIOCore: replace `mode_t` with `CInt`

compnerd:
Windows does not have a `mode_t` type alias, instead using the
de-sugared `CInt` type.  De-sugar the instances to permit building on
Windows.

### DIFF
--- a/Sources/NIOCore/FileHandle.swift
+++ b/Sources/NIOCore/FileHandle.swift
@@ -19,6 +19,12 @@ import Darwin
 import Glibc
 #endif
 
+#if os(Windows)
+public typealias NIOPOSIXFileMode = CInt
+#else
+public typealias NIOPOSIXFileMode = mode_t
+#endif
+
 /// A `NIOFileHandle` is a handle to an open file.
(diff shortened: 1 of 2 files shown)


compnerd:
Sure a type alias would work.  Any thoughts on a name?

Sources/NIOCore/FileHandle.swift
```diff
@@ -19,6 +19,12 @@ import Darwin
 import Glibc
 #endif
 
+#if os(Windows)
+public typealias POSIXFileMode = CInt
+#else
+public typealias POSIXFileMode = mode_t
```
Lukasa, line 25: Nit: We need to prefix this with `NIO`, otherwise this is a good name.

compnerd: Yeah, that is a good idea ... should've realized that.

Lukasa (changes requested):
One nit and then we're good.

Lukasa (approved):

PR #2135, opened 2022-05-31
Remove `@preconcurrency` from internal `EventLoopFuture` methods

dnadoba:
### Motivation:
We want to remove `@preconcurrency` from our internal methods and fix the resulting sendable warnings/errors.

### Modifications:

- Remove  `@preconcurrency` from internal methods and remove code duplication that is no longer necessary.  
- Introduce a new `UnsafeMutableTransferBox` that can be used to make non-`Sendable` values `Sendable` and mutable. It can be used to capture local mutable values in a `@Sendable` closure and mutate them from within the closure.
- Use `UnsafeMutableTransferBox` where we currently capture mutable variables from `@Sendable` closures.

### Result:

All internal only uses of `@preconcurrency` are now gone. Less code duplication.


dnadoba:

dnadoba:

Lukasa (approved):
Nice, LGTM.

dnadoba:
If we remove `@preconcurency` from `EventLoopFuture._whenComplete` we have a memory regression in our `test_1000_udpconnections.swift` allocation test of 1 allocation per iteration. I have added `@preconcurency` for this method again with a `TODO` and reported the issue here: https://github.com/apple/swift/issues/59249
We can still decided later if we accept this regression or not and remove `@preconcurency` in a separate PR.

PR #2137, opened 2022-05-31
wip: Reduce ARC traffic with inout wrapper

simonjbeaumont:

### Motivation:

### Modifications:

### Result:

### DIFF
--- a/Sources/NIOCore/ChannelPipeline.swift
+++ b/Sources/NIOCore/ChannelPipeline.swift
@@ -1468,6 +1468,15 @@ public final class ChannelHandlerContext: ChannelInvoker {
         precondition(self.inboundHandler != nil || self.outboundHandler != nil, "ChannelHandlers need to either be inbound or outbound")
     }
 
+    private func executeIfNotNil(
+        _ context: inout ChannelHandlerContext?,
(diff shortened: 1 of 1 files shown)


simonjbeaumont:

simonjbeaumont:

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1503,7 +1512,7 @@ public final class ChannelHandlerContext: ChannelInvoker {
 
     /// Signal to the next `ChannelHandler` that a read burst has finished.
     public func fireChannelReadComplete() {
-        self.next?.invokeChannelReadComplete()
+        executeIfNotNil(&self.next) { $0.invokeChannelReadComplete() }
```
Lukasa, line 1515: Sadly I don't think we can make this optimisation: it'll violate the law of exclusivity. It's possible for us to synchronously reconfigure the pipeline, which will write to this property.

simonjbeaumont: Yeah, I was looking at something referenced in an old issue which suggested this is something we could try although I can see that when I apply this particular variant across some tests they explode with access issues.

I'm studying the other issue though and think I just overlooked something and there may be something we can do here.

PR #2145, opened 2022-06-01
Adopt `Sendable` in `ConvenienceOptionSupport.swift`

dnadoba:

### DIFF
--- a/Sources/NIOCore/ConvenienceOptionSupport.swift
+++ b/Sources/NIOCore/ConvenienceOptionSupport.swift
@@ -53,6 +53,10 @@ extension ChannelOptions.Types {
     }
 }
 
+#if swift(>=5.5) && canImport(_Concurrency)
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/ConvenienceOptionSupport.swift
```diff
@@ -80,7 +84,7 @@ extension ChannelOptions.Types.ConvenienceOptionValue where ValueType == () {
 // MARK: TCP - data
 extension ChannelOptions {
     /// A TCP channel option which can be applied to a bootstrap using convenience notation.
-    public struct TCPConvenienceOption: Hashable {
+    public struct TCPConvenienceOption: Hashable, NIOSendable {
```
glbrntt, line 87: Does the wrapepd `ConvenienceOption` also need to be `NIOSendable`?

Lukasa (approved):

PR #2150, opened 2022-06-01
NIOCore: extract and modify POSIX flag constants

compnerd:
Windows does not have the same set of POSIX flag constants.  In
particular, there is no equivalent to group permissions as the ACL model
on Windows is far richer and requires the full specification of the
DACLs.  Additionally, isolate `O_CLOEXEC` to aid Windows which does not
have this flag.


Sources/NIOCore/FileHandle.swift
```diff
@@ -128,11 +128,17 @@ extension NIOFileHandle {
 
         public static let `default` = Flags(posixMode: 0, posixFlags: 0)
 
+#if os(Windows)
+        public static let rw_r__r = _S_IREAD | _S_IWRITE
+#else
+        public static let rw_r__r = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH
```
Lukasa, line 134: This should not need to be public. Additionally, it'd probably be better to give this a useful name, if only `defaultPermissions`.

compnerd: I can rename it, I don't have a problem with that.  Making it non-public though is a problem.

```
static property 'rw_r__r' is private and cannot be referenced from a default argument value
```

The use as a default value requires that the value be public as it is going to be inlined into the inlineable default value generator function.

Lukasa: It doesn't have to be `private`, `internal` is fine. And normally the default value generator function is only inlinable if the usage site is, which in this case it is not.

compnerd: `internal` demonstrates the same thing.

Lukasa: Huh, you're right. Very weird.

Lukasa (approved):

PR #2155, opened 2022-06-02
NIOCore: adjust system call helpers for Windows

compnerd:
Windows does not provide a 3 parameter `open` and needs to be passed all
the parameters as `_open` is a variadic function (as per the
specification).  This provides a Windows specific path for `open` and
`read` as on Darwin and Linux, `read` uses a non-standard return type
(`ssize_t`).  As it happens, the `size` parameter on Windows also uses
`unsigned int` rather than `size_t` which would break on different
bitnesses.


Sources/NIOCore/SystemCallHelpers.swift
```diff
@@ -121,12 +128,24 @@ enum SystemCalls {
         }
     }
 
+#if os(Windows)
+    @inline(never)
+    static func open(file filename: UnsafePointer<CChar>, oFlag oflag: CInt,
+                     mode pmode: CInt) throws -> CInt {
```
Lukasa, line 134: This doesn't need a new signature, just a new implementation. Can we only replace the implementation?

Lukasa (approved):

PR #2156, opened 2022-06-02
NIOCore: add explicit type cast for Windows compatibility

compnerd:
Windows does not have a `socklen_t` type, instead using `CInt` for the
value.  This adds an explicit type cast to repair the build on Windows.


Lukasa (approved):

Lukasa:

PR #2161, opened 2022-06-03
Implement the _failEarlyRangeCheck methods as no-ops

stepan-ulyanin:
Implements all versions of the `_failEarlyRangeCheck(_:bounds:)` Collection methods as no-ops.

### Motivation:
https://github.com/apple/swift-nio/issues/1827 specifies the missing Collection customization points that we would like to customize.

It seems that the `_failEarlyRangeCheck(_:bounds:)` should be implemented as no-ops.

### Modifications:

Adds three methods to the `CircularBuffer` conforming to `Collection`

### Result:

The default implementations of the `_failEarlyRangeCheck` methods are going to be customized to be no-ops in the `CircularBuffer` struct.

### DIFF
--- a/Sources/NIOCore/CircularBuffer.swift
+++ b/Sources/NIOCore/CircularBuffer.swift
@@ -251,6 +251,10 @@ extension CircularBuffer: Collection, MutableCollection {
 
         return (self[self.endIndex..<self.endIndex].makeIterator(), self.count)
     }
+    
+    public func _failEarlyRangeCheck(_ index: Index, bounds: Range<Index>) {}
+    public func _failEarlyRangeCheck(_ index: Index, bounds: ClosedRange<Index>) {}
+    public func _failEarlyRangeCheck(_ range: Range<Index>, bounds: Range<Index>) {}
 }
 
 // MARK: RandomAccessCollection implementation


Sources/NIOCore/CircularBuffer.swift
```diff
@@ -251,6 +251,10 @@ extension CircularBuffer: Collection, MutableCollection {
 
         return (self[self.endIndex..<self.endIndex].makeIterator(), self.count)
     }
+    
+    public func _failEarlyRangeCheck(_ index: Index, bounds: Range<Index>) {}
+    public func _failEarlyRangeCheck(_ index: Index, bounds: ClosedRange<Index>) {}
+    public func _failEarlyRangeCheck(_ range: Range<Index>, bounds: Range<Index>) {}
```
Lukasa, line 257: Can we mark these all `@inlinable`, and also note that these have empty bodies for performance reasons?

Lukasa: Fantastic, now can we add a comment above these saying that they're empty for performance reasons?

stepan-ulyanin: @Lukasa ok I added the comment in [135f5e7](https://github.com/apple/swift-nio/pull/2161/commits/135f5e744c99ce46070b45235014a6f89dbf2971).

Just for my understanding, now as these are implemented as no-ops, **there is no range check on index range access at runtime** (so we don't have these extra calls at runtime)?

The default behavior was:
> The default implementation for random access indices performs a range check.

If so, where were these methods called from? I was under the impression that these are supposed to be called by us not Swift?

Lukasa: No, these are called by Swift. Swift calls them from algorithm implementations on Swift. We're not supposed to call them.

Both `CircularBuffer` and `Array` do their own checking so we don't need to have Swift do it here.

Lukasa (approved):

Sources/NIOCore/CircularBuffer.swift
stepan-ulyanin: > No, these are called by Swift. Swift calls them from algorithm implementations on Swift. We're not supposed to call them.
> 
> Both `CircularBuffer` and `Array` do their own checking so we don't need to have Swift do it here.

Ah, ok that is what I was wondering about, thanks!

PR #2164, opened 2022-06-06
NIOCore: adjust the BSDSocketAPI for Windows

compnerd:
Windows does not provide a definition for `socklen_t`.  Create an
internal typealias for it to provide source stability to the internal
interfaces.


compnerd:
I don't see how this could possibly change the behaviour on non-Windows.  Could be a flake?

glbrntt:
The failure was one of the integration tests (`test_01_syscall_wrapper_fast.sh`) which often fails when the worker is overloaded.

Lukasa (approved):

PR #2165, opened 2022-06-06
NIOCore: duplicate the NIODevice implementation on Windows

compnerd:
Windows does not have an `ifaddr` interface, and instead we need to
construct the interface from the adapter and logical interface.
Replicate this for the construction of the deprecated type just for
repairing the build.

### DIFF
--- a/Sources/NIOCore/Interfaces.swift
+++ b/Sources/NIOCore/Interfaces.swift
@@ -87,6 +87,63 @@ public final class NIONetworkInterface {
     /// The index of the interface, as provided by `if_nametoindex`.
     public let interfaceIndex: Int
 
+#if os(Windows)
+    internal init?(_ pAdapter: UnsafeMutablePointer<IP_ADAPTER_ADDRESSES>,
+                   _ pAddress: UnsafeMutablePointer<IP_ADAPTER_UNICAST_ADDRESS>) {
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/Interfaces.swift
```diff
@@ -87,6 +87,63 @@ public final class NIONetworkInterface {
...
+            var mask: sockaddr_in = sockaddr_in()
+            mask.sin_family = ADDRESS_FAMILY(AF_INET)
+            _ = buffer.withUnsafeMutableBufferPointer {
+                try! NIOBSDSocket.inet_ntop(addressFamily: .inet,
+                                            addressBytes: &mask,
+                                            addressDescription: $0.baseAddress!,
+                                            addressDescriptionLength: socklen_t(INET_ADDRSTRLEN))
+            }
+            return SocketAddress(mask)
+        }
```
Lukasa, line 113: Ah yeah, I remember this from last time: these functions don't work as implemented. I'll merge this regardless, and follow-up with correct implementations.

Lukasa (approved):

Sources/NIOCore/Interfaces.swift
compnerd: Thanks!

PR #2168, opened 2022-06-06
NIOPosix: add additional imports and typealiases on Windows

compnerd:
Extend more of the internal interfaces for the BSD API coverage on
Windows to get the same API surface when possible.


Sources/NIOPosix/BSDSocketAPIWindows.swift
```diff
@@ -102,13 +105,14 @@ import struct WinSDK.SOCKADDR_STORAGE
...
 
+internal typealias in_addr = IN_ADDR
+internal typealias in_port_t = USHORT
+internal typealias sa_family_t = ADDRESS_FAMILY
 internal typealias sockaddr_in = SOCKADDR_IN
 internal typealias sockaddr_in6 = SOCKADDR_IN6
 internal typealias sockaddr_un = SOCKADDR_UN
 internal typealias sockaddr_storage = SOCKADDR_STORAGE
 
-public typealias linger = LINGER
```
glbrntt, line 110: Presumably this was unused?

compnerd: Yes, it was unused.

PR #2169, opened 2022-06-07
Add support for translating subnet prefixes to masks

Lukasa:
Motivation:

On Windows we have a need to transform a subnet prefix into a subnet
mask when enumerating devices. This is a straightforward transformation
that we can easily do in pure-Swift, so let's do that. Additionally, the
functions have small enough domains that we can just exhaustively test
them.

Modifications:

- Added functions to create SocketAddress objects from subnet prefixes.
- Added exhaustive tests for those functions.
- Use these functions in Windows code.

Result:

Correct transformations of prefixes into masks on Windows.


Sources/NIOCore/SocketAddresses.swift
```diff
@@ -394,6 +394,47 @@ public enum SocketAddress: CustomStringConvertible, NIOSendable {
...
 
+    /// Creates a new `SocketAddress` corresponding to the netmask for a subnet prefix.
+    ///
+    /// As an example, consider the subnet "127.0.0.1/8". The "subnet prefix" is "8", and the corresponding netmask is "255.0.0.0".
+    /// This initializer will produce a `SocketAddress` that contains "255.0.0.0".
+    ///
+    /// - parameters:
+    ///     - prefix: The prefix of the subnet.
+    /// - returns: A `SocketAddress` containing the associated netmask.
+    internal init(ipv4MaskForPrefix prefix: Int) {
```
glbrntt, line 405: Do we want to `precondition`/`assert` that `prefix` is valid?

glbrntt (approved):
One question but looks good otherwise.

Sources/NIOCore/SocketAddresses.swift
Lukasa: Yeah good question. I wasn't honestly sure. Given that these are `internal` I considered whether we should just allow them to produce gibberish, but maybe we should precondition instead.

Lukasa: Updated.

glbrntt (approved):

PR #2170, opened 2022-06-07
NIOPosix: reduce imports for Windows

compnerd:
Reduce unnecessary noise in the files after previous refactoring
shuffled the use of the imported interfaces.  We can centralise most of
the imports and internal typealiases which avoids cluttering the shared
interfaces with compatibility handling for Windows.


Lukasa (approved):

Lukasa:
Hit flaky test, re-running

PR #2177, opened 2022-06-08
NIOPosix: use internal enumeration for GAI resolver

compnerd:
Replace the use of raw constants with the internal enumeration.  This
ensures that the constant names are uniform and don't leak structural
information from the underlying information.


Lukasa:
There seem to be some tests that need fixing up. Are you equipped to tackle them, or would you like me to?

compnerd:
The help would be great!

Lukasa (approved):
Nice change, thank you!

PR #2178, opened 2022-06-08
NIOPosix: further adopt `NIOBSDSocket.OptionLevel`

compnerd:
Use `NIOBSDSocket.OptionLevel` for the option parameter to abstract over
the underlying system implementation.


Lukasa:
There are some test failures here too.

PR #2179, opened 2022-06-08
NIOPosix: port file validation to Windows

compnerd:
Use Win32 APIs to properly validate if a file is a pipe on Windows.
This enables providing the same semantics without leaking additional
Windows specifics.

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -13,6 +13,18 @@
 //===----------------------------------------------------------------------===//
 import NIOCore
 
+#if os(Windows)
+import ucrt
+
+import func WinSDK.GetFileType
+
+import let WinSDK.FILE_TYPE_PIPE
+import let WinSDK.INVALID_HANDLE_VALUE
+
+import struct WinSDK.DWORD
+import struct WinSDK.HANDLE
+#endif
+
 /// The type of all `channelInitializer` callbacks.
 internal typealias ChannelInitializerCallback = (Channel) -> EventLoopFuture<Void>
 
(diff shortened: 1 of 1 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1070,6 +1082,20 @@ public final class NIOPipeBootstrap {
...
     private func validateFileDescriptorIsNotAFile(_ descriptor: CInt) throws {
+#if os(Windows)
+        // NOTE: this is a *non-owning* handle, do *NOT* call `CloseHandle`
+        let hFile: HANDLE = HANDLE(bitPattern: _get_osfhandle(descriptor))!
+        if hFile == INVALID_HANDLE_VALUE {
+            throw IOError(errnoCode: EBADF, reason: "_get_osfhandle")
+        }
+
+        switch GetFileType(hFile) {
+        case DWORD(FILE_TYPE_PIPE):
```
Lukasa, line 1093: This is not quite the semantic of the non-Windows code, which just checks whether things are not a regular file, a directory, a symlink, or a block device. Sockets, character devices, and pipes all pass this check. Is it possible to replicate this logic?

compnerd: The sockets and file descriptors are completely separate domains, so its implicitly not a socket.

I suppose that we could add the check for the character devices and block devices, though, I'm worried that it is less than useful as pipes are completely different from the other file types (not to mention different from Unix-like platforms).

Lukasa: I'm ok leaving those checks out, but I think we need a code comment explaining _why_ the Windows code is written the way it is. Otherwise if it misbehaves in future I'll have no idea why the two branches do two (apparently) utterly different things.

Lukasa (approved):

PR #2191, opened 2022-06-13
Provide conversion APIs between TimeAmount and Swift.Duration

### LINKED ISSUE #2175: Provide Swift.Duration converters to NIO.TimeAmount
Swift 5.7 introduces time types, like `Swift.Duration`.

That type can be converted, although potentially lossy to `NIO.TimeAmount`.

It is represented as `seconds + attoseconds` so conversion to NIO's `nanoseconds: Int64` based value may be lossy, but in reality the conversion for "typical" uses is totally fine... Normally people just schedule a timeout of like 3 seconds or so after all 😉 

Let's provide some converters, even if failable.

simonjbeaumont:
### Motivation:

[SE-0329: Clock, Instant, and Duration](https://github.com/apple/swift-evolution/blob/main/proposals/0329-clock-instant-duration.md) now has a draft implementation in Swift 5.7. It would be ergonomic to provide APIs to convert `Swift.Duration` to and from `NIO.TimeAmount`.

### Modifications:

Add `TimeAmount.init(_: Duration)` and `Duration.init(_: TimeAmount)` extensions. Because `Duration` has higher precision and range, the conversion to `TimeAmount` is truncated and clamped to the maximum expressible nanosecond value, `Int64.max`.

### Result:

Resolves #2175.


ktoso:
This would resolve https://github.com/apple/swift-nio/issues/2175

Sources/NIOCore/EventLoop.swift
```diff
@@ -523,6 +523,33 @@ extension TimeAmount: AdditiveArithmetic {
...
 
+#if swift(>=5.7)
+extension TimeAmount {
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    /// Creates a new `TimeAmount` for the given ``Swift.Duration``.
+    ///
+    /// - parameters:
+    ///     - duration: ``Swift.Duration``.
+    /// - returns: the `TimeAmount` for the given amount, truncated to whole nanoseconds.
+    public static func from(_ duration: Swift.Duration) -> TimeAmount {
```
fabianfett, line 534: I think the preferred Swift way to do this is via initializers.

Sources/NIOCore/EventLoop.swift
```diff
@@ -523,6 +523,33 @@ extension TimeAmount: AdditiveArithmetic {
...
+
+@available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+extension Swift.Duration {
+    /// Construct a `Duration` given a number of nanoseconds represented as a `TimeAmount`.
+    ///
+    ///       let t: TimeAmount = .seconds(1)
+    ///       let d: Duration = .from(t)
+    ///
+    /// - Returns: A `Duration` representing a given number of nanoseconds.
+    public static func from(_ timeAmount: TimeAmount) -> Duration {
```
fabianfett, line 547: Same as above. I think this should be an init.

Tests/NIOCoreTests/TimeAmountTests.swift
```diff
@@ -43,4 +43,52 @@ class TimeAmountTests: XCTestCase {
         lhs -= rhs
         XCTAssertEqual(lhs, .nanoseconds(0))
     }
+
+#if swift(>=5.7)
```
fabianfett, line 47: The swift check has to live inside the function, since otherwise old Swift version don't compile.

fabianfett (commented):
Great addition!

Sources/NIOCore/EventLoop.swift
simonjbeaumont: Yes, I was back-and-forth on the spelling and wasn't happy enough to make this a non-draft.

I think I ended up with the static because I could use the existing `static Duration.nanoseconds(_:) -> Duration` as a shorthand but you're right ofc. I see that `Duration` has initialisers from `timeval` and `timespec` already. The best fit will probably be another one from `NIO.TimeAmount`. The many `1_000_000_000` were just making my eyes hurt :)

Tests/NIOCoreTests/TimeAmountTests.swift
simonjbeaumont: Is that because of the unguarded addition to `TimeAmountTests+XCTest.swift`. I'm happy to move the guard in the function but wanted to check if it would instead be cleaner to instead add another guard in the `+XCTest.swift` to mirror this one. That way the test won't appear at all on older versions, cf. being an empty test. WDYT?

glbrntt: > Is that because of the unguarded addition to `TimeAmountTests+XCTest.swift`. 

Exactly.

> I'm happy to move the guard in the function but wanted to check if it would instead be cleaner to instead add another guard in the `+XCTest.swift` to mirror this one. That way the test won't appear at all on older versions, cf. being an empty test. WDYT?

I don't think it matters all that much which way it's done. I suspect updating the script to do this is probably more hassle than it's worth though.

Sources/NIOCore/EventLoop.swift
```diff
@@ -523,6 +523,50 @@ extension TimeAmount: AdditiveArithmetic {
...
+}
+
+@available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+internal extension Swift.Duration {
+    /// The duration represented as nanoseconds, clamped to maximum expressible value.
+    var nanosecondsClamped: Int64 {
+        let components = self.components
+
+        let secondsComponentNanos = components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
+        let attosCompononentNanos = components.attoseconds.dividedReportingOverflow(by: 1_000_000_000)
```
Lukasa, line 554: I don't think we need `dividedReportingOverflow` here: we have a statically known quotient so we know this cannot overflow.

Lukasa (changes requested):
I've left a brief note in the implementation: additionally, I agree with @fabianfett's suggestion that this lives in a `_NIOBeta57` target and product for a while, as there is a risk of things breaking.

glbrntt:
> I've left a brief note in the implementation: additionally, I agree with @fabianfett's suggestion that this lives in a `_NIOBeta57` target and product for a while, as there is a risk of things breaking.

Why a 5.7 specific module rather than a something more general purpose?

simonjbeaumont:
> > I've left a brief note in the implementation: additionally, I agree with @fabianfett's suggestion that this lives in a `_NIOBeta57` target and product for a while, as there is a risk of things breaking.
> 
> Why a 5.7 specific module rather than a something more general purpose?

For now I created a `_NIOBeta57` module, but we can rename it to whatever.

Lukasa:
Because the module would only be temporary, we'd remove it when the 5.7 beta lands. Because it's underscored we reserve the right to break it by removing it (which we will). The reason it's not in, say, NIOCore is that the API surface isn't final: the Swift team absolutely may break it down the road, and we don't want users who are using our stable APIs to be broken by that.

glbrntt:
> Because the module would only be temporary, we'd remove it when the 5.7 beta lands. Because it's underscored we reserve the right to break it by removing it (which we will). The reason it's not in, say, NIOCore is that the API surface isn't final: the Swift team absolutely may break it down the road, and we don't want users who are using our stable APIs to be broken by that.

Ah I was more wondering why we'd have a temporary API breaking module rather than a `_NIOBeta` or similar that we can keep around the next time we want to add things ahead of a release.

Lukasa:
Oh, I see. We could do that I suppose. No strong preference really.

Package.swift
```diff
@@ -91,7 +93,7 @@ var targets: [PackageDescription.Target] = [
     .executableTarget(name: "NIOAsyncAwaitDemo",
             dependencies: ["NIOPosix", "NIOCore", "NIOHTTP1"]),
     .testTarget(name: "NIOCoreTests",
-                dependencies: ["NIOCore", "NIOEmbedded", "NIOFoundationCompat"]),
+                dependencies: ["NIOCore", "NIOEmbedded", "NIOFoundationCompat", "_NIOBeta57"]),
```
Lukasa, line 96: Can we move the tests to a specific test target?

Lukasa (changes requested):

Package.swift
simonjbeaumont: Done.

Lukasa (approved):

PR #2195, opened 2022-06-15
Adopt `Sendable` in `NIOEmbedded`

dnadoba:

`EmbeddedChannel` and `EmbeddedEventLoop` should not be `Sendable`. However, they inherit from `Channel` and `EventLoop` respectively. Therefore, they need to be `Sendable` and are already implicitly `Sendable`. Given these constraints, we can’t mark `EmbeddedChannel` and `EmbeddedEventLoop` as non-Sendable.


Lukasa (approved):

Lukasa:
The compile error on 5.5 is real and needs to be fixed.

PR #2196, opened 2022-06-15
NIOPosix: remove use of `setNonBlockingIO` on Windows pipes

compnerd:
On Windows anonymous pipes are not easy to make non-blocking.
Furthermore, doing so in any reasonably useful manner requires using
undocumented tricks which makes it vaguely but not entirely similar to
Unix semantics.  Prevent this operation on Windows for the time being.

### DIFF
--- a/Sources/NIOPosix/PipePair.swift
+++ b/Sources/NIOPosix/PipePair.swift
@@ -45,11 +45,16 @@ final class PipePair: SocketProtocol {
         self.inputFD = SelectableFileHandle(inputFD)
         self.outputFD = SelectableFileHandle(outputFD)
         try self.ignoreSIGPIPE()
+        // FIXME: there are underdocumented, ill-convcieved ways to emulate this
+        // vaugely, but, it might just not be worth it and be better to always
+        // just use overlapped IO on Winodws.
+#if !os(Windows)
         for fileHandle in [inputFD, outputFD] {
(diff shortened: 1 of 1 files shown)


Lukasa (commented):
If we can't make these non-blocking, should we remove all functionality that requires them from Windows entirely, root-and-branch? It should be easy enough to do by simply making the type entirely unavailable.

compnerd:
Well, pipes are fine, it just is that anonymous pipes don't have a way to be marked as async, though all IO on Windows is technically async (you can use IO Completion Ports, or overlapped IO).

Lukasa:
Yes, I'm familiar with IOCP. The problem here is that NIOPOSIX will not be using IOCP (it's intended to be the POSIX compat layer), so futzing with this in order to get it to work that way is less than ideal. If the POSIX APIs on Windows don't support non-blocking operation of anonymous pipes then we should forbid using them on that platform.

compnerd:
The one thing that I am concerned about is surfacing the API differences ... that has always caused trouble as it ends up causing downstream packages to break when they chance as there isn't always sufficient testing to ensure that all the paths are updated properly.

Lukasa:
What surfacing of the different API surfaces do you mean?

compnerd:
If there are public interfaces that may reference anything.  I suppose the alternate way to approach this: what exactly do you have in mind to remove?

PR #2199, opened 2022-06-17
Adopt `Sendable` in `NIOTestUtils`

dnadoba:

The parameter `decoderFactory` of the static methods `ByteToMessageDecoderVerifier.verifyDecoder` do not need to be `@escaping`. I have made them non-escaping as part of `Sendable` adoption because we would otherwise need to think about if they should be `@Sendable` too.

`VerificationError` is interesting, see the code comment for more information.


Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -207,3 +208,12 @@ extension ByteToMessageDecoderVerifier {
         }
     }
 }
+
+#if swift(>=5.5) && canImport(_Concurrency)
+/// `VerificationError` conforms to `Error` and therefore `Sendable`.
+/// `VerificationError` has a stored property `errorCode` of type `ErrorCode` which can store `NIOAny` which is not and can not be `Sendable`.
+/// In addtion, `ErrorCode` can also store a user defined `OutputType` which is not required to be `Sendable` but we could require it to be `Sendable`.
+/// We have to choises, we could either lie and confrom `ErrorCode` to `Sendable` with `@unchecked` or do the same but for `VerificationError`.
```
Lukasa, line 216:
```suggestion
/// We have two choices, we could either lie and conform `ErrorCode` to `Sendable` with `@unchecked` or do the same but for `VerificationError`.
```

Sources/NIOTestUtils/ByteToMessageDecoderVerifier.swift
```diff
@@ -207,3 +208,12 @@ extension ByteToMessageDecoderVerifier {
...
+/// As `VerificationError` already conforms to `Sendable` this sounds like the best option and this still allows us to adopt `Sendable` for `ErrorCode` later.
```
Lukasa, line 217: Can you clarify that the reason `VerificationError` already conforms to `Sendable` is that it's of type `Error`, which must conform.

Lukasa (approved):

Issue #1681, opened 2020-10-19
Typo in IdleChannelHandler is assigning a scheduled task to the wrong member field


abidon:
By browsing the code I think I found a typo in `IdleChannelHandler.makeAllTimeoutTask` (see [here](https://github.com/apple/swift-nio/blob/2.23.0/Sources/NIO/ChannelHandlers.swift#L302-L310)).

When makeAllTimeoutTask is rescheduling, the resulting `Scheduled<T>` is assigned to `scheduledReaderTask` instead of `scheduledAllTask`.

Lukasa:
Yup, looks like a typo to me. I think it ought to be possible to reproduce a bug there in a unit test, which would make a nice contribution along with a fix. Do you think you could do that?

stepan-ulyanin:
Working on this one

PR #2200, opened 2022-06-17
Fix the typos in the task assignments in makeAllTimeoutTask

stepan-ulyanin:
Assigns the scheduled task to `scheduledAllTask` property (instead of the `scheduledReaderTask`) of `IdleStateHandler` in `makeAllTimeoutTask` method.

### Motivation:
There was a typo in the property name, per https://github.com/apple/swift-nio/issues/1681

### Modifications:

The typo is fixed, the comments are also updated.

### Result:

The scheduled all timeout task will be stored in the right property, namely `scheduledAllTask`.

### DIFF
--- a/Sources/NIOCore/ChannelHandlers.swift
+++ b/Sources/NIOCore/ChannelHandlers.swift
@@ -310,13 +310,13 @@ public final class IdleStateHandler: ChannelDuplexHandler, RemovableChannelHandl
 
             let diff = .now() - latestLast
             if diff >= timeout {
-                // Reader is idle - set a new timeout and trigger an event through the pipeline
-                self.scheduledReaderTask = context.eventLoop.scheduleTask(in: timeout, self.makeAllTimeoutTask(context, timeout))
+                // Both reader and writer are idle - set a new timeout and trigger an event through the pipeline
+                self.scheduledAllTask = context.eventLoop.scheduleTask(in: timeout, self.makeAllTimeoutTask(context, timeout))
 
                 context.fireUserInboundEventTriggered(IdleStateEvent.all)
             } else {
-                // Read occurred before the timeout - set a new timeout with shorter delay.
-                self.scheduledReaderTask = context.eventLoop.scheduleTask(deadline: latestLast + timeout, self.makeAllTimeoutTask(context, timeout))
+                // Read or write occurred before the timeout - set a new timeout with shorter delay.
+                self.scheduledAllTask = context.eventLoop.scheduleTask(deadline: latestLast + timeout, self.makeAllTimeoutTask(context, timeout))
             }
         }
     }


stepan-ulyanin:
@Lukasa any suggestions on testing here?

stepan-ulyanin:
@Lukasa thanks a bunch, will let you know how it goes

stepan-ulyanin:
> 3\. that

@Lukasa I finally got back to this one. I followed your suggestion and found out that the events are properly fired only when I run the test through `lldb`. If I run the test without stepping through, I get no events being fired and captured into the array.

Here are the steps I am running, annotated with the steps from your suggestions, maybe you can spot an issue?

```swift
// Create the EmbeddedChannel
let embeddedChannel = EmbeddedChannel()

// Add the IdleStateHandler
try await embeddedChannel.pipeline.addHandler(idleStateHandler)

// Add the recording handler
try await embeddedChannel.pipeline.addHandler(testHandler)

// Activate the channel
try await embeddedChannel.connect(to: .makeAddressResolvingHost("127.0.0.1", port: 0))

// Advance time by 3 seconds
embeddedChannel.embeddedEventLoop.advanceTime(by: .seconds(3))

// write through the channel
try embeddedChannel.writeInbound(ByteBuffer())

// Advance time by 4 more seconds
embeddedChannel.embeddedEventLoop.advanceTime(by: .seconds(4))

// Remove the handler from the channel
try await embeddedChannel.pipeline.removeHandler(idleStateHandler)

// Confirm that no user events were fired yet
print(testHandler.events)

// Advance time by 4 more seconds
embeddedChannel.embeddedEventLoop.advanceTime(by: .seconds(4))

// Confirm that no user events were fired
print(testHandler.events)  // prints [ ]
```

Lukasa:
The issue isn't running the test in `lldb`, it's using an async test. `EmbeddedChannel` is totally synchronous, so you don't want to write a test that's `async`. You can add and remove the channel handlers using the synchronous operations on the channel pipeline, which should remove the futures in your code.

stepan-ulyanin:
> The issue isn't running the test in `lldb`, it's using an async test. `EmbeddedChannel` is totally synchronous, so you don't want to write a test that's `async`. You can add and remove the channel handlers using the synchronous operations on the channel pipeline, which should remove the futures in your code.

Ah great, thanks for this, I will work on this more and report back asap.

stepan-ulyanin:
@Lukasa I dug around a little more and have one last thing I can't fully figure out. We are advancing the time on the embedded event loop, hence we are executing the scheduled tasks, for example the read timeout task that we schedule after a timeout of a number of seconds.

How does this event ever get to the following handler from the idle state handler? I was under the impression that you would need to fire off the 
```
context.fireUserInboundEventTriggered(IdleStateEvent.read)
```
But `.now() - self.lastReadTime` in:
https://github.com/apple/swift-nio/blob/d05d2fd12573db3cb060ddf2f85433aeef0bb00f/Sources/NIOCore/ChannelHandlers.swift#L263-L268
is only greater than the timeout if `.now` is a number of real seconds ahead of the `lastReadTime`, not the embedded event loop seconds, am I missing something? 

I have been only able to propagate the event to the next handler by `sleep(K)`ing in the test, which we obviously don't want.

Lukasa:
Oh shoot you're right, this class uses a built-in clock. I guess this patch got a bit bigger!

Can you update the idle state handler to have an `internal` constructor that accepts a `now()` function as an argument, and then default it to `.now()` for the public `init`? That way we can keep the default behaviour but override the behaviour in testing, where we can defer to the `EmbeddedEventLoop`'s idea of `now()`.

stepan-ulyanin:
> Oh shoot you're right, this class uses a built-in clock. I guess this patch got a bit bigger!
> 
> Can you update the idle state handler to have an `internal` constructor that accepts a `now()` function as an argument, and then default it to `.now()` for the public `init`? That way we can keep the default behaviour but override the behaviour in testing, where we can defer to the `EmbeddedEventLoop`'s idea of `now()`.

You got it, thanks for the tip!

stepan-ulyanin:
@Lukasa Ok, I have been able to see the `IdleStateEvent.read` fired from an orphan `scheduledReaderTask` after we remove the idle state handler from the pipeline.

One last thing. The orphan event gets fired from:
https://github.com/apple/swift-nio/blob/bedb9fe613f2c4b460d6607a274ff2a7c9555aba/Sources/NIOCore/ChannelHandlers.swift#L312-L316
However, it never reaches the next handler (my test handler), the 
`userInboundEventTriggered(context:,event:)` is never called with the event. I actually see the event triggered in my logs that I added to the `makeReadTimeoutTask`. Any idea why?

Sources/NIOCore/ChannelHandlers.swift
```diff
@@ -237,19 +248,18 @@ public final class IdleStateHandler: ChannelDuplexHandler, RemovableChannelHandl
...
     private func shouldReschedule(_ context: ChannelHandlerContext) -> Bool {
-        if context.channel.isActive {
-            return true
-        }
-        return false
+        return context.channel.isActive
     }
 
-    private func makeReadTimeoutTask(_ context: ChannelHandlerContext, _ timeout: TimeAmount) -> (() -> Void) {
+    private typealias TimeoutTask = () -> Void
```
stepan-ulyanin, line 260: It was quite hard to follow some of these functions, so I added a couple of type aliases. Let me know if I should remove these @Lukasa

Sources/NIOCore/ChannelHandlers.swift
```diff
@@ -189,11 +189,22 @@ public final class IdleStateHandler: ChannelDuplexHandler, RemovableChannelHandl
     private var scheduledReaderTask: Optional<Scheduled<Void>>
     private var scheduledWriterTask: Optional<Scheduled<Void>>
     private var scheduledAllTask: Optional<Scheduled<Void>>
+    private var now: () -> NIODeadline
+    
+    public convenience init(readTimeout: TimeAmount? = nil, writeTimeout: TimeAmount? = nil, allTimeout: TimeAmount? = nil) {
```
stepan-ulyanin, line 194: I didn't want to open the `now` parameter to the public so I opted in for a public convenience initializer here with the same public API as before.

Sources/NIOCore/ChannelHandlers.swift
```diff
@@ -294,42 +304,45 @@ public final class IdleStateHandler: ChannelDuplexHandler, RemovableChannelHandl
...
-                // Read occurred before the timeout - set a new timeout with shorter delay.
-                self.scheduledReaderTask = context.eventLoop.scheduleTask(deadline: latestLast + timeout, self.makeAllTimeoutTask(context, timeout))
+                // Read or write occurred before the timeout - set a new timeout with shorter delay.
+                self.scheduledAllTask = context.eventLoop.scheduleTask(deadline: latestLast + timeout, self.makeAllTimeoutTask(context, timeout))
             }
         }
     }
 
-    private func schedule(_ context: ChannelHandlerContext, _ amount: TimeAmount?, _ body: @escaping (ChannelHandlerContext, TimeAmount) -> (() -> Void) ) -> Scheduled<Void>? {
+    private typealias TimeoutTaskFactory = (ChannelHandlerContext, TimeAmount) -> TimeoutTask
```
stepan-ulyanin, line 334: Not sure what a good name for this could be, `Factory` sounds a little Javaesque, @Lukasa let me know if you have a better name in mind, or if I should revert these type aliases back. 😅

Tests/NIOPosixTests/IdleStateHandlerTest.swift
```diff
@@ -14,7 +14,7 @@
 
 import XCTest
 @testable import NIOCore
-import NIOEmbedded
+@testable import NIOEmbedded
```
stepan-ulyanin, line 17: I needed an access to the `_now` of the `EmbeddedEventLoop`.

Tests/NIOPosixTests/IdleStateHandlerTest.swift
```diff
@@ -180,4 +180,42 @@ class IdleStateHandlerTest : XCTestCase {
...
+        try embeddedChannel.pipeline.syncOperations.connect(
+            to: .init(ipAddress: "127.0.0.1", port: 0),
+            promise: nil
+        )
+        
+        embeddedChannel.embeddedEventLoop.advanceTime(by: .seconds(3))
+        try embeddedChannel.writeInbound(ByteBuffer())
+        
+        embeddedChannel.embeddedEventLoop.advanceTime(by: .seconds(3))
+        try embeddedChannel.pipeline.removeHandler(idleStateHandler).wait()
```
stepan-ulyanin, line 215: I couldn't find a `removeHandler` in `syncOperations` so opted in for a future that we wait.

Lukasa:
Ah shoot, we invalidate the channel handler context when we remove the handler, so our testing strategy isn't going to work. This is kinda annoying: it means the code is fundamentally untestable, as there are no observable side-effects.

In this case, we can remove all the testing code and just fix up the typos.

stepan-ulyanin:
The PR is stale and the original issue was closed, closing the PR as well

PR #2204, opened 2022-06-24
Deprecate `NIOAtomics` in favor of `Atomics`

### LINKED ISSUE #1948: Replace NIOAtomics with swift-atomics
We should use the standard code now it's available.

stevapple:
Deprecate `NIOAtomics` in favor of `Atomics`, and replace all usages of `NIOAtomic`.

### Motivation:

Since we have an official `Atomics` library, we can move from our own `Atomics` implementation to it.
Resolves #1948.

### Modifications:

- `NIOAtomic` and `UnsafeEmbeddedAtomic` wrapper types are now deprecated;
- Swift NIO implementations and tests are moving from `NIOAtomic` to `ManagedAtomic`.

### Result:

We no longer need to maintain our own atomic types.
NIO will gain a dependency on `swift-atomics` package.


stevapple:
Note that `ordering` is set to `.relaxed` everywhere to preserve `NIOAtomic`'s behavior. We can fine-tune the behavior later.

Package.swift
```diff
@@ -135,6 +139,7 @@ let package = Package(
         .library(name: "NIOTestUtils", targets: ["NIOTestUtils"]),
     ],
     dependencies: [
+        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.0.0"),
```
Lukasa, line 142: I think we should immediately jump forward to requiring 1.0.2 or later.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -452,7 +452,8 @@ final class PendingDatagramWritesManager: PendingWritesManager {
         assert(self.isOpen)
         self.state.append(pendingWrite)
 
-        if self.state.bytes > waterMark.high && channelWritabilityFlag.compareAndExchange(expected: true, desired: false) {
+        if self.state.bytes > waterMark.high,
```
Lukasa, line 455: NIO style is to prefer `&&` unless a conditional binding or `case` statement is used.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -315,7 +315,8 @@ final class PendingStreamWritesManager: PendingWritesManager {
         assert(self.isOpen)
         self.state.append(.init(data: data, promise: promise))
 
-        if self.state.bytes > waterMark.high && channelWritabilityFlag.compareAndExchange(expected: true, desired: false) {
+        if self.state.bytes > waterMark.high,
```
Lukasa, line 318: Same note here about `&&`.

Lukasa (commented):
Fantastic, I'm really excited to see this come along. I have a few exceedingly tiny nits but otherwise I think this is a perfect patch and ready to go in. cc @weissi in case he's curious, and @lorentey because I know this is a thing he's wanted for a long time!

Lukasa:

Sources/NIOPosix/PendingDatagramWritesManager.swift
stevapple: I changed the style here because comma is more natural with a line break (or we’ll get an extremely long line here).

Shall I just ignore the length and keep them in the same line?

Lukasa: We can break a line after `&&`, and stylistically that's totally fine for NIO.

stevapple:
`Sendable` appears to be a problem for Swift 5.5, where `@preconcurrency` is not available.

I’ll ping on apple/swift-atomics#45 for this.

dnadoba:
`EventCounterHandler` has adopted `Sendable` in https://github.com/apple/swift-nio/pull/2199 which is not yet part of any release. We could therefore just remove `Sendable` in Swift  <= 5.5 as an alternative without actually breaking API.

stevapple:
cc @Lukasa @dnadoba

Sources/NIOConcurrencyHelpers/NIOAtomic.swift
```diff
@@ -193,6 +193,7 @@ extension UInt: NIOAtomicPrimitive {
 /// By necessity, all atomic values are references: after all, it makes no
 /// sense to talk about managing an atomic value when each time it's modified
 /// the thread that modified it gets a local copy!
+@available(*, deprecated, message:"please use Atomics.ManagedAtomic instead")
```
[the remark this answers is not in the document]
dnadoba: Thinking about this a bit more, maybe we should just link to the GitHub repository e.g.:
```suggestion
@available(*, deprecated, message:"please use ManagedAtomic from https://github.com/apple/swift-atomics instead")
```
@Lukasa are you aware of any prior art on how to link to packages in this case?

dnadoba (commented):
Thanks a lot @stevapple! Looks good from my side but we definitely need to wait for Cory's review.

Sources/NIOConcurrencyHelpers/NIOAtomic.swift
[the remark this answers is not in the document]
dnadoba: Then lets use `please use ManagedAtomic from https://github.com/apple/swift-atomics instead` @stevapple

Lukasa (approved):

Lukasa:

PR #2205, opened 2022-06-28
initialise EL pointer arrays

weissi:
### Motivation:

Although the uninitialised elements are never accessed (and therefore this isn't strictly speaking a real bug), `UBP`s should generally be initialised.

### Modifications:

Initialise the `iovec` and other pointer arrays in the EventLoop.

### Result:

Fewer uninitialised variables.

### DIFF
--- a/Sources/NIOPosix/PendingWritesManager.swift
+++ b/Sources/NIOPosix/PendingWritesManager.swift
@@ -29,7 +29,7 @@ private struct PendingStreamWrite {
 /// - returns: A tuple of the number of items attempted to write and the result of the write operation.
 private func doPendingWriteVectorOperation(pending: PendingStreamWritesState,
                                            iovecs: UnsafeMutableBufferPointer<IOVector>,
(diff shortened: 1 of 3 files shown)


Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -64,7 +64,7 @@ private func doPendingWriteVectorOperation(pending: PendingStreamWritesState,
     }
     defer {
         for i in 0..<numberOfUsedStorageSlots {
-            storageRefs[i].release()
+            storageRefs[i]?.release()
```
Lukasa, line 67: Any reason this isn't also a force unwrap?

weissi:
@Lukasa actually, I don't think this is a good idea anymore :). If anything, we should use `initialize` on individual elements when we set them and `deinitialize` after having used them. We'd then ~~hope~~ verify that `initialize` is the same as just `=` and `deinitialize` is a no-op. That way we'd always correctly pair `initialize` and `deinitialize`.

PR #2206, opened 2022-06-28
Adopt `noasync` for `ThreadSpecificVariable`

dnadoba:
Mark `ThreadSpecificVariable.currentValue` as not available from asynchronous context.

### Motivation:
From [SE-0340 (Unavailable From Async Attribute)](https://github.com/apple/swift-evolution/blob/main/proposals/0340-swift-noasync.md#introduction):
> The Swift concurrency model allows tasks to resume on different threads from the one they were suspended on. For this reason, API that relies on thread-local storage, locks, mutexes, and semaphores, should not be used across suspension points.

### Modifications:

add `@available(*, noasync)` to  `ThreadSpecificVariable.currentValue` in Swift 5.7+

### Result:

Prevent accidental misuse of `ThreadSpecificVariable`

Note: The `noasync` diagnostics in the latest Swift 5.7 nightly compiler emits errors instead of warnings but this is already fixed and will be part of the next 5.7 nightly toolchain. https://github.com/apple/swift/issues/59748

### DIFF
--- a/Sources/NIOPosix/Thread.swift
+++ b/Sources/NIOPosix/Thread.swift
@@ -191,20 +191,20 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
     @available(*, noasync, message: "threads can change between suspension points and therefore the thread specific value too")
     public var currentValue: Value? {
         get {
-            get()
+            self.get()
         }
         set {
-            set(newValue)
+            self.set(newValue)
         }
     }
     #else
     /// The value for the current thread.
     public var currentValue: Value? {
         get {
-            get()
+            self.get()
         }
         set {
-            set(newValue)
+            self.set(newValue)
         }
     }
     #endif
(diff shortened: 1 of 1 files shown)


Sources/NIOPosix/Thread.swift
```diff
@@ -185,25 +185,46 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
...
+    #else
+    /// The value for the current thread.
+    public var currentValue: Value? {
+        get {
+            get()
+        }
+        set {
+            set(newValue)
+        }
+    }
```
Lukasa, line 209: Please use explicit `self`.

Sources/NIOPosix/Thread.swift
```diff
@@ -185,25 +185,46 @@ public final class ThreadSpecificVariable<Value: AnyObject> {
...
+    #endif
+    
+    /// Get the current value for the calling thread.
+    func get() -> Value? {
+        guard let raw = self.key.get() else { return nil }
+        // parenthesize the return value to silence the cast warning
+        return (Unmanaged<BoxedType>
+            .fromOpaque(raw)
+            .takeUnretainedValue()
+            .value.1 as! Value)
```
Lukasa, line 219: Can we clean up this indentation?

dnadoba: It is already indented through Xcodes Re-Indent command and I don't see any mistakes. What would you like to be different?

dnadoba:

Lukasa (approved):

PR #2207, opened 2022-06-28
Adopt `noasync` for `Lock.lock()` and `Lock.unlock()`

dnadoba:
Mark `Lock.lock()` and `Lock.unlock()` as not available from asynchronous context.

### Motivation:
From [SE-0340 (Unavailable From Async Attribute)](https://github.com/apple/swift-evolution/blob/main/proposals/0340-swift-noasync.md#introduction):
> The Swift concurrency model allows tasks to resume on different threads from the one they were suspended on. For this reason, API that relies on thread-local storage, locks, mutexes, and semaphores, should not be used across suspension points.

### Modifications:

add `@available(*, noasync)` to  `Lock.lock()` and `Lock.unlock()` in Swift 5.7+

### Result:

Prevent accidental misuse of `Lock`

Note: The `noasync` diagnostics in the latest Swift 5.7 nightly compiler emits errors instead of warnings but this is already fixed and will be part of the next 5.7 nightly toolchain. https://github.com/apple/swift/issues/59748

### DIFF
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -62,11 +62,26 @@ public final class Lock {
         mutex.deallocate()
     }
 
+    #if swift(>=5.7)
+    /// Acquire the lock.
+    ///
+    /// Whenever possible, consider using `withLock` instead of this method and
+    /// `unlock`, to simplify lock handling.
+    @available(*, noasync, message: "use 'withLock(_:)' if you do not cross suspension points")
+    public func lock() {
+        self._lock()
+    }
+    #else
     /// Acquire the lock.
     ///
     /// Whenever possible, consider using `withLock` instead of this method and
     /// `unlock`, to simplify lock handling.
     public func lock() {
+        self._lock()
+    }
+    #endif
+    
+    private func _lock() {
 #if os(Windows)
         AcquireSRWLockExclusive(self.mutex)
 #else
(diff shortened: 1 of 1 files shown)


Lukasa:
I think this is excessive. While it’s unacceptable to use Lock across a suspension point, it’s totally fine to use it in all other circumstances. Indeed, we actively do in a number of our async projects. So I’m inclined not to want to do this.

fabianfett:
If this would land, this would break AHC and Postgres. The Postgres code in question:

```swift
    func next() async throws -> PostgresRow? {
        self.lock.lock()
        switch self.state.next() {
        case .returnNil:
            self.lock.unlock()
            return nil

        case .returnRow(let data, signalDemandTo: let source):
            self.lock.unlock()
            source?.demand()
            return PostgresRow(
                data: data,
                lookupTable: self.lookupTable,
                columns: self.columns
            )

        case .throwError(let error):
            self.lock.unlock()
            throw error

        case .hitSlowPath:
            return try await withCheckedThrowingContinuation { continuation in
                let slowPathAction = self.state.next(for: continuation)
                self.lock.unlock()
                switch slowPathAction {
                case .signalDemand(let source):
                    source.demand()
                case .none:
                    break
                }
            }
        }
    }
```

This perf optimization is not possible, if this PR lands.

dnadoba:
It will still be possible to call `noasync` methods in async context indirectly. You will however need to wrap them in another function/closure like this:
```
_ = { lock.lock() }()
_ = { lock.unlock() }()
```

Lukasa:
My inclination is to still avoid this change. I don't know that it provides sufficient value to justify its extra cost on users. Given that the diagnostic doesn't warn against the thing we actually intend to prevent, and that we do genuinely use these functions in async context (and intend to continue to do so), I'm still of the opinion that this is an unnecessarily heavy-handed approach.

PR #2213, opened 2022-06-30
Adopt `Sendable` for types in `NIOHTTP1`

dnadoba:

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypes.swift
+++ b/Sources/NIOHTTP1/HTTPTypes.swift
@@ -36,7 +36,7 @@ internal enum KeepAliveState {
 
 /// A representation of the request line and header fields of a HTTP request.
 public struct HTTPRequestHead: Equatable {
-    fileprivate final class _Storage {
+    private final class _Storage {
         var method: HTTPMethod
         var uri: String
         var version: HTTPVersion
@@ -123,11 +123,6 @@ public struct HTTPRequestHead: Equatable {
     }
 }
 
-#if swift(>=5.6)
-@available(*, unavailable)
-extension HTTPRequestHead._Storage: Sendable {}
-#endif
-
 #if swift(>=5.5) && canImport(_Concurrency)
 extension HTTPRequestHead: @unchecked Sendable {}
 #endif
@@ -180,7 +175,7 @@ extension HTTPResponseHead {
 
 /// A representation of the status line and header fields of a HTTP response.
 public struct HTTPResponseHead: Equatable {
-    fileprivate final class _Storage {
+    private final class _Storage {
         var status: HTTPResponseStatus
         var version: HTTPVersion
         init(status: HTTPResponseStatus, version: HTTPVersion) {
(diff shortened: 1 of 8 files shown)


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -36,7 +36,7 @@ internal enum KeepAliveState {
 
 /// A representation of the request line and header fields of a HTTP request.
 public struct HTTPRequestHead: Equatable {
-    private final class _Storage {
+    fileprivate final class _Storage {
```
dnadoba, line 39: I change the access level from `private` to `fileprivate` to be able to mark `_Storage` as non-sendable which is only possible in an extension. No strong optioned on that though, so happy to make it `private` again.

Lukasa: I think we don't need to mark internal types as non-sendable, so I'm happy for us to throw this away.

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
```diff
@@ -397,3 +401,8 @@ public final class NIOHTTPClientResponseAggregator: ChannelInboundHandler, Remov
         self.buffer = context.channel.allocator.buffer(capacity: 0)
     }
 }
+
+
+#if swift(>=5.5) && canImport(_Concurrency)
+extension NIOHTTPClientResponseAggregator: @unchecked Sendable {}
```
Lukasa, line 407: There's no way this type is `Sendable`.

Lukasa (changes requested):

Sources/NIOHTTP1/HTTPTypes.swift
dnadoba: `private` again https://github.com/apple/swift-nio/pull/2213/commits/9d78041cee755bf0870aac569bb583545d0e8321

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
```diff
@@ -301,6 +301,10 @@ public final class NIOHTTPServerRequestAggregator: ChannelInboundHandler, Remova
     }
 }
 
+#if swift(>=5.5) && canImport(_Concurrency)
+extension NIOHTTPServerRequestAggregator: @unchecked Sendable {}
```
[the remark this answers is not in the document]
dnadoba: Ups 🫣... fixed in https://github.com/apple/swift-nio/pull/2213/commits/9d78041cee755bf0870aac569bb583545d0e8321

Sources/NIOHTTP1/NIOHTTPObjectAggregator.swift
dnadoba: Ups 🫣... fixed in https://github.com/apple/swift-nio/pull/2213/commits/9d78041cee755bf0870aac569bb583545d0e8321

Lukasa (approved):
Nice work!

PR #2217, opened 2022-07-05
Adopt `Sendable` for Types in `NIOWebSocket`

dnadoba:


Sources/NIOWebSocket/WebSocketProtocolErrorHandler.swift
```diff
@@ -42,3 +42,8 @@ public final class WebSocketProtocolErrorHandler: ChannelInboundHandler {
         context.fireErrorCaught(error)
     }
 }
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension WebSocketProtocolErrorHandler: Sendable {}
```
dnadoba, line 48: We could make `WebSocketProtocolErrorHandler` conform to `Sendable` as it doesn't have any state but I think we don't want to guarantee it forever.

Lukasa (approved):

PR #2219, opened 2022-07-05
NIOCore: handle error description on Windows

compnerd:
We would previously attempt to convert an error to a description without
consideration for the provenance on Windows.  This would result in a
failure if the error code was not from `errno`.  Account for the source
of the error and translate it appropriately.  We can not retrieve
descriptions on errors that may originate from Windows or WinSock as
well as the C library.

### DIFF
--- a/Sources/NIOCore/IO.swift
+++ b/Sources/NIOCore/IO.swift
@@ -14,7 +14,20 @@
 
 #if os(Windows)
 import ucrt
+import func WinSDK.FormatMessageW
+import func WinSDK.LocalFree
+import let WinSDK.FORMAT_MESSAGE_ALLOCATE_BUFFER
+import let WinSDK.FORMAT_MESSAGE_FROM_SYSTEM
+import let WinSDK.FORMAT_MESSAGE_IGNORE_INSERTS
+import let WinSDK.LANG_NEUTRAL
+import let WinSDK.SUBLANG_DEFAULT
 import typealias WinSDK.DWORD
+import typealias WinSDK.WCHAR
+import typealias WinSDK.WORD
+
+internal func MAKELANGID(_ p: WORD, _ s: WORD) -> DWORD {
+  return DWORD((s << 10) | p)
+}
 #elseif os(Linux) || os(Android)
 import Glibc
 #elseif os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/IO.swift
```diff
@@ -107,13 +121,46 @@ private func reasonForError(errnoCode: CInt, reason: String) -> String {
...
 }
 
+#if os(Windows)
+private func reasonForWinError(_ code: DWORD) -> String {
+    let dwFlags: DWORD = DWORD(FORMAT_MESSAGE_ALLOCATE_BUFFER)
+                       | DWORD(FORMAT_MESSAGE_FROM_SYSTEM)
+                       | DWORD(FORMAT_MESSAGE_IGNORE_INSERTS)
+
+    var buffer: UnsafeMutablePointer<WCHAR>?
+    let dwResult: DWORD = withUnsafeMutablePointer(to: &buffer) {
```
Lukasa, line 131: This code is kinda wild: it allocates a pointer with a value of `nil`, then takes a pointer to that pointer and casts it to pointer to 2 WCHARs, then passes that to `FormatMessageW`, telling it that it should allocate at least zero bytes. This code scares the hell of out of me.

I appreciate that Win32 is the villain here, but can we add a comment explaining that because we've set `FORMAT_MESSAGE_ALLOCATE_BUFFER` that `FormatMessageW` will allocate a buffer for us, but that it requires us to type-pun the input because it only takes an LPWSTR instead of a pointer to one?

compnerd: Sure, I can add a comment, but yes, this is the call in this case.  The fact that it takes a `LPWSTR` here is just mindboggling.

Lukasa (approved):

Sources/NIOCore/IO.swift
```diff
@@ -14,7 +14,21 @@
...
+import let WinSDK.FORMAT_MESSAGE_ALLOCATE_BUFFER
+import let WinSDK.FORMAT_MESSAGE_FROM_SYSTEM
+import let WinSDK.FORMAT_MESSAGE_IGNORE_INSERTS
+import let WinSDK.LANG_NEUTRAL
+import let WinSDK.SUBLANG_DEFAULT
 import typealias WinSDK.DWORD
+import typealias WinSDK.WCHAR
+import typealias WinSDK.WORD
+
+@_transparent
```
Lukasa, line 28: Can we remove `@_transparent`? It's fine if this generates a function call and I'd prefer we didn't use attributes the Swift team might rename if we can possibly avoid it.

Lukasa (changes requested):

Sources/NIOCore/IO.swift
compnerd: I suppose; it is only there to deal with limitations of the clang importer.  This is a function-like macro, just that it cannot be imported by the swift compiler.

PR #2220, opened 2022-07-06
Adopt `Sendable` for `NIOChatServer` and `MIOMulticastChat` Examples

dnadoba:
The examples already compile and don't produce sendable warnings. However, with strict concurrency checking (`-warn-concurrency`) turned on, they produce some `Sendable` warnings which are easily fixable. 
This doesn't fix all e.g. accessing `CommandLine.arguments` still produces a warnings (Reference to static property 'arguments' is not concurrency-safe because it involves shared mutable state) but it fixes the warnings which can be easily fixed today.

### DIFF
--- a/Sources/NIOChatServer/main.swift
+++ b/Sources/NIOChatServer/main.swift
@@ -85,7 +85,7 @@ final class ChatHandler: ChannelInboundHandler {
         var buffer = context.channel.allocator.buffer(capacity: read.readableBytes + 64)
         buffer.writeString("(\(context.remoteAddress!)) - ")
         buffer.writeBuffer(&read)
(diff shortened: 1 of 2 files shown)


dnadoba:
Failure in 5.7 looks like a regression unrelated to this PR. I think its a bug in the computer and but just started investigating

dnadoba:
Swift 5.7 CI is failing because of:
https://github.com/apple/swift/issues/59909
https://github.com/apple/swift/issues/59910
https://github.com/apple/swift/issues/59911
https://github.com/apple/swift/issues/59913

Lukasa (approved):

PR #2224, opened 2022-07-07
Enhance and rename AsyncEmbeddedEventLoop

Lukasa:
Motivation

AsyncEmbeddedEventLoop is an important part of our ongoing testing story
for NIO. However, it suffers from two problems.

The first is a usability one. As we discovered during the original
implementation, following EmbeddedEventLoop's pattern of not having any
EventLoop "Tasks" execute until run() meant that simple constructs like
`EventLoopFuture.wait` and `EventLoopFuture.get` didn't work at all,
forcing us to add an annoying `awaitFuture` method.

When playing with implementing EmbeddedChannel, this got worse, as I/O
methods also don't run immediately if called from the testing thread. We
couldn't easily work around this issue, and it meant that common
patterns (like calling `channel.writeAndFlush`) would deadlock!

This is unacceptable, so a change had to be made.

While we're here, we received feedback that the name is unclear to
users. Given that this particular event loop is in no sense "embedded",
we no longer need the name, so we can take this opportunity to use a
better one.

Note that this is a patch change only because the feature has not yet shipped.

Modifications

Changed `func execute` to immediately execute its task body, and to
dequeue all pending tasks at this time. Essentially, it's the equivalent
to run(). This is a major change in its behaviour.

Renamed the loop to `NIOAsyncTestingEventLoop`.

Result

Better names, easier to use.


FranzBusch (approved):
LGTM! I really like the new naming

ktoso (approved):
LGTM both name and semantics sound good to me!

I think I missed to chime in a few weeks ago, but still  looks good 🥳

Sources/NIOEmbedded/AsyncTestingEventLoop.swift
```diff
@@ -57,7 +57,7 @@ import NIOConcurrencyHelpers
 /// `loop.run()`. As a result, if you need to await an `EventLoopFuture` created on this loop you
 /// should use `awaitFuture`.
```
dnadoba, line 58: This is no longer true, right? Should we now deprecate `awaitFuture`?

Lukasa: I think we can now just remove it.

Lukasa: Oh, in fact this patch _does_ remove it: the comment is just vestigial.

dnadoba (approved):

Lukasa:
API breakage output:

```
17:15:45 No breaking changes detected in _NIOBeta
17:15:45 
17:15:45 2 breaking changes detected in NIOEmbedded:
17:15:45   💔 API breakage: class NIOAsyncEmbeddedEventLoop has been removed
17:15:45   💔 API breakage: struct NIOAsyncEmbeddedEventLoopError has been removed
17:15:45 
17:15:45 No breaking changes detected in NIOWebSocket
17:15:45 
17:15:45 No breaking changes detected in NIOTestUtils
17:15:45 
17:15:45 No breaking changes detected in NIOTLS
17:15:45 
17:15:45 No breaking changes detected in NIOPosix
17:15:45 
17:15:45 No breaking changes detected in NIOFoundationCompat
17:15:45 
17:15:45 No breaking changes detected in NIOConcurrencyHelpers
17:15:45 
17:15:45 No breaking changes detected in _NIOConcurrency
17:15:45 
17:15:45 No breaking changes detected in NIOCore
17:15:45 
17:15:45 No breaking changes detected in NIOHTTP1
17:15:45 
17:15:45 No breaking changes detected in NIO
```

This is the intended behaviour, so I'm merging over the failures.

PR #2225, opened 2022-07-21
Workaround `Sendable` warning from Swift 5.7

dnadoba:
Workaround for https://github.com/apple/swift/issues/59911
```swift
/src/Tests/NIOHTTP1Tests/HTTPTest.swift:148:21: warning: 'buf' mutated after capture by sendable closure
                buf.writeString("\(c)")
```
The variable `buf` is mutated but only before being capture by the sendable closure. We now capture a copy explicitly which suppress this warning.

All other Sendable warnings/errors are fixed with the latest Swift 5.7 toolchain


Lukasa (approved):
Sensible workaround, thanks.

PR #2226, opened 2022-07-23
NIOCore: Implemented all three variants of _failEarlyRangeCheck methods for ByteBufferView

anishagg17:
Implements all versions of the `_failEarlyRangeCheck(_:bounds:)`  for `ByteBufferView`

### Motivation:

#1385 specifies the missing Collection customization points that we would like to customize.

### Modifications:

Adds three methods to the `ByteBufferView` conforming to Collection

### Result:

The default implementations of the `_failEarlyRangeCheck` methods are going to be customized according to needs in the `ByteBufferView`.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-views.swift
+++ b/Sources/NIOCore/ByteBuffer-views.swift
@@ -138,20 +138,15 @@ public struct ByteBufferView: RandomAccessCollection, NIOSendable {
         return (iterator, bytesToWrite)
     }
 
+    // These are implemented as no-ops for performance reasons.
     @inlinable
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
Thanks for this!

Much like with #2161, I think we can safely implement these three as no-ops: we do bounds-checking on all subscript paths, so these are simply duplicative of those checks. We can make the case that just as Array disables them for perf reasons, so will we.

anishagg17:
Turned into no-op functions ✅

Lukasa (approved):

Lukasa:
Thanks for this!

Lukasa:

PR #2228, opened 2022-07-26
`CurrentEventLoop` Proposal

dnadoba:
Many methods on `EventLoop`, `EventLoopFuture` and `Scheduled` take an escaping function type as an argument. 
Most of these function types do now need to conform to `Sendable` because the function may cross an isolation domain.
However, a common pattern in NIO programs guarantees that they do **not** cross an isolation domain. 
This PR makes it possible to encode this guarantee into the type system. 
The goal of this is not necessary to land the code but use it as starting point to talk about the semantics we want and need in the world of strict sendability checking. 
We also want to see if this problem can be and should be solved in a library or needs language support.

[Full proposal is at `docs/CurrentEventLoop.md`](https://github.com/apple/swift-nio/blob/0f28b22c746be2e06210a5a54e350f8d06a27c9b/docs/CurrentEventLoop.md)

cc @Lukasa @weissi @ktoso @DougGregor


Sources/NIOCore/ScheduledOnCurrentEventLoop.swift
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension ScheduledOnCurrentEventLoop: Sendable {}
+#endif
+
+extension Scheduled {
+    @inlinable public func iKnowIAmOnTheEventLoopOfThisScheduled(
```
ktoso, line 29: The name is trying a bit too hard to be funny -- can we have something that explains what it is doing?

Specifically this is `assumeOnScheduledEventLoop()` or something; `assume...` sounds good to me as a "assume or crash", so that could be adopted as general wording.

Sources/NIOCore/CurrentEventLoopPromise.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension CurrentEventLoopPromise: Sendable {}
+#endif
+
+extension EventLoopPromise {
+    @inlinable public func iKnowIAmOnTheEventLoopOfThisPromise(
```
ktoso, line 20: Same nitpick as below, it is trying a bit _too_ hard to be funny and does not explain why I'd be doing this.

Can this instead be `assumeOnPromiseEventLoop()` or `assumeOnEventLoopOfPromise()` or if you need to be super explicit `assume...OrCrash()`? The "iKnow" sound cute but don't help understand and parse the code while reading it.

Sources/NIOCore/CurrentEventLoopPromise.swift
```diff
@@ -0,0 +1,44 @@
+public struct CurrentEventLoopPromise<Value> {
```
ktoso, line 1: docs? :)

Sources/NIOCore/CurrentEventLoop.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension CurrentEventLoop: Sendable {}
+#endif
+
+extension EventLoop {
+    @inlinable public func iKnowIAmOnThisEventLoop(
```
ktoso, line 14: (Commented on naming below already), I don't think those `iKnow...` work well

ktoso (approved):
I like the idea -- the way of offering a non-sendable API-par types sounds great and should help in handler development, LGTM :) 

I'd bikeshed tne name a bit more, proposed a few ideas.

Sources/NIOCore/CurrentEventLoop.swift
```diff
@@ -0,0 +1,83 @@
+public struct CurrentEventLoop {
+    public let wrapped: EventLoop
+    @inlinable internal init(_ eventLoop: EventLoop) {
+        self.wrapped = eventLoop
+    }
+}
+
+#if swift(>=5.6)
+@available(*, unavailable)
+extension CurrentEventLoop: Sendable {}
```
ktoso, line 10: Add some documentation on it why it is not sendable - not everyone may understand this unavailable trick :)

Lukasa (commented):
I'm in favour of this proposal. I think it's a good idea to plan to adopt it in the future, including dealing with the naming bikeshed and adding all of the appropriate APIs to CurrentEventLoopFuture.

Lukasa:
This was implemented and shipped as the isolated views.

PR #2229, opened 2022-07-28
Fix strict concurrency checking diagnostics in `MTELG`

dnadoba:
### Motivation:

We want to incrementally comply with the strict sendability checking rules.

### Modifications:

- refactor code from using `var`s to one `let`
- storing shutdown callbacks as `Sendable` functions in `RunState`

### Result:

less warnings/errors in strict concurrency mode and we also got rid of one force-unwrap 🎉


Lukasa (approved):
Sounds good to me!

PR #2230, opened 2022-07-29
Implement a back-pressure aware `AsyncSequence` source

FranzBusch:
# Motivation
We ran into multiple use-cases (https://github.com/apple/swift-nio/pull/2067, https://github.com/grpc/grpc-swift/blob/main/Sources/GRPC/AsyncAwaitSupport/PassthroughMessageSource.swift) already where we want to vend an `AsyncSequence` where elements are produced from the sync world while the consumer is in the async world. Furthermore, we need the `AsyncSequence` to properly support back-pressure.
Since we already identified that this is something fundamental for our ecosystem and that current `AsyncSequence` sources are not providing the proper semantics or performance, it would be great to find a single solution that we can use everywhere.

We evaluated the current `AsyncSequnce` sources and decided against using them because of the following reasons:
- `AsyncChannel` is used in `async` to `async` communication which we don't need here
- `AsyncStream` is very close but is missing out on a few points
  - It only exposes back-pressure on the `yield()` edge and not on the `next()` edge
  - It only allows to `yield()` a single element instead of a sequence
  - It is not enforcing a single `AsyncIterator` tightly enough 

Before diving into the code, I think it is good to understand the goals of this `AsyncSequence`:
- The `AsyncSequence` should support a single unicast `Subscriber`
- The `AsyncSequence` should allow a pluggable back-pressure strategy
- The `AsyncSequence` should allow to yield a sequence of elements to avoid aquiring the lock for every element.
- We should make sure to do as few thread hops as possible to signal the producer to demand more elements.

# Modification
This PR introduces a new `AsyncSequence` called `NIOBackPressuredAsyncSequence`. The goal of that sequence to enable sync to async element streaming with back-pressure support.

# Result
We can now power our sync to async use-cases with this new `AsyncSequence`.

# Future work
There are couple of things left that I wanna land in a follow up PR:
1. An adaptive back-pressure strategy that grows and shrinks depending on the speed of consumption
2. A throwing version of this sequence
3. Potentially an async version that suspends on `yield()` and resumes when more elements should be demanded.


Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,988 @@
...
+/// The goal of this sequence is to bridge a stream of elements from the _synchronous_ world (e.g. elements from a Channel pipeline) into the _asynchronous_ world.
+/// Furthermore, it provides facilities to implement a back-pressure strategy which observes the number of elements that are yielded and demanded. This allows to signal the source to request more data.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOBackPressuredAsyncSequence<
+    Element,
+    Strategy: NIOBackPressuredAsyncSequenceStrategy,
+    Delegate: NIOBackPressuredAsyncSequenceDelegate
+> {
+    /// Simple struct for the return type of ``NIOBackPressuredAsyncSequence/makeSourceAndSequence(backPressureStrategy:delegate:)``.
+    public struct Initialized {
```
FranzBusch, line 61: Open to nit-picking on this name 😀

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+                self.lock.unlock()
+
+                return element
+
+            case .returnNil:
+                self.lock.unlock()
+                return nil
+
+            case .suspendTask:
+                // It is safe to hold the lock across this method
```
phausler, line 409: that isn't true here; the cancellation case will happen in two ways; either a) it will cancel asynchronously from another task (which is safe) or b) it will cancel immediately because the task is already cancelled at the point of entry (which is unsafe because you will become re-rentrant upon the lock)

FranzBusch: I was super unsure about this since the documentation is a bit bogus about it. It states both

> This differs from the operation cooperatively checking for cancellation and reacting to it in that the cancellation handler is always and immediately invoked when the task is canceled

and
> Doesn’t check for cancellation, and always executes the passed operation.

From what I understood there is that the operation is at least always executed. What you are saying is that if the `Task` is cancelled then `operation` is never called and this will result in a deadlock here, right?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+    /// This method is called after `next` was called on an iterator of the ``NIOBackPressuredAsyncSequence``.
+    ///
+    /// - Parameter bufferDepth: The current depth of the internal buffer.
+    /// - Returns: Returns wether new elements should be demanded.
+    mutating func didNext(bufferDepth: Int) -> Bool
+}
+
+/// The delegate of ``NIOBackPressuredAsyncSequence``.
+///
+/// - Important: All methods of this protocol are called inside the `Lock` of the ``NIOBackPressuredAsyncSequence``.
```
phausler, line 46: OOF that is a steep requirement... is there a way we can make these isolated instead? similar to how we implemented AsyncBuffer in the async algorithms package?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
FranzBusch: Okay I just tested the following

```swift
func main() async {
    do {
        try await  Task.sleep(nanoseconds: 2000000000)
    } catch {
        print("got cancelled")
        await withTaskCancellationHandler {
            print("operation")
        } onCancel: {
            print("cancelled")
        }

    }
}

let task = Task {
    await main()
}

Task {
    try! await Task.sleep(nanoseconds: 1000000000)
    print("cancelling")
    task.cancel()
}
```

and this prints

```
cancelling
got cancelled
cancelled
operation
```

Which is what you probably meant here. Thanks for spotting this! Now the question is how this should be done instead :D

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+/// when an ``NIOBackPressuredAsyncSequence/AsyncIterator`` requested the next value. The responsibility of the strategy is
+/// to determine wether more elements need to be demanded .
+///
+/// If more elements need to be requested, either the ``NIOBackPressuredAsyncSequenceSourceDelegate/demand()``
+/// method will be called or a ``NIOBackPressuredAsyncSequence/Source/YieldResult`` will be returned that indicates
+/// to demand more.
+///
+/// - Important: The methods of this protocol are guaranteed to be called serially. Furthermore, the implementation of these
+/// methods should not do any locking or call out to any other Task/Thread.
+public protocol NIOBackPressuredAsyncSequenceStrategy {
```
phausler, line 30: This is quite a strange shape/names-of-methods.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
FranzBusch: Hm, thinking out loud about this a bit. Is checking for `Task.isCancelled` before enough or does that introduce a race condition.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+
+            case .returnNil:
+                self.lock.unlock()
+                return nil
+
+            case .suspendTask:
+                // It is safe to hold the lock across this method
+                // since the closure is guaranteed to be run straight away
+                return await withTaskCancellationHandler {
+                    return await withCheckedContinuation { continuation in
```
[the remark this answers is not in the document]
FranzBusch: Yes our stance for now was to start with the checked one and once we are quite certain it is safe to move to the `unsafe` one.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+
+        @inlinable
+        /* fileprivate */ internal func sequenceDeinitialized() {
+            self.lock.withLock {
+                let action = self.stateMachine.sequenceDeinitialized()
+
+                switch action {
+                case .callDidTerminate:
+                    self.delegate?.didTerminate()
+                    self.delegate = nil
```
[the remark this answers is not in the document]
FranzBusch: But that is not a problem right?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+    /* fileprivate */ internal final class Storage: @unchecked Sendable {
+        /// The lock that protects our state.
+        @usableFromInline
+        /* private */ internal let lock = Lock()
+        /// The state machine.
+        @usableFromInline
+        /* private */ internal var stateMachine: StateMachine
+        /// The delegate.
+        @usableFromInline
+        /* private */ internal var delegate: Delegate?
```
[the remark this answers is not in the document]
FranzBusch: We are not using `weak` references due to their performance implications. That's why we are explicitly `nil`-ing out the `delegate` after we called `didTerminate`

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
phausler: from my experience deinit inside a lock can cause tons of issues. I have had to deal with dozens of bugs where other projects did stuff like that and it always is a recipe for disaster/bugs-later

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
Lukasa: To be clear, the failure mode of the unsafe variant is so horrifying that I'll be refusing to use it until this implementation has baked for at least a year.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
FranzBusch: The `AsyncBuffer` is an `actor` which we can't really use here because that would require `async` calls to the strategy.
In the end, we need something is running inside the state machine of the `NIOBackPressuredAsyncSequence` since it is basically an extension of our internal state but pluggable. 

I agree with you that this can be quite dangerous to hold, that's why we intend to supply two blessed implementations. We don't expect people to generally write their own strategies. cc @Lukasa

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
FranzBusch: Thanks for catching this @phausler, I moved the `withTaskCancellationHandler` around the whole method body now. This should make sure that only either of the two gets to run without lock re-entrancy.

Sources/NIOCore/AsyncSequences/NIOHighLowWatermarkBackPressuredAsyncSequenceStrategy.swift
```diff
@@ -0,0 +1,45 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
```

Tests/NIOCoreTests/AsyncSequences/NIOBackPressuredAsyncSequenceTests.swift
```diff
@@ -0,0 +1,544 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
```

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+    ///
+    /// - Parameter bufferDepth: The current depth of the internal buffer.
+    /// - Returns: Returns wether more elements should be demanded.
+    mutating func didYield(bufferDepth: Int) -> Bool
+
+    /// This method is called after `next` was called on an iterator of the ``NIOBackPressuredAsyncSequence``.
+    ///
+    /// - Parameter bufferDepth: The current depth of the internal buffer.
+    /// - Returns: Returns wether new elements should be demanded.
+    mutating func didNext(bufferDepth: Int) -> Bool
```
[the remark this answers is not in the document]
FranzBusch: I had `didDemand` first but that was also super confusing with `demand` existing on the delegate. Would renaming it to `didCallNext()` improve it? Other thing I can think about is, `didRegisterDemand` or `didRequest`. Open to better ideas :D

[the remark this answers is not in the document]
Lukasa: Maybe `didConsume`?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1003 @@
...
+        /// Otherwise, the element will get buffered.
+        ///
+        /// This can be called more than once and returns to the caller immediately
+        /// without blocking for any awaiting consumption from the iteration.
+        ///
+        /// - Parameter element: The element to yield to the stream.
+        /// - Returns: A ``NIOBackPressuredAsyncSequence/Source/YieldResult`` that indicates if the yield was successfull
+        /// and if more elements should be demanded.
+        @inlinable
+        public func yield<S: Sequence>(_ sequence: S) -> YieldResult where S.Element == Element {
```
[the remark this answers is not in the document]
FranzBusch: Added a part to the doc comment

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1004 @@
...
+                self.state = .finished
+
+                return .callDidTerminate
+
+            case .streaming(_, _, .some(let continuation), _, _):
+                // We have an outstanding continuation that needs to resumed
+                // and we can transition to finished here and inform the delegate
+                self.state = .finished
+
+                return .resumeContinuationWithNilAndCallDidTerminate(continuation)
```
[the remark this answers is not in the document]
FranzBusch: We can't throw a `CancellationError` here since our `NIOBackPressuredAsyncSequence` is non throwing. AFAIK for the sequences that are not throwing we should resume with `nil` for the throwing counter pat we can throw the `CancellationError`

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1004 @@
...
+                } else {
+                    self.state = .streaming(
+                        backPressureStrategy: backPressureStrategy,
+                        buffer: buffer,
+                        continuation: continuation,
+                        hasOutstandingDemand: shouldDemandMore,
+                        iteratorInitialized: iteratorInitialized
+                    )
+
+                    // This is weird somebody yielded an empty sequence but we can handle it
```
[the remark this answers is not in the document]
FranzBusch: I thought about simplifying this before already however I think regardless of how you change it around it won't get much shorter. I don't think we can return early with the current `hasOutstandingDemand` value. We need to inform the strategy even if an empty `yield` happened. Furthermore, the strategy might be doing any weird kind of thing where it returns a different demand behaviour even-though the `bufferDepth` is exactly the same.

I reordered this a bit a couple of times already and it always results in similar amounts of code.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
ktoso: If you can't throw returning nil sounds ok for this... Arguably this might want to get a throwing pair type but you'll be forced to replicate a lot of the type I guess :-(

FranzBusch:
The 5.7 CI failure seems kinda bogus:
> reading 'async let' can throw but is not marked with 'try'

But our `async let` cannot throw since the `AsyncSequence` is non-throwing. Trying to reproduce a simpler version of this.

Lukasa:
Is that bogus? The `await` could throw a `CancellationError`, could it not?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1063 @@
...
+    @inlinable
+    public static func makeSourceAndSequence(
+        backPressureStrategy: Strategy,
+        delegate: Delegate
+    ) -> NewSequence {
+        let sequence = Self(
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let source = Source(storage: sequence.internalClass.storage)
```
Lukasa, line 140: Can we add a file private getter to `sequence` to return the storage so we don't need two accessors here?

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1063 @@
...
+            /// The state once either any element was yielded or `next()` was called.
+            case streaming(
+                backPressureStrategy: Strategy,
+                buffer: CircularBuffer<Element>,
+                continuation: CheckedContinuation<Element?, Never>?,
+                hasOutstandingDemand: Bool,
+                iteratorInitialized: Bool
+            )
+
+            /// The sate once the underlying source signalled that it is finished.
```
Lukasa, line 526:
```suggestion
            /// The state once the underlying source signalled that it is finished.
```

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
Lukasa: Why do we need to inform the strategy on empty yield?

phausler:
> The 5.7 CI failure seems kinda bogus:
> 
> > reading 'async let' can throw but is not marked with 'try'
> 
> But our `async let` cannot throw since the `AsyncSequence` is non-throwing. Trying to reproduce a simpler version of this.

`async let` unfortunately does not support rethrowing... so basically that will claim it always throws...

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
phausler: It a non throwing `AsyncSequence` is running in a cancelled `Task` it will return nil at the earliest moment it can. (That may be deferred to the next call of `next()` in some cases since cancellation is cooperative, but a best effort should be made)

FranzBusch:
> > The 5.7 CI failure seems kinda bogus:
> > > reading 'async let' can throw but is not marked with 'try'
> > 
> > 
> > But our `async let` cannot throw since the `AsyncSequence` is non-throwing. Trying to reproduce a simpler version of this.
> 
> `async let` unfortunately does not support rethrowing... so basically that will claim it always throws...

Okay I did a bit of investigation. On 5.6 and 5.7 this is actually building perfectly fine. However, on swift/main this seems to be broken and suddenly it requires `try`. I will raise a bug against the compiler tomorrow, but this doesn't look like a problem that is blocking this PR since it is only against `main`.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1063 @@
...
+///
+/// The goal of this sequence is to bridge a stream of elements from the _synchronous_ world
+/// (e.g. elements from a ``Channel`` pipeline) into the _asynchronous_ world.
+/// Furthermore, it provides facilities to implement a back-pressure strategy which
+/// observes the number of elements that are yielded and consumed. This allows to signal the source to request more data.
+///
+/// - Important: This sequence is a unicast sequence that only supports a single ``NIOBackPressuredAsyncSequence/AsyncIterator``.
+/// If you try to create more than one iterator it will result in a `fatalError`.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOBackPressuredAsyncSequence<
```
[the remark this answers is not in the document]
FranzBusch: I am unsure, is this going to be our only blessed sequence? What if we are going to introduce the `async yield()` version of this sequence? Are we going to call that `NIOAsyncAsyncSequence`? :D

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
FranzBusch: IMO if the producer `yielded` to the source he is expecting that the strategy he configured is getting called and has a chance to evaluate with the current `bufferDepth`, if for some weird reason somebody implemented a strategy that signals demand on every other `yield()` invocation than early exiting on empty yields would not work.
In general, I think we are not losing anything by calling the strategy on empty yields, rather, we are just avoiding bugs down the road. Moreover, the empty `yield()` case is probably not happing too often anyhow.

FranzBusch:
Raised a bug against the swift repo for the compiler regression that we are seeing here with `async let` https://github.com/apple/swift/issues/60351

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
Lukasa: Is the yield one not going to have backpressure? :wink:

I think `BackPressured` is a bad name. We're going to have to disambiguate them, but having backpressure support is not what disambiguates this one.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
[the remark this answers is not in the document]
Lukasa: I think the exact opposite is true here.

Users don't have _any_ expectations about this type, because this type doesn't exist today and is not a direct analog of anything that currently exists. We define what the semantics of the `yield` operation are as it relates to the strategy. It is deeply not obvious to me that we might call the `didYield` method when nothing was actually yielded.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1068 @@
...
+                return .init(shouldProduceMore: shouldProduceMore)
+
+            case .streaming(var backPressureStrategy, var buffer, .some(let continuation), _, let iteratorInitialized):
+                // The buffer should always be empty if we hold a continuation
+                precondition(buffer.isEmpty)
+
+                self.state = .modifying
+
+                buffer.append(contentsOf: sequence)
+                let element = buffer.popFirst()
```
dnadoba, line 765: We could eliminate `buffer.popFirst()` by first getting the first element of the sequence and appending the remaining elements to the buffer. Not straight forward because `Sequence` doesn't guarantee it can be iterated multiple times but possible. I have done something similar for an `AsyncSequence` already:
https://github.com/apple/swift-nio/blob/ece5057615d1bee848341eceafdf04ca54d60177/Sources/NIOCore/AsyncAwaitSupport.swift#L299-L334

This will be especially nice for the case where `sequence.count == 1` as it will eliminate the roundtrip through the circular buffer. Not sure how much performance this will actually bring in practice though, so just a nice to have improvement.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1068 @@
...
+    /// This method is called after the `Subscriber` consumed an element.
+    /// More specifically this method is called after `next` was called on an iterator of the ``NIOBackPressuredAsyncSequence``.
+    ///
+    /// - Parameter bufferDepth: The current depth of the internal buffer.
+    /// - Returns: Returns whether the producer should produce more elements.
+    mutating func didConsume(bufferDepth: Int) -> Bool
+}
+
+/// The delegate of ``NIOBackPressuredAsyncSequence``.
+public protocol NIOBackPressuredAsyncSequenceDelegate {
```
dnadoba, line 46: Same here, `NIOBackPressuredAsyncSequenceDelegate` needs to inherit from Sendable because it is at least passed from the isolation context where it is initialized to the isolation context of the `NIOBackPressuredAsyncSequence`.

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
```diff
@@ -0,0 +1,1068 @@
...
+        }
+    }
+
+    /// This class is needed to hook the deinit to observe once all references to the ``NIOBackPressuredAsyncSequence`` are dropped.
+    ///
+    /// If we get move-only types we should be able to drop this class and use the `deinit` of the ``AsyncIterator`` struct itself.
+    ///
+    /// - Important: This is safe to be unchecked ``Sendable`` since the `storage` is ``Sendable`` itself.
+    @usableFromInline
+    /* fileprivate */ internal final class InternalClass: @unchecked NIOSendable {
```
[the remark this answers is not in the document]
FranzBusch: This requires `Sendable` in 5.6 sadly

Sources/NIOCore/AsyncSequences/NIOBackPressuredAsyncSequence.swift
FranzBusch: While I get what you want to optimise for here this would actually end up in something like this:

```swift
var iterator = sequence.makeIterator()

guard let firstElement = iterator.next() else {
    return .init(shouldProduceMore: hasOutstandingDemand)
}

while let nextElement = iterator.next() {
    buffer.append(nextElement)
}
```

This would actually result in worse code since it would repeatedly call `append`. Every call to append might grow the buffer and it has to adjust the index. Furthermore, for trivial types an implementation of `append(contentsOf:)` could do just a slab copy of the sequence.

After talking with @Lukasa about this a bit, I am going forward with early exiting if the buffer is empty. Furthermore, I am going to adopt the `Deque` from swift-collections since this provides a specialized `append(contentsOf:)` implementation.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+/// If more elements need to be produced, either the ``NIOAsyncSequenceProducerDelegate/produceMore()``
+/// method will be called or a ``NIOAsyncSequenceProducer/Source/YieldResult`` will be returned that indicates
+/// to produce more.
+///
+/// - Important: The methods of this protocol are guaranteed to be called serially. Furthermore, the implementation of these
+/// methods **MUST NOT** do any locking or call out to any other Task/Thread.
+public protocol NIOAsyncSequenceProducerBackPressureStrategy: Sendable {
+    /// This method is called after new elements were yielded by the producer to the source.
+    ///
+    /// - Parameter bufferDepth: The current depth of the internal buffer.
```
ktoso, line 34: Can you explain this a bit more? Depth isn't really that self explanatory in this context.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+            case .streaming(var backPressureStrategy, var buffer, .some(let continuation), let hasOutstandingDemand, let iteratorInitialized):
+                // The buffer should always be empty if we hold a continuation
+                precondition(buffer.isEmpty)
+
+                self.state = .modifying
+
+                buffer.append(contentsOf: sequence)
+
+                guard let element = buffer.popFirst() else {
+                    return .init(shouldProduceMore: hasOutstandingDemand)
```
glbrntt, line 772: We return while in the `.modifying` state here.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+                    continuation: nil,
+                    hasOutstandingDemand: false,
+                    iteratorInitialized: iteratorInitialized
+                )
+
+                return .suspendTask
+
+            case .streaming(_, _, .some, _, _):
+                // We have multiple AsyncIterators iterating the sequence
+                preconditionFailure("This should never happen since we only allow a single Iterator to be created")
```
glbrntt, line 947: Would we hit this if the consuming side did the following?

```swift
let iter = sequence.makeAsyncIterator()
async let foo = iter.next()
async let bar = iter.next()
```

Sources/NIOCore/AsyncSequences/NIOHighLowWatermarkBackPressuredAsyncSequenceStrategy.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+/// A high-low watermarked back-pressure strategy for a ``NIOAsyncSequenceProducer``.
+///
+/// This strategy does the following:
+/// - On yield it keeps on demanding more as long as the `highWatermark` isn't reached.
+/// - On next it starts to demand once the the `lowWatermark` is reached.
+public struct NIOHighLowWatermarkAsyncSequenceProducerBackPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategy {
```
glbrntt, line 21: This is quite a name...  

How about nesting it inside `NIOAsyncSequenceProducer` as `WatermarkedStrategy`?

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+            /// The initial state before either a call to `yield()` or a call to `next()` happened
+            case initial(
+                backPressureStrategy: Strategy,
+                iteratorInitialized: Bool
+            )
+
+            /// The state once either any element was yielded or `next()` was called.
+            case streaming(
+                backPressureStrategy: Strategy,
+                buffer: Deque<Element>,
```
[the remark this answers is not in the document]
FranzBusch: I discussed this with @Lukasa yesterday and he was fine in bringing in this dependency. The back story is that `Deque` is more performant than `CiruclarBuffer` since it implements a few customization points.

Sources/NIOCore/AsyncSequences/NIOHighLowWatermarkBackPressuredAsyncSequenceStrategy.swift
FranzBusch: I thought about this but decided against it for now since I had the feeling that it is very undiscoverable and it doesn't make it much better since in the end you have to spell out `NIOAsyncSequenceProducer<Element, NIOAsyncSequenceProducer.WatermarkedStrategy, Delegate>` open to change this though!

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+/// when a ``NIOAsyncSequenceProducer/AsyncIterator`` requested the next value. The responsibility of the strategy is
+/// to determine whether more elements need to be produced .
+///
+/// If more elements need to be produced, either the ``NIOAsyncSequenceProducerDelegate/produceMore()``
+/// method will be called or a ``NIOAsyncSequenceProducer/Source/YieldResult`` will be returned that indicates
+/// to produce more.
+///
+/// - Important: The methods of this protocol are guaranteed to be called serially. Furthermore, the implementation of these
+/// methods **MUST NOT** do any locking or call out to any other Task/Thread.
+public protocol NIOAsyncSequenceProducerBackPressureStrategy: Sendable {
```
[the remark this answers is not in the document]
FranzBusch: I think for most of the cases the two strategies that we are going to implement are used.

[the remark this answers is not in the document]
glbrntt: Do we absolutely want this to be generic then? 

Just wondering whether it would make sense instead to provide an enum-backed-struct with those two strategies and add more (or an escape hatch) if required. As a side effect it would make it mildly easier to construct the async source/sequence and avoid the naming issues.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
glbrntt: Ah of course. Can we explicitly mark the `AsyncIterator` as not being sendable?

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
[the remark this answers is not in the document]
FranzBusch: Open to do this change. Would like to get @Lukasa opinion on this as well, as it would basically mean we need to implement all strategies inside NIO.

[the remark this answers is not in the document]
Lukasa: I'm inclined to keep the generic: specialisation should allow the strategies to more-or-less go away. It does make the type hard to spell, I agree, but we don't want users to expose this type in their API regardless.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
FranzBusch: Updated the docs to explain this bit more.

> The current depth of the internal buffer of the sequence. The buffer contains all the yielded but not yet consumed elements.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+                    return .produceMore
+
+                case .returnStopProducing:
+                    return .stopProducing
+
+                case .returnDropped:
+                    return .dropped
+
+                case .resumeContinuationAndReturnProduceMore(let continuation, let element):
+                    // It is safe to resume the continuation while holding the lock
```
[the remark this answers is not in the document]
FranzBusch: This is somewhat documented on the `resume` method:

> After resume enqueues the task, control immediately returns to the caller. The task continues executing when its executor is able to reschedule it.

What this roughly translates to being is: `dispatchQueue.async {}`. It is safe to hold a lock across this, since the enqueued job will just wait on the next `lock.lock()`. Furthermore, we are releasing the lock right after resuming so there really shouldn't be too much waiting.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+    /// a ``NIOAsyncSequenceProducer``. The source MUST be held by the caller and
+    /// used to signal new elements or finish. The sequence MUST be passed to the actual subscriber and MUST NOT be held by the
+    /// caller. This is due to the fact that deiniting the sequence is used as part of a trigger to terminate the underlying source.
+    ///
+    /// - Parameters:
+    ///   - backPressureStrategy: The back-pressure strategy of the sequence.
+    ///   - delegate: The delegate of the sequence
+    /// - Returns: A ``NIOAsyncSequenceProducer/Source`` and a ``NIOAsyncSequenceProducer``.
+    @inlinable
+    public static func makeSourceAndSequence(
```
ktoso, line 142: The call sites are a bit rough but could be made simpler rather easily:

```
        let result = NIOAsyncSequenceProducer<
             Int,
             MockNIOElementStreamBackPressureStrategy,
             MockNIOBackPressuredStreamSourceDelegate
         >.makeSourceAndSequence(
             backPressureStrategy: self.backPressureStrategy,
             delegate: self.delegate
         )
```

via:

```
    @inlinable
     public static func makeSourceAndSequence(
        of elementType: Element.Type = Element.self, 
        backPressureStrategy: Strategy,
         delegate: Delegate
     ) -> NewSequence {
```

so it is spelled more nicer on callsites:

```
NIOAsyncSequenceProducer.makeSequence(
             of: Int.self
             backPressureStrategy: self.backPressureStrategy,
             delegate: self.delegate)
```

not sure the "AND" in `SourceAndSequence` explains all that much, I'd consider dropping... In general always worried when I see APIs mention "X AND Y".

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+        ///
+        /// If there are still buffered elements and there is an ``NIOAsyncSequenceProducer/AsyncIterator`` consuming the sequence,
+        /// then termination of the sequence only happens once all elements have been consumed by the ``NIOAsyncSequenceProducer/AsyncIterator``.
+        /// Otherwise, the buffered elements will be dropped.
+        ///
+        /// - Note: Calling this function more than once has no effect.
+        @inlinable
+        public func finish() {
+            self.storage.finish()
+        }
```
ktoso, line 290: Always unhappy about the "finish" wording... :-( but that's what AsyncStream sadly established; LGTM

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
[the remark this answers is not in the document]
ktoso: Yes, this is safe 👍

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+                }
+            }
+
+            delegate?.didTerminate()
+        }
+
+        @inlinable
+        /* fileprivate */ internal func next() async -> Element? {
+            // We are locking manually here since we want to hold the lock
+            // across the suspension point.
```
ktoso, line 437: this comment is a bit of a lie; you cannot hold across a suspension; what you're doing though is fine, since you unlock still in the SYNCHRONOUS piece of the withContinuation -- that is fine; it does not actually hold over the suspension point, due to how continuations work

ktoso: (best to remove this comment; the one below is correct tho)

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+
+            case .finished:
+                // We are already finished so there is nothing left to clean up.
+                // This is just the references dropping afterwards.
+                return .none
+
+            case .modifying:
+                preconditionFailure("Invalid state")
+            }
+        }
```
ktoso, line 698: like the state machine 👍

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+                // This means we can just transition to finished
+                self.state = .finished
+
+                return .callDidTerminate
+
+            case .streaming(_, let buffer, .some(let continuation), _, _):
+                // We have a continuation, this means our buffer must be empty
+                // Furthermore, we can now transition to finished
+                // and resume the continuation with `nil`
+                precondition(buffer.isEmpty)
```
ktoso, line 839: Please add a short message; those are so annoying to debug when they'd happen and you don't get a backtrace or even a message; always nicer to at least have a short string to grep for

Sources/NIOCore/AsyncSequences/NIOHighLowWatermarkBackPressuredAsyncSequenceStrategy.swift
ktoso: I don't hate the name... it's not going to be spelled much, but the nested version does read nicer! :-)

I'd propose `NIOAsyncSequenceProducer.HighLowWatermarkStrategy` - the "high low" bit really explains what this is; I would not be trimming it out of the name; though the nesting helps make it a bit easier to read, I'd +1 that

ktoso (approved):
Semantics look good! I like the delegate calls and the provided high/low strategy, looking good!

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+/// - Important: This sequence is a unicast sequence that only supports a single ``NIOAsyncSequenceProducer/AsyncIterator``.
+/// If you try to create more than one iterator it will result in a `fatalError`.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOAsyncSequenceProducer<
+    Element: Sendable,
+    Strategy: NIOAsyncSequenceProducerBackPressureStrategy,
+    Delegate: NIOAsyncSequenceProducerDelegate
+>: Sendable {
+    /// Simple struct for the return type of ``NIOAsyncSequenceProducer/makeSourceAndSequence(backPressureStrategy:delegate:)``.
+    public struct NewSequence {
```
ktoso, line 85: This should explain what the `sequence` and the `source` are for, add some paragraph about "use the `source` to yield values, and the `sequence` to consume them; the source can be freely shared with other parts of the program" or something like that

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+                iteratorInitialized: Bool
+            )
+
+            /// The state once there can be no outstanding demand. This can happen if:
+            /// 1. The ``NIOAsyncSequenceProducer/AsyncIterator`` was deinited
+            /// 2. The underlying source finished and all buffered elements have been consumed
+            case finished
+
+            /// Internal state to avoid CoW.
+            case modifying
```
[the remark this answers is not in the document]
FranzBusch: I quickly tried to create a reproducer for this and you are right that in a simple case the compiler is smart enough to avoid the CoW here. I have also quickly talked with @Lukasa and he explained me that in more complicated cases, especially if `@inlinable` methods are in play the compiler is not always able to optimise this away.

Since the code is already there and it helps the compiler, I would prefer to keep it.

ktoso:
By the way, 👏 for naming it a `Source` sounds very good to me 👍

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1064 @@
...
+    /* private */ internal let internalClass: InternalClass
+
+    @usableFromInline
+    /* private */ internal var storage: Storage {
+        self.internalClass.storage
+    }
+
+    /// Initializes a new ``NIOAsyncSequenceProducer`` and a ``NIOAsyncSequenceProducer/Source``.
+    ///
+    /// - Important: This method returns a tuple containing a ``NIOAsyncSequenceProducer/Source`` and
```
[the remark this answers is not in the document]
FranzBusch: This looks fine in DocC. AFAIK `-` and `>` have the same result.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
FranzBusch: Great suggestion! Changed it around, feels way better to create a sequence now.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
FranzBusch: Added some explanation on the properties of this type.

phausler:
>  Furthermore, we need the AsyncSequence to properly support back-pressure.

AsyncSequence does support back-pressure properly; each call to next is a back-pressure demand of 1. 

> It [AsyncSequence] only exposes back-pressure on the yield() edge and not on the next() edge

That is not true, it exposes that back-pressure by the aforementioned demand. What it does not express is a numeric buffer pressure translation to the yielded elements; which in general terms cannot be done safely (but perhaps in certain cases can be; like this is approaching).

FranzBusch:
> AsyncSequence does support back-pressure properly; each call to next is a back-pressure demand of 1.

What I meant with this is that the `AsyncSequence` implementation that we are coming up with here needs to expose the back-pressure to the producer.

> That is not true, it exposes that back-pressure by the aforementioned demand. What it does not express is a numeric buffer pressure translation to the yielded elements; which in general terms cannot be done safely (but perhaps in certain cases can be; like this is approaching).

`AsyncStream` is exposing two inits:
1. The continuation based init: When using this you can only yield new elements to the continuation but are not informed about consumption. This makes implementing a high/low back-pressure strategy for the underlying producer impossible.
2. The unfolding init: This one comes close, but it requires buffering outside of the `AsyncStream` and doing a thread jump on every call to `next()`.

In the end, the shape we created here is fulfilling all the goals we set out to achieve. Don't get me wrong, I still see use-cases for the other root `AsyncSequence` that exist in the ecosystem. They just didn't fit ours.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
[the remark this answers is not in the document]
ktoso: I see, thanks!

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -59,12 +64,12 @@ public protocol NIOAsyncSequenceProducerDelegate: Sendable {
...
     /// - The ``NIOAsyncSequenceProducer/AsyncIterator`` is deinited.
     /// - The ``NIOAsyncSequenceProducer`` deinited and no iterator is alive.
-    /// - The subscriber's `Task` is cancelled.
+    /// - The consuming `Task` is cancelled (e.g. `for await let element in`).
     /// - The source finished and all remaining buffered elements have been consumed.
     func didTerminate()
 }
 
-/// This is an ``AsyncSequence`` that supports a unicast ``AsyncIterator``.
+/// This is an ``Swift/AsyncSequence`` that supports a unicast ``Swift/AsyncIterator``.
```
ktoso, line 72: FYI: These will cause you docc warnings now; your call if you want to have them warn or just not do such links for now.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -83,7 +88,13 @@ public struct NIOAsyncSequenceProducer<
...
     Delegate: NIOAsyncSequenceProducerDelegate
 >: Sendable {
-    /// Simple struct for the return type of ``NIOAsyncSequenceProducer/makeSourceAndSequence(backPressureStrategy:delegate:)``.
+    /// Simple struct for the return type of ``NIOAsyncSequenceProducer/makeSequence(of:backPressureStrategy:delegate:)``.
+    ///
+    /// This struct contains two properties:
+    /// 1. The ``NIOAsyncSequenceProducer/NewSequence/source`` which should be retained by the producer and is used
+    /// to yield new elements to the sequence.
+    /// 2. The ``NIOAsyncSequenceProducer/NewSequence/sequence`` which is the actual ``Swift/AsyncSequence`` and
+    /// should be passed to the consumer.
```
ktoso, line 97: FYI: no need to fully qualify the names, you could just ``sequence`` and it'll resolve properly :)

Thanks for the docs here!

ktoso (approved):

Tests/NIOCoreTests/AsyncSequences/NIOBackPressuredAsyncSequenceTests.swift
```diff
@@ -0,0 +1,554 @@
+//===----------------------------------------------------------------------===//
```
glbrntt, line 1: Can we have some tests for the `HighLowWatermarkBackPressureStrategy`?

glbrntt (approved):
Nice work @FranzBusch! This all looks good but I think we should add a couple of tests for the `HighLowWatermarkBackPressureStrategy` before we merge this. As far as I could see it was only a mock strategy being tested.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
FranzBusch: I know, we have these already in various places of NIO docs.

FranzBusch:
@ktoso @glbrntt I moved the `HighLowWatermarkStrategy` outside of the `NIOAsyncSequenceProducer` because that made it basically impossible to use. `NIOAsyncSequenceProducer` is optional and trying to create a `HighLowWatermarkStrategy` always required to spell out the generic parameters which ended up in a recursive loop. So back to `NIOAsyncSequenceProducerHighLowWatermarkStrategy` we go :D

ktoso:
Haha right... well, thats fine then :)

FranzBusch:
The 5.5 test failure is due to a flakey test that this change did not impact:
> /code/Tests/NIOPosixTests/EventLoopTest.swift:827: error: EventLoopTest.testCancelledScheduledTasksDoNotHoldOnToRunClosure : failed - condition not satisfied in time

Lukasa:

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer+HighLowWatermarkBackPressureStrategy.swift
```diff
@@ -0,0 +1,46 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+/// A high-low watermarked back-pressure strategy for a ``NIOAsyncSequenceProducer``.
+///
+/// This strategy does the following:
+/// - On yield it keeps on demanding more as long as the `highWatermark` isn't reached.
+/// - On next it starts to demand once the the `lowWatermark` is reached.
```
Lukasa, line 20: Can we elaborate on this a little bit: can we say something like "as long as the number of buffered elements does not exceed `highWatermark`" and so on?

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer+HighLowWatermarkBackPressureStrategy.swift
```diff
@@ -0,0 +1,46 @@
...
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOAsyncSequenceProducerHighLowWatermarkBackPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategy {
```
Lukasa, line 22: I think we should have a namespace for these. Something like `enum NIOAsyncSequenceProducerStrategies`. That'll let us give this a better name.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer+HighLowWatermarkBackPressureStrategy.swift
```diff
@@ -0,0 +1,46 @@
...
+    private let highWatermark: Int
+
+    /// Initializes a new ``NIOAsyncSequenceProducerHighLowWatermarkBackPressureStrategy``.
+    ///
+    /// - Parameters:
+    ///   - lowWatermark: The low watermark where demand should start.
+    ///   - highWatermark: The high watermark where demand should be stopped.
+    public init(lowWatermark: Int, highWatermark: Int) {
+        self.lowWatermark = lowWatermark
+        self.highWatermark = highWatermark
```
Lukasa, line 33: Can we precondition that `highWatermark` is equal to or greater than `lowWatermark`?

Lukasa (approved):

Lukasa (changes requested):
Whoops, mis-clicked.

Lukasa (approved):

FranzBusch:
The main nightly build is still broken because of this issue: https://github.com/apple/swift/issues/60351. Since this already got a fixed in mainline Swift, I am going to merge over the failing build.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,1073 @@
...
+        }
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncSequenceProducer {
+    /// A struct to interface between the synchronous code of the producer and the asynchronous consumer.
+    /// This type allows the producer to synchronously `yield` new elements to the ``NIOAsyncSequenceProducer``
+    /// and to `finish` the sequence.
+    public struct Source {
```
fabianfett, line 238: Why is this a `struct`? That object has reference semantics, with sideeffects into other threads. I'm very much for using structs as much as we can, but to me this looks like 🐷💄 😅

PR #2231, opened 2022-08-01
Fix strict concurrency checking diagnostics in `NonBlockingFileIO`

dnadoba:
### Motivation:

We want to incrementally comply with the strict sendability checking rules.

### Modifications:
Move `ByteBuffer` allocation to thread modifying buffer by moving the mutable variable into the sendable and escaping closure 

### Result:

less warnings/errors in strict concurrency mode

### Alternatives considered
`ByteBuffer` is `Sendable` and we could therefore allocate a non-mutable buffer outside the closure, capture it by value and create a local mutable copy in the closure. However, I fear that we might trigger CoW as I'm not sure if the compiler eagerly release the initial buffer instance.

### DIFF
--- a/Sources/NIOPosix/NonBlockingFileIO.swift
+++ b/Sources/NIOPosix/NonBlockingFileIO.swift
@@ -442,8 +442,9 @@ public struct NonBlockingFileIO: NIOSendable {
             return eventLoop.makeSucceededFuture(allocator.buffer(capacity: 0))
         }
 
-        var buf = allocator.buffer(capacity: byteCount)
+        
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
Yeah, this is a straightforward cleanup.

Lukasa (changes requested):
Seems like this has caused issues in the alloc counter test. We need to take a look at this. There are two likely possibilities: either the test wasn't waiting for things to settle (not tearing down the background threads) or it was explicitly measuring allocations without letting the background threads run. In either case we should consider extending the test to avoid losing the allocations that we moved.

PR #2235, opened 2022-08-04
initial adoption of DocC based documentaiton

tomerd:
motivation: use DocC for docs

change:
* add DocC catalog to NIO target, as the "landing page"
* remove jazzy doc generation script
* add Package.swift with tools-version 5.6 so that DocC can be used, and add a dependency on the DocC plugin
* udpate soundness script

_


Package@swift-5.6.swift
```diff
@@ -0,0 +1,146 @@
+// swift-tools-version:5.6
```
tomerd, line 1: happy to also switch to the other style (file per "old" tools-version), lets finalize the discussion inhttps://github.com/apple/swift-statsd-client/pull/36#issuecomment-1205800362 and then apply everywhere

Sources/NIO/Docs.docc/index.md
```diff
@@ -0,0 +1,183 @@
...
+
+To achieve its goals SwiftNIO extensively uses "non-blocking I/O": hence the name! Non-blocking I/O differs from the more common blocking I/O model because the application does not wait for data to be sent to or received from the network: instead, SwiftNIO asks for the kernel to notify it when I/O operations can be performed without waiting.
+
+SwiftNIO does not aim to provide high-level solutions like, for example, web frameworks do. Instead, SwiftNIO is focused on providing the low-level building blocks for these higher-level applications. When it comes to building a web application, most users will not want to use SwiftNIO directly: instead, they'll want to use one of the many great web frameworks available in the Swift ecosystem. Those web frameworks, however, may choose to use SwiftNIO under the covers to provide their networking support.
+
+The following sections will describe the low-level tools that SwiftNIO provides, and provide a quick overview of how to work with them. If you feel comfortable with these concepts, then you can skip right ahead to the other sections of this README.
```
yim-lee, line 47:
```suggestion
The following sections will describe the low-level tools that SwiftNIO provides, and provide a quick overview of how to work with them. If you feel comfortable with these concepts, then you can skip right ahead to the other sections of this document.
```

Sources/NIO/Docs.docc/index.md
```diff
@@ -0,0 +1,183 @@
+# ``NIO`` <!-- SwiftNIO -->
```
yim-lee, line 1: Unrelated: should be doc be updated to cover async/await support?

tomerd: good idea,probably follow on PR or commits? this is all taken from the current readme

yim-lee: > probably follow on PR or commits?

yeah

Package@swift-5.6.swift
Lukasa, line 1: Can we use the style that the rest of the NIO repos use, where we have a Package@swift-5.4.swift and a Package@swift-5.5.swift? This lets us age out old package support really easily.

Package@swift-5.6.swift
```diff
@@ -0,0 +1,146 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import PackageDescription
+
+let swiftAtomics: PackageDescription.Target.Dependency = .product(name: "Atomics", package: "swift-atomics")
```
Lukasa, line 18: Relatedly, this is conflicting with #2230 which added a dependency on swift-collections, so you'll need to rebuild your Package.swift.

Lukasa (changes requested):
Thanks! Generally looks good, some notes in the diff.

Package@swift-5.6.swift
tomerd: @Lukasa ^^

Lukasa (approved):
Nice one, thanks!

Lukasa:
Flaky test: `ChannelTests.testConnectWithECONNREFUSEDGetsTheRightError`.

Lukasa:

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.4
+// swift-tools-version:5.6
```
buttaface, line 1: Strangely, this change now __requires__ that `--enable-test-discovery` be passed when compiling the tests on linux/Android, or [it errors like this with 5.6, 5.7, and trunk](https://github.com/buttaface/swift-android-sdk/runs/7742640026?check_suite_focus=true#step:13:496):
```
swift-nio/Tests/LinuxMain.swift:146:1: error: expressions are not allowed at the top level
(LinuxMainRunnerImpl() as LinuxMainRunner).run()
```
Anyone know what's going on?

Lukasa: I believe this is a bug in SwiftPM.

buttaface: Alright, I'll file a bug for SPM.

0xTim: @Lukasa Now that NIO supports 5.4 as a minimum can LinuxMain.swift not just be removed?

Lukasa: Sadly no. NIO tests with `--warnings-as-errors`, and we have number of tests for deprecated code. The test discovery stubs in SwiftPM generate warnings when those tests are present up until 5.6, which is why our CI continues to use `LinuxMain.swift` for 5.4 and 5.5. But we are at least _finally_ on a path to dropping it.

buttaface: Filed, apple/swift-package-manager#5717.

tomerd: this _should_ be fixed in 5.7 and main. but perhaps not available in the nightly toolchain yet?

buttaface: It is fixed with the latest Aug. 9 trunk snapshot build for linux, but I'm using a June trunk snapshot because of another breaking regression for me. As you noted in that issue, should only be a problem with 5.6 now.

tomerd: yea will be hard to justify pulling into 5.6 given workaround.

buttaface: Which workaround? Adding @main is a non-starter if someone is testing 5.3 and earlier, when the attribute wasn't there, apple/swift@df99de8. As for overriding it with `--enable-test-discovery`, adding that only for 5.6.3 for all future CI testing is going to be annoying and has other consequences.

PR #2237, opened 2022-08-08
Add throwing version of `NIOAsyncSequenceProducer`

FranzBusch:
# Motivation
We recently introduced a `NIOAsyncSequenceProducer` to bridge a stream of elements from the NIO world into the async world. The introduced type was a non-throwing `AsyncSequence`. To support all use-cases we also need to offer a throwing variant of the type.

# Modification
- Introduce a new `NIOThrowingAsyncSequenceProducer` that is identical to the `NIOAsyncSequenceProducer` except that it has a `Failure` generic parameter and that the `next()` method is throwing.
- Extract the `StateMachine` from both `AsyncSequenceProducer`s and unify them.
- There is one modification in behaviour: `didTerminate` is now only called after `nil` or the error has been consumed from the sequence.

# Result
We now have a throwing variant of the `NIOAsyncSequenceProducer`.


Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
```diff
@@ -288,6 +288,10 @@ final class NIOAsyncSequenceProducerTests: XCTestCase {
     func testFinish_whenFinished() async throws {
         self.source.finish()
 
+        XCTAssertEqual(self.delegate.didTerminateCallCount, 0)
+
+        _ = await self.sequence.first { _ in true }
+
```
FranzBusch, line 294: @Lukasa @glbrntt This is a slight change in behaviour where I am now calling `didTerminate` only after the `nil` or thrown `Error` has been consumed. I was going back and forth about this but thought this was the best way. Before we already called `didTerminate` once the last element has been consumed. WDYT?

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,503 @@
...
+            self.storage.finish(failure)
+        }
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOThrowingAsyncSequenceProducer {
+    /// This is the underlying storage of the sequence. The goal of this is to synchronize the access to all state.
+    @usableFromInline
+    /* fileprivate */ internal final class Storage: @unchecked Sendable {
```
FranzBusch, line 279: @Lukasa This type could also be unified between the two implementations. I didn't do that yet because the return type of `yield() -> Source.YieldResult` is currently depending on the type. There are two ways forward:

1. Not unify this type
2. Move the `YieldResult` outside of the generic types and unify the `Storage` type.

Let me know what you prefer!

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -0,0 +1,503 @@
...
+
+                delegate?.didTerminate()
+            }
+        }
+    }
+}
+
+/// The ``NIOThrowingAsyncSequenceProducer/AsyncIterator`` MUST NOT be shared across `Task`s. With marking this as
+/// unavailable we are explicitly declaring this.
+@available(*, unavailable)
```
glbrntt, line 501: Don't forget the `@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)`

Lukasa (changes requested):
This change has the effect of duplicating nearly 500 lines of code. Did you consider the idea of having _only_ the throwing version, and having the non-throwing version be a simple wrapper that uses `try!` on the `next()` function as well as a non-failing set of `yield`s?

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -158,93 +137,94 @@ public struct NIOAsyncSequenceProducer<
...
+            >.AsyncIterator
+        ) {
+            self.throwingIterator = throwingIterator
         }
 
         @inlinable
         public func next() async -> Element? {
-            await self.internalClass.next()
+            do {
+                return try await self.throwingIterator.next()
```
Lukasa, line 192: I think this can safely be `try!`.

Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
Lukasa: I think this is fine.

FranzBusch:

Lukasa (approved):

PR #2238, opened 2022-08-08
Provide NIOAsyncTestingChannel

Lukasa:
Motivation

Testing versions of NIO code that involve interfacing with Swift
Concurrency is currently a difficult business. In particular,
EmbeddedChannel is not available in Swift concurrency, making it
difficult to write tests where you fully control the I/O.

To that end, we should provide a variation of EmbeddedChannel that makes
testing these things possible.

Modifications

Provide an implementation of NIOAsyncTestingChannel.

Results

Users can write tests confidently with async/await.


Lukasa:
cc @0xTim, this should help with the WebSocket concerns you had.

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -0,0 +1,508 @@
...
+    ///
+    /// `NIOAsyncTestingChannel` is most useful in testing and usually in unit tests, you want to consume all inbound and
+    /// outbound data to verify they are what you expect. Therefore, when you `finish` an `NIOAsyncTestingChannel` it will
+    /// return if it's either `.clean` (no left overs) or that it has `.leftOvers`.
+    public enum LeftOverState {
+        /// The `NIOAsyncTestingChannel` is clean, ie. no inbound, outbound, or pending outbound data left on `finish`.
+        case clean
+
+        /// The `NIOAsyncTestingChannel` has inbound, outbound, or pending outbound data left on `finish`.
+        case leftOvers(inbound: CircularBuffer<NIOAny>, outbound: CircularBuffer<NIOAny>, pendingOutbound: [NIOAny])
```
glbrntt, line 52: Do we want to continue using `CircularBuffer` in new public API or should we start switching to `Deque` where possible? 

(I know these come from the channel core and we probably don't want to incur the additional allocation cost from copying them over but I'm still interested in the answer to the general case)

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -0,0 +1,508 @@
...
+        try! await self._pipeline.addHandlers(handlers)
+
+        // This will never throw...
+        try! await self.register()
+    }
+
+    /// - see: `Channel.setOption`
+    @inlinable
+    public func setOption<Option: ChannelOption>(_ option: Option, value: Option.Value) -> EventLoopFuture<Void> {
+        if self.eventLoop.inEventLoop {
```
glbrntt, line 410: Should we just copy the implementation of `setOptionSync` and `fatalError` here?

0xTim:
Great once this is merged I'll try and get my Websocket PR working

glbrntt (commented):
Looks good modulo a couple of small things.

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: Yeah that's a good question. In this instance we should leave it due to the existing use in the embedded channel, but I think that we should be using `Deque` in new API if we don't have a good reason not to.

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: I'm inclined to leave this as is: it's a common pattern.

Lukasa:
Hmm, we've regressed allocations in `EmbeddedChannel` due to the extra atomics. Let me see if I can tweak the numbers a bit.

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -0,0 +1,521 @@
...
+/// event. The user is responsible for making sure the first
+/// `ChannelInboundHandler` expects data of that type.
+///
+/// `NIOAsyncTestingChannel` automatically collects arriving outbound data and makes it
+/// available one-by-one through `readOutbound`.
+///
+/// - note: `NIOAsyncTestingChannel` is currently only compatible with
+///   `NIOAsyncTestingEventLoop`s and cannot be used with `SelectableEventLoop`s from
+///   for example `MultiThreadedEventLoopGroup`.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
FranzBusch, line 39: Maybe we can add a little diagram here to visualise the flow of data, just a quick draw:
```suggestion
///
///                            readOutbound
///                                 │
///                                 │
///                                 ▼
///                          ┌──────────────┐
///            writeInbound  │OutboundBuffer│
///                          └──────────────┘
///                 │               ▲
///                 │               │
///                 ▼               │
///             ┌────────────────────────┐
///             │ NIOAsyncTestingChannel │
///             └────────────────────────┘
///                 │                ▲
///                 │                │
/// fireChannelRead │                │  write & flush
///                 │                │
///                 ▼                │
///             ┌────────────────────────┐
///             │     ChannelHandler     │
///             └────────────────────────┘
///                 │                ▲
/// fireChannelRead │                │
///                 ▼                │
///           ┌─────────────┐
///           │InboundBuffer│  writeOutbound
///           └─────────────┘
///                 ▲
///                 │
///                 │
///            readInbound
///
@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -0,0 +1,521 @@
...
+        try await self.writeAndFlush(NIOAny(data))
+
+        return try await self.testingEventLoop.executeInContext {
+            return self.channelcore.outboundBuffer.isEmpty ? .empty : .full(self.channelcore.outboundBuffer)
+        }
+    }
+
+    /// This method will throw the error that is stored in the `NIOAsyncTestingChannel` if any.
+    ///
+    /// The `NIOAsyncTestingChannel` will store an error if some error travels the `ChannelPipeline` all the way past its end.
```
FranzBusch, line 338: Can we specify what end of the pipeline?

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -0,0 +1,521 @@
...
+        return t
+    }
+
+    /// Create a new instance.
+    ///
+    /// During creation it will automatically also register itself on the `NIOAsyncTestingEventLoop`.
+    ///
+    /// - parameters:
+    ///     - loop: The `NIOAsyncTestingEventLoop` to use.
+    public init(loop: NIOAsyncTestingEventLoop = NIOAsyncTestingEventLoop()) {
```
FranzBusch, line 374: NIT: Could we move all the inits above the methods?

FranzBusch (approved):
Great stuff! I think this will make async testing way better. Just left a couple of smaller comments inline.

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: Errors can only go to one end of the pipeline.

FranzBusch (approved):

PR #2242, opened 2022-08-12
Add a soundness plugin

glbrntt:
Motivation:

We have a handful of scripts which are cargo-culted across the NIO
repos. This includes the soundness script for checking license headers
and unacceptable language. As regex was added to Swift 5.7 we can use it
to replace the bash in the existing script and provide a product which
can be used in the other NIO repos.

Modifications:

- Add a 'nio-soundness' plugin which can check for unacceptable
  language and valid license headers.
- The CLI is hand rolled to avoid bringing in a dependency on
  swift-argument-parser.
- Update the soundness script to use the new tool.
- Update the soundness script to use zsh instead of bash as the bash
  shipped on macOS is old and doesn't support globbing '**'.

Result:

Less bash, more Swift.


glbrntt:
We'll have to update our soundness job to use 5.7 once it's release if we want to use this.

ktoso:
> It will also need zsh...

I mean... I love zsh as any other person; but it's not a great idea to require it for scripts like this, why the requirement?

ktoso:
Very nice by the way, would love to share this implementation and not have to copy paste the plugin into all other repos...
Could we perhaps make this a thing in swift-server?

scripts/soundness.sh
```diff
@@ -34,140 +30,24 @@ else
...
-        diff -u <(cat "$line" | replace_acceptable_years | head -n $expected_lines) "$tmp"
-        exit 1
-      fi
-    done
-    printf "\033[0;32mokay.\033[0m\n"
-  )
-done
-
-rm "$tmp"
+swift package plugin check-nio-soundness check-language \
```
ktoso, line 33: FYI you don't need the "plugin", can be `swift package check-nio-soundness`

I'd love to make this a plugin we can share within our projects tbh, lots of projects, not just nio, use those scripts -- we could make it `swift package soundness check-language` etc

scripts/soundness.sh
```diff
@@ -34,140 +30,24 @@ else
...
+  --exclude-file ${root}/CODE_OF_CONDUCT.md \
+  ${root}/**/*
```
ktoso, line 35: Is that the piece you need `zsh` for? Worth doing a small `find` dance and keeping the bash compatibility IMHO, keeping CI requirements simpler always worth it 🤔 (also for contributors which may or may not have zsh)

glbrntt: `zsh` has been default on macOS for a couple of years so I don't know how big a deal using `zsh` is. You're right that doing a `find` dance would probably be the safer bet here.

ktoso: Sure, but CIs and "random contributors" may very well be on Linux with default bash shells.

Plugins/NIOSoundness/Soundness.swift
```diff
@@ -0,0 +1,251 @@
...
+#if swift(>=5.7)
+import PackagePlugin
+import Foundation
+
+@main
+@available(macOS 13.0, *)
+struct Soundness: CommandPlugin {
+    func performCommand(context: PackagePlugin.PluginContext, arguments: [String]) async throws {
+        // Check for help first.
+        if arguments.contains("--help") {
```
tomerd, line 23: SwiftPM plugin APIs include a simple argument parser that can help make these tasks easier. cc @abertelrud

scripts/soundness.sh
ktoso: Oh nice, I also totally missed that... thanks for pointing it out :)

PR #2244, opened 2022-08-14
Implement additional file operation in NonBlockingFileIO

artemredkin:
Motivation:
Basic set of function that NonBlockingFileIO provides is enough to read and write files, but as we get more and more into fully asynchronous world we need other non-blocking file-related function.

Modifications:
 - Adds support for getting file information using `lstat`
 - Adds ability to create a symlink, delete symlink and read it's destination
 - Adds ability to list directories
 - Adds ability to rename and remove files

### DIFF
--- a/Sources/NIOPosix/NonBlockingFileIO.swift
+++ b/Sources/NIOPosix/NonBlockingFileIO.swift
@@ -627,4 +627,134 @@ public struct NonBlockingFileIO: Sendable {
         }
     }
 
+#if !os(Windows)
+    /// Returns information about a file at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - note: If `path` is a symlink, information about the link, not the file it points to.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to get information about.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing file information.
+    public func lstat(path: String, eventLoop: EventLoop) -> EventLoopFuture<stat> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            var s = stat()
+            try Posix.lstat(pathname: path, outStat: &s)
+            return s
+        }
+    }
+
+    /// Creates a symbolic link to a  `destination` file  at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the link.
+    ///     - destination: Target path where this link will point to.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the rename was successful or fails on error.
+    public func symlink(path: String, to destination: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.symlink(pathname: path, destination: destination)
+        }
+    }
+
+    /// Returns target of the symbolic link at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the link to read.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing link target.
+    public func readlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<String> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let maxLength = Int(PATH_MAX)
+            let pointer = UnsafeMutableBufferPointer<CChar>.allocate(capacity: maxLength)
+            defer {
+                pointer.deallocate()
+            }
+            let length = try Posix.readlink(pathname: path, outPath: pointer.baseAddress!, outPathSize: maxLength)
+            return String(decoding: UnsafeRawBufferPointer(pointer).prefix(length), as: UTF8.self)
+        }
+    }
+
+    /// Removes symbolic link at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the link to remove.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the rename was successful or fails on error.
+    public func unlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.unlink(pathname: path)
+        }
+    }
+
+    /// List contents of the directory at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the directory to list the content of.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing the directory entries.
+    public func listDirectory(path: String, eventLoop: EventLoop) -> EventLoopFuture<[NIODirectoryEntry]> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let dir = try Posix.opendir(pathname: path)
+            var entries: [NIODirectoryEntry] = []
+            do {
+                while let entry = try Posix.readdir(dir: dir) {
+                    let name = withUnsafeBytes(of: entry.pointee.d_name) { pointer -> String in
+                        let ptr = pointer.baseAddress!.assumingMemoryBound(to: CChar.self)
+                        return String(cString: ptr)
+                    }
+                    entries.append(NIODirectoryEntry(ino: UInt64(entry.pointee.d_ino), type: entry.pointee.d_type, name: name))
+                }
+                try? Posix.closedir(dir: dir)
+            } catch {
+                try? Posix.closedir(dir: dir)
+                throw error
+            }
+            return entries
+        }
+    }
+
+    /// Renames the file at `path` to `newName` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be renamed.
+    ///     - newName: New file name.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the rename was successful or fails on error.
+    public func rename(path: String, newName: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.rename(pathname: path, newName: newName)
+        }
+    }
+
+    /// Removes the file at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be removed.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the remove was successful or fails on error.
+    public func remove(path: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.remove(pathname: path)
+        }
+    }
+#endif
 }
+
+#if !os(Windows)
+/// A `NIODirectoryEntry` represents a single directory entry.
+public struct NIODirectoryEntry: Hashable {
+    // File number of entry
+    public var ino: UInt64
+    // File type
+    public var type: UInt8
+    // File name
+    public var name: String
+
+    public init(ino: UInt64, type: UInt8, name: String) {
+        self.ino = ino
+        self.type = type
+        self.name = name
+    }
+}
+#endif
--- a/Sources/NIOPosix/System.swift
+++ b/Sources/NIOPosix/System.swift
@@ -110,11 +110,27 @@ private let sysSocketpair: @convention(c) (CInt, CInt, CInt, UnsafeMutablePointe
 #if os(Linux)
 private let sysFstat: @convention(c) (CInt, UnsafeMutablePointer<stat>) -> CInt = fstat
 private let sysStat: @convention(c) (UnsafePointer<CChar>, UnsafeMutablePointer<stat>) -> CInt = stat
+private let sysLstat: @convention(c) (UnsafePointer<CChar>, UnsafeMutablePointer<stat>) -> CInt = lstat
+private let sysSymlink: @convention(c) (UnsafePointer<CChar>, UnsafePointer<CChar>) -> CInt = symlink
+private let sysReadlink: @convention(c) (UnsafePointer<CChar>, UnsafeMutablePointer<CChar>, Int) -> CLong = readlink
 private let sysUnlink: @convention(c) (UnsafePointer<CChar>) -> CInt = unlink
+private let sysOpendir: @convention(c) (UnsafePointer<CChar>) -> OpaquePointer? = opendir
+private let sysReaddir: @convention(c) (OpaquePointer) -> UnsafeMutablePointer<dirent>? = readdir
+private let sysClosedir: @convention(c) (OpaquePointer) -> CInt = closedir
+private let sysRename: @convention(c) (UnsafePointer<CChar>?, UnsafePointer<CChar>?) -> CInt = rename
+private let sysRemove: @convention(c) (UnsafePointer<CChar>?) -> CInt = remove
 #elseif os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(Android)
 private let sysFstat: @convention(c) (CInt, UnsafeMutablePointer<stat>?) -> CInt = fstat
 private let sysStat: @convention(c) (UnsafePointer<CChar>?, UnsafeMutablePointer<stat>?) -> CInt = stat
+private let sysLstat: @convention(c) (UnsafePointer<CChar>?, UnsafeMutablePointer<stat>?) -> CInt = lstat
+private let sysSymlink: @convention(c) (UnsafePointer<CChar>?, UnsafePointer<CChar>?) -> CInt = symlink
+private let sysReadlink: @convention(c) (UnsafePointer<CChar>?, UnsafeMutablePointer<CChar>?, Int) -> CLong = readlink
 private let sysUnlink: @convention(c) (UnsafePointer<CChar>?) -> CInt = unlink
+private let sysOpendir: @convention(c) (UnsafePointer<CChar>?) -> UnsafeMutablePointer<DIR>? = opendir
+private let sysReaddir: @convention(c) (UnsafeMutablePointer<DIR>?) -> UnsafeMutablePointer<dirent>? = readdir
+private let sysClosedir: @convention(c) (UnsafeMutablePointer<DIR>?) -> CInt = closedir
+private let sysRename: @convention(c) (UnsafePointer<CChar>?, UnsafePointer<CChar>?) -> CInt = rename
+private let sysRemove: @convention(c) (UnsafePointer<CChar>?) -> CInt = remove
 #endif
 #if os(Linux) || os(Android)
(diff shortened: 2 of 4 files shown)


Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,84 @@ public struct NonBlockingFileIO: NIOSendable {
         }
     }
 
+    public func lstat(path: String, eventLoop: EventLoop) -> EventLoopFuture<stat> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let pointer = UnsafeMutablePointer<stat>.allocate(capacity: 1)
```
Lukasa, line 632: No need for the stack allocated pointer, you can just do `var s = stat()` and then pass it in using `&`.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,84 @@ public struct NonBlockingFileIO: NIOSendable {
...
+    public func symlink(path: String, to destination: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.symlink(pathname: path, destination: destination)
+        }
+    }
+
+    public func readlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<String> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let maxLength = 1026
+            let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: maxLength)
```
Lukasa, line 650: Prefer `String(unsafeUninitializedCapacity:initializingWith:)`

Lukasa (commented):
A few quick comments as I was spinning through.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,84 @@ public struct NonBlockingFileIO: NIOSendable {
...
+
+    public func symlink(path: String, to destination: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.symlink(pathname: path, destination: destination)
+        }
+    }
+
+    public func readlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<String> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let maxLength = 1026
```
dnadoba, line 649: If `1026` is on purpose, can you document why exactly this length

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: we don't get a number a number bytes in the resulting pointer, we just expect that it will be `\0`-terminated

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: replaced with `MAX_PATH`

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: 🙏

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,150 @@ public struct NonBlockingFileIO: NIOSendable {
...
+    }
+
+    /// Creates a symbolic link to a  `destination` file  at `path` on a private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// - parameters:
+    ///     - path: The path of the link.
+    ///     - destination: Target path where this link will point to.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the rename was successful or fails on error.
+    public func symlink(path: String, to destination: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
```
artemredkin, line 654: I wonder if we should stick with close-to-syscall naming or go with foundation-esque naming? `createSymbolicLink`, `attributesOfAnItem` or similar...

Lukasa: I'm inclined for us to stick to close-to-syscall naming: this is NIOPosix, after all, we aren't pretending to cross-platform compatibility here.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,150 @@ public struct NonBlockingFileIO: Sendable {
...
+    /// - returns: An `EventLoopFuture` containing the file list.
+    public func listDirectory(path: String, eventLoop: EventLoop) -> EventLoopFuture<[String]> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let dir = try Posix.opendir(pathname: path)
+            var entries: [String] = []
+            do {
+                while let entry = try Posix.readdir(dir: dir) {
+                    let name = withUnsafeBytes(of: entry.pointee.d_name) { pointer -> String in
+                        let ptr = pointer.baseAddress!.assumingMemoryBound(to: CChar.self)
+                        return String(cString: ptr)
```
Lukasa, line 739: Prefer using buffer pointers and `d_namlen`.

Lukasa: Relatedly, should this be slightly richer? Should we include the type info from `dirent`?

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+    @inline(never)
+    public static func unlink(pathname: String) throws {
         _ = try syscall(blocking: false) {
             sysUnlink(pathname)
         }
     }
 
+    @inline(never)
+    public static func mkdir(pathname: String, mode: mode_t) throws {
+        _ = try syscall(blocking: true) {
```
Lukasa, line 695: Blocking should be `false`, this cannot return `EAGAIN` or `EWOULDBLOCK`.

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+            sysMkdir(pathname, mode)
+        }
+    }
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    @inline(never)
+    public static func mkpath(pathname: String, mode: mode_t) throws {
+        _ = try syscall(blocking: true) {
```
Lukasa, line 703: Blocking should be `false`, this cannot return `EAGAIN` or `EWOULDBLOCK`.

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+            sysMkpath(pathname, mode)
+        }
+    }
+#endif
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+    @inline(never)
+    public static func opendir(pathname: String) throws -> UnsafeMutablePointer<DIR> {
+        guard let result = sysOpendir(pathname) else {
```
Lukasa, line 712: This is missing the retry on `EINTR` logic. Can I propose adding a syscall helper for functions that return pointers?

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+            let err = errno
+            throw IOError(errnoCode: err, reason: #function)
+        }
+        return result
+    }
+
+    @inline(never)
+    public static func readdir(dir: UnsafeMutablePointer<DIR>) throws -> UnsafeMutablePointer<dirent>? {
+        guard let result = sysReaddir(dir) else {
```
Lukasa, line 721: Same note here on `EINTR`.

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+    @inline(never)
+    public static func closedir(dir: UnsafeMutablePointer<DIR>) throws {
+        _ = try syscall(blocking: true) {
+            sysClosedir(dir)
+        }
+    }
+#elseif os(Linux) || os(FreeBSD) || os(Android)
+    @inline(never)
+    public static func opendir(pathname: String) throws -> OpaquePointer {
+        guard let result = sysOpendir(pathname) else {
```
Lukasa, line 740: Same note here.

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+            let err = errno
+            throw IOError(errnoCode: err, reason: #function)
+        }
+        return result
+    }
+
+    @inline(never)
+    public static func readdir(dir: OpaquePointer) throws -> UnsafeMutablePointer<dirent>? {
+        guard let result = sysReaddir(dir) else {
```
Lukasa, line 749: Same note here.

Lukasa (changes requested):

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,150 @@ public struct NonBlockingFileIO: Sendable {
...
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            if createIntermediates {
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+                try Posix.mkpath(pathname: path, mode: mode)
+#else
+                // ensure that intermediate directories exist
+                let startIndex = path.startIndex
+                var endIndex = path.index(after: startIndex)
+                while endIndex != path.endIndex {
+                    if path[endIndex] == "/" {
```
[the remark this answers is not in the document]
artemredkin: would you be open to having a darwin-specific call that has `withIntermediateDirectories`? since we have a helper for it provided by the OS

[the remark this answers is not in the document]
Lukasa: Maybe let's just kick this out of this PR entirely and do a separate PR entirely focused on that function, so we can get the implementation right.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,150 @@ public struct NonBlockingFileIO: Sendable {
...
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` containing link target.
+    public func readlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<String> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let maxLength = Int(PATH_MAX)
+            let pointer = UnsafeMutablePointer<CChar>.allocate(capacity: maxLength)
+            defer {
+                pointer.deallocate()
+            }
+            try Posix.readlink(pathname: path, outPath: pointer, outPathSize: maxLength)
```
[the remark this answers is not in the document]
artemredkin: done, thanks!

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
+    @inline(never)
+    public static func symlink(pathname: String, destination: String) throws {
+        _ = try syscall(blocking: true) {
+            sysSymlink(destination, pathname)
+        }
+    }
+
+    @inline(never)
+    public static func readlink(pathname: String, outPath: UnsafeMutablePointer<CChar>, outPathSize: Int) throws {
+        _ = try syscall(blocking: true) {
```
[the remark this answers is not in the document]
artemredkin: done, thanks!

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +663,121 @@ internal enum Posix {
...
-    internal static func unlink(pathname: String) throws {
+    internal static func lstat(pathname: String, outStat: UnsafeMutablePointer<stat>) throws {
+        _ = try syscall(blocking: false) {
+            sysLstat(pathname, outStat)
+        }
+    }
+
+    @inline(never)
+    public static func symlink(pathname: String, destination: String) throws {
+        _ = try syscall(blocking: true) {
```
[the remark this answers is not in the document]
artemredkin: done, thanks!

Sources/NIOPosix/NonBlockingFileIO.swift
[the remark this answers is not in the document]
artemredkin: agreed, removed all mkdir-related stuff

Sources/NIOPosix/System.swift
artemredkin: removed for now

Sources/NIOPosix/System.swift
artemredkin: removed for now

Sources/NIOPosix/System.swift
artemredkin: done, thanks!

Sources/NIOPosix/System.swift
artemredkin: done, thanks!

Sources/NIOPosix/System.swift
artemredkin: done, thanks!

Sources/NIOPosix/System.swift
artemredkin: done, thanks!

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: on linux it's a null-terminated string if I'm not mistaken, for darwin added using a buffer pointer, plus added a struct with 2 additional fields instead of just string

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,132 @@ public struct NonBlockingFileIO: Sendable {
...
+    ///     - path: The path of the link to remove.
+    ///     - eventLoop: The `EventLoop` on which the returned `EventLoopFuture` will fire.
+    /// - returns: An `EventLoopFuture` which is fulfilled if the rename was successful or fails on error.
+    public func unlink(path: String, eventLoop: EventLoop) -> EventLoopFuture<Void> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.unlink(pathname: path)
+        }
+    }
+
+    public struct DirectoryEntry {
```
Lukasa, line 690: This type needs a doc comment, both on the type and on the properties. All the properties need to be public, we need a public memberwise initializer, and this should probably be `Hashable` too.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,132 @@ public struct NonBlockingFileIO: Sendable {
...
+    public func listDirectory(path: String, eventLoop: EventLoop) -> EventLoopFuture<[DirectoryEntry]> {
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            let dir = try Posix.opendir(pathname: path)
+            var entries: [DirectoryEntry] = []
+            do {
+                while let entry = try Posix.readdir(dir: dir) {
+                    let name = withUnsafeBytes(of: entry.pointee.d_name) { pointer -> String in
+                        // Per man readdir(3):
+                        // struct dirent {
+                        //  char d_name[256]; /* Null-terminated filename */
```
Lukasa, line 714: Can you move this comment into the `os(Linux)` block and then an equivalent manual page comment from macOS into the other block?

Lukasa: Relatedly @buttaface, do you know if Android behaves more like macOS or more like Linux here?

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +740,91 @@ internal enum Posix {
...
+    public static func opendir(pathname: String) throws -> UnsafeMutablePointer<DIR> {
+        return try syscall {
+            sysOpendir(pathname)
+        }
+    }
+
+    @inline(never)
+    public static func readdir(dir: UnsafeMutablePointer<DIR>) throws -> UnsafeMutablePointer<dirent>? {
+        return try syscallOptional {
+            sysReaddir(dir)
```
Lukasa, line 781: Can we set `errno = 0` before this call?

Sources/NIOPosix/System.swift
```diff
@@ -644,12 +740,91 @@ internal enum Posix {
...
+    public static func opendir(pathname: String) throws -> OpaquePointer {
+        return try syscall {
+            sysOpendir(pathname)
+        }
+    }
+
+    @inline(never)
+    public static func readdir(dir: OpaquePointer) throws -> UnsafeMutablePointer<dirent>? {
+        return try syscallOptional {
+            sysReaddir(dir)
```
Lukasa, line 802: Similarly, can we set `errno = 0` here before this call?

Lukasa (changes requested):

Sources/NIOPosix/NonBlockingFileIO.swift
buttaface: It appears to [behave the same as linux](https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/include/dirent.h#71), so please add an Android conditional here too.

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: Should this be namespaced or a top-level type? wdyt?

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: should we just use `d_namlen` in all cases instead?

artemredkin: actually, it seems that `d_name` is null-terminated on darwin too:
```
A directory entry has a struct dirent at the front of it, containing its
inode number, the length of the entry, and the length of the name
contained in the entry.  These are followed by the name padded to a 4
byte boundary with null bytes.  All names are guaranteed null terminated.
```

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: should `listDirectory` be `readdir` then? Although we do have `openFile` and not `open`...

Sources/NIOPosix/System.swift
artemredkin: added this to `syscallOptional `

Sources/NIOPosix/System.swift
artemredkin: added this to `syscallOptional `

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: converted to read null-terminated string in all cases

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: added doc comment, hashable conformance and public memberwise init, made fields public and moved outside of `NonBlockingFileIO` type for now

Sources/NIOPosix/System.swift
```diff
@@ -249,6 +265,91 @@ internal func syscall<T: FixedWidthInteger>(blocking: Bool,
     }
 }
 
+@inline(__always)
+@discardableResult
+internal func syscall<T>(where function: String = #function,
```
artemredkin, line 270: this helper and `syscall` variant that operates on the `OpaquePointer` seem to be platform-specific and used once (at least for now), do you think it's worth inlining them or making them plaftorm-specific?

Lukasa: I think it's fine to keep them separate. It makes it easier to follow the general pattern.

Sources/NIOPosix/NonBlockingFileIO.swift
Lukasa: I'm open to leaving this as-is.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -627,4 +627,134 @@ public struct NonBlockingFileIO: Sendable {
...
+        return self.threadPool.runIfActive(eventLoop: eventLoop) {
+            try Posix.remove(pathname: path)
+        }
+    }
+#endif
+}
+
+#if !os(Windows)
+/// A `DirectoryEntry` represents a single directory entry.
+public struct DirectoryEntry: Hashable {
```
Lukasa, line 746: This needs to be namespaced with `NIO` in front of it.

Lukasa (approved):

Sources/NIOPosix/NonBlockingFileIO.swift
artemredkin: fixed, thanks!

PR #2249, opened 2022-08-24
Functions passed to non-`Sendable` `ChannelHandler`s do not need to conform to `Sendable`

dnadoba:
### Motivation
A non-`Sendable` `ChannelHandler` can *not* be transfered to a different isolation domain. If it wants to be added to a `Channel` it need to be initialized on the `EventLoop` of the `Channel`. The functions modified in this PR are guaranteed to be called on the `EventLoop` of their `Channel`. They therefore do not need to be `Sendable`.

### Modification
Remove `Sendable` conformance requirement for functions passed to non-`Sendable` `ChannelHandler`s

### Result
Correct semantics and less code duplication. 

This is an API stable change because we only relax requirements for types passed into `ChannelHandler`s and they are luckily not expose as properties.


Lukasa (approved):

dnadoba:
Nightly allocation tests fail because of a compiler crash unrelated to this change: https://github.com/apple/swift/issues/60797

PR #2251, opened 2022-08-26
Implement a `NIOAsyncWriter`

FranzBusch:
# Motivation
We previously added the `NIOAsyncSequenceProducer` to bridge between the NIO channel pipeline and the asynchronous world. However, we still need something to bridge writes from the asynchronous world back to the NIO channel pipeline.

# Modification
This PR adds a new `NIOAsyncWriter` type that allows us to asynchronously `yield` elements to it. On the other side, we can register a `NIOAsyncWriterDelegate` which will get informed about any written elements. Furthermore, the synchronous side can toggle the writability of the `AsyncWriter` which allows it to implement flow control.
A main goal of this type is to be as performant as possible. To achieve this I did the following things:
- Make everything generic and inlinable
- Use a class with a lock instead of an actor
- Provide methods to yield a sequence of things which allows users to reduce the amount of times the lock gets acquired.

# Result
We now have the means to bridge writes from the asynchronous world to the synchronous

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -274,22 +274,22 @@ extension NIOAsyncWriter {
 
         @inlinable
         /* fileprivate */ internal func writerDeinitialized() {
-            let _delegate: Delegate? = self._lock.withLock {
+            let delegate: Delegate? = self._lock.withLock {
                 let action = self._stateMachine.writerDeinitialized()
 
                 switch action {
                 case .callDidTerminate:
-                    let _delegate = self._delegate
+                    let delegate = self._delegate
                     self._delegate = nil
 
-                    return _delegate
+                    return delegate
 
                 case .none:
                     return nil
                 }
             }
 
-            _delegate?.didTerminate(failure: nil)
+            delegate?.didTerminate(failure: nil)
         }
 
         @inlinable
@@ -301,10 +301,10 @@ extension NIOAsyncWriter {
 
             switch action {
             case .callDidYieldAndResumeContinuations(let suspendedYields):
-                let _delegate = self._delegate
+                let delegate = self._delegate
                 self._lock.unlock()
 
-                _delegate?.didYield(contentsOf: suspendedYields.lazy.map { $0.elements }.flatMap { $0 })
+                delegate?.didYield(contentsOf: suspendedYields.lazy.map { $0.elements }.flatMap { $0 })
                 suspendedYields.forEach { $0.continuation.resume() }
 
             case .none:
@@ -327,9 +327,9 @@ extension NIOAsyncWriter {
 
                 switch action {
                 case .callDidYield:
-                    let _delegate = self._delegate
+                    let delegate = self._delegate
                     self._lock.unlock()
-                    _delegate?.didYield(contentsOf: sequence)
+                    delegate?.didYield(contentsOf: sequence)
 
                 case .throwError(let error):
                     self._lock.unlock()
@@ -361,32 +361,32 @@ extension NIOAsyncWriter {
 
         @inlinable
         /* fileprivate */ internal func finish(with failure: Failure?) {
-            let _delegate: Delegate? = self._lock.withLock {
+            let delegate: Delegate? = self._lock.withLock {
                 let action = self._stateMachine.finish()
 
                 switch action {
                 case .callDidTerminate:
-                    let _delegate = self._delegate
+                    let delegate = self._delegate
                     self._delegate = nil
 
-                    return _delegate
+                    return delegate
 
                 case .resumeContinuationsWithErrorAndCallDidTerminate(let suspendedYields, let error):
-                    let _delegate = self._delegate
+                    let delegate = self._delegate
                     self._delegate = nil
                     // It is safe to resume the continuation while holding the lock
                     // since the task will get enqueued on its executor and the resume method
                     // is returning immediately
                     suspendedYields.forEach { $0.continuation.resume(throwing: error) }
 
-                    return _delegate
+                    return delegate
 
                 case .none:
                     return nil
                 }
             }
 
-            _delegate?.didTerminate(failure: failure)
+            delegate?.didTerminate(failure: failure)
         }
     }
 }
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -1,0 +1,750 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import NIOConcurrencyHelpers
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+/// The delegate of the ``NIOAsyncWriter``.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public protocol NIOAsyncWriterDelegate: Sendable {
+    /// The `Element` type of the delegate and the writer.
+    associatedtype Element: Sendable
+    /// The `Failure` type of the delegate and the writer.
+    associatedtype Failure: Error = Never
+
+    /// This method is called once a sequence was yielded to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
+    /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
+    /// until the ``NIOAsyncWriter`` becomes writable again.
+    func didYield<S: Sequence>(contentsOf sequence: S) where S.Element == Element
+
+    /// This method is called once the ``NIOAsyncWriter`` is terminated.
+    ///
+    /// Termination happens if:
+    /// - The ``NIOAsyncWriter`` is deinited.
+    /// - ``NIOAsyncWriter/finish(completion:)`` is called.
+    func didTerminate(completion: NIOAsyncWriterCompletion<Failure>)
+}
+
+/// The signal that the writer doesn’t produce additional elements, either due to normal completion or an error.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public enum NIOAsyncWriterCompletion<Failure: Error>: Sendable {
+    /// The writer finished normally.
+    case finished
+    /// The writer stopped publishing due to the indicated error.
+    case failure(Failure)
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// Errors thrown by the ``NIOAsyncWriter``.
+public struct NIOAsyncWriterError: Error, Hashable {
+    @usableFromInline
+    internal enum _Code: Hashable, Sendable {
+        case alreadyFinished
+    }
+
+    @usableFromInline
+    let _code: _Code
+
+    public static let alreadyFinished: Self = .init(_code: .alreadyFinished)
+}
+
+/// A ``NIOAsyncWriter`` is a type used to bridge elements from the Swift Concurrency domain into
+/// a synchronous world. The `Task`s that are yielding to the ``NIOAsyncWriter`` are the producers.
+/// Whereas the ``NIOAsyncWriterDelegate`` is the consumer.
+///
+/// Additionally, the ``NIOAsyncWriter`` allows the consumer to toggle the writability by calling ``NIOAsyncWriter/Sink/toggleWritability()``.
+/// This allows the implementation of flow control on the consumer side. Any call to ``NIOAsyncWriter/yield(contentsOf:)`` or ``NIOAsyncWriter/yield(_:)``
+/// will suspend if the ``NIOAsyncWriter`` is not writable and only be resumed after the ``NIOAsyncWriter`` becomes writable again.
+///
+/// - Note: It is recommended to never directly expose this type from APIs, but rather wrap it. This is due to the fact that
+/// this type has three generic parameters where at least two should be known statically and it is really awkward to spell out this type.
+/// Moreover, having a wrapping type allows to optimize this to specialized calls if all generic types are known.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOAsyncWriter<
+    Element,
+    Failure,
+    Delegate: NIOAsyncWriterDelegate
+>: Sendable where Delegate.Element == Element, Delegate.Failure == Failure {
+    /// Simple struct for the return type of ``NIOAsyncWriter/makeWriter(elementType:failureType:isWritable:delegate:)``.
+    ///
+    /// This struct contains two properties:
+    /// 1. The ``sink`` which should be retained by the consumer and is used to toggle the writability.
+    /// 2. The ``writer`` which is the actual ``NIOAsyncWriter`` and should be passed to the producer.
+    public struct NewWriter {
+        public let sink: Sink
+        public let writer: NIOAsyncWriter
+
+        @usableFromInline
+        /* fileprivate */ internal init(
+            sink: Sink,
+            writer: NIOAsyncWriter
+        ) {
+            self.sink = sink
+            self.writer = writer
+        }
+    }
+
+    /// This class is needed to hook the deinit to observe once all references to the ``NIOAsyncWriter`` are dropped.
+    ///
+    /// If we get move-only types we should be able to drop this class and use the `deinit` of the ``NIOAsyncWriter`` struct itself.
+    ///
+    /// - Important: This is safe to be unchecked ``Sendable`` since the `storage` is ``Sendable`` and `immutable`.
+    @usableFromInline
+    /* fileprivate */ internal final class InternalClass: @unchecked Sendable {
+        @usableFromInline
+        internal let storage: Storage
+
+        @inlinable
+        init(storage: Storage) {
+            self.storage = storage
+        }
+
+        @inlinable
+        deinit {
+            storage.writerDeinitialized()
+        }
+    }
+
+    @usableFromInline
+    /* private */ internal let internalClass: InternalClass
+
+    @usableFromInline
+    /* private */ internal var storage: Storage {
+        self.internalClass.storage
+    }
+
+    /// Initializes a new ``NIOAsyncWriter`` and a ``NIOAsyncWriter/Sink``.
+    ///
+    /// - Important: This method returns a struct containing a ``NIOAsyncWriter/Sink`` and
+    /// a ``NIOAsyncWriter``. The sink MUST be held by the caller and is used to toggle the writability.
+    /// The writer MUST be passed to the actual producer and MUST NOT be held by the
+    /// caller. This is due to the fact that deiniting the sequence is used as part of a trigger to terminate the underlying sink.
+    ///
+    /// - Parameters:
+    ///   - elementType: The element type of the sequence.
+    ///   - failureType: The failure type of the sequence.
+    ///   - isWritable: The initial writability state of the writer.
+    ///   - delegate: The delegate of the writer.
+    /// - Returns: A ``NIOAsyncWriter/NewWriter``.
+    public static func makeWriter(
+        elementType: Element.Type = Element.self,
+        failureType: Failure.Type = Failure.self,
+        isWritable: Bool,
+        delegate: Delegate
+    ) -> NewWriter {
+        let writer = Self(
+            isWritable: isWritable,
+            delegate: delegate
+        )
+        let sink = Sink(storage: writer.storage)
+
+        return .init(sink: sink, writer: writer)
+    }
+
+    private init(
+        isWritable: Bool,
+        delegate: Delegate
+    ) {
+        let storage = Storage(
+            isWritable: isWritable,
+            delegate: delegate
+        )
+        self.internalClass = .init(storage: storage)
+    }
+
+    /// Yields a sequence of new elements to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` is writable the sequence will get forwarded to the ``NIOAsyncWriterDelegate`` immediately.
+    /// Otherwise, the call to ``NIOAsyncWriter/yield(contentsOf:)`` will get suspended until the ``NIOAsyncWriter``
+    /// becomes writable again.
+    ///
+    /// If the ``NIOAsyncWriter`` is finished while a call to ``NIOAsyncWriter/yield(contentsOf:)`` is suspended the
+    /// call will throw a ``NIOAsyncWriterError/alreadyFinished`` error.
+    ///
+    /// This can be called more than once and from multiple `Task`s at the same time.
+    ///
+    /// - Parameter contentsOf: The sequence to yield.
+    @inlinable
+    public func yield<S: Sequence>(contentsOf sequence: S) async throws where S.Element == Element {
+        try await self.storage.yield(contentsOf: sequence)
+    }
+
+    /// Yields an element to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` is writable the element will get forwarded to the ``NIOAsyncWriterDelegate`` immediately.
+    /// Otherwise, the call to ``NIOAsyncWriter/yield(contentsOf:)`` will get suspended until the ``NIOAsyncWriter``
+    /// becomes writable again.
+    ///
+    /// If the ``NIOAsyncWriter`` is finished while a call to ``NIOAsyncWriter/yield(contentsOf:)`` is suspended the
+    /// call will throw a ``NIOAsyncWriterError/alreadyFinished`` error.
+    ///
+    /// This can be called more than once and from multiple `Task`s at the same time.
+    ///
+    /// - Parameter element: The element to yield.
+    @inlinable
+    public func yield(_ element: Element) async throws {
+        try await self.yield(contentsOf: CollectionOfOne(element))
+    }
+
+    /// Finishes the writer.
+    ///
+    /// Calling this function signals the writer that any suspended or subsequent calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will return a ``NIOAsyncWriterError/alreadyFinished`` error.
+    ///
+    /// - Note: Calling this function more than once has no effect.
+    /// - Parameter completion: A ``NIOAsyncWriterCompletion`` which indicates whether publishing has finished normally
+    /// or failed with an error.
+    @inlinable
+    public func finish(completion: NIOAsyncWriterCompletion<Failure>) {
+        self.storage.finish(completion: completion)
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncWriter {
+    /// The underlying sink of the ``NIOAsyncWriter``. This type allows to toggle the writability of the ``NIOAsyncWriter``.
+    public struct Sink {
+        @usableFromInline
+        /* private */ internal var storage: Storage
+
+        @inlinable
+        init(storage: Storage) {
+            self.storage = storage
+        }
+
+        /// Toggles the writability of the ``NIOAsyncWriter``.
+        ///
+        /// If the writer becomes writable again all suspended yields will be resumed and the produced elements will be forwarded via
+        /// the ``NIOAsyncWriterDelegate/didYield(contentsOf:)`` method.
+        public func toggleWritability() {
+            self.storage.toggleWritability()
+        }
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncWriter {
+    /// This is the underlying storage of the writer. The goal of this is to synchronize the access to all state.
+    @usableFromInline
+    /* fileprivate */ internal final class Storage: @unchecked Sendable {
+        /// The lock that protects our state.
+        @usableFromInline
+        /* private */ internal let lock = Lock()
+        /// The counter used to assign an ID to all our yields.
+        @usableFromInline
+        /* private */ internal let yieldIDCounter = ManagedAtomic<UInt64>(0)
+        /// The state machine.
+        @usableFromInline
+        /* private */ internal var stateMachine: StateMachine
+        /// The delegate.
+        @usableFromInline
+        /* private */ internal var delegate: Delegate?
+
+        @usableFromInline
+        /* fileprivate */ internal init(
+            isWritable: Bool,
+            delegate: Delegate
+        ) {
+            self.stateMachine = .init(isWritable: isWritable)
+            self.delegate = delegate
+        }
+
+        @inlinable
+        /* fileprivate */ internal func writerDeinitialized() {
+            let delegate: Delegate? = self.lock.withLock {
+                let action = self.stateMachine.writerDeinitialized()
+
+                switch action {
+                case .callDidTerminate:
+                    let delegate = self.delegate
+                    self.delegate = nil
+
+                    return delegate
+
+                case .none:
+                    return nil
+                }
+            }
+
+            delegate?.didTerminate(completion: .finished)
+        }
+
+        @inlinable
+        /* fileprivate */ internal func toggleWritability() {
+            // We are manually locking here since we need to use both the delegate and the
+            // suspendedYields. Doing this with a withLock becomes very unhandy.
+            self.lock.lock()
+            let action = self.stateMachine.toggleWritability()
+
+            switch action {
+            case .callDidYieldAndResumeContinuations(let suspendedYields):
+                let delegate = self.delegate
+                self.lock.unlock()
+
+                delegate?.didYield(contentsOf: suspendedYields.lazy.map { $0.elements }.flatMap { $0 })
+                suspendedYields.forEach { $0.continuation.resume() }
+
+            case .none:
+                self.lock.unlock()
+                return
+            }
+        }
+
+        @inlinable
+        /* fileprivate */ internal func yield<S: Sequence>(contentsOf sequence: S) async throws where S.Element == Element {
+            // Using relaxed is fine here since we do not need any strict ordering just a
+            // unique ID for every yield.
+            let yieldID = self.yieldIDCounter.loadThenWrappingIncrement(ordering: .relaxed)
+
+            try await withTaskCancellationHandler {
+                // We are manually locking here to hold the lock across the withCheckedContinuation call
+                self.lock.lock()
+
+                let action = self.stateMachine.yield(yieldID: yieldID)
+
+                switch action {
+                case .callDidYield:
+                    let delegate = self.delegate
+                    self.lock.unlock()
+                    delegate?.didYield(contentsOf: sequence)
+
+                case .throwError(let error):
+                    self.lock.unlock()
+                    throw error
+
+                case .suspendTask:
+                    try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<Void, Error>) in
+                        self.stateMachine.yield(
+                            contentsOf: sequence,
+                            continuation: continuation,
+                            yieldID: yieldID
+                        )
+
+                        self.lock.unlock()
+                    }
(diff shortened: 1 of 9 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import NIOConcurrencyHelpers
```
glbrntt, line 16: Can we move these into the compiler guards?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+    /// This method is called once a sequence was yielded to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
+    /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
+    /// until the ``NIOAsyncWriter`` becomes writable again.
+    func didYield<S: Sequence>(contentsOf sequence: S) where S.Element == Element
+
+    /// This method is called once the ``NIOAsyncWriter`` is terminated.
+    ///
+    /// Termination happens if:
```
glbrntt, line 36: Is this always called exactly once?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public enum NIOAsyncWriterCompletion<Failure: Error>: Sendable {
+    /// The writer finished normally.
+    case finished
+    /// The writer stopped publishing due to the indicated error.
+    case failure(Failure)
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+/// Errors thrown by the ``NIOAsyncWriter``.
```
glbrntt, line 52:
```suggestion
/// Errors thrown by the ``NIOAsyncWriter``.
@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+public struct NIOAsyncWriter<
+    Element,
+    Failure,
+    Delegate: NIOAsyncWriterDelegate
+>: Sendable where Delegate.Element == Element, Delegate.Failure == Failure {
+    /// Simple struct for the return type of ``NIOAsyncWriter/makeWriter(elementType:failureType:isWritable:delegate:)``.
+    ///
+    /// This struct contains two properties:
+    /// 1. The ``sink`` which should be retained by the consumer and is used to toggle the writability.
+    /// 2. The ``writer`` which is the actual ``NIOAsyncWriter`` and should be passed to the producer.
```
glbrntt, line 86: Can we move the docs onto the properties?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+        }
+    }
+
+    /// This class is needed to hook the deinit to observe once all references to the ``NIOAsyncWriter`` are dropped.
+    ///
+    /// If we get move-only types we should be able to drop this class and use the `deinit` of the ``NIOAsyncWriter`` struct itself.
+    ///
+    /// - Important: This is safe to be unchecked ``Sendable`` since the `storage` is ``Sendable`` and `immutable`.
+    @usableFromInline
+    /* fileprivate */ internal final class InternalClass: @unchecked Sendable {
```
glbrntt, line 107: If `Storage` is `Sendable` and immutable then this doesn't need to be `@unchecked`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+        }
+
+        @inlinable
+        deinit {
+            storage.writerDeinitialized()
+        }
+    }
+
+    @usableFromInline
+    /* private */ internal let internalClass: InternalClass
```
glbrntt, line 123: nit: we generally prefix with an underscore for private-but-usable-from-inline properties.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+    associatedtype Element: Sendable
+    /// The `Failure` type of the delegate and the writer.
+    associatedtype Failure: Error = Never
+
+    /// This method is called once a sequence was yielded to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
+    /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
+    /// until the ``NIOAsyncWriter`` becomes writable again.
+    func didYield<S: Sequence>(contentsOf sequence: S) where S.Element == Element
```
glbrntt, line 32: Do we always yield a complete sequence?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+    /// Finishes the writer.
+    ///
+    /// Calling this function signals the writer that any suspended or subsequent calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will return a ``NIOAsyncWriterError/alreadyFinished`` error.
+    ///
+    /// - Note: Calling this function more than once has no effect.
+    /// - Parameter completion: A ``NIOAsyncWriterCompletion`` which indicates whether publishing has finished normally
+    /// or failed with an error.
+    @inlinable
+    public func finish(completion: NIOAsyncWriterCompletion<Failure>) {
```
glbrntt, line 212: Is it actually worth having a _public_ `NIOAsyncWriterCompletion` here rather than say, `finish()` and `finish(error:)`?

`finish(completion: .finished)` and `finish(completion: .failure(.someError))` seem unnecessarily verbose.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+        self.storage.finish(completion: completion)
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncWriter {
+    /// The underlying sink of the ``NIOAsyncWriter``. This type allows to toggle the writability of the ``NIOAsyncWriter``.
+    public struct Sink {
+        @usableFromInline
+        /* private */ internal var storage: Storage
```
glbrntt, line 222: is this intentionally `var`?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+        @inlinable
+        init(storage: Storage) {
+            self.storage = storage
+        }
+
+        /// Toggles the writability of the ``NIOAsyncWriter``.
+        ///
+        /// If the writer becomes writable again all suspended yields will be resumed and the produced elements will be forwarded via
+        /// the ``NIOAsyncWriterDelegate/didYield(contentsOf:)`` method.
+        public func toggleWritability() {
```
glbrntt, line 233: Is this the right API? I wonder if setting the writability explicitly would be more useful (if anything it makes the tests more obvious). In `actor` versions it _had_ to be toggle because calls to set the writability could be reordered.

glbrntt (commented):
Only just got to the meat of this PR; will pick up the rest of the review later. Leaving my comments so far.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: We can but we are not aligned on this. Even in my previous `NIOAsyncSequenceProducer` PR we haven't done this.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Yes this is guaranteed to be called exactly one same as the `didTerminate` on the `NIOAsyncSeuqenceProducerDelegate`

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: We always `yield` the complete sequence that the user yielded to us. However, the user might have also yielded a single element which we will yield as a `CollectionOfOne` sequence.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I adopted the spelling from `Combine`. We also have the spelling from `AsyncThrowingStream.Continuation` which has `finish(throwing error: Failure? = nil`.
I like your approach the best though having to separate methods

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
glbrntt: IMHO if it isn't required it shouldn't be imported: if we don't have Swift > 5.5.2 then we shouldn't import these

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
glbrntt: Can we document that?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I kept this in because to get the writability of a `Channel` you need to load the value from the underlying atomic. Since, that load is normally done using a `relaxed` load I wasn't sure if this is safe to do in the `channelWritabilityChanged` method. However, after talking with @Lukasa this is safe to do so I am changed the API to be `setWritability(to:)`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,750 @@
...
+            self.lock.lock()
+            let action = self.stateMachine.toggleWritability()
+
+            switch action {
+            case .callDidYieldAndResumeContinuations(let suspendedYields):
+                let delegate = self.delegate
+                self.lock.unlock()
+
+                delegate?.didYield(contentsOf: suspendedYields.lazy.map { $0.elements }.flatMap { $0 })
+                suspendedYields.forEach { $0.continuation.resume() }
```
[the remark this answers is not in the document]
FranzBusch: I prefer keeping the `forEach` since it is more concise for a very simple task. However, in general I agree to prefer `for ...in` over `forEach` for anything that is more complex

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,763 @@
...
+                        suspendedYield.continuation,
+                        CancellationError()
+                    )
+
+                } else {
+                    // There is no suspended yield. This can mean that we either already yielded
+                    // or that the call to `yield` is coming afterwards. We need to store
+                    // the ID here. However, if the yield already happened we will never remove the
+                    // stored ID. The only way to avoid doing this would be storing every ID
+                    cancelledYields.append(yieldID)
```
glbrntt, line 703: Do we need to be `.modifying` here?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,763 @@
...
+                // We are already finished and still tried to write something
+                return .throwError(NIOAsyncWriterError.alreadyFinished)
+
+            case .modifying:
+                preconditionFailure("Invalid state")
+            }
+        }
+
+        @inlinable
+        /* fileprivate */ internal mutating func yield<S: Sequence>(
```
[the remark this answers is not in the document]
FranzBusch: I see. I kinda followed what I did for the `NIOAsyncSequenceProducer` where I have `next()` and `next(for:)`. If you want I can change it. Up to you!

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Good catch!

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,763 @@
...
+        }
+
+        /// Sets the writability of the ``NIOAsyncWriter``.
+        ///
+        /// If the writer becomes writable again all suspended yields will be resumed and the produced elements will be forwarded via
+        /// the ``NIOAsyncWriterDelegate/didYield(contentsOf:)`` method. If the writer becomes unwritable all
+        /// subsequent calls to ``NIOAsyncWriterDelegate/didYield(contentsOf:)`` will suspend.
+        ///
+        /// - Parameter writability: The new writability of the ``NIOAsyncWriter``.
+        public func setWritability(to writability: Bool) {
```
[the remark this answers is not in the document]
FranzBusch: Hm, I am not 100% convinced. If we change it to a property we also have to provide a `getter` which we would need to go through the lock to get the actual value. Since, I don't expect anyone to be actually querying for the writable of the `Sink` I am kinda hesitant to make that change.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+///
+/// Additionally, the ``NIOAsyncWriter`` allows the consumer to toggle the writability by calling ``NIOAsyncWriter/Sink/toggleWritability()``.
+/// This allows the implementation of flow control on the consumer side. Any call to ``NIOAsyncWriter/yield(contentsOf:)`` or ``NIOAsyncWriter/yield(_:)``
+/// will suspend if the ``NIOAsyncWriter`` is not writable and only be resumed after the ``NIOAsyncWriter`` becomes writable again.
+///
+/// - Note: It is recommended to never directly expose this type from APIs, but rather wrap it. This is due to the fact that
+/// this type has three generic parameters where at least two should be known statically and it is really awkward to spell out this type.
+/// Moreover, having a wrapping type allows to optimize this to specialized calls if all generic types are known.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOAsyncWriter<
```
fabianfett, line 96: I'm not sure, I'm a fan of this being a struct. The AsyncWriter has clearly reference semantics.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=5.5.2) && canImport(_Concurrency)
+import Atomics
+import NIOConcurrencyHelpers
+
+/// The delegate of the ``NIOAsyncWriter``.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public protocol NIOAsyncWriterDelegate: Sendable {
```
fabianfett, line 21: Is there any situation, in which this could be a `struct`? Can we require this to inherit from `AnyObject`?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+    @inlinable
+    public func finish(with failure: Failure) {
+        self._storage.finish(with: failure)
+    }
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncWriter {
+    /// The underlying sink of the ``NIOAsyncWriter``. This type allows to toggle the writability of the ``NIOAsyncWriter``.
+    public struct Sink {
```
fabianfett, line 251: What's the benefit of this type? Ensure that a consumer of the `NIOAsyncWriter` can only call `setWritability` on it? In my mind the `OutboundChannelHandler` implements the Delegate protocol and holds a reference to the writer. `OutboundChannelHandler` can then easily toggle writability on the `NIOAsyncWriter`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+        /* private */ internal let _lock = Lock()
+        /// The counter used to assign an ID to all our yields.
+        @usableFromInline
+        /* private */ internal let _yieldIDCounter = ManagedAtomic<UInt64>(0)
+        /// The state machine.
+        @usableFromInline
+        /* private */ internal var _stateMachine: StateMachine
+        /// The delegate.
+        @usableFromInline
+        /* private */ internal var _delegate: Delegate?
```
fabianfett, line 289: I assume, holding the `Delegate`, or not holding the `Delegate` is part of the internal state. Should the delegate be referenced from the StateMachine? Then the stateMachine should return the Delegate when needed for some actions. This way, we should get rid of a bunch of `?`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+}
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncWriter {
+    /// This is the underlying storage of the writer. The goal of this is to synchronize the access to all state.
+    @usableFromInline
+    /* fileprivate */ internal final class Storage: @unchecked Sendable {
+        /// The lock that protects our state.
+        @usableFromInline
+        /* private */ internal let _lock = Lock()
```
fabianfett, line 280: What does the underscore mean here?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+    /* private */ internal struct StateMachine {
+        /// This is a small helper struct to encapsulate the three different values for a suspended yield.
+        @usableFromInline
+        /* private */ internal struct SuspendedYield {
+            /// The yield's ID.
+            @usableFromInline
+            var yieldID: UInt64
+            /// The yield's produced sequence of elements.
+            @usableFromInline
+            var elements: AnySequence<Element>
```
fabianfett, line 431: Should the AsyncWriter be generic over the Sequence type as well? How often do user's want to send different Sequence types? For the SequenceOfOne/Two case users could always opt for an enum that implements Sequence.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+        /* fileprivate */ internal func yield<S: Sequence>(contentsOf sequence: S) async throws where S.Element == Element {
+            // Using relaxed is fine here since we do not need any strict ordering just a
+            // unique ID for every yield.
+            let yieldID = self._yieldIDCounter.loadThenWrappingIncrement(ordering: .relaxed)
+
+            try await withTaskCancellationHandler {
+                // We are manually locking here to hold the lock across the withCheckedContinuation call
+                self._lock.lock()
+
+                let action = self._stateMachine.yield(yieldID: yieldID)
```
fabianfett, line 351: What's the benefit of having the `yieldIDCounter` as an atomic, if we acquire a lock afterwards anyway?

fabianfett (commented):
Great start. Left some broader questions for now.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Fair point, I think making this a class isn't bad. I went with the struct first approach but open to change it

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Yeah sometimes you want to use a struct for this to have an intermediate type that is implementing this. I think forcing this to inherit from `AnyObject` is overly restrictive.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: The benefit of this type is that the producer and the delegate are holding two separate types. This allows us to make assumption of when the `NIOAsyncWriter` actually gets deallocated since this means we can tell the delegate that we terminated. I stated in the doc comments of `makeWriter` that you should hold the `Sink` in the place where you consume the writes and the `AsyncWriter` in the async world.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: The underscore her is due to the fact that we would like make the lock `private` but can't since we want to add `@usableFromInline`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Yeah, I didn't like the point of holding an `AnySequence` here. If we do that, then we should get rid of the yielding of a single element.

fabianfett: > If we do that, then we should get rid of the yielding of a single element.
I think we must and I consider this a good thing :P

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,763 @@
...
+public struct NIOAsyncWriterError: Error, Hashable {
+    @usableFromInline
+    internal enum _Code: Hashable, Sendable {
+        case alreadyFinished
+    }
+
+    @usableFromInline
+    let _code: _Code
+
+    public static let alreadyFinished: Self = .init(_code: .alreadyFinished)
```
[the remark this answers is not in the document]
glbrntt: Up to you. This PR is already pretty large. I'm happy to leave it for now.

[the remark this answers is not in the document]
glbrntt: This needs to become `func` now so we capture the file and line it was called, otherwise it'll always be this file and line

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,789 @@
...
+    let _code: _Code
+
+    @usableFromInline
+    var file: String
+
+    @usableFromInline
+    var line: Int
+
+    @inlinable
+    init(_code: _Code, file: String = #fileID, line: Int = #line) {
```
glbrntt, line 66: We should remove the defaults here so we ensure they're always provided by the helpers

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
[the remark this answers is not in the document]
glbrntt: You're right, we don't need a getter here.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
glbrntt: > Should the AsyncWriter be generic over the Sequence type as well? How often do user's want to send different Sequence types?

I'm not sure that's a good idea in practice. I think the typical usage pattern would be for a framework to create a writer and pass it to a caller. That would mean the framework would be responsible for choosing what the sequence type is _or_ the caller has to ask the framework for a writer with a given sequence type (which is a pretty awkward API).

> For the SequenceOfOne/Two case users could always opt for an enum that implements Sequence.

Not offering a yield-one API and requiring users to implement an enum like this makes for a really user hostile API IMHO.

---

FWIW I also don't like `AnySequence` here but I think we should keep it for the time being (it is an implementation detail and can be changed later). 

One alternative would be to always store `[Element]` and special case the API to avoid copies when an array is provided and copy into an array when anything else is provided. 

Taking that further we could be to pull a 'NIOAny' trick and have an enum of `array`, `one`, and `anySequence`.

That said, I think we should focus on getting the API right first and then compare perf.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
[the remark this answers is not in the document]
glbrntt: Don't feel strongly about it, a small comment would help though.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I agree with what George said. Making the whole thing generic of the sequence that is yielded severely limits the usage of the type and makes it very awkward to hold. I would stick with `AnySequence` for now and as George said we can always make it faster by applying some of our tricks.

Lukasa (changes requested):
I think we're mishandling the delegate throughout this code.

Because we only call the delegate from _outside_ the lock, we do not guarantee to maintain ordering of yields. We can suspend immediately after we drop the lock, and then have a completely unrelated Task come in and yield results, get an immediate call to `didYield`, and then drop into the delegate, beating the first Task.

This didn't matter in the previous delegate, which was only being called with signals that were themselves well-ordered, but it does matter here. We need a fundamentally different behaviour that maintains ordering, probably by way of keeping the lock held across these calls. That will require some thought as to how it interacts with `setWritability`, which we need to be careful wtih.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Yeah happy to change this. Especially now that we are calling the delegate while holding the lock

glbrntt (approved):
Looks great! Left a few documentation nits/fixes.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,823 @@
...
+///
+/// - Important: The methods on the delegate are called while a lock inside of the ``NIOAsyncWriter`` is held. This is done to
+/// guarantee the ordering of the writes. However, this means you **MUST** avoid calling ``NIOAsyncWriter/Sink/setWritability(to:)``
+/// from within ``NIOAsyncWriterDelegate/didYield(contentsOf:)`` or ``NIOAsyncWriterDelegate/didTerminate(failure:)``.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public protocol NIOAsyncWriterDelegate: Sendable {
+    /// The `Element` type of the delegate and the writer.
+    associatedtype Element: Sendable
+    /// The `Failure` type of the delegate and the writer.
+    associatedtype Failure: Error = Never
```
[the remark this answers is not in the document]
FranzBusch: AFAIK, they existed for quite some time; however, there are various quirks in the current type checker that require you to still explicitly specify them in the implementation which is kinda sad.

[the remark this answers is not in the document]
glbrntt: Seems to be an undocumented feature: https://github.com/apple/swift/issues/51269

glbrntt (approved):

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,851 @@
...
+extension NIOAsyncWriter {
+    /// This is the underlying storage of the writer. The goal of this is to synchronize the access to all state.
+    @usableFromInline
+    /* fileprivate */ internal final class Storage: @unchecked Sendable {
+        /// The lock that protects our state.
+        @usableFromInline
+        /* private */ internal let _lock = Lock()
+        /// The counter used to assign an ID to all our yields.
+        @usableFromInline
+        /* private */ internal let _yieldIDCounter = ManagedAtomic<UInt64>(0)
```
Lukasa, line 329: Can we wrap this in a `struct` that gives us back an object that isn't a `UInt64` that has the API surface we want? The way we use this made me think for a while that it was supposed to be a monotonically increasing value, but in actuality it just needs to be a _unique_ value. Having a type that encapsulates this logic and doesn't encourage comparison would be helpful.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
Lukasa: As discussed offline, I think we should change this to remove the `AnySequence`.

FranzBusch:
@Lukasa I changed the API to always yield a `Deque` now and I added a performance test for this. Let me know what you think!

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,934 @@
...
+
+    /// Finishes the writer.
+    ///
+    /// Calling this function signals the writer that any suspended or subsequent calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will return a ``NIOAsyncWriterError/alreadyFinished`` error.
+    ///
+    /// - Note: Calling this function more than once has no effect.
+    /// - Parameter failure: The failure indicating why the writer finished.
+    @inlinable
+    public func finish(with failure: Failure) {
```
glbrntt, line 268: Minor bikeshedding... I don't love the use of `with` here. To me, `finish(with:)` suggests a completion value (rather than completing as a result of a failure).

Some previous art:
- `AsyncThrowingStream.Continuation` has `finish(throwing:)`
- `AsyncThrowingChannel` has `fail(_:)`

I prefer `finish(throwing:)` so that it's similar to `finish()` over `fail(_:)`.

glbrntt: Agree `throwing:` is weird if it isn't thrown. It does at least make it obvious when reading that it's an error. I think `finish(failure:)`/`finish(error:)` would be fine.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,934 @@
...
+///
+/// - Note: It is recommended to never directly expose this type from APIs, but rather wrap it. This is due to the fact that
+/// this type has three generic parameters where at least two should be known statically and it is really awkward to spell out this type.
+/// Moreover, having a wrapping type allows to optimize this to specialized calls if all generic types are known.
+///
+/// - Note: This struct has reference semantics. Once all copies of a writer have been dropped ``NIOAsyncWriterSinkDelegate/didTerminate(failure:)`` will be called.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+public struct NIOAsyncWriter<
+    Element,
+    Failure,
```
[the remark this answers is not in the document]
FranzBusch: They already are. The delegate is currently restricting the `Element` to be `Sendable` which propagates to this generic type through the constraints. Furthermore, `Failure` is restricted to `Error` which inherits from `Sendable`.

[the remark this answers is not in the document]
FranzBusch: Actually taking this back since it produces a warning since it is a duplicate constraint.

FranzBusch:
@Lukasa @glbrntt @gjcairo I just pushed a new commit that changes a few things after our offline discussion:
- Allow sending an `Error` with the `Sink/finish(error:)` method. Any suspended or subsequent call to `NIOAsyncWriter/yield` will throw this error. It provides a way to surface an underlying error to the callers of the `NIOAsyncWriter`. We have been doing this in `grpc-swift` to surface errors from the channel pipeline 
- I removed the generic `Failure` parameter from the `NIOAsyncWriter` since we agreed that there is little value in providing a generic here over just an `Error`. Furthermore, since we want to allow to send an `Error` from the `Sink/finish(error:)` method it would mean we would have to introduce another generic parameter.
- Added a few more tests for the new state and updated the documentation.

Lukasa (approved):

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,1060 @@
...
+
+    /// Yields a sequence of new elements to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` is writable the sequence will get forwarded to the ``NIOAsyncWriterSinkDelegate`` immediately.
+    /// Otherwise, the sequence will be buffered and the call to ``NIOAsyncWriter/yield(contentsOf:)`` will get suspended until the ``NIOAsyncWriter``
+    /// becomes writable again. If the calling `Task` gets cancelled at any point the call to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// will be resumed.
+    ///
+    /// If the ``NIOAsyncWriter`` is finished while a call to ``NIOAsyncWriter/yield(contentsOf:)`` is suspended the
+    /// call will throw a ``NIOAsyncWriterError/alreadyFinished`` error.
```
glbrntt, line 217: I don't think this is accurate.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,1060 @@
...
+
+    /// Yields an element to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` is writable the element will get forwarded to the ``NIOAsyncWriterSinkDelegate`` immediately.
+    /// Otherwise, the element will be buffered and the call to ``NIOAsyncWriter/yield(_:)`` will get suspended until the ``NIOAsyncWriter``
+    /// becomes writable again. If the calling `Task` gets cancelled at any point the call to ``NIOAsyncWriter/yield(_:)``
+    /// will be resumed.
+    ///
+    /// If the ``NIOAsyncWriter`` is finished while a call to ``NIOAsyncWriter/yield(_:)`` is suspended the
+    /// call will throw a ``NIOAsyncWriterError/alreadyFinished(file:line:)`` error.
```
glbrntt, line 235: I don't think this is accurate.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,1060 @@
...
+    @inlinable
+    public func yield(_ element: Element) async throws {
+        try await self.yield(contentsOf: CollectionOfOne(element))
+    }
+
+    /// Finishes the writer.
+    ///
+    /// Calling this function signals the writer that any suspended calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will be resumed. Any subsequent calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will throw.
```
glbrntt, line 249: Same here re: yielding

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -0,0 +1,1060 @@
...
+    @inlinable
+    public func finish() {
+        self._storage.writerFinish(error: nil)
+    }
+
+    /// Finishes the writer.
+    ///
+    /// Calling this function signals the writer that any suspended calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will be resumed. Any subsequent calls to ``NIOAsyncWriter/yield(contentsOf:)``
+    /// or ``NIOAsyncWriter/yield(_:)`` will throw.
```
glbrntt, line 264: Same here re: yielding

glbrntt (commented):
This is great @FranzBusch -- I think a couple of docs are out of date and should be fixed in a followup.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I think this one is actually correct.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Same here

PR #2258, opened 2022-08-31
Call finish once the Source is deinited

FranzBusch:
# Motivation
We **MUST** call `finish()` when the `Source` deinits otherwise we can have a suspended continuation that never gets resumed.

# Modification
Introduce an internal class to both `Source`s and call `finish()` in their `deinit`s.

# Result
We are now resuming all continuations.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
@@ -90,7 +90,7 @@ public struct NIOAsyncSequenceProducer<
     Strategy: NIOAsyncSequenceProducerBackPressureStrategy,
     Delegate: NIOAsyncSequenceProducerDelegate
 >: Sendable {
-    /// Simple struct for the return type of ``NIOAsyncSequenceProducer/makeSequence(of:backPressureStrategy:delegate:)``.
+    /// Simple struct for the return type of ``NIOAsyncSequenceProducer/makeSequence(elementType:backPressureStrategy:delegate:)``.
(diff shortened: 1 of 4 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -198,30 +198,51 @@ extension NIOAsyncSequenceProducer {
...
+
+            @inlinable
+            init(throwingSource: ThrowingSource) {
+                self._throwingSource = throwingSource
+            }
+
+            @inlinable
+            deinit {
+                // We need to call finish here to resume any suspended continuation.
+                _throwingSource.finish()
```
fabianfett, line 228:
```suggestion
                self._throwingSource.finish()
```

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -198,30 +198,51 @@ extension NIOAsyncSequenceProducer {
...
-            Element,
-            Never,
-            Strategy,
-            Delegate
-        >.Source
-
-        @usableFromInline
-        /* fileprivate */ internal init(
-            throwingSource: NIOThrowingAsyncSequenceProducer<
+        /* fileprivate */ internal final class InternalClass: @unchecked Sendable {
```
fabianfett, line 208: Do we need `@unchecked` here?

Lukasa (approved):
Great change, thanks @FranzBusch

PR #2261, opened 2022-09-02
NTRIP RTCM body correction

cloewen8:
Fixes RTCM body data being interpreted as an HTTP header.

**Relies on pull request https://github.com/apple/swift-nio/pull/2260**

### Motivation:

Some NTRIP servers (called casters) produce an HTTP response that lacks a CRLF between the status-line and RTCM-based body.

> One error results in the body of the response being interpreted as a header, causing the execution to never produce a result or to produce an incorrect result (infinite loop, timeout, or invalid header and cut off body).
In practice, this only ever appears to occur when the body contains RTCM data. This data always starts with a D3 hex character.

### Modifications:

The http-parser main loop was modified to check if a header starts with \xC3 and then \x93 (or Ó). If it does, the read is rolled back and the state is changed to s_headers_almost_done.
If the header only starts with \xC3 but not \x93, the HPE_INVALID_HEADER_TOKEN error is produced, as \xC3 is not a tokenizable character (see the tokens array comment or RFC 2616).

### Result:

This allows RTCM responses to be properly parsed while still producing the original HTTP header error.

**It may be possible to further restrict this behavior to ICY responses only. Further verification is needed on my end to see if this is viable.**


Lukasa (changes requested):
Initially I'd like us to follow along in #2260. As a more general note to the shape of this patch, I'd like us to require users to opt-in to the NTRIP parser leniency so that users not expecting NTRIP won't accidentally tolerate malformed HTTP.

PR #2263, opened 2022-09-05
Update HTTP parser to LLHTTP

Lukasa:
### Motivation:

The node.js HTTP parser library that we use has been unmaintained for some time. We should move to the maintained replacement, which is [llhttp](https://github.com/nodejs/llhttp). This patch will update our dependency and bring us over to the new library, as well as make any changes we need.

### Modifications:

This patch comes in 4 parts, each contained in a separate commit in the PR.

The first commit drops the existing http_parser code and updates some of the repo state for using llhttp.
The second commit rewrites the update script to bring in llhttp instead of http_parser.
The third runs the actual script. You can skip reviewing this except to sanity check the outcome.
The fourth commit updates the NIO code and the tests to get everything working.

In general the substance of the product modifications was minimal. The logic around keeping track of where we are in the buffer and how upgrades work has changed a bit, so that required some fiddling. I also had to add an error reporting path for the delegates to be able to throw specific errors that llhttp no longer checks for. Finally, I removed two tests that were a little overzealous and that llhttp does not police.

### Result:

Back on the supported path.


3a4oT:
Just a side question, why a project like NIO doesn't have an HTTP parser in Swift? I would assume because of performance. But how slow is it?

Lukasa:
It's not performance related, it's correctness and time related.

Having a HTTP/1.1 parser that disagrees with common reverse proxies about how to parse a HTTP/1.1 request or response is a big security risk: confused deputy issues are common and troubling. Writing a parser that does this well is a lot of work, and it's not been anywhere near to the most valuable thing we could do with our time, so we haven't.

As discussed in #2260, we have a long-term intention to bring the parser into Swift. However, before we do it we'd need to produce an exhaustive test suite that codifies our current behaviour and use that to confirm we do not diverge from that behaviour. This work does not produce a lot of value in itself, and while it does improve the risk from the unsafe code that llhttp provides, the fact that llhttp is slow-moving and widely used reduces the size of that risk substantially.

All-in-all, having a HTTP parser in Swift is important, but no so important that it has become a big enough priority for us to do the work.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -350,41 +360,58 @@ private class BetterHTTPParser {
...
             // if we chose to abort (eg. wrong HTTP version) the error will be in self.httpErrno, otherwise http_parser
             // will tell us...
             // self.parser must be non-nil here because we can't be re-entered here (ByteToMessageDecoder guarantee)
-            // If we have a richer error than the errno code, and the errno is unknown, we'll use it. Otherwise, we use the
+            // If we have a richer error than the errno code, and the errno is internal, we'll use it. Otherwise, we use the
             // error from http_parser.
-            let err = self.httpErrno ?? http_errno(rawValue: parserErrno)
-            if err == HPE_UNKNOWN, let richerError = self.richerError {
+            let err = self.httpErrno ?? parserErrno
+            if (err == HPE_INTERNAL || err == HPE_USER), let richerError = self.richerError {
```
glbrntt, line 406: Why also `HPE_USER` here?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -733,52 +770,52 @@ extension HTTPParserError {
...
+             HPE_CB_HEADERS_COMPLETE,
+             HPE_CB_MESSAGE_COMPLETE,
+             HPE_CB_CHUNK_HEADER,
+             HPE_CB_CHUNK_COMPLETE,
+             HPE_USER,
+             HPE_CB_URL_COMPLETE,
+             HPE_CB_STATUS_COMPLETE,
+             HPE_CB_HEADER_FIELD_COMPLETE,
+             HPE_CB_HEADER_VALUE_COMPLETE:
+            // The downside of enums here, we don't have a case for these. Map them to .unknown for now.
```
glbrntt, line 818: Bit annoying... I assume there's limited value in deprecating and replacing `HTTPParserError`?

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Two different codepaths. Annoyingly, we get HPE_INTERNAL from one, and HPE_USER from the other.

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: I don't see much value in it tbh.

glbrntt (approved):

Tests/NIOHTTP1Tests/HTTPDecoderTest.swift
```diff
@@ -917,4 +917,79 @@ class HTTPDecoderTest: XCTestCase {
...
+
+        var buffer = ByteBuffer(string: "GET / HTTP/1.1\r\nHost: example.com\r\n")
+
+        XCTAssertNoThrow(try self.channel.writeInbound(buffer))
+        XCTAssertNoThrow(XCTAssertNil(try self.channel.readInbound(as: HTTPServerRequestPart.self)))
+
+        buffer.clear()
+        buffer.writeRepeatingByte(UInt8(ascii: "x"), count: 1024)
+
+        for _ in 0..<80 {
```
glbrntt, line 932: nit: 80 and 1024 are a bit magic here, can you comment with a reference to `maximumHeaderFieldSize` in these tests?

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -354,41 +371,58 @@ private class BetterHTTPParser {
...
+                                                startPointer,
+                                                bytesToRead)
+
+                if rc == HPE_PAUSED_UPGRADE {
+                    // This is a special pause. We don't need to stop here (our other code will prevent us
+                    // parsing past this point, but we do need a special hook to work out how many bytes were read.
+                    // The force-unwrap is safe: we know we hit an "error".
+                    bytesRead = UnsafeRawPointer(c_nio_llhttp_get_error_pos(parserPtr)!) - startPointer
+                    c_nio_llhttp_resume_after_upgrade(parserPtr)
+                    rc = HPE_OK
```
FranzBusch, line 397: Just for my understanding could you explain me what exactly is going on here and why are we setting `rc` to `HPE_OK`?

FranzBusch (approved):
The changes look good and the tests are confirming that we are now doing the right thing. I added a question for my understanding.

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Sure!

llhttp has a "pause on upgrade" behaviour that will stop the parser at the point that it has completely parse a message that will trigger an upgrade, usually a 101 response. This is sensible: whatever bytes fall behind the parser from the point are not likely to be HTTP and so probably shouldn't be parsed.

In our case, however, we have different logic that handles this "stop parsing" behaviour, because http_parser didn't have the "pause on upgrade" behaviour so we invented it ourself. This essentially mimics the previous behaviour: we work out how many bytes actually got consumed, then we just pretend that everything went great and let the rest of the code kick in.

PR #2264, opened 2022-09-07
helpful error message when compiling without test discovery on >= 5.5

weissi:
Motivation:

When compiling on Swift >= 5.5, we require `--enable-test-discovery` to be passed to `swift build --build-tests`/`swift test`. But the error message was terrible.

Modification:

Give a helpful error message when compiling without `--enable-test-discovery`.

Result:

Less puzzled users.


Lukasa (approved):
Thanks!

PR #2265, opened 2022-09-07
NIOConcurrency: add NIOLockedValueBox

weissi:
Motivation:

`NIOLockedValueBox` can help programmers to not forget to actually take a `Lock` when they should.

Modifications:

Add `NIOLockedValueBox` (which currently just uses a `Lock` (so it'll be 3 allocations) but can easily be changed to tail-allocate a `pthread_mutex_t` (to make it 1 allocation))

Result:

Hopefully fewer concurrency bugs.


Sources/NIOConcurrencyHelpers/LockedValue.swift
```diff
@@ -0,0 +1,53 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) YEARS Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
```

Sources/NIOConcurrencyHelpers/LockedValue.swift
```diff
@@ -0,0 +1,53 @@
...
+    internal var _storage: _Storage
+
+    /// Initialize the `Value`.
+    public init(_ value: Value) {
+        self._storage = _Storage(_value: value)
+    }
+
+    /// Access the `Value`, allowing mutation of it.
+    @inlinable
+    public func withNIOLockedValueBox<T>(_ mutate: (inout Value) throws -> T) rethrows -> T {
```
dnadoba, line 48: I don't really like the name of this method. I think the NIO prefix is not need here as the type itself is already prefixed with NIO. I propose we rename it to just `withValue` or `withLockedValue`.

Sources/NIOConcurrencyHelpers/LockedValue.swift
```diff
@@ -0,0 +1,53 @@
...
+/// Provides locked access to `Value`.
+///
+/// - note: ``NIOLockedValueBox`` has reference semantics and holds the ``Value``
+///         alongside a lock behind a reference.
+///
+/// This is no different than creating a ``Lock`` and protecting all
+/// accesses to a value using the lock. But it's easy to forget to actually
+/// acquire/release the lock in the correct place. ``NIOLockedValueBox`` makes
+/// that much easier.
+public struct NIOLockedValueBox<Value> {
```
dnadoba, line 24: `NIOLockedValueBox` should conform to `Sendable`.

dnadoba (commented):
I'm a bit late to the party but we haven't shipped this yet so we can still change things quite easily. Happy to do the changes if we agree.

Sources/NIOConcurrencyHelpers/LockedValue.swift
Lukasa: Fine by me.

Sources/NIOConcurrencyHelpers/LockedValue.swift
Lukasa: :+1: Agree.

Sources/NIOConcurrencyHelpers/LockedValue.swift
weissi: @dnadoba whoops, thanks for catching! I renamed this a bunch so I actually wanted `withLockedValue` but too much find & replace made it this atrocity. Fixing.

Sources/NIOConcurrencyHelpers/LockedValue.swift
weissi: it does because it's a `struct` with only sendable members but I'll make it explicit.

Sources/NIOConcurrencyHelpers/LockedValue.swift
weissi: addressed in https://github.com/apple/swift-nio/pull/2270

PR #2266, opened 2022-09-08
rename class Lock to struct NIOLock

weissi:
### Motivation:

It's silly that we allocate _twice_ for each `Lock` because we should just tail-allocate the `pthread_mutex_t`. Unfortunately, that's impossible without changing the API for the worse (`Lock.makeLock()` or something weird and no options left for the existing `Lock.init()`).

Also `Lock` is a type that can very easily clash with another library.

### Modifications:

- deprecate `Lock`
- introduce `public struct NIOLock` (backed by a `class Storage` for now)

### Result:

- We can file a subsequent issue (easy to implement) which will tail-allocate the `pthread_mutex_t` and we can cut in half all `NIOLock` allocations.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOLock.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOLock.swift
@@ -1,0 +1,148 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2022 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+import Darwin
+#elseif os(Windows)
+import ucrt
+import WinSDK
+#else
+import Glibc
+#endif
+
+/// A threading lock based on `libpthread` instead of `libdispatch`.
+///
+/// - note: ``NIOLock`` has reference semantics.
+///
+/// This object provides a lock on top of a single `pthread_mutex_t`. This kind
+/// of lock is safe to use with `libpthread`-based threading models, such as the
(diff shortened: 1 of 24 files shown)


Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -0,0 +1,127 @@
...
+import Glibc
+#endif
+
+/// A threading lock based on `libpthread` instead of `libdispatch`.
+///
+/// This object provides a lock on top of a single `pthread_mutex_t`. This kind
+/// of lock is safe to use with `libpthread`-based threading models, such as the
+/// one used by NIO. On Windows, the lock is based on the substantially similar
+/// `SRWLOCK` type.
+public struct NIOLock {
```
Lukasa, line 30: Can we add to this comment that this type has reference semantics?

Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// Whenever possible, consider using `withLock` instead of this method and
+    /// `unlock`, to simplify lock handling.
+    public func lock() {
+#if os(Windows)
+        AcquireSRWLockExclusive(self._storage.mutex)
+#else
+        let err = pthread_mutex_lock(self._storage.mutex)
+        precondition(err == 0, "\(#function) failed in pthread_mutex with error \(err)")
+#endif
```
Lukasa, line 89: Can we delegate the methods in here to methods on `self._storage`? Swift plays a bit fast and loose with the guarantees around object lifetimes, but in principle we could accidentally `deinit` the lock here after loading it from `_storage`, whereas `self` is generally guaranteed to be alive.

Lukasa (changes requested):

Lukasa:

Lukasa (approved):
Nice one, LGTM.

Lukasa (approved):

Lukasa:
A couple of deprecation warnings are preventing us merging this.

dnadoba (approved):

toluRV:
@weissi @Lukasa  This merge seems to be causing build errors on windows. I created a PR here: https://github.com/apple/swift-nio/pull/2278

PR #2270, opened 2022-09-20
address additional NIOLockedValueBox review comments

weissi:
### Modifications:

Fix additional review comments in https://github.com/apple/swift-nio/pull/2265#discussion_r975563309

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
@@ -36,7 +36,7 @@ public struct NIOLockedValueBox<Value> {
     }
     
     @usableFromInline
-    internal var _storage: _Storage
+    internal let _storage: _Storage
 
     /// Initialize the `Value`.
     public init(_ value: Value) {
@@ -45,9 +45,14 @@ public struct NIOLockedValueBox<Value> {
 
     /// Access the `Value`, allowing mutation of it.
     @inlinable
-    public func withNIOLockedValueBox<T>(_ mutate: (inout Value) throws -> T) rethrows -> T {
+    public func withLockedValue<T>(_ mutate: (inout Value) throws -> T) rethrows -> T {
         return try self._storage._lock.withLock {
             return try mutate(&self._storage._value)
         }
     }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension NIOLockedValueBox: Sendable where Value: Sendable {}
+extension NIOLockedValueBox._Storage: @unchecked Sendable {}
+#endif
--- a/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift
+++ b/Tests/NIOConcurrencyHelpersTests/NIOConcurrencyHelpersTests.swift
@@ -1016,13 +1016,13 @@ class NIOConcurrencyHelpersTests: XCTestCase {
         let lv = NIOLockedValueBox<State>(State())
         spawnAndJoinRacingThreads(count: 50) { _ in
             for _ in 0..<1000 {
-                lv.withNIOLockedValueBox { state in
+                lv.withLockedValue { state in
                     state.count += 1
                 }
             }
         }
 
-        XCTAssertEqual(50_000, lv.withNIOLockedValueBox { $0.count })
+        XCTAssertEqual(50_000, lv.withLockedValue { $0.count })
     }
 }
 


weissi:
@dnadoba API breakage is obvs expected, just FYI.

Lukasa (approved):

Sources/NIOConcurrencyHelpers/LockedValue.swift
```diff
@@ -45,9 +45,15 @@ public struct NIOLockedValueBox<Value> {
...
         return try self._storage._lock.withLock {
             return try mutate(&self._storage._value)
         }
     }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension NIOLockedValueBox: Sendable {}
+
+extension NIOLockedValueBox._Storage: @unchecked Sendable {}
```
dnadoba, line 58: `NIOLockedValueBox` can only be `Sendable` if `Value` conforms to `Sendable`.

and a nit: `NIOLockedValueBox._Storage` isn't actually `Sendable` as the interface provides unprotected access to `value`. Only `NIOLockedValueBox` is really `Sendable`.
This doesn't matter that much as `_Storage` is not public and meant to be an implementation detail anyway.
```suggestion
extension NIOLockedValueBox: @unchecked Sendable where Value: Sendable {}
```

dnadoba (changes requested):
Really happy with the new method name! Thanks!

Sources/NIOConcurrencyHelpers/LockedValue.swift
weissi: The point of `NIOLockedValueBox` is to always be sendable, crucially also if `Value` is not sendable.

weissi: @dnadoba so your suggestion isn’t correct. I understand your nit pick that the Storage isn’t actually sendable (which is correct because - as you point out - if provides unprotected access to the value) but it’s underscored which means “private” so I don’t think that’s an issue. If you really want to have the unchecked one level up, then I can do that but don’t really see the point.

dnadoba: I don't agree. I think the point of `NIOLockedValueBox` is to make a `Sendable` value type (e.g. an `Int`) into a shared but synchronized reference type.

`NIOLockedValueBox` doesn't make a non-`Sendable` reference type `Sendable`.
One reason is that it might not hold the only reference to it. One unsafe example is a `Counter` class:
```swift
class Counter {
    var count: Int
    init(count: Int) {
        self.count = count
    }
}

let counter = Counter(count: 0)
let counterLocked = NIOLockedValueBox(counter)
Task {
    counterLocked.withLockedValue {
        $0.count += 1
    }
}
counter.count += 1
```
This is a data race but compiles with your current implementation. If we make `NIOLockedValueBox` conditionally `Sendable` the compiler catches this mistake and produces a warning (with `-warn-concurrency`)
```
Capture of 'counterLocked' with non-sendable type 'NIOLockedValueBox<Counter>' in a `@Sendable` closure
```
---
If you really want to go around the `Sendable` rules, you still can! However, you will need to wrap your non-`Sendable` type in an `UnsafeTransfer` type which is unconditionally `Sendable`. This type was mentioned in [SE-0302](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md#adaptor-types-for-legacy-codebases) and we use it `swift-nio` too:
https://github.com/apple/swift-nio/blob/26afcecdc2142f1cd0d9b7f4d25b3a72938c3368/Sources/NIOCore/NIOSendable.swift#L27-L46
It is a bit more to type but this is the usual tradeoff in Swift if you want to do unsafe things.

dnadoba: After thinking about this a bit more, I think we might actually want to require `Value` to be `Sendable` (i.e. `struct NIOLockedValueBox<Value: Sendable> { ... }`). `NIOLockedValueBox` can otherwise not cross an isolation domain and would not require any synchronization through locking. In this cases `NIOLockedValueBox` would not be useful and the plain `Value` type can be used instead.

[For reference, `Task` requires both of its generic arguments to be `Sendable` too.](https://developer.apple.com/documentation/swift/task)

weissi: @dnadoba Yes, you can use `NIOLockedValueBox` to send an otherwise not sendable reference across threads. And with that functionality, you can create data races but only if _you_ escape the reference (either from the `withLockedValue` or before `init` like you do in your example). In other words, if you bypass the `NIOLockedValueBox`.

So yes, if it gets misused, it can be used to disable Swift's sendable checking. I'm happy to add more commentary.

Again, requiring `Sendable` from `Value` makes this vastly less useful. Yes, we could force users to combine it as a `NIOLockedValueBox<UnsafeTransfer<...>>` but that doesn't feel right either because the place where you have the `Unsafe` isn't actually the unsafe place :).

FranzBusch: I agree with @weissi here that requiring `Sendable` here is not very helpful because you often want to use this type to make things `@unchecked Sendable`.

weissi: Right, the job of this class is pretty much to turn

```
class Foo: @unchecked Sendable {
   var something: RandomType
   let lock: Lock
}
```

into

```
class Foo: @unchecked Sendable {
   var something: NIOLockedValueBox<RandomType>
}
```

to prevent mistakes like

```
class Foo: @unchecked Sendable {
   var something: RandomType
   let lock: Lock

   var someThingsColor: Colour {
       return self.something.colour // oops, forgot the lock
   }
}
```

weissi: We could also just strip the `Sendable` entirely actually.

FranzBusch: I don't know about that what if you just create a local `LockedValueBox` that you pass to two `Task`s without `self`, e.g.:

```swift
let box = ...

Task {
    box.withValue {}
}

Task {
    box.withValue {}
}
```

IMO having this conform to `@unchecked Sendable` is correct.

FranzBusch: @Lukasa Yeah you are absolutely right, after thinking about this again even `@unchecked Sendable` is wrong. I guess the conditional conformance is the best we can do at this point.

weissi: Let's just strip `Sendable` here. As mentioned above, it is correct (although unlikely) that this type _can_ be mis-used w.r.t. sendability because it allows to absorb a reference.

Conceptually, `NIOLockedValueBox` is 

> - Reference types that internally manage access to their state

_but_ it cannot finally police that its managed state never leaves the domain.

Given that where you want to use this, you'll need an `@unchecked Sendable` anyway we don't need `Sendable` here, I stripped it.

Everybody okay with this?

Lukasa: WFM.

Lukasa (approved):

Sources/NIOConcurrencyHelpers/LockedValue.swift
dnadoba: We don't need `@unchecked Sendable` if the wrapped type is `Sendable`. E.g. if we slightly transform your example from above:
```swift
class Foo: @unchecked Sendable {
   var something: NIOLockedValueBox<RandomType>
}
```
assuming `RandomType` is Sendable, we can make `Foo` final and `something` a let:
```swift
final class Foo: Sendable {
    let something: NIOLockedValueBox<Int>
    init(something: NIOLockedValueBox<Int>) {
        self.something = something
    }
}
```
The compiler is now smart enough to automatically figure out that this type is `Sendable` and we don't need `@unchecked` anymore. This is because `Foo` is now a
> - Reference types with no mutable storage
---
@weissi what use case do you have in mind where you want to use `NIOLockedValueBox` with a non-`Sendable` `Value`? 

I have quickly looked through AHCs usage of Locks and it only protects `Sendable` values (with the exception of `HTTPClientTaskDelegate` in `HTTPClient.Task`. However, `HTTPClientTaskDelegate` should and can inherit from `Sendable` as `cancel()` is called from an arbitrary `EventLoop` and all implementations are already thread safe). 

In one instance we actually need `NIOLockedValueBox` to be `Sendable`, otherwise we can't replace our unsafe usage of a `Lock` and `UnsafeMutableTransferBox` here:

https://github.com/swift-server/async-http-client/blob/7f998f5118de081fa2034833ce95bb5925dc6459/Sources/AsyncHTTPClient/HTTPClient.swift#L172-L188

weissi: > assuming RandomType is Sendable,

usually not

> what use case do you have in mind where you want to use NIOLockedValueBox with a non-Sendable Value

for example, check the `SSLContextCache` in AHC. The `LRUCache` isn't `Sendable`, could it be? Probably. We'd need to require both `Key` and `Value` to be `Sendable`. Is it worth doing that? Maybe, probably not.

> In one instance we actually need NIOLockedValueBox to be Sendable, otherwise we can't replace our unsafe usage of a Lock and UnsafeMutableTransferBox

Yes, there are use cases for a standalone box that can safely transport `Sendable` values across threads. If you feel strongly about it, I can make `NIOLockedValueBox` sendable or you make a follow-up PR to do it so. How I personally envision to use this it only makes sense to have it either always sendable (including non-Sendable value) or not at all. Most of my `Value`s in there are internal types and aren't `Sendable` and I don't want to make them `Sendable` for two reasons:

1. it's work
2. there might be a house of cards collapsing later if I have to later add something non-Sendable (which I consider to be very likely) to my `Value`. Then I have to strip the `Sendable` from my `Value` which then leads to me having to add `@unchecked Sendable` to the classes holding the `NIOLockedValueBox` and all other types containing that. I'd rather decide at the beginning that _I_ am taking responsibility for Sendability from the get go for a given class by using a `NIOLockedValueBox`. 

But I'm not disputing that there are neat uses like the one you linked where you happen to have a `Sendable` value (`Error?` in your case) and you want to use `NIOLockedValueBox` as a safe transporter, completely separate the the `class Foo: @unchecked Sendable { let state: NIOLockedValueBox<NonSendableState> }` use case that I'm thinking.

Lukasa: The `SSLContextCache` is a perfect example of why this should be given the conditional `Sendable` conformance. By its very definition, SSLContext must be `Sendable`: otherwise we couldn't use it across multiple threads (and indeed it is). Similarly, `TLSConfiguration` is a pure value, and so is also trivially `Sendable` (and indeed it is). More broadly, to do the job it wants to do, `SSLContextCache` must be able to send keys and values across concurrency domains, and that implies `Sendable`.

weissi: Absolutely, but we can't rely on SSLContextCache's state to be or remain `Sendable` by automatic compiler inference.

For example, `LRUCache` isn't `Sendable` from what I can tell. Yes, _today_ we can make `LRUCache` `Sendable` ( if Key/Values are `Sendable`) but I don't think that adds a lot of value. Because LRUCache could see a refactoring which makes it no longer sendable, e.g. it could become a `class` or so.

Anyway, as I said before: If you feel strongly, I can add it here or somebody makes a follow up.

weissi: Expressed differently: I don't _want_ `LRUCache` to be `Sendable`, even if it could be [today]. If somebody relied on its sendability, they're likely doing a bad thing.

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
```diff
@@ -45,9 +45,14 @@ public struct NIOLockedValueBox<Value> {
...
+    public func withLockedValue<T>(_ mutate: (inout Value) throws -> T) rethrows -> T {
         return try self._storage._lock.withLock {
             return try mutate(&self._storage._value)
         }
     }
 }
+
+#if compiler(>=5.5) && canImport(_Concurrency)
+extension NIOLockedValueBox: Sendable where Value: Sendable {}
+extension NIOLockedValueBox._Storage: @unchecked Sendable {}
```
dnadoba, line 57: Sorry for nitpicking but is there any benefit making `_Storage` `@unchcked Sendable` instead of moving the `@unchecked` to the conditional conformance above and leaving `_Storage` `non-Sendable`?
I'm fine with either way as it is internal. Just merge it as we can revisit this at any time but I'm just curious

dnadoba (approved):
Thanks @weissi for this new type! I think this will make our code a lot safer!

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
weissi: thought I should use of the automatic inference where I can. Just in the (_very unlikely_...) case that somebody accidentally adds something else to `NIOLockedValueBox`, the compiler would now barf :P. It won't happen but you know.

weissi:

dnadoba:
@weissi FYI: nightly failure is expected because of a compiler issue https://github.com/apple/swift/issues/60797

Issue #2275, opened 2022-10-04
`NIOLock` missing `withLockVoid(_:)`


MahdiBM:
`Lock` has been recently deprecated in favor of the new `NIOLock`.
The warnings mention that "`Lock` has been renamed to `NIOLock`". Considering `Lock` had `withLockVoid(_:)`, I would expected `NIOLock` to have that function as well, but it's missing.

I'm not sure if this is considered "api breaking" or not, but it was an unexpected change to me since the warnings claimed that this is only a "rename".

Lukasa:
Yeah, this isn't an API break but it's an unfortunate oversight. Would you like to open a PR to add the function?

MahdiBM:
Surely. Thanks for the answer.

adam-fowler:
Yeah just found this when moving across to NIOLock in Soto

PR #2276, opened 2022-10-04
Add `withLockVoid(_:)` to `NIOLock`

MahdiBM:
### Motivation:

`Lock` has recently been deprecated in favor of `NIOLock`.
The warnings claim that this is supposed to be a mere rename, but the API of `NIOLock` misses `withLockVoid(_:)` compared to `Lock`, which is unexpected to some users.

### Modifications:

`NIOLock` now has the `withLockVoid(_:)` too.

### Result:

`Lock` and `NIOLock` will have a more similar API and users can move to `NIOLock` easier.
This will also resolve https://github.com/apple/swift-nio/issues/2275.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOLock.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOLock.swift
@@ -140,6 +140,12 @@ extension NIOLock {
         }
         return try body()
     }
+
+    // specialise Void return (for performance)
+    @inlinable
+    public func withLockVoid(_ body: () throws -> Void) rethrows -> Void {
+        try self.withLock(body)
+    }
 }
 
 #if compiler(>=5.5) && canImport(_Concurrency)


Lukasa (approved):
Thanks for this!

Lukasa:

Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -140,6 +140,12 @@ extension NIOLock {
         }
         return try body()
     }
+
+    // specialise Void return (for performance)
```
dnadoba, line 144: Why is this more performant? This just calls down to `self.withLock(_:)`.

MahdiBM: I wouldn't know, i just copy pasted the exact function that `Lock` had, including the comment above it 😅

dnadoba: @Lukasa should we remove this doc comment?

Lukasa: It's a regular comment, not a doc comment. You can remove it if you'd like.

Lukasa: The comment likely stopped being true when the method was marked `@inlinable`.

weissi: Why don't we just suggest to use `withLock`? `withLockVoid` is pointless since about Swift 4.2 or so when `@inlinable` became available publicly.

Lukasa: The honest answer is that I don’t think it’s worth the churn. While `withLockVoid` isn’t meaningfully better, plenty of existing code uses it and it makes the deprecation more painful than it needs to be.

Lukasa: (Also it doesn’t totally seamlessly get replaced by `withLock`, due to single statement return value inference. That can cause you to need to explicitly discard the result, which this avoids.)

Issue #2274, opened 2022-10-01
Cannot parse HTTP responses after parsing HEAD response


adam-fowler:
### Expected behavior
Expect parsing responses from HEAD requests to not break subsequent HTTP requests.

### Actual behavior
After receiving response from a HEAD request, subsequent requests fail with the error `failed: caught error: "stream ended at an unexpected time"`.

### Steps to reproduce
The code below use AsyncHTTPClient to replicate the issue, but I have seen this with the Hummingbird test HTTP client as well

### If possible, minimal yet complete reproducer code (or URL to code)

```swift
let client = HTTPClient(eventLoopGroupProvider: .createNew)
_ = try client.execute(.HEAD, url: "http://httpbin.org/get").wait()
_ = try client.get(url: "http://httpbin.org/get").wait()
```

### SwiftNIO version/commit hash

Release 2.42.0
Although it looks like the responsible commit is the update to the LLHTTP parser 6918034260d1b0ff46a24a6e3ced445bc2d0ed4b. I couldn't test this with AHC though as the latest AHC requires the NIOLock commit. I have used Hummingbird to verify this if the commit responsible for the issue.

### System & version information

Darwin Adams-MBP-M1-Max.local 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64
swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: arm64-apple-macosx12.0

Lukasa:
Thanks for filing this, I'm going to dig into it today.

Lukasa:
Ok this appears to be an llhttp issue: nodejs/llhttp#202. In the meantime, I've got a targeted temporary workaround that I plan to push up shortly.

adam-fowler:
Thought it might be related to content-length in a HEAD response. Can't believe this wasn't caught earlier

Lukasa:
This issue now tracks unwinding our temporary fix when the upstream addresses their issue.

Lukasa:
Upstream has been fixed, we can do this now.

PR #2277, opened 2022-10-04
Correctly manage Content-Length on HEAD responses

Lukasa:
Motivation

When we receive a HEAD response, it's possible that the response contains a content-length. llhttp has a bug
(https://github.com/nodejs/llhttp/issues/202) that prevents it from properly managing that issue, which causes us to incorrectly parse responses.

Modifications

Forcibly set llhttp's content-length value to 0.

Result

Correctly handle HTTP framing around llhttp's issues.

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -113,6 +113,9 @@ private class BetterHTTPParser {
         }
         self.settings.pointee.on_message_complete = { opaque in
             BetterHTTPParser.fromOpaque(opaque).didReceiveMessageCompleteNotification()
+            // Temporary workaround for https://github.com/nodejs/llhttp/issues/202, should be removed
+            // when that issue is fixed.
+            opaque?.pointee.content_length = 0
             return 0
         }
         self.withExclusiveHTTPParser { parserPtr in
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -114,7 +114,7 @@ private class BetterHTTPParser {
         self.settings.pointee.on_message_complete = { opaque in
             BetterHTTPParser.fromOpaque(opaque).didReceiveMessageCompleteNotification()
             // Temporary workaround for https://github.com/nodejs/llhttp/issues/202, should be removed
-            // when that issue is fixed.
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -113,6 +113,9 @@ private class BetterHTTPParser {
         }
         self.settings.pointee.on_message_complete = { opaque in
             BetterHTTPParser.fromOpaque(opaque).didReceiveMessageCompleteNotification()
+            // Temporary workaround for https://github.com/nodejs/llhttp/issues/202, should be removed
+            // when that issue is fixed.
+            opaque?.pointee.content_length = 0
```
FranzBusch, line 118: Can we track this in one of our GH issues as well?

FranzBusch (approved):
LGTM!

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: Can we repurpose #2274 to track this?

FranzBusch: We sure can!

dnadoba (approved):

fabianfett:
@Lukasa Should we add this to the workarounds.md?

Lukasa:
Yes. I was hoping that upstream would turn around a fix quickly, but it seems like they're not in a rush to do so.

Lukasa:
Just shipped as 2.43.0.

adam-fowler:
Sorry to be annoying but can we get this in a v2.42.1. Builds dependent on swift 5.4 are currently broken as the fix is not available for 5.4. I know you've moved on from supporting 5.4 but it would be good to leave the build in a working state (as much as you can know about).

Lukasa:
How many builds are dependent on 5.4? Swift 5.4 is by this point 18 months old and I'd have expected users to move forward to at least 5.5 by now.

I'm not _strongly_ opposed to back porting this fix, but in general we are not going to continue to backport every bugfix to 5.4. That's what it means for us to have dropped support.

Lukasa:
See #2289.

adam-fowler:
I understand that, but this was a fairly sizeable component that was replaced just before you dropped a Swift version.
Regarding number of builds dependent on 5.4 I don't know.

adam-fowler:
thanks

Lukasa:
Shipped as 2.42.1.

PR #2282, opened 2022-10-06
Ensure fatalError for the NIOAsyncSequenceProducer when finished

FranzBusch:
# Motivation
Our `NIOAsyncSequenceProducer` is a unicast `AsyncSequence`; therefor, we must ensure that only a single iterator is every created. Our failure mode in the case another iterator is created is to `fatalError`. Currently, this `fatalError` is not produced if the sequence is in the `finished` state. This results in hard to debug behaviour since the user will get a basically useless iterator.

# Modification
Ensure that the `finished` state also keeps track of if an iterator was initialised and produces the correct `fatalError`.

# Result
We are now consistent in how many iterators can be created for every state.


Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -643,7 +644,7 @@ extension NIOThrowingAsyncSequenceProducer {
                     failure: failure
                 )
 
-            case .finished:
+            case .finished(iteratorInitialized: false):
                 // It is strange that an iterator is created after we are finished
                 // but it can definitely happen, e.g.
                 // Sequence.init -> source.finish -> sequence.makeAsyncIterator
```
fabianfett, line 650: I think we should change the state here, shouldn't we?

FranzBusch: Yep good catch. So annoying that you can't write tests for `fatalError`s easily.

Lukasa (approved):

PR #2284, opened 2022-10-07
Get NIO compiling with GM Xcode

### LINKED ISSUE #2279: _NIOBeta doesn't compile with Swift 5.7
_NIOBeta is assuming `Swift.Duration` is a valid symbol in Swift 5.7 when it is not.

### System and version information
swift-nio: [dee448a](https://github.com/apple/swift-nio/commit/dee448ac8b7dd0f80a48bc0b2e59aec68c651b88)
swift: swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: arm64-apple-macosx12.0

Lukasa:
Motivation

It's good if we compile with the GM Xcode.

Modifications

Guard Duration usability to at least Swift 5.7.1.

Result

NIO compiles on GM Xcode again.

Note that this is technically an API break for those using Xcode 14 Betas. As those users can (and should) update to 14.1 Beta, and we don't define API stability for Beta xcodes, this is acceptable.
Resolves #2279.

### DIFF
--- a/Sources/_NIOBeta/TimeAmount+Duration.swift
+++ b/Sources/_NIOBeta/TimeAmount+Duration.swift
@@ -14,7 +14,7 @@
 
 import NIOCore
 
-#if swift(>=5.7)
+#if (os(macOS) && swift(>=5.7.1)) || (!os(macOS) && swift(>=5.7))
 extension TimeAmount {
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):
LGTM!

PR #2287, opened 2022-10-11
Improve `NIOAsyncSequenceProducer` docs

FranzBusch:
# Motivation
Using the `NIOAsyncSequenceProducer` requires a bunch of knowledge around when and how the delegate is getting called to ensure a correct back-pressure implementation. We should enhance the docs a bit more to surface some of the invariants that we expect.

# Modification
Extend the docs for `NIOAsyncSequenceProducer`.

# Result
Better docs.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
@@ -47,16 +47,22 @@ public protocol NIOAsyncSequenceProducerBackPressureStrategy: Sendable {
 }
 
 /// The delegate of ``NIOAsyncSequenceProducer``.
+///
+/// - Important: The calls to ``NIOAsyncSequenceProducerDelegate/produceMore()`` and ``NIOAsyncSequenceProducerDelegate/didTerminate()``
+/// are being done on arbitrary threads. To ensure that your conforming type is able to implement back-pressure correctly your must synchronize
+/// your calls to ``NIOAsyncSequenceProducer/Source/yield(contentsOf:)`` and callbacks on this delegate.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public protocol NIOAsyncSequenceProducerDelegate: Sendable {
     /// This method is called once the back-pressure strategy of the ``NIOAsyncSequenceProducer`` determined
-    /// that the producer needs to produce more elements.
+    /// that the producer needs to start producing more elements.
     ///
     /// - Note: ``NIOAsyncSequenceProducerDelegate/produceMore()`` will never be called after
     /// ``NIOAsyncSequenceProducerDelegate/didTerminate()`` was called.
     ///
     /// - Important: This is only called as a result of the consumer calling ``NIOAsyncSequenceProducer/AsyncIterator/next()``.
     /// It is never called as a result of a producer calling ``NIOAsyncSequenceProducer/Source/yield(_:)``.
+    /// Furthermore, it will also only be called if ``NIOAsyncSequenceProducer/Source/yield(_:)`` returned a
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -244,7 +250,8 @@ extension NIOAsyncSequenceProducer {
 
         /// The result of a call to ``NIOAsyncSequenceProducer/Source/yield(_:)``.
         public enum YieldResult: Hashable {
-            /// Indicates that the caller should produce more elements.
+            /// Indicates that the caller should produce more elements for now. The delegate's ``NIOAsyncSequenceProducerDelegate/produceMore()``
+            /// method will get called once production should be resumed.
             case produceMore
             /// Indicates that the caller should stop producing elements.
```
fabianfett, line 256:
```suggestion
            /// Indicates that the caller should produce more elements for now. The delegate's ``NIOAsyncSequenceProducerDelegate/produceMore()`` will not get called, since the demand was already signaled through this ``YieldResult``
            case produceMore
            /// Indicates that the caller should stop producing elements. The delegate's ``NIOAsyncSequenceProducerDelegate/produceMore()``
            /// method will get called once production should be resumed.
```

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -47,16 +47,22 @@ public protocol NIOAsyncSequenceProducerBackPressureStrategy: Sendable {
...
-    /// that the producer needs to produce more elements.
+    /// that the producer needs to start producing more elements.
     ///
     /// - Note: ``NIOAsyncSequenceProducerDelegate/produceMore()`` will never be called after
     /// ``NIOAsyncSequenceProducerDelegate/didTerminate()`` was called.
     ///
     /// - Important: This is only called as a result of the consumer calling ``NIOAsyncSequenceProducer/AsyncIterator/next()``.
     /// It is never called as a result of a producer calling ``NIOAsyncSequenceProducer/Source/yield(_:)``.
+    /// Furthermore, it will also only be called if ``NIOAsyncSequenceProducer/Source/yield(_:)`` returned a
+    /// ``NIOAsyncSequenceProducer/Source/YieldResult/stopProducing`` to indicate that new elements should be produced now.
```
fabianfett, line 65: Can we move this into the first paragraph? I consider this very important.

fabianfett (changes requested):
The last comment on YieldResult got onto the wrong case.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -47,10 +47,16 @@ public protocol NIOAsyncSequenceProducerBackPressureStrategy: Sendable {
 }
 
 /// The delegate of ``NIOAsyncSequenceProducer``.
+///
+/// - Important: The calls to ``NIOAsyncSequenceProducerDelegate/produceMore()`` and ``NIOAsyncSequenceProducerDelegate/didTerminate()``
+/// are being done on arbitrary threads. To ensure that your conforming type is able to implement back-pressure correctly your must synchronize
+/// your calls to ``NIOAsyncSequenceProducer/Source/yield(contentsOf:)`` and callbacks on this delegate.
```
fabianfett, line 53: Do we want to add?

> We recommend dispatching from the arbitrary thread that called `produceMore` or `didTerminate` onto the thread that is calling source.yield. This way you synchronize the receiving the YieldResult after a yield call and `produceMore` and `didTerminate` on the same thread.

fabianfett (approved):
Awesome! Thank you!

PR #2291, opened 2022-10-13
Replace `NIOSendable` with `Sendable`

dnadoba:
We now require Swift 5.5.2 which includes `Sendable` so we no longer require the `NIOSendable` workaround.
`NIOSendable` is just a `typealias` of `Sendable` so this is a non source breaking change.

We have already replaced the usages of `NIOSendable` in all of our packages and we can now deprecate `NIOSendable` altogether.


Lukasa:

PR #2301, opened 2022-10-25
Add benchmarks for `NIOAsyncWriter` and `NIOAsyncSequenceProducer`

FranzBusch:
# Motivation
We landed the async bridge types a while back but never added allocation and performance tests. Since we expect these types to be used performance critical paths we really should cover those with tests.

# Modification
Extends the allocation counter scaffolding to support async tests. Furthermore, add allocations tests for both the writer and producer. Lastly, I a also added a performance test for the producer.

# Result
We now have baseline tests for the `NIOAsyncWriter` and `NIOAsyncSequenceProducer`


IntegrationTests/allocation-counter-tests-framework/template/scaffolding.swift
```diff
@@ -99,3 +99,69 @@ public func measure(identifier: String, _ body: () -> Int) {
...
+        AtomicCounter.reset_malloc_bytes_counter()
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+        autoreleasepool {
+            let group = DispatchGroup()
+            group.enter()
+            Task {
+                _ = await fn()
+                group.leave()
+            }
+            group.wait()
```
fabianfett, line 117: do you maybe want to factor this into an internal run method?

Lukasa (approved):

PR #2302, opened 2022-10-25
Add correct C directory function declarations for Android

buttaface:
### Motivation

Get the Android build working again

### Modifications

- Modify close/open/readdir arguments, as they can be null
- Remove mkpath_np, as it's not there on Android

### Result

Android builds again and the same tests pass

#2244 [broke building NIO on my Android CI today](https://github.com/buttaface/swift-android-sdk/actions/runs/3319673203/jobs/5485126146#step:13:154), this should fix it.


buttaface:
Nightly CI failure is unrelated to this patch.

PR #2303, opened 2022-10-28
Don't unconditionally remove the HTTPServerUpgradeHandler

glbrntt:
Motivation:

The `HTTPServerUpgradeHandler` removes itself from the pipeline after an upgrade and unbuffers any unconsumed reads. If an upgrade starts but does not complete successfully then the pipeline may be left in an unknown state.

If, for example, the failure occurs before the user provided upgrade handler is run then unbuffered writes may be unwrapped by the wrong channel handler as the wrong type leading to a crash.

Modifications:

- Only remove the upgrade handler and forward buffered writes if all parts of the upgrade complete successfully.
- If part of the upgrade fails then fire an error into the channel pipeline without removing the server upgrade handler.
- Remove a few unnecessary `map`s.
- Make `httpEncoder` non-optional since and remove associated dead code since it can never be `nil`.

Result:

Upgrade handling is safer.

### DIFF
--- a/Sources/NIOHTTP1/HTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/HTTPServerUpgradeHandler.swift
@@ -66,7 +66,7 @@ public final class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableCha
     private let upgraders: [String: HTTPServerProtocolUpgrader]
     private let upgradeCompletionHandler: (ChannelHandlerContext) -> Void
 
-    private let httpEncoder: HTTPResponseEncoder?
+    private let httpEncoder: HTTPResponseEncoder
     private let extraHTTPHandlers: [RemovableChannelHandler]
 
     /// Whether we've already seen the first request.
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP1/HTTPServerUpgradeHandler.swift
```diff
@@ -235,17 +235,22 @@ public final class HTTPServerUpgradeHandler: ChannelInboundHandler, RemovableCha
                 self.removeExtraHandlers(context: context).flatMap {
                     self.sendUpgradeResponse(context: context, upgradeRequest: request, responseHeaders: finalResponseHeaders)
                 }.flatMap {
-                    self.removeHandler(context: context, handler: self.httpEncoder)
-                }.map {
+                    context.pipeline.removeHandler(self.httpEncoder)
+                }.flatMap { () -> EventLoopFuture<Void> in
```
glbrntt, line 239: The semantics here are slightly awkward: 
- `self.upgradeCompletionHandler(_:)` can fire followed by
- `upgrader.upgrade(context:upgradeRequest:)` which can fail

I think this is okay though: `upgradeCompletionHandler` is documented as being called after the HTTP upgrade is complete (which it is: the upgrade response was successfully sent.) This is distinct from the pipeline upgrade completing which is signalled by `HTTPServerUpgradeEvents.upgradeComplete` so I think this is fine, if a little unclear.

Lukasa (commented):
I'm not sure if this has helped: the pipeline is still left in an undefined state. What's the specific problem we're trying to solve?

Lukasa (approved):
Oh, apologies, on re-reading this I see that the main effect is actually to fire an error in the pipeline. This is substantially more useful.

PR #2304, opened 2022-10-28
Make `NIOWebSocketServerUpgrader` `Sendable`

glbrntt:
Motivation:

The `NIOWebSocketServerUpgrader` is marked as not being `Sendable`, but it has no mutable state and makes no assumptions about the event loop it is invoked on, so it should be `Sendable`. The two user-provided callbacks which are injected in `init` are marked as being `Sendable` but require Swift 5.7 to express this safely.

Modifications:

- Add `@unchecked Sendable` conformance to `NIOWebSocketServerUpgrader`

Result:

- `NIOWebSocketServerUpgrader` is Sendable.

### DIFF
--- a/Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
+++ b/Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
@@ -61,7 +61,10 @@ fileprivate extension HTTPHeaders {
 ///
 /// This upgrader assumes that the `HTTPServerUpgradeHandler` will appropriately mutate the pipeline to
 /// remove the HTTP `ChannelHandler`s.
-public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader {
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

weissi:
@glbrntt what's the motivation for this? Just because the current implementation _can be Sendable_ should mean that this _should_ be Sendable. In general, ChannelHandlers aren't sendable and this will require all future implementations to be stateless too.

dnadoba:
We have used it in our `NIOWebSocketServer` example as if it was already `Sendable` because we share it between all channels:
https://github.com/apple/swift-nio/blob/558e4f2fb854dacc22d1774e7b6528d06a4667ad/Sources/NIOWebSocketServer/main.swift#L203-L224

Making it non-`Sendable` would break this example and any code that is in the wild using it like we have recommended since its inception.

PR #2305, opened 2022-11-02
Add missing `Sendable` requirements

dnadoba:
`NIOWebSocketServerUpgrader` is `Sendable` and therefore needs all of its callbacks to be `Sendable` as well. We have already stated that for one `init` but forgot another init due to an [incomplete revert ](https://github.com/apple/swift-nio/pull/2249)and therefore subsequent incomplete re-revert.

### DIFF
--- a/Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
+++ b/Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
@@ -61,12 +61,11 @@ fileprivate extension HTTPHeaders {
 ///
 /// This upgrader assumes that the `HTTPServerUpgradeHandler` will appropriately mutate the pipeline to
 /// remove the HTTP `ChannelHandler`s.
-public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader, @unchecked Sendable {
+public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader {
     // This type *is* Sendable but we can't express that properly until Swift 5.7. In the meantime
     // the conformance is `@unchecked`.
     
     #if swift(>=5.7)
-    // FIXME: remove @unchecked when 5.7 is the minimum supported version.
     private typealias ShouldUpgrade = @Sendable (Channel, HTTPRequestHead) -> EventLoopFuture<HTTPHeaders?>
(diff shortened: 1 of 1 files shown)


Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
```diff
@@ -64,7 +64,15 @@ fileprivate extension HTTPHeaders {
 public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader, @unchecked Sendable {
     // This type *is* Sendable but we can't express that properly until Swift 5.7. In the meantime
     // the conformance is `@unchecked`.
-
+    
+    #if swift(>=5.7)
+    // FIXME: remove @unchecked when 5.7 is the minimum supported version.
```
glbrntt, line 69: Do we still need `@unchecked` or do we say that it's only `Sendable` if Swift >= 5.7?

dnadoba: Let's just say it is `Sendable` in Swift >= 5.7. Done in https://github.com/apple/swift-nio/pull/2305/commits/808638ca747c61c98eea66ad8c8faa1049c6ec47

glbrntt (approved):

Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
dnadoba: hm... thinking about this again, we actually break API because we have made it `Sendable` in Swift >= 5.5 with the latest release. Swift 5.5 is a bit more strict and can produces Sendable errors even without `-warn-concurrency`. I highly doubt that someone hast written new code in the last couple of hours that will now break because they rely on `NIOWebSocketServerUpgrader` being `Sendable` and only use a Swift 5.5 compiler but I would like to get @Lukasa opinion first before we merge.

Lukasa: Let's keep this unchecked and leave the FIXME.

dnadoba: reverted in https://github.com/apple/swift-nio/pull/2305/commits/d1786d06a4b608c4bea80cf4856331bb92d02fc8

Issue #2283, opened 2022-10-06
`NIOAsyncTestingChannel` not working as expected


adam-fowler:
### Expected behavior
`NIOAsyncTestingChannel` works with a channel handler that has a `Task` with a suspension point that lasts longer than a few microseconds

### Actual behavior
If Task suspends for longer than a few microseconds then `NIOAsyncTestingChannel` will return `nil`

### If possible, minimal yet complete reproducer code (or URL to code)

The following ChannelHandler should return a number incremented by 1 but it returns `nil`

```swift
final class AsyncChannelHandler: ChannelDuplexHandler {
    typealias InboundIn = Int
    typealias OutboundIn = Never
    typealias OutboundOut = Int

    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
        let number = self.unwrapInboundIn(data)
        let promise = context.eventLoop.makePromise(of: Int.self)
        promise.completeWithTask {
            // if you reduce this to 1000 it works
            try await Task.sleep(nanoseconds: 1_000_000)
            return number + 1
        }
        promise.futureResult.whenComplete { result in
            switch result {
                case .success(let number):
                    context.writeAndFlush(self.wrapOutboundOut(number), promise: nil)
                case .failure(let error):
                    print("\(error)")
            }
        }
    }
}

func testAsyncTestingChannel() async throws {
    let channel = NIOAsyncTestingChannel()
    try await channel.pipeline.addHandler(AsyncChannelHandler())
    // write to channel
    try await channel.writeInbound(1)
    // run
    await channel.testingEventLoop.run()
    // read from channel
    let outbound = try await channel.readOutbound(as: Int.self)
    print(outbound?.description ?? "nil")
}
```

### SwiftNIO version/commit hash

cdeffe8c5bb88288ee624bb81fefe0508abe5a29

### System & version information

Swift version: swift-driver version: 1.62.8 Apple Swift version 5.7 (swiftlang-5.7.0.127.4 clang-1400.0.29.50)
Target: x86_64-apple-macosx12.0
System: Darwin Adams-iMac-Pro-2.local 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:17:10 PDT 2022; root:xnu-8020.140.49~2/RELEASE_X86_64 x86_64

adam-fowler:
Actually just found you don't need a suspension point. The task just needs to take some amount of time eg
```swift
promise.completeWithTask {
    var number = number
    for i in 0..<10000 {
[...]

PR #2307, opened 2022-11-04
Add `NIOAsyncTestingChannel.waitForOut/InboundWrite()`

dnadoba:
### Motivation:

We want users to be able to wait for an I/O operation to happen.

### Modifications:

- add `EmbeddedChannelCore.in/outboundBufferConsumer` to make it possible to observer buffer appends
- add `NIOAsyncTestingChannel.waitForIn/OutboundWrite()` which will internally use `EmbeddedChannelCore.in/outboundBufferConsumer` to wait for appends if the buffer is empty

### Result:

Users can now wait for data to be appended to the in/outbound buffers.
Resolves https://github.com/apple/swift-nio/issues/2283


Lukasa (approved):
Nice change, this LGTM.

PR #2308, opened 2022-11-05
MPTCP support on Linux

Lukasa:
Motivation

MPTCP provides multipath capability for TCP connections. This allows TCP connections to consume multiple independent network paths, providing devices with a number of capabilities to improve throughput, latency, or reliability.

MPTCP is not totally transparent, and requires servers to support the functionality as well as clients. To that end, we should expose some MPTCP capability.

Importantly, MPTCP uses a number of new socket flags and options. To enable us to support this when it is available but gracefully fail when it is not, we've hardcoded a number of Linux kernel constants instead of relying on libc to expose them. This is safe to do on Linux because its syscall layer is ABI stable.

Modifications

- Add ClientBootstrap and ServerBootstrap flags for MPTCP
- Plumb MPTCP through the stack
- Add new socket options for MPTCP

Result

MPTCP is supported on Linux

### DIFF
--- a/Sources/NIOCore/BSDSocketAPI.swift
+++ b/Sources/NIOCore/BSDSocketAPI.swift
@@ -236,6 +236,11 @@ extension NIOBSDSocket.OptionLevel {
                 NIOBSDSocket.OptionLevel(rawValue: IPPROTO_TCP)
     #endif
 
+    /// Socket options that apply to MPTCP sockets.
+    ///
+    /// These only work on Linux currently.
+    public static let mptcp = NIOBSDSocket.OptionLevel(rawValue: 284)
+
     /// Socket options that apply to all sockets.
     public static let socket: NIOBSDSocket.OptionLevel =
             NIOBSDSocket.OptionLevel(rawValue: SOL_SOCKET)
@@ -314,6 +319,15 @@ extension NIOBSDSocket.Option {
             NIOBSDSocket.Option(rawValue: TCP_CONNECTION_INFO)
 }
 #endif
+
+// MPTCP options
+//
+// These values are hardcoded as they're fairly new, and not available in all
+// header files yet.
+extension NIOBSDSocket.Option {
+    /// Get info about an MPTCP connection
+    public static let mptcp_info = NIOBSDSocket.Option(rawValue: 1)
+}
 
 // Socket Options
 extension NIOBSDSocket.Option {
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -261,6 +261,10 @@ public final class ServerBootstrap {
     /// failing if the system does not support it. Users are recommended to
     /// only enable this in response to configuration or feature detection.
     ///
+    /// > Note: Enabling this setting will re-enable Nagle's algorithm, even if it
+    /// > had been disabled. This is a temporary workaround for a Linux kernel
+    /// > limitation.
+    ///
     /// - parameters:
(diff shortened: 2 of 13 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -253,6 +255,26 @@ public final class ServerBootstrap {
...
+    /// - parameters:
+    ///     - value: Whether to enable MPTCP or not.
+    public func enableMPTCP(_ value: Bool) -> Self {
+        self.enableMPTCP = value
+
+        // This is a temporary workaround until we get some stable Linux kernel
+        // versions that support TCP_NODELAY and MPTCP.
+        if value {
+            self._serverChannelOptions.remove(key: ChannelOptions.tcpOption(.tcp_nodelay))
+        }
```
dnadoba, line 273: Will we be able to detect the linux kernel version at runtime in the future to make this a non-breaking change?

Lukasa: Yes, in principle we can do this.

dnadoba (approved):
LGTM

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -253,6 +255,26 @@ public final class ServerBootstrap {
...
+    /// failing if the system does not support it. Users are recommended to
+    /// only enable this in response to configuration or feature detection.
+    ///
+    /// - parameters:
+    ///     - value: Whether to enable MPTCP or not.
+    public func enableMPTCP(_ value: Bool) -> Self {
+        self.enableMPTCP = value
+
+        // This is a temporary workaround until we get some stable Linux kernel
+        // versions that support TCP_NODELAY and MPTCP.
```
glbrntt, line 270: Can we note this behaviour in the docs?

glbrntt (approved):

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -236,6 +236,11 @@ extension NIOBSDSocket.OptionLevel {
                 NIOBSDSocket.OptionLevel(rawValue: IPPROTO_TCP)
     #endif
 
+    /// Socket options that apply to MPTCP sockets.
+    ///
+    /// These only work on Linux currently.
+    public static let mptcp = NIOBSDSocket.OptionLevel(rawValue: 284)
```
FranzBusch, line 242: Does it make sense to mark them as unavailable from Darwin OSes?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -253,6 +255,26 @@ public final class ServerBootstrap {
...
 
+    /// Enables multi-path TCP support.
+    ///
+    /// This option is only supported on some systems, and will lead to bind
+    /// failing if the system does not support it. Users are recommended to
+    /// only enable this in response to configuration or feature detection.
+    ///
+    /// - parameters:
+    ///     - value: Whether to enable MPTCP or not.
+    public func enableMPTCP(_ value: Bool) -> Self {
```
FranzBusch, line 266: Since we know these are not available on Darwin OSes does it make sense to mark them as unavailable?

Sources/NIOCore/BSDSocketAPI.swift
Lukasa: I'm inclined to leave them: it adds compiler errors for no good reason. We could set it to a nonsense value, I suppose.

FranzBusch (approved):

PR #2309, opened 2022-11-07
add witnesses for ByteBufferView.reserveCapacity(_:), append(_:), and append(contentsOf:)

kelvin13:
this would improve the efficiency of `ByteBufferView`’s `RangeReplaceableCollection` interface.

see [discussion](https://forums.swift.org/t/shoehorning-bytebuffer-bytebufferview-into-rangereplaceablecollection/61269/2).

### DIFF
--- a/Sources/NIOCore/ByteBuffer-views.swift
+++ b/Sources/NIOCore/ByteBuffer-views.swift
@@ -158,6 +158,27 @@ extension ByteBufferView: RangeReplaceableCollection {
         self = ByteBufferView(ByteBuffer())
     }
 
+    /// Reserves enough space in the underlying `ByteBuffer` to store the specified
+    /// number of bytes.
+    ///
+    /// See the documentation for ``ByteBuffer.reserveCapacity(_:)`` for more details.
+    @inlinable
+    public mutating func reserveCapacity(_ minimumCapacity: Int) {
+        self._buffer.reserveCapacity(minimumCapacity)
+    }
+
+    /// Writes a single byte to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append(_ byte: UInt8) {
+        self._buffer.writeBytes(CollectionOfOne<UInt8>(byte))
+    }
+
+    /// Writes a sequence of bytes to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append<Bytes: Sequence>(contentsOf bytes:Bytes) where Bytes.Element == UInt8 {
+        self._buffer.writeBytes(bytes)
+    }
+
     @inlinable
     public mutating func replaceSubrange<C: Collection>(_ subrange: Range<Index>, with newElements: C) where ByteBufferView.Element == C.Element {
         precondition(subrange.startIndex >= self.startIndex && subrange.endIndex <= self.endIndex,


dnadoba:

dnadoba (approved):
Thanks @kelvin13 for the additions!

dnadoba:

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,27 @@ extension ByteBufferView: RangeReplaceableCollection {
         self = ByteBufferView(ByteBuffer())
     }
 
+    /// Reserves enough space in the underlying `ByteBuffer` to store the specified
+    /// number of bytes.
+    ///
+    /// See the documentation for ``ByteBuffer.reserveCapacity(_:)`` for more details.
+    @inlinable
+    public mutating func reserveCapacity(_ minimumCapacity: Int) {
+        self._buffer.reserveCapacity(minimumCapacity)
```
Lukasa, line 167: We don't necessarily care here, but `ByteBuffer`'s `reserveCapacity` method doesn't have exactly the semantics we want. Specifically it completely disregards the reader/writer indices, so if this `ByteBufferView` is a slice of a larger buffer, we need to be a little careful. In this instance, we probably want to do something like this:

```swift
let currentBufferCapacity = self._buffer.capacity
let currentViewSize = self.count
let requestedExtraCapacity = max(minimumCapacity - currentViewSize, 0)
self._buffer.reserveCapacity(currentBufferCapacity + requestedExtraCapacity)
```

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,27 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+    /// Writes a single byte to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append(_ byte: UInt8) {
+        self._buffer.writeBytes(CollectionOfOne<UInt8>(byte))
+    }
+
+    /// Writes a sequence of bytes to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append<Bytes: Sequence>(contentsOf bytes: Bytes) where Bytes.Element == UInt8 {
+        self._buffer.writeBytes(bytes)
```
Lukasa, line 179: This does not correctly manipulate `self._range`, and so fails to move `endIndex`. Can you add a test for this function? That test should reveal this issue.

Lukasa (changes requested):
Thanks for this! A couple of tweaks and we should be golden.

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: done, and done. i’m not familiar with the preconditions of `moveWriterIndex(to:)`, so i hope i am using it correctly…

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,39 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+            self._buffer.reserveCapacity(self._buffer.capacity + additionalCapacity)
+        }
+    }
+
+    /// Writes a single byte to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append(_ byte: UInt8) {
+        // ``CollectionOfOne`` has no witness for
+        // ``Sequence.withContiguousStorageIfAvailable(_:)``. so we do this instead:
+        Swift.withUnsafeBytes(of: byte)
```
Lukasa, line 178: We don't need unsafe code here, we can just use `writeInteger(byte)`.

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,39 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+        {
+            self._buffer.setBytes($0, at: self.endIndex)
+        }
+        self._range = self._range.startIndex ..< self._range.endIndex.advanced(by: 1)
+        self._buffer.moveWriterIndex(to: self.endIndex)
```
Lukasa, line 183: If we use `writeInteger` as shown above, we can avoid the need to move the writer index.

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,39 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+    /// Writes a single byte to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append(_ byte: UInt8) {
+        // ``CollectionOfOne`` has no witness for
+        // ``Sequence.withContiguousStorageIfAvailable(_:)``. so we do this instead:
+        Swift.withUnsafeBytes(of: byte)
+        {
+            self._buffer.setBytes($0, at: self.endIndex)
+        }
+        self._range = self._range.startIndex ..< self._range.endIndex.advanced(by: 1)
```
Lukasa, line 182: It's probably moderately clearer (and more consistent with this file) to use `range.upperBound` and `range.lowerBound` instead of `startIndex` and `endIndex`.

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,39 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+        self._buffer.moveWriterIndex(to: self.endIndex)
+    }
+
+    /// Writes a sequence of bytes to the underlying `ByteBuffer`.
+    @inlinable
+    public mutating func append<Bytes: Sequence>(contentsOf bytes: Bytes) where Bytes.Element == UInt8 {
+        let written = self._buffer.setBytes(bytes, at: self.endIndex)
```
Lukasa, line 189: If we can use `writeBytes` we don't need to modify the writer index.

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -158,6 +158,39 @@ extension ByteBufferView: RangeReplaceableCollection {
...
+        self._range = self._range.startIndex ..< self._range.endIndex.advanced(by: written)
```
Lukasa, line 190: Same note about upper and lower bound.

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: what happens if `_range.endIndex` is less than `_buffer.writerIndex`?

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: replaced with `setInteger` per discussion below

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: see discussion below

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: done

Sources/NIOCore/ByteBuffer-views.swift
kelvin13: done

Lukasa (approved):
Great, thanks so much!

Lukasa:

Lukasa:

Issue #2310, opened 2022-11-10
Special case `EventLoopPromise.succeed()` when `Value` is `Void`


glbrntt:
To succeed an `EventLoopPromise<Void>` users must pass a void-value to it, e.g.:
- `promise.succeed(())` 
- `promise.succeed(Void())`

It would be nice if we could spell this as `promise.succeed()`.

mob-connection:
Maybe something like this ?
```
extension EventLoopPromise where Value == Void {
    @inlinable
    public func succeed() {
        succeed(Void())
    }

}
```

glbrntt:
Yes, something like that. Would you like to open a PR for this @mob-connection?

PR #2311, opened 2022-11-11
Special case EventLoopPromise.succeed() when Value is Void

mob-connection:
Motivation:

To succeed an `EventLoopPromise<Void>` users must pass a void-value to it, e.g.:
`promise.succeed(())`
`promise.succeed(Void())`
It would be nice if we could spell this as `promise.succeed()`.

For [issue](https://github.com/apple/swift-nio/issues/2310)

Modifications:

- add `func succeed()` for `Void` type via `extension`

Result: 

Users can use `succeed` without `Void` type declaration


FranzBusch (approved):

FranzBusch:
LGTM! Thanks for the contribution

dnadoba (approved):

dnadoba:

dnadoba:
Thanks @mob-connection for this patch and sorry for the delay!

PR #2313, opened 2022-11-14
Fix failing build on MacOS

FranzBusch:
# Motivation
Currently, the main branch fails to build for me since the `testSimpleMPTCP()` can't compile. This is due to the fact that the `getMPTCPInfo()` method is only available on Linux.

# Modification
Add `#if os(Linux)` around the `testSimpleMPTCP()`.

# Result
Main builds on MacOS again.


Lukasa (approved):

PeterAdams-A (approved):
Seems reasonable.

PR #2317, opened 2022-11-16
Add `NIOBSDSocket.ProtocolSubtype`

dnadoba:
### Motivation:

We want to support custom IP protocols through POSIX Raw Socket API. To allow this, users will need to specify the IP sub protocol type they want to use.

### Modifications:

- Add `NIOBSDSocket.ProtocolSubtype` with a lot of well known protocols
- Plumb `NIOBSDSocket.ProtocolSubtype` through `DatagramChannel` which we will reuse for the Raw Socket API

### Result:

`DatagramChannel ` is prepared to be used in a `RawSocketBootstrap`


Sources/NIOPosix/BSDSocketAPICommon.swift
```diff
@@ -127,6 +127,333 @@ extension NIOBSDSocket.Option {
         NIOBSDSocket.Option(rawValue: Posix.IPV6_RECVPKTINFO)
 }
 
+extension NIOBSDSocket {
+    /// In the Internet Protocol version 4 (IPv4) [RFC791] there is a field
+    /// called "Protocol" to identify the next level protocol.  This is an 8
+    /// bit field.  In Internet Protocol version 6 (IPv6) [RFC8200], this field
+    /// is called the "Next Header" field.
+    public struct ProtocolSubtype: RawRepresentable {
```
Lukasa, line 135: Can you add `Hashable`?

Sources/NIOPosix/BSDSocketAPICommon.swift
```diff
@@ -127,6 +127,333 @@ extension NIOBSDSocket.Option {
...
+//
+//
+//     if let description {
+//         print("/// \(description)")
+//     }
+//     print("public static let \(identifier) = Self(rawValue: \(number))")
+// }
+extension NIOBSDSocket.ProtocolSubtype {
+    /// internet protocol, pseudo protocol number
+    public static let ip = Self(rawValue: 0)
```
Lukasa, line 178: These protocol subtypes are not necessarily stable from OS release to OS release. As much as possible, can we use the symbolic names?

Lukasa (changes requested):

Sources/NIOPosix/BSDSocketAPICommon.swift
dnadoba: Where can I find the symbolic names?

Lukasa: Some of these have that effect, but some do not. The protocol subtype field has uses outside of raw sockets.

In this case, these all match `IPPROTO_*` constants in the various libcs. I suspect `netinet/in.h` has most of them.

dnadoba: I have made `NIOBSDSocket.ProtocolSubtype` internal and moved all known protocol types to a new type called `NIOIPProtocol` which is public. Should I still use the `IPPROTO_*` constants for those as well?

Lukasa: I think if we can point to the IANA registry that's sufficient.

dnadoba: Not sure I can follow. What do you mean by pointing to the IANA registry?

Lukasa: IANA (the Internet Assigned Numbers Authority) holds registries for most extension points in IETF protocols. In the case of the IPv4 next-level protocol field or IPv6 next header field, that registry is [this one](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml). You can skip most of that registry: I'd limit myself to assignments with a reference to an RFC or IANA. But that's the authoritative reference for the numbers, so we should reference it as the source for the numbers we choose.

Sources/NIOCore/IPProtocol.swift
```diff
@@ -0,0 +1,338 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// In the Internet Protocol version 4 (IPv4) [RFC791] there is a field
+/// called "Protocol" to identify the next level protocol.  This is an 8
+/// bit field.  In Internet Protocol version 6 (IPv6) [RFC8200], this field
+/// is called the "Next Header" field.
+public struct NIOIPProtocol: RawRepresentable, Hashable {
```
Lukasa, line 19: For the sake of convenience it might be nice to add a `CustomStringConvertible` implementation that switches over the named `static let`s and prints those names as well as the raw number.

Sources/NIOCore/IPProtocol.swift
```diff
@@ -0,0 +1,338 @@
...
+    public typealias RawValue = UInt8
+    public var rawValue: RawValue
+    
+    @inlinable
+    public init(rawValue: RawValue) {
```
Lukasa, line 24: Can we add an `init(_ rawValue: Int)` as well? It's useful to be able to construct this from a general `Int`.

Sources/NIOPosix/BSDSocketAPICommon.swift
dnadoba: Done in https://github.com/apple/swift-nio/pull/2317/commits/30e96bba9f72346fd43e94bdb9f5b4e68398eb05

Sources/NIOPosix/BSDSocketAPICommon.swift
```diff
@@ -127,6 +127,28 @@ extension NIOBSDSocket.Option {
...
+        var rawValue: RawValue
+        
+        init(rawValue: RawValue) {
+            self.rawValue = rawValue
+        }
+    }
+}
+
+extension NIOBSDSocket.ProtocolSubtype {
+    static let ip = Self(rawValue: 0)
```
Lukasa, line 142: Should we add an `MPTCP` `static let` here to avoid using the raw value elsewhere?

Sources/NIOPosix/BSDSocketAPIPosix.swift
```diff
@@ -212,10 +212,10 @@ extension NIOBSDSocket {
 
     // The protocol subtype for MPTCP.
     // Returns nil if mptcp is not supported.
-    static var mptcpProtocolSubtype: Int? {
+    static var mptcpProtocolSubtype: NIOBSDSocket.ProtocolSubtype? {
```
Lukasa, line 215: Should we remove this in favour of `NIOBSDSocket.ProtocolSubtype.mptcp`?

Sources/NIOCore/IPProtocol.swift
```diff
@@ -0,0 +1,119 @@
...
+    /// Robust Header Compression - [RFC5858]
+    public static let rohc = Self(rawValue: 142)
+    /// Ethernet - [RFC8986]
+    public static let ethernet = Self(rawValue: 143)
+    /// AGGFRAG encapsulation payload for ESP - [RFC-ietf-ipsecme-iptfs-19]
+    public static let aggfrag = Self(rawValue: 144)
+    // Use for experimentation and testing (253) - [RFC3692]
+    public static let useForExperimentationAndTesting1 = Self(rawValue: 253)
+    // Use for experimentation and testing (254) - [RFC3692]
+    public static let useForExperimentationAndTesting2 = Self(rawValue: 254)
```
Lukasa, line 118: I'd skip these bottom two names: we can just leave them undefined.

Sources/NIOPosix/BSDSocketAPICommon.swift
Lukasa, line 142: I don't think `0` implies `IP`: I think it just implies "default".

Sources/NIOCore/IPProtocol.swift
Lukasa: I think so. It remains helpful for cases where you can easily get hold of an `Int`, and represents good practice in the Swift ecosystem (where `Int` is the currency numerical type). The crash at runtime is a red herring: if the user has an `Int`, we just force them to write the crash in _their_ code instead of ours. The type very clearly communicates that it can only hold a `UInt8`, so there's no ambiguity, and we can safely save the user the hassle of writing that code.

dnadoba: Done in 35cb11f9b16375ccbedccd4907afc8f2af668b6b

Sources/NIOCore/IPProtocol.swift
dnadoba: Done in 904c5fd6380eb9e5667378a39176cab829e2361f

Sources/NIOPosix/BSDSocketAPICommon.swift
dnadoba: Done in dddf452e0e9843312528554df56be8d8cbe0fbd2

Sources/NIOPosix/BSDSocketAPIPosix.swift
dnadoba: Done in fadd5c120b7626be22b7731daab94c171a1fe610

Sources/NIOPosix/BSDSocketAPICommon.swift
dnadoba: Done in fadd5c120b7626be22b7731daab94c171a1fe610

Lukasa:

Lukasa:
Looks like the syscall wrapper test has broken. Probably this is the result of another dependency getting added to actually building that project (listed in `IntegrationTests/tests_02_syscall_wrappers/defines.sh`). You probably need to add `IPProtocol.swift` to that list.

PR #2318, opened 2022-11-17
Make `NIOHTTP1TestServer` Sendable

dnadoba:
`NIOHTTP1TestServer` is actually `Sendable` through an `EventLoop` which is also used as an `EventLoopGroup`.


fabianfett (approved):
We should wait for @Lukasa though.

Lukasa (approved):
Yes, good catch: this type is _supposed_ to be `Sendable`! Hard to use it in tests otherwise!

PR #2320, opened 2022-11-22
Add `RawSocketBootstrap`

dnadoba:
### Motivation:

We want to support custom IP protocols through POSIX Raw Socket API. To allow this, users will need to bootstrap a `DatagramChannel` with a socket in raw mode.

### Modifications:

- add `RawSocketBoostrap` which configures a `DatagramChannel` with a raw socket.
- add end-to-end test 

### Result:

NIO can send and receive IP packets.

### Alternatives
`RawSocketBootstrap` is inspired by the socket mode name (`SOCK_RAW`). Instead we could also call it `IPBootstrap` to reflect that the resulting channel will operate on IP packets.


Tests/NIOPosixTests/RawSocketBootstrapTests.swift
```diff
@@ -0,0 +1,60 @@
...
+        
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try elg.syncShutdownGracefully()) }
+        let channel = try RawSocketBootstrap(group: elg)
+            .channelInitializer {
+                $0.pipeline.addHandler(DatagramReadRecorder<ByteBuffer>(), name: "ByteReadRecorder")
+            }
+            .bind(host: "127.0.0.1", ipProtocol: .reservedForTesting).wait()
+        defer { XCTAssertNoThrow(try channel.close().wait()) }
+        try channel.configureForRecvMmsg(messageCount: 10)
```
Lukasa, line 41: We should probably have a test that doesn't use `recvmmsg` as well to confirm we handle that path appropriately.

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+final class RawSocketBootstrapTests: XCTestCase {
+    func testWriteAndRead() throws {
+        #if canImport(Darwin)
+        try XCTSkipIf(geteuid() != 0, "Raw Socket API requires root privileges on Darwin")
+        #endif
+        
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try elg.syncShutdownGracefully()) }
+        let channel = try RawSocketBootstrap(group: elg)
```
Lukasa, line 35: We probably need a test of the `ip_hdrincl` path.

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
```diff
@@ -0,0 +1,60 @@
...
+            _ = try channel.write(AddressedEnvelope(
+                remoteAddress: SocketAddress(ipAddress: "127.0.0.1", port: 0),
+                data: ByteBuffer(string: message)
+            ))
+        }
+        channel.flush()
+        
+        let receivedMessages = Set(try channel.waitForDatagrams(count: 10).map(\.data).map { buffer in
+            String(
+                decoding: buffer.readableBytesView.dropFirst(4 * 5), // skip the IPv4 header
```
Lukasa, line 53: Ideally we'd actually validate that this is an IPv4 header.

Lukasa (changes requested):
Thanks, this looks great! I think we need a few extra tests though, as we're currently not coming close to testing all the new code that was added here.

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
dnadoba: Tested now in `RawSocketBootstrapTests.testConnect`

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
```diff
@@ -0,0 +1,60 @@
...
+        try XCTSkipIf(geteuid() != 0, "Raw Socket API requires root privileges on Darwin")
+        #endif
+        
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { XCTAssertNoThrow(try elg.syncShutdownGracefully()) }
+        let channel = try RawSocketBootstrap(group: elg)
+            .channelInitializer {
+                $0.pipeline.addHandler(DatagramReadRecorder<ByteBuffer>(), name: "ByteReadRecorder")
+            }
+            .bind(host: "127.0.0.1", ipProtocol: .reservedForTesting).wait()
```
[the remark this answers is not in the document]
dnadoba: Tested now in `RawSocketBootstrapTests.testConnect`

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
dnadoba: Tested now in `RawSocketBootstrapTests.testIpHdrincl`

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
dnadoba: Tested in now in all tests of `RawSocketBootstrapTests`

Sources/NIOPosix/RawSocketBootstrap.swift
```diff
@@ -0,0 +1,197 @@
...
+///             channel.pipeline.addHandler(MyChannelHandler())
+///         }
+///     let channel = try! bootstrap.bind(host: "127.0.0.1", ipProtocol: .icmp).wait()
+///     /* the Channel is now ready to send/receive IP packets */
+///
+///     try channel.closeFuture.wait()  // Wait until the channel un-binds.
+/// ```
+///
+/// The `Channel` will operate on `AddressedEnvelope<ByteBuffer>` as inbound and outbound messages.
+public final class RawSocketBootstrap {
```
weissi, line 36:
```suggestion
public final class NIORawSocketBootstrap {
```

as per the public API guidelines, this need to have a `NIO` prefix

Tests/NIOPosixTests/IPv4Header.swift
```diff
@@ -0,0 +1,277 @@
...
+            let identification: UInt16 = buffer.readInteger(),
+            // fragmentOffset is in host byte order as well but it is always zero in our tests
+            // and fragmentOffset is 13 bits in size so we can't just use readInteger(endianness: .host)
+            let flagsAndFragmentOffset: UInt16 = buffer.readInteger(),
+            let timeToLive: UInt8 = buffer.readInteger(),
+            let `protocol`: UInt8 = buffer.readInteger(),
+            let headerChecksum: UInt16 = buffer.readInteger(),
+            let sourceIpAddress: UInt32 = buffer.readInteger(),
+            let destinationIpAddress: UInt32 = buffer.readInteger()
+        else { return nil }
```
weissi, line 127: this is incorrect because it forgets to reset the `buffer` if you half-parsed something.

Also this doesn't follow the usual pattern of `read*`/`write*` methods. And lastly, there's an implementation of this already:
- check `PCAP.swift` on this PR: https://github.com/apple/swift-nio-extras/pull/158 : https://github.com/apple/swift-nio-extras/pull/158/files#diff-8b0f17fa27d694721788c7adab20151f44ff3672becc16dd728bd530c0de44ba
- also here (this code is changed by the above PR): https://github.com/apple/swift-nio-extras/blob/main/Tests/NIOExtrasTests/WritePCAPHandlerTest.swift#L746

Tests/NIOPosixTests/IPv4Header.swift
```diff
@@ -0,0 +1,277 @@
...
+        assert({
+            var buffer = ByteBuffer()
+            self._write(to: &buffer)
+            let newValue = Self(buffer: &buffer)
+            return self == newValue
+        }())
+        self._write(to: &buffer)
+    }
+    
+    private func _write(to buffer: inout ByteBuffer) {
```
weissi, line 219: normally we write these as

```
extension ByteBuffer {
   private func writeTCP4Packet...
```

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
```diff
@@ -0,0 +1,192 @@
...
+            let header = try XCTUnwrap(IPv4Header(buffer: &data))
+            XCTAssertEqual(header.version, 4)
+            XCTAssertEqual(header.protocol, .reservedForTesting)
+            #if canImport(Darwin)
+            // On BSD the IP header will only contain the size of the ip packet body not the header.
+            // This is known bug which can't be fixed without breaking old apps which already workaround the issue
+            // like we do.
+            XCTAssertEqual(Int(header.totalLength), data.readableBytes)
+            // On BSD the checksum is always zero
+            XCTAssertEqual(header.headerChecksum, 0)
```
weissi, line 70: I don't think we should make this platform specific. I think we should have a `bsdHeaderChecksum` and `bsdTotalLength` etc properties. And maybe maybe a `headerChecksumForPlatformRawSockets` or so properties that then (platform specifically) return the correct one.

dnadoba: I thought about this too but couldn't come up with good names. These properties should return the correct value (as specified in [RFC-791](https://www.rfc-editor.org/rfc/rfc791#page-11)). The problem though is not only naming but also that the workaround (i.e. adding 20 to `totalLength`) requires knowing if we have just parsed the IPv4 packet from a `ByteBuffer` we got from a raw socket or not. If we create a `IPv4Header` ourselves e.g. for sending through the Raw Socket API, the totalLength will be correct and adding 20 will result in a wrong value.

The options I see are:
- store this information in the type and dynamically return the correct value
    - This works awful if we reuse the same value to e.g. echo a packet
- use different types for receiving and sending.
    - stores this information statically and we will need to explicitly convert the type when we want to echo a packet where we can get the `totalLength` fixed up
- use some rather long name which conveys that this is only for the receiving side e.g. `platformIndependetTotalLengthForRecivedPacket`
- add some documentation and leave it to the call site to handle it


I kind of want to use the "rather long name" variant but don't yet have name I like much.

Lukasa: I also don't think we need to worry _too_ much about this, as this is testing code. It's ok if the parsing is a bit janky.

Tests/NIOPosixTests/IPv4Header.swift
dnadoba: Done in https://github.com/apple/swift-nio/pull/2320/commits/20ed3802642a475b760beb33d8a121430f0224a6

Tests/NIOPosixTests/IPv4Header.swift
dnadoba: Done in https://github.com/apple/swift-nio/pull/2320/commits/20ed3802642a475b760beb33d8a121430f0224a6

Sources/NIOPosix/RawSocketBootstrap.swift
dnadoba: Done in https://github.com/apple/swift-nio/pull/2320/commits/981d0b88b242aa1aa5ed14385699dde39c95cfa1

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
dnadoba: I want to eventually make `IPv4Header` public API so we can still discuss this here. However, I will make this in a separate PR so it is not blocking `RawSocketBootstrap` to land.

Lukasa (approved):
I'm happy enough with this change, but it'd be good to confirm that @weissi is happy too.

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
weissi: I'd go for some long names to explain this. And +1 for the public API, this should be reconciled with the already-existing IPv4/6 parsing code in swift-nio-extras.

Tests/NIOPosixTests/IPv4Header.swift
```diff
@@ -0,0 +1,316 @@
...
+    }
+}
+
+extension ByteBuffer {
+    @discardableResult
+    mutating func readIPv4Header() -> IPv4Header? {
+        #if canImport(Darwin)
+        var initialState = self
+        guard
+            let versionAndIhl: UInt8 = self.readInteger(),
```
weissi, line 187: just a question: Why don't you actually use `self.readMultipleIntegers` here?

dnadoba: https://www.unix.com/man-page/mojave/4/ip/
> Note that the ip_off and ip_len fields are in host byte order.

relevant comment is at line 189. `self.readMultipleIntegers` sadly doesn't support specifying endianness per integer but I also think it shouldn't as this is a bug in the OS.

Tests/NIOPosixTests/RawSocketBootstrapTests.swift
dnadoba: Added them in https://github.com/apple/swift-nio/pull/2320/commits/fabaf9b804afd085ff5d2c2f84ca8d2764e7b762

Tests/NIOPosixTests/IPv4Header.swift
```diff
@@ -0,0 +1,347 @@
...
+        self.sourceIpAddress = .init(rawValue: 0)
+        self.destinationIpAddress = .init(rawValue: 0)
+    }
+}
+
+
+
+extension ByteBuffer {
+    @discardableResult
+    mutating func readIPv4Header() -> IPv4Header? {
```
weissi, line 153: As discussed offline this should IMHO really not be platform specific.
I think we should have readIPv4Header which reads it correctly. And then we can have another one which is readIPv4HeaderWithDarwinRawSocketQuirks() or something which does the weird thing.
Finally we can then have a readIPv4HeaderFromOSRawSocket()  or something which chooses (in a platform specific way) the right one for the platform

But any readFoo that can't read something written by a writeFoo even if on a different platform is a major red flag

weissi: There's no reason we shouldn't unit test the darwin quirks on Linux too.

dnadoba: Added in https://github.com/apple/swift-nio/pull/2320/commits/1ac14d225b8a8b50f486a1117029d0d313066545

Tests/NIOPosixTests/IPv4Header.swift
```diff
@@ -0,0 +1,365 @@
...
+
+extension IPv4Header {
+    var platformIndependentTotalLengthForReceivedPacketFromRawSocket: UInt16 {
+        #if canImport(Darwin)
+        // On BSD the IP header will only contain the size of the ip packet body, not the header.
+        // This is known bug which can't be fixed without breaking old apps which already workaround the issue
+        // like e.g. we do now too.
+        return totalLength + 20
+        #elseif os(Linux)
+        return totalLength
```
buttaface, line 333: This pull [broke building the tests for Android](https://github.com/buttaface/swift-android-sdk/actions/runs/3600314592/jobs/6064844908#step:13:377). I [just confirmed that adding Android here fixes it again](https://github.com/buttaface/swift-android-sdk/pull/83/checks), buttaface/swift-android-sdk#83.

@dnadoba, mind adding these Android checks in a subsequent commit?

dnadoba: Oh yeah sure, sorry about that. I think we can also just remove os(Linux) and just use an `#else`

PR #2323, opened 2022-11-25
cap read+pread POSIX read sizes at Int32.max

rnro:
### Motivation:

According to `read`/`pread` man page, reads may return `EINVAL` if "The value provided for `nbyte` exceeds `INT_MAX`". In practice on Darwin it is observed that this happens for `>Int32.max`.

We workaround the issiue in the `NonBlockingFileIO` level to keep the lower levels as simple as possible.

### Modifications:

`NonBlockingFileIO` `read0` amends read `byteCount`s to be `Int32.max` if they are larger than that value.

### Result:

We will no longer receive `EINVAL` signals and fail preconditions for reads larger than` Int32.max`. reads larger than `Int32.max` will return up to `Int32.max` bytes.

### DIFF
--- a/Sources/NIOPosix/System.swift
+++ b/Sources/NIOPosix/System.swift
@@ -602,17 +602,15 @@ internal enum Posix {
 
     @inline(never)
     internal static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<ssize_t> {
-        let cappedSize = size < Int32.max ? size : size_t(Int32.max)
         return try syscallForbiddingEINVAL {
-            sysRead(descriptor, pointer, cappedSize)
+            sysRead(descriptor, pointer, size)
         }
     }
 
     @inline(never)
     internal static func pread(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t, offset: off_t) throws -> IOResult<ssize_t> {
-        let cappedSize = size < Int32.max ? size : size_t(Int32.max)
         return try syscallForbiddingEINVAL {
-            sysPread(descriptor, pointer, cappedSize, offset)
+            sysPread(descriptor, pointer, size, offset)
         }
     }
 
--- a/Sources/NIOPosix/System.swift
+++ b/Sources/NIOPosix/System.swift
@@ -602,15 +602,17 @@ internal enum Posix {
 
     @inline(never)
     internal static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<ssize_t> {
+        let cappedSize = size < Int32.max ? size : size_t(Int32.max)
         return try syscallForbiddingEINVAL {
-            sysRead(descriptor, pointer, size)
+            sysRead(descriptor, pointer, cappedSize)
         }
     }
 
     @inline(never)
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/System.swift
```diff
@@ -602,15 +602,17 @@ internal enum Posix {
 
     @inline(never)
     internal static func read(descriptor: CInt, pointer: UnsafeMutableRawPointer, size: size_t) throws -> IOResult<ssize_t> {
+        let cappedSize = size < Int32.max ? size : size_t(Int32.max)
```
weissi, line 605: I don't think we should be capping this so low-level. This could lead to weird problems. IMHO we should make sure not to issue too-big `read` calls?

Lukasa: Agreed, we should guard this in NonBlockingFileIO.

Lukasa:

Lukasa (changes requested):
As noted in the PR feedback, we should only cap this at the usage site, not at the syscall layer.

Sources/NIOPosix/System.swift
rnro: Thanks for the input folks! I'm happy to make the change but could you possibly expand on the reasoning to inform future commits? It seemed to me that since this change was defensive against a low-level behaviour it made sense to 'protect' at as low a level as possible to be less intrusive and less likely that someone accidentally sidesteps the protection. 

For example Is it more important that we allow people to workaround this if desired or that there is higher visibility for this behaviour change?

Lukasa: The syscall layer is not API: it's an implementation detail of NIO. That means the only people who call it are us. We tend to set ourselves the rule that we should solve problems at the location where we have the most information about what solution is optimal, and usually that's at as high a level as possible. It is not necessarily the case that it's safe to truncate arbitrary reads, but for NonBlockingFileIO we _know_ it's safe to do so.

Additionally, this helps with debugging. The syscall layer in NIO is essentially a direct translation of C syscalls into Swift. It does minimal data transformations. That means you don't have to look into it to see if it screwed with your data: you can assume that whatever you passed to it made it to the OS unchanged. That lets you avoid needing to have debugging sessions spend too much time in that layer of the code. This is a good principle to persist.

rnro: Thanks Cory!

Tests/NIOPosixTests/NonBlockingFileIOTest.swift
```diff
@@ -282,6 +282,14 @@ class NonBlockingFileIOTest: XCTestCase {
         XCTAssertEqual(2, numCalls)
     }
 
+    func testReadMoreThanIntMaxBytesDoesntThrow() throws {
+        // here we try to read way more data back from the file than it contains but it serves the purpose
+        // even on a small file the OS will return EINVAL if you try to read > INT_MAX bytes
+        try withTemporaryFile(content: "some-dummy-content", { (filehandle, path) -> Void in
```
Lukasa, line 288: Can we expand this test to confirm the content is appropriate? (In this case, empty.)

rnro: 👍I have added a content check here

Lukasa (approved):
Nice! This LGTM.

PR #2326, opened 2022-12-02
Remove dependency on NIO umbrella target

dnadoba:
`NIOHTTP1` and `NIOWebSocket` don’t `import NIO` and therefore should not depend on it.


Lukasa (changes requested):
Unfortunately, we can't merge this.

These dependencies are load-bearing: they have the effect of supporting invalid dependency trees that expect NIO to be present and loadable. Removing them is not currently possible.

dnadoba:
hm... yeah you are right. Forgot about that issue

glbrntt:
@dnadoba probably worth filing an issue to fix this in the next major version

dnadoba:
https://github.com/apple/swift-nio/issues/2327

PR #2330, opened 2022-12-12
Remove implicit ByteBuffer copy in copyMemory

Lukasa:
Motivation:

withVeryUnsafeBytes has a reference to `self` that it is using. This necessarily triggers a copy when we subsequently use `self` inside the block. The effect of that copy is an extra retain/release pair, and a call to beginAccess, that we simply don't need.

Modifications:

Replace the call to `_setBytesAssumingUniqueBufferAccess` with a manually inlined equivalent.

Result:

Avoids the extra copy of `self`, removes a retain/release and a beginAccess.


FranzBusch (approved):
Nice one!

dnadoba (approved):

PR #2331, opened 2022-12-12
Make EventLoopFuture.wait() unavailable from async

Lukasa:
Motivation:

ELF.wait() waits for a condition variable to become true, which can frequently lead to extremely long waits. This is a bad thing to call on a Swift Concurrency thread, especially as we have ELF.get() which is preferable.

Modifications:

Make ELF.wait() unavailable from async.

Result:

Users are encouraged to use the correct primitive.

### DIFF
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -1237,7 +1237,7 @@ extension EventLoopFuture {
     ///
     /// - returns: The value of the `EventLoopFuture` when it completes.
     /// - throws: The error value of the `EventLoopFuture` if it errors.
-    @available(*, noasync, message: "threads can change between suspension points and therefore the thread specific value too")
+    @available(*, noasync, message: "wait() can block indefinitely, prefer get()", renamed: "get()")
     @inlinable
     public func wait(file: StaticString = #file, line: UInt = #line) throws -> Value {
         return try self._wait(file: file, line: line)
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -1223,6 +1223,7 @@ extension EventLoopFuture {
 // MARK: wait
 
 extension EventLoopFuture {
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1232,10 +1233,37 @@ extension EventLoopFuture {
     /// threads: it is primarily useful for testing, or for building interfaces between blocking
     /// and non-blocking code.
     ///
+    /// This is also forbidden in async contexts: prefer ``EventLoopFuture/get``.
+    ///
     /// - returns: The value of the `EventLoopFuture` when it completes.
     /// - throws: The error value of the `EventLoopFuture` if it errors.
+    @available(*, noasync, message: "threads can change between suspension points and therefore the thread specific value too")
```
dnadoba, line 1240:
```suggestion
    @available(*, noasync, message: "threads can change between suspension points and therefore the thread specific value too", renamed: "get()")
```
This is will produce a fixit that will allow user to easily replace it with `.get()`
```
warning: instance method 'wait' is unavailable from asynchronous contexts; threads can change between suspension points and therefore the thread specific value too; this is an error in Swift 6
    try future.wait()
               ^~~~
               get
```
Users still need to add `await` infront but the compiler will nicely provide a fixit for that too.
```
error: expression is 'async' but is not marked with 'await'
    try future.get()
    ^~~~~~~~~~~~~~~~
        await 
```

---

 I understand that we want to encourage users to use `get()` but I'm not sure I quite understand the `message`:

> threads can change between suspension points and therefore the thread specific value too
 
`wait()` doesn't use any thread specific value and AFAIK it is "safe" to use in async context (except deadlocks) but is just not performant.

Lukasa: Yeah that is just an overhang from a previous usage I didn't replace.

dnadoba (approved):
LGTM!

ktoso:
LGTM :)

PR #2333, opened 2022-12-13
Measure allocations applying WS mask

Lukasa:
Motivation:

We're missing some alloc tests on the WS mask.

Modifications:

- Add extra alloc tests for the WS mask

Result:

We'll have more tests


Lukasa:

dnadoba (approved):

PR #2334, opened 2022-12-16
add easier async to future conversion

weissi:
### Motivation:

During the migration we unfortunately do need async -> future conversions from time to time. Currently, we do this with `EventLoopPromise.completeWithTask` which is great but also inconvenient in many cases.

### Modifications:

Add `EventLoop.makeFutureWithTask` which allows a direct `EventLoopFuture` -> `async throws` conversion.

### Result:

Less code, better code, fewer bugs.

### DIFF
--- a/Sources/NIOCore/AsyncAwaitSupport.swift
+++ b/Sources/NIOCore/AsyncAwaitSupport.swift
@@ -330,3 +330,13 @@ struct AsyncSequenceFromIterator<AsyncIterator: AsyncIteratorProtocol>: AsyncSeq
         self.iterator
     }
 }
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension EventLoop {
+    @inlinable
+    public func makeFutureWithTask<Return>(_ body: @Sendable @escaping () async throws -> Return) -> EventLoopFuture<Return> {
+        let promise = self.makePromise(of: Return.self)
+        promise.completeWithTask(body)
+        return promise.futureResult
+    }
+}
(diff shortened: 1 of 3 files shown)


Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -330,3 +330,13 @@ struct AsyncSequenceFromIterator<AsyncIterator: AsyncIteratorProtocol>: AsyncSeq
         self.iterator
     }
 }
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension EventLoop {
+    @inlinable
+    public func makeFutureWithTask<Return>(_ body: @Sendable @escaping () async throws -> Return) -> EventLoopFuture<Return> {
```
jakepetroules, line 337: Can we call it `performWithTask`?

Lukasa: I don’t think we should add this extension on ELG: it’s rare that you’re without an opinion as to what loop your promise should be attached to. For that same reason, we don’t offer a `makePromise` API on ELG.

jakepetroules: OK, sounds reasonable. Any thoughts on the name?

jakepetroules (approved):

Sources/NIOCore/AsyncAwaitSupport.swift
Lukasa: I probably have a slight preference for the name proposed by @weissi, as it aligns with existing prior art for making Futures.

Lukasa (approved):

PR #2336, opened 2022-12-19
Improve performance of tests

dnadoba:
### Motivation:

`NIOConcurrencyHelpersTests.testLargeContendedAtomicSum` takes ~42 seconds on linux.
It turns out that a concurrent queue on Linux will only initially spawn up to `System.coreCount` threads. Afterwards it will create one thread per second.

### Modifications:
Limit the number of dispatch items added to the queue to `System.coreCount`. 
This limits the usefulness of this test. However, this test only tests deprecated functionality.

### Result:

`NIOConcurrencyHelpersTests.testLargeContendedAtomicSum` only takes ~0.02 seconds instead of ~42 seconds.
Overall this reduces the time it takes to execute all tests on Linux from ~86 seconds to ~44 seconds.


glbrntt (approved):
Nice investigation and catch!

PR #2337, opened 2022-12-19
Fix main nightly CI

dnadoba:
- Allow `SIGILL` and `SIGTRAP`. `SIGTRAP` is the signal on arm64 for fatalError/precondition/assert
- allocations `1_reqs_1000_conn` have regressed but only to the same numbers as Swift 5.7. not yet investigated

### DIFF
--- a/IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
+++ b/IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
@@ -69,8 +69,9 @@ for mode in debug release; do
             fail "exited successfully but was supposed to fail"
         else
             exit_code=$?
-            # expecting illegal instruction as it should fail with an unacceptable errno
-            assert_equal $(( 128 + 4 )) $exit_code  # 4 == SIGILL
+            # expecting irrecoverable error as process should be terminated through fatalError/precondition/assert
+            assert_greater_than_or_equal $exit_code $(( 128 + 4 ))  # 4 == SIGILL aka illegal instruction, expected on x86
+            assert_less_than_or_equal $exit_code $(( 128 + 5 ))  # 5 == SIGTRAP aka trace trap, expected on arm64
             if [[ "$mode" == "debug" ]]; then
                 grep -q unacceptable\ errno "$temp_file"
             fi
--- a/IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
+++ b/IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
@@ -69,9 +69,17 @@ for mode in debug release; do
             fail "exited successfully but was supposed to fail"
         else
             exit_code=$?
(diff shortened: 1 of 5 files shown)


IntegrationTests/tests_02_syscall_wrappers/test_02_unacceptable_errnos.sh
```diff
@@ -69,8 +69,9 @@ for mode in debug release; do
             fail "exited successfully but was supposed to fail"
         else
             exit_code=$?
-            # expecting illegal instruction as it should fail with an unacceptable errno
-            assert_equal $(( 128 + 4 )) $exit_code  # 4 == SIGILL
+            # expecting irrecoverable error as process should be terminated through fatalError/precondition/assert
+            assert_greater_than_or_equal $exit_code $(( 128 + 4 ))  # 4 == SIGILL aka illegal instruction, expected on x86
+            assert_less_than_or_equal $exit_code $(( 128 + 5 ))  # 5 == SIGTRAP aka trace trap, expected on arm64
```
FranzBusch, line 74: Should we conditionlize this on the platform rather? Otherwise the wrong exit code might be accepted on the wrong platform

dnadoba: I honestly don't think this really matters. Both exit codes should be fine on either platform.

dnadoba: Could find some documentation about it here: https://developer.apple.com/documentation/xcode/understanding-the-exception-types-in-a-crash-report#EXCBREAKPOINT-SIGTRAP-and-EXCBADINSTRUCTION-SIGILL

dnadoba: So it kind of would make sense to conditionlize on the platform but it seems to be rather hard/fragile to actually get the platform we run on in a common format. 
E.g. this is what I get on ubuntu:
```
root@79504e2c266b:/src# uname -m
aarch64
root@79504e2c266b:/src# uname -p
aarch64
```
And this is what I get on macOS:
```
davidnadoba@Davids-MacBook-Pro release-validator % uname -m
arm64
davidnadoba@Davids-MacBook-Pro release-validator % uname -p
arm
```
Open for better ideas though.
Note that we will also need to do this from within Swift.

Lukasa: I think we can just support `uname -m` and treat aarch64 and arm64 as the same.

dnadoba: Done in https://github.com/apple/swift-nio/pull/2337/commits/35e1f6da282d721d779d9aa6cb606e51ca1f1066

Lukasa (approved):

Lukasa:
Thanks so much for this! :sparkles:

PR #2344, opened 2023-01-11
    Prepare mmsghdr structure properly.

ser-0xff:
Motivation:

    According to the Linux man page the msg_len field supposed to be used to return a number of bytes sent for the particular message.
    It does not make a sense to initialize it with a size of the message.

    Modifications:

    Change msg_leg field initialization, use 0 instead of message size.

    Result:

    Use sendmmsg() call properly.

### Motivation:

### Modifications:

### Result:


Lukasa (approved):
Thanks for this!

Lukasa:

Lukasa (changes requested):
Several of the tests (at least in `PendingWritesManagerTests`) are expecting this value to be non-zero. Can you update those tests?

PR #2345, opened 2023-01-11
Implement `remoteAddress0` and `localAddress0` on `EmbeddedChannel`

FranzBusch:
# Motivation
`EmbeddedChannel` is often used in testing and currently any code under testing that uses `context.localAddress` cannot be mocked, since `EmbeddedChannelCore` is always throwing.

# Modification
Use the same values for `localAddress` in `localAddress0()`. Same for `remoteAddress`

# Result
We can now properly test code that needs local/remote addresses with `EmbeddedChannel`

### DIFF
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -300,15 +300,29 @@ class EmbeddedChannelCore: ChannelCore {
     var inboundBufferConsumer: Deque<(NIOAny) -> Void> = []
 
     @usableFromInline
+    var localAddress: SocketAddress?
+
+    @usableFromInline
+    var remoteAddress: SocketAddress?
+
+    @usableFromInline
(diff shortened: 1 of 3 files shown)


Sources/NIOEmbedded/Embedded.swift
```diff
@@ -615,10 +629,18 @@ public final class EmbeddedChannel: Channel {
     public var embeddedEventLoop: EmbeddedEventLoop = EmbeddedEventLoop()
 
     /// - see: `Channel.localAddress`
-    public var localAddress: SocketAddress? = nil
+    public var localAddress: SocketAddress? = nil {
```
Lukasa, line 632: Is there any reason for this not to be a computed property, instead of a `didSet`?

Lukasa (changes requested):

Lukasa (approved):
Good stuff, thanks!

PR #2347, opened 2023-01-12
Avoid actually allocating a giant buffer

Lukasa:
Motivation:

Our CI system is beginning to struggle with allocating a giant buffer in testSliceOfMassiveBufferWithAdvancedReaderIndexIsOk. We can work around this by faking out the allocation.

Modifications:

- Provide a fake allocator that doesn't actually allocate memory.
- Rewrite the test to use it.

Result:

Test still validates the behaviour but doesn't touch memory anymore.

### DIFF
--- a/Tests/NIOCoreTests/ByteBufferTest.swift
+++ b/Tests/NIOCoreTests/ByteBufferTest.swift
@@ -1754,14 +1754,24 @@ class ByteBufferTest: XCTestCase {
             throw XCTSkip("This test is only supported on 64-bit systems.")
         }
 
+        // This allocator assumes that we'll never call realloc.
+        let fakeAllocator = ByteBufferAllocator(
+            hookedMalloc: { _ in .init(bitPattern: 0xdeadbeef) },
+            hookedRealloc: { _, _ in fatalError() },
+            hookedFree: { precondition($0 == .init(bitPattern: 0xdeadbeef)!) },
+            hookedMemcpy: {_, _, _ in }
+        )
+
         let targetSize = Int(UInt32.max)
-        var buffer = self.allocator.buffer(capacity: targetSize)
+        var buffer = fakeAllocator.buffer(capacity: targetSize)
 
         // Move the reader index forward such that we hit the slow path.
         let offset = Int(_UInt24.max) + 1
         buffer.moveWriterIndex(to: offset)
         buffer.moveReaderIndex(to: offset)
-        buffer.writeInteger(UInt32(0))
+
+        // Pretend we wrote a UInt32.
+        buffer.moveWriterIndex(forwardBy: 4)
 
         // We're going to move the readerIndex forward by 1, and then slice.
         buffer.moveReaderIndex(forwardBy: 1)


Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1754,14 +1754,24 @@ class ByteBufferTest: XCTestCase {
             throw XCTSkip("This test is only supported on 64-bit systems.")
```
dnadoba, line 1754: Can we now run this on 32-bit systems as well?

dnadoba (approved):
LGTM!

Tests/NIOCoreTests/ByteBufferTest.swift
Lukasa: Unfortunately no: we still need to be able to have a buffer size that is 32-bits long, but we store the sizes in `Int`.

FranzBusch (approved):

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1754,14 +1754,24 @@ class ByteBufferTest: XCTestCase {
...
+
         let targetSize = Int(UInt32.max)
-        var buffer = self.allocator.buffer(capacity: targetSize)
+        var buffer = fakeAllocator.buffer(capacity: targetSize)
 
         // Move the reader index forward such that we hit the slow path.
         let offset = Int(_UInt24.max) + 1
         buffer.moveWriterIndex(to: offset)
         buffer.moveReaderIndex(to: offset)
-        buffer.writeInteger(UInt32(0))
```
weissi, line 1764: @Lukasa you can still make this write happen by having the fake allocator do

```
let ptr = malloc(4)
return ptr - offset
```

And `free` would do `free(ptr + offset)`.

Lukasa: Sure, but the write isn't important: we never validated the bytes came back.

Tests/NIOCoreTests/ByteBufferTest.swift
buttaface: Unfortunately, the two `0xdeadbeef` lines you added below [broke compilation on 32-bit platforms like Android armv7 too](https://github.com/buttaface/swift-android-sdk/actions/runs/3909881597/jobs/6681462361#step:13:963), so this test will either need to be conditionally compiled only on 64-bit platforms or modified in some other way.

Lukasa: Presumably the easiest fix is to just use a different sentinel value. The number doesn't matter: do you want to open a quick PR to use a different number?

buttaface: As in change it to a 32-bit number that could be stored in an Int32 also? It's your idea, please go ahead.

Lukasa: Should be fixed by #2348.

buttaface: Yep, [compiles now](https://github.com/buttaface/swift-android-sdk/actions/runs/3917797023/jobs/6697815544), thanks.

Issue #2354, opened 2023-01-23
Flaky test: `NIOAsyncSequenceProducerTests.testTaskCancel_whenStreaming_andNotSuspended`


glbrntt:
`NIOAsyncSequenceProducerTests.testTaskCancel_whenStreaming_andNotSuspended` occasionally fails:

```
11:22:16 Test Case 'NIOAsyncSequenceProducerTests.testTaskCancel_whenStreaming_andNotSuspended' started at 2023-01-23 11:22:16.706
11:22:16 /code/Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift:400: error: NIOAsyncSequenceProducerTests.testTaskCancel_whenStreaming_andNotSuspended : XCTAssertEqual failed: ("1") is not equal to ("0") - 
11:22:16 Test Case 'NIOAsyncSequenceProducerTests.testTaskCancel_whenStreaming_andNotSuspended' failed (0.003 seconds)
```

Test code:

https://github.com/apple/swift-nio/blob/1a0f46b8cd0f6a0e2fa8ec68768dfb2e14aca18f/Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift#L388-L405

From a quick look it appears `didTerminate()` can be called if the async iterator is `deinit`ed, which can happen after `yield(contentsOf:)` is called and before reading `self.delegate.didTerminateCallCount` on line 400.

PR #2355, opened 2023-01-23
Fix flaky testTaskCancel_whenStreaming_andNotSuspended

FranzBusch:
# Motivation
Currently `testTaskCancel_whenStreaming_andNotSuspended` is flaky since `didTerminate` can be called after the iterator is dropped. Fixes https://github.com/apple/swift-nio/issues/2354

# Modification
Let's modify that slightly so we hight the condition we want to hit.

# Result
No more flaky tests.

### DIFF
--- a/Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
+++ b/Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
@@ -391,9 +391,16 @@ final class NIOAsyncSequenceProducerTests: XCTestCase {
         let sequence = try XCTUnwrap(self.sequence)
         let task: Task<Int?, Never> = Task {
             let iterator = sequence.makeAsyncIterator()
-            return await iterator.next()
+            let value = await iterator.next()
+
+            // Sleeping here a bit to make sure we hit the case where
+            // we are streaming and still retain the iterator.
+            try? await Task.sleep(nanoseconds: 1_000_000)
+
+            return value
         }
-        try await Task.sleep(nanoseconds: 1_000_000)
+
+        try await Task.sleep(nanoseconds: 2_000_000)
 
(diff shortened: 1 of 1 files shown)


glbrntt (approved):

Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
```diff
@@ -391,8 +391,14 @@ final class NIOAsyncSequenceProducerTests: XCTestCase {
         let sequence = try XCTUnwrap(self.sequence)
         let task: Task<Int?, Never> = Task {
             let iterator = sequence.makeAsyncIterator()
-            return await iterator.next()
+            let value = await iterator.next()
+
+            // Sleeping here a bit to make sure we hit the
```
Lukasa, line 396: Truncated comment.

Tests/NIOCoreTests/AsyncSequences/NIOAsyncSequenceTests.swift
```diff
@@ -391,8 +391,14 @@ final class NIOAsyncSequenceProducerTests: XCTestCase {
         let sequence = try XCTUnwrap(self.sequence)
         let task: Task<Int?, Never> = Task {
             let iterator = sequence.makeAsyncIterator()
-            return await iterator.next()
+            let value = await iterator.next()
+
+            // Sleeping here a bit to make sure we hit the 
+            try? await Task.sleep(nanoseconds: 500_000)
```
Lukasa, line 397: Is 500 microseconds sufficient?

FranzBusch: Changed to 1 ms and 2 ms. That should be enough

Lukasa (approved):

PR #2357, opened 2023-01-24
POC output with URring

ser-0xff:
Use URing on Linux.

URing on Linux provides an API to demultiplex events and perform socket operations.
NIO originally designed to demultiplex events and perform all IO in a single thread, while URing socket operations are asynchronous. It requires more significant changes in the NIO taking into account the requirement to additionally manage memory buffers for pending read and write operations.

The one idea was to try to minimise PR as possible, so that PR send outbound data using URing, but still read inbound data using base socket API. Reading inbound data with URing is a next step.

Some highlights regarding changes in the PR

1) The URing API using on Linux can be enabled by the conditional compilation.
In some places I saw '#if defined(SWIFTNIO_USE_IO_URING) && (os(Linux) || os(Android))
As of now we have no possibility even to compile it for Android, not saying to test, so all URing related code is under just os(Linux). Let's decide later what can we do with Android.

2) NIO used some buffers stored in the EventLoop (iovecs, storageRefs etc) which are shared across all channels registered in the event loop. That approach does not work with URing because of its asynchronous nature. There are not too many ways to workaround it for URing, so I moved them to the channel, so with URing each channel has own iovecs and storageRefs buffers. Probably we could cache them... not clear what is better... let's discuss...

3) URing has no analogue for the sendfile() system call.
Instead they suggest to splice() the file via intermediate pipe. It works, but requires an intermediate pipe. There are not too many options how to manage intermediate pipes, but I can't see a good one. So, as of now we just cache intermediate pipes and reuse them...

4) URing has no analogue for sendmmsg() system call.
Unfortunately did not found a good way to workaround that. Currently just send messages one by one. One possible optimisation here is to collect all messages with the same destination address and same ancillary data and send them with one sendmsg(), but did not found such obvious optimisation in the NIO, so probably it is not a case.

5) I had to disable few tests which checks the order of system calls. Have no idea how we could make them work for URing, because URing is asynchronous. Let's discuss them later. All other tests works on my environment, but I will not be surprised if they will not work somewhere else.

Will add some information later within comments to PR if will recall something important.

### DIFF
--- a/Sources/NIOPosix/LinuxUring.swift
+++ b/Sources/NIOPosix/LinuxUring.swift
@@ -12,15 +12,17 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if SWIFTNIO_USE_IO_URING
-#if os(Linux)
+#if SWIFTNIO_USE_IO_URING && os(Linux)
 
 import CNIOLinux
 import NIOCore
 
 @usableFromInline
 enum CQEEventType: UInt8 {
-    case poll = 1, pollModify, pollDelete // start with 1 to not get zero bit patterns for stdin
+    case poll = 1, // start with 1 to not get zero bit patterns for stdin
+    pollModify,
+    pollDelete,
+    write
 }
 
 internal enum URingError: Error {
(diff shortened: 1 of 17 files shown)


Sources/NIOPosix/SelectorGeneric.swift
```diff
@@ -137,21 +198,33 @@ internal class Selector<R: Registration>  {
     // Here we add the stored properties that are used by the specific backends
     #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
     typealias EventType = kevent
+    #elseif SWIFTNIO_USE_IO_URING && os(Linux)
+    typealias EventType = URingEvent
+    var eventFD: CInt = -1 // -1 == we're closed
+    var ring = URing()
+    let multishot = URing.io_uring_use_multishot_poll // if true, we run with streaming multishot polls
+    let pipeCache = Cache<Pipe>(maxSize: 8,
```
Lukasa, line 206: I'm inclined to drop the support for sendfile emulation in v1 of this patch. I'm suspicious that almost no-one is using it right now _anyway_, so I think it's probably fine if we begin by just throwing runtime errors if people try to use it. We can always make a follow-up PR to add support back in if it turns out to be necessary.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -329,6 +440,81 @@ final class PendingStreamWritesManager: PendingWritesManager {
         return self.state.currentBestWriteMechanism
     }
 
+#if SWIFTNIO_USE_IO_URING && os(Linux)
+
+    func triggerAppropriateAsyncWriteOperations(scalarBufferAsyncWriteOperation: (UnsafeRawBufferPointer) throws -> Void,
```
Lukasa, line 445: This is just one example, but this PR has a lot of API surface guarded behind availability of IOUring. I'd like us to minimise this as much as possible, so I think we can do a few things to produce fewer compile conditions.

Firstly, let's make as much API available on all platforms as possible. If the type signature is cross-platform, we should include it. In this case, there's nothing in this function that requires liburing, so we can safely just make this available unconditionally. 

Secondly, where the _implementation_ can also be written safely on all platforms, we should keep it! The goal is to funnel down to the places where we actually interact with io_uring. At those levels, we should have thin layer that calls into an underlying implementation if it's present, or that crashes if called on the wrong platform.

This will hopefully minimise the amount of `#if` we have to scatter around the codebase and make the whole thing substantively easier to read.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -423,12 +443,24 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
...
+        let addresses = UnsafeMutableBufferPointer<sockaddr_storage>.allocate(capacity: 1)
+        let controlMessageStorage = UnsafeControlMessageStorage.allocate(msghdrCount: 1)
+#else
+        let msgs = eventLoop.msgs
+        let iovecs = eventLoop.iovecs
+        let addresses = eventLoop.addresses
+        let storageRefs = eventLoop.storageRefs
+        let controlMessageStorage = eventLoop.controlMessageStorage
+#endif
+        self.pendingWrites = PendingDatagramWritesManager(msgs: msgs,
```
Lukasa, line 459: I don't love having the pending writes managers take ownership of pointers allocated outside their scope. Can we have two different initialisers, one where they allocate the pointers themselves and one where they don't, and then keep track of what happened with a flag? Then we can drop most of the compile time conditionals, which will make our lives a lot easier.

Lukasa (commented):
Thanks for this patch! I've left a few structural notes in the diff. Note that I haven't done a detailed code review yet, I'm just laying out some thoughts. As we work on the structure I'll continue to re-review the implementation. I've also not deeply looked into the changes to SelectorUring and LinuxUring yet, those are my next round of review. Generally though, this is looking really solid!

Sources/NIOPosix/SelectorGeneric.swift
ser-0xff: Nice, will give us a chance to remove some code.

Sources/NIOPosix/PendingWritesManager.swift
ser-0xff: The idea was do not let code which will be unused on the particular platform even compile.
But if you think it is fine I will remove those conditional compilations in such places.

Sources/NIOPosix/SocketChannel.swift
ser-0xff: I do not love that either...
Had an impression having a different initialiser will require bigger changes in tests...
But may be we could use something different?
For example we could introduce a cache of IOVector buffers in the EventLoop, and then each channel could use that cache. Without URing that cache will always have a one buffer. With URing - probably more than one but less than total number of channels. We could save some memory in that case. And the buffer management logic will not require any conditional compilation at all in that case. What do you think?

Sources/NIOPosix/SelectorGeneric.swift
ser-0xff: Frankly speaking, in a case if we are going to remove possibility to send a file, then the best approach from my point of view would be to disable a possibility to create a file region parcel on the compile time.
But it will require a conditional compilation, and if you do not want to have many of them in the code, then a runtime error is only a way... not a perfect from my point of view because end user fill find that later...

Sources/NIOPosix/PendingWritesManager.swift
Lukasa: I think it's fine.

Sources/NIOPosix/SelectorGeneric.swift
Lukasa: The bigger problem is not the conditional compilation, it's the fact that we don't want to break the API for this. So sadly for now this needs to stay as-is. But it'll be a useful sanity check to find out if anyone actually is sending files.

Sources/NIOPosix/LinuxUring.swift
```diff
@@ -38,84 +40,135 @@ internal extension TimeAmount {
...
+        }
+
+        @inlinable init(rawValue: UInt64) {
+            let unpacked = IntegerBitPacking.unpackUInt32UInt16UInt8(rawValue >> 1)
+            self = .init(CInt(unpacked.0),
+                        SelectorRegistrationID(rawValue: UInt32(unpacked.1)),
+                        CQEEventType(rawValue:unpacked.2)!)
+        }
+
+        @inlinable func asUInt64() -> UInt64 {
```
Lukasa, line 111: Strictly preferred Swift style is to spell this as an `init` on `UInt64`.

Sources/NIOPosix/LinuxUring.swift
```diff
@@ -312,239 +335,314 @@ final internal class URing {
...
 
-    internal func _debugPrint(_ s: @autoclosure () -> String)
-    {
-        #if SWIFTNIO_IO_URING_DEBUG_URING
-        print("L [\(NIOThread.current)] " + s())
-        #endif
+    func prep_writev(fileDescriptor: CInt, iovecs: UnsafeBufferPointer<IOVector>, registrationID: SelectorRegistrationID) {
+        let userData = UserData(fileDescriptor, registrationID, CQEEventType.write).asUInt64()
+
+        func bytesToWrite(_ iovecs: UnsafeBufferPointer<IOVector>) -> Int {
```
Lukasa, line 362: It might be a little nicer to spell this in an `extension` on `UnsafeBufferPointer where Element == IOVector`.

Sources/NIOPosix/LinuxUring.swift
```diff
@@ -312,239 +335,314 @@ final internal class URing {
...
+            var bytes = 0
+            for idx in 0..<iovecs.count {
```
Lukasa, line 364: Prefer `for iovec in iovecs`.

Sources/NIOPosix/LinuxUring.swift
```diff
@@ -38,84 +40,135 @@ internal extension TimeAmount {
...
-        self.fileDescriptor = f
-        self.registrationID = s
+    init(_ fileDescriptor: CInt, _ registrationID: UInt16) {
+        self.fileDescriptor = fileDescriptor
+        self.registrationID = registrationID
     }
 }
 
+private func _debugPrint(_ s: @autoclosure () -> String) {
+    #if SWIFTNIO_IO_URING_DEBUG
```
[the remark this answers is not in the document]
ser-0xff: That condition was before I started to work on it, just left it...

Sources/NIOPosix/SocketChannel.swift
ser-0xff: I could probably try to implement that cache in a separate branch,
then we will merge it, and then will merge into uring-out branch, and then we will continue with it.
What do you think?

Sources/NIOPosix/LinuxUring.swift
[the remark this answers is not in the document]
hassila: I think it's a remnant of the original bringup to not having to define out every debugPrint that is in place IIRC.

There are additionally three more specific defines for more in-depth debugging, but they were in few places and are completely defined out (https://github.com/search?q=repo%3Aapple%2Fswift-nio%20SWIFTNIO_IO_URING_DEBUG&type=code) .

Perhaps there's a better way to do this (maybe with the macro support coming up? Basically, one would like to be able to build with debug logging during development but to completely remove traces of it for production as this may be a fairly performance sensitive thing).

Sources/NIOPosix/SocketChannel.swift
hassila: So that's https://github.com/apple/swift-nio/pull/2358 - just commenting to tie it together with this comment chain.

ser-0xff:
I got the point regarding the conditional compilation.
How do you think should we move the conditional compilation in [that] (https://github.com/ordo-one/swift-nio/blob/uring-out/Sources/NIOPosix/SelectorGeneric.swift#L161) place more deeply leaving the async API available even for a case if the backend is synchronous?

Lukasa:
I'm inclined to suggest that we have most backends implement the functions with `fatalError`, and only have the uring selector implement them. Then we can ensure that at runtime we only call those functions when we know we have the uring selector, which we should know either as a compile-time static or as event loop state.

Lukasa:
Yes, I think that works fine.

Sources/NIOPosix/SelectorGeneric.swift
Lukasa: Just a gentle ping that you can delete this code for now.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -381,7 +381,10 @@ extension PendingDatagramWritesState {
 /// the availability of the functions.
 final class PendingDatagramWritesManager: PendingWritesManager {
 
+    /// storage for IOVector and the references to the buffers used when we perform gathering writes. Only present
+    /// on Linux because Darwin does not support gathering datagram writes.
     private var bufferPool: Pool<PooledBuffer>
+    private var buffer: PooledBuffer?
```
Lukasa, line 387: Why do we need the local buffer reference?

ser-0xff: There are 2 reasons for that:
1) we will need to put that pooled buffer back to the pool when async operation will be completed
2) we should keep a memory block which is used for the async operation while the operation will not be completed

And we also have 2 options how to do it:
1) tie that pooled buffer to the asynchronous write request
2) keep it locally, so when async request will be completed

ser-0xff:
Could you have a look at [PR](https://github.com/ordo-one/swift-nio/pull/2)?
I tried to disable possibility to send files with URing, but end up with a lot of conditional compilation in tests.
As far as I understand test runner parse swift code and generate a code calling tests, so I have no idea how could disable those tests when running with URing.

Lukasa:
I'm entirely happy with us adding this conditional compilation into the tests, as it's small and clear. :+1:

Sources/NIOPosix/SelectorGeneric.swift
ser-0xff: Done within [commit](https://github.com/ordo-one/swift-nio/commit/7fffd46b8a806b0939405d9bc660dab365fc4bd6)

hassila:
So what would be good next steps if we want to move this forward?

ser-0xff:
I merged most recent changes from the main into the feature branch and stabilised tests, so all tests (except skipped are passed now).
Had a lot of fun with the test stabilisation,
it still seems to me the idea to have both SYNC and ASYNC code at the same time in the source code is not good and it would be much better separate them with #ifdefs. The point here is that on runtime the only one approach should work, but codebase has a lot of conditional branches, and sometimes it happen that SYNC IO is triggered even when ASYNC is supposed to because source code was not fixed properly. It would not even happen if we would completely move the SYNC functionality under #ifdef. Not sure I found all places where SYNC IO can be triggered in ASYNC mode.
But still, all tests pass now, would be nice if you could have a look at PR.

PR #2358, opened 2023-01-25
Pool buffers for ivecs and storage refs in the event loop.

ser-0xff:
Pool and reuse IOVector and storage ref buffers in the event pool.

### Motivation:

Having an asynchronous sockets backend (like URing on Linux) require extend a lifetime of memory buffers used as arguments for more than a code block.

### Modifications:

Event loop have a pool of buffers to be used for output vector write operations.

### Result:

In a case of synchronous sockets backend that pool will always have a single buffer, for asynchronous backends every async call will take buffer from the pool and release the buffer when it will not be needed.


Sources/NIOCore/Pool.swift
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public class Pool<Element> {
```
Lukasa, line 15: We don't want to be in the business of vending this type as public API. It can safely be moved into `NIOPosix`.

Sources/NIOCore/Pool.swift
```diff
@@ -0,0 +1,52 @@
...
+    private let maxSize: Int
+    private let ctor: () -> Element
+    private let dtor: (Element) -> Void
+    private var elements: [Element]
+
+    public init(maxSize: Int, ctor: @escaping () -> Element, dtor: @escaping (Element) -> Void) {
```
Lukasa, line 21: Rather than using separate constructor/destructor functions, we'd typically want to define a protocol to which `Element` conforms that implements the relevant methods. Something like:

```
protocol PoolElement {
    init()

    func evictedFromPool()
}

File: Sources/NIOCore/Pool.swift [hunk 3]
```diff
@@ -0,0 +1,52 @@
...
+        self.maxSize = maxSize
+        self.ctor = ctor
+        self.dtor = dtor
+        self.elements = [Element]()
+    }
+
+    deinit {
+        while !elements.isEmpty {
```
Lukasa, line 29: This can just be a `for` loop, we don't need to deinit the elements immediately.

Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -77,6 +76,13 @@ private func doPendingDatagramWriteVectorOperation(pending: PendingDatagramWrite
     var c = 0
     var toWrite: Int = 0
 
+    let buffer = bufferPool.get()
+    defer { bufferPool.put(buffer) }
+
+    let count = (buffer.count / (MemoryLayout<IOVector>.stride + MemoryLayout<Unmanaged<AnyObject>>.stride))
+    let iovecs = (buffer.baseAddress!).bindMemory(to: IOVector.self, capacity: count)
+    let storageRefs = (buffer.baseAddress! + (count * MemoryLayout<IOVector>.stride)).bindMemory(to: Unmanaged<AnyObject>.self, capacity: count)
```
Lukasa, line 84: This code is very precise. Can we factor it out into its own method to which we can add assertions on appropriate length?

Lukasa: So what you're doing here is allocating a single buffer that you then slice into the relevant buffers we want. That's fine, but I'd like to encapsulate all of that dance into a struct that we can use to hide this mess of memory binding. It will also let us add a bunch of pseudo-bounds-checks to our construction code, so we can confirm that we're not actually binding or using memory that is out of bounds of the original allocation. Does that make sense?

ser-0xff: well,
I would look at that with a very simple metric: number of code lines and number of extra entities (we will need to introduce an additional structure). Here we have just 3 lines of code, adding a struct with that functionality will give more than 3 lines... But if you think it is better in that particular case - let's do it.

Sources/NIOCore/Pool.swift
ser-0xff: I had some debates regarding such approach which is I see is common for swift, but sometimes it does not work well from my point of view. So we are going to extend some type with 'PoolElement' protocol. But is it really needed? Why the particular type should expose the 'PoolElement' functionality and why it should even know that it is going to be pooled with some pool? Often it is only the particular pool who knows what does it going to pool and how...
As you can see in the PR there are 2 pool
- pool to be used in the event loop, which just allocates buffer
- pool to be used in tests, which allocates buffers with markers and check those markers when buffer is not needed any more...
The PoolElement protocol does not give a possibility to have a different pooling logic for the same type in different pools... or at least I can't see a simple way for that...
Or am I missing something?

Sources/NIOPosix/PendingDatagramWritesManager.swift
Lukasa: Generally I don't think lines of code added is the particular metric I'm most concerned about. What I want is to build abstractions that discourage misuse. A related opportunity is to actively reference count the pointers, but I think this is a one-step-at-a-time kind of deal, so I'm interested for now in giving us a sensible abstraction.

Sources/NIOCore/Pool.swift
Lukasa: The thing you're missing is that the way to pool the "same thing" in different ways is to define a new type with the new behaviour. Swift `struct`s are essentially "free": they're trivial aggregates. That means whenever we want a different behaviour we can wrap the type we want in a new structure that adds that new behaviour.


In general if we were defining a public API I'd be a bit more cautious, but as this is purely an internal data structure I think it's probably nicer to avoid unstructured callbacks.

ser-0xff: So, I made another branch with a Pool<> which operates with something implementing 'PoolElement'.
I did a PoolableBuffer struct implementing default behaviour: https://github.com/ordo-one/swift-nio/blob/aa23e675eb95292f98c08391cf30bec37fed8292/Sources/NIOPosix/SelectableEventLoop.swift#L33
My understanding was we can define a different behaviour for a poolable buffer adding some bounds checks, like I did at https://github.com/ordo-one/swift-nio/blob/aa23e675eb95292f98c08391cf30bec37fed8292/Tests/NIOPosixTests/ChannelTests.swift#L76, but TestPooledBuffer can not be used as a pooled type, the code does not compile...
Am I missing something or did I get you wrong?

Lukasa: When you say the code doesn't compile, what did you try to do and what happened?

ser-0xff: Well, my understanding was you want to have 2 different types to be used as an argument type for the pool.
1) PooledBuffer, with a common allocation logic
2) TestPooledBuffer with allocation and bounds check
and depending on the case we could use pools with different types.
But the problem here we can not keep something like Pool<PoolElement> in the PendingWritesManager, because compiler (and I understand it) want to know a particular type here.
So, we should specify a particular at https://github.com/ordo-one/swift-nio/blob/pool-iovecs-buffers2/Sources/NIOPosix/SelectableEventLoop.swift#L125
And after that we should always use Pool<PooledBuffer> as an argument for the PendingWritesManager initialiser, and can't provide something different, like for example Pool<TestPooledBuffer>....

Lukasa: To be clear, I agree with this: I think the pooled buffer should have bounds checks in debug mode, but that's a different flow. I'm not proposing we use a protocol so that we can support multiple different types in the core code, just so that we can provide a nicer internal abstraction for ourselves.

ser-0xff: Then I am totally confused...
In a very first message in that thread you suggest to use a protocol for the pool element.
Not you tell you are not proposing to use protocol.
So what should we do now?

Lukasa: Sorry, that might not have been clear. I _am_ proposing a protocol. When I said "I'm not proposing a protocol", that was part of a sentence that said "I'm not proposing it because X, but because Y".

ser-0xff: So could you give some hints on how can have a protocol, supposed to be implemented by types pooled in the Pool'<>', and have 2 different implementations of that protocol for prod and tests?

hassila: Do I understand correctly that only a single type is suggested but with assertions for the invariants etc that simply won’t be part of optimized builds?

Lukasa: Yes, exactly so.

Lukasa: Yes, much closer. The extra thing to add is a bunch of `assert`s in the `get` function to ensure that the bounds are respected there too.

ser-0xff: Do you mean PooledBuffer.get()?

Lukasa: I do.

ser-0xff: What kind of assertions do you think off?

Lukasa: Confirming that the start addresses and end addresses don't overlap, and that they fit within the buffer.

ser-0xff: I pushed a commit with some bounds checks.
https://github.com/apple/swift-nio/commit/a437a52ed134c90f30152e8811c59fe0cb51f295
Could you have a look?

Lukasa: Looks great!

ser-0xff: Do you think we can merge these changes?
What will be better, create a new PR from that branch, or apply those changes to the branch where that PR is made from?

Lukasa: I think you can merge those changes into this branch.

Sources/NIOCore/Pool.swift
ser-0xff: Fixed

Sources/NIOPosix/Pool.swift
```diff
@@ -0,0 +1,53 @@
...
+    public func get() -> Element {
+        if elements.isEmpty {
+            return Element()
+        }
+        else {
+            return elements.removeLast()
+        }
+    }
+
+    public func put(_ e: Element) {
```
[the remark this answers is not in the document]
ser-0xff: Fixed all above.

Lukasa:

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -58,10 +58,12 @@ struct PooledBuffer: PoolElement {
...
         let storageRefs = (self.buffer + (count * MemoryLayout<IOVector>.stride)).bindMemory(to: Unmanaged<AnyObject>.self, capacity: count)
-        assert((iovecs >= self.buffer) && (iovecs <= (self.buffer + self.bufferSize)))
-        assert((storageRefs >= self.buffer) && (storageRefs <= (self.buffer + self.bufferSize)))
-        assert((iovecs + count) == storageRefs)
-        assert((storageRefs + count) <= (self.buffer + bufferSize))
+        assert(iovecs >= self.buffer.bindMemory(to: IOVector.self, capacity: count))
+        assert(iovecs <= (self.buffer + self.bufferSize).bindMemory(to: IOVector.self, capacity: count))
+        assert(storageRefs >= self.buffer.bindMemory(to: Unmanaged<AnyObject>.self, capacity: count))
+        assert(storageRefs <= (self.buffer + self.bufferSize).bindMemory(to: Unmanaged<AnyObject>.self, capacity: count))
+        assert(UnsafeMutableRawPointer(iovecs + count) == UnsafeMutableRawPointer(storageRefs))
```
Lukasa, line 65: Don't bind the memory in these assertions, the duplicate bind makes it hard to reason about. Erase the pointers to raw instead.

Lukasa:

Lukasa:

Sources/NIOPosix/SelectableEventLoop.swift
ser-0xff: It does not compile with swift 5.5/5.6, CI works fails...

ser-0xff: or do you mean having it like
```
        assert(UnsafeMutableRawPointer(iovecs) >= self.buffer)
        assert(UnsafeMutableRawPointer(iovecs) <= (self.buffer + self.bufferSize))
        assert(UnsafeMutableRawPointer(storageRefs) >= self.buffer)
        assert(UnsafeMutableRawPointer(storageRefs) <= (self.buffer + self.bufferSize))
        assert(UnsafeMutableRawPointer(iovecs + count) == UnsafeMutableRawPointer(storageRefs))
        assert(UnsafeMutableRawPointer(storageRefs + count) <= (self.buffer + bufferSize))

```

Lukasa:

Lukasa:

Lukasa:

Lukasa (approved):
For the sake of keepings things moving forward, I'm happy for us to merge this as-is. However, I'm going to provide a follow-up PR myself to refactor this a bit to make it a little safer. In particular, I want us to avoid the risk of leaking memory, so I'm going to change the layout of the `PoolElement` somewhat.

PR #2362, opened 2023-02-03
Add a pooled recv buffer allocator

glbrntt:
Motivation:

Channels can read `ChannelOptions.maxMessagesPerRead` times from a socket in each read cycle. They typically re-use the same buffer for each read and rely on it CoWing if necessary. If we read more than once in a cycle then we will always CoW the buffer. Instead of reusing one buffer we can reuse a pool of buffers limited by `maxMessagesPerRead` and cycle through each, reducing the chance of CoWing the buffers.

Modifications:

- Add `NIOPoolableRecvByteBufferAllocator` which extends `RecvByteBufferAllocator` to provide the size of the next buffer.
- Add an internal pool which uses a `NIOPoolableRecvByteBufferAllocator`
- Add a wrapper which is either backed by poolable allocator or non-poolable allocator.

Results:

Fewer allocations


Lukasa:
> If we read more than once in a cycle then we will always CoW the buffer.

Side note: I don't think this is entirely true. It's only true if a `ChannelHandler` holds onto that buffer. Most don't: if they have swift-nio-ssl in the way then it will always drop the inbound buffer.

tomerd:

tomerd:

tomerd:

tomerd:

tomerd:

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,198 @@
...
+
+extension RecvBufferAllocator {
+    @usableFromInline
+    enum _Backing {
+        case pooled(PooledRecvBufferAllocator)
+        case unpooled(RecvByteBufferAllocator)
+        case modifying
+
+        init(_ allocator: RecvByteBufferAllocator, capacity: Int) {
+            if let poolable = allocator as? NIOPoolableRecvByteBufferAllocator {
```
Lukasa, line 142: This is pretty painfully expensive. Perhaps we should amend `RecvByteBufferAllocator` to include the new method with a default implementation that returns `nil`, and have a different pooling strategy for the fallback case?

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,198 @@
...
+            case .modifying:
+                fatalError("Left in .modifying state")
+            }
+        }
+
+        mutating func record(actualReadBytes: Int) -> Bool {
+            let mayGrow: Bool
+            switch self {
+            case .pooled(var allocator):
+                self = .modifying
```
Lukasa, line 166: We shouldn't need a modifying case.

Lukasa (changes requested):

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,198 @@
...
+            self.buffers.append(self.recvAllocator.buffer(allocator: allocator))
+            index = self.nextIndex()
+        } else {
+            index = self.nextIndex()
+            let nextBufferSize = self.recvAllocator.nextBufferSize()
+            self.buffers[index].clear(minimumCapacity: nextBufferSize)
+        }
+
+        let result = try body(&self.buffers[index])
+        return (self.buffers[index], result)
```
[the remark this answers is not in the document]
glbrntt: Good point re: reusing buffers without a CoW; didn't think of that case.

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,142 @@
...
+                    if let result = try self.buffers[index].modifyIfUniquelyOwned(minimumCapacity: nextBufferSize, body) {
+                        return (self.buffers[index], result)
+                    }
+                }
+            }
+        }
+
+        // No available buffers or the allocator does not offer up buffer sizes; directly
+        // allocate a new one.
+        var newBuffer = self.recvAllocator.buffer(allocator: allocator)
```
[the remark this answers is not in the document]
glbrntt: Ah, so if the last recorded size wouldn't cause the buffer to grow _and_ the buffer can be reused then we've regressed. Yeah I think we can fix that.

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,142 @@
...
+        precondition(capacity > 0)
+        self.capacity = capacity
+        self.buffer = nil
+        self.buffers = []
+        self.recvAllocator = recvAllocator
+    }
+
+    /// Returns the number of buffers in the pool.
+    var count: Int {
+        return self.buffer == nil ? self.buffers.count : 1
```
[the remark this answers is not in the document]
Lukasa: Yeah that wasn't at all obvious to me. Can you add a comment to that effect?

Sources/NIOPosix/BaseStreamSocketChannel.swift
```diff
@@ -109,21 +109,24 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
...
             // Reset reader and writerIndex and so allow to have the buffer filled again. This is better here than at
             // the end of the loop to not do an allocation when the loop exits.
-            buffer.clear()
-            switch try buffer.withMutableWritePointer(body: { try self.socket.read(pointer: $0) }) {
+            switch readResult {
             case .processed(let bytesRead):
                 if bytesRead > 0 {
-                    let mayGrow = recvAllocator.record(actualReadBytes: bytesRead)
-
+                    _ = self.recvBufferPool.record(actualReadBytes: bytesRead)
```
Lukasa, line 129: Can we just not return anything from the record call here?

Sources/NIOPosix/PooledRecvBufferAllocator.swift
```diff
@@ -0,0 +1,192 @@
...
+            } else if self.buffers.count < self.capacity {
+                self.buffers.append(newBuffer)
+                self.lastUsedIndex = self.buffers.index(before: self.buffers.endIndex)
+                return try self.modifyLast(body)
+            } else {
+                let result = try body(&newBuffer)
+                return (newBuffer, result)
+            }
+        }
+    }
```
Lukasa, line 174: I think this method is beginning to get a bit unwieldy. Can we refactor it into smaller helper functions that each handle one case?

Lukasa (approved):
Nice one, thanks @glbrntt!

PR #2363, opened 2023-02-07
Make PooledBuffer safer.

Lukasa:
Motivation:

PooledBuffer is an inherently unsafe type, but its original incarnation was less safe than it needed to be. In particular, we can rewrite it to ensure that it is compatible with automatic reference counting.

Modifications:

- Rewrite PooledBuffer to use ManagedBuffer
- Clean up alignment math
- Use scoped accessors
- Add hooks for future non-scoped access

Result:

Safer, clearer code

### DIFF
--- a/Sources/NIOPosix/Pool.swift
+++ b/Sources/NIOPosix/Pool.swift
@@ -50,3 +50,155 @@ class Pool<Element: PoolElement> {
         }
     }
 }
+
+/// A ``PooledBuffer`` is used to track an allocation of memory required
+/// by a `Channel` or `EventLoopGroup`.
+///
+/// ``PooledBuffer`` is a reference type with inline storage. It is intended to
+/// be bound to a single thread, and ensures that the allocation it stores does not
+/// get freed before the buffer is out of use.
+struct PooledBuffer: PoolElement {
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/Pool.swift
```diff
@@ -50,3 +50,155 @@ class Pool<Element: PoolElement> {
...
+    /// instance. If you don't require the pointer after the closure returns, use ``withUnsafePointers``.
+    ///
+    /// If you escape the pointer from the closure, you _must_ call `storageManagement.retain()` to get ownership to
+    /// the bytes and you also must call `storageManagement.release()` if you no longer require those bytes. Calls to
+    /// `retain` and `release` must be balanced.
+    ///
+    /// - parameters:
+    ///     - body: The closure that will accept the yielded pointers and the `storageManagement`.
+    /// - returns: The value returned by `body`.
+    func withUnsafePointersWithStorageManagement<ReturnValue>(
```
Lukasa, line 96: For reviewers: this function isn't in use right now, but I added it because it's a necessary component for io_uring support, where the buffer will need to outlive the closure-scoped usage site.

weissi (approved):
Nice improvement!

PR #2365, opened 2023-02-10
NIOAsyncWriter: Provide a fast path for single element writes

FranzBusch:
# Motivation
We are currently always allocating a new `Deque` when we get a single element write in the `NIOAsyncWriter`

# Modification
Provide a fast path method on the `NIOAsyncWriterSinkDelegate` protocol which will be called when we receive a single element write and are currently writable.

# Result
Performance win for the single write cases.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -37,6 +37,18 @@ public protocol NIOAsyncWriterSinkDelegate: Sendable {
     /// - Important: You **MUST NOT** call ``NIOAsyncWriter/Sink/setWritability(to:)`` from within this method.
     func didYield(contentsOf sequence: Deque<Element>)
 
+    /// This method is called once a single element was yielded to the ``NIOAsyncWriter``.
+    ///
+    /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
+    /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
+    /// until the ``NIOAsyncWriter`` becomes writable again. All buffered writes, while the ``NIOAsyncWriter`` is not writable,
+    /// will be coalesced into a single sequence.
+    ///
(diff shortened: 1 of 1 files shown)


FranzBusch:

FranzBusch:
>  asyncwriter_single_writes_1M_times | 0.342436316 | 0.414022724 | current

This is the interesting bit from the perf test.

FranzBusch:
> 💔 API breakage: func NIOAsyncWriterSinkDelegate.didYield(_:) has been added as a protocol requirement

Breaking change checker is wrong here since we added a default implementation.

Lukasa (approved):
LGTM

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -37,6 +37,18 @@ public protocol NIOAsyncWriterSinkDelegate: Sendable {
...
+    ///
+    /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
+    /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
+    /// until the ``NIOAsyncWriter`` becomes writable again. All buffered writes, while the ``NIOAsyncWriter`` is not writable,
+    /// will be coalesced into a single sequence.
+    ///
+    /// - Note: This a fast path that you can optionally implement. By default this will just call ``NIOAsyncWriterSinkDelegate/didYield(contentsOf:)``.
+    ///
+    /// - Important: You **MUST NOT** call ``NIOAsyncWriter/Sink/setWritability(to:)`` from within this method.
+    func didYield(_ element: Element)
```
dnadoba, line 50: I'm wondering if we can instead use a single `Deque` we reuse specifically for single yields. If users don't hold onto it we would only allocate once or CoW if they still hold it.

FranzBusch: I was thinking the same but that doesn't really work out. The problem in high throughput is that we are expecting users to hop onto the EL in their `didYield` implementation. At the same time the producing thread is calling `yield` again on the writer. Most of the times we are still holding onto the `Deque` in the EL thread at this point. This is the safer fast path that we can guarantee to hit.

dnadoba (approved):

PR #2367, opened 2023-02-14
TCP channel throughput benchmark.

ser-0xff:
Add a benchmark measuring TCP channel throughput.

### Motivation:

The product we are working on exchange with significant number of small messages via TCP/IP. Would be interesting to know the throughput we can achieve. Also, we are integrating the Linux URing API, will be interesting to look at how the performance will be changed with URing.

### Modifications:

Benchmark which sends 100K messages of 500 bytes each and measure a time between the very first message sent on the sender side and last message received on the receiver side.

### Result:

Benchmark for TCP channel throughput.

### DIFF
--- a/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
+++ b/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
@@ -1,3 +1,17 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
 import NIOCore
 import NIOPosix
 
--- a/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
+++ b/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
@@ -1,0 +1,129 @@
+import NIOCore
+import NIOPosix
+
+/// Test measure a TCP channel throughput.
+/// Server send 100K messages to the client,
+/// measure the time from the very first message sent by the server
+/// to the last message received by the clien.
+
+class TCPThroughputBenchmark: Benchmark {
+
+    private let messages: Int
+    private let messageSize: Int
+
+    private var group: EventLoopGroup!
+    private var serverChannel: Channel!
+    private var serverHandler: ServerHandler!
+    private var clientChannel: Channel!
+
+    private var message: ByteBuffer!
+    private var isDonePromise: EventLoopPromise<Void>!
+
+    final class ServerHandler: ChannelInboundHandler {
+        public typealias InboundIn = ByteBuffer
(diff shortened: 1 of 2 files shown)


Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+    }
+
+    final class StreamDecoder: ByteToMessageDecoder {
+        public typealias InboundIn = ByteBuffer
+        public typealias InboundOut = ByteBuffer
+
+        public func decode(context: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+            if buffer.readableBytes >= MemoryLayout<UInt16>.size {
```
Lukasa, line 44: The idiomatic way to spell this would be to use conditional binding:

```swift
if let messageSize = buffer.getInteger(at: buffer.readerIndex, as: UInt16.self) {
```

Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -0,0 +1,129 @@
...
+                self.serverHandler = ServerHandler()
+                connectionEstablished.succeed()
+                return channel.pipeline.addHandler(self.serverHandler)
+            }
+            .bind(host: "127.0.0.1", port: 0)
+            .wait()
+
+        self.clientChannel = try ClientBootstrap(group: group)
+            .channelInitializer { channel in
+                channel.pipeline.addHandler(ByteToMessageHandler(StreamDecoder())).flatMap { v in
```
Lukasa, line 97:
```suggestion
                channel.pipeline.addHandler(ByteToMessageHandler(StreamDecoder())).flatMap { _ in
```

Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -0,0 +1,129 @@
...
+                    channel.pipeline.addHandler(ClientHandler(self))
+                }
+            }
+            .connect(to: serverChannel.localAddress!)
+            .wait()
+
+        var message = self.serverChannel.allocator.buffer(capacity: self.messageSize)
+        message.writeInteger(UInt16(messageSize), as:UInt16.self)
+        for idx in 0..<(self.messageSize - MemoryLayout<UInt16>.stride) {
+            message.writeInteger(UInt8(idx%(Int(UInt8.max)+1)), endianness:.little, as:UInt8.self)
```
Lukasa, line 107:
```suggestion
            message.writeInteger(UInt8(truncatingIfNeeded: idx), as: UInt8.self)
```

Lukasa (commented):
A few general nits, but otherwise this makes sense!

hassila:
Maybe more iterations than 100k? In general it'd be better to have a minimum runtime for benchmarks imho to avoid significant fluctuations. I'd expect the runtime of this to be a very small fraction of a second...

Lukasa:
We can always measure the performance number and tweak as part of this PR process.

Lukasa:

Lukasa:

Lukasa:
773ms is long enough of a performance run to be suitable in this case I think.

Can you add the appropriate license header to your new file?

ser-0xff:
Done.

Lukasa:

Lukasa (approved):

Lukasa:
Thanks for this!

PR #2369, opened 2023-02-17
Building swift-nio with Swift 5.7 for iOS using Xcode 14.0 at 2.48.0.

carolinacass:
Motivation:
swift- nio was failing builds that should pass

Modifications:
Adding available to the necessary sections

### DIFF
--- a/Sources/NIOCrashTester/CrashTestSuites.swift
+++ b/Sources/NIOCrashTester/CrashTestSuites.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if !os(iOS) && !os(tvOS) && !os(watchOS)
 let crashTestSuites: [String: Any] = [
     "EventLoopCrashTests": EventLoopCrashTests(),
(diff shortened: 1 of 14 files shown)


Tests/NIOPosixTests/EventLoopTest+XCTest.swift
```diff
@@ -22,6 +22,7 @@ import XCTest
 /// Do NOT edit this file directly as it will be regenerated automatically when needed.
 ///
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
Lukasa, line 25: Were these two changes made by hand-editing?

carolinacass: yes

Lukasa: Ok, we can't make manual changes to these files (see the comment above). We'll either need to change the tool, or lift the availability guard from the type and move them into the method bodies using runtime checks.

Lukasa (approved):

PR #2370, opened 2023-02-17
OnLoopSendable: Sendable containers if on EventLoop

weissi:
### Motivation:

In SwiftNIO programs -- at least today -- it's often required to get a non-`Sendable` value from one place to another where the compiler can't prove that origin & destination are actually the same EventLoop.

### Modifications:

- Provide `NIOOnLoopSendable` which can be sent but checks the `EventLoop` on `init`, `get` and `set` of the value.
- Provide `NIOOnLoopSendableBox` which is a mutable reference type which checks the `EventLoop` on `init`, `get` and `set`. Additionally it allows an un-EL-checked `init()` iff we're initialising it with `nil`.

### Result:

Easier interop between future-based NIO programs and `Sendable` checks in Swift Concurrency programs.


Sources/NIOCore/NIOOnLoopSendable.swift
```diff
@@ -0,0 +1,78 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+public struct NIOOnLoopSendable<Value>: @unchecked Sendable {
```
Lukasa, line 15: Can you please add documentation comments to these that explains what they're for, how to use them, and how they differ?

Sources/NIOCrashTester/CrashTests+OnLoopSendable.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import NIOCore
+import NIOPosix
+
```
Lukasa, line 16: Can you comment these tests out on iOS/watchOS/tvOS please?

Sources/NIOCore/NIOOnLoopSendable.swift
```diff
@@ -0,0 +1,78 @@
...
+            return self._value
+        }
+        set {
+            self._eventLoop.preconditionInEventLoop()
+            self._value = newValue
+        }
+    }
+}
+
+public class NIOOnLoopSendableBox<Value>: @unchecked Sendable {
```
dnadoba, line 42:
```suggestion
final public class NIOOnLoopSendableBox<Value>: @unchecked Sendable {
```

Sources/NIOCore/NIOOnLoopSendable.swift
Lukasa, line 15: I wonder if instead of `NIOOnLoopSendable` we should say `NIOLoopBoundValue`. How does that sound?

Sources/NIOCore/NIOOnLoopSendable.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    public init<NonOptionalValue>(
+        valueType: NonOptionalValue.Type = NonOptionalValue.self,
+        eventLoop: EventLoop
+    ) where Optional<NonOptionalValue> == Value {
+        // We are deliberately _not_ checking if we're on `EventLoop` here because we're only putting in the `.none`
+        // value of `Optional`.
+        self._eventLoop = eventLoop
+        self._value = .none
+    }
```
dnadoba, line 64: So this can be used to initialise a variable from one thread but read/write it from another thread without any memory barrier. This sound as if the compiler might be allowed to reorder this. cc: @Lukasa 

In addition, we need to make it clear that we initialise `value` to nil. Preferable in the symbol name or at least in the docs.

Sources/NIOCore/NIOOnLoopSendable.swift
weissi: Yes, I will once we settle on a name

weissi: I do like LoopBound(Value), probably even more without "Value", because `NIOLoopBound<ChannelHandler>` sounds like actual English :)

weissi: The counterpart `NIOLoopBoundBox<ChannelHandler>` isn't ideal but we could opt for `NIOLoopBoundBoxed<ChannelHandler>`

Lukasa: Yeah, so I proposed `NIOLoopBoundValue` in part because it clarifies the (potential) value semantics of this type, while `NIOLoopBoundBox` is always referential. Could also be `NIOLoopBoundReference<>`. But I'm happy enough to drop `Value`.

Sources/NIOCore/NIOOnLoopSendable.swift
Lukasa: There's no reordering risk here: you cannot publish a reference to `self` until it is fully initialized, so we cannot publish `self` until such time as this object is properly constructed.

weissi: @dnadoba This isn't quite correct. The value is fully initialised before `init` returns. This reference isn't and cannot be published before `init` returns and is guaranteed to be local to just one thread up until after `init` has returned. And crucially, it is never read or written anymore by anybody except for the EL thread.

To then ever read or write it, you need to get it onto _the_ EventLoop (unless you're already there) which requires a memory barrier because EventLoops have to guarantee total ordering.

Sources/NIOCore/NIOOnLoopSendable.swift
weissi: I'd be in favour of dropping because that's kinda the default (like Array/Optional/... which are equally potentially fully value typed). I'm ambivalent about if we use Box or Reference. I went for Box because of NIOLockedValueBox.

Sources/NIOCore/NIOOnLoopSendable.swift
dnadoba: Thanks for explaining! I think it would be great to explain this also in the body of the function.

dnadoba:
I'm wondering if we should add `unavailable` inits `where Value: Sendable` to guide users not to use them if they already have `Sendable` values.

Sources/NIOCore/NIOOnLoopSendable.swift
```diff
@@ -0,0 +1,78 @@
...
+public class NIOOnLoopSendableBox<Value>: @unchecked Sendable {
+    @usableFromInline
+    let _eventLoop: EventLoop
+
+    @usableFromInline
+    var _value: Value
+
+    @inlinable
+    public init(_ value: Value, eventLoop: EventLoop) {
+        eventLoop.preconditionInEventLoop()
```
dnadoba, line 51: Isn't this precondition overly restrictive? Why only allow `Optinal`s to be initialised to `.none` without being on the given `eventLoop` but nothing else?

Lukasa: Because the `none` value is definitely safe to send, as it doesn't contain the non-Sendable value.

weissi:
> I'm wondering if we should add `unavailable` inits `where Value: Sendable` to guide users not to use them if they already have `Sendable` values.

I like the idea but on the other hand I think that would be trying to parent the user a little bit too hard. Especially in internal code there might be some struct that happens to be sendable right now but you know shouldn't be. I wouldn't want the compilation to fail for that.

Nothing bad happens if you do put a Sendable value in it, yeah, it's pointless but there are many pointless things that we don't make unavailable.

Sources/NIOCore/NIOOnLoopSendable.swift
weissi: I'll expand the comment.

Sources/NIOCore/NIOOnLoopSendable.swift
weissi: @Lukasa renamed to `NIOBoundValue(Box)`, lmkwyt

weissi: done

Sources/NIOCrashTester/CrashTests+OnLoopSendable.swift
weissi: @Lukasa the other crash tests aren't, also they are only in `NIOCrashTester` which is a binary which isn't compiled on those platforms, right>?

Lukasa: See #2369.

weissi: oh my, let's merge @carolinacass 's PR and I'll fix my merge conflicts with it.

weissi: @Lukasa actually I copied the merge-conflicting code so maybe an auto-merge will fix it. I think this is done.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,120 @@
...
+        eventLoop.preconditionInEventLoop()
+        self._eventLoop = eventLoop
+        self._value = value
+    }
+
+    /// Initialise a ``NIOLoopBoundBox`` with `nil`, this does _not_ require you to be running on `eventLoop`.
+    public init<NonOptionalValue>(
+        valueType: NonOptionalValue.Type = NonOptionalValue.self,
+        eventLoop: EventLoop
+    ) where Optional<NonOptionalValue> == Value {
```
glbrntt, line 93: Usability nit: I'm not sure `NIOLoopBound(valueType: Foo.self, eventLoop: loop)` is not immediately obvious that it's for an optional value.

The other small usability concern is that you can also do `NIOLoopBoundBox<Foo?>(nil, eventLoop: loop)` off of the event loop and still crash when `NIOLoopBound(valueType: Foo.self, eventLoop: loop)` wouldn't.

I wonder if instead of this `init` we should overload the other init:

```swift
public init<NonOptionalType>(_ value: NonOptionalType?, eventLoop: EventLoop) where Optional<NonOptionalType> == Value {
    if value == nil { 
        // all good 
    } else { 
        // check on EL.
    }
}
```

We may have to mark the other `init` as `dis

weissi: @glbrntt What about we spell the `nil` `init` one as a `static func makeEmpty()`?

I think it's confusing to have some `init`s okay and others not if you're off the EL. And the other issue with `init`s is that we can't easily add extra words like "empty" or something.

My new suggestion:

- all `init`s (there'll be just one) require to be called _on_ the EL
- there's a static factory that can be called off the EL which initialises an empty one.

WDYT? Also CC @Lukasa & @dnadoba

weissi: implementation suggestion here: https://github.com/apple/swift-nio/pull/2370/commits/c58431796c5b481df61368ab579cd43c4dd8ea55

weissi: (also added to this PR as a separate commit)

glbrntt (approved):

weissi:
@Lukasa / @dnadoba WDYT? Happy with it as is? For this PR ,I'll wait until you two also give me a green tick just because it defines new API that's not immediately obious.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// ``NIOLoopBound`` is an always-``Sendable``, value-typed container allowing you access to ``Value`` if and only if
+/// on the right EventLoop``.
+///
+/// ``NIOLoopBound`` is useful to transport a value of a non-``Sendable`` type that needs to go from one place in
+/// your code to another where you (but not the compiler) know is on one and the same ``EventLoop``. Usually this
+/// involves `@Sendable` closures. This type is safe because it verfies (using `eventLoop.preconditionInEventLoop()`)
```
Lukasa, line 20:
```suggestion
/// involves `@Sendable` closures. This type is safe because it verifies (using `eventLoop.preconditionInEventLoop()`)
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+        }
+        set {
+            self._eventLoop.preconditionInEventLoop()
+            self._value = newValue
+        }
+    }
+}
+
+/// ``NIOLoopBoundBox`` is an always-``Sendable``, reference-typed container allowing you access to ``Value`` if and
+/// only if on the right EventLoop``.
```
Lukasa, line 58:
```suggestion
/// only if you are accessing it on the right EventLoop``.
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// ``NIOLoopBound`` is an always-``Sendable``, value-typed container allowing you access to ``Value`` if and only if
+/// on the right EventLoop``.
```
Lukasa, line 16:
```suggestion
/// you are accessing it on the right EventLoop``.
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+}
+
+/// ``NIOLoopBoundBox`` is an always-``Sendable``, reference-typed container allowing you access to ``Value`` if and
+/// only if on the right EventLoop``.
+///
+/// ``NIOLoopBoundBox`` is useful to transport a value of a non-``Sendable`` type that needs to go from one place in
+/// your code to another where you (but not the compiler) know is on one and the same ``EventLoop``. Usually this
+/// involves `@Sendable` closures. This type is safe because it verfies (using `eventLoop.preconditionInEventLoop()`)
```
Lukasa, line 62:
```suggestion
/// involves `@Sendable` closures. This type is safe because it verifies (using `eventLoop.preconditionInEventLoop()`)
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+/// that this is actually true.
+///
+/// A ``NIOLoopBoundBox`` can only be read from or written to when you a provably
```
Lukasa, line 65:
```suggestion
/// A ``NIOLoopBoundBox`` can only be read from or written to when you are provably
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+/// (through `eventLoop.preconditionInEventLoop()`) on the ``EventLoop`` associated with the ``NIOLoopBound``. Accessing
```
Lukasa, line 66:
```suggestion
/// (through `eventLoop.preconditionInEventLoop()`) on the ``EventLoop`` associated with the ``NIOLoopBoundBox``. Accessing
```

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -0,0 +1,125 @@
...
+/// or constructing it from any other place will crash your program with a precondition as it would be undefined
+/// behaviour to do so.
+///
+/// If constructing a ``NIOLoopBoundBox`` with a `value`, it is also required the program to already be on `eventLoop`
```
Lukasa, line 70:
```suggestion
/// If constructing a ``NIOLoopBoundBox`` with a `value`, it is also required for the program to already be on `eventLoop`
```

weissi:
thanks @Lukasa applied all your suggestions

Lukasa (approved):

Sources/NIOCore/NIOLoopBound.swift
dnadoba: "Empty" isn't something I would associated with an Optional but rather with a Collection. Maybe we call it `makeNone()` or even just `.none`?
(We also have `ExpressibleByNilLiteral` which would allow us to use the `nil` literal [but the documentation discouraged its usage](https://developer.apple.com/documentation/swift/expressiblebynilliteral) and we couldn't specify the event loop which we must in our case.)

weissi: `makeNone` sounds kinda weird to me. `makeBoxWithNil`? Don't think `nil` literal would work because we need the EL

PR #2372, opened 2023-02-20
Add support for UDP_SEGMENT

glbrntt:
Motivation:

On Linux, the UDP_SEGMENT socket option allows for large buffers to be written to the kernel and segmented by the kernel (or in some cases the NIC) into smaller datagrams. This can substantially decrease the number of syscalls.

This can be set on a per message basis on a per socket basis. This change adds per socket configuration.

Modifications:

- Add a CNIOLinux function to check whether UDP_SEGMENT is supported on that particular Linux.
- Add a helper to `System` to check whether UDP_SEGMENT is supported on the current platform.
- On Linux only:
  - add the sol_udp socket option level
  - add the udp_gso socket option
- Add the `DatagramGenericSegmentationOffload` channel option.
- Get/Set the option in `DatagramChannel`

Results:

UDP GSO is supported on Linux.

### DIFF
--- a/Sources/CNIOLinux/shim.c
+++ b/Sources/CNIOLinux/shim.c
@@ -150,4 +150,21 @@ size_t CNIOLinux_CMSG_SPACE(size_t payloadSizeBytes) {
 
 const int CNIOLinux_SO_TIMESTAMP = SO_TIMESTAMP;
 const int CNIOLinux_SO_RCVTIMEO = SO_RCVTIMEO;
+
+int CNIOLinux_supports_udp_segment() {
+    #ifndef UDP_SEGMENT
+    return -1;
+    #else
+    int fd = socket(AF_INET, SOCK_DGRAM, 0);
+    if (fd == -1) {
+        return -1;
+    }
+
+    int gso_size = 512;
+    int rc = setsockopt(fd, IPPROTO_UDP, UDP_SEGMENT, &gso_size, sizeof(gso_size));
+    close(fd);
+    return rc;
+    #endif
+}
+
 #endif
--- a/Sources/NIOCore/BSDSocketAPI.swift
+++ b/Sources/NIOCore/BSDSocketAPI.swift
@@ -245,6 +245,15 @@ extension NIOBSDSocket.OptionLevel {
     /// Socket options that apply to all sockets.
     public static let socket: NIOBSDSocket.OptionLevel =
             NIOBSDSocket.OptionLevel(rawValue: SOL_SOCKET)
+
+    /// Socket options that apply only to UDP sockets.
+    #if os(Linux) || os(Android)
+    public static let udp: NIOBSDSocket.OptionLevel =
+            NIOBSDSocket.OptionLevel(rawValue: CInt(IPPROTO_UDP))
+    #else
+    public static let udp: NIOBSDSocket.OptionLevel =
+            NIOBSDSocket.OptionLevel(rawValue: IPPROTO_UDP)
+    #endif
 }
 
 // IPv4 Options
@@ -325,6 +334,15 @@ extension NIOBSDSocket.Option {
     /// Get information about the TCP connection.
     public static let tcp_connection_info: NIOBSDSocket.Option =
             NIOBSDSocket.Option(rawValue: TCP_CONNECTION_INFO)
+}
+#endif
+
+#if os(Linux)
+extension NIOBSDSocket.Option {
+    // Note: UDP_SEGMENT is not available on all Linux platforms so the value is hardcoded.
+
+    /// Use UDP segmentation offload (UDP_SEGMENT, or 'GSO'). Only available on Linux.
+    public static let udp_segment = NIOBSDSocket.Option(rawValue: 103)
 }
 #endif
 
--- a/Sources/NIOCore/ChannelOption.swift
+++ b/Sources/NIOCore/ChannelOption.swift
@@ -191,6 +191,19 @@ extension ChannelOptions {
             public init() { }
         }
 
+        /// ``DatagramSegmentSize`` controls the 'UDP_SEGMENT' socket option (sometimes reffered to as 'GSO') which allows for
+        /// large writes to be sent via `sendmsg` and `sendmmsg` and segmented into separate datagrams by the kernel (or in some cases, the NIC).
+        /// The size of segments the large write is split into is controlled by the value of this option (note that writes do not need to be a
+        /// multiple of this option).
+        ///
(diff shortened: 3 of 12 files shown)


FranzBusch (approved):
LGTM!

Sources/CNIOLinux/shim.c
```diff
@@ -150,4 +150,21 @@ size_t CNIOLinux_CMSG_SPACE(size_t payloadSizeBytes) {
 
 const int CNIOLinux_SO_TIMESTAMP = SO_TIMESTAMP;
 const int CNIOLinux_SO_RCVTIMEO = SO_RCVTIMEO;
+
+int CNIOLinux_supports_udp_segment() {
+    #ifndef UDP_SEGMENT
```
Lukasa, line 155: This guard isn't enough: we need to compile out the function below if `UDP_SEGMENT` isn't defined, or the compiler will still fail.

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -245,6 +245,12 @@ extension NIOBSDSocket.OptionLevel {
     /// Socket options that apply to all sockets.
     public static let socket: NIOBSDSocket.OptionLevel =
             NIOBSDSocket.OptionLevel(rawValue: SOL_SOCKET)
+
+    #if os(Linux)
+    /// Socket options that apply only to UDP sockets.
+    public static let sol_udp: NIOBSDSocket.OptionLevel =
```
Lukasa, line 251: Per the naming scheme, I think this is just `udp`. Also, prefer `IPPROTO_UDP` to `SOL_UDP`. That way we can also make this available on Darwin and Windows.

Sources/NIOCore/ChannelOption.swift
```diff
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if os(Linux)
+import CNIOLinux
```
Lukasa, line 16: Is this import needed?

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -552,6 +563,19 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
                 // Receiving packet info is only supported for IP
                 throw ChannelError.operationUnsupported
             }
+        case _ as ChannelOptions.Types.DatagramGenericSegmentationOffload:
+            if System.supportsUDPSegmentationOffload {
+                #if os(Linux)
```
Lukasa, line 568: Instead of putting the compile time checks in here, can we put them into functions in `BSDSocketAPI` with different implementations in the various placse? I'd like us to avoid too much multiplication of `#if os` throughout this file, so if we can split these branches into functions that exist on all platforms but that are only implemented non-trivially on the platforms we care about, that would be very helpful.

Sources/CNIOLinux/shim.c
Lukasa, line 155: Also, why are we compile guarding for the absence of this flag here, where we rely on it being present in `ChannelOption.swift`?

Sources/NIOCore/BSDSocketAPI.swift
glbrntt: Will change this. What's the difference between `IPPROTO_UDP` and `SOL_UDP` all of the references and docs I found referred to the latter.

Sources/NIOCore/ChannelOption.swift
glbrntt: Nope, I previously had the is-supported check was on `DatagramGenericSegmentationOffload` but remember we have `System` which seemed a better fit.

Sources/CNIOLinux/shim.c
Lukasa: Yes, the preprocessor doesn't skip the rest of the code, and the compiler doesn't stop parsing after `return`. [Godbolt](https://godbolt.org/z/dxvjKvTrc).

Lukasa: Yes, Linux has a stable ABI so the value of this flag is hardcoded and cannot change. We've done this with other socket options.

Sources/NIOCore/BSDSocketAPI.swift
Lukasa: In terms of its value, literally nothing: both evaluate to 17. `SOL_UDP` is kinda a bastardisation: the `SOL_TCP` and `SOL_UDP` are both aliases to `IPPROTO_TCP` and `IPPROTO_UDP`, but the latter two spellings are more common.

Sources/NIOPosix/BSDSocketAPIPosix.swift
```diff
@@ -267,4 +267,35 @@ extension NIOBSDSocketControlMessage {
     }
 }
 
+extension NIOBSDSocket {
+    static func setUDPSegment(socket: NIOBSDSocket.Handle, segmentSize: CInt) throws {
```
glbrntt, line 271: Is something like this what you had in mind @Lukasa?

Sources/NIOCore/ChannelOption.swift
```diff
@@ -191,6 +191,19 @@ extension ChannelOptions {
             public init() { }
         }
 
+        /// ``DatagramGenericSegmentationOffload`` controls the 'UDP_SEGMENT' socket option (sometimes reffered to as 'GSO') which allows for
+        /// large writes to be sent via `sendmsg` and `sendmmsg` and segmented into separate datagrams by the kernel (or in some cases, the NIC).
+        /// The size of segments the large write is split into is controlled by the value of this options (note that writes do not need to be a
```
Lukasa, line 196:
```suggestion
        /// The size of segments the large write is split into is controlled by the value of this option (note that writes do not need to be a
```

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -508,7 +508,14 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
                 // Receiving packet info is only supported for IP
                 throw ChannelError.operationUnsupported
             }
-            break
+        case _ as ChannelOptions.Types.DatagramGenericSegmentationOffload:
+            guard System.supportsUDPSegmentationOffload else {
+                throw ChannelError.operationUnsupported
+            }
+            let segmentSize = value as! ChannelOptions.Types.DatagramGenericSegmentationOffload.Value
+            try self.socket.withUnsafeHandle {
```
Lukasa, line 516: Can we make this a `func` on the socket type itself? That way we can avoid having the unsafe code in this location, which would be nice.

Lukasa (approved):
Nice one, LGTM!

PR #2373, opened 2023-02-20
Don't retain a task when all we want is a time

Lukasa:
Motivation:

To know when we next need to wake up, we keep track of what the next deadline will be. This works great, but in order to keep track of this UInt64 we save off an entire ScheduledTask. This object is quite wide (6 pointers wide), and two of those pointers require ARC traffic, so doing this saving produces unnecessary overhead.

Worse, saving this task plays poorly with task cancellation. If the saved task is cancelled, this has the effect of "retaining" that task until the next event loop tick. This is unlikely to produce catastrophic bugs in real programs, where the loop does tick, but it violates our tests which rigorously assume that we will always drop a task when it is cancelled. In specific manufactured cases it's possible to produce leaks of non-trivial duration.

Modifications:

- Wrote a weirdly complex test.
- Moved the implementation of Task.readyIn to a method on NIODeadline
- Saved a NIODeadline instead of a ScheduledTask

Result:

Minor performance improvement in the core event loop processing, minor correctness improvement.

### DIFF
--- a/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
+++ b/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
@@ -378,21 +378,14 @@ internal struct ScheduledTask {
     let task: () -> Void
     private let failFn: (Error) ->()
     @usableFromInline
-    internal let _readyTime: NIODeadline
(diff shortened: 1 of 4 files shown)


Lukasa:

PR #2375, opened 2023-02-21
Not Holding OnToRunClosure() test updates

carolinacass:
Motivation:
testCancelledScheduledTasksDoNotHoldOnToRunClosure() was not allowed enough time and timing off at moments, causing it to fail occasionally

Modifications:
Added a ConditionLock throughout the code to make sure it only unlocks when the code has waited enough time for us not to hit the precondition failure


Tests/NIOEmbeddedTests/AsyncTestingEventLoopTests.swift
```diff
@@ -471,22 +471,35 @@ final class NIOAsyncTestingEventLoopTests: XCTestCase {
...
+            let scheduled = make(deallocated: deallocated)
             scheduled.cancel()
-            assert(weakThing == nil, within: .seconds(1))
-            await XCTAssertThrowsError(try await scheduled.futureResult.get()) { error in
+            if deallocated.lock(whenValue: 1, timeoutSeconds: 60) {
+                deallocated.unlock()
+            } else {
+                XCTFail("Timed out waiting for lock")
+            }
+            XCTAssertThrowsError(try scheduled.futureResult.wait()) { error in
```
Lukasa, line 502: Prefer the original `await` construction we used here.

Tests/NIOPosixTests/EventLoopTest+XCTest.swift
```diff
@@ -22,6 +22,7 @@ import XCTest
 /// Do NOT edit this file directly as it will be regenerated automatically when needed.
 ///
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
```
Lukasa, line 25: We can't add these: where did these come from? I think we resolved this in your prior PR.

Lukasa (changes requested):

Tests/NIOEmbeddedTests/AsyncTestingEventLoopTests.swift
```diff
@@ -471,21 +471,34 @@ final class NIOAsyncTestingEventLoopTests: XCTestCase {
                 XCTAssertNoThrow(try eventLoop.syncShutdownGracefully())
             }
 
-            class Thing {}
+            class Thing {
```
Lukasa, line 474:
```suggestion
            class Thing: @unchecked Sendable {
```

Tests/NIOEmbeddedTests/EmbeddedEventLoopTest.swift
```diff
@@ -319,21 +320,34 @@ public final class EmbeddedEventLoopTest: XCTestCase {
             XCTAssertNoThrow(try eventLoop.syncShutdownGracefully())
         }
 
-        class Thing {}
+        class Thing {
```
Lukasa, line 323:
```suggestion
        class Thing: @unchecked Sendable {
```

Tests/NIOPosixTests/EventLoopTest.swift
```diff
@@ -810,21 +811,34 @@ public final class EventLoopTest : XCTestCase {
             XCTAssertNoThrow(try group.syncShutdownGracefully())
         }
 
-        class Thing {}
+        class Thing {
```
Lukasa, line 814:
```suggestion
        class Thing: @unchecked Sendable {
```

Lukasa (approved):

PR #2376, opened 2023-02-27
Fix memory binding.

ser-0xff:
Fix memory binding.

### Motivation:

According to the documentation the 'capacity' argument represent the amount of memory counted as instances of T. Number of bytes allocated for the memory block does not always the same value as a number of instances.

### Modifications:

bindMemory(:, capacity:) called with capacity equal to number of items in the vector instead of size of vector in bytes.

### Result:

Proper memory binding.


Lukasa (approved):

Lukasa:

Lukasa:
This LGTM but needs to be rebased onto main.

Lukasa:
This still needs to be rebased onto main, and sadly I can't do it as you haven't made the branch editable by admins.

ser-0xff:
I synced the branch with swift-nio/main and asked to give you rights resync it.

Lukasa:

Lukasa (approved):

PR #1963, opened 2021-09-21
Fix HttpPart description comment typo

wplong11:
Fixes a typo `mssage` -> `message`

### Motivation:

Typos are bad.

### Modifications:

Simple correction to the description comment for HttpPart

### Result:

Correct typo

PR #2377, opened 2023-02-28
Fix an upcoming compiler warning on implicit raw pointer casts.

atrick:
Fix an upcoming compiler warning on implicit raw pointer casts.

## Motivation:

setOption forms a raw pointer to a generic argument. The compiler will warn on this as of:

[proposal] Constrain implicit raw pointer conversion... #1963 https://github.com/apple/swift-evolution/pull/1963

/Sources/NIOPosix/BaseSocket.swift:286:31: warning: forming 'UnsafeRawPointer' to a variable of type 'T'; this is likely incorrect because 'T' may contain an object reference.

                option_value: &val,
                              ^

## Modifications:

Ideally, this would be fixed by adding a BitwiseCopyable constraint to the 'value' parameter of 'BaseSocker.setOption'. That would not only eliminate the warning, but would make the API safer. But BitwiseCopyable isn't quite ready for public use.

In the meantime, a reasonable workaround is to use an explicit `withUnsafeBytes` to form the raw pointer.

## Result:

This silences the new compiler warning.


Lukasa:

Lukasa (approved):
Thanks for the fix Andy!

atrick:
@Lukasa do you know of any process for getting fixes into the swift-nio tag that the swift toolchain itself builds? It's only at 2.31.2, which dates back to 2021.

Lukasa:
I think @yim-lee is the keeper of the knowledge.

yim-lee:
@atrick Do you mean [update-checkout](https://github.com/apple/swift/blob/main/utils/update_checkout/update-checkout-config.json)? That will need to be updated by owner of the project(s) that depend on `swift-nio`. I don't work on a Swift toolchain related project with the dependency though. @shahmishal @tomerd Do you know?

shahmishal:
> That will need to be updated by owner of the project(s) that depend on swift-nio. 
Correct, @atrick any specific reason to update the tag?

atrick:
@shahmishal This PR has a trivial fix for a scary warning that the compiler will produce. I'm guessing people don't want to see that every time they build the swift toolchain. AFAICT swift-docc and swiftpm depend on swift-nio.

yim-lee:
> AFAICT swift-docc and swiftpm depend on swift-nio.

SwiftPM doesn't depend on `swift-nio` directly. I don't think any of its dependencies require it either.

PR #2378, opened 2023-02-28
Buffer pool for message headers and addresses.

ser-0xff:
Add a pool of msg_hdr and sockaddr_storage buffers.

### Motivation:

With synchronous calls it was enough to have a single buffer and use it for all calls. Asynchronous operations require to keep data while operation will not be completed, so they can be reused only when operation will be completed.

### Modifications:

Add a pool with sentinel to check buffer out of bound access.

### Result:

Have a possibility to use asynchronous operations in the future.


Sources/NIOPosix/Pool.swift
```diff
@@ -202,3 +202,117 @@ extension Int {
         self = (self + alignmentGuide) & (~alignmentGuide)
     }
 }
+
+final class PooledMsgBuffer: PoolElement {
```
Lukasa, line 206: I think we want this to be a struct, not a class.

Sources/NIOPosix/Pool.swift
```diff
@@ -202,3 +202,117 @@ extension Int {
...
+
+    deinit {
+        self.controlMessageStorage.deallocate()
+    }
+
+    func evictedFromPool() {
+        self.validateSentinel()
+    }
+
+    func withUnsafePointers<ReturnValue>(
```
Lukasa, line 308: Can you add the `withStorageManagement` version of this function? We'll need it for the eventual io_uring case.

Sources/NIOPosix/Pool.swift
```diff
@@ -202,3 +202,117 @@ extension Int {
...
+    }
+
+    private func validateSentinel() {
+        self.storage.withUnsafeMutableTypedPointers { _, _, sentinelPointer in
+            precondition(sentinelPointer.pointee == Self.sentinelValue, "Detected memory handling error!")
+        }
+    }
+
+    private var storage: BackingStorage
+    public var controlMessageStorage: UnsafeControlMessageStorage
```
Lukasa, line 289: I think this thing is why this is currently a class. Presumably we're going to have to pool this too, so my gut instinct is that we should omit the change to this right now, and have a separate change to clean up the lifecycle management and move this to a pool.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -440,10 +437,7 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
     init(socket: Socket, parent: Channel? = nil, eventLoop: SelectableEventLoop) throws {
         self.vectorReadManager = nil
         try socket.setNonBlocking()
-        self.pendingWrites = PendingDatagramWritesManager(bufferPool: eventLoop.bufferPool,
-                                                          msgs: eventLoop.msgs,
-                                                          addresses: eventLoop.addresses,
-                                                          controlMessageStorage: eventLoop.controlMessageStorage)
+        self.pendingWrites = PendingDatagramWritesManager(bufferPool: eventLoop.bufferPool, msgBufferPool: eventLoop.msgBufferPool)
```
Lukasa, line 440: Can we save the `msgBufferPool` on `self` here? That way we can avoid reaching into the EL to grab it later on, which is a nice win.

Lukasa (commented):
Broadly looks great! I've left a few notes in the diff.

Sources/NIOPosix/Pool.swift
ser-0xff: Needed to deallocate ControlMessageStorage

Sources/NIOPosix/Pool.swift
ser-0xff: Do you mean remove the controlMessageStorage from the PooledMsgBuffer and move it back to the EL for now?

Lukasa: Yes.

Sources/NIOPosix/Pool.swift
ser-0xff: Done.

Sources/NIOPosix/Pool.swift
ser-0xff: Done.

Sources/NIOPosix/Pool.swift
ser-0xff: Moved controlMessageStorage back to EL.

Sources/NIOPosix/SocketChannel.swift
ser-0xff: That change was needed to refer controlMessageStorage from the PooledMsgBuffer. Now controlMessageStorage moved back to the EL, so probably that change not needed any more. What do you think?

Lukasa (approved):
Nice one, this LGTM.

Lukasa:
Can you update this to in-line with current `main`?

Lukasa:

ser-0xff:
Synced with main

Lukasa:
Unfortunately this has fallen out of date again, can you give me permissions on this branch so I can bring it up-to-date and merge it?

PR #2379, opened 2023-03-01
Remove redundant availability guards

glbrntt:
Motivation:

EventLoopTest has an availability annotation for the various Darwin platforms. This availability check is repeated in a few tests which yields warnings.

Modifications:

Remove the additional test-level availability guards.

Result:

Warning free.


Lukasa (changes requested):
I think I'd rather we dropped the one on the class than we dropped these, as I believe these ones are gating the test on iOS. Can you sync up with @carolinacass to validate that this change still passes on iOS?

PR #2381, opened 2023-03-01
mark syncShutdownGracefully noasync

rnro:
Motivation:

The code as-is blocks the calling thread which shouldn't be done in an `async` content.

Modifications:

Mark `EventLoopGroup.syncShutdownGracefully()` and `NIOThreadPool.syncShutdownGracefully()` noasync on Swift > 5.7


Lukasa (approved):

Lukasa:
Can you run `generate_linux_tests.rb`?

PR #2385, opened 2023-03-06
Add support for UDP_GRO

glbrntt:
Motivation:

Support was added for UDP_SEGMENT in #2372 which allows for large UDP datagrams to be written to a socket by letting the kernel or NIC segment the data across multiple datagrams. This reduces traversals across the network stack which can lead to performance improvements. UDP_GRO is the receive-side counterpart allowing the kernel/NIC to aggregate datagrams and reduce network stack traversals.

Modifications:

- Add a function in CNIOLinux to check whether UDP_GRO is supported
- Add the relevant socket and channel options
- Add tests

Result:

- UDP_GRO can be enabled where supported and applications may receive large buffers.

### DIFF
--- a/Sources/CNIOLinux/include/CNIOLinux.h
+++ b/Sources/CNIOLinux/include/CNIOLinux.h
@@ -22,6 +22,7 @@
 #include <sys/sysinfo.h>
 #include <sys/socket.h>
 #include <sched.h>
+#include <stdbool.h>
 #include <errno.h>
 #include <pthread.h>
 #include <netinet/ip.h>
@@ -107,7 +108,7 @@ size_t CNIOLinux_CMSG_SPACE(size_t);
 extern const int CNIOLinux_SO_TIMESTAMP;
 extern const int CNIOLinux_SO_RCVTIMEO;
 
-int CNIOLinux_supports_udp_segment();
-int CNIOLinux_supports_udp_gro();
+bool CNIOLinux_supports_udp_segment();
+bool CNIOLinux_supports_udp_gro();
 #endif
 #endif
--- a/Sources/CNIOLinux/include/CNIOLinux.h
+++ b/Sources/CNIOLinux/include/CNIOLinux.h
@@ -108,5 +108,6 @@ extern const int CNIOLinux_SO_TIMESTAMP;
 extern const int CNIOLinux_SO_RCVTIMEO;
 
 int CNIOLinux_supports_udp_segment();
+int CNIOLinux_supports_udp_gro();
 #endif
 #endif
--- a/Sources/NIOPosix/BSDSocketAPIPosix.swift
+++ b/Sources/NIOPosix/BSDSocketAPIPosix.swift
@@ -297,5 +297,35 @@ extension NIOBSDSocket {
         throw ChannelError.operationUnsupported
         #endif
     }
+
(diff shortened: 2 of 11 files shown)


Sources/CNIOLinux/include/CNIOLinux.h
```diff
@@ -108,5 +108,6 @@ extern const int CNIOLinux_SO_TIMESTAMP;
 extern const int CNIOLinux_SO_RCVTIMEO;
 
 int CNIOLinux_supports_udp_segment();
+int CNIOLinux_supports_udp_gro();
```
Lukasa, line 111: I only just realised these two functions have a really weird return value convention. Can we `#include <stdbool.h>` and then have these return `bool`? It'll make the rest of the code much clearer.

Sources/NIOPosix/BSDSocketAPIPosix.swift
```diff
@@ -297,5 +297,35 @@ extension NIOBSDSocket {
         throw ChannelError.operationUnsupported
         #endif
     }
+
+    static func setUDPReceiveOffload(_ enabled: Bool, socket: NIOBSDSocket.Handle) throws {
+        #if os(Linux)
+        var isEnabled: CInt = enabled ? 1 : 0
```
Lukasa, line 303: Does this need to be `var`?

Lukasa (commented):
Mostly looks great, a few tiny notes!

Sources/NIOPosix/BSDSocketAPIPosix.swift
glbrntt: I think so: `option_value` in `setsockopt` takes an `UnsafeRawPointer`

Lukasa (approved):

buttaface:
These tests are failing for me on Ubuntu 20.04 x86_64 and [the CI is not running them](https://ci.swiftserver.group/job/swift-nio2-swift57-prb/855/consoleText), which may be why that was missed:
```
DatagramChannelTests.testChannelCanReceiveLargeBufferWithGROUsingScalarReads : Test skipped: required false value but got true - UDP_SEGMENT (GSO) is not supported on this platform
```
This is the error I see when they _are_ run locally:
```
/home/butta/swift-nio/Tests/NIOPosixTests/DatagramChannelTests.swift:1378: error: DatagramChannelTests.testChannelCanReceiveLargeBufferWithGROUsingScalarReads : XCTAssertEqual failed: ("1000") is not equal to ("10000") -Test Case 'DatagramChannelTests.testChannelCanReceiveLargeBufferWithGROUsingScalarReads' failed (0.007 seconds)
```

Lukasa:
Thanks for this. We’re tracking this on our side as well, as our internal CI platform has also flagged this regression.

Lukasa:
Also as a breadcrumb, can you provide the output of `uname -a`?

buttaface:
```
Linux Butta-server 5.4.0-132-generic #148-Ubuntu SMP Mon Oct 17 16:02:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

Lukasa:
10/10 for the hostname.

glbrntt:
We also see failures on 5.4.

FWIW these tests do pass with a newer kernel (5.15).

PR #2387, opened 2023-03-09
Allow UDP GRO tests to fail in some circumstances

glbrntt:
Motivation:

The recently added UDP GRO tests fail on some older Linux Kernel versions. We believe that UDP GRO support on loopback was limited in early versions so we should tolerate those failures.

However, we've verified that on 5.15 and newer that GRO is supported so we should not tolerate failure in those cases.

Modifications:

- Add shims to CNIOLinux to get system info via uname
- Verify GRO works before running GRO tests and if it doesn't then validate the kernel version isn't greater than 5.15.

Results:

Less flaky tests.

### DIFF
--- a/Tests/NIOPosixTests/DatagramChannelTests.swift
+++ b/Tests/NIOPosixTests/DatagramChannelTests.swift
@@ -16,6 +16,9 @@ import NIOConcurrencyHelpers
 import NIOCore
 @testable import NIOPosix
 import XCTest
+#if os(Linux)
+import CNIOLinux
+#endif
 
 extension Channel {
     func waitForDatagrams(count: Int) throws -> [AddressedEnvelope<ByteBuffer>] {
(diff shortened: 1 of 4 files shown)


Tests/NIOPosixTests/DatagramChannelTests.swift
```diff
@@ -1409,4 +1413,136 @@ class DatagramChannelTests: XCTestCase {
...
+        //
+        // It does however make testing GRO somewhat challenging. We need to know when we can assert
+        // that datagrams will be aggregated. To do this we run a simple check on loopback (as we
+        // use this for all other UDP_GRO tests) and check whether datagrams are aggregated on the
+        // receive side. If they aren't then we we don't bother with further testing and instead
+        // validate that our kernel is older than 5.15.
+        try XCTSkipUnless(System.supportsUDPSegmentationOffload, "UDP_SEGMENT (GSO) is not supported on this platform")
+        try XCTSkipUnless(System.supportsUDPReceiveOffload, "UDP_GRO is not supported on this platform")
+        let sender = try! self.buildChannel(group: self.group)
+        let receiver = try! self.buildChannel(group: self.group)
```
Lukasa, line 1436: Presumably we don't want to leak these channels?

Tests/NIOPosixTests/DatagramChannelTests.swift
```diff
@@ -1409,4 +1413,136 @@ class DatagramChannelTests: XCTestCase {
...
+                } else {
+                    self.major = nil
+                    self.minor = nil
+                }
+            }
+        }
+
+        init(utsname info: utsname) {
+            self.machine = withUnsafePointer(to: info.machine) { pointer in
+                pointer.withMemoryRebound(to: UInt8.self, capacity: MemoryLayout.size(ofValue: pointer)) {
```
Lukasa, line 1516: This doesn't look like this can possibly be right: the size of value seems absolutely wrong. Can I suggest we use `withUnsafeBytesOf` instead?

Lukasa (approved):

Lukasa:
5.8 and nightly failures are a result of https://forums.swift.org/t/update-to-swift-package-init-templates/63145/9.

PR #2388, opened 2023-03-10
Initial PoC of Custom Actor Executor support

Lukasa:
This PR tracks an example of custom actor executor support, as well as updating our AsyncAwait example to clearly annotate (via `precondition[Not]OnEventLoop`) whether the code in question is executing on the event loop or not.

The initial conclusion here is that this demonstrates the strengths _and_ limitations of custom actor executors for the NIO model. Actors running on an event loop executor clearly run _on_ the event loop, such that the NIO fast-path successfully fires. However, free functions and non-actor types do not inherit this execution context, which limits the utility of our existing async sequences (which are not actors and cannot be EL bound).

### DIFF
--- a/Sources/NIOCore/ActorExecutor.swift
+++ b/Sources/NIOCore/ActorExecutor.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if swift(>=5.9)
+#if compiler(>=5.9)
 /// A helper protocol that can be mixed in to a NIO ``EventLoop`` to provide an
 /// automatic conformance to `SerialExecutor`.
 ///
--- a/Sources/NIOCore/ActorExecutor.swift
+++ b/Sources/NIOCore/ActorExecutor.swift
@@ -1,0 +1,82 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.9)
+/// A helper protocol that can be mixed in to a NIO ``EventLoop`` to provide an
+/// automatic conformance to `SerialExecutor`.
+///
+/// Implementers of `EventLoop` should consider conforming to this protocol as
+/// well on Swift 5.9 and later.
+@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+public protocol NIOSerialEventLoopExecutor: EventLoop, SerialExecutor { }
+
+@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+extension NIOSerialEventLoopExecutor {
+    @inlinable
+    public func enqueue(_ job: consuming ExecutorJob) {
+        let unownedJob = UnownedJob(job)
+        self.execute {
+            unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
+        }
+    }
+
+    @inlinable
+    public func asUnownedSerialExecutor() -> UnownedSerialExecutor {
+        UnownedSerialExecutor(ordinary: self)
+    }
+
+    @inlinable
+    public var executor: any SerialExecutor {
+        self
+    }
+}
+
+/// A type that wraps a NIO ``EventLoop`` into a `SerialExecutor`
+/// for use with Swift concurrency.
+///
+/// This type is not recommended for use because it risks problems with unowned
+/// executors. Adopters are recommended to conform their own event loop
+/// types to `SerialExecutor`.
+final class NIODefaultSerialEventLoopExecutor {
+    @usableFromInline
+    let loop: EventLoop
+
+    @inlinable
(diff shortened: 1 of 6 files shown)


FranzBusch:
Great PoC!

> However, free functions and non-actor types do not inherit this execution context, which limits the utility of our existing async sequences (which are not actors and cannot be EL bound).

Can we try out marking our `NIOAsyncSequenceProducer` methods with `@_inheritActorContext` and having an example how this works with `NIOAsyncChannel`? In theory if the actor that is consuming the `NIOAsyncSequenceProducer` is using the NIO executor we should get a fast path right?

John-Connolly:
This is super interesting I'm eager to see what you come up with. It would be ideal if swift had a way to express that everything in this closure will get executed on a specific executor unless ```Task.detached``` is called or another task with a specified actor is started.  I'm unsure how to model a ```Request -> Response``` scenario without that. The only solution i've come up with is to pass and ```isolated Actor``` as a parameter to every function I want to be run on a executor.

Sources/NIOAsyncAwaitDemo/AsyncChannelIO.swift
```diff
@@ -24,17 +24,21 @@ struct AsyncChannelIO<Request, Response> {
     }
 
     func start() async throws -> AsyncChannelIO<Request, Response> {
+        self.channel.eventLoop.preconditionNotInEventLoop()
```
ktoso, line 27: Hm yeah all those are cases where we'd want to stay on the EL... this is useful to point at as specific examples, thanks Cory!

I could propose unholy horrible workarounds, but let's not... 😅

Sources/NIOCore/ActorExecutor.swift
```diff
@@ -0,0 +1,82 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.9)
+/// A helper protocol that can be mixed in to a NIO ``EventLoop`` to provide an
+/// automatic conformance to `SerialExecutor`.
+///
+/// Implementers of `EventLoop` should consider conforming to this protocol as
+/// well on Swift 5.9 and later.
```
FranzBusch, line 20: NIT:
```suggestion
/// Implementers of ``EventLoop`` should consider conforming to this protocol as
/// well on Swift 5.9 and later.
```

Sources/NIOCore/EventLoop.swift
```diff
@@ -429,6 +440,15 @@ extension EventLoop {
     }
 }
 
+extension EventLoop {
+    #if swift(>=5.9)
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    public var executor: any SerialExecutor {
+        NIODefaultSerialEventLoopExecutor(self)
+    }
```
FranzBusch, line 448: I am wondering if we really should default to this or are better off with a `fatalError`. We are making sure our ELs implement this sensible but if somebody created their own EL and we use the default I am a bit concerned that users will just run into crashes for which we could provide a better `fatalError` message.

FranzBusch: Yeah probably it's okay since all our ELs do the right thing and there aren't many other ELs out there.

Lukasa:
Applied your nit.

FranzBusch (approved):

Sources/NIOCore/EventLoop.swift
ktoso: Seems okey to leave this IMHO 👍

ktoso (commented):
Looks good 👍 Too bad we couldn't put it onto EL right away but this is okey

ktoso (approved):

Sources/NIOCore/ActorExecutor.swift
```diff
@@ -0,0 +1,82 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if swift(>=5.9)
```
FranzBusch, line 15: Just circling back here. If we want there is a way to create a custom serial executor without warnings on 5.9 and blow like this

```
#if swift(<5.9)
final class MyExecutor: SerialExecutor {
    func enqueue(_ job: UnownedJob) {
        job._runSynchronously(on: self.asUnownedSerialExecutor())
    }

    func asUnownedSerialExecutor() -> UnownedSerialExecutor {
        UnownedSerialExecutor(ordinary: self)
    }
}
#else
@available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
final class MyExecutor: SerialExecutor {
    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
    func enqueue(_ job: __owned ExecutorJob) {
      job.runSynchronously(on: asUnownedSerialExecutor())
    }

    func asUnownedSerialExecutor() -> UnownedSerialExecutor {
        UnownedSerialExecutor(ordinary: self)
    }
}
#endif
```

Lukasa: Thanks Franz. It's a good suggestion but I'm inclined not to do it: using both `_runSynchronously` _and_ `__owned` makes me really quite nervous about the stability of that code. @weissi are you happy with having this API on 5.9 only?

weissi: I'm happy with having APIs that are available everywhere from a certain Swift version. But it should be `#if compiler` not `#if swift`

PR #2394, opened 2023-03-22
Make System.coreCount aware of cgroup2

PeterAdams-A:
Motivation:

The count already appears to be aware of cgroup1
but doesn't have logic to support cgroup2.

Modifications:

Add logic to parse /sys/fs/cgroup/cpu.max
For details see - https://stackoverflow.com/questions/65551215/get-docker-cpu-memory-limit-inside-container/65554131#65554131

Result:

Core count will adapt if constrained by a cgroup on more modern kernels.

### DIFF
--- a/Sources/NIOCore/Linux.swift
+++ b/Sources/NIOCore/Linux.swift
@@ -61,7 +61,7 @@ enum Linux {
     }
 
     /// Get the available core count according to cgroup1 restrictions.
-    /// Round up the the next whole number.
+    /// Round up to the next whole number.
     static func coreCountCgroup1Restriction(quota quotaPath: String = Linux.cfsQuotaPath,
                                             period periodPath: String = Linux.cfsPeriodPath) -> Int? {
         guard
--- a/Sources/NIOCore/Linux.swift
+++ b/Sources/NIOCore/Linux.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2023 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Sources/NIOCore/Utilities.swift
```diff
@@ -93,7 +93,9 @@ public enum System {
             .map { $0.ProcessorMask.nonzeroBitCount }
             .reduce(0, +)
 #elseif os(Linux) || os(Android)
-        if let quota = Linux.coreCount(quota: Linux.cfsQuotaPath, period: Linux.cfsPeriodPath) {
+        if let quota2 = Linux.coreCount(maxPath: Linux.cfsCpuMaxPath) {
```
FranzBusch, line 96: Should we adapt the documentation of `coreCount` to also include details of how c groups are used?

Sources/NIOCore/Linux.swift
```diff
@@ -70,5 +72,15 @@ enum Linux {
         else { return nil }
         return (quota - 1 + period) / period // always round up if fractional CPU quota requested
     }
+
+    // cgroup2
+    static func coreCount(maxPath: String) -> Int? {
```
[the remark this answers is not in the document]
PeterAdams-A: I've tried a rename, we'll see how people like the names.  Making the name unique also allows defaulting the arguments so I've done that too - let me know if you think that's a bad idea.

Sources/NIOCore/Utilities.swift
PeterAdams-A: I've added a brief bit of extra commentary.  There is scope for broader review here.  The idea of using this number to setup your threads loop group is mentioned - in most cases this is probably the wrong thing to do.

I also wonder if the logic could be improved - I think it's probably possible to be both in a cgroup and a cpuset.  I do wonder if CPUSet combined with multiple threads is a good idea or not - suspect anyone doing this has probably put in quite a lot of thought.

Lukasa (approved):
I'm happy with this

FranzBusch (approved):

Sources/NIOCore/Linux.swift
```diff
@@ -59,7 +60,10 @@ enum Linux {
         return cpuset.map(countCoreIds).reduce(0, +)
     }
 
-    static func coreCount(quota quotaPath: String,  period periodPath: String) -> Int? {
+    /// Get the available core count according to cgroup1 restrctions.
```
MaxDesiatov, line 63:
```suggestion
    /// Get the available core count according to cgroup1 restrictions.
```

Sources/NIOCore/Linux.swift
```diff
@@ -70,5 +74,16 @@ enum Linux {
         else { return nil }
         return (quota - 1 + period) / period // always round up if fractional CPU quota requested
     }
+
+    /// Get the available core count according to cgroup2 restrctions.
```
MaxDesiatov, line 78: typo nit
```suggestion
    /// Get the available core count according to cgroup2 restrictions.
```

FranzBusch (approved):

PeterAdams-A:

PR #2395, opened 2023-03-24
Mildly rework the NIOLock storage

Lukasa:
Motivation:

NIOLock uses a storage object constructed from a ManagedBuffer. In general this is fine, but it's a tricky API to use safely and we want to avoid violating any of its guarantees.

Modifications:

- Store the value in the header and the lock in the elements
- Add debug assertions on alignment

Result:

We'll be a bit more confident of the use of NIOLock


weissi (approved):
nice one, thanks!

Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -51,6 +53,8 @@ extension LockOperations {
     
     @inlinable
     static func destroy(_ mutex: UnsafeMutablePointer<LockPrimitive>) {
+    mutex.assertValidAlignment()
```
glbrntt, line 56: nit: indentation

Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -61,6 +65,8 @@ extension LockOperations {
     
     @inlinable
     static func lock(_ mutex: UnsafeMutablePointer<LockPrimitive>) {
+    mutex.assertValidAlignment()
```
glbrntt, line 68: nit: indentation

Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -71,6 +77,8 @@ extension LockOperations {
     
     @inlinable
     static func unlock(_ mutex: UnsafeMutablePointer<LockPrimitive>) {
+    mutex.assertValidAlignment()
```
glbrntt, line 80: nit: indentation

dnadoba (approved):

Issue #2398, opened 2023-04-05
`NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` should throw an error on cancelation


dnadoba:
### Expected behavior
`NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` should throw an error on cancelation

### Actual behavior
`NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` returns nil

PR #2399, opened 2023-04-05
Throw `CancellationError` if `NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` is cancelled instead of returning `nil`

dnadoba:
Fixes https://github.com/apple/swift-nio/issues/2398
### Motivation:

`NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` currently returns nil if the current `Task` is canceled. This behaviour is undesirable. For example `AsyncSequence<ByteBuffer>.collect(upTo:)` would return an incomplete buffer instead of throwing an error which will very likely result in corrupted data or parsing errors down the line.

### Modifications:

Throw `CancellationError` if task is cancelled.

### Result:

`NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` correctly signals that it was prematurely terminated through cancellation.

For the non-throwing version `NIOAsyncSequenceProducer` we should think about not returning `nil` as well and instead waiting until the source finishes the `AsyncSequence`.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
@@ -868,7 +868,7 @@ extension NIOThrowingAsyncSequenceProducer {
         enum CancelledAction {
             /// Indicates that ``NIOAsyncSequenceProducerDelegate/didTerminate()`` should be called.
             case callDidTerminate
-            /// Indicates that the continuation should be resumed with a `CancelationError` and
+            /// Indicates that the continuation should be resumed with a `CancellationError` and
             /// that ``NIOAsyncSequenceProducerDelegate/didTerminate()`` should be called.
             case resumeContinuationWithCancellationErrorAndCallDidTerminate(CheckedContinuation<Element?, Error>)
             /// Indicates that nothing should be done.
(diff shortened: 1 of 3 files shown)


Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -870,7 +870,7 @@ extension NIOThrowingAsyncSequenceProducer {
             case callDidTerminate
             /// Indicates that the continuation should be resumed with `nil` and
             /// that ``NIOAsyncSequenceProducerDelegate/didTerminate()`` should be called.
-            case resumeContinuationWithNilAndCallDidTerminate(CheckedContinuation<Element?, Error>)
+            case resumeContinuationWithCancellationErrorAndCallDidTerminate(CheckedContinuation<Element?, Error>)
```
Lukasa, line 873: Can you update the comment above?

dnadoba:

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
dnadoba: Changed in https://github.com/apple/swift-nio/pull/2399/commits/1e41b2bf459f73ba70b5d0608b8267a34009b638

Lukasa (approved):

PR #2401, opened 2023-04-11
Throw `CancellationError` instead of returning `nil` during early cancellation.

dnadoba:
### Motivation:
Follow up PR for https://github.com/apple/swift-nio/pull/2399

We currently still return `nil` if the current `Task` is canceled before the first call to `NIOThrowingAsyncSequenceProducer.AsyncIterator.next()` but it should throw `CancellationError` too.

In addition, the generic `Failure` type turns out to be a problem. Just throwing a `CancellationError` without checking that `Failure` type is `any Swift.Error` or `CancellationError` introduced a type safety violation as we throw an unrelated type.

### Modifications:

- throw `CancellationError` on eager cancellation
-  deprecates the generic `Failure` type of `NIOThrowingAsyncSequenceProducer`. It now must always be `any Swift.Error`. For backward compatibility we will still return nil if `Failure` is not `any Swift.Error` or `CancellationError`.

### Result:

`CancellationError` is now correctly thrown instead of returning `nil` on eager cancelation. Generic `Failure` type is deprecated.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
@@ -98,6 +98,7 @@ public struct NIOThrowingAsyncSequenceProducer<
     ///   - backPressureStrategy: The back-pressure strategy of the sequence.
     ///   - delegate: The delegate of the sequence
     /// - Returns: A ``NIOThrowingAsyncSequenceProducer/Source`` and a ``NIOThrowingAsyncSequenceProducer``.
+    @available(*, deprecated, message: "Support for a generic Failure type is deprecated. Failure type must be `any Swift.Error`.")
     @inlinable
     public static func makeSequence(
         elementType: Element.Type = Element.self,
@@ -105,6 +106,51 @@ public struct NIOThrowingAsyncSequenceProducer<
         backPressureStrategy: Strategy,
         delegate: Delegate
     ) -> NewSequence {
+        let sequence = Self(
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let source = Source(storage: sequence._storage)
+
+        return .init(source: source, sequence: sequence)
+    }
+    
+    /// Initializes a new ``NIOThrowingAsyncSequenceProducer`` and a ``NIOThrowingAsyncSequenceProducer/Source``.
+    ///
+    /// - Important: This method returns a struct containing a ``NIOThrowingAsyncSequenceProducer/Source`` and
+    /// a ``NIOThrowingAsyncSequenceProducer``. The source MUST be held by the caller and
+    /// used to signal new elements or finish. The sequence MUST be passed to the actual consumer and MUST NOT be held by the
+    /// caller. This is due to the fact that deiniting the sequence is used as part of a trigger to terminate the underlying source.
+    ///
+    /// - Parameters:
+    ///   - elementType: The element type of the sequence.
+    ///   - failureType: The failure type of the sequence. Must be `Swift.Error`
+    ///   - backPressureStrategy: The back-pressure strategy of the sequence.
+    ///   - delegate: The delegate of the sequence
+    /// - Returns: A ``NIOThrowingAsyncSequenceProducer/Source`` and a ``NIOThrowingAsyncSequenceProducer``.
+    @inlinable
+    public static func makeSequence(
+        elementType: Element.Type = Element.self,
+        failureType: Failure.Type = Error.self,
+        backPressureStrategy: Strategy,
+        delegate: Delegate
+    ) -> NewSequence where Failure == Error {
+        let sequence = Self(
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let source = Source(storage: sequence._storage)
+
+        return .init(source: source, sequence: sequence)
+    }
+
+    /// only used internally by``NIOAsyncSequenceProducer`` to reused most of the code
+    @inlinable
+    internal static func makeNonThrowingSequence(
+        elementType: Element.Type = Element.self,
+        backPressureStrategy: Strategy,
+        delegate: Delegate
+    ) -> NewSequence where Failure == Never {
         let sequence = Self(
             backPressureStrategy: backPressureStrategy,
             delegate: delegate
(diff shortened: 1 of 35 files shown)


Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -499,7 +545,20 @@ extension NIOThrowingAsyncSequenceProducer {
...
 
                     case .resumeContinuationWithCancellationErrorAndCallDidTerminate(let continuation):
-                        continuation.resume(throwing: CancellationError())
+                        // We have deprecated the generic Failure type in the public API and Failure should
+                        // now be `Swift.Error`. However, if users have not migrated to the new API the could
+                        // still use a custom generic Error type and this cast might fail.
+                        // In addition, we use `NIOThrowingAsyncSequenceProducer` in the implementation of the
+                        // non-throwing variant `NIOAsyncSequenceProducer` where `Failure` will be `Never` and
+                        // this cast will fail as well.
+                        // Everything is marked @inlinable and the Failure type is know at compile time,
```
glbrntt, line 554:
```suggestion
                        // Everything is marked @inlinable and the Failure type is known at compile time,
```

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -880,9 +939,26 @@ extension NIOThrowingAsyncSequenceProducer {
             switch self._state {
             case .initial(_, let iteratorInitialized):
                 // This can happen if the `Task` that calls `next()` is already cancelled.
-                self._state = .finished(iteratorInitialized: iteratorInitialized)
-
-                return .callDidTerminate
+                
+                // We have deprecated the generic Failure type in the public API and Failure should
+                // now be `Swift.Error`. However, if users have not migrated to the new API the could
```
glbrntt, line 944:
```suggestion
                // now be `Swift.Error`. However, if users have not migrated to the new API they could
```

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -880,9 +939,26 @@ extension NIOThrowingAsyncSequenceProducer {
...
+                // still use a custom generic Error type and this cast might fail.
+                // In addition, we use `NIOThrowingAsyncSequenceProducer` in the implementation of the
+                // non-throwing variant `NIOAsyncSequenceProducer` where `Failure` will be `Never` and
+                // this cast will fail as well.
+                // Everything is marked @inlinable and the Failure type is know at compile time,
```
glbrntt, line 949:
```suggestion
                // Everything is marked @inlinable and the Failure type is known at compile time,
```

glbrntt (approved):

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -113,6 +114,51 @@ public struct NIOThrowingAsyncSequenceProducer<
...
+        let sequence = Self(
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let source = Source(storage: sequence._storage)
+
+        return .init(source: source, sequence: sequence)
+    }
+
+    /// only used internally by``NIOAsyncSequenceProducer`` to reused most of the code
```
Lukasa, line 147:
```suggestion
    /// only used internally by``NIOAsyncSequenceProducer`` to reuse most of the code
```

Lukasa (changes requested):
We clearly have no tests for the typed error case, as we haven't had to deprecate anything. Can you add tests for that case that confirm we appropriately do _not_ throw `CancellationError` in them, thereby testing the new function?

dnadoba:
@Lukasa tests added in https://github.com/apple/swift-nio/pull/2401/commits/28e4c70d31e1c69c24c8a5424a40f264fde94635

Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
```diff
@@ -463,6 +463,32 @@ final class NIOThrowingAsyncSequenceProducerTests: XCTestCase {
...
+        // sure our task runs when the demand is registered
+        let backPressureStrategy = MockNIOElementStreamBackPressureStrategy()
+        let delegate = MockNIOBackPressuredStreamSourceDelegate()
+        let new = NIOThrowingAsyncSequenceProducer.makeSequence(
+            elementType: Int.self,
+            failureType: CustomError.self,
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let sequence = new.sequence
```
Lukasa, line 480: Can we make sure to use `withExtendedLifetime` to preserve the producer side? That'll help me be confident that this code doesn't have this effect because the producer is being dropped.

Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
```diff
@@ -515,9 +541,37 @@ final class NIOThrowingAsyncSequenceProducerTests: XCTestCase {
...
+        struct CustomError: Error {}
+        let backPressureStrategy = MockNIOElementStreamBackPressureStrategy()
+        let delegate = MockNIOBackPressuredStreamSourceDelegate()
+        let new = NIOThrowingAsyncSequenceProducer.makeSequence(
+            elementType: Int.self,
+            failureType: CustomError.self,
+            backPressureStrategy: backPressureStrategy,
+            delegate: delegate
+        )
+        let sequence = new.sequence
```
Lukasa, line 562: Same note about `withExtendedLifetime` here.

Lukasa (changes requested):

Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
dnadoba: Added in https://github.com/apple/swift-nio/pull/2401/commits/c7863f6f89974138bc44d50a6a7d5932c283ea3e
We can't wrap the whole thing in `withExtendedLifetime` because we call an async method and the closure `withExtendedLifetime` takes isn't async. (another use case for reasync)

Lukasa: Thanks. I'm happy with the patch as written. More broadly, if you ever want to say "I need a value to live to this point in the code", the idiom has ended up becoming to just write `withExtendedLifetime(theValue) { }`. Your version works too, so I'm happy with this as-is, but just an FYI for the future that you could have simply written that at the bottom of the file.

dnadoba: Ah, nice. Thank you!

PR #2402, opened 2023-04-12
Handle reentrant reads in ALPNHandler

FranzBusch:
# Motivation
I spotted a bug in the ALPNHandler where it doesn't properly unbuffer reentrant reads. This can lead to dropped reads.

# Modification
Instead of buffering into an array we are now buffering into a Deque and unbuffer as long as there are reads in the Deque.

# Result
No more dropped reads.

### DIFF
--- a/Sources/NIOTLS/ApplicationProtocolNegotiationHandler.swift
+++ b/Sources/NIOTLS/ApplicationProtocolNegotiationHandler.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import NIOCore
+import DequeModule
 
 /// The result of an ALPN negotiation.
 ///
@@ -62,7 +63,7 @@ public final class ApplicationProtocolNegotiationHandler: ChannelInboundHandler,
 
     private let completionHandler: (ALPNResult, Channel) -> EventLoopFuture<Void>
     private var waitingForUser: Bool
-    private var eventBuffer: [NIOAny]
+    private var eventBuffer: Deque<NIOAny>
 
     /// Create an `ApplicationProtocolNegotiationHandler` with the given completion
     /// callback.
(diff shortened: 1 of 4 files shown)


Sources/NIOTLS/ApplicationProtocolNegotiationHandler.swift
```diff
@@ -72,7 +73,7 @@ public final class ApplicationProtocolNegotiationHandler: ChannelInboundHandler,
     public init(alpnCompleteHandler: @escaping (ALPNResult, Channel) -> EventLoopFuture<Void>) {
         self.completionHandler = alpnCompleteHandler
         self.waitingForUser = false
-        self.eventBuffer = []
+        self.eventBuffer = .init()
```
glbrntt, line 76: Does `Deque` not conform to `ExpressibleByArrayLiteral`?

Tests/NIOTLSTests/ApplicationProtocolNegotiationHandlerTests.swift
```diff
@@ -222,4 +243,37 @@ class ApplicationProtocolNegotiationHandlerTests: XCTestCase {
 
         XCTAssertTrue(try channel.finish().isClean)
     }
+
+    func testUnbufferingHandlesReentrantReads() throws {
+        let channel = EmbeddedChannel()
+        let loop = channel.eventLoop as! EmbeddedEventLoop
```
glbrntt, line 249: This is just `let loop = channel.embeddedEventLoop` but looks like you never use the fact it's an embedded loop anyways:

```suggestion
        let loop = channel.eventLoop
```

glbrntt (approved):

Tests/NIOTLSTests/ApplicationProtocolNegotiationHandlerTests.swift
FranzBusch: Yeah this was just a copy and paste from the old tests here. I fixed this up

PR #2403, opened 2023-04-12
Add `AsyncChannel` based `ServerBootstrap.bind()` methods

FranzBusch:
# Motivation
In my previous PR, we added a new async bridge from a NIO `Channel` to Swift Concurrency primitives in the from of the `NIOAsyncChannel`. This type alone is already helpful in bridging `Channel`s to Concurrency; however, it is hard to use since it requires to wrap the `Channel` at the right time otherwise we will drop reads. Furthermore, in the case of protocol negotiation this becomes even trickier since we need to wait until it finishes and then wrap the `Channel`.

# Modification
This PR introduces a few things:
1. New methods on the `ServerBootstrap` which allow the creation of `NIOAsyncChannel` based channels. This can be used in all cases where no protocol negotiation is involved.
2. A new protocol and type called `NIOProtocolNegotiationHandler` and `NIOProtocolNegotiationResult` which is used to identify channel handlers that are doing protocol negotiation.
3. New methods on the `ServerBootstrap` that are aware of protocol negotiation.

# Result
We can now easily and safely create new `AsyncChannel`s from the `ServerBootstrap`


Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -67,10 +67,59 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
+    init(
+        eventLoop: EventLoop,
+        closeRatchet: CloseRatchet,
+        protocolNegotiationClosure: @escaping (InboundIn) -> EventLoopFuture<ProducerElement>
+    ) where InboundIn == Channel {
         self.eventLoop = eventLoop
         self.closeRatchet = closeRatchet
+        self.transformation = .protocolNegotiation { channel in
+            return protocolNegotiationClosure(channel)
+                .flatMapErrorThrowing { error in
```
glbrntt, line 105: I think we need to hop to the channel event loop here as the `protocolNegotiationClosure` could return us a future on any loop (and `pipeline.fireErrorCaught` asserts we're on the right EL).

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -86,12 +135,50 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
-        self.buffer.append(self.unwrapInboundIn(data))
+        let unwrapped = self.unwrapInboundIn(data)
+
+        switch self.transformation {
+        case .sync(let transformation):
+            do {
+                try self.buffer.append(transformation(unwrapped))
+            } catch {
+                context.close(promise: nil)
+                context.fireErrorCaught(error)
```
glbrntt, line 146: We usually fire the error before closing

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -86,12 +135,50 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
+                return
+            }
+        case .protocolNegotiation(let protocolNegotiation):
+            let unsafeSelf = UnsafeTransfer(self)
+            let unsafeContext = UnsafeTransfer(context)
+            protocolNegotiation(unwrapped)
+                .whenComplete { result in
```
glbrntt, line 153: Can you add a comment why the unsafe transfers are okay here? At the moment it relies on the negotiation result being on the correct EL.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,484 @@ public final class ServerBootstrap {
...
+    ///   - host: The host to bind on.
+    ///   - port: The port to bind on.
+    ///   - serverBackpressureStrategy: The back pressure strategy used by the server socket channel.
+    ///   - childBackpressureStrategy: The back pressure strategy used by the child channels.
+    ///   - isChildChannelOutboundHalfClosureEnabled: Indicates if half closure is enabled on the child channels. If half closure is enabled
+    ///   then finishing the ``NIOAsyncChannelWriter`` will lead to half closure.
+    /// - Returns: A ``NIOAsyncChannel`` of connection ``NIOAsyncChannel``s.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func bind<ChildChannelInbound: Sendable, ChildChannelOutbound: Sendable>(
```
glbrntt, line 499: Does this (and other `bind`s) need type hints?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,484 @@ public final class ServerBootstrap {
...
+                    return register(eventLoop, serverChannel).map { asyncChannel }
+                } catch {
+                    return eventLoop.makeFailedFuture(error)
+                }
+            }.flatMapError { error in
+                serverChannel.close0(error: error, mode: .all, promise: nil)
+                return eventLoop.makeFailedFuture(error)
+            }
+        }.flatMap {
+            $0
```
glbrntt, line 715: I think if you use `eventLoop.flatSubmit` you can drop this

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,484 @@ public final class ServerBootstrap {
...
+    ///   - host: The host to bind on.
+    ///   - port: The port to bind on.
+    ///   - protocolNegotiationHandlerType: The protocol negotiation handler type that is awaited on. A handler of this type
+    ///   must be added to the pipeline in the child channel initializer.
+    ///   - serverBackpressureStrategy: The back pressure strategy used by the server socket channel.
+    /// - Returns: A ``NIOAsyncChannel`` of  the protocol negotiation results. It is expected that the protocol negotiation handler
+    /// is going to wrap the child channels into ``NIOAsyncChannel`` which are returned as part of the negotiation result.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func bindWithProtocolNegotiation<Handler: NIOProtocolNegotiationHandler>(
```
glbrntt, line 735: Your other `bind`s for protocol negotiation don't have the `WithProtocolNegotiation` -- I think dropping it is fine since you're require to specify the `protocolNegotiationHandlerType`.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,484 @@ public final class ServerBootstrap {
...
+                    // the user to wrap the child channels themselves when the negotiation is done
+                    // and return the wrapped async channel as part of the negotiation result.
+                    let asyncChannel = try NIOAsyncChannel<Handler.NegotiationResult, Never>(
+                        synchronouslyWrapping: serverChannel,
+                        backpressureStrategy: serverBackpressureStrategy,
+                        protocolNegotiationClosure: { (channel: Channel) in
+                            channel.eventLoop
+                                .makeCompletedFuture {
+                                    try channel.pipeline.syncOperations.handler(type: protocolNegotiationHandlerType)
+                                }
```
glbrntt, line 924: Isn't this just `channel.pipelline.handler(type: protocolNegotiationHandlerType)`?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,484 @@ public final class ServerBootstrap {
...
+                                .flatMap { handler -> EventLoopFuture<NIOProtocolNegotiationResult< Handler.NegotiationResult>> in
```
glbrntt, line 925:
```suggestion
                                .flatMap { handler -> EventLoopFuture<NIOProtocolNegotiationResult<Handler.NegotiationResult>> in
```

Sources/NIOPosix/Bootstrap.swift
FranzBusch: Where would I use the `flatSubmit` here?

Sources/NIOPosix/Bootstrap.swift
FranzBusch: Thanks for catching this! This was necessary at some point when I had no way to disambiguate them.

Sources/NIOPosix/Bootstrap.swift
glbrntt: Instead of the `submit` on L699

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -491,6 +491,8 @@ extension ServerBootstrap {
     ///   - port: The port to bind on.
     ///   - serverBackpressureStrategy: The back pressure strategy used by the server socket channel.
     ///   - childBackpressureStrategy: The back pressure strategy used by the child channels.
+    ///   - childChannelInboundType: The child channel's inbound type.
+    ///   - childChannelOutboundType: The child channel's inbound type.
```
glbrntt, line 495: inbound -> outbound

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -599,6 +615,8 @@ extension ServerBootstrap {
     ///   - socket: The _Unix file descriptor_ representing the bound stream socket.
     ///   - serverBackpressureStrategy: The back pressure strategy used by the server socket channel.
     ///   - childBackpressureStrategy: The back pressure strategy used by the child channels.
+    ///   - childChannelInboundType: The child channel's inbound type.
+    ///   - childChannelOutboundType: The child channel's inbound type.
```
glbrntt, line 619: inbound -> outbound

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,501 @@ public final class ServerBootstrap {
...
+        ) {
+            try SocketAddress(unixDomainSocketPath: unixDomainSocketPath)
+        }
+    }
+
+    /// Bind the `ServerSocketChannel` to a UNIX Domain Socket.
+    /// Waiting for protocol negotiation to finish before yielding the channel to the returned ``NIOAsyncChannel``.
+    ///
+    /// - Parameters:
+    ///   - unixDomainSocketPath: The _Unix domain socket_ path to bind to. `unixDomainSocketPath` must not exist, it will be created by the system.
```
gjcairo, line 821: Same note about:
> `unixDomainSocketPath` must not exist, it will be created by the system.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +482,501 @@ public final class ServerBootstrap {
...
+                serverChannelInit(serverChannel)
+            }.flatMap {
+                do {
+                    try serverChannel.pipeline.syncOperations.addHandler(
+                        AcceptHandler(childChannelInitializer: childChannelInit, childChannelOptions: childChannelOptions),
+                        name: "AcceptHandler"
+                    )
+
+                    // In the case of protocol negotiation we cannot wrap the child channels into
+                    // `NIOAsyncChannel`s for the user since we don't know the type. We rather exepect
```
gjcairo, line 934:
```suggestion
                    // `NIOAsyncChannel`s for the user since we don't know the type. We rather expect
```

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_spi(AsyncChannel) import NIOCore
+import DequeModule
+
+
+@_spi(AsyncChannel)
+public final class NIOTypedApplicationProtocolNegotiationHandler<NegotiationResult>: ChannelInboundHandler, RemovableChannelHandler, NIOProtocolNegotiationHandler {
```
gjcairo, line 20: We should probably add some docs for this type and the public properties below.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -104,6 +104,44 @@ public final class NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Senda
...
+        self.outboundWriter.finish()
+    }
+
+    @inlinable
+    @_spi(AsyncChannel)
+    public init(
+        synchronouslyWrapping channel: Channel,
+        backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
+        isOutboundHalfClosureEnabled: Bool = false,
+        protocolNegotiationClosure: @escaping (Channel) -> EventLoopFuture<Inbound>
```
Lukasa, line 133: I'm a bit nervous about having these two methods differ only on the label of a trailing closure, which is commonly omitted.

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -67,10 +67,61 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
+    }
+
+    @inlinable
+    init(
+        eventLoop: EventLoop,
+        closeRatchet: CloseRatchet
+    ) where InboundIn == ProducerElement {
+        self.eventLoop = eventLoop
+        self.closeRatchet = closeRatchet
+        self.transformation = .sync { $0 }
```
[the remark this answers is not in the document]
FranzBusch: I tried this initially but this doesn't work because the compiler can't infer that if the `transformation` is `none` that `InboundIn == ProducerElement`. The compiler should be smart enough here though to see that the closure is a no-op in this case and be able to optimise it away.

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStream.swift
```diff
@@ -47,7 +47,71 @@ public struct NIOAsyncChannelInboundStream<Inbound: Sendable>: Sendable {
...
+        backpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark?,
+        closeRatchet: CloseRatchet,
+        protocolNegotiationClosure: @escaping (Channel) -> EventLoopFuture<Inbound>
+    ) throws {
+        channel.eventLoop.preconditionInEventLoop()
+        let handler = NIOAsyncChannelInboundStreamChannelHandler<Channel, Inbound>(
+            eventLoop: channel.eventLoop,
+            closeRatchet: closeRatchet,
+            protocolNegotiationClosure: protocolNegotiationClosure
+        )
```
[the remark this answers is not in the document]
Lukasa: Agree with Gus here, this should be able to be refactored.

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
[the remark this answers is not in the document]
Lukasa: There's no way the compiler makes that optimization, but it's ok, the identity function is going to be pretty cheap, and this is an implementation detail we can fix.

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
FranzBusch: Added comments to the type. The property is already documented in the protocol so no need to document it here as well.

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
```diff
@@ -0,0 +1,141 @@
...
+        // Importantly firing a channel read can lead to new reads being buffered due to reentrancy!
+        while let datum = self.eventBuffer.popFirst() {
+            context.fireChannelRead(datum)
+        }
+
+        context.fireChannelReadComplete()
+    }
+}
+
+#if swift(>=5.6)
```
Lukasa, line 138: We only support 5.6.

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
Lukasa, line 20: This has loads of common implementation with the existing ALPN handler. Presumably we can factor much of that out to common code.

FranzBusch: Just pushed a change to create a shared state machine between the two handlers. They still have a bit of implementation in them but this shares the bulk of the underlying logic now.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +483,448 @@ public final class ServerBootstrap {
...
+                    )
+
+                    // We are wrapping the inbound channels into `NIOAsyncChannel`s with the transformation
+                    // closure of the `NIOAsyncChannel` that allows us to wrap them without adding
+                    // wrapping/unwrapping handlers to the pipeline.
+                    let asyncChannel = try NIOAsyncChannel<NIOAsyncChannel<ChildChannelInbound, ChildChannelOutbound>, Never>(
+                        synchronouslyWrapping: serverChannel,
+                        backpressureStrategy: serverBackpressureStrategy,
+                        transformationClosure: { channel in
+                            try NIOAsyncChannel(
```
adam-fowler, line 688: It appears this is being called on the server channel eventloop, but it expects to be called child channel eventloop.

FranzBusch: Thanks. Good catch!

adam-fowler: You should run your tests with an EventLoopGroup with more than one thread or at least have one test with this. Three is a good number to ensure you don't have a server and a client interleaving EventLoop usage. I was running with two threads and only caught this because I had a test which had two client requests in it which opened two child channels

Lukasa: This is a very good piece of advice. I also recommend adding `preconditionInEventLoop` checks for any code that expects it, because they're fairly cheap and they flush out these bugs really early.

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -86,10 +149,67 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
+                    // We forward the read only after the transformation has been completed. This is super important
+                    // since we are setting up the NIOAsyncChannel handlers in the transformation and
+                    // we must make sure to only generate reads once they are setup. Reads can only
+                    // happen after the child channels hit `channelRead0` that's why we are holding the read here.
+                    context.fireChannelRead(data)
+                }
+
+        case .protocolNegotiation(let protocolNegotiation):
+            // The unsafe transfers here are required because we need to use self in whenComplete
+            // We are making sure to be on our event loop so we can safely use self in whenComplete
```
Lukasa, line 179: Same here, can we use `LoopBound`?

Tests/NIOPosixTests/AsyncChannelBootstrapTests+XCTest.swift
```diff
@@ -0,0 +1,37 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+//
+// AsyncChannelBootstrapTests+XCTest.swift
+//
+import XCTest
+
+///
+/// NOTE: This file was generated by generate_linux_tests.rb
```
Lukasa, line 20: We can remove this file, we no longer need it.

Tests/NIOTLSTests/NIOTypedApplicationProtocolNegotiationHandlerTests+XCTest.swift
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+//
+// NIOTypedApplicationProtocolNegotiationHandlerTests+XCTest.swift
+//
+import XCTest
+
+///
+/// NOTE: This file was generated by generate_linux_tests.rb
```
Lukasa, line 20: We can remove this file, we no longer need it.

Tests/NIOTLSTests/NIOTypedApplicationProtocolNegotiationHandlerTests.swift
```diff
@@ -0,0 +1,209 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2017-2023 Apple Inc. and the SwiftNIO project authors
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -482,6 +483,454 @@ public final class ServerBootstrap {
...
+            isChildChannelOutboundHalfClosureEnabled: isChildChannelOutboundHalfClosureEnabled
+        ) { (eventLoop, serverChannel) in
+            serverChannel.registerAndDoSynchronously { serverChannel in
+                serverChannel.bind(to: address)
+            }
+        }
+    }
+
+    private typealias MakeServerChannel = (_ eventLoop: SelectableEventLoop, _ childGroup: EventLoopGroup, _ enableMPTCP: Bool) throws -> ServerSocketChannel
+    private typealias Register = (EventLoop, ServerSocketChannel) -> EventLoopFuture<Void>
```
Lukasa, line 652: Do either of these need `@Sendable`?

FranzBusch: Yes `Register` needs to `Sendable` since it is run on the server socket EL. However, `MakeServerChannel` is run inline and not escaping.

FranzBusch:
> I wonder if it's a good time for us to explore (3). However, as this is SPI we can merge this without that change and investigate it later.

I would prefer to merge this now to unblock the HTTP2 work and will track this as an open item before we make the APIs public

Lukasa (approved):

PR #2408, opened 2023-04-19
Don't have channels stop reading on errors they tolerate.

Lukasa:
Motivation:

When an error is hit during a read loop, a channel is able to tolerate that error without closing. This is done for a number of reasons, but the most important one is accepting sockets for already-closed connections, which can trigger all kinds of errors on the read path.

Unfortunately, there was an edge-case in the code for handling this case. If one or more reads in the loop had succeeded before the error was caught, the inner code would be expecting a call to readIfNeeded, but the outer code wouldn't make it. This would lead to autoRead channels being wedged open.

Modifications:

This patch extends the Syscall Abstraction Layer to add support for server sockets. It adds two tests: one for the basic accept flow, and then one for the case discussed above.

This patch also refactors the code in BaseSocketChannel.readable0 to more clearly show the path through the error case. There were a number of early returns and partial conditionals that led to us checking the same condition in a number of places. This refactor makes it clearer that it is possible to exit this code in the happy path, with a tolerated error, which should be considered the same as reading _something_.

Result:

Harder to wedge a channel open.

### DIFF
--- a/Sources/NIOPosix/BaseSocketChannel.swift
+++ b/Sources/NIOPosix/BaseSocketChannel.swift
@@ -1091,36 +1091,40 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
             // peer closed / shutdown the connection.
             if let channelErr = err as? ChannelError, channelErr == ChannelError.eof {
                 readStreamState = .eof
-                // Directly call getOption0 as we are already on the EventLoop and so not need to create an extra future.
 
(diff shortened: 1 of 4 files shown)


Lukasa:
Flipping this to draft while I diagnose some weirdness I saw locally.

Lukasa:
Nevermind, this is working fine.

dnadoba (approved):
Great catch! LGTM

FranzBusch (approved):

PR #2411, opened 2023-04-26
Extend the integration test harness to track FDs

Lukasa:
Motivation

This patch extends the NIO integration test harness to track file descriptors, in particular to search for leaks. This change has been validated on Linux and Darwin, and in both cases correctly diagnoses FD leaks.

The goal is to enable us to regression test for things like

Modifications

- Add support for hooking socket and close calls.
- Wire up this support into the test harness.
- Extend the test harness to handle the logging.
- Add new regression test for #2047.

Results

We can write regression tests for FD leaks.

### DIFF
--- a/IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-darwin.c
+++ b/IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-darwin.c
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <assert.h>
 #if __APPLE__
 
 #define _GNU_SOURCE
@@ -23,6 +24,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
+#include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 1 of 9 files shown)


IntegrationTests/allocation-counter-tests-framework/template/HookedFunctionsDoHook/Sources/HookedFunctions/src/hooked-functions-darwin.c
```diff
@@ -148,6 +150,37 @@ int replacement_posix_memalign(void **memptr, size_t alignment, size_t size) {
...
+}
+
+int replacement_accept4(int sockfd, struct sockaddr *addr, socklen_t *addrlen, int flags) {
+    // Should never be called.
+    assert(false);
+}
+
+int replacement_close(int fildes) {
+    track_closed_fd(fildes);
+    JUMP_INTO_LIBC_FUN(close, fildes);
```
dnadoba, line 181: Why do we need to use `JUMP_INTO_LIBC_FUN` for close but not for the others?

Lukasa: We use it for all of them: the rest just call into a thunk that has the macro.

The reason is that the macro expands to a complex structure that ends in `return`. This means we can only use `JUMP_INTO_LIBC_FUN` as the _last_ statement in a function. For `close` that's fine, we want to call the real `close` last, but for the others we want to call the real function first. Hence the thunk.

dnadoba: We call the thunks in `hooked-functions-unix.c` but not in this file (`hooked-functions-darwin.c`), or do we?
Also `JUMP_INTO_LIBC_FUN` in this file just calls the function returns the result.

Lukasa: Apologies, you're right. The same answer applies though: we don't want the immediate return.

dnadoba: I don't quite follow. Why can't we just write:
```suggestion
    return close(files);
```
instead?

The other functions in this file appear to call the libc function without any thunks. Am I overlooking something?

dnadoba: hm... it is only halfway consistent because we are missing the thunks for `accept`, `socket` and in theory `accept4`. We don't share the code and we can't just copy and paste it over because behaviour differs slightly and of the missing thunks. `JUMP_INTO_LIBC_FUN` is even a no-op in this file so I'm wondering if the "consistency" here just makes everything harder to understand without any clear benefit. 

But we can make it fully consistent and I think even share most of the code. However, this probably more effort than it is worth it. No hard feelings on that though after I have fully understood that this is just a no-op. Feel free to leave it as is.

dnadoba (approved):

FranzBusch (approved):
LGTM! Hope we can have all of this in Swift at some point

PR #2412, opened 2023-04-27
AsyncChannelBootstrapTests bind to 0 instead

FranzBusch:
# Motivation

The `AsyncChannelBootstrapTests` were failing when running parallel since they were all using the same port.

# Modification

This PR uses port `0` to let the system assign a port instead. It also runs the formatter on this file.

# Result

No more tests failures.


glbrntt (approved):
Thank you!

PR #2413, opened 2023-04-28
Bind to port 0 in a few more places

glbrntt:
Motivation:

Follow up to #2412 which changed the async channel bootstrap tests to
bind to port 0 instead of a fixed port. A few instances were missed,
this fixes up the remainder.

Modifications:

- Bind to port 0

Result:

Tests should run in parallel without failures.


Lukasa (approved):

dnadoba (approved):
Awesome! Love parallel testing!

PR #2414, opened 2023-04-28
Handle close(output) in the pipeline handler.

Lukasa:
Motivation:

Currently the server pipeline handler ignores close. This generally works, except in the rare case that the user calls close(mode: .output). In this instance they have signalled that they'll never write again, and they're likely expecting a final close shortly after.

However, it is possible that the pipeline handler has suspended reads at the same time. On Linux this isn't an issue, because we'll still be told about the eventual socket close. However, on Apple platforms we won't: we've masked off the reads, and we can't listen to EVFILT_EXCEPT due to some other issues. This means that on Apple platforms the server pipeline handler can accidentally wedge the Channel open and prevent it from closing.

We should take this opportunity to have the server pipeline handler be smart about close(mode: .output). What _should_ happen here is that the pipeline handler should immediately refuse to deliver further requests on the Channel. If one is in-flight, it can continue, but everything else should be dropped. This is because the server cannot possibly respond to further requests.

Modifications:

- Add new states to the server pipeline handler
- Drop buffered requests and new data after close(mode: .output)
- Add tests

Result:

Server pipeline handler behaves way better.

### DIFF
--- a/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
+++ b/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
@@ -140,9 +140,8 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
             case .requestEndPending, .requestAndResponseEndPending:
                 self = .sentCloseOutputRequestEndPending
             case .sentCloseOutput, .sentCloseOutputRequestEndPending:
-                // Weird to duplicate fail. Tolerate this in release mode but forbid it in debug.
-                assertionFailure("sent close output twice")
-                self = .sentCloseOutput
+                // Weird to duplicate fail, but we tolerate it in both cases.
+                ()
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -414,6 +447,28 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+            // If there's a read pending, we should deliver it after we forward the close on.
+            shouldRead = self.readPending
+        }
+
+        context.close(mode: mode, promise: promise)
+
+        // Double-check readPending here in case something weird happened.
+        if shouldRead && self.readPending {
+            self.readPending = false
+            context.read()
```
weissi, line 468: Mind adding a comment here that because of the state transition to `closeOutputSent()` we will _not_ actually deliver any reads to any handlers further back? We're just dropping them on the floor, that's a good idea but I first didn't get that :)

Lukasa: Sure, I've just done that.

weissi (approved):
awesome, thanks!

PR #2415, opened 2023-04-28
NIOThrowingAsyncSequenceProducer throws when cancelled

fabianfett:
NIOThrowingAsyncSequenceProducer throws when cancelled

### Motivation:

Currently `NIOThrowingAsyncSequenceProducer` currently does not throw a CancellationError, if the AsyncSequence is cancelled while the buffer is not empty.

### Modifications:

- Make `NIOThrowingAsyncSequenceProducer` throw when cancelled

### Result:

- Postgres queries that are cancelled throw and do not give the impression that they succeed.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
@@ -511,6 +511,21 @@ extension NIOThrowingAsyncSequenceProducer {
                         return nil
                     }
 
+                case .returnCancellationError:
+                    self._lock.unlock()
+                    // We have deprecated the generic Failure type in the public API and Failure should
+                    // now be `Swift.Error`. However, if users have not migrated to the new API they could
+                    // still use a custom generic Error type and this cast might fail.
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -603,6 +618,9 @@ extension NIOThrowingAsyncSequenceProducer {
                 failure: Failure?
             )
 
+            /// The state once a call to next has been cancelled. To reach this state we must cancel the source.
```
Lukasa, line 621: This comment is a bit confused: presumably we're cancelling the iterator, not the source.

fabianfett:
```
            /// The state once a call to next has been cancelled. Cancel the source when entering this state.
```
@Lukasa Is this better?

Lukasa: Yeah, LGTM.

Lukasa (approved):
Thanks!

PR #2417, opened 2023-05-03
addition of assertSuccess() and assertFailure() on EventLoopFuture

### LINKED ISSUE #2359: EventLoopFuture should have an `assertSuccess()` (and maybe `assertFailure()` methods)
Futures are in many places used in ways that make sure that a certain future will only ever succeed (and never fail). In those cases, it'd be nice to write

```
someFuture
    .assertSucceess()
    .map { value in
        [...]
    }
```

dkz2:
Added `assertSuccess`, `assertFailure`, `preconditionSuccess` and `preconditionFailure` methods to `EventLoopFuture` for safer and more explicit error handling.

### Motivation:

Enhance developer experience by providing a clear way to handle expected success or failure of futures.

Closes #2359.

### Modifications:

Implemented `assertSuccess`, `assertFailure`, `preconditionSuccess` and `preconditionFailure`  for `EventLoopFuture`.
Added corresponding test cases to verify behavior of the new methods.

### Result:

Developers will now have access to `assertSuccess`, `assertFailure`, `preconditionSuccess` and `preconditionFailure` methods on`EventLoopFuture`.

### DIFF
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -1098,80 +1098,6 @@ extension EventLoopFuture {
         }
         return CallbackList()
     }
-
-    #if swift(>=5.7)
-    /// Adds an observer callback to this `EventLoopFuture` that is called when the
-    /// `EventLoopFuture` has a success result, asserting a failure if it encounters an error.
-    ///
-    /// - parameters:
-    ///     - callback: The callback that is called with the successful result of the `EventLoopFuture`.
-    @inlinable
-    @preconcurrency
-    public func assertSuccess(_ callback: @escaping @Sendable (Value) -> Void) {
-        self._assertSuccess(callback)
-    }
-    @usableFromInline typealias AssertSuccessCallback = @Sendable (Value) -> Void
-    #else
-    /// Adds an observer callback to this `EventLoopFuture` that is called when the
-    /// `EventLoopFuture` has a success result, asserting a failure if it encounters an error.
-    ///
-    /// - parameters:
-    ///     - callback: The callback that is called with the successful result of the `EventLoopFuture`.
-    @inlinable
-    public func assertSuccess(_ callback: @escaping (Value) -> Void) {
-        self._assertSuccess(callback)
-    }
-    @usableFromInline typealias AssertSuccessCallback = (Value) -> Void
-    #endif
-
-    @inlinable
-    func _assertSuccess(_ callback: @escaping AssertSuccessCallback) {
-        self.whenComplete { result in
-            switch result {
-            case .success(let value):
-                callback(value)
-            case .failure(let error):
-                assertionFailure("Expected success, but got failure: \(error)")
-            }
-        }
-    }
-
-    #if swift(>=5.7)
-    /// Adds an observer callback to this `EventLoopFuture` that is called when the
-    /// `EventLoopFuture` has a failure result, asserting a failure if it encounters a success.
-    ///
-    /// - parameters:
-    ///     - callback: The callback that is called with the error of the `EventLoopFuture`.
-    @inlinable
-    @preconcurrency
-    public func assertFailure(_ callback: @escaping @Sendable (Error) -> Void) {
-        self._assertFailure(callback)
-    }
-    @usableFromInline typealias AssertFailureCallback = @Sendable (Error) -> Void
-    #else
-    /// Adds an observer callback to this `EventLoopFuture` that is called when the
-    /// `EventLoopFuture` has a failure result, asserting a failure if it encounters a success.
-    ///
-    /// - parameters:
-    ///     - callback: The callback that is called with the error of the `EventLoopFuture`.
-    @inlinable
-    public func assertFailure(_ callback: @escaping (Error) -> Void) {
-        self._assertFailure(callback)
-    }
-    @usableFromInline typealias AssertFailureCallback = (Error) -> Void
-    #endif
-
-    @inlinable
-    func _assertFailure(_ callback: @escaping AssertFailureCallback) {
-        self.whenComplete { result in
-            switch result {
-            case .success(let value):
-                assertionFailure("Expected failure, but got success: \(value)")
-            case .failure(let error):
-                callback(error)
-            }
-        }
-    }
 }
 
 // MARK: and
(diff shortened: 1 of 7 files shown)


Lukasa (changes requested):
Thanks for this! I am inclined to want to say that these methods do slightly too much. Having the callback as an argument muddys the water about their actual semantic. If these methods instead behave like a regular future function (where they return a new future) then we can allow them to compose with the rest of the future handling functions, which might be nicer.

@weissi I'm also curious whether you think assertion is sufficient, or whether we need precondition variants too.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1098,6 +1098,80 @@ extension EventLoopFuture {
...
     }
+
+    #if swift(>=5.7)
+    /// Adds an observer callback to this `EventLoopFuture` that is called when the
+    /// `EventLoopFuture` has a success result, asserting a failure if it encounters an error.
+    ///
+    /// - parameters:
+    ///     - callback: The callback that is called with the successful result of the `EventLoopFuture`.
+    @inlinable
+    @preconcurrency
```
dnadoba, line 1109: As these are new methods they don't need `@preconcurrency` and also no special treatment in Swift 5.6. We can therefore just have a single declaration and implementation. 

I'm with @Lukasa though and we should make these methods return an `EventLoopFuture` instead. Therefore we will drop the `@Sendable` function argument all together and AFAICT don't have to deal with `Sendable` at all.

weissi:
> Thanks for this! I am inclined to want to say that these methods do slightly too much. Having the callback as an argument muddys the water about their actual semantic. If these methods instead behave like a regular future function (where they return a new future) then we can allow them to compose with the rest of the future handling functions, which might be nicer.

Agreed!

 
> @weissi I'm also curious whether you think assertion is sufficient, or whether we need precondition variants too.

I'd think we should have both `precondition*` and `assert*` variants

Sources/NIOCore/EventLoopFuture.swift
dkz2: Okay i see, that is more for the older methods that were implemented before some language features were introduced,  I will work on addressing the needed changes. Thank you all.

dkz2:
> > Thanks for this! I am inclined to want to say that these methods do slightly too much. Having the callback as an argument muddys the water about their actual semantic. If these methods instead behave like a regular future function (where they return a new future) then we can allow them to compose with the rest of the future handling functions, which might be nicer.
> 
> Agreed!
> 
> > @weissi I'm also curious whether you think assertion is sufficient, or whether we need precondition variants too.
> 
> I'd think we should have both `precondition*` and `assert*` variants

Added `precondition*` variants, also removed all unneeded requirements.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
+    /// If the original future fails, it triggers an assertion failure, causing a runtime error during development.
+    @inlinable
+    public func assertSuccess() -> EventLoopFuture<Value> {
+        let promise = self.eventLoop.makePromise(of: Value.self)
+        self.whenComplete { result in
+            switch result {
+            case .success(let value):
+                promise.succeed(value)
+            case .failure(let error):
+                assertionFailure("Expected success, but got failure: \(error)")
```
Lukasa, line 2204: Can we pass `file` and `line` through from the line that calls `assertSuccess`? That'll make the crash more generally useful.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
+    /// Returns a new `EventLoopFuture` that asserts the original future's failure.
+    ///
+    /// If the original future succeeds, it triggers an assertion failure, causing a runtime error during development.
+    @inlinable
+    public func assertFailure() -> EventLoopFuture<Value> {
+        let promise = self.eventLoop.makePromise(of: Value.self)
+        self.whenComplete { result in
+            switch result {
+            case .success(let value):
+                assertionFailure("Expected failure, but got success: \(value)")
```
Lukasa, line 2220: Same note here on `file` and `line`.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
 
+// MARK: assertion
+
+extension EventLoopFuture {
+    /// Returns a new `EventLoopFuture` that asserts the original future's success.
+    ///
+    /// If the original future fails, it triggers an assertion failure, causing a runtime error during development.
+    @inlinable
+    public func assertSuccess() -> EventLoopFuture<Value> {
+        let promise = self.eventLoop.makePromise(of: Value.self)
```
glbrntt, line 2198: No need to create a promise in any of these functions; you can use `always` instead:

```swift
return self.always { result in 
    switch result { 
    case .success:
        ()
    case .failure(let error):
        assertionFailure("Expected success, but got failure: \(error)")    
    }
}
```

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
+            }
+        }
+        return promise.futureResult
+    }
+
+    /// Returns a new `EventLoopFuture` that preconditions the original future's success.
+    ///
+    /// If the original future fails, it triggers a precondition failure, causing a runtime error during development.
+    @inlinable
+    public func ensureSuccess() -> EventLoopFuture<Value> {
```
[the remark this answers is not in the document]
glbrntt: You can disambiguate by using `Swift.preconditionFailure(...)` to refer to the global one.

Sources/NIOCore/EventLoopFuture.swift
weissi: Indeed, but we should pass `file` and `line` to `assertionFailure` such that it shows where the user put `.assertX` and not always this line

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
 }
 
+// MARK: assertion
+
+extension EventLoopFuture {
+    /// Returns a new `EventLoopFuture` that asserts the original future's success.
+    ///
+    /// If the original future fails, it triggers an assertion failure, causing a runtime error during development.
+    @inlinable
+    public func assertSuccess() -> EventLoopFuture<Value> {
```
weissi, line 2197: here and in the other functions we should take `file: ... =  #file, line = ... #line` and pass that to the (assertion/precondition)Failure functions

Sources/NIOCore/EventLoopFuture.swift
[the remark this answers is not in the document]
dkz2: Sounds good making the changes!

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -2187,6 +2187,79 @@ extension EventLoopFuture {
...
+    /// If the original future fails, it triggers a precondition failure, causing a runtime error during development.
+    @inlinable
+    public func ensureSuccess() -> EventLoopFuture<Value> {
+        let promise = self.eventLoop.makePromise(of: Value.self)
+        self.whenComplete { result in
+            switch result {
+            case .success(let value):
+                promise.succeed(value)
+            case .failure(let error):
+                preconditionFailure("Expected success, but got failure: \(error)")
```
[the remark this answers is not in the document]
dkz2: yes to all!

Sources/NIOCore/EventLoopFuture.swift
dkz2: oh okay i see no need to create a new promise / or complete it bc `always` can take care of that for us, working on the changes. Thank you!

Sources/NIOCore/EventLoopFuture.swift
dkz2: Done to all!

Sources/NIOCore/EventLoopFuture.swift
dkz2: changes commited!

Lukasa (approved):
Nice one, this LGTM.

Lukasa:

dkz2:
Thank you all for the help and guidance @weissi @Lukasa @glbrntt @dnadoba, very much appreciated!

PR #2419, opened 2023-05-03
Add unprocessedBytes property on NIOSingleStepByteToMessageProcessor

fabianfett:
Add unprocessedBytes property on NIOSingleStepByteToMessageProcessor

### Motivation:

Some protocols require no messages to be in the incoming buffer after a certain message was received. For example if both endpoints agree on an upgrade to TLS no remaining bytes should be in the incoming buffer (as those would have been sent unencrypted or those may have been injected by a man in the middle). We should allow to ask the `NIOSingleStepByteToMessageProcessor` if there are bytes remaining in its incoming buffer.

### Modifications:

- Add unprocessedBytes property on NIOSingleStepByteToMessageProcessor

### Result:

- Safer network communication

### DIFF
--- a/Tests/NIOCoreTests/SingleStepByteToMessageDecoderTest.swift
+++ b/Tests/NIOCoreTests/SingleStepByteToMessageDecoderTest.swift
@@ -30,19 +30,6 @@ public final class NIOSingleStepByteToMessageDecoderTest: XCTestCase {
         }
     }
 
-    private final class ForeverDecoder: NIOSingleStepByteToMessageDecoder {
-        typealias InboundOut = Never
-
-        func decode(buffer: inout ByteBuffer) throws -> InboundOut? {
-            return nil
-        }
-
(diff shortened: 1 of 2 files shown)


Tests/NIOCoreTests/SingleStepByteToMessageDecoderTest.swift
```diff
@@ -30,19 +30,6 @@ public final class NIOSingleStepByteToMessageDecoderTest: XCTestCase {
...
-
-        func decode(buffer: inout ByteBuffer) throws -> InboundOut? {
-            return nil
-        }
-
-        func decodeLast(buffer: inout ByteBuffer, seenEOF: Bool) throws -> InboundOut? {
-            XCTAssertTrue(seenEOF)
-            return try self.decode(buffer: &buffer)
-        }
-    }
```
fabianfett, line 44: totally unused. This is why I removed it.

dnadoba (approved):
LGTM

Lukasa (approved):

PR #2421, opened 2023-05-07
Tolerate sending data after close(mode: .output)

Lukasa:
Motivation

We shouldn't crash on somewhat likely user error.

Modifications

Pass on writes after close(mode: .output) instead of crashing.

Result

User code is more robust to weird edge cases.

### DIFF
--- a/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
+++ b/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
@@ -111,7 +111,11 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
                 // We got a response while still receiving a request, which we have to
                 // wait for.
                 self = .requestEndPending
-            case .requestEndPending, .idle, .sentCloseOutput, .sentCloseOutputRequestEndPending:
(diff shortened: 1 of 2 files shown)


dnadoba (approved):
LGTM

FranzBusch:
In general, this looks good to me. I am just wondering if we should rather fail the write here instead of letting it pass through the rest of the pipeline until it gets failed by the channel. Is there any reason why we prefer the latter?

Lukasa:
Yeah, I wondered about that, as well as about throwing errors.

Ultimately I decided to stick close to the current behaviour. Right now the HTTPServerPipelineHandler passes writes on in essentially all situations, even when they're sequence breaking. We never suppress them, never delay them, etc. The only rule we have enforced thus far is appropriate ordering within a request/response stream, which currently crashes.

I think dropping the write is probably the wrong thing to do. If we're right about the state the connection is in, the write will fail. If we're wrong, it _won't_ fail and we should let it be delivered. Additionally, today the HTTPServerPipelineHandler only polices `.end` messages: it doesn't concern itself with `.body` or `.head`. That further means that unless we tweak that behaviour we'll only be policing `.end`, which is weirdly inconsistent.

In the end I convinced myself that this is the minimal patch that restores the prior behaviour.

PR #2429, opened 2023-05-17
`Embedded`: `getOption(ChannelOptions.allowRemoteHalfClosure)` should not `fatalError`

felixschlegel:
## Motivation

[In `swift-nio-ssl`, I am currently working on allowing half-closures](https://github.com/apple/swift-nio-ssl/pull/428/) which rely on querying the underlying channel if `ChannelOptions.Types.AllowRemoteHalfClosureOption` is enabled. As a lot of
`swift-nio-ssl`'s tests rely on `EmbeddedChannel`, which did not support this option, they failed failed due to a `fatalError`.

## Modifications

* add a `public var allowRemoteHalfClosure` to `EmbeddedChannel`
* enable setting/getting
  `ChannelOptions.Types.AllowRemoteHalfClosureOption` in
`EmbeddedChannel` (only modifies the `allowRemoteHalfClosure` variable
* add test for new behaviour

### DIFF
--- a/Sources/NIOEmbedded/AsyncTestingChannel.swift
+++ b/Sources/NIOEmbedded/AsyncTestingChannel.swift
@@ -171,7 +171,14 @@ public final class NIOAsyncTestingChannel: Channel {
     public var isActive: Bool { return channelcore.isActive }
 
     /// - see: `ChannelOptions.Types.AllowRemoteHalfClosureOption`
-    public var allowRemoteHalfClosure: Bool = false
+    public var allowRemoteHalfClosure: Bool {
+        get {
+            return channelcore.allowRemoteHalfClosure
+        }
+        set {
+            channelcore.allowRemoteHalfClosure = newValue
+        }
+    }
 
     /// - see: `Channel.closeFuture`
     public var closeFuture: EventLoopFuture<Void> { return channelcore.closePromise.futureResult }
--- a/Sources/NIOEmbedded/AsyncTestingChannel.swift
+++ b/Sources/NIOEmbedded/AsyncTestingChannel.swift
@@ -170,6 +170,9 @@ public final class NIOAsyncTestingChannel: Channel {
     /// - note: An ``NIOAsyncTestingChannel`` starts _inactive_ and can be activated, for example by calling `connect`.
     public var isActive: Bool { return channelcore.isActive }
 
+    /// - see: `ChannelOptions.Types.AllowRemoteHalfClosureOption`
+    public var allowRemoteHalfClosure: Bool = false
+
     /// - see: `Channel.closeFuture`
     public var closeFuture: EventLoopFuture<Void> { return channelcore.closePromise.futureResult }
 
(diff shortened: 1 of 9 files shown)


Sources/NIOEmbedded/Embedded.swift
```diff
@@ -788,7 +788,8 @@ public final class EmbeddedChannel: Channel {
 
     @inlinable
     internal func getOptionSync<Option: ChannelOption>(_ option: Option) -> Option.Value {
-        if option is ChannelOptions.Types.AutoReadOption {
+        if option is ChannelOptions.Types.AutoReadOption
+        || option is ChannelOptions.Types.AllowRemoteHalfClosureOption {
```
Lukasa, line 792: I know I gave you this advice, so this is really my fault, but I no longer think this behaviour is quite right. 😁 

I think we should allow the user of `EmbeddedChannel` to have a `public var` that they can set which provides the value of `AllowRemoteHalfClosureOption`. That lets tests configure whether the behaviour should be supported or not. This also lets us support `setOption` for this option.

Relatedly, the default value should be `false`.

Lukasa (changes requested):

Tests/NIOEmbeddedTests/EmbeddedChannelTest.swift
```diff
@@ -481,13 +481,22 @@ class EmbeddedChannelTest: XCTestCase {
...
+    }
+
+    func testSetGetChannelOptionAllowRemoteHalfClosueIsSupported() {
+        let channel = EmbeddedChannel()
+        let options = channel.syncOptions
+        XCTAssertNotNil(options)
+        // allowRemoteHalfClosure should be false by default
+        XCTAssertEqual(try options?.getOption(ChannelOptions.allowRemoteHalfClosure), false)
+        XCTAssertNoThrow(try options?.setOption(ChannelOptions.allowRemoteHalfClosure, value: true))
+        XCTAssertEqual(try options?.getOption(ChannelOptions.allowRemoteHalfClosure), true)
```
Lukasa, line 499: Can you extend this test to add a case for changing the value of `allowRemoteHalfClosure` on the `EmbeddedChannel` directly?

Lukasa (changes requested):
Great, this looks really good. Can I ask you to make the same enhancement to `AsyncTestingChannel` as well? That will keep the two interfaces in step.

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -170,6 +170,9 @@ public final class NIOAsyncTestingChannel: Channel {
     /// - note: An ``NIOAsyncTestingChannel`` starts _inactive_ and can be activated, for example by calling `connect`.
     public var isActive: Bool { return channelcore.isActive }
 
+    /// - see: `ChannelOptions.Types.AllowRemoteHalfClosureOption`
+    public var allowRemoteHalfClosure: Bool = false
```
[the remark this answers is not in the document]
felixschlegel: Thanks for pointing that out! Have moved `allowRemoteClosure` to `EmbeddedChannelCore` now!

Lukasa (approved):
Fantastic!

Lukasa:

Lukasa:
Can you update the allocation limits?

Lukasa (approved):

Issue #2427, opened 2023-05-16
Flaky test: NIOAsyncWriterTests.testWriterDeinitialized_whenWriterFinished


Lukasa:
Failed in #2425 on Swift 5.6: https://ci.swiftserver.group/job/swift-nio2-swift56-prb/1026/consoleFull.

```
10:11:24 Test Case 'NIOAsyncWriterTests.testWriterDeinitialized_whenWriterFinished' started at 2023-05-16 09:11:24.341
10:11:24 /code/Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift:157: error: NIOAsyncWriterTests.testWriterDeinitialized_whenWriterFinished : XCTAssertEqual failed: ("1") is not equal to ("0") - 
10:11:24 Test Case 'NIOAsyncWriterTests.testWriterDeinitialized_whenWriterFinished' failed (0.002 seconds)
```

PR #2431, opened 2023-05-22
Fix flaky test in NIOAsyncWriter

FranzBusch:
# Motivation
We had flaky test in the writer where we were relying on some timing.

# Modification
Change the test so that it is deterministic.

# Result
Fixes https://github.com/apple/swift-nio/issues/2427


Lukasa (approved):

Lukasa:
Seems like a nightly allocation regression, mind fixing it up?

Issue #2432, opened 2023-05-24
`NIOWebSocket.WebSocketMaskingKey.random()` crashes under Thread Sanitizer


gwynne:
Details of the crash itself and the conditions under which it occurs can be found in apple/swift#66099.

My suggestion on the NIO side of things is to add an additional test to the `WebSocketMaskingKeyTests` case which covers the `WebSocketMakingKey.random()` method; currently only `.random(using:)` is covered.

### Expected behavior
Everything works normally.

### Actual behavior
The first (on Linux) or second (on macOS) call to `WebSocketMaskingKey.random()` crashes when ThreadSanitizer is enabled.

### Steps to reproduce

Copied from apple/swift#66099:

1. `git clone https://github.com/apple/swift-nio.git && cd swift-nio`
2. `sed -i '' -E -e 's/\(using: &generator\)/()/' Tests/NIOWebSocketTests/WebSocketMaskingKeyTests.swift`
3. `swift test --sanitize=thread --filter=testRandomMaskingKey`

### SwiftNIO version/commit hash

Observed against `main` as of commit 27620851c43b0ae5ac1a66e2a428448e71f4e74d and tagged release `2.53.0`.

### System & version information

See apple/swift#66099 for details, in short it's on macOS and Linux with Swift versions 5.8, latest 5.9 nightly, and latest main nightly.

PR #2433, opened 2023-05-24
Add test coverage of WebSocketMaskingKey.random()

gwynne:
This solves the NIO side of #2432.

### Motivation:

There is currently no test coverage of the `WebSocketMaskingKey.random()` convenience wrapper for `WebSocketMaskingKey.random(using:)`, which aside from simply being incomplete also has the somewhat ironic effect of masking apple/swift#66099.

### Modifications:

Added a test which exercises the code.

### Result:

The code coverage is increased by a fraction of a percent and an apparently rather obscure Thread Sanitizer issue is less hidden.

### DIFF
--- a/Tests/NIOWebSocketTests/WebSocketMaskingKeyTests.swift
+++ b/Tests/NIOWebSocketTests/WebSocketMaskingKeyTests.swift
@@ -32,10 +32,9 @@ final class WebSocketMaskingKeyTests: XCTestCase {
         }, "at least 1 of 1000 random masking keys should not be all zeros")
     }
     
-    func testRandomMaskingKeyWithDefaultGenerator() {
-        let key1 = WebSocketMaskingKey.random()
-        let key2 = WebSocketMaskingKey.random()
-        
-        XCTAssertNotEqual(key1, key2)
+    func testRandomMaskingKeyIsNotAlwaysZeroWithDefaultGenerator() {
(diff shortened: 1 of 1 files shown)


Tests/NIOWebSocketTests/WebSocketMaskingKeyTests.swift
```diff
@@ -31,4 +31,11 @@ final class WebSocketMaskingKeyTests: XCTestCase {
             WebSocketMaskingKey.random(using: &generator) != [0, 0, 0, 0]
         }, "at least 1 of 1000 random masking keys should not be all zeros")
     }
+    
+    func testRandomMaskingKeyWithDefaultGenerator() {
+        let key1 = WebSocketMaskingKey.random()
+        let key2 = WebSocketMaskingKey.random()
+        
+        XCTAssertNotEqual(key1, key2)
```
Lukasa, line 39: This test has a non-zero chance of failing: a random sequence is not forced to avoid repeating values. Can you replace this with a variant of the test directly above? I'm fine with you calling `random` deterministically at least twice to produce your test case for the Swift repo.

gwynne: Done, good call 👍

Lukasa (approved):

PR #2437, opened 2023-06-01
Add NIOAsyncChannel based connect methods to ClientBootstrap

FranzBusch:
# Motivation
In my previous PR, I added new `bind` methods to `ServerBootstrap` that vend `NIOAsyncChannel` or support an async protocol negotiation. This PR focuses on adding new `connect` methods to `ClientBootstrap` which offer the same functionality.

# Modification
This PR adds new `connect` methods that either vend a `NIOAsyncChannel` or an asynchronous protocol negotiation result. To make this work I had to change the `HappyEyeballs` resolver so that it can return a generic value on resolving. Lastly, I adapted the bootstrap tests to use the new `ClientBootstrap` capabilities which now demonstrate a client/server protocol negotiation dance.

# Result
We can now bootstrap TCP clients with `NIOAsyncChannel`s

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -1502,7 +1502,7 @@ extension ClientBootstrap {
             port: port,
             eventLoop: eventLoop,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { handler in
+            postRegisterTransformation: { handler, eventLoop in
                 handler.protocolNegotiationResult.flatMap { result in
                     result.resolve(on: eventLoop)
                 }
@@ -1570,7 +1570,7 @@ extension ClientBootstrap {
             eventLoop: eventLoop,
             socket: socket,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { handler in
+            postRegisterTransformation: { handler, eventLoop in
                 handler.protocolNegotiationResult.flatMap { result in
                     result.resolve(on: eventLoop)
                 }
@@ -1589,7 +1589,7 @@ extension ClientBootstrap {
             eventLoop: eventLoop,
             protocolFamily: protocolFamily,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { handler in
+            postRegisterTransformation: { handler, eventLoop in
                 handler.protocolNegotiationResult.flatMap { result in
                     result.resolve(on: eventLoop)
                 }
@@ -1610,7 +1610,7 @@ extension ClientBootstrap {
             registration: { channel in
                 channel.registerAndDoSynchronously(body)
             },
-            postRegisterTransformation: { handler in
+            postRegisterTransformation: { handler, eventLoop in
                 handler.protocolNegotiationResult.flatMap { result in
                     result.resolve(on: channel.eventLoop)
                 }
@@ -1644,7 +1644,7 @@ extension ClientBootstrap {
             port: port,
             eventLoop: eventLoop,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { output in
+            postRegisterTransformation: { output, eventLoop in
                 eventLoop.makeSucceededFuture(output)
             }
         )
@@ -1668,7 +1668,7 @@ extension ClientBootstrap {
             eventLoop: eventLoop,
             protocolFamily: address.protocol,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { output in
+            postRegisterTransformation: { output, eventLoop in
                 eventLoop.makeSucceededFuture(output)
             }, { channel in
                 return self.connect(freshChannel: channel, address: address)
@@ -1713,7 +1713,7 @@ extension ClientBootstrap {
             eventLoop: eventLoop,
             socket: socket,
             channelInitializer: channelInitializer,
-            postRegisterTransformation: { output in
+            postRegisterTransformation: { output, eventLoop in
                 eventLoop.makeSucceededFuture(output)
             }
         )
@@ -1725,7 +1725,7 @@ extension ClientBootstrap {
         port: Int,
         eventLoop: EventLoop,
         channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<ChannelInitializerResult>,
-        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult>
+        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult, EventLoop) -> EventLoopFuture<PostRegistrationTransformationResult>
     ) async throws -> PostRegistrationTransformationResult {
         let resolver = self.resolver ?? GetaddrinfoResolver(
             loop: eventLoop,
@@ -1757,7 +1757,7 @@ extension ClientBootstrap {
         eventLoop: EventLoop,
         socket: NIOBSDSocket.Handle,
         channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<ChannelInitializerResult>,
-        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult>
+        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult, EventLoop) -> EventLoopFuture<PostRegistrationTransformationResult>
     ) async throws -> PostRegistrationTransformationResult {
         let channel = try SocketChannel(eventLoop: eventLoop as! SelectableEventLoop, socket: socket)
 
@@ -1778,7 +1778,7 @@ extension ClientBootstrap {
         eventLoop: EventLoop,
         protocolFamily: NIOBSDSocket.ProtocolFamily,
         channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<ChannelInitializerResult>,
-        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult>,
+        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult, EventLoop) -> EventLoopFuture<PostRegistrationTransformationResult>,
         _ body: @escaping (Channel) -> EventLoopFuture<Void>
     ) -> EventLoopFuture<(Channel, PostRegistrationTransformationResult)> {
         let channel: SocketChannel
@@ -1802,7 +1802,7 @@ extension ClientBootstrap {
         channel: SocketChannel,
         channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<ChannelInitializerResult>,
         registration: @escaping @Sendable (Channel) -> EventLoopFuture<Void>,
-        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult>
+        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult, EventLoop) -> EventLoopFuture<PostRegistrationTransformationResult>
     ) -> EventLoopFuture<PostRegistrationTransformationResult> {
         let channelInitializer = { channel in
             return self.channelInitializer(channel)
@@ -1834,7 +1834,7 @@ extension ClientBootstrap {
                         result
                     }
                 }.flatMap { (result: ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult> in
-                    postRegisterTransformation(result)
+                    postRegisterTransformation(result, eventLoop)
                 }.flatMapError { error in
                     eventLoop.assertInEventLoop()
                     channel.close0(error: error, mode: .all, promise: nil)
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -1344,7 +1344,7 @@ extension ClientBootstrap {
     /// - Parameters:
     ///   - host: The host to connect to.
     ///   - port: The port to connect to.
-    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
     ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
     ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
     ///   - inboundType: The channel's inbound type.
@@ -1380,7 +1380,7 @@ extension ClientBootstrap {
     ///
     /// - Parameters:
     ///   - address: The address to connect to.
-    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
     ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
     ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
     ///   - inboundType: The channel's inbound type.
@@ -1414,7 +1414,7 @@ extension ClientBootstrap {
     ///
     /// - Parameters:
     ///   - unixDomainSocketPath: The _Unix domain socket_ path to connect to.
-    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
     ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
     ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
     ///   - inboundType: The channel's inbound type.
@@ -1448,7 +1448,7 @@ extension ClientBootstrap {
     ///
     /// - Parameters:
     ///   - descriptor: The _Unix file descriptor_ representing the connected stream socket.
-    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
     ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
     ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
     ///   - inboundType: The channel's inbound type.
@@ -1628,8 +1628,8 @@ extension ClientBootstrap {
     /// - Parameters:
     ///   - host: The host to connect to.
     ///   - port: The port to connect to.
-    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
-    ///   the protocol.
+    ///   - channelInitializer: A closure to initialize the channel. The return value of this closure is returned from the `connect`
+    ///   method.
     /// - Returns: The result of the channel initializer.
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
     @_spi(AsyncChannel)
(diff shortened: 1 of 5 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,482 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    ///   - port: The port to connect to.
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
+    ///   the protocol.
+    /// - Returns: The protocol negotiation result.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func connect<Handler: NIOProtocolNegotiationHandler>(
+        host: String,
+        port: Int,
+        channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<Handler>
```
FranzBusch, line 1629: This PR uses a different approach than the `ServerBootstrap` where we passed the type of the protocol negotiation handler. I am going to check and see if we can make this approach work for `ServerBootstrap` as well

FranzBusch:
5.8/5.9 failures is an allocation test that is two allocations of (`1000_rst_connections`)

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,482 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    ///   - isOutboundHalfClosureEnabled:  Indicates if half closure is enabled on the channel. If half closure is enabled
+    ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
+    /// - Returns: A `NIOAsyncChannel` for the established connection.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func connect<ChildChannelInbound: Sendable, ChildChannelOutbound: Sendable>(
+        host: String,
+        port: Int,
+        channelInboundType: ChildChannelInbound.Type = ChildChannelInbound.self,
+        channelOutboundType: ChildChannelOutbound.Type = ChildChannelOutbound.self,
```
Lukasa, line 1374: I think these type names are wrong: there are no child channels here. Same for the other `connect` functions added in this PR.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,482 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+            return self.initializeAndRegisterNewChannel(
+                eventLoop: eventLoop,
+                protocolFamily: protocolFamily,
+                channelInitializer: channelInitializer
+            ) {
+                $0.eventLoop.makeSucceededFuture(())
+            }
+        }
+        return try await connector.resolveAndConnect().map { $0.1 }.get()
+    }
```
Lukasa, line 1654: Substantial amount of code is duplicated between this block and the block above. Can we find a way to define common functions that abstract the difference?

Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -303,10 +304,43 @@ internal class HappyEyeballsConnector {
...
+        self.state = .idle
+        self.resolutionPromise = self.loop.makePromise()
+        self.error = NIOConnectionError(host: host, port: port)
+
+        precondition(resolutionDelay.nanoseconds > 0, "Resolution delay must be greater than zero, got \(resolutionDelay).")
+        self.resolutionDelay = resolutionDelay
+
+        precondition(connectionDelay >= .milliseconds(100) && connectionDelay <= .milliseconds(2000), "Connection delay must be between 100 and 2000 ms, got \(connectionDelay)")
+        self.connectionDelay = connectionDelay
+    }
```
Lukasa, line 337: These two inits entirely duplicate their code, for one line of difference. Can we have one call the other?

Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -315,6 +349,19 @@ internal class HappyEyeballsConnector {
...
+    /// returns: An `EventLoopFuture` that fires with a connected `Channel`.
+    @inlinable
+    func resolveAndConnect() -> EventLoopFuture<Channel> where ChannelBuilderResult == Void {
+        // We dispatch ourselves onto the event loop, rather than do all the rest of our processing from outside it.
+        self.loop.execute {
+            self.timeoutTask = self.loop.scheduleTask(in: self.connectTimeout) { self.processInput(.connectTimeoutElapsed) }
+            self.processInput(.resolve)
+        }
+        return resolutionPromise.futureResult.map { $0.0 }
+    }
```
Lukasa, line 363: This function is just the same as the one above with `.map { $0.0 }` at the end, can we call that one and just chain this on?

FranzBusch:
@Lukasa Reduced the code duplication quite substantially with another set of even more generic methods

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                return try NIOAsyncChannel(
+                    synchronouslyWrapping: channel,
+                    backpressureStrategy: channelBackpressureStrategy,
+                    isOutboundHalfClosureEnabled: isOutboundHalfClosureEnabled,
+                    inboundType: channelInboundType,
+                    outboundType: channelOutboundType
+                )
+            }
+        }
+        postRegisterTransformation: { result in
```
glbrntt, line 1395:
```suggestion
        } postRegisterTransformation: { result in
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
```
glbrntt, line 1494:
```suggestion
        } postRegisterTransformation: { result in
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    private func initializeAndRegisterNewChannel<ChannelInbound: Sendable, ChannelOutbound: Sendable>(
+        eventLoop: EventLoop,
+        protocolFamily: NIOBSDSocket.ProtocolFamily,
+        channelInboundType: ChannelInbound.Type = ChannelInbound.self,
+        channelOutboundType: ChannelOutbound.Type = ChannelOutbound.self,
+        channelBackpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
+        isOutboundHalfClosureEnabled: Bool = false,
+        _ body: @escaping (Channel) -> EventLoopFuture<Void>
```
glbrntt, line 1507: What is `body` for here?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    }
+
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    private func initializeAndRegisterChannel<ChannelInbound: Sendable, ChannelOutbound: Sendable>(
+        channel: SocketChannel,
+        channelInboundType: ChannelInbound.Type = ChannelInbound.self,
+        channelOutboundType: ChannelOutbound.Type = ChannelOutbound.self,
+        channelBackpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
+        isOutboundHalfClosureEnabled: Bool = false,
+        _ body: @escaping (Channel) -> EventLoopFuture<Void>
```
glbrntt, line 1537: What is `body` for here?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    ) async throws -> Handler.NegotiationResult {
+        let eventLoop = self.group.next()
+        return try await self.connect(
+            host: host,
+            port: port,
+            eventLoop: eventLoop,
+            channelInitializer: channelInitializer,
+            postRegisterTransformation: { handler in
+                handler.protocolNegotiationResult.flatMap { result in
+                    ServerBootstrap.waitForFinalResult(result, eventLoop: eventLoop)
```
glbrntt, line 1585: Feels weird calling a `ServerBootstrap` function here; might make more sense to move the function to `NIOProtocolNegotiationResult` instead? 

```swift
extension NIOProtocolNegotiationResult {
 func resolve(on eventLoop: EventLoop) -> EventLoopFuture<NegotiationResult> { 
    // ...
  }
}
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    private func initializeAndRegisterChannel<ChannelInitializerResult, PostRegistrationTransformationResult>(
+        channel: SocketChannel,
+        channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<ChannelInitializerResult>,
+        registration: @escaping @Sendable (Channel) -> EventLoopFuture<Void>,
+        postRegisterTransformation: @escaping @Sendable (ChannelInitializerResult) -> EventLoopFuture<PostRegistrationTransformationResult>
+    ) -> EventLoopFuture<PostRegistrationTransformationResult> {
+        let channelInitializer = { channel in
+            return self.channelInitializer(channel)
+                .flatMap { channelInitializer(channel) }
```
glbrntt, line 1786: Not sure how I feel about applying two channel initializers... Presumably one is the existing "normal" initialiser and the other is passed on `connect`? I can't imagine a situation where – as a user – I'd want to apply both.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                        return channel
+                            .bind(to: bindTarget)
+                            .flatMap {
+                                channelInitializer(channel)
+                            }
+                    } else {
+                        return channelInitializer(channel)
+                    }
+                }.flatMap { (result: ChannelInitializerResult) in
+                    eventLoop.assertInEventLoop()
```
glbrntt, line 1809: Is this assertion correct? There's nothing stopping a user from returning from an initialiser on a different event loop so I think we actually need to hop back to our event loop here instead.

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -33,43 +33,101 @@ private final class LineDelimiterDecoder: ByteToMessageDecoder {
         }
         return .needMoreData
     }
+
+    func encode(data: ByteBuffer, out: inout ByteBuffer) throws {
+        var data = data
+        out.writeBuffer(&data)
```
glbrntt, line 39:
```suggestion
        out.writeImmutableBuffer(data)
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1352,483 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+    ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
+    /// - Returns: A `NIOAsyncChannel` for the established connection.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func connect<ChannelInbound: Sendable, ChannelOutbound: Sendable>(
+        host: String,
+        port: Int,
+        channelInboundType: ChannelInbound.Type = ChannelInbound.self,
+        channelOutboundType: ChannelOutbound.Type = ChannelOutbound.self,
+        channelBackpressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark? = nil,
```
glbrntt, line 1376: I'm not sure that these three need the `channel` prefix.

Sources/NIOPosix/Bootstrap.swift
FranzBusch: I agree that the use-case is weird but I don't think we can do much from an API design point. We have the builder style pattern that allows you to add a child channel initialiser and we want to have a typed one on this method. I don't think ignoring or fatalErroring on the builder set initialiser is good. What do you think?

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -33,43 +33,101 @@ private final class LineDelimiterDecoder: ByteToMessageDecoder {
...
+        self.proposedALPN = proposedALPN
+    }
+
+    func handlerAdded(context: ChannelHandlerContext) {
+        guard context.channel.isActive else {
+            return
+        }
+
+        if let proposedALPN = self.proposedALPN {
+            self.proposedALPN = nil
```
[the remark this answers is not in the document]
FranzBusch: Since the channel can already be active in handlerAdded we might have to write the "negotiate request" here but we should only write it once. This code here ensures that between handlerAdded and channelActive that we only write it out once and only once we are active.

Sources/NIOPosix/Bootstrap.swift
FranzBusch: They totally could but I carried this assert forward from the current bootstraps where we do have that assert as well. I would keep these APIs aligned in behavior. WDYT?

Lukasa: The assertion is right: `.flatMap` will coerce us back to the original EL.

Lukasa (approved):

FranzBusch:

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
 
+// MARK: Async connect methods with NIOAsyncChannel
+
+extension ClientBootstrap {
+    /// Specify the `host` and `port` to connect to for the TCP `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - host: The host to connect to.
+    ///   - port: The port to connect to.
+    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
```
glbrntt, line 1347:
```suggestion
    ///   - backpressureStrategy: The back pressure strategy used by the channel.
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                )
+            }
+        }
+    }
+
+    /// Specify the `address` to connect to for the TCP `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - address: The address to connect to.
+    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
```
glbrntt, line 1383:
```suggestion
    ///   - backpressureStrategy: The back pressure strategy used by the channel.
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                )
+            }
+        }
+    }
+
+    /// Specify the `unixDomainSocket` path to connect to for the UDS `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - unixDomainSocketPath: The _Unix domain socket_ path to connect to.
+    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
```
glbrntt, line 1417:
```suggestion
    ///   - backpressureStrategy: The back pressure strategy used by the channel.
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                )
+            }
+        }
+    }
+
+    /// Use the existing connected socket file descriptor.
+    ///
+    /// - Parameters:
+    ///   - descriptor: The _Unix file descriptor_ representing the connected stream socket.
+    ///   - channelBackpressureStrategy: The back pressure strategy used by the channel.
```
glbrntt, line 1451:
```suggestion
    ///   - backpressureStrategy: The back pressure strategy used by the channel.
```

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+// MARK: Async connect methods with arbitrary payload
+
+extension ClientBootstrap {
+    /// Specify the `host` and `port` to connect to for the TCP `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - host: The host to connect to.
+    ///   - port: The port to connect to.
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
+    ///   the protocol.
```
glbrntt, line 1632: This needs updating: it's not necessarily used for negotiation here.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+                eventLoop.makeSucceededFuture(output)
+            }
+        )
+    }
+
+    /// Specify the `address` to connect to for the TCP `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - address: The address to connect to.
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
```
glbrntt, line 1657: This needs updating: it's not necessarily used for negotiation here.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+            }, { channel in
+                return self.connect(freshChannel: channel, address: address)
+            }).get().1
+    }
+
+    /// Specify the `unixDomainSocket` path to connect to for the UDS `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - unixDomainSocketPath: The _Unix domain socket_ path to connect to.
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
```
glbrntt, line 1682: This needs updating: it's not necessarily used for negotiation here.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1352,6 +1336,522 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
+            to: address,
+            channelInitializer: channelInitializer
+        )
+    }
+
+    /// Use the existing connected socket file descriptor.
+    ///
+    /// - Parameters:
+    ///   - descriptor: The _Unix file descriptor_ representing the connected stream socket.
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
```
glbrntt, line 1702: This needs updating: it's not necessarily used for negotiation here.

glbrntt (approved):
Looks good but left some doc nits.

glbrntt (approved):
Looks good but left some doc nits.

PR #2440, opened 2023-06-06
Adopt the Swift CoC

Lukasa:
Motivation:

We're centralizing on the Swift code of conduct, so we'll x-reference that instead of holding our own.

Modifications:

Hyperlink out to Swift.

Result:

Shared CoC across the projects.


FranzBusch (approved):

Lukasa:

PR #2443, opened 2023-06-15
Align ServerBootstrap bind methods with the initializer style

FranzBusch:
# Motivation

After experimenting with the bind/connect APIs of the other bootstraps, we realized that passing the type of the handler is not working very well and choose to add an initializer closure to all the bind/connect methods which returns a concrete result.

# Modification

This PR aligns the `ServerBootstrap` APIs with the new initializer based approach. To make this work I had to refactor the `NIOAsyncChannel` SPIs a bit. Overall, this is a net reduction in code.

# Result

We now have aligned bootstrap APIs.


glbrntt (approved):

FranzBusch:
The API breakages are expected in the SPI

Lukasa (approved):

PR #2448, opened 2023-06-22
Add support to xrOS

brennobemoura:
This the release of visionOS Beta SDK, I made this contribution so we can start to use the SwiftNIO with full compatibility in xrOS.

### Motivation:

Some parts of the SwiftNIO codebase has conditional compilation or platform version verification. What I did here was to add the `xrOS 1.0` so the Swift compiler targeting to visionOS SDK can use the right code and not falloff to the `else` statements.

### Modifications:

From this `available(macOS x.x, iOS x.x, tvOS x.x, watchOS x.x, *)` to this `available(macOS x.x, iOS x.x, tvOS x.x, watchOS x.x, xrOS 1, *)`

### Result:

The Swift 5.9 compiles successfully and the compiled code runs great on visionOS.

### DIFF
--- a/IntegrationTests/allocation-counter-tests-framework/template/scaffolding.swift
+++ b/IntegrationTests/allocation-counter-tests-framework/template/scaffolding.swift
@@ -136,7 +136,7 @@ func measureAndPrint(desc: String, trackFDs: Bool, fn: () -> Int) -> Void {
     measurements.printRemainingAllocations(description: desc)
     measurements.printTotalAllocatedBytes(description: desc)
     measurements.printLeakedFDs(description: desc)
-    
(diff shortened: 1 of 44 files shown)


FranzBusch:
I don't think we need to add those. The `*` is a catch-all for unknown platforms and correctly handles this for us. We only need to add them for future releases where new symbols are added to one of the SDKs

brennobemoura:
Actually, this rule applies even for `#available`?

FranzBusch:
@brennobemoura can you close this PR then?

brennobemoura:
I run this locally here and it works. Wow!

PR #2449, opened 2023-06-22
Add support for Musl libc

MaxDesiatov:
Since Musl is sufficiently different from Glibc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html), it requires a different import, which now should be applied to files that have `import Glibc` in them.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOLock.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOLock.swift
@@ -17,8 +17,10 @@ import Darwin
 #elseif os(Windows)
 import ucrt
 import WinSDK
-#else
+#elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #endif
 
 #if os(Windows)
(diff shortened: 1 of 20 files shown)


FranzBusch:
Can we get some CI setup for this, otherwise I see us regressing really fast here. @MaxDesiatov is there any docker image we can use?

MaxDesiatov:
There's no Docker image right now as toolchain PRs aren't ready yet. I'm fine with this regressing until that's sorted out, I'm submitting this PR primarily to simplify local builds for development. Maintaining patches and branches manually across a dozen of repositories is unwieldy and I'd rather have these small changes merged ASAP so that iterating on the toolchain locally is easier.

Lukasa (approved):

finagolfin:
There are some mistakes in this pull, which [just broke my Android CI](https://github.com/finagolfin/swift-android-sdk/actions/runs/5354700464/jobs/9712076566). Android uses `#import Glibc`, but since it's actually Bionic, there were places in NIO where it differentiated by using `os(Linux)` or `os(Android)`, which some of these changes broke. The changes simply adding `#if canImport(Glibc)` before importing `Glibc` are fine, but some of the others are mistaken.

I will submit a pull correcting those later today.

PR #2450, opened 2023-06-23
Close Channel when input & output are closed

felixschlegel:
### Motivation:

Suppose we have a channel whose input and output are closed.
On Linux, this currently results in a full socket closure (which means
TCP `RST` is being sent). This behaviour is currently different
on Darwin because we don't close our socket here when our input and output are
closed.

We should not have two different platform-dependent behaviours here.

As we cannot suppress the TCP `RST` on Linux, we have to align the
Darwin behaviour to the Linux behaviour meaning we close
our entire NIO channel when both the input and the output are closed.

### Modifications:

* `BaseStreamSocketChannel`: escalate to full closure when both its
  input and its output are closed
* add new test
  `ChannelTests.testInputAndOutputClosedResultsInFullClosure()`
  verifying the desired behaviour

### Result:

`BaseStreamSocketChannel` will now close entirely when both input and
output are closed.


Lukasa (changes requested):
Can you extend the test to confirm that the server shuts down too?

Lukasa (approved):

PR #2451, opened 2023-06-23
Correct a few Musl additions from #2449 for Android, plus error if libc not found

finagolfin:
### Motivation:

Fix build errors on Android

### Modifications:

- Fix previous Musl modifications that assumed Glibc wasn't imported on Android
- Add errors for all libc imports, so new platform ports error out early
- Update my github username

### Result:

NIO builds natively on Android again, with all the same tests passing

### DIFF
--- a/Sources/NIOPosix/BSDSocketAPICommon.swift
+++ b/Sources/NIOPosix/BSDSocketAPICommon.swift
@@ -66,7 +66,7 @@ extension NIOBSDSocket.SocketType: Hashable {
 extension NIOBSDSocket.SocketType {
     /// Supports datagrams, which are connectionless, unreliable messages of a
     /// fixed (typically small) maximum length.
-    #if canImport(Glibc)
+    #if os(Linux) && !canImport(Musl)
         internal static let datagram: NIOBSDSocket.SocketType =
                 NIOBSDSocket.SocketType(rawValue: CInt(SOCK_DGRAM.rawValue))
     #else
@@ -76,7 +76,7 @@ extension NIOBSDSocket.SocketType {
 
     /// Supports reliable, two-way, connection-based byte streams without
     /// duplication of data and without preservation of boundaries.
-    #if canImport(Glibc)
+    #if os(Linux) && !canImport(Musl)
         internal static let stream: NIOBSDSocket.SocketType =
                 NIOBSDSocket.SocketType(rawValue: CInt(SOCK_STREAM.rawValue))
     #else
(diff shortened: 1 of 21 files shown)


Sources/NIOPosix/BSDSocketAPICommon.swift
```diff
@@ -66,7 +66,7 @@ extension NIOBSDSocket.SocketType: Hashable {
 extension NIOBSDSocket.SocketType {
     /// Supports datagrams, which are connectionless, unreliable messages of a
     /// fixed (typically small) maximum length.
-    #if canImport(Glibc)
+    #if os(Linux) && !canImport(Musl)
```
finagolfin, line 69: @MaxDesiatov, I presume this was your intent when you changed these two lines in this file.

Sources/NIOPosix/System.swift
```diff
@@ -37,7 +37,7 @@ import CNIOWindows
 
 internal typealias MMsgHdr = CNIOWindows_mmsghdr
 #else
-let badOS = { fatalError("unsupported OS") }()
```
finagolfin, line 40: I have no idea why @weissi used this formulation instead of erroring, as I added now.

Sources/NIOPosix/System.swift
```diff
@@ -111,7 +111,7 @@ private let sysIfNameToIndex: @convention(c) (UnsafePointer<CChar>?) -> CUnsigne
 private let sysSocketpair: @convention(c) (CInt, CInt, CInt, UnsafeMutablePointer<CInt>?) -> CInt = socketpair
 #endif
 
-#if canImport(Glibc)
+#if os(Linux) && !canImport(Musl)
```
finagolfin, line 114: @MaxDesiatov, I don't know why you changed this for Musl, these functions aren't there? Because this change you made doesn't define them for Musl at all.

Lukasa (commented):
I’m generally happy with this but would like @MaxDesiatov to confirm as well.

Sources/NIOPosix/System.swift
weissi: That was from before Swift had `#error(...)`.

Sources/NIOPosix/System.swift
finagolfin: @MaxDesiatov, do you define these for Musl? If so, feel free to add a commit to this pull with those Musl defintions, and we can get that fix in too.

Lukasa (approved):

Lukasa:

PR #2457, opened 2023-07-02
Fix last remaining test on Android

finagolfin:
### Motivation:

Get all tests passing on Android AArch64

### Modifications:

- Check if `localhost` is set for IPv6 in `BootstrapTest/testClientBindWorksOnSocketsBoundToEitherIPv4OrIPv6Only`, and use `ip6-localhost` if not

### Result:

All tests pass natively on Android AArch64

The problem is that the localhost definitions in `/etc/hosts` on Android look like this, so [I've long disabled this test on my Android CI](https://github.com/finagolfin/swift-android-sdk/commit/7938a8102844000e1398ab4196eec649d8f6074e#diff-2305028dd5b4c6982108b8f4044fd87f05517499e57ca03c7d8e4b4230ccd949R10):
```
> ag localhost /etc/hosts
127.0.0.1       localhost
::1             ip6-localhost
```
Whereas this test passed on linux because it defines `localhost` for both protocols:
```
> ag localhost /etc/hosts
127.0.0.1 localhost
::1 localhost ip6-localhost ip6-loopback
```
~Binging it online, I see that Darwin doesn't appear to define `ip6-localhost`, so I kept that the same.~ I only tested this change natively on Android, so let's see if the CI catches anything on linux and macOS.

### DIFF
--- a/Tests/NIOPosixTests/BootstrapTest.swift
+++ b/Tests/NIOPosixTests/BootstrapTest.swift
@@ -644,10 +644,21 @@ class BootstrapTest: XCTestCase {
 
             var maybeChannel1: Channel? = nil
             // Try 2: Connect to "localhost", this will do Happy Eyeballs.
+            var localhost = "localhost"
+            // Some platforms don't define "localhost" for IPv6, so check that
+            // and use "ip6-localhost" instead.
+            if !isIPv4 {
+                let hostResolver = GetaddrinfoResolver(loop: group.next(), aiSocktype: .stream, aiProtocol: .tcp)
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
Thanks for this patch!

I worry about this code being a little brittle. Can I suggest that we instead do a DNS resolution on `localhost` and `ip6-localhost` to check whether the names are defined, and what they resolve to? That will make this code a lot more predictable, and work on systems regardless of how specifically they've handled `/etc/hosts`.

finagolfin:
Made the DNS resolution change you asked for, let me know what you think.

Lukasa (approved):

PR #2458, opened 2023-07-03
Async support NIOPipeBootstrap

FranzBusch:
# Motivation
We want to support async bootstrapping with all our bootstraps.

# Modification
This PR adds support for the `NIOPipeBootstrap` and adds the three variants of methods to it (abstract output, `NIOAsyncChannel` based and protocol negotiation based)

# Result
We now support asynchronous interaction with the `NIOPipeBootstrap`

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -3084,19 +3084,19 @@ public final class NIOPipeBootstrap {
     ///
     /// This method is useful for specialilsed use-cases where you want to use `NIOPipeBootstrap` for say a serial line.
     ///
-    /// - Note: If this method returns a succeeded future, SwiftNIO will close `fileDescriptor` when the `Channel`
-    ///         becomes inactive. You _must not_ do any further operations with `fileDescriptor`, including `close`.
+    /// - Note: If this method returns a succeeded future, SwiftNIO will close `inputOutput` when the `Channel`
+    ///         becomes inactive. You _must not_ do any further operations with `inputOutput`, including `close`.
     ///         If this method returns a failed future, you still own the file descriptor and are responsible for
     ///         closing it.
     ///
     /// - Parameters:
-    ///   - fileDescriptor: The _Unix file descriptor_ for the input & output.
+    ///   - inputOutput: The _Unix file descriptor_ for the input & output.
     /// - Returns: an `EventLoopFuture<Channel>` to deliver the `Channel`.
-    public func takingOwnershipOfInputOutputDescriptor(_ fileDescriptor: CInt) -> EventLoopFuture<Channel> {
-        let inputFD = fileDescriptor
-        let outputFD = try! Posix.dup(descriptor: fileDescriptor)
+    public func takingOwnershipOfDescriptor(inputOutput: CInt) -> EventLoopFuture<Channel> {
+        let inputFD = inputOutput
+        let outputFD = try! Posix.dup(descriptor: inputOutput)
 
-        return self.takingOwnershipOfInputOutputDescriptor(inputDescriptor: inputFD, outputDescriptor: outputFD).flatMapErrorThrowing { error in
+        return self.takingOwnershipOfDescriptors(input: inputFD, output: outputFD).flatMapErrorThrowing { error in
             try! Posix.close(descriptor: outputFD)
             throw error
         }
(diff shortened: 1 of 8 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3169,6 +3169,324 @@ public final class NIOPipeBootstrap {
...
+    ///   - fileDescriptor: The _Unix file descriptor_ for the input & output.
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
+    ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
+    ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
+    ///   - inboundType: The channel's inbound type.
+    ///   - outboundType: The channel's outbound type.
+    /// - Returns: A `NIOAsyncChannel` for the bound connection.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func withInputOutputDescriptor<Inbound: Sendable, Outbound: Sendable>(
```
glbrntt, line 3195: Hmm `with` suggests a block here which isn't the case (I see these are carried over from the existing bootstrap) -- I wonder if we should try to use a better name here?

Relatedly this takes ownership of the descriptor which is only clear from the docs. We should probably try to weave that into the name as well. I wonder if something like `takingOwnershipOfInputOutputDescriptor(...)` makes sense?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3169,6 +3169,324 @@ public final class NIOPipeBootstrap {
...
+    ///   - outputDescriptor: The _Unix file descriptor_ for the output (ie. the write side).
+    ///   - backpressureStrategy: The back pressure strategy used by the channel.
+    ///   - isOutboundHalfClosureEnabled: Indicates if half closure is enabled on the channel. If half closure is enabled
+    ///   then finishing the `NIOAsyncChannelWriter` will lead to half closure.
+    ///   - inboundType: The channel's inbound type.
+    ///   - outboundType: The channel's outbound type.
+    /// - Returns: A `NIOAsyncChannel` for the bound connection.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func withPipes<Inbound: Sendable, Outbound: Sendable>(
```
glbrntt, line 3240: same note re: with and taking ownership

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3169,6 +3169,324 @@ public final class NIOPipeBootstrap {
...
+    ///   the protocol.
+    /// - Returns: The protocol negotiation result.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func withInputOutputDescriptor<Handler: NIOProtocolNegotiationHandler>(
+        _ fileDescriptor: CInt,
+        channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<Handler>
+    ) async throws -> Handler.NegotiationResult {
+        let inputFD = fileDescriptor
+        let outputFD = try! Posix.dup(descriptor: fileDescriptor)
```
glbrntt, line 3290: It's a bit sad we do a syscall here.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3169,6 +3169,324 @@ public final class NIOPipeBootstrap {
...
+    ///   - channelInitializer: A closure to initialize the channel which must return the handler that is used for negotiating
+    ///   the protocol.
+    /// - Returns: The protocol negotiation result.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel)
+    public func withPipes<Handler: NIOProtocolNegotiationHandler>(
+        inputDescriptor: CInt,
+        outputDescriptor: CInt,
+        channelInitializer: @escaping @Sendable (Channel) -> EventLoopFuture<Handler>
+    ) async throws -> Handler.NegotiationResult {
```
glbrntt, line 3328: Can the body of this be written in terms of the variant that returns `Output`? There's a lot of duplication.

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -78,6 +78,7 @@ private final class TLSUserEventHandler: ChannelInboundHandler, RemovableChannel
 
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
         let buffer = self.unwrapInboundIn(data)
+        print(buffer)
```
glbrntt, line 81: nit: leftover debugging

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -102,6 +103,7 @@ private final class ByteBufferToStringHandler: ChannelDuplexHandler {
...
```
glbrntt, line 106: nit: leftover debugging

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -586,8 +588,108 @@ final class AsyncChannelBootstrapTests: XCTestCase {
             }
         }
     }
+    
+    // MARK: - Pipe Bootstrap
+    
+    func testPipeBootstrap() throws {
+        XCTAsyncTest {
```
glbrntt, line 595: I'm pretty sure we no longer need `XCTAsyncTest`, `async` tests should work from 5.6

Sources/NIOPosix/Bootstrap.swift
FranzBusch: cc @Lukasa @weissi WDYT?

Lukasa: I like George's suggestion here.

Sources/NIOPosix/Bootstrap.swift
Lukasa: `dup` is pretty cheap, so it's not really a problem.

Sources/NIOPosix/Bootstrap.swift
FranzBusch: Do we want to deprecate and rename the old methods as well to align ELF and async APIs?

FranzBusch: Renamed all methods and deprecated the old ones

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3108,17 +3108,17 @@ public final class NIOPipeBootstrap {
...
-    /// - parameters:
-    ///     - inputDescriptor: The _Unix file descriptor_ for the input (ie. the read side).
-    ///     - outputDescriptor: The _Unix file descriptor_ for the output (ie. the write side).
-    /// - returns: an `EventLoopFuture<Channel>` to deliver the `Channel`.
-    public func withPipes(inputDescriptor: CInt, outputDescriptor: CInt) -> EventLoopFuture<Channel> {
+    /// - Parameters:
+    ///   - inputDescriptor: The _Unix file descriptor_ for the input (ie. the read side).
+    ///   - outputDescriptor: The _Unix file descriptor_ for the output (ie. the write side).
+    /// - Returns: an `EventLoopFuture<Channel>` to deliver the `Channel`.
+    public func takingOwnershipOfInputOutputDescriptor(inputDescriptor: CInt, outputDescriptor: CInt) -> EventLoopFuture<Channel> {
```
glbrntt, line 3121: Can we call this one `takingOwnershipOfDescriptors(input:output:)`?  There's a lot of repetition otherwise...

glbrntt: Might make sense for the other one to be `takingOwnershipOfDescriptor(inputOutput:)` 🤷‍♂️ (sorry, I know I suggested `takingOwnershipOfInputOutputDescriptor(_:)` ...)

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -3167,6 +3167,312 @@ public final class NIOPipeBootstrap {
...
+        try await self.takingOwnershipOfInputOutputDescriptor(
+            inputDescriptor: inputDescriptor,
+            outputDescriptor: outputDescriptor,
+            channelInitializer: channelInitializer,
+            postRegisterTransformation: { $0.makeSucceededFuture($1) }
+        )
+    }
+    
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @_spi(AsyncChannel) // Should become private
```
glbrntt, line 3425: maybe `_` prefix the name?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -2968,7 +2968,7 @@ extension DatagramBootstrap: Sendable {}
 ///                       .channelInitializer { channel in
 ///                           channel.pipeline.addHandler(MyChannelHandler())
 ///                       }
-///                       .withPipes(inputDescriptor: STDIN_FILENO, outputDescriptor: STDOUT_FILENO)
+///                       .takingOwnershipOfInputOutputDescriptor(inputDescriptor: STDIN_FILENO, outputDescriptor: STDOUT_FILENO)
```
glbrntt, line 2971:
```suggestion
///                       .takingOwnershipOfInputOutputDescriptor(input: STDIN_FILENO, output: STDOUT_FILENO)
```

glbrntt (approved):
Left a fixit but looks good otherwise!

PR #2459, opened 2023-07-03
Make `NIO[Throwing]AsyncSequenceProducer.Source` `Sendable`

felixschlegel:


Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -329,3 +329,5 @@ extension NIOAsyncSequenceProducer {
 @available(*, unavailable)
 extension NIOAsyncSequenceProducer.AsyncIterator: Sendable {}
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOAsyncSequenceProducer.Source: Sendable {}
```
FranzBusch, line 333: We have seen a bunch of use-cases where we really want to construct the source in one isolation domain and move it over to another. This is currently producing warnings due to the missing `Sendable` conformance. The initial intention was to have it non `Sendable` since it should be tied to a `Channel` and constructed on the `Channel`s EL. However, the type now is getting adopted wider and the missing conformance is problematic. Since the underlying type is fully thread safe, I am fine with adding this now. cc @Lukasa

FranzBusch (approved):

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
Lukasa: Any reason not to do the throwing one as well?

Lukasa (approved):

PR #2462, opened 2023-07-06
Restructure `Package.swift`

FranzBusch:
This might be a controversial PR! 🍿

# Motivation
It always bothered me how our `Package.swift` is structured. First, we had a huge list of `targets` at the top that distracted from the `products`. The `products` are the important part for our adopters since those are the actual modules they can use. Secondly, the formatting of the targets made it hard to follow and resulted in unnecessary merge conflicts.

# Modification
This PR moves the targets back to the bottom and splits all the various array elements into separate elements.

# Result
Easier to understand `Package.swift` and less merge conflicts.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,110 +18,6 @@ import PackageDescription
 let swiftAtomics: PackageDescription.Target.Dependency = .product(name: "Atomics", package: "swift-atomics")
 let swiftCollections: PackageDescription.Target.Dependency = .product(name: "DequeModule", package: "swift-collections")
 
-var targets: [PackageDescription.Target] = [
-    .target(name: "NIOCore",
-            dependencies: ["NIOConcurrencyHelpers", "CNIOLinux", "CNIOWindows", swiftCollections, swiftAtomics]),
-    .target(name: "_NIODataStructures"),
-    .target(name: "NIOEmbedded",
-            dependencies: ["NIOCore",
-                           "NIOConcurrencyHelpers",
-                           "_NIODataStructures",
-                           swiftAtomics,
-                           swiftCollections]),
-    .target(name: "NIOPosix",
-            dependencies: ["CNIOLinux",
-                           "CNIODarwin",
-                           "CNIOWindows",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -13,115 +13,10 @@
...
-    .testTarget(name: "NIOWebSocketTests",
-                dependencies: ["NIOCore", "NIOEmbedded", "NIOWebSocket"]),
-    .testTarget(name: "NIOTestUtilsTests",
-                dependencies: ["NIOTestUtils", "NIOCore", "NIOEmbedded", "NIOPosix"]),
-    .testTarget(name: "NIOFoundationCompatTests",
-                dependencies: ["NIOCore", "NIOFoundationCompat"]),
-    .testTarget(name: "NIOTests",
-                dependencies: ["NIO"]),
-]
+import PackageDescription
```
dnadoba, line 19: The `import` should be at the top before the dependency definitions.

Package.swift
```diff
@@ -143,5 +38,351 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.0.2"),
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"),
     ],
-    targets: targets
+    targets: [
```
glbrntt, line 41: I think it would be useful to group targets into 'blocks', e.g. "modules", "examples", and "tests" and separate each with a `// MARK:` or similar.

FranzBusch: Added `// MARK:`s and grouped them

Package.swift
```diff
@@ -143,5 +39,359 @@ let package = Package(
...
+        // MARK: Targets
```
0xTim, line 43:
```suggestion
        // MARK: - Targets
```
Makes use of Xcode's formatting for splitting up stuff

Lukasa (approved):
Sure, why not.

FranzBusch:
Merging over the known allocation regression.

PR #2463, opened 2023-07-07
Add async TCP echo example

FranzBusch:
# Motivation
We introduced new async APIs to our bootstraps that make it easy to use NIO from Swift Concurrency. We should provide examples showing how to use those new APIs. In the future, we want to make those examples the primary ones and remove the non-async based ones.

# Modification
This PR introduces two new targets: `NIOTCPEchoServer` and `NIOTCPEchoClient`. Both targets are executable and provide either the server or client piece of the example.

# Result
New and shiny async example

### DIFF
--- a/Sources/NIOTCPEchoClient/Client.swift
+++ b/Sources/NIOTCPEchoClient/Client.swift
@@ -74,7 +74,9 @@ struct Client {
         for try await inboundData in channel.inboundStream {
             print("Connection(\(number)): Received response (\(inboundData))")
 
-            // We only expect a single response so we can exit here
+            // We only expect a single response so we can exit here.
+            // Once, we exit out of this loop and the references to the `NIOAsyncChannel` are dropped
+            // the connection is going to close itself.
             break
         }
     }
@@ -92,10 +94,10 @@ private final class NewlineDelimiterCoder: ByteToMessageDecoder, MessageToByteEn
     func decode(context: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
         let readableBytes = buffer.readableBytesView
 
-        if let firstLine = readableBytes.firstIndex(of: UInt8(ascii: "\n")).map({ readableBytes[readableBytes.startIndex ..< $0] }) {
+        if let firstLine = readableBytes.firstIndex(of: self.newLine).map({ readableBytes[..<$0] }) {
             buffer.moveReaderIndex(forwardBy: firstLine.count + 1)
             // Fire a read without a newline
-            context.fireChannelRead(self.wrapInboundOut(String(buffer: ByteBuffer(firstLine.dropLast()))))
+            context.fireChannelRead(self.wrapInboundOut(String(buffer: ByteBuffer(firstLine))))
             return .continue
         } else {
             return .needMoreData
@@ -104,7 +106,7 @@ private final class NewlineDelimiterCoder: ByteToMessageDecoder, MessageToByteEn
 
     func encode(data: String, out: inout ByteBuffer) throws {
         out.writeString(data)
-        out.writeString("\n")
+        out.writeInteger(self.newLine)
     }
 }
 #else
--- a/Sources/NIOTCPEchoClient/Client.swift
+++ b/Sources/NIOTCPEchoClient/Client.swift
@@ -23,30 +23,15 @@ struct Client {
     /// The port to connect to.
     private let port: Int
     /// The client's event loop group.
-    private let eventLoopGroup: any EventLoopGroup
+    private let eventLoopGroup: MultiThreadedEventLoopGroup
 
     static func main() async throws {
-        // We are creating this event loop group at the top level and it will live until
-        // the process exits. This means we also don't have to shut it down.
-        //
-        // Note that we start this group with 1 thread. In general, most NIO programs
-        // should use 1 thread as a default unless they're planning to be a network
-        // proxy or something that is expecting to be dominated by packet parsing. Most
-        // servers aren't, and NIO is very fast, so 1 NIO thread is quite capable of
-        // saturating the average small to medium sized machine.
-        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
-
         let client = Client(
             host: "localhost",
             port: 8765,
-            eventLoopGroup: eventLoopGroup
+            eventLoopGroup: .singleton
         )
         try await client.run()
-
-        print("Done sending requests; exiting in 5 seconds")
-        // We are only sleeping here to keep the terminal output visible a bit longer in Xcode
-        // so that users can see that something happened. This should normally not be done!
-        try await Task.sleep(for: .seconds(5))
     }
 
     /// This method sends a bunch of requests.
@@ -63,22 +48,25 @@ struct Client {
     }
 
     private func sendRequest(number: Int) async throws {
-        let channel = try await ClientBootstrap(group: eventLoopGroup)
-            .channelInitializer { channel in
+        let channel = try await ClientBootstrap(group: self.eventLoopGroup)
+            .connect(
+                host: self.host,
+                port: self.port
+            ) { channel in
                 channel.eventLoop.makeCompletedFuture {
                     // We are using two simple handlers here to frame our messages with "\n"
                     try channel.pipeline.syncOperations.addHandler(ByteToMessageHandler(NewlineDelimiterCoder()))
                     try channel.pipeline.syncOperations.addHandler(MessageToByteHandler(NewlineDelimiterCoder()))
+
+                    return try NIOAsyncChannel(
+                        synchronouslyWrapping: channel,
+                        configuration: .init(
+                            inboundType: String.self,
+                            outboundType: String.self
+                        )
+                    )
                 }
             }
-            .connect(
-                host: self.host,
-                port: self.port,
-                channelConfiguration: .init(
-                    inboundType: String.self,
-                    outboundType: String.self
-                )
-            )
 
         print("Connection(\(number)): Writing request")
         try await channel.outboundWriter.write("Hello on connection \(number)")
@@ -97,7 +85,7 @@ private final class NewlineDelimiterCoder: ByteToMessageDecoder, MessageToByteEn
     typealias InboundIn = ByteBuffer
     typealias InboundOut = String
 
-    private let newLine = "\n".utf8.first!
+    private let newLine = UInt8(ascii: "\n")
 
     init() {}
 
--- a/Sources/NIOTCPEchoClient/Client.swift
+++ b/Sources/NIOTCPEchoClient/Client.swift
@@ -36,18 +36,20 @@ struct Client {
         // saturating the average small to medium sized machine.
         let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
 
-        let server = Client(
+        let client = Client(
             host: "localhost",
             port: 8765,
             eventLoopGroup: eventLoopGroup
         )
-        try await server.run()
+        try await client.run()
 
         print("Done sending requests; exiting in 5 seconds")
+        // We are only sleeping here to keep the terminal output visible a bit longer in Xcode
+        // so that users can see that something happened. This should normally not be done!
         try await Task.sleep(for: .seconds(5))
     }
 
-    /// This method starts the server and handles incoming connections.
+    /// This method sends a bunch of requests.
     func run() async throws {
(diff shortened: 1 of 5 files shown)


Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        let server = Client(
+            host: "localhost",
+            port: 8765,
+            eventLoopGroup: eventLoopGroup
+        )
+        try await server.run()
+
+        print("Done sending requests; exiting in 5 seconds")
+        try await Task.sleep(for: .seconds(5))
```
glbrntt, line 47: why do we sleep?

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,91 @@
...
+        // the process exits. This means we also don't have to shut it down.
+        //
+        // Note that we start this group with 1 thread. In general, most NIO programs
+        // should use 1 thread as a default unless they're planning to be a network
+        // proxy or something that is expecting to be dominated by packet parsing. Most
+        // servers aren't, and NIO is very fast, so 1 NIO thread is quite capable of
+        // saturating the average small to medium sized machine.
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+
+        let server = Client(
```
glbrntt, line 39: Pretty sure this is a client ;)

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        for try await inboundData in channel.inboundStream {
+            print("Connection(\(number)): Received response (\(String(buffer: inboundData)))")
+
+            // We only expect a single response so we can exit here
+            break
+        }
+    }
+}#else
+@main
```
glbrntt, line 85:
```suggestion
}
#else
@main
```

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,96 @@
...
+                port: self.port,
+                childChannelConfiguration: .init(
+                    inboundType: ByteBuffer.self,
+                    outboundType: ByteBuffer.self
+                )
+            )
+
+        // We are handling each incoming connection in a separate child task. It is important
+        // to use a discarding task group here which automatically discards finished child tasks
+        // otherwise this task group would end up leaking memory of all finished connection tasks.
```
glbrntt, line 62: I think we should elaborate on this a bit because it makes it sound like task groups leak! We don't leak memory if we use a task group, we just don't free it until the task group exits or we consume the results.

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,91 @@
...
+    static func main() async throws {
+        // We are creating this event loop group at the top level and it will live until
+        // the process exits. This means we also don't have to shut it down.
+        //
+        // Note that we start this group with 1 thread. In general, most NIO programs
+        // should use 1 thread as a default unless they're planning to be a network
+        // proxy or something that is expecting to be dominated by packet parsing. Most
+        // servers aren't, and NIO is very fast, so 1 NIO thread is quite capable of
+        // saturating the average small to medium sized machine.
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
weissi, line 37: Sure, in this case you don't have to shut it down but why leak stuff if it's trivial to not leak?

`defer { try! elg.syncShutdownGracefully() }`

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,91 @@
...
+                    inboundType: ByteBuffer.self,
+                    outboundType: ByteBuffer.self
+                )
+            )
+
+        print("Connection(\(number)): Writing request")
+        try await channel.outboundWriter.write(ByteBuffer(string: "Hello on connection \(number)"))
+
+        for try await inboundData in channel.inboundStream {
+            print("Connection(\(number)): Received response (\(String(buffer: inboundData)))")
```
weissi, line 78: The string conversion here is incorrect. You can receive in arbitrarily sized chunks. So either you frame or you write each chunk as bytes (and crucially not String).

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,96 @@
...
+    static func main() async throws {
+        // We are creating this event loop group at the top level and it will live until
+        // the process exits. This means we also don't have to shut it down.
+        //
+        // Note that we start this group with 1 thread. In general, most NIO programs
+        // should use 1 thread as a default unless they're planning to be a network
+        // proxy or something that is expecting to be dominated by packet parsing. Most
+        // servers aren't, and NIO is very fast, so 1 NIO thread is quite capable of
+        // saturating the average small to medium sized machine.
+        let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
weissi, line 37: unnecessary (but in this case benign leak). But be mindful that this code will end up copy&pasted potentially into functions that get called multiple times.

weissi: If you really want a singleton, it should be a lazy global variable and named appropriately.

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /// This method handles a single connection by echoing back all inbound data.
+    private func handleConnection(channel: NIOAsyncChannel<ByteBuffer, ByteBuffer>) async {
+        // Note that this method is non-throwing and we are catching any error.
+        // We do this since we don't want to tear down the whole server when a single connection
+        // encounters an error.
+        do {
+            for try await inboundData in channel.inboundStream {
+                print("Received request (\(String(buffer: inboundData)))")
```
weissi, line 81: same issue here, the printing is incorrect as we're not framing

Sources/NIOTCPEchoClient/Client.swift
weissi: So moving this into a global variable and naming it `globalSingletonGroup` or so would be much better because copy & paste wouldn't lead to super bad leaks. The code as is is just incorrect because `eventLoopGroup` will get `deinit`ed and it's illegal to `deinit` this without shutting it down (don't think that's enforced today though).

Thinking about it more: I've long come to the conclusion that NIO needs to provides singletons for users who don't need tight control and I think https://github.com/apple/swift-nio/pull/2471 is the answer :)

weissi (changes requested):
Just leaking an `EventLoopGroup` in a _function_ is illegal (will get deinited without shutdown which is a lifecycle violation)

Sources/NIOTCPEchoClient/Client.swift
FranzBusch: If a user runs this in Xcode he gets a couple of seconds to see the output. Otherwise it is just gone immediately.

glbrntt: Oof, that's pretty annoying. Could you add a comment explaining that and that it's not actually necessary? Examples get copy-pasted a lot so we don't want users to actually do this!

Sources/NIOTCPEchoClient/Client.swift
FranzBusch: Holding this PR until we land the singleton work. Thanks @weissi!

Sources/NIOTCPEchoClient/Client.swift
FranzBusch: I added a newline based decoder/encoder to both pipelines. I think that is actually good to have in an example since if somebody copies it over they are aware of the need for framing right away.

FranzBusch:
@weissi I updated this PR to use your newly landed singletons.

Sources/NIOTCPEchoClient/Client.swift
weissi: Yeah, can we not do that? Let's find a better way that demonstrates this without just delaying it here...

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+
+/// A simple newline based encoder and decoder.
+private final class NewlineDelimiterCoder: ByteToMessageDecoder, MessageToByteEncoder {
+    typealias InboundIn = ByteBuffer
+    typealias InboundOut = String
+
+    private let newLine = "\n".utf8.first!
```
weissi, line 96:
```suggestion
    private let newLine = UInt8(ascii: "\n")
```

Sources/NIOTCPEchoClient/Client.swift
FranzBusch: I mean I am happy to just not do this.

Sources/NIOTCPEchoClient/Client.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    init() {}
+
+    func decode(context: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        let readable = buffer.withUnsafeReadableBytes { $0.firstIndex(of: self.newLine) }
```
weissi, line 93: could we please not use `*Unsafe*` in the examples. `buffer.readableBytesView.firstIndex(of: UInt8(ascii: "\n"))` gives you what you need.

weissi: or probably something like

```
let readableBytes = buffer.readableBytesView

if let firstLine = readableBytes.firstIndex(of: UInt8(ascii: "\n")).map({ readableBytes[readableBytes.startIndex ..< $0] }) {
    buffer.moveReaderIndex(forwardBy: firstLine.count)
    context.fireChannelRead(self.wrapInboundOut(ByteBuffer(firstLine.dropLast() /* kill newline */)))
    return .continue
} else {
    return .needMoreData
}
```

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,125 @@
...
+private final class NewlineDelimiterCoder: ByteToMessageDecoder, MessageToByteEncoder {
+    typealias InboundIn = ByteBuffer
+    typealias InboundOut = String
+
+    private let newLine = UInt8(ascii: "\n")
+
+    init() {}
+
+    func decode(context: ChannelHandlerContext, buffer: inout ByteBuffer) throws -> DecodingState {
+        let readable = buffer.withUnsafeReadableBytes { $0.firstIndex(of: self.newLine) }
```
weissi, line 104: same here, unnecessary Unsafe

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,125 @@
...
+                port: self.port
+            ) { channel in
+                channel.eventLoop.makeCompletedFuture {
+                    // We are using two simple handlers here to frame our messages with "\n"
+                    try channel.pipeline.syncOperations.addHandler(ByteToMessageHandler(NewlineDelimiterCoder()))
+                    try channel.pipeline.syncOperations.addHandler(MessageToByteHandler(NewlineDelimiterCoder()))
+
+                    return try NIOAsyncChannel(
+                        synchronouslyWrapping: channel,
+                        configuration: .init(
```
weissi, line 52: `.init` is pretty ugly, could we get a type name?

Sources/NIOTCPEchoServer/Server.swift
```diff
@@ -0,0 +1,125 @@
...
+    func run() async throws {
+        let channel = try await ServerBootstrap(group: eventLoopGroup)
+            .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
+            .bind(
+                host: self.host,
+                port: self.port
+            ) { channel in
+                channel.eventLoop.makeCompletedFuture {
+                    // We are using two simple handlers here to frame our messages with "\n"
+                    try channel.pipeline.syncOperations.addHandler(ByteToMessageHandler(NewlineDelimiterCoder()))
```
weissi, line 47: all these `try`s should be `try!`s, they can't fail unless programmer error, right?

FranzBusch: They could throw if the channel is closed while we add the handler. I don't wanna show adopters a `try!` when it could fail in rare cases and I don't think the `try` is too bad.

glbrntt (approved):

weissi (approved):
lgtm

PR #2464, opened 2023-07-10
Add configuration for `NIOAsyncChannel`

FranzBusch:
# Motivation
While adopting the new `NIOAsyncChannel` type we saw an exploding number of parameters passed to the various connect/bind/configure methods. All these methods had in common that we had to pass configuration for the `NIOAsyncChannel`.

# Modification
This PR introduces a new type `NIOAsyncChannel.Configuration` which groups all four configuration parameters into a struct.

# Result
We can now write more concise methods that use a single configuration object.

### DIFF
--- a/Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
+++ b/Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
@@ -225,8 +225,10 @@ final class AsyncChannelBootstrapTests: XCTestCase {
             .bind(
                 host: "127.0.0.1",
                 port: 0,
-                childChannelInboundType: String.self,
-                childChannelOutboundType: String.self
+                childChannelConfiguration: .init(
+                    inboundType: String.self,
+                    outboundType: String.self
+                )
             )
 
         try await withThrowingTaskGroup(of: Void.self) { group in
@@ -565,7 +567,7 @@ final class AsyncChannelBootstrapTests: XCTestCase {
         let channel = try await DatagramBootstrap(group: eventLoopGroup)
             .bind(
                 to: .init(ipAddress: "127.0.0.1", port: 0),
-                channelInitializer:  { channel -> EventLoopFuture<Channel> in channel.eventLoop.makeSucceededFuture(channel) }
+                channelInitializer: { channel -> EventLoopFuture<Channel> in channel.eventLoop.makeSucceededFuture(channel) }
             )
 
         let port = channel.localAddress!.port!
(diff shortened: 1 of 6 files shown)


Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -1026,11 +1021,8 @@ final class AsyncChannelBootstrapTests: XCTestCase {
                 case "string":
                     return channel.eventLoop.makeCompletedFuture {
                         try channel.pipeline.syncOperations.addHandler(ByteBufferToStringHandler())
-                        let asyncChannel = try NIOAsyncChannel(
-                            synchronouslyWrapping: channel,
-                            isOutboundHalfClosureEnabled: true,
```
rnro, line 1031: Isn't this dropped when we use the default configuration?
```
isOutboundHalfClosureEnabled: true,
```

Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -1039,11 +1031,8 @@ final class AsyncChannelBootstrapTests: XCTestCase {
                     return channel.eventLoop.makeCompletedFuture {
                         try channel.pipeline.syncOperations.addHandler(ByteBufferToByteHandler())
 
-                        let asyncChannel = try NIOAsyncChannel(
-                            synchronouslyWrapping: channel,
-                            isOutboundHalfClosureEnabled: true,
```
rnro, line 1044: Same here about the config being dropped

FranzBusch:

rnro (approved):

glbrntt (approved):

FranzBusch:
- API breakage is expected
- 5.9 is a known allocation regression
- nightly seems to have broken on the allocation tests since a couple of commits. Going to look into this later today

finagolfin:
This [broke building NIO HTTP2 on my Android CI today](https://github.com/finagolfin/swift-android-sdk/actions/runs/5517966208/jobs/10061316013#step:28:288) because it builds against NIO HEAD, which I can reproduce on linux.

Since this previous expansive constructor has shipped in NIO tags, maybe it should be deprecated, not removed?

glbrntt:
We only make API guarantees about tagged releases. In this case we're modifying `@_spi` which is not part of our declared API so we're free to remove it between tagged releases. I appreciate this is annoying but it should be resolved soon.

finagolfin:
OK, I was unaware of `@_spi`, hence the confusion.

Issue #2142, opened 2022-06-01
ecosystem issue: EventLoopGroupProvider and the 'similar create or share ELG' pattern


weissi:
The NIO ecosystem suffers from one composability issue: `EventLoopGroupProvider` and similar constructs. The issue is that taking `EventLoopGroupProvider` (or similar) means that a library can either own or not own an `EventLoopGroup`.

That doesn't sound too bad but unfortunately, it's very bad for the shutdown case.

A library (like say AsyncHTTPClient) would obviously want to have a method called `thing.shutdown() -> EventLoopFuture<Void>` to match the other functions. Unfortunately, that can't be done because the library may own the `EventLoopGroup`, therefore `shutdown()` has to may have to shut it down which then however means that there is no more `EventLoop` left to run the future on.

The current hack is to instead provide a `shutdown(group: DispatchGroup, _ completion: @escaping (Error?) -> Void)` method. But that's both ugly (doesn't fit the rest) and wasteful (needs to spin up a dispatch thread for no reason).

Unfortunately, this issue won't just magically be fixed by the universal adoption of Swift Concurrency either. Yes, it's possible to provide a `func shutdown() async throws -> Void` but that'll then bounce from a Concurrency executor to the EventLoop to a DispatchQueue thread back to some Concurrency executor. That's still not great.

Also, there's another issue if `EventLoop(Group)`s frequently go away: What do we do with people who caught a reference to an EL(G) and then (after the shutdown) `execute` stuff on it? Currently we print this weird warning that we'll crash in future releases. That's cool if EL(G)s are mostly global/created in main but it really doesn't work if they're frequently created/destroyed.

Also there's a question w.r.t. Custom Executors (once they actually arrive in Swift): Will they support this use case?

```
func myFunc() async throws -> Void {
    let httpClient = HTTPClient(eventLoopGroup: .createNew)
    // on some Concurrency executor
    do {
        let result = try await httpClient.get(...)
        // on a NIO EventLoop Custom Executor
        print(result)
        // we should be sure that this is actually legal to shut down the executor we're currently
        try await httpClient.shutdown() // this is also ugly because no async in defer
    } catch {
        try await httpClient.shutdown() // this is also ugly because no async in defer
    }
}
```

---

My recommendations are:
- deprecate `EventLoopGroupProvider`
- deprecate the pattern that an instance of a library (like `HTTPClient`) creates its own EL(G) threads

[...]

PR #2471, opened 2023-07-13
NIOSingletons: Use NIO in easy mode

### LINKED ISSUE #2472: NIOSingletons: Make EventLoopGroup shutdown throw
Once we have https://github.com/apple/swift-nio/pull/2471

The `EventLoopGroup`s returned from `NIOSingletons` should throw if anybody attempts to shut them down. We just need some `_internalParameter` to `MultiThreadedEventLoopGroup.init` that disallows shutdown. There's precedent for this: The `EventLoop`s you get from `MultiThreadedEventLoopGroup.next()` also throw on shutdown.

weissi:
### Motivation:

SwiftNIO allows and encourages to precisely manage all operating system resources like file descriptors & threads. That is a very important property of the system but in many places -- especially since Swift Concurrency arrived -- many simpler SwiftNIO programs only require a single, globally shared EventLoopGroup. Often even with just one thread.

Long story short: Many, probably most users would happily trade precise control over the threads for not having to pass around `EventLoopGroup`s. Today, many of those users resort to creating (and often leaking) threads because it's simpler. Adding a `.globalSingle` static var which _lazily_ provides singleton `EventLoopGroup`s and `NIOThreadPool`s is IMHO a much better answer.

Finally, this aligns SwiftNIO a little more with Apple's SDKs which have a lot of global singletons that hold onto system resources (`Dispatch`'s thread pool, `URLSession.shared`, ...). At least in `Dispatch`'s case the Apple SDKs actually make it impossible to manage the resources, there can only ever be one global pool of threads. That's fine for app development but wouldn't be good enough for certain server use cases, therefore I propose to add `NIOSingleton`s as an _option_ to the user. Confident programmers (especially in libraries) are still free and encouraged to manage all the resources deterministically and explicitly.

Companion PRs: 
 - https://github.com/apple/swift-nio-transport-services/pull/180
 - https://github.com/swift-server/async-http-client/pull/697

### Modifications:

- Add the `NIOSingletons` type
- Add `MultiThreadedEventLoopGroup.singleton`
- Add `NIOThreadPool.singleton`

### Result:

- Easier use of NIO that requires fewer parameters for users who don't require full control.
- Helps with #2142
- Fixes #2472
- Partially addresses #2473

### DIFF
--- a/Sources/NIOCore/GlobalSingletons.swift
+++ b/Sources/NIOCore/GlobalSingletons.swift
@@ -1,0 +1,180 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+#if canImport(Darwin)
+import Darwin
+#elseif os(Windows)
+import ucrt
+import WinSDK
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#else
+#error("Unsupported C library")
+#endif
+
+/// SwiftNIO provided singleton resources for programs & libraries that don't need full control over all operating
+/// system resources. This type holds sizing (how many loops/threads) suggestions.
+///
+/// Users who need very tight control about the exact threads and resources created may decide to set
+/// `NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false`. All singleton-creating facilities should check
+/// this setting and if `false` restrain from creating any global singleton resources. Please note that disabling the
+/// global singletons will lead to a crash if _any_ code attempts to use any of the singletons.
+public enum NIOGlobalSingletons {
+}
+
+extension NIOGlobalSingletons {
+    /// A suggestion of how many ``EventLoop``s the global singleton ``EventLoopGroup``s are supposed to consist of.
+    ///
+    /// The thread count is ``System/coreCount`` unless the environment variable `NIO_SINGLETON_GROUP_LOOP_COUNT`
+    /// is set or this value was set manually by the user.
+    ///
+    /// - note: This value must be set _before_ any singletons are used and must only be set once.
+    public static var suggestedGlobalSingletonGroupLoopCount: Int {
+        set {
+            Self.userSetSingletonThreadCount(rawStorage: globalRawSuggestedLoopCount, userValue: newValue)
+        }
+
+        get {
+            return Self.getTrustworthyThreadCount(rawStorage: globalRawSuggestedLoopCount,
+                                                  environmentVariable: "NIO_SINGLETON_GROUP_LOOP_COUNT")
+        }
+    }
+
+    /// A suggestion of how many threads the global singleton thread pools that can be used for synchronous, blocking
+    /// functions (such as ``NIOThreadPool``) are supposed to consist of
+    ///
+    /// The thread count is ``System/coreCount`` unless the environment variable
+    /// `NIO_SINGLETON_BLOCKING_POOL_THREAD_COUNT` is set or this value was set manually by the user.
+    ///
+    /// - note: This value must be set _before_ any singletons are used and must only be set once.
+    public static var suggestedBlockingPoolThreadCount: Int {
+        set {
+            Self.userSetSingletonThreadCount(rawStorage: globalRawSuggestedBlockingThreadCount, userValue: newValue)
+        }
+
+        get {
+            return Self.getTrustworthyThreadCount(rawStorage: globalRawSuggestedBlockingThreadCount,
+                                                  environmentVariable: "NIO_SINGLETON_BLOCKING_POOL_THREAD_COUNT")
+        }
+    }
+
+    /// A suggestion for whether the global singletons should be enabled. This is `true` unless changed by the user.
+    ///
+    /// This value cannot be changed using an environment variable.
+    ///
+    /// - note: This value must be set _before_ any singletons are used and must only be set once.
+    public static var globalSingletonsEnabledSuggestion: Bool {
+        get {
+            let (exchanged, original) = globalRawSingletonsEnabled.compareExchange(expected: 0,
+                                                                                   desired: 1,
+                                                                                   ordering: .relaxed)
+            if exchanged {
+                // Never been set, we're committing to the default (enabled).
+                assert(original == 0)
+                return true
+            } else {
+                // This has been set before, 1: enabled; -1 disabled.
+                assert(original != 0)
+                assert(original == -1 || original == 1)
+                return original > 0
+            }
+        }
+
+        set {
+            let intRepresentation = newValue ? 1 : -1
+            let (exchanged, original) = globalRawSingletonsEnabled.compareExchange(expected: 0,
+                                                                                   desired: intRepresentation,
+                                                                                   ordering: .relaxed)
+            guard exchanged else {
+                fatalError("""
+                           Bug in user code: Global singleton enabled suggestion has been changed after \
+                           user or has been changed more than once. Either is an error, you must set this value very \
+                           early and only once.
+                           """)
+            }
+        }
+    }
+}
+
+// DO NOT TOUCH THESE DIRECTLY, use `userSetSingletonThreadCount` and `getTrustworthyThreadCount`.
+private let globalRawSuggestedLoopCount = ManagedAtomic(0)
+private let globalRawSuggestedBlockingThreadCount = ManagedAtomic(0)
+private let globalRawSingletonsEnabled = ManagedAtomic(0)
+
+extension NIOGlobalSingletons {
+    private static func userSetSingletonThreadCount(rawStorage: ManagedAtomic<Int>, userValue: Int) {
+        precondition(userValue > 0, "illegal value: needs to be strictly positive")
+
+        // The user is trying to set it. We can only do this if the value is at 0 and we will set the
+        // negative value. So if the user wants `5`, we will set `-5`. Once it's used (set getter), it'll be upped
+        // to 5.
+        let (exchanged, _) = rawStorage.compareExchange(expected: 0, desired: -userValue, ordering: .relaxed)
+        guard exchanged else {
+            fatalError("""
+                       Bug in user code: Global singleton suggested loop/thread count has been changed after \
+                       user or has been changed more than once. Either is an error, you must set this value very early \
+                       and only once.
+                       """)
+        }
+    }
+
+    private static func validateTrustedThreadCount(_ threadCount: Int) {
+        assert(threadCount > 0,
+               "BUG IN NIO, please report: negative suggested loop/thread count: \(threadCount)")
+        assert(threadCount <= 1024,
+               "BUG IN NIO, please report: overly big suggested loop/thread count: \(threadCount)")
+    }
+
+    private static func getTrustworthyThreadCount(rawStorage: ManagedAtomic<Int>, environmentVariable: String) -> Int {
+        let returnedValueUnchecked: Int
+
+        let rawSuggestion = rawStorage.load(ordering: .relaxed)
+        switch rawSuggestion {
+        case 0: // == 0
+            // Not set by user, not yet finalised, let's try to get it from the env var and fall back to
+            // `System.coreCount`.
+            let envVarString = getenv(environmentVariable).map { String(cString: $0) }
+            returnedValueUnchecked = envVarString.flatMap(Int.init) ?? System.coreCount
+        case .min ..< 0: // < 0
+            // Untrusted and unchecked user value. Let's invert and then sanitise/check.
+            returnedValueUnchecked = -rawSuggestion
+        case 1 ... .max: // > 0
+            // Trustworthy value that has been evaluated and sanitised before.
+            let returnValue = rawSuggestion
+            Self.validateTrustedThreadCount(returnValue)
+            return returnValue
+        default:
+            // Unreachable
+            preconditionFailure()
+        }
+
+        // Can't have fewer than 1, don't want more than 1024.
+        let returnValue = max(1, min(1024, returnedValueUnchecked))
+        Self.validateTrustedThreadCount(returnValue)
+
+        // Store it for next time.
+        let (exchanged, _) = rawStorage.compareExchange(expected: rawSuggestion,
+                                                        desired: returnValue,
+                                                        ordering: .relaxed)
+        if !exchanged {
+            // We lost the race, this must mean it has been concurrently set correctly so we can safely recurse
+            // and try again.
+            return Self.getTrustworthyThreadCount(rawStorage: rawStorage, environmentVariable: environmentVariable)
+        }
+        return returnValue
+    }
+}
--- a/Sources/NIOCore/GlobalSingletons.swift
+++ b/Sources/NIOCore/GlobalSingletons.swift
@@ -30,20 +30,20 @@ import Musl
 /// system resources. This type holds sizing (how many loops/threads) suggestions.
 ///
 /// Users who need very tight control about the exact threads and resources created may decide to set
-/// `NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false`. All singleton-creating facilities should check
+/// `NIOSingletons.singletonsEnabledSuggestion = false`. All singleton-creating facilities should check
 /// this setting and if `false` restrain from creating any global singleton resources. Please note that disabling the
 /// global singletons will lead to a crash if _any_ code attempts to use any of the singletons.
-public enum NIOGlobalSingletons {
+public enum NIOSingletons {
 }
 
-extension NIOGlobalSingletons {
+extension NIOSingletons {
     /// A suggestion of how many ``EventLoop``s the global singleton ``EventLoopGroup``s are supposed to consist of.
     ///
     /// The thread count is ``System/coreCount`` unless the environment variable `NIO_SINGLETON_GROUP_LOOP_COUNT`
     /// is set or this value was set manually by the user.
     ///
     /// - note: This value must be set _before_ any singletons are used and must only be set once.
-    public static var suggestedGlobalSingletonGroupLoopCount: Int {
+    public static var groupLoopCountSuggestion: Int {
         set {
             Self.userSetSingletonThreadCount(rawStorage: globalRawSuggestedLoopCount, userValue: newValue)
         }
@@ -61,7 +61,7 @@ extension NIOGlobalSingletons {
     /// `NIO_SINGLETON_BLOCKING_POOL_THREAD_COUNT` is set or this value was set manually by the user.
     ///
     /// - note: This value must be set _before_ any singletons are used and must only be set once.
-    public static var suggestedBlockingPoolThreadCount: Int {
+    public static var blockingPoolThreadCountSuggestion: Int {
         set {
             Self.userSetSingletonThreadCount(rawStorage: globalRawSuggestedBlockingThreadCount, userValue: newValue)
         }
@@ -77,7 +77,7 @@ extension NIOGlobalSingletons {
     /// This value cannot be changed using an environment variable.
     ///
     /// - note: This value must be set _before_ any singletons are used and must only be set once.
-    public static var globalSingletonsEnabledSuggestion: Bool {
+    public static var singletonsEnabledSuggestion: Bool {
         get {
             let (exchanged, original) = globalRawSingletonsEnabled.compareExchange(expected: 0,
                                                                                    desired: 1,
@@ -96,9 +96,9 @@ extension NIOGlobalSingletons {
 
         set {
             let intRepresentation = newValue ? 1 : -1
-            let (exchanged, original) = globalRawSingletonsEnabled.compareExchange(expected: 0,
-                                                                                   desired: intRepresentation,
-                                                                                   ordering: .relaxed)
+            let (exchanged, _) = globalRawSingletonsEnabled.compareExchange(expected: 0,
+                                                                            desired: intRepresentation,
+                                                                            ordering: .relaxed)
             guard exchanged else {
                 fatalError("""
                            Bug in user code: Global singleton enabled suggestion has been changed after \
@@ -115,7 +115,7 @@ private let globalRawSuggestedLoopCount = ManagedAtomic(0)
 private let globalRawSuggestedBlockingThreadCount = ManagedAtomic(0)
 private let globalRawSingletonsEnabled = ManagedAtomic(0)
 
-extension NIOGlobalSingletons {
+extension NIOSingletons {
     private static func userSetSingletonThreadCount(rawStorage: ManagedAtomic<Int>, userValue: Int) {
         precondition(userValue > 0, "illegal value: needs to be strictly positive")
 
--- a/Sources/NIOCore/GlobalSingletons.swift
+++ b/Sources/NIOCore/GlobalSingletons.swift
@@ -55,7 +55,7 @@ extension NIOSingletons {
     }
 
     /// A suggestion of how many threads the global singleton thread pools that can be used for synchronous, blocking
-    /// functions (such as ``NIOThreadPool``) are supposed to consist of
+    /// functions (such as `NIOThreadPool`) are supposed to consist of
     ///
     /// The thread count is ``System/coreCount`` unless the environment variable
     /// `NIO_SINGLETON_BLOCKING_POOL_THREAD_COUNT` is set or this value was set manually by the user.
--- a/Sources/NIOCrashTester/CrashTests+EventLoop.swift
+++ b/Sources/NIOCrashTester/CrashTests+EventLoop.swift
@@ -105,5 +105,77 @@ struct EventLoopCrashTests {
             try! el.submit {}.wait()
         }
     }
+
+    let testUsingTheSingletonGroupWhenDisabled = CrashTest(
+        regex: #"Fatal error: Cannot create global singleton MultiThreadedEventLoopGroup because the global singletons"#
+    ) {
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
+        try? NIOGlobalSingletons.posixEventLoopGroup.next().submit {}.wait()
+    }
+
+    let testUsingTheSingletonBlockingPoolWhenDisabled = CrashTest(
+        regex: #"Fatal error: Cannot create global singleton NIOThreadPool because the global singletons have been"#
+    ) {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            try? group.syncShutdownGracefully()
+        }
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
+        try? NIOGlobalSingletons.posixBlockingThreadPool.runIfActive(eventLoop: group.next(), {}).wait()
+    }
+
+    let testDisablingGlobalSingletonsEnabledValueTwice = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton enabled suggestion has been changed after"#
+    ) {
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
+    }
+
+    let testEnablingGlobalSingletonsEnabledValueTwice = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton enabled suggestion has been changed after"#
+    ) {
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = true
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = true
+    }
+
+    let testEnablingThenDisablingGlobalSingletonsEnabledValue = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton enabled suggestion has been changed after"#
+    ) {
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = true
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
+    }
+
+    let testSettingTheGlobalSingletonEnabledValueAfterUse = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton enabled suggestion has been changed after"#
+    ) {
+        try? MultiThreadedEventLoopGroup.globalSingleton.next().submit({}).wait()
+        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = true
+    }
+
+    let testSettingTheSuggestedSingletonGroupCountTwice = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton suggested loop/thread count has been changed after"#
+    ) {
+        NIOGlobalSingletons.suggestedGlobalSingletonGroupLoopCount = 17
+        NIOGlobalSingletons.suggestedGlobalSingletonGroupLoopCount = 17
+    }
+
+    let testSettingTheSuggestedSingletonGroupChangeAfterUse = CrashTest(
+        regex: #"Fatal error: Bug in user code: Global singleton suggested loop/thread count has been changed after"#
+    ) {
+        try? MultiThreadedEventLoopGroup.globalSingleton.next().submit({}).wait()
+        NIOGlobalSingletons.suggestedGlobalSingletonGroupLoopCount = 17
+    }
+
+    let testSettingTheSuggestedSingletonGroupLoopCountToZero = CrashTest(
+        regex: #"Precondition failed: illegal value: needs to be strictly positive"#
+    ) {
+        NIOGlobalSingletons.suggestedGlobalSingletonGroupLoopCount = 0
+    }
+
+    let testSettingTheSuggestedSingletonGroupLoopCountToANegativeValue = CrashTest(
+        regex: #"Precondition failed: illegal value: needs to be strictly positive"#
+    ) {
+        NIOGlobalSingletons.suggestedGlobalSingletonGroupLoopCount = -1
+    }
 }
 #endif
--- a/Sources/NIOCrashTester/CrashTests+EventLoop.swift
+++ b/Sources/NIOCrashTester/CrashTests+EventLoop.swift
@@ -109,8 +109,8 @@ struct EventLoopCrashTests {
     let testUsingTheSingletonGroupWhenDisabled = CrashTest(
         regex: #"Fatal error: Cannot create global singleton MultiThreadedEventLoopGroup because the global singletons"#
     ) {
-        NIOGlobalSingletons.globalSingletonsEnabledSuggestion = false
-        try? NIOGlobalSingletons.posixEventLoopGroup.next().submit {}.wait()
+        NIOSingletons.singletonsEnabledSuggestion = false
+        try? NIOSingletons.posixEventLoopGroup.next().submit {}.wait()
     }
 
     let testUsingTheSingletonBlockingPoolWhenDisabled = CrashTest(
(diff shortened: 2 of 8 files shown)


Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,74 @@
...
+/// resources that applications (and some libraries) can use to obtain never-shut-down singleton resources.
+///
+/// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
+/// first use.
+public enum NIOSingletons {}
+
+extension NIOSingletons {
+    /// How many threads will be used for the multi-threaded `EventLoopGroup`s.
+    ///
+    /// Uses `NIO_SINGLETONS_MULTI_THREADED_ELG_THREAD_COUNT`.
```
glbrntt, line 44: Could you add a note making it explicit that this is an env variable and what the default is if it's not set?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,74 @@
...
+    public static var suggestedMultiThreadedEventLoopGroupThreadCount: Int {
+        return globalSuggestedMultiThreadedEventLoopGroupThreadCount
+    }
+
+    /// How many threads will be used for the blocking `NIOThreadPool`s.
+    ///
+    /// Uses `NIO_SINGLETONS_BLOCKING_POOL_THREAD_COUNT`.
```
glbrntt, line 51: Could you add a note making it explicit that this is an env variable and what the default is if it's not set?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,74 @@
...
+///
+/// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
+/// first use.
+public enum NIOSingletons {}
+
+extension NIOSingletons {
+    /// How many threads will be used for the multi-threaded `EventLoopGroup`s.
+    ///
+    /// Uses `NIO_SINGLETONS_MULTI_THREADED_ELG_THREAD_COUNT`.
+    public static var suggestedMultiThreadedEventLoopGroupThreadCount: Int {
```
glbrntt, line 45: It feels odd having this defined in `NIOSingletonResources` when MTELG is defined in `NIOPosix` -- is this used elsewhere too (e.g. for NIOTS)? If so we should consider a less specific name.

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,61 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_exported import NIOSingletonResources
+import NIOPosix
+import NIOCore
+
+extension NIOSingletons {
+    /// A globally shared, lazily intialized EventLoopGroup  that uses `epll` as the selector mechanism.
```
glbrntt, line 20:
```suggestion
    /// A globally shared, lazily initialized ``EventLoopGroup``  that uses `epoll` as the selector mechanism.
```

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,61 @@
...
+    public static var multiThreadedPosixEventLoopGroup: MultiThreadedEventLoopGroup {
+        return globalMultiThreadedPosixEventLoopGroup
+    }
+
+    /// A globally shared, lazily intialized EventLoopGroup with just one thread that uses `epoll` as the selector mechanism.
```
glbrntt, line 25:
```suggestion
    /// A globally shared, lazily initialized ``EventLoopGroup`` with just one thread that uses `epoll` as the selector mechanism.
```

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,61 @@
...
+    public static var singleThreadedPosixEventLoopGroup: MultiThreadedEventLoopGroup {
+        return globalSingleThreadedPosixEventLoopGroup
+    }
+
+    /// A globally shared, lazily intialized EventLoop that uses `epoll` as the selector mechanism. Backed by `NIOSingletons.singleThreadedPosixEventLoopGroup`.
```
glbrntt, line 30:
```suggestion
    /// A globally shared, lazily initialized ``EventLoop`` that uses `epoll` as the selector mechanism. Backed by ``NIOSingletons/singleThreadedPosixEventLoopGroup``.
```

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,61 @@
...
+    public static var posixEventLoop: EventLoop {
+        return Self.singleThreadedPosixEventLoopGroup.next()
+    }
+
+    /// A globally shared, lazily intialized `NIOThreadPool` that can be used for blocking I/O and other blocking operations.
```
glbrntt, line 35:
```suggestion
    /// A globally shared, lazily initialized ``NIOThreadPool`` that can be used for blocking I/O and other blocking operations.
```

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -138,25 +139,48 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
     /// - arguments:
     ///     - numberOfThreads: The number of `Threads` to use.
     public convenience init(numberOfThreads: Int) {
-        self.init(numberOfThreads: numberOfThreads, selectorFactory: NIOPosix.Selector<NIORegistration>.init)
+        self.init(numberOfThreads: numberOfThreads,
+                  canBeShutDown: true,
+                  selectorFactory: NIOPosix.Selector<NIORegistration>.init)
+    }
+
+    public convenience init(_canBeShutDown canBeShutDown: Bool, numberOfThreads: Int) {
```
glbrntt, line 147: Is this ever used with `_canBeShutDown: true`? A `static func` might be better here as you can underscore the method name (`init` with `_` never feels private enough to me)

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,61 @@
...
+private let globalSingleThreadedPosixEventLoopGroup: MultiThreadedEventLoopGroup = {
+    let group = MultiThreadedEventLoopGroup(_canBeShutDown: false, numberOfThreads: 1)
+    _ = Unmanaged.passUnretained(group).retain() // never gonna let you down.
+    return group
+}()
+
+private let globalPosixBlockingPool: NIOThreadPool = {
+    let pool = NIOThreadPool(numberOfThreads: NIOSingletons.suggestedBlockingPoolThreadCount)
+    pool.start()
+    _ = Unmanaged.passUnretained(pool).retain() // Make sure this is never deallocated (not strictly necessary).
```
glbrntt, line 57: Why do we do this if it's held by a global?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,74 @@
...
+import Darwin
+#elseif os(Windows)
+import ucrt
+import WinSDK
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#else
+#error("The concurrency NIOLock module was unable to identify your C library.")
```
simonjbeaumont, line 25: Copy paste error?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,74 @@
...
+///
+/// SwiftNIO allows and encourages the precise management of all operating system resources such as threads and file descriptors.
+/// Certain resources (such as the main `EventLoopGroup`) however are usually globally shared across the program. This means
+/// that many programs have to carry around an `EventLoopGroup` despite the fact the don't require the ability to fully return
+/// all the operating resources which would imply shutting down the `EventLoopGroup`. This type is the global handle for singleton
+/// resources that applications (and some libraries) can use to obtain never-shut-down singleton resources.
+///
+/// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
+/// first use.
+public enum NIOSingletons {}
```
glbrntt, line 39: Bikeshedding... I don't love the "singletons" in the name. I think "shared" or "global" express the purpose bit more.

Sources/NIOSingletonResources/SingletonType.swift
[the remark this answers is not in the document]
weissi: whoops fixed thanks

Sources/NIOSingletonResources/SingletonType.swift
weissi: Yes correct, that's used in NIOTS too and is suggested to be used in all multi-thread/queue/...'ed ELGs.

Sources/NIOSingletonResources/SingletonType.swift
weissi: I wanted to be 100% explicit that this is a singleton so I thought that's better. "Shared"/global is a little overloaded because people usually recommend to "share a global ELG" even if that's not this singleton.

Sources/NIOSingletonsPosix/PosixSingletons.swift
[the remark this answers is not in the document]
weissi: thanks fixed

Sources/NIOSingletonsPosix/PosixSingletons.swift
weissi: I'm not 100% sure if there's an actual language guarantee to not dealloc globals. Maybe after `main` or maybe if the compiler develops some ultra smarts to figure out you're not using it anymore.

Yes, not today but I thought let's better be safe than sorry, this isn't gonna get deinit'ed.

Sources/NIOSingletonResources/SingletonType.swift
weissi: done

Sources/NIOSingletonsPosix/PosixSingletons.swift
weissi: done thanks

Sources/NIOSingletonResources/SingletonType.swift
weissi: good idea, thanks, done

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
weissi: great idea, thanks, done

Sources/NIOSingletonResources/SingletonType.swift
weissi: c.f. https://github.com/apple/swift-nio-transport-services/pull/180/files#diff-d899aed582799046dcdaf6578f5f68ec07f205961e912528abb567f3413a4191R42

Sources/NIOSingletonResources/SingletonType.swift
FranzBusch: Adding to this: Why do we want to introduce new libraries for these singletons instead of adding them to the current libraries? Also from a discoverability standpoint wouldn't it be better for users if they could write something `MTELG.shared` instead of going through `NIOSingletons.multiThreadedPosixEventLoopGroup`

weissi: SwiftNIO is a low-level high-performance library that wants to give you control. Singletons are the antithesis of that so I really don't want to pollute `NIOCore`/`NIOPosix` with things that really go against the philosophy. I think it's a real feature that you need to let down your pants and say `import NIOSingletons`, that's `git grep`'able so you can easily check if things where the bodies are buried.

glbrntt: `shared`/`sharedInstance` are commonly used to refer to singletons, especially in Apple frameworks:
- https://developer.apple.com/documentation/swift/managing-a-shared-resource-using-a-singleton/
- https://developer.apple.com/search/?q=shared&type=Documentation
- https://developer.apple.com/search/?q=sharedInstance&type=Documentation

weissi: @glbrntt yeah, I know but GRPC/AHC and others use `.shared(...)` for "sharing another ELG" not the global singleton. So I thought it's confusing in NIO land.

I won't fight this but I think NIOSingletons is good because it says what it does. Sure, we can make it `NIOSharedResources` or something but that's hiding the fact that it's a singleton. Maybe we make the module `NIOSingletons*` and the types `NIOSharedResources` or smth?

glbrntt: > SwiftNIO is a low-level high-performance library that wants to give you control. Singletons are the antithesis of that so I really don't want to pollute `NIOCore`/`NIOPosix` with things that really go against the philosophy. 

I think it depends on how strongly we want to encourage users to reach for the singletons. If our position is "use them unless you know better" then I don't think we should hide them away in a separate module where they're hard to discover.

I like Franz's suggestion of putting them on the type which could be extended further:

```swift
extension EventLoopGroup where Self == MultiThreadedEventLoopGroup {
    public static var sharedMultiThreadedEventLoopGroup: MultiThreadedEventLoopGroup {
        MultiThreadedEventLoopGroup.shared
    }
}

let bootstrap = ClientBootstrap(group: .sharedMultiThreadedEventLoopGroup)
```

Sources/NIOSingletonResources/SingletonType.swift
glbrntt: Cool, can we name it something less MTELG-specific then? Maybe `suggestedEventLoopGroupSize`/`suggestedEventLoopGroupLoopCount`?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,75 @@
...
+
+extension NIOSingletons {
+    /// How many threads will be used for the multi-threaded ``EventLoopGroup``s.
+    ///
+    /// The thread count is ``System.coreCoung`` unless the environment variable `NIO_SINGLETONS_MULTI_THREADED_ELG_THREAD_COUNT`
```
glbrntt, line 44:
```suggestion
    /// The thread count is ``System/coreCount`` unless the environment variable `NIO_SINGLETONS_MULTI_THREADED_ELG_THREAD_COUNT`
```

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,75 @@
...
+        return globalSuggestedBlockingPoolThreadCount
+    }
+}
+
+private let globalSuggestedMultiThreadedEventLoopGroupThreadCount: Int = {
+    if let threadCountEnv = getenv("NIO_SINGLETONS_MULTI_THREADED_ELG_THREAD_COUNT"),
+        let threadCount = Int(String(cString: threadCountEnv)), threadCount > 0 {
+        return threadCount
+    } else {
+        return System.coreCount
```
FranzBusch, line 63: We should really land this PR https://github.com/apple/swift-nio/pull/2394. Otherwise we have the same problems in containers where we create a very wide ELG

weissi: Right yes, but they can go in in any order.

Sources/NIOSingletonResources/SingletonType.swift
weissi: So if we stick the extension into `NIOSingletonsPosix` then I think the extension is a good idea.

I like `MTELG.sharedMultiThreadedEventLoopGroup` and `MTELG.sharedSingleThreadedEventLoopGroup` but I think they should be in a singletons module.

Sources/NIOSingletonResources/SingletonType.swift
weissi: `loopCount` is great! "size" is a little unclear I think

weissi: fixed

Sources/NIOSingletonResources/SingletonType.swift
weissi: really a `/`? I changed that but that looks odd ;). Thanks for the typo, fixed that too ofc

Sources/NIOSingletonResources/SingletonType.swift
glbrntt: I think this comes back to discoverability. 

I'm not opposed to a separate module, I just think it needs to be easy for users to find. For the posix singletons we could re-export them from the `NIO` module. AHC and the like should depend on `NIOCore`/`NIOPosix` etc so that would solve that issue. One problem is that NIOTS can't re-export its singletons from NIO...

weissi: Right, I don't think _users_ should find it.

But let me be more precise:
- I would expect most users (say users of AHC / GRPC / Vapor / ...) to use async/await exclusively. Today, they do indeed need to think/create about ELGs but with this PR I'd hope to get the `import NIO*` totally out of their codebases (yes, it'll take a while until that's fully true)
- I would expect library developers (i.e. people who write AHC / GRPC / Vapor / ...) to know how to use NIO and I'd expect them to be somewhat deliberate about their ELG choices. But today they're in a tough spot: They need to put the ELG stuff into their API and the users need to fill it in. But I'd hope that this PR fixes that. If I made a new library FooClient, I think I would do `actor FooClient { init(eventLoopGroup: ELG = NIOSingletons.multiThreadedPosixGroup) { ... } )`. That way the library developer will need to `import NIOSingletons` in _one file_ and that's it. On the "inside" of their library they'll use `self.group` (and they shouldn't reach for the singletons).

Lukasa: I'm inclined to want us to add extensions to the appropriate types, as suggested by Franz and George.

Candidly, I don't think a singletons module is going to help us avoid the problems @weissi describes. It is not really possible to hit the combined goal of a) make it easy for novice users to just use a singleton ELG when it is appropriate and also b) make it really hard for library developers to misuse the singletons. We have tried (and failed) to communicate issues this subtle before, and I'd expect us to fail again. Where users _can_ pull an ELG out of thin air, they _will_.

With that tension in mind I think we get a better result by moving more in one direction than the other. And we're already all the way towards optimizing for avoiding misuse. So if we want to try something else, I'd favour optimizing hard for the novice use-case, and trusting experienced developers to deploy best practices. This is in line with Swift's overall pattern anyway. The cost to expert develoers of messing this up is basically only twofold: harder testing, and lower performance. Neither of these is a game-breaker.

weissi: But who do you expect to even use the singleton ELGs? Going forward I would not expect to find much NIO code in typical applications, I think they will just use a HTTP client/server, a GRPC client/server, ... All of these can default/have the option to use the singletons without a user ever typing either `import NIO` or `import NIOSingletons`, right?

Sources/NIOSingletonResources/SingletonType.swift
```diff
@@ -0,0 +1,75 @@
...
+    public static var suggestedMultiThreadedEventLoopGroupLoopCount: Int {
+        return globalSuggestedMultiThreadedEventLoopGroupLoopCount
+    }
+
+    /// How many threads will be used for the blocking `NIOThreadPool`s.
+    ///
+    /// The thread count is determined automatically unless the environment variable `NIO_SINGLETONS_BLOCKING_POOL_THREAD_COUNT` is set.
+    public static var suggestedBlockingPoolThreadCount: Int {
+        return globalSuggestedBlockingPoolThreadCount
+    }
```
Lukasa, line 55: I think if we're doing this, we should make these _settable_. Otherwise these numbers aren't very helpful as public API and we should drop them. But overriding their default behaviours is a reasonable thing to want to do at program startup.

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// The number of threads is determined by ``NIOSingletons.suggestedMultiThreadedEventLoopGroupThreadCount``.
+    public static var multiThreadedPosixEventLoopGroup: MultiThreadedEventLoopGroup {
+        return globalMultiThreadedPosixEventLoopGroup
+    }
+
+    /// A globally shared, lazily initialized ``MultiThreadedEventLoopGroup`` with just one thread that uses `epoll` as the selector mechanism.
+    public static var singleThreadedPosixEventLoopGroup: MultiThreadedEventLoopGroup {
+        return globalSingleThreadedPosixEventLoopGroup
+    }
```
Lukasa, line 30: I think we shouldn't offer both of these. If we're optimizing for non-expert users, we should make sensible choices for them: a user capable of answering the question "multi or single threaded" is also in a position to be able to find out how to arrange that behaviour.

Sources/NIOSingletonResources/SingletonType.swift
weissi: Okay, that's quite interesting. I think I like that.

So essentially
- What the user sets (assuming it's "early enough")
- If user hasn't set anything use what's in the `NIO_bla_blah_COUNT` env var
- If the env var isn't set fall back onto `System.coreCount`?

Sources/NIOSingletonsPosix/PosixSingletons.swift
Lukasa: Also this makes the Rick Astley callback work a little better because we have two stanzas instead of 3.

Sources/NIOSingletonResources/SingletonType.swift
Lukasa: Yup, that works for me.

Sources/NIOSingletonResources/SingletonType.swift
Lukasa: As an immediate piece of feedback I think the 80% answer is "the NIO core team", where the spelling is immaterial. It's not all that easy for the wider ecosystem to adopt these changes in a sprightly fashion. They probably won't. It'll take a long time for this to percolate out.

The broader answer is that library authors will use them, but the majority of library authors are in the long tail. SSWG authors are likely to have their use of ELGs called out in SSWG review. Everyone else in the long tail doesn't get that feedback, and so they'll do whatever seems most "obvious". I worry more about them than about the most plugged-in folks.

But to be clear, I'm happy with these living in a `NIOSingletons` module if you feel strongly. I think it gets extremely ugly extremely fast though, as each new EL type must define its own Singletons module. We end up with NIOSingletons, NIOTSSingletons, NIOWindowsSingletons, etc. I'm not sure that's making it easy to DTRT.

Sources/NIOSingletonsPosix/PosixSingletons.swift
weissi: Right, I was thinking about this too. Yes, one option gives us a nicer API and we can likely recommend people to just go single-threaded.

What made choose to offer both is that it's entirely reasonable to start off single threaded and later decide or evaluate (not enough perf or so) if more threads would help. I liked that that experiment is then simply a single `s/single/multi/` and done.

If we only offered a single-threaded singleton then the step to trying a multi-threaded one would be quite steep because suddenly you need your own singletons (or pass around from main) and what not.

Of course we could just default to multi threaded (with a settable thread count if people really want to reduce it to 1) and not offer a single-threaded singleton ELG. I'd be fine with that.

The last option is to have just one singleton ELG that defaults to 1 thread but can be configured to have more than one thread (via env var / early-enough setting). The issue with that is that people will only test on the single-threaded one and suddenly a user changes some environment variable and now we're multi threaded which might break stuff as it's a config that has never run...

weissi: > Also this makes the Rick Astley callback work a little better because we have two stanzas instead of 3.

We could also look up the next stanza ;p

Sources/NIOSingletonResources/SingletonType.swift
gwynne: I have to side with @weissi in terms of naming - "shared" means too many other things, none of which (as inaccurate as some of those usages may be) would be correct for this.

Sources/NIOSingletonsPosix/PosixSingletons.swift
gwynne: You know the rules, and so do I...

Sources/NIOSingletonResources/SingletonType.swift
fabianfett: Do we want to fallback to `System.coreCount`? I would have expected to fallback to just `1`?

Sources/NIOSingletonResources/SingletonType.swift
weissi: > The broader answer is that library authors will use them, but the majority of library authors are in the long tail. SSWG authors are likely to have their use of ELGs called out in SSWG review. Everyone else in the long tail doesn't get that feedback, and so they'll do whatever seems most "obvious".

Agree with all that. Do you think the most obvious would be to not `import NIOSingletonsPosix` and instead create their own MTELG and leak it?

> But to be clear, I'm happy with these living in a NIOSingletons module if you feel strongly. I think it gets extremely ugly extremely fast though, as each new EL type must define its own Singletons module. We end up with NIOSingletons, NIOTSSingletons, NIOWindowsSingletons, etc. I'm not sure that's making it easy to DTRT.

I don't feel strongly about the particular name `NIOSingletons` vs. `NIOSharedResources` or smth. But I do prefer a separate module. Of course, you correctly point out that we'll end up with one `NIOSingletonsELType` for every EventLoop type. I didn't think that's a problem given that we have 2 today and in the future it'll be 3 or 4 I reckon. But it still is a good point.

If we all feel that having it in `NIO` (more precisely `NIOPosix` but nobody that we're discussing here uses that) then I'm happy to move it over. When I set out to start the PR I was very much not thinking of 'polluting' NIO with singleton badness but I can obviously see where you're coming from.

weissi: Separate module for the singletons (`import NIOSingletonsPosix` or similar required to get hold of them)

Sources/NIOSingletonsPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,66 @@
...
+
+    /// A globally shared, lazily initialized ``EventLoop`` that uses `epoll` as the selector mechanism. Backed by `NIOSingletons.singleThreadedPosixEventLoopGroup`.
+    public static var posixEventLoop: EventLoop {
+        return Self.singleThreadedPosixEventLoopGroup.next()
+    }
+
+    /// A globally shared, lazily initialized ``NIOThreadPool`` that can be used for blocking I/O and other blocking operations.
+    ///
+    ///     /// The number of threads is determined by ``NIOSingletons.suggestedBlockingPoolThreadCount``.
+    public static var posixBlockingPool: NIOThreadPool {
```
adam-fowler, line 40: `posixBlockingPool` doesn't mean much to me. Can we name this something different, `posixBlockingThreadPool` maybe? I'm guessing `posixThreadPool` is out of the question

adam-fowler (commented):
For libraries it would be great if we can replace the NIOEventLoopGroupProvider with an EventLoopGroup that defaults to one of the global ones. Then not have to worry about EventLoopGroup shutdown at all.

Sources/NIOSingletonsPosix/PosixSingletons.swift
Lukasa: I think it's extremely unclear to a library developer whether 1 or 1+ is appropriate. At the very least it isn't clear to me. What's appropriate for grpc-swift or AHC?

More importantly, the risk we have here is of multiplicative singletons. In your proposal we already have four:

1. NIOPosix.singleThreaded
2. NIOPosix.multiThreaded
3. NIOTS.singleThreaded
4. NIOTS.multiThreaded

Presuming a world where we have IOCP support we have two more, for 6.

I have no idea how a library developer is expected to pick which of these they should use. Even if we offered a "best available", we still have to offer "best available multithreaded" and "best available singlethreaded", and nothing prevents a developer from picking a different one. We rapidly trip into a world where we've created 2x core count threads for I/O and we haven't achieved a substantial simplification.

If we want NIO to have an easy mode, let's give it an easy mode. No choices. Users who aren't happy with that are still free to create ELGs, they are one line of code to create.

Lukasa: Put another way, "consider how much you have in the ChannelPipeline" is not easy mode.

weissi: Yes, it's not always easy. If you really think we should have just one then it has to be the multi-threaded.

I think it would be truly awful to give users one that's single-threaded by default (i.e. single threaded in all their tests) but can be turned into a multi-threaded ones by user config.

Given that a multi-threaded group doesn't cause harm it's gotta be the default if we only do one. I thought it might be nice to _additionally_ provide a single-threaded one in cases where you know that one thread will be enough. Happy for that affordance to go away especially given that you can do `singletonGroup.next()` to get  a single-threaded one.

weissi: Single multi-threaded works for me.

Lukasa: Fine by me.

Sources/NIOSingletonResources/SingletonType.swift
weissi: Okay, to break the stalemate here, why don't we follow what NIO does already:

- `import NIO` gives you everything -- right for most
- `import NIOCore/Posix/...` gives you only what you need

So `NIO` would `@_exported import NIOSingletonsPosix`.

---

Of course, this still suffers from the issue that `import NIO` potentially gives you access to multiple groups: The "correct" one that's part of your framework as well as the singleton one. Developers will be confused which one to choose and will likely start to default to the singleton ones even in libraries that have a group (like gRPC/AHC/most other libs). IMHO that's bad and unfortunate but most of you disagree so fine.

FranzBusch: I don't see a big advantage in having a separate target in that case then. For users that just import `NIO` right now they presumably don't care much about what is in our various targets and just want to get it building. Presumably this camp also doesn't have to touch ELGs much or shouldn't once we get them to move over to Concurrency more and more.

On the other hand, where we currently have targeted imports is in most of our libraries and I don't see why we would need to import another target just to get the singleton queues. I would argue that developers who use our targeted imports don't need to import another module just to access the singletons.

weissi: I don't like it, I think it's the right move (singletons are bad for a reason) but I need the functionality.

So whatever, I put them into `NIOPosix`. Added `MultiThreadedEventLoopGroup.globalSingleton` and `NIOThreadPool.globalSingleton`.

Sources/NIOSingletonResources/SingletonType.swift
weissi: @Lukasa Thanks, that's now implemented

Sources/NIOSingletonsPosix/PosixSingletons.swift
weissi: Done, implemented

Sources/NIOCore/GlobalSingletons.swift
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static func validateTrustedThreadCount(_ threadCount: Int) {
+        assert(threadCount > 0,
+               "BUG IN NIO, please report: negative suggested loop/thread count: \(threadCount)")
+        assert(threadCount <= 1024,
+               "BUG IN NIO, please report: overly big suggested loop/thread count: \(threadCount)")
+    }
+
+    private static func getTrustworthyThreadCount(rawStorage: ManagedAtomic<Int>, environmentVariable: String) -> Int {
```
FranzBusch, line 95: I lowkey love how we solved this!

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -138,29 +140,73 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
...
-        self.init(numberOfThreads: numberOfThreads, selectorFactory: NIOPosix.Selector<NIORegistration>.init)
+        self.init(numberOfThreads: numberOfThreads,
+                  canBeShutDown: true,
+                  selectorFactory: NIOPosix.Selector<NIORegistration>.init)
+    }
+
+    /// Create a ``MultiThreadedEventLoopGroup`` that cannot be shut down and must not be `deinit`ed.
+    ///
+    /// This is only useful for global singletons.
+    public static func _makePerpetualGroup(threadNamePrefix: String,
```
FranzBusch, line 151: Does this need to be `public` now that we have it in the same module?

Sources/NIOPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,87 @@
...
+        return globalSingletonMTELG
+    }
+
+    /// A globally shared, lazily initialized ``NIOThreadPool`` that can be used for blocking I/O and other blocking operations.
+    ///
+    ///     /// The number of threads is determined by ``NIOSingletons.suggestedBlockingPoolThreadCount``.
+    public static var posixBlockingThreadPool: NIOThreadPool {
+        return globalPosixBlockingPool
+    }
+}
```
FranzBusch, line 32: Do we still see value in having this extension? Why would users choose to do

```
NIOGlobalSingletons.posixEventLoopGroup
```

over

```
MultiThreadedEventLoopGroup.globalSingleton
```

IMO we can just have the latter and we could rename the `NIOGlobalSingletons` to `NIOGlobalSingletonsConfiguration`

Sources/NIOPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// SwiftNIO allows and encourages the precise management of all operating system resources such as threads and file descriptors.
+    /// Certain resources (such as the main ``EventLoopGroup``) however are usually globally shared across the program. This means
+    /// that many programs have to carry around an ``EventLoopGroup`` despite the fact the don't require the ability to fully return
+    /// all the operating resources which would imply shutting down the ``EventLoopGroup``. This type is the global handle for singleton
+    /// resources that applications (and some libraries) can use to obtain never-shut-down singleton resources.
+    ///
+    /// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
+    /// first use.
+    ///
+    /// The loop count of this group is determined by ``NIOGlobalSingletons/suggestedGlobalSingletonGroupLoopCount``.
```
FranzBusch, line 46: NIT:
```suggestion
    /// A globally shared, singleton ``MultiThreadedEventLoopGroup``.
    ///
    /// SwiftNIO allows and encourages the precise management of all operating system resources such as threads and file descriptors.
    /// Certain resources (such as the main ``EventLoopGroup``) however are usually globally shared across the program. This means
    /// that many programs have to carry around an ``EventLoopGroup`` despite the fact they don't require the ability to fully return
    /// all the operating resources which would imply shutting down the ``EventLoopGroup``. This type is the global handle for singleton
    /// resources that applications (and some libraries) can use to obtain never-shut-down singleton resources.
    ///
    /// Programs and libraries that do not use these singletons will not incur extra resource usage, these resources are lazily initialized on
    /// first use.
    ///
    /// The loop count of this group is determined by ``NIOGlobalSingletons/suggestedGlobalSingletonGroupLoopCount``.
```

Sources/NIOPosix/PosixSingletons.swift
```diff
@@ -0,0 +1,87 @@
...
+        return NIOGlobalSingletons.posixEventLoopGroup
+    }
+}
+
+extension NIOThreadPool {
+    /// A globally shared, singleton ``NIOThreadPool``.
+    ///
+    /// SwiftNIO allows and encourages the precise management of all operating system resources such as threads and file descriptors.
+    /// Certain resources (such as the main ``NIOThreadPool``) however are usually globally shared across the program. This means
+    /// that many programs have to carry around an ``NIOThreadPool`` despite the fact the don't require the ability to fully return
```
FranzBusch, line 57:
```suggestion
    /// that many programs have to carry around an ``NIOThreadPool`` despite the fact they don't require the ability to fully return
```

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
weissi: It doesn't _need_ to be but I wanted to keep it as "SPI" to allow other large projects to make their customised singletons too if they already have an existing, similar solution.

Sources/NIOPosix/PosixSingletons.swift
weissi: I like having them in one place but of course it's not strictly necessary. I thought about renaming it to something like `Configuration` or `Suggestions` or so. But then I thought it's a good place for other extensions to hang their more specialised stuff off of.

For example, if we did change our mind and thought there should be a single threaded one too. Or maybe the first thread of the shared multi group is specially exposed as `NIOSingletons.posixSingleEventLoop` or something then we have a place in the API. If it's just "configuration" or "suggestions" or so, then that becomes awkard.

weissi: tl;dr I think it gives us unexpensive/unoffensive design space. Users ain't gonna look there and we don't love singletons, let's give them their naughty corner but in there they have free reign

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
glbrntt: Will this _actually_ be used? Or is it maybe we'd use this in the future? If it's the former then we should just make it API (should it be "unsafe" too?). If it's the latter then let's keep it internal until it's needed.

Sources/NIOPosix/PosixSingletons.swift
glbrntt: Can we have an extension on `EventLoopGroup` à la https://github.com/apple/swift-nio/pull/2471#discussion_r1262318001 as well?

weissi: @glbrntt Not quite sure what that buys us:

- For functions that take `MultiThreadedEventLoopGroup` today, that already works
- For functions that take `any EventLoopGroup`, we can only have _one_ `globalSingleton` and I'm not sure it's right to reserve that for `MultiThreadedEventLoopGroup`.

For the second point, let's say we have

```
extension EventLoopGroup where Self == MultiThreadedEventLoopGroup {
    public static var globalSingleton: MultiThreadedEventLoopGroup {
        return ...
    }
}

extension EventLoopGroup where Self == NIOTSEventLoopGroup {
    public static var globalSingleton: NIOTSEventLoopGroup {
        return ...
    }
}
```

if you were then to use `ClientBootstrap(group: .globalSingleton)` you'd get an ambiguity warning.

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
glbrntt: Only (minor) concern was that if it's public API and you hold it wrong then you can hit an assertion failure. 

If it remains underscored then I don't mind at all. Presumably you're fine with the caveats of it not being public API.

Sources/NIOPosix/PosixSingletons.swift
glbrntt: It buys usability. Users don't need to know the type of the group, they can `.`-tab-complete for `any EventLoopGroup`. It's fine having to spell out the type on the `EventLoopGroup` extension: 

```swift
extension EventLoopGroup where Self == MultiThreadedEventLoopGroup {
    public static var globalMultiThreadedEventLoopGroup: MultiThreadedEventLoopGroup { // ... }
}

extension EventLoopGroup where Self == NIOTSEventLoopGroup {
    public static var globalNIOTSEventLoopGroup: NIOTSEventLoopGroup { // ... }
}
```

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
glbrntt: fine by me!

Sources/NIOPosix/PosixSingletons.swift
weissi: AAhh, sorry, misread that. Yes, I'm happy if we use distinct names like `globalMultiThreadedEventLoopGRoup`.

weissi: @glbrntt added

Sources/NIOSingletonsPosix/PosixSingletons.swift
weissi: @adam-fowler The default way is now `NIOThreadPool.globalSingleton` which I suppose is clearer?

Sources/NIOCore/GlobalSingletons.swift
```diff
@@ -0,0 +1,180 @@
...
+                                                  environmentVariable: "NIO_SINGLETON_BLOCKING_POOL_THREAD_COUNT")
+        }
+    }
+
+    /// A suggestion for whether the global singletons should be enabled. This is `true` unless changed by the user.
+    ///
+    /// This value cannot be changed using an environment variable.
+    ///
+    /// - note: This value must be set _before_ any singletons are used and must only be set once.
+    public static var globalSingletonsEnabledSuggestion: Bool {
```
glbrntt, line 80: Is this effectively a kill-switch for singletons? I.e. set this and blow up if singletons are used?

glbrntt (commented):
Mostly docc nits but otherwise this is looking good. Still not a big fan of doubling up global _and_ singleton though but not a hill I'll die on.

weissi:
> For libraries it would be great if we can replace the NIOEventLoopGroupProvider with an EventLoopGroup that defaults to one of the global ones. Then not have to worry about EventLoopGroup shutdown at all.

@adam-fowler Yes, `NIOEventLoopGroupProvider` was always a bad idea. See #2480 for the deprecation of it.

Sources/NIOCrashTester/CrashTests+EventLoop.swift
```diff
@@ -105,5 +105,77 @@ struct EventLoopCrashTests {
             try! el.submit {}.wait()
         }
     }
+
+    let testUsingTheSingletonGroupWhenDisabled = CrashTest(
```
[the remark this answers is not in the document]
weissi: 💥

Sources/NIOCore/GlobalSingletons.swift
weissi: @glbrntt It's a suggestion because we can't _actually_ control that no code's creating singletons... We can make NIO code behave but we don't actually know :)

weissi: kill switch: Yes, I think this is important to test stuff that thinks/hopes/promises to correctly wire through the `group`. Without the kill switch it's super duper hard to test (need to check the thread count etc)

weissi:
> Mostly docc nits but otherwise this is looking good. Still not a big fan of doubling up global _and_ singleton though but not a hill I'll die on.

Just `NIOSingletons` and `MultiThreadedEventLoopGroup.singleton` ?

glbrntt:
> > Mostly docc nits but otherwise this is looking good. Still not a big fan of doubling up global _and_ singleton though but not a hill I'll die on.
> 
> Just `NIOSingletons` and `MultiThreadedEventLoopGroup.singleton` ? I'm happy with that FWIW.
> 
> I just don't think we should do literally `.shared` because of the EventLoopGroupProvider `.shared(...)` and `.shared(.shared)` is kinda weird :P. `.shared(...singleton)` does make sense.

Yeah I'm okay with that.

Sources/NIOCore/GlobalSingletons.swift
```diff
@@ -0,0 +1,180 @@
...
+}
+
+extension NIOGlobalSingletons {
+    /// A suggestion of how many ``EventLoop``s the global singleton ``EventLoopGroup``s are supposed to consist of.
+    ///
+    /// The thread count is ``System/coreCount`` unless the environment variable `NIO_SINGLETON_GROUP_LOOP_COUNT`
+    /// is set or this value was set manually by the user.
+    ///
+    /// - note: This value must be set _before_ any singletons are used and must only be set once.
+    public static var suggestedGlobalSingletonGroupLoopCount: Int {
```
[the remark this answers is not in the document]
weissi: done

glbrntt (approved):
LGTM modulo CI failures. I think this ended up in a pretty good shape 🙂

FranzBusch (approved):
LGTM!

weissi:

weissi:
ping @Lukasa, I didn't see that you actually hadn't approved this yet. Happy to make changes if there's anything. Hasn't been released yet so we've got the freedom.

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
natikgadzhi: `@_spi(WD40)` scuse me just passing by.

PR #2474, opened 2023-07-13
WIP: Reduce the amount of bootstrap async methods

FranzBusch:
# Motivation
When adding new async methods for all the bootstrap we had to create 3 sets of method for every bootstrap. This resulted in lots of code and only provided little benefit for users.

# Modification
This PR removes all bind/connect methods except the most generic ones. This leaves it up to the user to wrap their channels in a `NIOAsyncChannel` or to await the protocol negotiation result.

# Result
Less code to maintain on our side.


Tests/NIOPosixTests/AsyncChannelBootstrapTests.swift
```diff
@@ -278,9 +280,9 @@ final class AsyncChannelBootstrapTests: XCTestCase {
 
             group.addTask {
                 try await withThrowingTaskGroup(of: Void.self) { group in
-                    for try await childChannel in channel.inboundStream {
+                    for try await negotiationResult in channel.inboundStream {
                         group.addTask {
-                            switch childChannel {
+                            switch try await negotiationResult.resolve() {
```
FranzBusch, line 285: This part here is the important change. Instead of having a `NIOAsyncChannel` of negotiation results we now have a channel of futures of negotiation results. This means that users have to resolve the result and can then decide what to do with it. It doesn't look too bad from a users perspective and allows us to significantly reduce the amount of code in NIO. cc @Lukasa @glbrntt @rnro

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -353,6 +353,36 @@ public enum NIOProtocolNegotiationResult<NegotiationResult: Sendable> {
...
+
+extension EventLoopFuture {
+    @available(macOS 10.15, *)
+    @_spi(AsyncChannel)
+    /// Resolves the protocol negotiation result.
+    ///
+    /// This method waits for the final protocol negotiation result and the returns the value.
+    public func resolve<NegotiationResult: Sendable>() async throws -> NegotiationResult where Value == NIOProtocolNegotiationResult<NegotiationResult> {
+        try await self.flatMap { $0 .resolve(on: self.eventLoop) }.get()
+    }
```
FranzBusch, line 383: I am not 100% sold on this yet. It is pretty hard to discover. We could instead introduce a new like `NIOUnresolvedProtocolNegotiationResult` which would hold this method and make the `NIOProtocolNegotiationHandler` require to return that instead of the current `EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>`

FranzBusch:
- 5.9 is a known allocation regression
- Nightly is a known crash
- API breakage is also expected

@Lukasa Could you merge this over the failing API breakage CI please

PR #2475, opened 2023-07-14
Adding ByteBuffer.hexDump in xxd and hexdump -C formats

### LINKED ISSUE #2447: ByteBuffer should have a bunch of hex dump helpers
I frequently need at least the following kinds of functionality, no idea how many times I've implemented them in random projects:

1. hex dumps of a `ByteBuffer` one (long) line `xxd -r -p` compatible (example `42 41 42 41`)
2. same as 1 but length-limited (as in produce xxd compatible output but only up to 1024 bytes, if more, keep beginning and end and replace middle with [...] or so)
3. produce multi-line `hexdump -C` compatible output (in full)
4. product multi-line `hexdump -C`-compatible output but length limited

and probably there are others. This is mostly an API-design problem I think because if we're smart we can probably fit this all into one method like `buffer.hexDump(renderer: .xxdCompatible(maxLength: 1024))` or something.

natikgadzhi:
This PR adds a bunch of `hexDump` functions to `ByteBuffer`. 

Closes #2447 

### Motivation:

@weissi says he ends up writing hex dump helper functions in most projects with NIO, so they would be useful to have straight out of the box.

### Modifications:

- [x] Adds a `String(byte: UInt, padding: Int)` initializer to format bytes.
- [x] Changes the output format of `ByteBuffer.Storage.dumpBytes` and re-uses it.
- [x] Adds `hexDump` implementations for different formats.
- [x] Tests for hexDump functions that this PR adds.

### TODO
- [x] Rework reading data from the buffer in long-format dump and avoid using `get*`.
- [x] Add test cases for situations where reader and writer indices are in non-default states.
- [x] Add a test case for a single line output in long-format dump.
- [x] Adds `ByteBuffer.HexDumpFormat` enum and a public function to get a hex dump of a `ByteBuffer` providing the desired format.
- [x] Documentation for public hexDump functions.
- [x] Add `hexDumpLong(maxLength:)` implementation for long-format clipped hex dumps
- [x] Cleanup the code and wire up the clipped length dump in the public `hexDump()` function
- [x] Perhaps more tests?

### Result:

We'll have _one_ new public function on `ByteBuffer` that accepts desired `HexDumpFormat` and returns the formatted hex dump string.


Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -675,7 +673,7 @@ public struct ByteBuffer {
     }
 
     /// Yields a buffer pointer containing this `ByteBuffer`'s readable bytes. You may hold a pointer to those bytes
-    /// even after the closure returned iff you model the lifetime of those bytes correctly using the `Unmanaged`
```
weissi, line 678: "iff" means "if and only if" and is correct here

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,93 @@ extension ByteBuffer {
...
+    }
+
+    /// Returns a `String` of hexadecimal digits of bytes in the Buffer,
+    /// with formatting compatible with output of `hexdump -C`.
+    func hexDumpWithASCII() -> String {
+        var result = ""
+
+        // hexdump -C dumps into lines of upto 16 bytes each
+        // with a last line containing only the offset column.
+        for lineIdx in 0 ... self.readableBytes/16 {
```
weissi, line 1165: if you call `get*` later, your indices should be between `self.readerIndex ..< self.writerIndex` not starting at 0.

You can easily generate a test case that will have a higher than 0 readerIndex by doing:

```swift
var buffer = ByteBuffer(string: "XXXXXHello World")
let xxxxx = buffer.readString(length: 5)!
XCTAssertEqual("XXXXX", xxxxx)
XCTAssertEqual(5, buffer.readerIndex)
XCTAssertNil(buffer.getInteger(at: 0, as: UInt8.self)) // <-- nil becasuse 0 is not a readable index now
```

Sources/NIOCore/ByteBuffer-core.swift
natikgadzhi: 🤦🏼 TIL, thank you!

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,105 @@ extension ByteBuffer {
...
+extension ByteBuffer {
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer,
+    /// in a format that's compatible with `xxd -r -p`.
+    ///
+    /// - parameters:
+    ///     - offset:  The offset from the beginning of the buffer, in bytes. Defaults to 0.
+    ///     - readableOnly: Whether to only dump the `readableBytes` part (before the `writerIndex`), or to dump the full buffer, including empty bytes in the end of the allocated space. Defaults to true.
+    func hexDumpShort(offset: Int = 0, readableOnly: Bool = true) -> String {
+        let length = (readableOnly ? self.writerIndex : self.capacity) - offset
```
natikgadzhi, line 1132: The actual `dumpBytes` uses unsafe pointer and can dump any indices, so the dump will start at `self.bytes[offset]`.

`readableOnly` is an attempt to provide a convenience API to only dump the meaningful bytes that have been previously written, hence using `writerIndex`. 

Unlike `readableBytes`, even if the buffer was previously read from, and `readerIndex` moved, `writerIndex` is more likely to be useful.

Pros:
- Perhaps it's easier to dump the part of the buffer that was previously written to, less cleanup to do.

Cons:
- Naming is not yet final, and I think it does not make much sense. `readableOnly` implies dumping readable bytes, i.e. between `readerIndex..<writerIndex`, which is NOT what this is doing. 
- Unreliable. Someone could `setWriterIndex()` before calling this, and miss some parts of the buffer bytes that were useful to them.


I think if we keep this in the final version, this needs to be documented. Or we can remove that `readableOnly` option altogether, and just dump all buffer bytes to capacity, or accept `length` as an argument, default to full capacity. 

/cc @weissi, @glbrntt

natikgadzhi (commented):
I've cleaned up the short hexdump helpers for now, but will work on the long-format more. Not yet ready for a thorough review.

Sources/NIOCore/ByteBuffer-core.swift
natikgadzhi: @weissi, okay, I think I got the workaround for `readableBytes` — using the same mechanism as the `Storage.dumpBytes`. https://github.com/apple/swift-nio/pull/2475/files#diff-3e3180bab09c17b6054fc18892bb5ee1cd211963a7a2d42265cb905269b98240R1178

Sources/NIOCore/ByteBuffer-core.swift
weissi: @natikgadzhi the `readableOnly` parameter needs to go. Nothing must _ever_ access non-readable bytes, that'd be undefined behaviour, might crash the program and even worse things could happen. The non-readable bytes must not be accessible through any (non-unsafe) API, including this one.

Said differently, accessing non-redable bytes is like purposefully reading past the end of a pointer. Something like

```
void *ptr = malloc(10); // get 10 bytes
memset(ptr, 'A', 10); // write 10 As
printf("some unreadable byte: %c"\n", ptr + 100); // read 100 bytes past the pointer (which holds 10 allocated bytes only)
```

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,100 @@ extension ByteBuffer {
...
+        return self._storage.dumpBytes(slice: self._storage.fullSlice,
```
weissi, line 1133: - would you mind moving the byte dumping out of `ByteBuffer-core` into `ByteBuffer-aux` as it's not "core" functionality
- the hex dumping functionality should use `ByteBuffer`'s _public_ API. I.e. only things that can be accessed from the outside and especially no `self._*` things

weissi: Honestly, I would delete (or leave alone) the old `self._storage.dumpBytes()`. That was written to debug the underlying storage itself when we're replacing the storage. So essentially it's an internal debugging facility that we can delete.

This new hex dumping facility probably need no APIs but

- `self.withUnsafeReadableBytes { ... }` <- pointer to all readable bytes
- `self.readableBytesView` <- a `Collection` of all readable bytes
- `self.getInteger(...)` <- a third way to access the bytes
- `self.readerIndex`/`self.writerIndex`/`self.redableBytes` <- how many/which bytes are readable

all of them are public.

Sources/NIOCore/ByteBuffer-core.swift
natikgadzhi: First off, I'm so grateful for your help and guidance! Coming from the world of web requests and Ruby here, and have a few things to learn. I realize that the whole thing would probably take you under an hour to implement yourself, so I appreciate your time that you spend on coaching me, @weissi. Thank you! 👏🏼 

I'll reply to the conversation threads and try to capture your feedback and work through it.

> @natikgadzhi the readableOnly parameter needs to go. Nothing must ever access non-readable bytes, that'd be undefined behaviour, might crash the program and even worse things could happen. The non-readable bytes must not be accessible through any (non-unsafe) API, including this one.

I take a few things from this: 
- [ ] Remove `readableOnly` as a concept. We will only dump the readable bytes portion.
- [ ] As a rule, only use public API for reading data. Also, dumping is not as complex as to need unsafe access, it feels like. 
- [ ] Agree on what gets dumped, and what is a good public API before building. I want to clarify that: 
   - `hexDump()` should dump readable bytes, which means from `readerIndex` through `writerIndex`. 
   - If we choose to accept `offset` parameter, document that it offsets from the current `readerIndex`, similar to how all `get*(at: Int)` functions currently work.
   - Clarify that it's the user's responsibility to set the `readerIndex` to the desired location before dumping.
   - Clarify that `hexDump` should not move the readerIndex. That is an open questin to me. I _think_ dumping should not change the index (it can just `get*` and that'll do the trick), but I want to confirm that with you. /cc @weissi. 


As for unsafe access for the full buffer allocated length — yep, thank you for the note! I thought that as long as it's within buffer's `capacity`, it would be safe to access and dump, but hey, I'm a newbie here.

Sources/NIOCore/ByteBuffer-core.swift
natikgadzhi: > Honestly, I would delete (or leave alone) the old self._storage.dumpBytes().
Sure. Since hex dump functions will cover similar functionality (but at a higher level), I'll delete the old `dumpBytes` from Storage and clean up the places where we use it to make sure i don't break anything.

> Even better: make a new file for it ByteBuffer+HexDumps.swift
EZ. I thought of moving them, and just started in the `core` file to quickly prototype. 

> he hex dumping functionality should use ByteBuffer's public API. I.e. only things that can be accessed from the outside and especially no self._* things

Yep, got it. 

> self.withUnsafeReadableBytes { ... } <- pointer to all readable bytes

This will be the easiest rewrite from what's currently in the pull request. 

At first, I started out using `readableBytes`, but I used it in a way that would introduce a bug, and I only read a single byte with `getInteger` because I didn't know if reading the whole buffer would make a copy of the data (and hence did not understand how memory efficient that would be).

Let me go clean this up.

Sources/NIOCore/ByteBuffer-core.swift
glbrntt: > I take a few things from this:
> 
> * [ ]  Remove `readableOnly` as a concept. We will only dump the readable bytes portion.
> * [ ]  As a rule, only use public API for reading data. Also, dumping is not as complex as to need unsafe access, it feels like.

That's right

> * [ ]  Agree on what gets dumped, and what is a good public API before building. I want to clarify that:
>   
>   * `hexDump()` should dump readable bytes, which means from `readerIndex` through `writerIndex`.
>   * If we choose to accept `offset` parameter, document that it offsets from the current `readerIndex`, similar to how all `get*(at: Int)` functions currently work.
>   * Clarify that it's the user's responsibility to set the `readerIndex` to the desired location before dumping.
>   * Clarify that `hexDump` should not move the readerIndex. That is an open questin to me. I _think_ dumping should not change the index (it can just `get*` and that'll do the trick), but I want to confirm that with you. /cc @weissi.

Methods for reading from a `ByteBuffer` are usually implemented as `get`/`read` pairs where:
- `getThing(offset: Int) -> Thing?`, reads `Thing` from the given offset or returns `nil` if there weren't enough bytes to read `Thing` or the bytes of `Thing` weren't within the range of the readable bytes. `get` methods don't advance the reader index.
- `readThing() -> Thing?` usually calls `getThing(offset: readerIndex)` and advances the `readerIndex` if `Thing` was read.

As a tool for debugging, we shouldn't be advancing the reader index here. I also don't think we should be able to specify an offset. We should just do a hex dump of all readable bytes in the buffer without advancing the reader index.

Sources/NIOCore/ByteBuffer-core.swift
glbrntt: > Honestly, I would delete (or leave alone) the old `self._storage.dumpBytes()`. That was written to debug the underlying storage itself when we're replacing the storage. So essentially it's an internal debugging facility that we can delete.
> 
> This new hex dumping facility probably need no APIs but
> 
> * `self.withUnsafeReadableBytes { ... }` <- pointer to all readable bytes
> * `self.readableBytesView` <- a `Collection` of all readable bytes
> * `self.getInteger(...)` <- a third way to access the bytes
> * `self.readerIndex`/`self.writerIndex`/`self.redableBytes` <- how many/which bytes are readable
> 
> all of them are public.

Another implicit API which might be of use is just grabbing a mutable copy of the buffer as the backing storage will only copy-on-write (CoW) if you modify the _underlying data_. This may simplify the length limited cases:

```swift
let originalBuffer = ...
// buffer and originalBuffer share underlying storage for data but do not share
// their indexes
var buffer = originalBuffer
// moves readerIndex on buffer but not originalBuffer
let prefix = buffer.readSlice(length: 64)
let prefixHex = prefix.hexDump()

buffer.advanceReaderIndex(by: ...)
let suffixHex = buffer.hexDump()
```

Sources/NIOCore/ByteBuffer-conversions.swift
```diff
@@ -35,7 +35,18 @@ extension String {
         var buffer = buffer
         self = buffer.readString(length: buffer.readableBytes)!
     }
-    
+
+    /// Creates a `String` from a given Int with base (radix) of 16, with the provided padding size.
+    /// - parameter padding: The desired length of the resulting string.
+    @inlinable
+    public init<T>(byte: T, padding: Int) where T : BinaryInteger {
```
glbrntt, line 42: We can keep this `internal` for the time being

Sources/NIOCore/ByteBuffer-conversions.swift
```diff
@@ -35,7 +35,18 @@ extension String {
...
+        self.init(byte, radix: 16)
+        var hexString = String(byte, radix: 16)
+        while hexString.count < padding {
+            hexString = "0" + hexString
+        }
```
glbrntt, line 47: This is expensive! Instead calculate the amount of padding required and use `String(repeating: "0", count: amountOfPadding)` to create the padding and then combine it with the `hexString`

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,100 @@ extension ByteBuffer {
...
+                                                 length: clipLength)
+
+            return startHex + " ... " + endHex
+        }
+    }
+
+    /// Returns a `String` of hexadecimal digits of bytes in the Buffer,
+    /// with formatting compatible with output of `hexdump -C`.
+    func hexDumpLong(offset: Int = 0, readableOnly: Bool = true) -> String {
+        var result = ""
```
glbrntt, line 1173: While this is a debugging tool it'd be nice if we could reduce allocations where possible. As we know how many bytes are in the buffer and what the output looks like we can estimate how long the result will be and use `result.reserveCapacity(...)` to allocate that space ahead of time (otherwise we end up repeatedly doubling the size of the resulting string as we append to it which can get quite expensive).

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,100 @@ extension ByteBuffer {
...
+        let lastIndex = (readableOnly ? self.writerIndex : self.capacity)
+        let length = lastIndex - offset
+
+        // Grab the pointer to the whole buffer storage. We'll use offset and length when we index into it.
+        let bytes = UnsafeRawBufferPointer(start: self._storage.bytes, count: Int(self.capacity))
+
+        // hexdump -C dumps into lines of upto 16 bytes each
+        // with a last line containing only the offset column.
+        for line in 0 ... length/16 {
```
glbrntt, line 1182: This might be a little simpler if you use slices:

```swift
var buffer = self
var lineOffset = 0
while buffer.readableBytes > 0 {
    let slice = buffer.readSlice(length: min(16, buffer.readableBytes))! 
    let lineLength = slice.readableBytes

    // ...

    lineOffset += lineLength
}

```

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1119,3 +1117,100 @@ extension ByteBuffer {
...
+            // The index range in the buffer storage for this line.
+            let range = (offset + line*16)..<(offset + line*16 + lineLength)
+
+            // Grab just the bytes for this particular line
+            let lineBytes = bytes[range]
+
+            // Make a hexdump of a single line of up to 16 bytes
+            // Pad it with empty space if there were not enough bytes in the buffer dump range
+            // And insert the separator space in the middle of the line
+            var lineDump = bytes[range].map { String(byte: $0, padding: 2) }.joined(separator: " ")
```
glbrntt, line 1201: This is effectively `hexDumpShort` -- if you use the `slice` as suggested above you can make use of it here too.

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1843,19 +1843,103 @@ class ByteBufferTest: XCTestCase {
...
         a0 a1 a2 a3 a4 a5 a6 a7 a8 a9 aa ab ac ad ae af b0 b1 b2 b3 b4 b5 b6 b7 b8 b9 ba bb bc bd be bf \
         c0 c1 c2 c3 c4 c5 c6 c7 c8 c9 ca cb cc cd ce cf d0 d1 d2 d3 d4 d5 d6 d7 d8 d9 da db dc dd de df \
-        e0 e1 e2 e3 e4 e5 e6 e7 e8 e9 ea eb ec ed ee ef f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 fa fb fc fd fe ff ]
+        e0 e1 e2 e3 e4 e5 e6 e7 e8 e9 ea eb ec ed ee ef f0 f1 f2 f3 f4 f5 f6 f7 f8 f9 fa fb fc fd fe ff
         """
         XCTAssertEqual(expected, actual)
     }
 
+    func testHexDumpShort() {
+        let buf = ByteBufferAllocator().buffer(string: "Hello")
```
glbrntt, line 1859: FWIW: you can just use `ByteBuffer(string: "Hello")`, no need to create an allocator.

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1843,19 +1843,103 @@ class ByteBufferTest: XCTestCase {
...
+        for f in UInt8.min...UInt8.max {
+            self.buf.writeInteger(f)
+        }
+        let actual = self.buf.hexDumpShort(limit: 10)
+        let expected = "00 01 02 03 04 ... fb fc fd fe ff"
+        XCTAssertEqual(expected, actual)
+    }
+
+    func testHexDumpLong() {
+        let buf = ByteBufferAllocator().buffer(string: "Goodbye, world! It was nice knowing you.\n")
```
glbrntt, line 1905: Cheerful! 😅

glbrntt (commented):
Great start @natikgadzhi! Sorry for the delay, I've been out of office for a while. I added a little to weissi's feedback.

Tests/NIOCoreTests/ByteBufferTest.swift
1proprogrammerchant: oooo!

Tests/NIOCoreTests/ByteBufferTest.swift
natikgadzhi: I'll get the feedback items one by one. Getting back into the PR after two weeks of people-centered work is rough ;)

Sources/NIOCore/ByteBuffer-core.swift
natikgadzhi: That saved me lots of headache just now.

natikgadzhi:
@glbrntt, I cleaned up what I already had — but I still have to write the clipped version of the hexdump, and then implement the format enum and the switcher method. Time to get some sleep, but I hope to get it to work tomorrow.

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+            let startHex = self.getSlice(at: 0, length: limit/2)!.hexDumpShort()
+            let endHex  = self.getSlice(at: self.readableBytes - limit/2, length: limit/2)!.hexDumpShort()
+            return startHex + " ... " + endHex
+        }
+    }
+
+    /// Returns a `String` of hexadecimal digits of bytes in the Buffer,
+    /// with formatting compatible with output of `hexdump -C`.
+    func hexDumpLong() -> String {
+        var result = ""
```
glbrntt, line 54: Can we reserve capacity on the result as we did for the short hex dump?

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+
+extension ByteBuffer {
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer,
+    /// in a format that's compatible with `xxd -r -p`.
+    /// `hexDumpShort()` always dumps all readable bytes, i.e. from `readerIndex` to `writerIndex`,
+    /// so you should set those indices to desired location to get the offset and length that you need to dump.
+    func hexDumpShort() -> String {
+        var hexString = ""
+        hexString.reserveCapacity(self.readableBytes*3)
```
glbrntt, line 23:
```suggestion
        hexString.reserveCapacity(self.readableBytes * 3)
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+            let slice = buffer.readSlice(length: min(16, buffer.readableBytes))!
+            let lineLength = slice.readableBytes
+
+            result += "\(String(byte: lineOffset, padding: 8))  "
+
+            // Start with a xxd-format hexdump that we already have, then pad it
+            // and insert a space in the middle.
+            var lineHex = slice.hexDumpShort()
+            lineHex.append(String(repeating: " ", count: 49 - lineHex.count))
+            lineHex.insert(" ", at: lineHex.index(lineHex.startIndex, offsetBy: 24))
```
glbrntt, line 69: Instead of doing one hex dump and shifting (which is expensive) can we create two slices and append the hex dump of each? This will be slightly more code but I think will be easier to follow and more closely match the output in structure.

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+            result += lineHex
+
+            // ASCII column renders the line as ASCII characters, or "." if the character is not printable.
+            result += "|" + String(decoding: slice.getBytes(at: 0, length: slice.readableBytes)!, as: Unicode.UTF8.self).map {
+                $0 >= " " && $0 < "~" ? $0 : "."
```
glbrntt, line 74: Rather than converting the bytes to string space and then mapping them we can do it more cheaply by operating in byte-space. We can also make use of the readable bytes view:

```swift
let printableBytes = slice.readableBytesView.map { 
    let printableRange = UInt8(ascii: " ") ..< UInt8(ascii: "~")
    return printableRange.contains($0) ? $0 : UInt8(ascii: ".")
}

result += "|" 
result += String(decoding: printableBytes, as: UTF8.self)
result += "|\n"
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+        /// A hexdump format compatible with the xxd utility.
+        /// - parameters:
+        ///     - maxLength: The maximum amount of bytes to dump before clipping for readability.
+        case xxdCompatible(maxLength: Int? = nil)
+
+        /// A hexdump format compatible with hexdump utility.
+        /// - parameters:
+        ///     - maxLength: The maximum amount of bytes to dump before clipping for readability.
+        case hexDumpCompatible(maxLength: Int? = nil)
+    }
```
glbrntt, line 110: Unfortunately we can't use an `enum` because adding a `case` is an API breaking change.

We can, however, have a struct which wraps an `enum`:


```swift
public struct HexDumpFormat {
    enum Value {
         case xxdCompatible(maxLength: Int?)
         case hexDumpCompatible(maxLength: Int?)
    }

    var value: Value
    init(_ value: Value) { self.value = value }

    public static let xxdCompatible = Self(.xxdCompatible(maxLength: nil))

    public static func xxdCompatible(maxLength: Int) -> Self { 
        Self(.xxdCompatible(maxLength: maxLength))
    }

    // etc...
}
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+
+            lineOffset += lineLength
+        }
+
+        // Add the last line with the last byte offset
+        result += String(byte: lineOffset, padding: 8)
+        return result
+    }
+
+//    func hexDumpLong(limit: Int) -> String {
```
glbrntt, line 85: If you pull out the loop from your existing `hexDumpLong` to have an `offset` you can build this function more easily.

glbrntt (commented):
This is coming along nicely!

Sources/NIOCore/ByteBuffer-hexdump.swift
1proprogrammerchant: why can't we use enum again

1proprogrammerchant: ahh thanks!

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// - parameters:
+    ///     - limit: The maximum amount of bytes presented in the dump.
+    func hexDumpShort(limit: Int) -> String {
+        // If the length of the resulting hex dump is smaller than the maximum allowed dump length,
+        // return the full dump. Otherwise, dump the first and last pieces of the buffer, then concatenate them.
+        if readableBytes <= limit {
+            return self.hexDumpShort()
+        } else {
+            let startHex = self.getSlice(at: 0, length: limit/2)!.hexDumpShort()
+            let endHex  = self.getSlice(at: self.readableBytes - limit/2, length: limit/2)!.hexDumpShort()
```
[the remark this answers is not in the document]
natikgadzhi: This is neat! Thank you.

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: I _think_ it's okay to switch `format.value` still in our code, right?

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: I felt greedy with those three lines of string concatenation and squashed them in one line of code. I think readability is still fine, but I appreciate how clear this one by line append code is.

natikgadzhi:
I've worked through most of your feedback, but left the clipped hexdump format implementation for tomorrow. I hope to find time and work on it in the morning, though. 

One tricky part (to me) is to figure out what to use for the placeholder when clipping, if the clipping happens within a line. 

One option would be to always clip on 16-byte increments (but that won't be precise). Another option would be to use "." or space to show that that byte was omitted. And always insert one empty line with "..." in the middle to show that this line signifies the clipped part. 

So it would be something like this: 

```
        let buf = ByteBuffer(string: "Goodbye, world! It was nice knowing you.\n")
        let expected = """
        00000000  47 6f 6f 64 62 79 65 2c  20 77 6f 72 6c           |Goodbye, worl   |

                                         ...

        00000010                                       6b 6e 6f 77  |            know|
        00000020  69 6e 67 20 79 6f 75 2e  0a                       |ing you..|
        00000029
        """
```

But that looks weird to me. Admittedly, I also didn't use hex dumps very often. What format should I implement, @glbrntt?

Sources/NIOCore/ByteBuffer-hexdump.swift
glbrntt: yes, that's absolutely fine because the `Value` is `internal`

Sources/NIOCore/ByteBuffer-hexdump.swift
glbrntt: IMO readability counts for a lot. I'd rather have five easy to understand lines than a single more complicated line.

1proprogrammerchant:
> 

I like this

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: Oh yes, I absolutely agree with you — that's why I commented on it. I sort of cleaned it up a bit, but still left a single line with string interpolation. But, I'm a Swift newbie, so if you see that three lines are more readable to the folks in the community, I would do that instead.

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: Did that in https://github.com/apple/swift-nio/pull/2475/commits/2f0d956ba6e85c10c77be9db51e359578b89d929 — not that I'm particularly proud of the resulting code.

natikgadzhi:
@glbrntt, @weissi, okay, it seems to work now! Time for a review.

@glbrntt, I did `reserveCapacity()` in `_hexDumpLines` helper function that's messy with result +=, but didn't reserve capacity in the clipped length variant. There are four consecutive string appends that will more than double the string size, but not 10x it — I think it's okay to not reserve capacity in that case?

weissi (commented):
Thank you so much, I love it! I had a quick read through and this looks broadly fine. But I didn't look into any of the details and will leave that up to @glbrntt and team.

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1843,19 +1843,102 @@ class ByteBufferTest: XCTestCase {
...
+        }
+        let actual = self.buf.hexDumpShort(limit: 10)
+        let expected = "00 01 02 03 04 ... fb fc fd fe ff"
+        XCTAssertEqual(expected, actual)
+    }
+
+    func testHexDumpLong() {
+        let buf = ByteBuffer(string: "Goodbye, world! It was nice knowing you.\n")
+        let expected = """
+        00000000  47 6f 6f 64 62 79 65 2c  20 77 6f 72 6c 64 21 20  |Goodbye, world! |
```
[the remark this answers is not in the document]
natikgadzhi: This line seems to be popular 🙃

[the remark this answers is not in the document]
weissi: :P . I didn't actually read the text 🙈🙈🙈🙈

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension ByteBuffer {
+
+    /// Describes a ByteBuffer hexDump format.
+    /// Can be either xxd output compatible, or hexdump compatible.
+    public struct HexDumpFormat {
```
Lukasa, line 19: Can I propose we make this `Hashable` and `Sendable`?

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+    ///
+    /// - parameters:
+    ///     - limit: The maximum amount of bytes presented in the dump.
+    func hexDumpShort(limit: Int) -> String {
+        guard self.readableBytes > limit else {
+            return self.hexDumpShort()
+        }
+
+        var buffer = self
+        let front = buffer.readSlice(length: limit / 2)!
```
Lukasa, line 76: Can you add a comment explaining why this force-unwrap is safe?

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+        buffer.moveReaderIndex(forwardBy: buffer.readableBytes - limit / 2)
```
Lukasa, line 77: It might be cleaner to use `moveReaderIndex(to:)` with a computation from `writerIndex`. Up to you though.

Sources/NIOCore/ByteBuffer-conversions.swift
```diff
@@ -35,7 +35,16 @@ extension String {
         var buffer = buffer
         self = buffer.readString(length: buffer.readableBytes)!
     }
-    
+
+    /// Creates a `String` from a given Int with base (radix) of 16, with the provided padding size.
+    /// - parameter padding: The desired length of the resulting string.
+    @inlinable
+    internal init<T>(byte: T, padding: Int) where T : BinaryInteger {
```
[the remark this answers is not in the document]
glbrntt: This one hasn't been addressed

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+            result += String(byte: lineOffset, padding: 8)
+            result += "  "
+
+            // Center column consists of:
+            // - xxd-compatible dump of the first 8 bytes
+            // - space
+            // - xxd-compatible dump of the rest 8 bytes
+            // If there are not enough bytes to dump, the column is padded with space.
+            let left = slice.getSlice(at: 0, length: min( 8, slice.readableBytes ))!
+            let right = slice.getSlice(at: 8, length: min( 8, slice.readableBytes - 8)) ?? ByteBuffer()
```
glbrntt, line 111: Minor thing here, we can avoid `getSlice` again by grabbing a mutable copy of the slice:

```swift
var otherSlice = slice
let left = otherSlice.readSlice(length: min(8, otherSlice.readableBytes))!
let right = otherSlice.readSlice(length: min(8, otherSlice.readableBytes))!
```

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1843,19 +1843,102 @@ class ByteBufferTest: XCTestCase {
...
+        """
+        let actual = buf.hexDump(format: .hexDumpCompatible)
+        XCTAssertEqual(expected, actual)
+    }
+
+    func testHexDumpLongEmptyBuffer() {
+        let buf = ByteBuffer()
+        let expected = "00000000"
+        let actual = buf.hexDump(format: .hexDumpCompatible)
+        XCTAssertEqual(expected, actual)
```
glbrntt, line 1939: Interestingly `echo -n '' | hexdump -C` doesn't print any output: we should do the same.

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+        result += front._hexDumpLines(offset: 0)
+        result += separator
+        result += back._hexDumpLines(offset: self.readableBytes - limit/2)
+        result += String(byte: self.readableBytes, padding: 8)
+        return result
+    }
+
+    /// hexDump this `ByteBuffer` in a preferred `HexDumpFormat`.
+    /// hexDump will dump only the readable bytes, from `readerIndex` to `writerIndex`. You're responsible for setting those indices or slicing into the part that you want to be dumped.
+    ///
```
glbrntt, line 168: I think we can skip the part about the reader and writer index; we generally advise users to not move them manually but to have them moved by using the various `read/`write` methods.

```suggestion
    /// Returns a hex dump of the readable bytes of this `ByteBuffer` in a preferred `HexDumpFormat`.
```

Sources/NIOCore/ByteBuffer-hexdump.swift
glbrntt, line 19:
```suggestion
    public struct HexDumpFormat: Sendable {
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension ByteBuffer {
+
+    /// Describes a ByteBuffer hexDump format.
+    /// Can be either xxd output compatible, or hexdump compatible.
+    public struct HexDumpFormat {
+        enum Value {
```
glbrntt, line 20:
```suggestion
        enum Value: Sendable {
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+
+        /// A hex dump format compatible with `xxd` CLI utility.
+        public static let xxdCompatible = Self(.xxdCompatible(maxLength: nil))
+
+        /// A hex dump format compatible with `hexdump` command line utility.
+        public static let hexDumpCompatible = Self(.hexDumpCompatible(maxLength: nil))
+
+        /// A hex dump format compatible with `xxd`, clipped to `maxLength` bytes dumped.
+        /// This format will dump first maxLength / 2 bytes, and the last maxLength / 2 bytes, replacing the rest with " ... ".
+        public static func xxdCompatible(maxLength: Int) -> Self {
```
glbrntt, line 36: I think we should use `maxBytes` instead of `maxLength`;  you could mistake length for being the length of the output.

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+            Self(.xxdCompatible(maxLength: maxLength))
+        }
+
+        /// A hex dump format compatible with `hexdump` command line tool.
+        /// This format will dump first maxLength / 2 bytes, and the last maxLength / 2 bytes, with a placeholder in between.
+        public static func hexDumpCompatible(maxLength: Int) -> Self {
```
glbrntt, line 42: I think we should use `maxBytes` instead of `maxLength`;  you could mistake length for being the length of the output.

glbrntt (changes requested):

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer.,
+    /// in a format that's compatible with `xxd -r -p`, but clips the output to the max length of `limit` bytes.
+    /// If the dump contains more than the `limit` bytes, this function will return the first `limit/2`
+    /// and the last `limit/2` of that, replacing the rest with `...`, i.e. `01 02 03 ... 09 11 12`.
+    ///
+    /// - parameters:
+    ///     - limit: The maximum amount of bytes presented in the dump.
+    func hexDumpShort(limit: Int) -> String {
+        guard self.readableBytes > limit else {
```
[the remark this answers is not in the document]
natikgadzhi: Yep! I started with an if statement, then thought I'd save on nesting and replaced them with a guard, but I agree the readability suffered. Fixing now.

Sources/NIOCore/ByteBuffer-hexdump.swift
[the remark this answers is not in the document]
natikgadzhi: This is such a clean way to guarantee the right part of the line is not nil without coalescing! TIL.

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: Renamed everywhere — it's easier to use just `maxBytes` instead of `limit` and `maxLength`.

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        var value: Value
+        init(_ value: Value) { self.value = value }
+
+        /// A hex dump format compatible with `xxd` CLI utility.
+        public static let xxdCompatible = Self(.xxdCompatible(maxLength: nil))
+
+        /// A hex dump format compatible with `hexdump` command line utility.
+        public static let hexDumpCompatible = Self(.hexDumpCompatible(maxLength: nil))
```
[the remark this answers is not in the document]
natikgadzhi: I like your idea, so I, uh, renamed all the things to this notation.

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: Went with @Lukasa's feedback and made them both `Hashable` and `Sendable`.

natikgadzhi:
@glbrntt, @Lukasa, I've pushed up a commit that addresses most items, I think. A few things are not done yet: 

- @glbrntt, you mentioned that feedback in `Hexdump-conversions.swift` is not addressed. I think I missed it. Here's what I see: https://github.com/apple/swift-nio/pull/2475#discussion_r1280188883 — and the ask to make the initalizer internal. I think I did both? What am I missing? To me personally, `String(byte, padding:10)` feels a bit misleading, I think I should rename `padding` to `size` of the string.
- @glbrntt, the detailed output format with maxBytes does not work as expected yet — I need to pad the output with spaces for clipped pieces in lines where there are some bytes bumped. 

Pushing up the rest, will clean up the remaining two items in a bit.

glbrntt:
> @glbrntt, @Lukasa, I've pushed up a commit that addresses most items, I think. A few things are not done yet:
> 
> * @glbrntt, you mentioned that feedback in `Hexdump-conversions.swift` is not addressed. I think I missed it. Here's what I see: [Adding ByteBuffer.hexDump in xxd and hexdump -C formats #2475 (comment)](https://github.com/apple/swift-nio/pull/2475#discussion_r1280188883) — and the ask to make the initalizer internal. I think I did both? What am I missing? To me personally, `String(byte, padding:10)` feels a bit misleading, I think I should rename `padding` to `size` of the string.

The comment which hasn't been address is: https://github.com/apple/swift-nio/pull/2475/files#r1280384327

> * @glbrntt, the detailed output format with maxBytes does not work as expected yet — I need to pad the output with spaces for clipped pieces in lines where there are some bytes bumped.
> 
> Pushing up the rest, will clean up the remaining two items in a bit.

👍

glbrntt:

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,205 @@
...
+        switch(format.value) {
+        case .plain(let maxBytes):
+            if let maxBytes {
+                return self.hexDumpPlain(maxBytes: maxBytes)
+            } else {
+                return self.hexDumpPlain()
+            }
+
+        case .detailed(let maxBytes):
+            if let maxBytes {
```
glbrntt, line 198: Same here, I think this needs to be `if let maxBytes = maxBytes`

glbrntt (commented):
Looks really good overall. I left a couple of tiny things to fix.

natikgadzhi:
I have a commit I need to push up that looks ugly AF, but matches the detailed limited format. Sick kid, so I was slow yesterday. Ready to wrap this up tonight. On Aug 8, 2023, at 7:26 AM, George Barnett ***@***.***> wrote:﻿
@glbrntt commented on this pull request.

Looks really good overall. I left a couple of tiny things to fix.

In Sources/NIOCore/ByteBuffer-hexdump.swift:
> +            hexString += String(byte, radix: 16, padding: 2)
+            hexString += " "
+        }
+
+        return String(hexString.dropLast())
+    }
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer.,
+    /// in a format that's compatible with `xxd -r -p`, but clips the output to the max length of `maxBytes` bytes.
+    /// If the dump contains more than the `maxBytes` bytes, this function will return the first `maxBytes/2`
+    /// and the last `maxBytes/2` of that, replacing the rest with `...`, i.e. `01 02 03 ... 09 11 12`.
+    ///
+    /// - parameters:
+    ///     - maxBytes: The maximum amount of bytes presented in the dump.
+    private func hexDumpPlain(maxBytes: Int) -> String {
+        // If the buffer lenght fits in the max bytes limit in the hex dump, just dump the whole thing.

⬇️ Suggested change
-        // If the buffer lenght fits in the max bytes limit in the hex dump, just dump the whole thing.
+        // If the buffer length fits in the max bytes limit in the hex dump, just dump the whole thing.





In Sources/NIOCore/ByteBuffer-hexdump.swift:
> +        return startHex + " ... " + endHex
+    }
+
+    // Dump hexdump -C compatible format without the last line containing the byte length of the buffer
+    private func _hexDumpLines(offset: Int) -> String {
+        var result = ""
+        // Each hex dump line represents 16 bytes, and takes up to 79 characters including the \n in the end.
+        // Plus the last line of the hex dump which is just 8 characters.
+        result.reserveCapacity(self.readableBytes / 16 * 79 + 8)
+
+        var buffer = self
+        // We're going to move the lineOffset index forward, so copy it into a variable
+        var lineOffset = offset
+
+        while buffer.readableBytes > 0 {
+            let slice = buffer.readSlice(length: min(16, buffer.readableBytes))! // Safe to force-unwrap because we're in a loop that guarantees there's at least one byte to read.

fwiw: it's always safe to unwrap buffer.readSlice(length: buffer.readableBytes)



In Sources/NIOCore/ByteBuffer-hexdump.swift:
> +    }
+
+    /// Returns a hex dump of  this `ByteBuffer` in a preferred `HexDumpFormat`.
+    ///
+    /// `hexDump` provides four formats:
+    ///     - `.plain` — plain hex dump format with hex bytes separated by spaces, i.e. `48 65 6c 6c 6f` for `Hello`. This format is compatible with `xxd -r`.
+    ///     - `.plain(maxBytes: Int)` — like `.plain`, but clipped to maximum bytes dumped.
+    ///     - `.detailed` — detailed hex dump format with both hex, and ASCII representation of the bytes. This format is compatible with what `hexdump -C` outputs.
+    ///     - `.detailed(maxBytes: Int)` — like `.detailed`, but  clipped to maximum bytes dumped.
+    ///
+    /// - parameters:
+    ///     - format: ``HexDumpFormat`` to use for the dump.
+    public func hexDump(format: HexDumpFormat) -> String {
+        switch(format.value) {
+        case .plain(let maxBytes):
+            if let maxBytes {

We support back to Swift 5.6 so I think you'll need to spell this out as if let maxBytes = maxBytes



In Sources/NIOCore/ByteBuffer-hexdump.swift:
> +    ///     - `.detailed` — detailed hex dump format with both hex, and ASCII representation of the bytes. This format is compatible with what `hexdump -C` outputs.
+    ///     - `.detailed(maxBytes: Int)` — like `.detailed`, but  clipped to maximum bytes dumped.
+    ///
+    /// - parameters:
+    ///     - format: ``HexDumpFormat`` to use for the dump.
+    public func hexDump(format: HexDumpFormat) -> String {
+        switch(format.value) {
+        case .plain(let maxBytes):
+            if let maxBytes {
+                return self.hexDumpPlain(maxBytes: maxBytes)
+            } else {
+                return self.hexDumpPlain()
+            }
+
+        case .detailed(let maxBytes):
+            if let maxBytes {

Same here, I think this needs to be if let maxBytes = maxBytes

glbrntt:
> I have a commit I need to push up that looks ugly AF, but matches the detailed limited format. Sick kid, so I was slow yesterday. Ready to wrap this up tonight.

Don't worry, there's no rush for this 🙂

natikgadzhi:
@glbrntt, okay, done with a little update! 

- I think I combed through your feedback. I left the safe to unwrap comments there, but I think I acted on the rest. 
- I've reworked how `.detailed` format prints out the dump to accommodate the format. It's a little bit tricky, and I had to write some repetitive code to get it to work. I'm not _entirely_ happy with it, and tried to document it thoroughly. I think there's a trade-off between the dump readability (format) and code complexity to get it dumped in said format. 
- Caught a bug in a case when the dump with maxBytes is long enough for the clipped pieces to take multiple lines and fixed it.

I think this is ready for a CI run and a review.

glbrntt:

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+        init(_ value: Value) { self.value = value }
+
+        /// A plain hex dump format compatible with `xxd` CLI utility.
+        public static let plain = Self(.plain(maxBytes: nil))
+
+        /// A hex dump format compatible with `hexdump` command line utility.
+        public static let detailed = Self(.detailed(maxBytes: nil))
+
+        /// A detailed hex dump format compatible with `xxd`, clipped to `maxLength` bytes dumped.
+        /// This format will dump first maxLength / 2 bytes, and the last maxLength / 2 bytes, replacing the rest with " ... ".
```
glbrntt, line 36:
```suggestion
        /// A detailed hex dump format compatible with `xxd`, clipped to `maxBytes` bytes dumped.
        /// This format will dump first `maxBytes / 2` bytes, and the last `maxBytes / 2` bytes, replacing the rest with " ... ".
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+        public static func plain(maxBytes: Int) -> Self {
+            Self(.plain(maxBytes: maxBytes))
+        }
+
+        /// A hex dump format compatible with `hexdump` command line tool.
+        /// This format will dump first maxLength / 2 bytes, and the last maxLength / 2 bytes, with a placeholder in between.
```
glbrntt, line 42:
```suggestion
        /// This format will dump first `maxBytes / 2` bytes, and the last `maxBytes / 2` bytes, with a placeholder in between.
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+        public static func detailed(maxBytes: Int) -> Self {
+            Self(.detailed(maxBytes: maxBytes))
+        }
+    }
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer,
```
glbrntt, line 48:
```suggestion
    /// Return a `String` of space separated hexadecimal digits of the readable bytes in the buffer,
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        for byte in self.readableBytesView {
+            hexString += String(byte, radix: 16, padding: 2)
+            hexString += " "
+        }
+
+        return String(hexString.dropLast())
+    }
+
+    /// Return a `String` of hexadecimal digits of bytes in the Buffer.,
```
glbrntt, line 64:
```suggestion
    /// Return a `String` of space delimited hexadecimal digits of the readable bytes in the buffer,
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+        let front = buffer.readSlice(length: maxBytes / 2)!
+        buffer.moveReaderIndex(to: buffer.writerIndex - maxBytes / 2)
+        let back = buffer.readSlice(length: buffer.readableBytes)!
+
+        let startHex = front.hexDumpPlain()
+        let endHex = back.hexDumpPlain()
+        return startHex + " ... " + endHex
+    }
+
+    /// Returns a String of detailed hex dump of this slice.
```
glbrntt, line 89:
```suggestion
    /// Returns a `String` containing a detailed hex dump of this buffer.
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        // Iterate over the bytes in the line, and dump them one by one, insert a separator if needed.
+        for (byteIndex, byte) in self.readableBytesView.enumerated() {
+            result += String(byte, radix: 16, padding: 2)
+            result += " "
+            if byteIndex + paddingBefore == 7 {
+                result += " "
+            }
+        }
```
glbrntt, line 125: This part is a little hard to follow because of the extra spaces being added conditionally. Can I suggest something like this?

```swift
    result += String(repeating: " ", count: paddingBefore * 3)

    // Add the left side of the central column
    if paddingBefore < 8 {
        for byte in self.readableBytesView.prefix(8 - paddingBefore) {
            result += String(byte, radix: 16, padding: 2)
            result += " "
        }
    }

    // Add an extra space for the centre column.
    result += " "

    // Add the right side of the central column.
    let bytesToSkip = paddingBefore < 8 ? 8 - paddingBefore : 0
    for byte in self.readableBytesView.dropFirst(bytesToSkip) {
        result += String(byte, radix: 16, padding: 2)
        result += " "
    }
```

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1843,19 +1843,108 @@ class ByteBufferTest: XCTestCase {
...
+        00000029
+        """
+        let actual = buf.hexDump(format: .detailed(maxBytes: 16))
+        XCTAssertEqual(expected, actual)
+    }
+
+    func testHexDumpDetailedWithMultilineFrontAndBack() {
+        let buf = ByteBuffer(string: """
+        Goodbye, world! It was nice knowing you.
+        I will miss this pull request with all of it's 94+ comments.
```
glbrntt, line 1913: We can keep it going if you like? 😅 (Also, thank you again, this is a really awesome addition!)

glbrntt (commented):
This is looking really good. I mostly left doc/typo fixes but also left a little feedback on `hexDumpLine`. Thanks for sticking with this!

Tests/NIOCoreTests/ByteBufferTest.swift
1proprogrammerchant: such good changes

Sources/NIOCore/ByteBuffer-hexdump.swift
natikgadzhi: Reworked it based on your feedback: https://github.com/apple/swift-nio/pull/2475/commits/d057109f9661b26df9cda8a9defb2618fd448a14 

That way, all `if` statements can go away, seems more readable?

natikgadzhi (commented):
Ready for the next round ;)

glbrntt (approved):
This looks good to me now, great job on this and thanks for sticking with it 🙂

Sources/NIOCore/ByteBuffer-conversions.swift
```diff
@@ -35,7 +35,17 @@ extension String {
...
     }
-    
+
+    /// Creates a `String` from a given `Int` with a given base (`radix`), padded with zeroes to the provided `padding` size.
+    ///
+    /// - parameters:
+    ///     - radix: radix base to use for conversion.
+    ///     - padding: the desired lenght of the resulting string.
+    @inlinable
+    internal init<T>(_ value: T, radix: Int, padding: Int) where T: BinaryInteger {
```
Lukasa, line 45: Can we give this a better generic type parameter name? Maybe `Value`?

yim-lee:

yim-lee:

natikgadzhi:
Huh, let me do a clean build and fix those.

natikgadzhi:
- Fixed a test failure of `storage.dumpBytes`. I dropped `[` and `]` earlier, but then reverted it, and forgot to revert the test. That was the only test failure on Swift 5.7 -> nightly. 
- Fixed a Swift 5.6 compile error in https://github.com/apple/swift-nio/pull/2475/commits/dc65676a4b703e684ea0f0a32ff75bce2a43e0c4 — the code is a tad cleaner that way, too.

It should be good now. @glbrntt can you poke the swift CI gods, please?

yim-lee:
@natikgadzhi I added you to the allow list, so your pushes to the PR should automatically trigger CI from now on. (I just fixed the soundness check job permissions)

natikgadzhi:
@yim-lee, thank you <3

1proprogrammerchant:
> @natikgadzhi I added you to the allow list, so your pushes to the PR should automatically trigger CI from now on. (I just fixed the soundness check job permissions)
> 
> @swift-server-bot test this please

cool person

glbrntt (approved):
Awesome, thanks @natikgadzhi!

Lukasa (approved):

PR #2477, opened 2023-07-18
Throw SocketAddressError (cf. fatalError) in SocketAddress.convert for unknown address family

simonjbeaumont:
### Motivation:

NIO currently fails with `fatalError` when converting a `sockaddr_storage` if it is not one of the socket addresses that NIO explicitly supports (`AF_UNIX`, `AF_INET`, or `AF_INET6`). However, since NIO offers API that allows users to create sockets out of band (`withConnectedSocket(_:)` and `withBoundSocket(_:)`), it's possible a user has provided a socket of a different family.

### Modifications:

Instead of crashing, `SocketAddress.convert()` now throws `SocketAddressError.unsupported`, and whether to crash (or just propagate the error) is moved up the stack.

### Result:

It's possible to bootstrap clients and servers with other kinds of sockets, even if NIO cannot convert them to one of the `SocketAddress` cases it knows about.

### DIFF
--- a/Sources/NIOPosix/DatagramVectorReadManager.swift
+++ b/Sources/NIOPosix/DatagramVectorReadManager.swift
@@ -178,7 +178,9 @@ struct DatagramVectorReadManager {
 #else
             precondition(self.messageVector[i].msg_hdr.msg_namelen != 0, "Unexpected zero length peer name")
 #endif
-            let address: SocketAddress = self.sockaddrVector[i].convert()
+            let address: SocketAddress
+            do { address = try self.sockaddrVector[i].convert() }
+            catch { fatalError("Socket address conversion failed.") }
             
             // Extract congestion information if requested.
             let metadata: AddressedEnvelope<ByteBuffer>.Metadata?
--- a/Sources/NIOPosix/DatagramVectorReadManager.swift
+++ b/Sources/NIOPosix/DatagramVectorReadManager.swift
@@ -136,10 +136,10 @@ struct DatagramVectorReadManager {
             return .none
         case .processed(let messagesProcessed):
             buffer.moveWriterIndex(to: messageSize * messagesProcessed)
-            return self.buildMessages(messageCount: messagesProcessed,
-                                      sliceSize: messageSize,
-                                      buffer: &buffer,
-                                      parseControlMessages: parseControlMessages)
+            return try self.buildMessages(messageCount: messagesProcessed,
+                                          sliceSize: messageSize,
+                                          buffer: &buffer,
+                                          parseControlMessages: parseControlMessages)
         }
     }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/DatagramVectorReadManager.swift
```diff
@@ -178,7 +178,9 @@ struct DatagramVectorReadManager {
 #else
             precondition(self.messageVector[i].msg_hdr.msg_namelen != 0, "Unexpected zero length peer name")
 #endif
-            let address: SocketAddress = self.sockaddrVector[i].convert()
+            let address: SocketAddress
+            do { address = try self.sockaddrVector[i].convert() }
+            catch { fatalError("Socket address conversion failed.") }
```
Lukasa, line 183: Prefer `try!`.

Lukasa, line 183: More broadly, should we tolerate this error in this location?

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -651,7 +651,7 @@ final class DatagramChannel: BaseSocketChannel<Socket> {
                     metadata = nil
                 }
 
-                let msg = AddressedEnvelope(remoteAddress: rawAddress.convert(),
+                let msg = AddressedEnvelope(remoteAddress: try rawAddress.convert(),
```
Lukasa, line 654: Can we lift the call to `convert` higher in the `.processed` block? Let's not trigger any side-effects until we know we're going to process the read.

Sources/NIOPosix/DatagramVectorReadManager.swift
simonjbeaumont: Good to know—I assumed that `try!` would be discouraged in review. Although I appreciate that all we do here is fail with a different string.

Changed in 805f1ca82.

simonjbeaumont: That's a great question.

I suppose we have to choose between failing hard or silently dropping these messages, given that we will be unable to construct an `AddressedEnvelope` without a `SocketAddress`.

I opted to fail because:
1. It's closest to the behaviour before this patch—where it would have still crashed, just inside `convert()`.
2. Above these lines there are a number of `precondition` checks so failing hard seemed to be "the done thing" here.

Happy to revisit though if you think this is the wrong choice here.

Sources/NIOPosix/SocketChannel.swift
simonjbeaumont: f7ca7428d

Sources/NIOPosix/DatagramVectorReadManager.swift
Lukasa: `try!` falls into the same bucket as `!`, which is that you need to write down why you know it won't crash in a comment.

Lukasa: I think there's a third option, which is to throw. The preconditions above are correctness checks, but this is not. Throwing here is consistent with the scalar read path.

simonjbeaumont: ****

simonjbeaumont:

Lukasa (approved):

FranzBusch:
5.9 allocation regression is known and to be investigated
nightly crash is also known and fixed in Swift already.

PR #2479, opened 2023-07-21
Add bootstrap APIs for VSOCK sockets

simonjbeaumont:
### Motivation:

The VSOCK address family facilitates communication between virtual machines and the host they are running that need a communications channel that is independent of virtual machine network configuration.

While NIO has support for building channels using sockets that were created out of band (`withConnectedSocket(_:)` and `withBoundSocket(_:)`, there are no APIs that facilitate the creation of VSOCK-based channels.

### Modifications:

Given `AF_VSOCK` is just another address family, extending `enum SocketAddress` with a new case would be the most natural approach. However this is a breaking API change, so this pull request introduces a new type, `VsockAddress`, and some bootstrap APIs.

Note that this depends on #2477 and #2476, which make NIO more tolerant of channels backed by sockets other than those represented by `SocketAddress`.

Linux and Darwin man pages differ in both API and behaviour:

- Support for `SOCK_DGRAM`: only supported on Linux.
- Getting the local CID via `ioctl(2)`: Called with `/dev/vsock` on Linux and with the socket on Darwin.
- Use of `VMADDR_CID_LOCAL`: only present on Linux.
- Use of `VMADDR_CID_ANY` with `connect(2)`: documented only on Darwin

To accommodate these, this change only covers `SOCK_STREAM` and provides a consistent API for getting the local context ID for a socket (but discourage its use on Linux in the documentation).

This change also includes some tweaks to `NIOEchoClient` and `NIOEchoServer`—when run with two integer parameters, these are considered to be the VSOCK context ID and port number.

### Result:

It's now possible to bootstrap clients and servers using `VsockAddress`. For example, you can do the following to run `NIOEchoServer` and `NIOEchoClient` in the same Linux container over loopback (note `VMADDR_CID_LOCAL = 1` on Linux in the following example): 

```console
[root@3255e403fa47 code]# $(swift build --show-bin-path)/NIOEchoServer -1 12345
Server started and listening on nil
```

...then, exec'ing into the same container:

```console
[root@3255e403fa47 code]# $(swift build --show-bin-path)/NIOEchoClient 1 12345
Please enter line to send to the server
Hello!
Client connected to nil
Received: 'Hello!' back from the server, closing channel.
Client closed
```

It's important to callout that these binaries used to rely on `localAddress` and `remoteAddress` being non-nil. When working with sockets that are not handled by `enum SocketAddress` these _will_ be nil, so it's worth calling out because it seems common that people make the non-nil assumption.

### DIFF
--- a/Sources/CNIODarwin/shim.c
+++ b/Sources/CNIODarwin/shim.c
@@ -21,6 +21,7 @@
 #include <assert.h>
 #include <netinet/ip.h>
 #include <netinet/in.h>
+#include <sys/ioctl.h>
 
 int CNIODarwin_sendmmsg(int sockfd, CNIODarwin_mmsghdr *msgvec, unsigned int vlen, int flags) {
     // Some quick error checking. If vlen can't fit into int, we bail.
@@ -90,4 +91,9 @@ const int CNIODarwin_IPTOS_ECN_CE = IPTOS_ECN_CE;
 const int CNIODarwin_IPV6_RECVPKTINFO = IPV6_RECVPKTINFO;
 const int CNIODarwin_IPV6_PKTINFO = IPV6_PKTINFO;
 
+uint32_t CNIODarwin_get_local_vsock_cid(int socket) {
+    uint32_t cid = 0;
+    int result = ioctl(socket, IOCTL_VM_SOCKETS_GET_LOCAL_CID, &cid);
+    return cid;
+}
 #endif  // __APPLE__
--- a/Sources/NIOCore/VsockAddress.swift
+++ b/Sources/NIOCore/VsockAddress.swift
@@ -1,0 +1,221 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+#if canImport(Darwin) || os(Linux)
+
+#if canImport(Darwin)
+import Darwin
+import CNIODarwin
+fileprivate let get_local_vsock_cid: @convention(c) (CInt) -> UInt32 = CNIODarwin_get_local_vsock_cid
+#elseif os(Linux)
+#if canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#endif
+import CNIOLinux
+fileprivate let get_local_vsock_cid: @convention(c) (CInt) -> UInt32 = CNIOLinux_get_local_vsock_cid
+#else
+#error("Unable to identify your C library.")
+#endif
+
+/// A vsock socket address.
+///
+/// A socket address is defined as a combination of a Context Identifier (CID) and a port number.
+/// The CID identifies the source or destination, which is either a virtual machine or the host.
+/// The port number differentiates between multiple services running on a single machine.
+///
+/// For well-known CID values and port numbers, see ``ContextID`` and ``Port`` respectively.
+public struct VsockAddress: Sendable {
+    private let _storage: Box<sockaddr_vm>
+
+    /// The libc socket address for a vsock socket.
+    public var address: sockaddr_vm { return _storage.value }
+
+    /// Get the context ID associated with the address.
+    public var cid: Int {
+        Int(Int32(bitPattern: self.address.svm_cid))
+    }
+
+    /// Get the port associated with the address.
+    public var port: Int {
+        Int(Int32(bitPattern: self.address.svm_port))
+    }
+
+    /// Creates a new vsock address.
+    ///
+    /// - parameters:
+    ///     - address: the `sockaddr_vm` that holds the context ID and port.
+    public init(address: sockaddr_vm) {
+        self._storage = Box(address)
+    }
+
+    public init(cid: UInt32, port: UInt32) {
+        var addr = sockaddr_vm()
+        addr.svm_family = sa_family_t(NIOBSDSocket.AddressFamily.vsock.rawValue)
+        addr.svm_cid = cid
+        addr.svm_port = port
+        self.init(address: addr)
+    }
+
+    /// Creates a new vsock address.
+    ///
+    /// - parameters:
+    ///   - cid: the context ID.
+    ///   - port: the target port.
+    /// - returns: the `SocketAddress` for the given context ID and port combination.
+    public init(cid: Int, port: Int) {
+        self.init(cid: UInt32(bitPattern: Int32(truncatingIfNeeded: cid)), port: UInt32(bitPattern: Int32(truncatingIfNeeded: port)))
+    }
+
+    /// Get the local context ID for a vsock socket.
+    ///
+    /// - parameters:
+    ///   - socket: The vsock socket to get the local context ID for.
+    ///
+    /// - NOTE: On Linux, you can use `VMADDR_CID_LOCAL` to retrieve the local context ID, which does not require a socket.
+    public static func localVsockContextID(_ socket: NIOBSDSocket.Handle) -> Int {
+        Int(get_local_vsock_cid(socket))
+    }
+}
+
+extension VsockAddress {
+    /// A vsock Context Identifier (CID).
+    ///
+    /// The CID identifies the source or destination, which is either a virtual machine or the host.
+    public struct ContextID: RawRepresentable, ExpressibleByIntegerLiteral, Hashable, Sendable {
+        public var rawValue: UInt32
+
+        public init(rawValue: UInt32) {
+            self.rawValue = rawValue
+        }
+
+        public init(integerLiteral value: UInt32) {
+            self.init(rawValue: value)
+        }
+
+        /// Wildcard, matches any address.
+        ///
+        /// On all platforms, using this value with `bind(2)` means "any address".
+        ///
+        /// On Darwin platforms, the man page states this can be used with `connect(2)` to mean "this host".
+        public static let any: Self = Self(rawValue: VMADDR_CID_ANY)
+
+        /// The address of the hypervisor.
+        public static let hypervisor: Self = Self(rawValue: UInt32(VMADDR_CID_HYPERVISOR))
+
+        /// The address of the host.
+        public static let host: Self = Self(rawValue: UInt32(VMADDR_CID_HOST))
+
+#if os(Linux)
+        /// The address for local communication (loopback).
+        ///
+        /// This directs packets to the same host that generated them.  This is useful for testing
+        /// applications on a single host and for debugging.
+        ///
+        /// The local context ID obtained with ``getLocalContextID(_:)`` can be used for the same
+        /// purpose, but it is preferable to use ``local``.
+        public static let local: Self = Self(rawValue: UInt32(VMADDR_CID_LOCAL))
+#endif
+
+        /// Get the context ID of the local machine.
+        ///
+        /// On Linux, consider using ``local`` when binding instead of this function.
+        public static func getLocalContextID(_ socket: NIOBSDSocket.Handle) -> Self {
+            Self(rawValue: get_local_vsock_cid(socket))
+        }
+    }
+
+    /// A vsock port number.
+    ///
+    /// The vsock port number differentiates between multiple services running on a single machine.
+    public struct Port: RawRepresentable, ExpressibleByIntegerLiteral, Hashable, Sendable {
+        public var rawValue: UInt32
+
+        public init(rawValue: UInt32) {
+            self.rawValue = rawValue
+        }
+
+        public init(integerLiteral value: UInt32) {
+            self.init(rawValue: value)
+        }
+
+        /// Used to bind to any port number.
+        public static let any: Self = Self(rawValue: VMADDR_PORT_ANY)
+    }
+
+    /// Creates a new vsock address.
+    ///
+    /// - parameters:
+    ///   - cid: the context ID.
+    ///   - port: the target port.
+    /// - returns: the `SocketAddress` for the given context ID and port combination.
+    public init(cid: ContextID, port: Port) {
+        self.init(cid: cid.rawValue, port: port.rawValue)
+    }
+}
+
+extension VsockAddress: CustomStringConvertible {
+    public var description: String {
+        "[VSOCK]\(self.cid):\(self.port)"
+    }
+}
+
+/// Element-wise Equatable conformance using only the struct fields defined in the man page (excluding lengths).
+extension VsockAddress: Equatable {
+    public static func == (lhs: VsockAddress, rhs: VsockAddress) -> Bool {
+        (
+            lhs.address.svm_family == rhs.address.svm_family
+            && lhs.address.svm_reserved1 == rhs.address.svm_reserved1
+            && lhs.address.svm_port == rhs.address.svm_port
+            && lhs.address.svm_cid == rhs.address.svm_cid
+        )
+    }
+}
+
+/// Element-wise Hashable conformance using only the struct fields defined in the man page (excluding lengths).
+extension VsockAddress: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.address.svm_family)
+        hasher.combine(self.address.svm_reserved1)
+        hasher.combine(self.address.svm_port)
+        hasher.combine(self.address.svm_cid)
+    }
+}
+
+extension VsockAddress: SockAddrProtocol {
+    public func withSockAddr<T>(_ body: (UnsafePointer<sockaddr>, Int) throws -> T) rethrows -> T {
+        try self.address.withSockAddr({ try body($0, $1) })
+    }
+}
+
+extension NIOBSDSocket.AddressFamily {
+    /// Address for vsock.
+    public static let vsock: NIOBSDSocket.AddressFamily =
+            NIOBSDSocket.AddressFamily(rawValue: AF_VSOCK)
+}
+
+extension NIOBSDSocket.ProtocolFamily {
+    /// Vsock protocol.
+    public static let vsock: NIOBSDSocket.ProtocolFamily =
+            NIOBSDSocket.ProtocolFamily(rawValue: PF_VSOCK)
+}
+
+extension sockaddr_vm: SockAddrProtocol {
+    func withSockAddr<R>(_ body: (UnsafePointer<sockaddr>, Int) throws -> R) rethrows -> R {
+        return try withUnsafeBytes(of: self) { p in
+            try body(p.baseAddress!.assumingMemoryBound(to: sockaddr.self), p.count)
+        }
+    }
+}
+#endif  // canImport(Darwin) || os(Linux)
--- a/Sources/NIOEchoServer/main.swift
+++ b/Sources/NIOEchoServer/main.swift
@@ -70,18 +70,22 @@ let defaultPort = 9999
 enum BindTo {
     case ip(host: String, port: Int)
     case unixDomainSocket(path: String)
+    case vsock(_: VsockAddress)
 }
 
 let bindTarget: BindTo
 switch (arg1, arg1.flatMap(Int.init), arg2.flatMap(Int.init)) {
-case (.some(let h), _ , .some(let p)):
-    /* we got two arguments, let's interpret that as host and port */
+case (_, .some(let cid), .some(let port)):
+    /* we got two arguments (Int, Int), let's interpret that as vsock cid and port */
+    bindTarget = .vsock(VsockAddress(cid: cid, port: port))
+case (.some(let h), _, .some(let p)):
+    /* we got two arguments (String, Int), let's interpret that as host and port */
     bindTarget = .ip(host: h, port: p)
-case (.some(let portString), .none, _):
-    /* couldn't parse as number, expecting unix domain socket path */
-    bindTarget = .unixDomainSocket(path: portString)
-case (_, .some(let p), _):
-    /* only one argument --> port */
+case (.some(let pathString), .none, .none):
+    /* we got one argument (String), let's interpret that unix domain socket path */
+    bindTarget = .unixDomainSocket(path: pathString)
+case (_, .some(let p), .none):
+    /* we got one argument (Int), let's interpret that as port on default host */
     bindTarget = .ip(host: defaultHost, port: p)
 default:
     bindTarget = .ip(host: defaultHost, port: defaultPort)
@@ -93,10 +97,17 @@ let channel = try { () -> Channel in
         return try bootstrap.bind(host: host, port: port).wait()
     case .unixDomainSocket(let path):
         return try bootstrap.bind(unixDomainSocketPath: path).wait()
+    case .vsock(let vsockAddress):
+        return try bootstrap.bind(to: vsockAddress).wait()
     }
 }()
 
-print("Server started and listening on \(channel.localAddress!)")
+switch bindTarget {
+case .ip, .unixDomainSocket:
+    print("Server started and listening on \(channel.localAddress!)")
+case .vsock(let vsockAddress):
+    print("Server started and listening on \(vsockAddress)")
+}
 
 // This will never unblock as we don't close the ServerChannel
 try channel.closeFuture.wait()
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -326,6 +326,25 @@ public final class ServerBootstrap {
         return self.bind(unixDomainSocketPath: unixDomainSocketPath)
     }
 
+    #if canImport(Darwin) || os(Linux)
+    public func bind(to vsockAddress: VsockAddress) throws -> EventLoopFuture<Channel> {
+        func makeChannel(_ eventLoop: SelectableEventLoop, _ childEventLoopGroup: EventLoopGroup, _ enableMPTCP: Bool) throws -> ServerSocketChannel {
+            if enableMPTCP {
+                throw ChannelError.operationUnsupported
+            }
+            let socket = try ServerSocket(protocolFamily: .vsock, setNonBlocking: true)
+            try socket.bind(to: vsockAddress)
+            try socket.listen(backlog: 128)
+            return try ServerSocketChannel(serverSocket: socket, eventLoop: eventLoop, group: childEventLoopGroup)
+        }
+        return bind0(makeServerChannel: makeChannel) { (eventLoop, serverChannel) in
+            let promise = eventLoop.makePromise(of: Void.self)
+            serverChannel.registerAlreadyConfigured0(promise: promise)
+            return promise.futureResult
+        }
+    }
+    #endif
+
     #if !os(Windows)
         /// Use the existing bound socket file descriptor.
         ///
@@ -1263,6 +1282,50 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
             return self.group.next().makeFailedFuture(error)
         }
     }
+
+    #if canImport(Darwin) || os(Linux)
+    public func connect(to vsockAddress: VsockAddress) throws -> EventLoopFuture<Channel> {
+        let eventLoop = group.next()
+        let channelInitializer = self.channelInitializer
+        let socket: Socket
+        let channel: SocketChannel
+        do {
+            socket = try Socket(protocolFamily: .vsock, type: .stream)
+            if try !socket.connect(to: vsockAddress) {
+                try socket.finishConnect()
+            }
+            channel = try SocketChannel(socket: socket, eventLoop: eventLoop as! SelectableEventLoop)
+            channel.closeFuture.whenComplete() { _ in
+                try? socket.close()
+            }
+        } catch {
+            return eventLoop.makeFailedFuture(error)
+        }
+
+        func setupChannel() -> EventLoopFuture<Channel> {
+            eventLoop.assertInEventLoop()
+            return self._channelOptions.applyAllChannelOptions(to: channel).flatMap {
+                channelInitializer(channel)
+            }.flatMap {
+                eventLoop.assertInEventLoop()
+                let promise = eventLoop.makePromise(of: Void.self)
+                channel.registerAlreadyConfigured0(promise: promise)
+                return promise.futureResult
+            }.map {
+                channel
+            }.flatMapError { error in
+                channel.close0(error: error, mode: .all, promise: nil)
+                return channel.eventLoop.makeFailedFuture(error)
+            }
+        }
+
+        if eventLoop.inEventLoop {
+            return setupChannel()
+        } else {
+            return eventLoop.flatSubmit { setupChannel() }
+        }
+    }
+    #endif
 
     #if !os(Windows)
         /// Use the existing connected socket file descriptor.
--- a/Sources/NIOPosix/ServerSocket.swift
+++ b/Sources/NIOPosix/ServerSocket.swift
@@ -39,6 +39,10 @@ import NIOCore
         switch protocolFamily {
         case .unix:
             cleanupOnClose = true
+        #if canImport(Darwin) || os(Linux)
+        case .vsock:
+            cleanupOnClose = true
+        #endif
         default:
             cleanupOnClose = false
         }
--- a/Tests/NIOPosixTests/VsockAddressTest.swift
+++ b/Tests/NIOPosixTests/VsockAddressTest.swift
@@ -1,0 +1,41 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+@testable import NIOCore
+@testable import NIOPosix
+
+class VsockAddressTest: XCTestCase {
+
+    func testDescriptionWorks() throws {
+        XCTAssertEqual(VsockAddress(cid: .host, port: 12345).description, "[VSOCK]2:12345")
+        XCTAssertEqual(VsockAddress(cid: .any, port: 12345).description, "[VSOCK]-1:12345")
+    }
+
+    func testSocketAddressEqualitySpecialValues() throws {
+        XCTAssertEqual(VsockAddress(cid: .any, port: 12345), .init(cid: UInt32(bitPattern: -1), port: 12345))
+        XCTAssertEqual(VsockAddress(cid: .hypervisor, port: 12345), .init(cid: 0, port: 12345))
+        XCTAssertEqual(VsockAddress(cid: .host, port: 12345), .init(cid: 2, port: 12345))
+    }
+
+    func testSocketAddressEquality() throws {
+        XCTAssertEqual(VsockAddress(cid: 0, port: 0), .init(cid: 0, port: 0))
+        XCTAssertEqual(VsockAddress(cid: 1, port: 0), .init(cid: 1, port: 0))
+        XCTAssertEqual(VsockAddress(cid: 0, port: 1), .init(cid: 0, port: 1))
+
+        XCTAssertNotEqual(VsockAddress(cid: 0, port: 0), .init(cid: 1, port: 0))
+        XCTAssertNotEqual(VsockAddress(cid: 0, port: 0), .init(cid: 0, port: 1))
+    }
+}
+
--- a/Package.swift
+++ b/Package.swift
@@ -46,6 +46,7 @@ let package = Package(
             name: "NIOCore",
             dependencies: [
                 "NIOConcurrencyHelpers",
+                "CNIODarwin",
                 "CNIOLinux",
                 "CNIOWindows",
                 swiftCollections,
--- a/Sources/CNIODarwin/include/CNIODarwin.h
+++ b/Sources/CNIODarwin/include/CNIODarwin.h
@@ -16,6 +16,7 @@
 
 #ifdef __APPLE__
 #include <sys/socket.h>
+#include <sys/vsock.h>
 #include <time.h>
 
 // Darwin platforms do not have a sendmmsg implementation available to them. This C module
@@ -49,6 +50,8 @@ const void *CNIODarwin_CMSG_DATA(const struct cmsghdr *);
 void *CNIODarwin_CMSG_DATA_MUTABLE(struct cmsghdr *);
 size_t CNIODarwin_CMSG_LEN(size_t);
 size_t CNIODarwin_CMSG_SPACE(size_t);
+
+uint32_t CNIODarwin_get_local_vsock_cid(int socket);
 
 #endif  // __APPLE__
 #endif  // C_NIO_DARWIN_H
--- a/Sources/CNIOLinux/include/CNIOLinux.h
+++ b/Sources/CNIOLinux/include/CNIOLinux.h
@@ -29,6 +29,7 @@
 #include <netinet/ip.h>
 #include <netinet/udp.h>
 #include "liburing_nio.h"
+#include <linux/vm_sockets.h>
 
 #if __has_include(<linux/mptcp.h>)
 #include <linux/mptcp.h>
@@ -113,6 +114,8 @@ bool CNIOLinux_supports_udp_segment();
 bool CNIOLinux_supports_udp_gro();
 
 int CNIOLinux_system_info(struct utsname* uname_data);
+
+uint32_t CNIOLinux_get_local_vsock_cid(int socket);
 
 #endif
 #endif
--- a/Sources/CNIOLinux/shim.c
+++ b/Sources/CNIOLinux/shim.c
@@ -182,4 +182,20 @@ int CNIOLinux_system_info(struct utsname* uname_data) {
     return uname(uname_data);
 }
 
+/// For consistency with Darwin API, on which this is the only way to get the local CID.
+///
+/// While Linux and Darwin both support `IOCTL_VM_SOCKETS_GET_LOCAL_CID`, they operate on different
+/// special files.
+///
+/// - On Darwin, `IOCTL_VM_SOCKETS_GET_LOCAL_CID` is called with the socket to get the local CID.
+///
+/// - On Linux, `IOCTL_VM_SOCKETS_GET_LOCAL_CID` is called with `/dev/vsock` and, while it is a
+/// supported way to get the local CID, the man page encourages the use of `VMADDR_CID_LOCAL` instead.
+///
+/// This Linux API exists solely for consistency with Darwin platforms and, as such, makes no use of
+/// the parameter and just returns `VMADDR_CID_LOCAL`.
+uint32_t CNIOLinux_get_local_vsock_cid(int __unused__) {
+    return VMADDR_CID_LOCAL;
+}
+
 #endif
--- a/Sources/NIOEchoClient/main.swift
+++ b/Sources/NIOEchoClient/main.swift
@@ -24,7 +24,7 @@ private final class EchoHandler: ChannelInboundHandler {
     private var receiveBuffer: ByteBuffer = ByteBuffer()
     
     public func channelActive(context: ChannelHandlerContext) {
-        print("Client connected to \(context.remoteAddress!)")
+        print("Client connected to \(context.remoteAddress?.description ?? "unknown")")
         
         // We are connected. It's time to send the message to the server to initialize the ping-pong sequence.
         let buffer = context.channel.allocator.buffer(string: line)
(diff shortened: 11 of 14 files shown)


simonjbeaumont:
Hm, looking at the CI failures. Locally the following works:

```console
% docker-compose -f ./docker/docker-compose.yaml -f ./docker/docker-compose.2204.59.yaml run shell
root@aaf1579beac7:/code# rm -rf .build
root@aaf1579beac7:/code# swift run NIOEchoServer -1 12345
...
Server started and listening on [VSOCK]-1:12345
```

...so does building the tests in the same container:

```console
root@aaf1579beac7:/code# swift build --build-tests
...
[433/433] Linking swift-nioPackageTests.xctest
Build complete! (47.44s)
```

...but in CI we get: 

```log
12:44:26 /code/Sources/NIOPosix/Bootstrap.swift:336:33: error: cannot convert value of type 'VsockAddress' to expected argument type 'SocketAddress'
12:44:26             try socket.bind(to: vsockAddress)
12:44:26                                 ^
12:44:26 error: fatalError
```

I'll investigate.

simonjbeaumont:
I'm also seeing this in the CI logs...

```
12:43:23 + uname -a
12:43:23 Linux 43442937c113 3.10.0-1160.15.2.el7.x86_64 #1 SMP Wed Feb 3 15:06:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

...which poses a bit of a problem, because we might need a newer Linux Kernel version:

```
VERSIONS

       Support for VMware (VMCI) has been available since Linux 3.9.
       KVM (virtio) is supported since Linux 4.8.  Hyper-V is supported
       since Linux 4.14.

       VMADDR_CID_LOCAL is supported since Linux 5.6.  Local
       communication in the guest and on the host is available since
       Linux 5.6.  Previous versions supported only local communication
       within a guest (not on the host), and with only some transports
       (VMCI and virtio).
```

Sources/CNIODarwin/shim.c
```diff
@@ -90,4 +91,9 @@ const int CNIODarwin_IPTOS_ECN_CE = IPTOS_ECN_CE;
 const int CNIODarwin_IPV6_RECVPKTINFO = IPV6_RECVPKTINFO;
 const int CNIODarwin_IPV6_PKTINFO = IPV6_PKTINFO;
 
+uint32_t CNIODarwin_get_local_vsock_cid(int socket) {
+    uint32_t cid = 0;
+    int result = ioctl(socket, IOCTL_VM_SOCKETS_GET_LOCAL_CID, &cid);
```
Lukasa, line 96: This call doesn't check the return value, which is unlikely to be right. Can we pass the `cid` as an out parameter and return a success or failure boolean? Similarly, this needs to loop on `EINTR`.

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+        ///
+        /// The local context ID obtained with ``getLocalContextID(_:)`` can be used for the same
+        /// purpose, but it is preferable to use ``local``.
+        public static let local: Self = Self(rawValue: UInt32(VMADDR_CID_LOCAL))
+#endif
+
+        /// Get the context ID of the local machine.
+        ///
+        /// On Linux, consider using ``local`` when binding instead of this function.
+        public static func getLocalContextID(_ socket: NIOBSDSocket.Handle) -> Self {
```
Lukasa, line 134: I don't think we need this function to be public.

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+
+    /// Creates a new vsock address.
+    ///
+    /// - parameters:
+    ///     - address: the `sockaddr_vm` that holds the context ID and port.
+    public init(address: sockaddr_vm) {
+        self._storage = Box(address)
+    }
+
+    public init(cid: UInt32, port: UInt32) {
```
Lukasa, line 63: Why do these two initializers not use your `ContextID` and `Port` types?

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+    /// The libc socket address for a vsock socket.
+    public var address: sockaddr_vm { return _storage.value }
+
+    /// Get the context ID associated with the address.
+    public var cid: Int {
+        Int(Int32(bitPattern: self.address.svm_cid))
+    }
+
+    /// Get the port associated with the address.
+    public var port: Int {
```
Lukasa, line 51: Why do these return `Int` instead of `ContextID` and `Port`?

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+/// A socket address is defined as a combination of a Context Identifier (CID) and a port number.
+/// The CID identifies the source or destination, which is either a virtual machine or the host.
+/// The port number differentiates between multiple services running on a single machine.
+///
+/// For well-known CID values and port numbers, see ``ContextID`` and ``Port`` respectively.
+public struct VsockAddress: Sendable {
+    private let _storage: Box<sockaddr_vm>
+
+    /// The libc socket address for a vsock socket.
+    public var address: sockaddr_vm { return _storage.value }
```
Lukasa, line 43: For now I think my preference is not to make this API, so we avoid committing to a non-portable representation here. That would make it easy to have this type on Windows if we need it.

Lukasa: Relatedly, can I propose we create this lazily and avoid the use of `Box`? We only need `Box` for `SocketAddress` to deal with existential values, but `ContextID` and `Port` together as stored properties should fit nicely.

Sources/NIOCore/VsockAddress.swift
Lukasa: Relatedly, these two can be mutable.

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+        self.init(cid: UInt32(bitPattern: Int32(truncatingIfNeeded: cid)), port: UInt32(bitPattern: Int32(truncatingIfNeeded: port)))
+    }
+
+    /// Get the local context ID for a vsock socket.
+    ///
+    /// - parameters:
+    ///   - socket: The vsock socket to get the local context ID for.
+    ///
+    /// - NOTE: On Linux, you can use `VMADDR_CID_LOCAL` to retrieve the local context ID, which does not require a socket.
+    public static func localVsockContextID(_ socket: NIOBSDSocket.Handle) -> Int {
```
Lukasa, line 87: We don't need this to be public.

Sources/NIOCore/VsockAddress.swift
```diff
@@ -0,0 +1,221 @@
...
+}
+
+extension VsockAddress: CustomStringConvertible {
+    public var description: String {
+        "[VSOCK]\(self.cid):\(self.port)"
+    }
+}
+
+/// Element-wise Equatable conformance using only the struct fields defined in the man page (excluding lengths).
+extension VsockAddress: Equatable {
```
Lukasa, line 175: If we change the storage representation to only store context ID and port then we can use the synthesised Equatable and Hashable implementations.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -326,6 +326,25 @@ public final class ServerBootstrap {
         return self.bind(unixDomainSocketPath: unixDomainSocketPath)
     }
 
+    #if canImport(Darwin) || os(Linux)
+    public func bind(to vsockAddress: VsockAddress) throws -> EventLoopFuture<Channel> {
```
Lukasa, line 330: Can we add a doc comment here please?

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -326,6 +326,25 @@ public final class ServerBootstrap {
...
+        func makeChannel(_ eventLoop: SelectableEventLoop, _ childEventLoopGroup: EventLoopGroup, _ enableMPTCP: Bool) throws -> ServerSocketChannel {
+            if enableMPTCP {
+                throw ChannelError.operationUnsupported
+            }
+            let socket = try ServerSocket(protocolFamily: .vsock, setNonBlocking: true)
+            try socket.bind(to: vsockAddress)
+            try socket.listen(backlog: 128)
```
Lukasa, line 337: Preferentially we'd have the `Channel` be responsible for doing this: that is, instead of calling `bind` and `listen` here we'd have the `Channel` do it in `bind0`. This would use a user event to trigger the operation, as we do in NIOTS.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1264,6 +1283,50 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
     }
 
+    #if canImport(Darwin) || os(Linux)
+    public func connect(to vsockAddress: VsockAddress) throws -> EventLoopFuture<Channel> {
+        let eventLoop = group.next()
+        let channelInitializer = self.channelInitializer
+        let socket: Socket
+        let channel: SocketChannel
+        do {
+            socket = try Socket(protocolFamily: .vsock, type: .stream)
```
Lukasa, line 1293: Preferentially we'd have the `Channel` be responsible for doing this: that is, instead of calling `connect` here we'd have the `Channel` do it in `connect0`. This would use a user event to trigger the operation, as we do in NIOTS.

Sources/NIOPosix/ServerSocket.swift
```diff
@@ -39,6 +39,10 @@ import NIOCore
         switch protocolFamily {
         case .unix:
             cleanupOnClose = true
+        #if canImport(Darwin) || os(Linux)
+        case .vsock:
+            cleanupOnClose = true
```
Lukasa, line 44: Why do we do cleanup on close here?

Sources/CNIODarwin/shim.c
simonjbeaumont: Done in the latest round of commits.

simonjbeaumont (commented):
I've addressed most of the comments in the latest round of commits—I think all of them that relate to the API, although there's some open discussion around the local context ID.

What remains is for me to try and use the user event to trigger the bind, rather than doing it in the bootstrap itself.

Sources/NIOCore/VsockAddress.swift
Lukasa: Yes, correct. Those types are constructable from integers.

Sources/NIOCore/VsockAddress.swift
Lukasa: The user doesn't need to do that, just add an initializer that does it.

Sources/NIOCore/VsockAddress.swift
Lukasa: Is there any reason to allow this public API to be anything other than a channel option?

Sources/NIOCore/VsockAddress.swift
Lukasa: Yeah, `SocketAddress` making the various `sockaddr` stuff API was a historical mistake we cannot unwind.

Yes, I was assuming (1).

Sources/NIOPosix/Bootstrap.swift
simonjbeaumont: I've pushed a commit which moves all of the logic to a user event.

There came a point where I had to bridge the ability to connect and bind to either a `SocketAddress` or a `VsockAddress`.

I ended up adding an overload of `connectSocket(to:)` in `BaseSocketChannel` and, consequently, implementations in `PipeChannel` and `DatagramChannel` which return `operationUnsupported`.

I then added a `final` function to `BaseSocketChannel` to bridge this functionality with an internal enum type.

It didn't seem as complicated with the bind side, because there connect had the non-blocking flow to handle.

Hopefully I got the right end of the stick here!

Sources/NIOCore/VsockAddress.swift
simonjbeaumont: What's the semantics we're trying to achieve that would be covered by a channel option?

IIUC we need this to be something a user can _get_ to support being able to connect to the local CID on platforms that don't have `VMADDR_CID_LOCAL` (Linux < 5.6, Darwin).

If this were a channel option, what address would users provide to `connect(to:)`?

I agree that the primary use case right now is testing, so I'd ideally make it `internal`. What blocks that is the split of the logic over `NIOCore` and `NIOPosix`.

Lukasa: We support get-only Channel Options, which are what we're looking for here I think.

Sources/NIOCore/VsockAddress.swift
simonjbeaumont: Makes sense. However, do we even want to provide an initialiser from `Int`?

Both `ContextID` and `Port` have `init(rawValue: UInt32)` and `init(integerValue: UInt32)` to satisfy `RawRepresentable` and `ExpressibleByIntegerLiteral` respectively.

If we were to provide an init from `Int` how would we spell it to communicate the semantics that they'll be some conversion going on?

- `init?(_: Int)`
- `init(_: Int) throws`
- `init(truncatingIfNecessary: Int)`

Maybe the best thing to do is to _not_ provide one?

Sources/CNIODarwin/shim.c
simonjbeaumont: Worth noting that, if we really want to do this using a new `System.ioctl`, that's going to be in `NIOPosix` so we'd need to move the API for getting the local Context ID out of `NIOCore` completely.

Sources/NIOCore/VsockAddress.swift
simonjbeaumont: Added an get-only option in the latest commit. 

Note that the `ChannelOption` is in `NIOPosix` but relies on a function in NIOCore so I had to keep one of these APIs public.

simonjbeaumont:
Note the tests will be skipped in CI because the kernel is too old. Here is an example run with the latest commit in a Linux environment:

```console
% swift test --filter Vsock
Build complete! (27.46s)
Test Suite 'Selected tests' started at 2023-07-25 16:05:03.138
Test Suite 'EchoServerClientTest' started at 2023-07-25 16:05:03.154
Test Case 'EchoServerClientTest.testEchoVsock' started at 2023-07-25 16:05:03.154
Test Case 'EchoServerClientTest.testEchoVsock' passed (0.051 seconds)
Test Suite 'EchoServerClientTest' passed at 2023-07-25 16:05:03.206
         Executed 1 test, with 0 failures (0 unexpected) in 0.051 (0.051) seconds
Test Suite 'VsockAddressTest' started at 2023-07-25 16:05:03.206
Test Case 'VsockAddressTest.testDescriptionWorks' started at 2023-07-25 16:05:03.206
Test Case 'VsockAddressTest.testDescriptionWorks' passed (0.0 seconds)
Test Case 'VsockAddressTest.testGetLocalCID' started at 2023-07-25 16:05:03.206
Test Case 'VsockAddressTest.testGetLocalCID' passed (0.001 seconds)
Test Case 'VsockAddressTest.testSocketAddressEquality' started at 2023-07-25 16:05:03.207
Test Case 'VsockAddressTest.testSocketAddressEquality' passed (0.0 seconds)
Test Case 'VsockAddressTest.testSocketAddressEqualitySpecialValues' started at 2023-07-25 16:05:03.207
Test Case 'VsockAddressTest.testSocketAddressEqualitySpecialValues' passed (0.0 seconds)
Test Suite 'VsockAddressTest' passed at 2023-07-25 16:05:03.207
         Executed 4 tests, with 0 failures (0 unexpected) in 0.001 (0.001) seconds
Test Suite 'Selected tests' passed at 2023-07-25 16:05:03.207
         Executed 5 tests, with 0 failures (0 unexpected) in 0.052 (0.052) seconds
```

Sources/NIOCore/VsockAddress.swift
Lukasa: There's no need to: we can simply crash, as the transformation from `Int` will do.

Sources/NIOCore/VsockAddress.swift
Lukasa: I think moving the whole API surface into NIOPosix is the right thing to do.

simonjbeaumont: OK, I've now moved everything to `NIOPosix` and made both `getLocalContextID(socketFD:)` and `getLocalContextID()` internal, leaving just `ChannelOptions.localVsockContextID` as the public API for getting the local context ID.

Sources/NIOCore/VsockAddress.swift
simonjbeaumont: Currently it feels like we have a safer API and, because we are conforming to `ExprissibleByIntegerLiteral` we have the ergonomics without the runtime crash:

```swift
_ = VsockAddress(cid: 4294967295, port: 12345)  // OK
_ = VsockAddress(cid: 4294967296, port: 12345)  // error: Integer literal '4294967296' overflows when stored into 'VsockAddress.ContextID'
```

Lukasa: Oh sorry, I meant an init from `Int` on the `ContextID` and `Port`, not on anywhere else.

Lukasa: I am indeed.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -326,6 +326,28 @@ public final class ServerBootstrap {
         return self.bind(unixDomainSocketPath: unixDomainSocketPath)
     }
 
+    #if canImport(Darwin) || os(Linux)
+    /// Bind the `ServerSocketChannel` to a VSOCK socket.
+    ///
+    /// - parameters:
+    ///   - vsockAddress: The VSOCK socket address to bind on.
+    public func bind(to vsockAddress: VsockAddress) -> EventLoopFuture<Channel> {
```
weissi, line 334: this is public API so shouldn't be `#if os(...)`'d. In the body you can `#if os(...)` and just return an unsupported operation error.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -993,6 +1015,33 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
         }
     }
 
+    #if canImport(Darwin) || os(Linux)
+    /// Specify the VSOCK address to connect to for the `Channel`.
+    ///
+    /// - parameters:
+    ///     - address: The VSOCK address to connect to.
+    /// - returns: An `EventLoopFuture<Channel>` for when the `Channel` is connected.
+    public func connect(to address: VsockAddress) -> EventLoopFuture<Channel> {
```
weissi, line 1024: this is public API so shouldn't be `#if os(...)`'d. In the body you can `#if os(...)` and just return an unsupported operation error.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -119,9 +116,25 @@ final class SocketChannel: BaseStreamSocketChannel<Socket> {
...
+
+    override func connectSocket(to address: SocketAddress) throws -> Bool {
+        if try self.socket.connect(to: address) {
+            return true
+        }
+        self.scheduleConnectTimeout()
+        return false
+    }
 
+#if canImport(Darwin) || os(Linux)
```
weissi, line 129: why are these `#if`'d?

Sources/NIOPosix/System.swift
```diff
@@ -153,6 +153,9 @@ private let sysKevent = kevent
 private let sysSendMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_mmsghdr>?, CUnsignedInt, CInt) -> CInt = CNIODarwin_sendmmsg
 private let sysRecvMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_mmsghdr>?, CUnsignedInt, CInt, UnsafeMutablePointer<timespec>?) -> CInt = CNIODarwin_recvmmsg
 #endif
+#if canImport(Darwin) || os(Linux)
```
weissi, line 156: `ioctl` should be available ~everywhere? If Windows doesn't have it maybe `#if !os(Windows)`?

Sources/NIOPosix/System.swift
```diff
@@ -861,6 +864,14 @@ internal enum Posix {
         }
     }
 #endif
+#if canImport(Darwin) || os(Linux)
```
weissi, line 867: same here

Sources/NIOPosix/Bootstrap+vsock.swift
```diff
@@ -0,0 +1,48 @@
...
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+#if canImport(Darwin) || os(Linux)
```
weissi, line 14: unnecessary `#if`. Also, there is public API `#if`'d which shouldn't be done

weissi (commented):
Thanks, this looks mostly good but I think almost all the `#if`s should go.

`#if` is a last resort for things that cannot be expressed in a different way. Especially if it's around whole types/functions. If something doesn't work on a certain platform then we should throw an an 'unsupported operation' error.

It's important to compile as much code on as many platforms as possible

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,256 @@
...
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+#if canImport(Darwin) || os(Linux)
```
simonjbeaumont, line 14: @weissi could you advise how we should structure these changes to safely not break builds on other platforms?

This file contains the definition of the `VsockAddress` type, which is used in the new public APIs, e.g. `Bootstrap.connect(to vsockAddress:)`. If we remove the guard from the public API, then `VsockAddress` also needs to be unconditionally available.

However, this file also contains the various extensions required on other types, e.g. `NIOBSDSocket`, `ProtocolFamily`, `AddressFamily`, `sockaddr_vm`, `sockaddr_storage`, `BaseSocket`, and `ChannelOptions`.

If I just remove the `#if canImport(Darwin) || or os(Linux) from this line, then I expect it will just fail to compile on other platforms.

Some of these require the C library be present, which is why this file has the (almost) usual import dance. What it does _not_ currently have is the `#error` I see in other places:

```swift
#error("The ... module was unable to identify your C library.")
```

Should we be having this `#error` here and do I need to expand the import dance to include Windows, even though we will probably throw an error in the  body of the APIs?

I'm a bit nervous about breaking the build for other platforms.

Sources/NIOPosix/SocketChannel.swift
simonjbeaumont: Because they are `#if`'d in `BaseSocketChannel`.

I'll try and remove them from everywhere and see how I get on.

Sources/NIOPosix/System.swift
simonjbeaumont: Right, again, I was avoiding adding things on platforms I wasn't testing the build on.

Looking now it seems the spelling is slightly different in e.g. Musl:

```c
int ioctl (int, int, ...);
```
(source: https://git.musl-libc.org/cgit/musl/tree/include/sys/ioctl.h)

cf.

```c
int ioctl(int, unsigned long, ...);
```

Windows looks to be different further—possibly it's `DeviceIoControl`—but this isn't something I know about.

I can try and do the dance to get Musl in there.

Sources/NIOPosix/Bootstrap.swift
weissi: > withBoundSocket(descriptor:) which does use #if !os(...) guard to not include it in the API.

`withBoundSocket` is an interesting case because it's an overload. There is always

```
public func withBoundSocket(_ socket: NIOBSDSocket.Handle) -> EventLoopFuture<Channel>
```

and on non-Windows there is also the deprecated

```
        @available(*, deprecated, renamed: "withBoundSocket(_:)")
        public func withBoundSocket(descriptor: CInt) -> EventLoopFuture<Channel> {
```

When Windows support was started, we wanted to migrate everybody to the new, former one because it works everywhere but because of public API we could only deprecate it of course (and not remove it). But because Windows had not been supported at all, we could actually remove it (by `#if`) on Windows.

Was it 100% necessary to `#if` it out on Windows? No. But it's also not a big issue because the API still exists, just with slightly different types. So I think that makes some sense.

Sources/NIOPosix/Bootstrap.swift
weissi: Yeah same reason as above

Sources/NIOPosix/SocketChannel.swift
weissi: thank you, yeah, let's push the `#if`s as far in as we can. The less `#if`'d code the better

Sources/NIOPosix/System.swift
weissi: fair enough, that sounds okay actually, `System.swift` is also very much in the guts. Happy for you to leave the `#if` if that makes it easier

Sources/NIOPosix/Bootstrap+vsock.swift
weissi: thanks so much! Yes, unavailable API is terrible for users because they'd need to replicate the `#if`s. If it doesn't work and throws an error that's much better.

You know if you tell Vapor to listen on a vsock on some unsupported platform and you get "unsupportedOperation" that's cool. If Vapor doesn't build or everybody has to perfectly copy the `#if`s into higher and higher level libraries/apps then that sucks.

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,256 @@
...
+        /// applications on a single host and for debugging.
+        ///
+        /// The local context ID obtained with `getLocalContextID(_:)` can be used for the same
+        /// purpose, but it is preferable to use `local`.
+        ///
+        /// This is equal to `VMADDR_CID_LCOAL`.
+        ///
+        /// - Seealso: https://man7.org/linux/man-pages/man7/vsock.7.html
+        public static let local: Self = Self(rawValue: 1)
+#endif
```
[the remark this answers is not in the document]
weissi: I'd probably just do a `Self(rawValue: 1)`. If it turns out to be wrong on other platforms, we can `#if` the correct value in there when that happens.

Sources/NIOPosix/VsockAddress.swift
weissi: Everything but `sockaddr_vm` should be available on all supported platforms. For `extension sockaddr_vm` I think it's fine to put an `#if` guard around that.

Sources/NIOPosix/VsockAddress.swift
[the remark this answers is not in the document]
simonjbeaumont: This comment also applies to the constants that are available on both Linux and Darwin, if we include this type on all platforms, e.g. Windows.

Specifically we have `.any`, `.host`, and `.hypervisor`, which make use of `VMADDR_CID_ANY`, `VMADDR_CID_HOST`, and `VMADDR_CID_HYPERVISOR`, respectively.

IIUC we have a few options:
1. Make these all computed throwing properties and `#if` in the body.
2. Don't use the constants imported from C, and instead hardcode their values.
3. Remove them altogether.
4. Move them into an extension which is guarded by an `#if`.

Thoughts, @weissi?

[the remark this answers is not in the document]
weissi: I'd say 2 or 3 :)

simonjbeaumont:
@weissi @Lukasa I've pushed the `#if` guards inside the API. This has resulted in one of three kinds of thing:
1. Public API that work everywhere and have no `#if` guards for any platforms.
2. Public API that only work on some platforms. These have the `#if` guard pushed into the implementation and result in a runtime error for the caller on other platforms.
3. Internal functions that only serve to support the supported platforms. These have the `#if` guard on the outside.

The latest commit includes these changes.

I'd be very grateful for someone to double check I've done this correctly 🙏

Sources/NIOEchoServer/main.swift
```diff
@@ -70,18 +70,25 @@ let defaultPort = 9999
...
 }
 
 let bindTarget: BindTo
 switch (arg1, arg1.flatMap(Int.init), arg2.flatMap(Int.init)) {
-case (.some(let h), _ , .some(let p)):
-    /* we got two arguments, let's interpret that as host and port */
+case (_, .some(let cid), .some(let port)):
+    /* we got two arguments (Int, Int), let's interpret that as vsock cid and port */
+    bindTarget = .vsock(VsockAddress(
+        cid: .init(rawValue: UInt32(bitPattern: Int32(truncatingIfNeeded: cid))),
```
weissi, line 81: nit, I'd suggest using the type name here instead of `.init` unless this is some cunning have the compiler pick the right, platform dependent type trick

Sources/NIOPosix/BaseSocketChannel.swift
```diff
@@ -1218,6 +1253,10 @@ class BaseSocketChannel<SocketType: BaseSocketProtocol>: SelectableChannel, Chan
         }
     }
 
+    public final func connect0(to address: SocketAddress, promise: EventLoopPromise<Void>?) {
```
weissi, line 1256: total nit, to make the diff easier to read, I'd move this implementation up and the new one down :)

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,290 @@
...
+    /// - Note: On Linux, ``local`` may be a better choice.
+    static func getLocalContextID(_ socketFD: NIOBSDSocket.Handle) throws -> Self {
+#if canImport(Darwin)
+        let request = CNIODarwin_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = socketFD
+#elseif os(Linux)
+        let request = CNIOLinux_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = open("/dev/vsock", O_RDONLY)
+        precondition(fd >= 0, "couldn't open /dev/vsock (\(errno))")
+        defer { close(fd) }
```
weissi, line 225: should use `try! sysClose`

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,290 @@
...
+    ///
+    /// - Note: On Linux, ``local`` may be a better choice.
+    static func getLocalContextID(_ socketFD: NIOBSDSocket.Handle) throws -> Self {
+#if canImport(Darwin)
+        let request = CNIODarwin_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = socketFD
+#elseif os(Linux)
+        let request = CNIOLinux_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = open("/dev/vsock", O_RDONLY)
+        precondition(fd >= 0, "couldn't open /dev/vsock (\(errno))")
```
weissi, line 224: this is unnecessary with `sysOpen` and you can just `try!` that then

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,290 @@
...
+    /// - On Linux, the `ioctl()` request operates on the `/dev/vsock` device.
+    ///
+    /// - Note: On Linux, ``local`` may be a better choice.
+    static func getLocalContextID(_ socketFD: NIOBSDSocket.Handle) throws -> Self {
+#if canImport(Darwin)
+        let request = CNIODarwin_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = socketFD
+#elseif os(Linux)
+        let request = CNIOLinux_IOCTL_VM_SOCKETS_GET_LOCAL_CID
+        let fd = open("/dev/vsock", O_RDONLY)
```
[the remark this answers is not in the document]
weissi: somewhat surprising that `ioctl`'ing counts as read-only :)

Sources/NIOPosix/VsockAddress.swift
weissi: or `Posix.open`, can't remember the naming

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,290 @@
...
+        var cid = Self.any.rawValue
+        try Posix.ioctl(fd: fd, request: request, ptr: &cid)
+        precondition(cid != Self.any.rawValue)
+        return Self(rawValue: cid)
+    }
+
+    /// Get the context ID of the local machine.
+    static func getLocalContextID() throws -> Self {
+        let socket = try Socket(protocolFamily: .vsock, type: .stream)
+        defer { try? socket.close() }
```
weissi, line 236: `try!`, if open works, close must work or else we have a big issue (somebody closed our fd under our feat)

Tests/NIOPosixTests/VsockAddressTest.swift
```diff
@@ -0,0 +1,80 @@
...
+        XCTAssertEqual(VsockAddress.ContextID(integerLiteral: 4294967295), 4294967295)
+        XCTAssertEqual(VsockAddress.Port(integerLiteral: 4294967295), 4294967295)
+    }
+
+    func testInitializeFromInt() throws {
+        XCTAssertEqual(VsockAddress.ContextID(0), 0)
+        XCTAssertEqual(VsockAddress.ContextID(4294967295), 4294967295)
+        XCTAssertEqual(VsockAddress.Port(0), 0)
+        XCTAssertEqual(VsockAddress.Port(4294967295), 4294967295)
+//        XCTAssertEqual(VsockAddress.ContextID(-1), 4294967295)  // compile-time error
```
weissi, line 39: some comments left

weissi (commented):
looks super good, thanks. Left a few comments, nothing major

Sources/NIOEchoServer/main.swift
simonjbeaumont: Done (Picked this habit up from another project which seemed to favour it).

Sources/NIOPosix/BaseSocketChannel.swift
simonjbeaumont: Done.

Sources/NIOPosix/VsockAddress.swift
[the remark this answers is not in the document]
simonjbeaumont: > * should use `sysOpen` and `| O_CLOEXEC`

Done: used `Posix.open` with `O_CLOEXEC`.

> somewhat surprising that `ioctl`'ing counts as read-only :)

🤷 The tests pass, so I guess it's fine in this case. The very nature of this `ioctl` is a "getter"; presumably other `ioctl` requests may require the fd be RW.

Sources/NIOPosix/VsockAddress.swift
simonjbeaumont: > this is unnecessary with `sysOpen` and you can just `try!` that then

Done.

Sources/NIOPosix/VsockAddress.swift
simonjbeaumont: Done.

Sources/NIOPosix/VsockAddress.swift
simonjbeaumont: Done—Good to know. Guess I need to tune into when it's OK to `try!`.

Tests/NIOPosixTests/VsockAddressTest.swift
simonjbeaumont: Left them there for posterity, but I removed now.

simonjbeaumont (commented):
@weissi thanks for taking the time to review. I addressed all the feedback.

Sources/NIOPosix/System.swift
```diff
@@ -153,6 +153,13 @@ private let sysKevent = kevent
 private let sysSendMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_mmsghdr>?, CUnsignedInt, CInt) -> CInt = CNIODarwin_sendmmsg
 private let sysRecvMmsg: @convention(c) (CInt, UnsafeMutablePointer<CNIODarwin_mmsghdr>?, CUnsignedInt, CInt, UnsafeMutablePointer<timespec>?) -> CInt = CNIODarwin_recvmmsg
 #endif
+#if !os(Windows)
+#if canImport(Musl)
```
glbrntt, line 157: Just to check: is the signature different for musl vs. glibc?

Sources/NIOPosix/System.swift
```diff
@@ -861,6 +868,14 @@ internal enum Posix {
         }
     }
 #endif
+#if !os(Windows)
+    @inline(never)
+    internal static func ioctl(fd: CInt, request: CUnsignedLong, ptr: UnsafeMutableRawPointer) throws {
+        _ = try syscall(blocking: false) {
+            sysIoctl(fd, request, ptr)
```
glbrntt, line 875: Is `request` is a `CInt` for musl then I think we'll need a `numericCast` here

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,289 @@
...
+
+// MARK: - Public API that's available on all platforms.
+
+/// A vsock socket address.
+///
+/// A socket address is defined as a combination of a Context Identifier (CID) and a port number.
+/// The CID identifies the source or destination, which is either a virtual machine or the host.
+/// The port number differentiates between multiple services running on a single machine.
+///
+/// For well-known CID values and port numbers, see `VsockAddress.ContextID` and `VsockAddress.Port` respectively.
```
glbrntt, line 36: I think for docc you want:

```suggestion
/// For well-known CID values and port numbers, see ``VsockAddress/ContextID`` and ``VsockAddress/Port`` respectively.
```

Sources/NIOPosix/VsockChannelEvents.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// This flow for connect is required because we cannot extend `enum SocketAddress` without
+    /// breaking public API.
+    struct ConnectToAddress: VsockChannelEvent {
+        public var address: VsockAddress
+
+        public init(_ address: VsockAddress) {
+            self.address = address
+        }
+    }
+}
```
glbrntt, line 45: Are these supposed to be public or internal?

simonjbeaumont (commented):
Thanks @glbrntt—I've updated the PR, if you want to take another look 🙏

Sources/NIOPosix/VsockAddress.swift
glbrntt: That's interesting and sounds a lot like a bug...

glbrntt (approved):
Looks good to me! Slightly hesitant to merge new API without @Lukasa also approving though.

Sources/NIOPosix/VsockAddress.swift
Lukasa: It's not a bug, you need a proper reference. Docc IDs are case insensitive, so you need to disambiguate between `VsockAddress.Port` and `VsockAddress.port`. If you let Xcode autocomplete type this for you, it'll put the right thing in.

Sources/NIOPosix/VsockAddress.swift
```diff
@@ -0,0 +1,289 @@
...
+        defer { try! Posix.close(descriptor: fd) }
+#endif
+        var cid = Self.any.rawValue
+        try Posix.ioctl(fd: fd, request: request, ptr: &cid)
+        precondition(cid != Self.any.rawValue)
+        return Self(rawValue: cid)
+    }
+
+    /// Get the context ID of the local machine.
+    static func getLocalContextID() throws -> Self {
```
Lukasa, line 233: Do we need this version? In all cases where we want to call this we have a socket already, so it seems like we should prefer the other one.

Sources/NIOPosix/VsockChannelEvents.swift
```diff
@@ -0,0 +1,43 @@
...
+
+/// A tag protocol that can be used to cover events used to boostrap VSOCK channels.
+protocol VsockChannelEvent: Hashable, Sendable {}
+
+enum VsockChannelEvents {
+    /// Fired as an outbound event when NIO would like to ask itself to bind the socket.
+    ///
+    /// This flow for connect is required because we cannot extend `enum SocketAddress` without
+    /// breaking public API.
+    struct BindToAddress: VsockChannelEvent {
```
Lukasa, line 24: Make these two events public please.

Sources/NIOPosix/VsockChannelEvents.swift
```diff
@@ -0,0 +1,43 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import NIOCore
+
+/// A tag protocol that can be used to cover events used to boostrap VSOCK channels.
+protocol VsockChannelEvent: Hashable, Sendable {}
```
Lukasa, line 17: Do we need the tag protocol?

Lukasa (changes requested):

Sources/NIOPosix/VsockAddress.swift
simonjbeaumont: > you need to disambiguate between VsockAddress.Port and VsockAddress.port. If you let Xcode autocomplete type this for you, it'll put the right thing in.

Well, Xcode autocomplete puts in <code>\`\`VsockAddress/Port\`\`</code> and resolves to the correct thing (jump to definition works fine).

Is there some other disambiguation I can use?

---

Here's the evidence that this doesn't work in CI (which I already tried).

Works fine currently:

```console
❯ docker-compose -f ./docker/docker-compose.yaml -f ./docker/docker-compose.2204.58.yaml run documentation-check
...
Generating documentation for 'NIOPosix'...
Building for debugging...
Build complete! (0.28s)
...

❯ echo $?
0
```

Then with this proposed change (even with Xcode autocompleting):

```console
❯ git diff
diff --git a/Sources/NIOPosix/VsockAddress.swift b/Sources/NIOPosix/VsockAddress.swift
index add18ffe3..442f938b6 100644
--- a/Sources/NIOPosix/VsockAddress.swift
+++ b/Sources/NIOPosix/VsockAddress.swift
@@ -33,7 +33,7 @@ fileprivate let vsockUnimplemented = "VSOCK support is not implemented for this
 /// The CID identifies the source or destination, which is either a virtual machine or the host.
 /// The port number differentiates between multiple services running on a single machine.
 ///
-/// For well-known CID values and port numbers, see `VsockAddress.ContextID` and `VsockAddress.Port` respectively.
+/// For well-known CID values and port numbers, see ``VsockAddress/ContextID`` and ``VsockAddress/Port`` respectively.
 public struct VsockAddress: Hashable, Sendable {
     /// The context ID associated with the address.
     public var cid: ContextID

❯ docker-compose -f ./docker/docker-compose.yaml -f ./docker/docker-compose.2204.58.yaml run documentation-check
...
Generating documentation for 'NIOPosix'...
Building for debugging...
Build complete! (0.27s)
Converting documentation...
/code/Sources/NIOPosix/VsockAddress.swift:36:84: error: Topic reference 'VsockAddress/Port' couldn't be resolved. No local documentation matches this reference.
Error: An error was encountered while compiling documentation
error: 'docc convert' invocation failed with a nonzero exit code: '1'

❯ echo $?
1
```

simonjbeaumont (commented):
@Lukasa I addressed most of the comments. The two that remain I responded to:

- https://github.com/apple/swift-nio/pull/2479/files#r1281514025
- https://github.com/apple/swift-nio/pull/2479/files#r1281511252

Sources/NIOPosix/VsockAddress.swift
simonjbeaumont: Update: found the syntax, which appears to be <code>\`\`VsockAddress/Port-swift.struct\`\`</code>

Sources/NIOPosix/VsockChannelEvents.swift
simonjbeaumont: Related to other thread, in which they were then made public: https://github.com/apple/swift-nio/pull/2479#discussion_r1281530933.

simonjbeaumont (commented):
OK, I think that's everything @Lukasa. Thanks for your patience 🙏

Lukasa (approved):

allevato:
I stumbled upon this PR because we were importing the latest swift-nio changes into our monorepo, where we're building exclusively with explicit modules and strict layering (`-fmodules-strict-decluse`). This change causes the following error in that mode:

```
error: module CNIODarwin does not depend on a module exporting 'sys/vsock.h'
#include <sys/vsock.h>
         ^
```

Unfortunately, it appears that this is because no module in Xcode's SDK includes `sys/vsock.h` (it's not in Darwin, nor in any other module defined by the module maps there). In a less strict/implicit module build, it would just get textually included.

We can probably reduce the strictness here on our side for this specific module, but do any of the Apple folks have thoughts here? I don't know what the state of explicit modules is for Swift/SPM builds, or whether they're planning to use `-fmodules-strict-decluse` as well—if so, this inclusion could cause issues there as well.

shahzadmajeed:
I'm seeing same build error on my end

shahzadmajeed:
@allevato what strictness level (flag to be specific) fixes this issue for you and how/where do you add that?

Lukasa:
Thanks folks. SwiftNIO has never formally promised to build with non-standard build flags, and this is one of those cases. Can I recommend that you use Feedback Assistant to ask for this header to be modularised? In the meantime I recommend unsetting the build flag.

shahzadmajeed:
Hi @Lukasa I just looked at my build command log and couldn't find that build flag. Where do you think this might be coming from?

Lukasa:
It is not likely to be the only flag that triggers this. How are you building?

shahzadmajeed:
I tried `swift build`, `xcodebuild archive` and direct Xcode run action.. all failed with same error.

Basically, i'm building a dynamic framework that links against static frameworks to create a single XCFramework. It builds fine when I use old versions but fails on latest version.

I will see if I can find other linker flags that are sent to xcodebuild invocation.

Lukasa:
Can you provide your build log?

shahzadmajeed:
Sure.. here is build log

[Build DomainKit_2023-08-17T10-27-28.txt](https://github.com/apple/swift-nio/files/12370744/Build.DomainKit_2023-08-17T10-27-28.txt)

and here is build settings for my DomainKit project (via. xcodebuild -showBuildSettings)

[DomainKitBuildSettings.txt](https://github.com/apple/swift-nio/files/12370772/DomainKitBuildSettings.txt)

Build Settings of NIO project

[SwiftNioBuildSettings.txt](https://github.com/apple/swift-nio/files/12370788/SwiftNioBuildSettings.txt)

Lukasa:
That's very strange, I can't find a reason for this compile error in those flags. Can you please set up that github example project?

allevato:
In @shahzadmajeed's case, I see that Clang is being invoked with the following flags:

```
-Wnon-modular-include-in-framework-module -Werror=non-modular-include-in-framework-module
```

which I believe is the default behavior for an Xcode project since the corresponding Xcode setting is named "*Allow* Non-modular Includes in Framework Modules" and setting it would disable those flags.

Lukasa:
Good catch: I was looking at the Swift compiler invocation, which is where the error is actually produced, not the C setting.

So again, the issue here is Tuist building this module as a framework.

Lukasa:
Our advice then would be to use the corresponding Xcode setting to allow the non-modular include.

allevato:
I've filed FB12999916 to cover the underlying problem in the SDK. (If any Apple folks can make sure the appropriate team sees it, that would be appreciated!)

shahzadmajeed:
I believe I tried building framework with non-modular headers before posting here but I can double check that.

shahzadmajeed:
@allevato when you say `Swift search paths` do you mean `Frameworks Search Paths`?

shahzadmajeed:
@Lukasa sorry for delay. I added you to my github private repo which is just a sample project to setup static frameworks. You should get an invite. I have added a section at the top on how to reproduce the error

PR #2483, opened 2023-07-27
Rename the `waitForFinalResult()` methods and provide an `ELF` convenience

FranzBusch:
# Motivation
The name `waitForFinalResult()` is bad in a couple of ways. First, we normally use `wait` for blocking operations. Secondly, `finalResult` is also just strange because the user only wants to know the _result_ and doesn't even know that there might be intermediates.

# Modification
This PR renames the `waitForFinalResult()` method to be a property on `NIOProtocolNegotiationResult`. Additionally, it provides a conditional extension on `ELF` to avoid a `get()`.

# Result
Better naming for our async interfaces

### DIFF
--- a/Sources/NIOCore/ChannelHandler.swift
+++ b/Sources/NIOCore/ChannelHandler.swift
@@ -354,25 +354,48 @@ public struct NIOProtocolNegotiationResult<NegotiationResult: Sendable> {
         case deferredResult(EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>)
     }
 
-    private let result: Result
+    private let _result: Result
 
-    public init(result: NegotiationResult) {
-        self.result = .finished(result)
-    }
-
-    public init(deferredResult: EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>) {
-        self.result = .deferredResult(deferredResult)
-    }
-
-    /// Waits for the final protocol negotiation result and  returns the value.
+    /// The final result of protocol negotiation.
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-    public func waitForFinalResult() async throws -> NegotiationResult {
-        switch self.result {
-        case .finished(let negotiationResult):
-            return negotiationResult
-        case .deferredResult(let eventLoopFuture):
-            return try await eventLoopFuture.flatMap { $0.result.resolve(on: eventLoopFuture.eventLoop) }.get()
+    var result: NegotiationResult {
+        get async throws {
+            switch self._result {
+            case .finished(let negotiationResult):
+                return negotiationResult
+            case .deferredResult(let eventLoopFuture):
+                return try await eventLoopFuture.flatMap { $0._result.resolve(on: eventLoopFuture.eventLoop) }.get()
+            }
         }
+    }
+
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/ChannelHandler.swift
```diff
@@ -354,25 +354,42 @@ public struct NIOProtocolNegotiationResult<NegotiationResult: Sendable> {
...
-    public func waitForFinalResult() async throws -> NegotiationResult {
-        switch self.result {
-        case .finished(let negotiationResult):
-            return negotiationResult
-        case .deferredResult(let eventLoopFuture):
-            return try await eventLoopFuture.flatMap { $0.result.resolve(on: eventLoopFuture.eventLoop) }.get()
-        }
+extension EventLoopFuture {
+    /// Get the result/error from the protocol negotiation.
+    @available(macOS 10.15, *)
```
glbrntt, line 389: This is missing some availability

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -354,25 +354,42 @@ public struct NIOProtocolNegotiationResult<NegotiationResult: Sendable> {
...
+    /// - Parameter deferredResult: The deferred result.
     public init(deferredResult: EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>) {
-        self.result = .deferredResult(deferredResult)
+        self._result = .deferredResult(deferredResult)
     }
+}
 
-    /// Waits for the final protocol negotiation result and  returns the value.
-    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-    public func waitForFinalResult() async throws -> NegotiationResult {
```
glbrntt, line 369: Do we use this anywhere? Might be easier to deprecate and remove later to avoid breaking our other repos etc.

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -354,25 +354,42 @@ public struct NIOProtocolNegotiationResult<NegotiationResult: Sendable> {
...
-        case .finished(let negotiationResult):
-            return negotiationResult
-        case .deferredResult(let eventLoopFuture):
-            return try await eventLoopFuture.flatMap { $0.result.resolve(on: eventLoopFuture.eventLoop) }.get()
-        }
+extension EventLoopFuture {
+    /// Get the result/error from the protocol negotiation.
+    @available(macOS 10.15, *)
+    @_spi(AsyncChannel)
+    public func getResult<NegotiationResult: Sendable>() async throws -> NegotiationResult where Value == NIOProtocolNegotiationResult<NegotiationResult> {
```
glbrntt, line 391: Why not also make this:

```swift
public var result: NegotiationResult {
    get async throws { ... }
}
```

FranzBusch: Because you cannot write a generic computed property AFAIK and you cannot write a parameterised extension like this `extension EventLoopFuture<NIOProtocolNegotiationResult<some Any>>`. I tried both avenues but we cannot express this at this time sadly.

rnro (approved):
This definitely looks more readable! I'm still not wild about `negotiationResult.getResult` but I think that's out of scope for this PR

glbrntt (approved):

PR #2486, opened 2023-07-28
Don't lose or delay writes before channelActive

Lukasa:
Motivation:

In rare circumstances, some users have reported hangs, delays, or lost writes when writing network proxies. These proxies have typically been derived from our TLSify example:
https://github.com/apple/swift-nio-examples/tree/main/TLSify

This proxy makes its outbound connections in handlerAdded. Generally this works without issue, as the connection takes long enough to set up that channelActive fires before the outbound connection has been made.

However, in some rare circumstances it is possible to both make an outbound connection, have it succeed, _and_ read from the network on that connection before NIO activates the inbound channel. This would cause users to lose writes.

This violates a cardinal NIO rule: writes should never be lost. If it is practical for us to delay them and apply them later, we should. This patch addresses several holes in this area to correctly handle these outbound writes.

The first hole is that we would fail writes if the Channel wasn't yet active. This constraint was unnecessary: it was added in a refactor to use an internal state machine, and while we had a test that this happened, it didn't protect us from anything. We can safely buffer a write at this point: and indeed, we _do_ buffer flushes at this point. So the check was unneeded.

The second hole was that, even if we buffered the writes, we frequently wouldn't unbuffer them again until a flush happened after channelActive. This didn't happen on _every_ path: in the rare case that connect() succeeded immediately, we would flush the writes. But in all other cases, including listening connections, we needed _another_ flush to happen to get the writes out.

This fix patches up both issues and reconciles the flows. Now we _always_ try to push out flushed writes when we activate the Channel, just as we always try to read.

Modifications:

- Remove a check that failed writes if we weren't active.
- Remove a test that validated that we did that.
- Attempt to emit flushed writes on channelActive.
- Make sure we mark ourselves writable if we couldn't write it all.
- Make `BaseSocket.getOption` non-final so we can hook it in the SAL.
- Make `AcceptHandler` non-private so we can use it in tests.
- Extend the SAL to support `getOption`
- Write several SAL tests to ensure we manage our writes.

Result:

No lost writes.

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -411,7 +411,7 @@ public final class ServerBootstrap {
         }
     }
 
-    class AcceptHandler: ChannelInboundHandler {
+    final class AcceptHandler: ChannelInboundHandler {
         public typealias InboundIn = SocketChannel
 
         private let childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -411,7 +411,7 @@ public final class ServerBootstrap {
         }
     }
 
-    private class AcceptHandler: ChannelInboundHandler {
+    class AcceptHandler: ChannelInboundHandler {
         public typealias InboundIn = SocketChannel
 
         private let childChannelInit: ((Channel) -> EventLoopFuture<Void>)?
--- a/Tests/NIOPosixTests/SALChannelTests.swift
+++ b/Tests/NIOPosixTests/SALChannelTests.swift
@@ -392,21 +392,21 @@ final class SALChannelTest: XCTestCase, SALTest {
         final class ConnectionRecorder: ChannelInboundHandler {
             typealias InboundIn = Any
             typealias InboundOut = Any
-
+            
             let readCount = ManagedAtomic(0)
-
+            
             func channelRead(context: ChannelHandlerContext, data: NIOAny) {
                 readCount.wrappingIncrement(ordering: .sequentiallyConsistent)
                 context.fireChannelRead(data)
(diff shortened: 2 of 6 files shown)


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -411,7 +411,7 @@ public final class ServerBootstrap {
         }
     }
 
-    private class AcceptHandler: ChannelInboundHandler {
+    class AcceptHandler: ChannelInboundHandler {
```
FranzBusch, line 414: Can we make this `final` while we are at it?

Tests/NIOPosixTests/SALChannelTests.swift
```diff
@@ -392,21 +392,21 @@ final class SALChannelTest: XCTestCase, SALTest {
         final class ConnectionRecorder: ChannelInboundHandler {
             typealias InboundIn = Any
             typealias InboundOut = Any
-
+
```
FranzBusch, line 395: We are adding a bunch of whitespaces here. Can you remove that

Tests/NIOPosixTests/SALChannelTests.swift
```diff
@@ -445,15 +445,442 @@ final class SALChannelTest: XCTestCase, SALTest {
...
+                        channel.write(secondWrite, promise: nil)
+                        channel.flush()
+                        return channel.eventLoop.makeSucceededVoidFuture()
+                    },
+                    childChannelOptions: childChannelOptions
+                ))
+        })
+    }
+
+    func testWriteBeforeChannelActiveServerStream_shortWriteLeadsToWritable_instantClose() throws {
```
weissi, line 821: thanks for all these super cool tests!

weissi (approved):
looks great to me, thanks!

FranzBusch (approved):

PR #2488, opened 2023-07-30
Add testing interfaces for the stream and writer of the `NIOAsyncChannel`

FranzBusch:
# Motivation
When using the `NIOAsyncChannel` to build higher level abstractions it is hard to test right now since one cannot drive the `NIOAsyncChannelInputStream` nor the `NIOAsyncChannelOutputWriter`.

# Modification
This PR adds new factory methods to both of those types which allow to either drive the input stream or record the outputs written to the output writer.

# Result
One can now successfully test their code written on top of the input stream and output writer.


glbrntt (approved):

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStream.swift
```diff
@@ -21,8 +21,57 @@ public struct NIOAsyncChannelInboundStream<Inbound: Sendable>: Sendable {
...
-    @usableFromInline let _producer: Producer
+    @usableFromInline
+    let _backing: _Backing
+
+    /// Creates a new stream with a source for testing.
+    ///
+    /// This is useful for writing unit tests where you want to drive a ``NIOAsyncChannelInboundStream``.
+    ///
+    /// - Returns: A tuple containing the input stream and a test source to drive it.
+    public static func makeTestingStream() -> (Self, TestSource) {
```
gjcairo, line 64: Should this be `@inlinable`?

gjcairo (approved):
LGTM, just one small comment

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStream.swift
FranzBusch: It doesn't really matter since this is only for testing purposes.

gjcairo: Yeah okay that's fair; the question came from the fact that `makeTestingWriter()` *is* `@inlinable` so it rang the inconsistency bell :P but maybe it's unnecessary on the writer then. I'm aware these are nits though :D

PR #2489, opened 2023-08-01
Fix leaked promise in `NIOTypedApplicationProtocolNegotiationHandler`

FranzBusch:
# Motivation
We sometimes leaked a promise in the `NIOTypedApplicationProtocolNegotiationHandler` when the handler was created and immediately deinited.

# Modification
This PR makes sure we always complete the negotiation promise of the handler.

# Result
No more leaked promises.

### DIFF
--- a/Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
+++ b/Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
@@ -80,6 +80,16 @@ public final class NIOTypedApplicationProtocolNegotiationHandler<NegotiationResu
         }
     }
 
+    deinit {
+        switch self.stateMachine.deinitHandler() {
+        case .failPromise:
+            self.negotiatedPromise.fail(ChannelError.inappropriateOperationForState)
+
+        case .none:
+            break
+        }
+    }
+
     @_spi(AsyncChannel)
     public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
         switch self.stateMachine.userInboundEventTriggered(event: event) {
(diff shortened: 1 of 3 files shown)


Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
```diff
@@ -80,6 +80,10 @@ public final class NIOTypedApplicationProtocolNegotiationHandler<NegotiationResu
         }
     }
 
+    deinit {
+        self.negotiatedPromise.fail(ChannelError.inappropriateOperationForState)
+    }
```
Lukasa, line 85: This solution seems wrong. We should rely on a channel lifecycle event to do this, probably `handlerRemoved`, and we should also know whether we already completed a promise or not.

Lukasa (changes requested):

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
FranzBusch: If we do the channel lifecycle we should also create the promise in handlerAdded but then we cannot really fulfill the protocol requirements and I didn’t want to make it optional in the protocol. That’s why I moved it to deinit. I also had it in handlerRemoved initially.

I agree on the second point.

Lukasa: I'm fine having this auto-fail on deinit _if_ we guarantee we only complete this once.

1proprogrammerchant: same here!

Lukasa (approved):

PR #2493, opened 2023-08-07
AsyncChannel: Provide throwing finish method on test stream

FranzBusch:
# Motivation
We recently provided testing utilities for the `NIOAsyncChannelInboundStream`. On thing that was missing is a way to finish the stream with an error.

# Modification
This PR provides a `finish()` method that takes an error which is thrown from the inbound stream.

# Result
Better way to test code relying on the AsyncChannel work.


Sources/NIOCore/AsyncChannel/AsyncChannelInboundStream.swift
```diff
@@ -40,23 +40,25 @@ public struct NIOAsyncChannelInboundStream<Inbound: Sendable>: Sendable {
         }
 
         /// Finished the inbound stream.
+        ///
+        /// - Parameter error: The error to throw, or nil, to finish normally.
         @inlinable
-        public func finish() {
-            self.continuation.finish()
+        public func finish(throwing error: Error?) {
```
glbrntt, line 46: I think we should default this to `nil`. Having to spell out `finish(throwing: nil)` for a 'happy-path' finish is a bit awkward IMO.

Lukasa (approved):

PR #2496, opened 2023-08-07
EventLoopFuture/Promise: only Sendable if Value is Sendable

weissi:
### Motivation:

`EventLoopFuture/Promise` are (incorrectly) marked as unconditionally `Sendable`. But this allows to send non-`Sendable` values which isn't right.

### Modifications:

Mark `EventLoopFuture/Promise` as conditionally `Sendable` if `Value` is `Sendable`.

This is technically API breaking but given the nature of the `Sendable` protocol should only result in warnings in the very most cases. The NIO release tools will assess compatibility.

### Result:

- More correctness
- More warnings for users (sorry NIO users)

### DIFF
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -2293,13 +2293,13 @@ public struct _NIOEventLoopFutureIdentifier: Hashable, Sendable {
     }
 }
 
-// EventLoopPromise is a reference type, but by its very nature is Sendable.
-extension EventLoopPromise: Sendable { }
+// EventLoopPromise is a reference type, but by its very nature is Sendable (if its Value is).
+extension EventLoopPromise: Sendable where Value: Sendable { }
 
-// EventLoopFuture is a reference type, but it is Sendable. However, we enforce
+// EventLoopFuture is a reference type, but it is Sendable (if its Value is). However, we enforce
 // that by way of the guarantees of the EventLoop protocol, so the compiler cannot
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
I'm remarkably nervous about merging this, but it must not merge until immediately after a release. We need some time for the release validation to bake. Marking "request changes" to block an accidental merge.

weissi:
> I'm remarkably nervous about merging this, but it must not merge until immediately after a release. We need some time for the release validation to bake. Marking "request changes" to block an accidental merge.

Good idea. Yes, I'd suggest to make a NIO minor release _JUST_ for this PR :). If something goes side-ways we can "repair" it by undoing it in a follow-up patch version (think in SemVer this is allowed if something was done accidentally).

ktoso:
Overall it’s a good idea, but I support the careful rollout 👍

weissi:
> Also very nervous about this, especially about Swift 5.6/5.7. But it is the right thing to do.

I mean we _could_ (not saying we necessarily) should `#if compiler(>5.9)` this whole thing and then nothing changes until 5.9 is released :P.

Lukasa:
Sadly that's still a source break :wink:

weissi:
> Sadly that's still a source break 😉

No disagreement here

Lukasa (approved):
OK, I think we can land this and see what happens.

Lukasa:
FYI @simonjbeaumont who is monitoring the release tools at this time.

Issue #2494, opened 2023-08-07
NIOTypedApplicationProtocolNegotiationHandler: control-flow observable `deinit` abuse (instead of correct lifecycle)


weissi:
https://github.com/apple/swift-nio/blob/1ec71beb8a27a1ddcc42aeeadce6273378f59581/Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift#L86

This fails a promise from deinit which is very bad because this means a `deinit` becomes observable via control flow. This sort of bug is infectious and will lead to hard-to-debug/workaround bugs that we can just avoid following what NIO does everywhere: Explicit and correct lifecycle management.

weissi:
originates from #2489

PR #2497, opened 2023-08-07
Avoid using deinit to fulfil the protocol negotiation promise

FranzBusch:
# Motivation

Fixes https://github.com/apple/swift-nio/issues/2494

# Modification
This PR avoids using `deinit` to fulfil the protocol negotiation promise and opts to trap instead when it is being accessed before the handler is added. This allows us to use `handlerAdded` and `handlerRemoved`.

# Result
No more `deinit` usage that can be observed.

### DIFF
--- a/Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
+++ b/Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
@@ -49,10 +49,14 @@ public final class NIOTypedApplicationProtocolNegotiationHandler<NegotiationResu
 
     @_spi(AsyncChannel)
     public var protocolNegotiationResult: EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>> {
-        self.negotiatedPromise.futureResult
+        return self.negotiatedPromise.futureResult
     }
 
-    private let negotiatedPromise: EventLoopPromise<NIOProtocolNegotiationResult<NegotiationResult>>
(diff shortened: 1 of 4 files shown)


Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
```diff
@@ -74,14 +77,18 @@ public final class NIOTypedApplicationProtocolNegotiationHandler<NegotiationResu
...
-    public convenience init(eventLoop: EventLoop, alpnCompleteHandler: @escaping (ALPNResult) -> EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>) {
-        self.init(eventLoop: eventLoop) { result, _ in
+    public convenience init(alpnCompleteHandler: @escaping (ALPNResult) -> EventLoopFuture<NIOProtocolNegotiationResult<NegotiationResult>>) {
+        self.init { result, _ in
             alpnCompleteHandler(result)
         }
     }
 
-    deinit {
-        switch self.stateMachine.deinitHandler() {
```
Lukasa, line 84: Should we drop the `deinitHandler` method?

Lukasa (changes requested):

Sources/NIOTLS/NIOTypedApplicationProtocolNegotiationHandler.swift
FranzBusch: Yes and the method is dropped from the state machine in this PR.

PR #2501, opened 2023-08-11
Enforce Sendability in `EventLoopFuture` and `EventLoopPromise` methods

FranzBusch:
# Motivation
With https://github.com/apple/swift-nio/pull/2496 we fixed a Sendability checking hole by removing the unconditional conformance of `EventLoopFuture/Promise` to `Sendable`. This was the correct thing to do; however, it has the fallout that a couple of methods are now rightly complaining that their values are send across isolation domains.

# Modification
This PR requires values on some `ELF/P` methods to be `Sendable` when we might potentially transfer the values across isolation domains/ELs. We have to be overly aggressive here because we do not know that some `ELF` method are staying on the same EL. For example `flatMap` gets a new `ELF` from the closure provided to it. If the `ELF` is on the same EL we do not need to hop; however, we can not guarantee this right now from a type level so we have to stay on the safe side and actually require the `NewValue` to be `Sendable`.

# Result
This PR makes us more correct from a Sendability perspective but produces warnings for some safe patterns that are currently in use.


Sources/NIOCore/EventLoopFuture+WithEventLoop.swift
```diff
@@ -41,25 +41,12 @@ extension EventLoopFuture {
...
-                    futureU.cascade(to: next)
-                    return CallbackList()
-                }
-            case .failure(let error):
-                return next._setValue(value: .failure(error))
-            }
-        }
-        return next.futureResult
+    public func flatMapWithEventLoop<NewValue: Sendable>(_ callback: @escaping @Sendable (Value, EventLoop) -> EventLoopFuture<NewValue>) -> EventLoopFuture<NewValue> {
+        // Is this the same thing and still fast?
```
FranzBusch, line 45: I am not sure about this one. Probably the previous implementation was faster but I wasn't really sure here

Sources/NIOCore/EventLoopFuture+WithEventLoop.swift
```diff
@@ -41,25 +41,12 @@ extension EventLoopFuture {
...
+        let eventLoop = self.eventLoop
+        return self.flatMap {
+            callback($0, eventLoop)
+        }.hop(to: self.eventLoop)
```
weissi, line 49: you can delete the `.hop(to:)`. `self.flatMap` will return a future on `self.eventLoop`

weissi: and the previous implementation probably also allocates less, depends on the optimiser a bit. Why not just revert?

FranzBusch: Right. Well this method became a lot easier then :D

Lukasa:
Fixed in 2.78.

PR #2508, opened 2023-08-21
Support users opting-out of us setting framing headers.

Lukasa:
Motivation:

Right now we always set the framing headers in HTTP requests and responses that we send. This is a good practice for most users, but it can cause issues, most notably in responses to CONNECT requests which requires that we do not set framing headers.

Unfortunately, in NIO's current HTTP/1.1 design it is not possible for us to suppress these headers. This is because the HTTP encoders come _earlier_ in the pipeline than the decoders, so the HTTP encoders do not know structurally what requests they are responding to.

While we could merge the encoders/decoders, that's a fairly substantial change. A better short-term change is to offer users the ability to turn off the NIO header manipulation feature. In this circumstance, users take on full responsibility for appropriately framing their HTTP messages.

Modifications

- Add config to HTTPRequestEncoder and HTTPResponseEncoder.
- Plumb this config through.
- Add a bunch of tests.

Result

Users have way more control of HTTP/1.1 messages.

### DIFF
--- a/Sources/NIOHTTP1/HTTPEncoder.swift
+++ b/Sources/NIOHTTP1/HTTPEncoder.swift
@@ -128,6 +128,16 @@ private func correctlyFrameTransportHeaders(hasBody: HTTPMethod.HasBody, headers
     }
 }
 
+/// Validates the response/request headers to ensure that we correctly send the body with chunked transfer
+/// encoding, when needed.
+private func messageIsChunked(headers: HTTPHeaders, version: HTTPVersion) -> Bool {
+    if version.major == 1 && version.minor >= 1 {
+        return headers["transfer-encoding"].first == "chunked" ? true : false
+    } else {
+        return false
+    }
+}
+
 /// A `ChannelOutboundHandler` that can serialize HTTP requests.
 ///
 /// This channel handler is used to translate messages from a series of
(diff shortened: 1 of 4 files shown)


Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -128,6 +128,16 @@ private func correctlyFrameTransportHeaders(hasBody: HTTPMethod.HasBody, headers
     }
 }
 
+/// Validates the response/request headers to ensure that we correctly send the body with chunked transfer
+/// encoding, when needed.
+private func messageIsChunked(headers: HTTPHeaders, version: HTTPVersion) -> Bool {
+    if version.major == 1 && version.minor >= 1 {
+        return headers["transfer-encoding"].first == "chunked" ? true : false
```
glbrntt, line 135: nit: `headers.first(name: "transfer-encoding")`

Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -177,18 +219,50 @@ public final class HTTPResponseEncoder: ChannelOutboundHandler, RemovableChannel
...
     public func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
         switch self.unwrapOutboundIn(data) {
         case .head(var response):
             assert(!(response.headers.contains(name: "content-length") &&
                         response.headers[canonicalForm: "transfer-encoding"].contains("chunked"[...])),
                      "illegal HTTP sent: \(response) contains both a content-length and transfer-encoding:chunked")
-            self.isChunked = correctlyFrameTransportHeaders(hasBody: response.status.mayHaveResponseBody ? .yes : .no,
-                                                            headers: &response.headers, version: response.version) == .chunked
+
+            if configuration.automaticallySetFramingHeaders {
```
glbrntt, line 260:
```suggestion
            if self.configuration.automaticallySetFramingHeaders {
```

glbrntt (approved):
Looks good! Left a couple of nits inline.

dnadoba (approved):

PR #2509, opened 2023-08-23
Make `ByteBuffer` Codable

### LINKED ISSUE #2507: ByteBuffer should be Codable
`ByteBuffer` should implement `Codable` in way that makes

```
struct FooBB: Codable {
   var bytes: ByteBuffer
}
```

encode/decode into the same `.json` as

```
struct FooData: Codable {
  var bytes: Data
}
```

does for

`JSONEncoder().encode(FooData(bytes: Data(buffer: fooBB.bytes)))`

Clarifying note: I very much do _not_ want to be able to encode as `FooBB` and decode as `FooData` (or vice versa) but I want a refactoring of `FooData` to `FooBB` to result in the same `.json` files.

natikgadzhi:
This pull request implements `Codable` for `ByteBuffer`.

### Motivation:

- Closes #2507 
- Implements a base64 encode/decode for ByteBuffer.

### Modifications:

- Moves `Base64` code into `_NIODataStructures`. I'm not sure we should put it there — I can move it into a new underscore module if you think we need to.
- Adds a couple of new tests for `Base64`
- Adds `Codable` to `ByteBuffer`.
- Formatting changes in `Base64`.

### TODO
- [ ] Would be good to add tests for `ByteBuffer` encoding and decoding. It's a bit awkward to manipulate encoders and decoders in tests. 👀 

### Review
- Does that look like a good direction overall? 
- Can we remove String backports from Base64? Since we're using Swift 5.6 for now, it doesn't yet allow us to use `String(withUnsafeUnitializedCapacity` so I guess we have to wait for a bit longer?


Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -675,6 +676,23 @@ extension ByteBuffer {
     }
 }
 
+extension ByteBuffer: Codable {
+
+    /// Creates a ByteByffer by decoding from a Base64 encoded single value container.
+    public init(from decoder: Decoder) throws {
+        let container = try! decoder.singleValueContainer()
+        let base64String = try! container.decode(String.self)
+        self = try! ByteBuffer(bytes: base64String.base64Decoded())
```
weissi, line 685: we can't `try!` here, the string might not be valid base64. And we should add a test that proves that.

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -675,6 +676,23 @@ extension ByteBuffer {
...
+    }
+
+    /// Encodes this buffer as a base64 string in a single value container.
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        let base64String = String(base64Encoding: self.readableBytesView)
+        try! container.encode(base64String)
```
weissi, line 692: this `try!` should also be a `try`. I don't think JSONEncoder will ever throw on a string but there might be other encoders that limit the size of the output of a single string value (I think).

Tests/NIODataStructuresTests/Base64Test.swift
```diff
@@ -28,6 +28,26 @@ class Base64Test: XCTestCase {
...
+        let encoded = String(base64Encoding: string.utf8)
+        let expected = "SGVsbG8sIHdvcmxkIQ=="
+        XCTAssertEqual(encoded, expected)
+    }
+
+    func testBase64DecodeHelloWorld() throws {
+        let encoded = "SGVsbG8sIHdvcmxkIQ=="
+        let decoded = try! encoded.base64Decoded()
+        XCTAssertEqual(decoded, "Hello, world!".utf8.map{ UInt8($0) })
+    }
```
weissi, line 49: need some more slightly less happy case tests but otherwise good

weissi (commented):
Thank you! That's great and I think we're almost there, just a few `try!`s that need to be `try` because we can't control what gets thrown at us

Sources/NIOCore/ByteBuffer-aux.swift
natikgadzhi: Done: https://github.com/apple/swift-nio/pull/2509/commits/3861cf8b76686fa907d44fb96bf6565483bc5185

Sources/NIOCore/ByteBuffer-aux.swift
natikgadzhi: Also done: https://github.com/apple/swift-nio/pull/2509/commits/3861cf8b76686fa907d44fb96bf6565483bc5185 

Yep, that was prototype-grade spike at first.

Tests/NIODataStructuresTests/Base64Test.swift
natikgadzhi: Added a few more: https://github.com/apple/swift-nio/pull/2509/commits/b5c867908cf99da9b083e339e26822f7afdcfb7b

natikgadzhi:
@Lukasa, ready for review!

Two things look suspicious to me: 
- I've added `Base64` to `_NIODataStructures`. It's small, so that seems fine, but perhaps we should rename it then? It seems like adding a separate module for just Base64 is an overkill?
- I've added more tests to ByteBufferTests, and that file starts looking very scary with 3.5+k LOC. Should we plan to break it up into smaller chunks, and work in improving setup and teardown for them so they are more readable?

Lukasa:
> It seems like adding a separate module for just Base64 is an overkill?

Nah, modules are cheap. I think it's a good idea to add a separate one.

> Should we plan to break it up into smaller chunks, and work in improving setup and teardown for them so they are more readable?

Conceptually that's a reasonable enough idea. We shouldn't do it in this PR though.

Lukasa:

natikgadzhi:
@Lukasa, how should we name the new module? I'd love to set it up in a way that we can use this module in the future for similar things, so I'm not sure it should be `_NIOBase64` for example. Nothing good comes to mind, though — I don't have enough experience with SwiftNIO to plan for that. 

`_NIOInternalAlgirithms` seems restrictive to algorithms?

Lukasa:
I don't think we need to stress too much about it. `_NIOBase64` is fine for now.

natikgadzhi:
@Lukasa then I think that's in good shape and ready for a review! 

I think I saw two CI failures on Swift 5.8 and nightly, but they seemed unrelated. Should I look into them?

natikgadzhi:
Send halp. It looks like CI is dying on `let ptr = allocator.malloc(size_t(bytes))!`, and I'm not entirely sure why. Would that be because with the new tests I'm allocating too much, and not cleaning up after myself in test cases somehow?

Sources/NIOWebSocket/NIOWebSocketClientUpgrader.swift
```diff
@@ -14,6 +14,7 @@
 
 import NIOCore
 import NIOHTTP1
+import _NIODataStructures
```
Lukasa, line 17: This needs to be `_NIOBase64`.

Package.swift
```diff
@@ -46,6 +46,7 @@ let package = Package(
             name: "NIOCore",
             dependencies: [
                 "NIOConcurrencyHelpers",
+                "_NIOBase64",
```
Lukasa, line 49: Need to add this dependency to `NIOWebSocket` as well.

Lukasa:
Let's begin by cleaning up those compiler errors first.

natikgadzhi:
🤦🏼 How the hell did I miss that =/ I could swear it built, but I was not paying enough attention, and didn't do a clean build. 

- Fixed `Package.swift` and imports based on your feedback
- Clean -> Build -> Tests work locally in Xcode
- `swift test` also works, with `WARNING: Syscall wrapper test: Over 100% overhead allowed. Running in debug assert configuration which allows 2000% overhead :(. Consider running in Release mode.`

Lukasa (approved):

weissi:
Amazing, thank you!

PR #2513, opened 2023-09-05
Simplify header separator and crlf

alex-taffe:
Everywhere else in the file uses `crlf` and `headerSeperator` when writing those static strings, we shouldn't be using string literals here

### Motivation:

Code cleanliness

### Modifications:

Swapped the string literals of `\r\n` and `: ` for their already defined constants

### Result:

No functional changes


Lukasa (approved):

Lukasa:

PR #2515, opened 2023-09-09
Add NIOLoopBound(Box) uncheckedValue

### LINKED ISSUE #2506: NIOLoopBound should get `unchecked` variants that are truly free in release mode
`NIOLoopBound` is super useful when adopting `Sendable` but the `eventLoop.preconditionInEventLoop()` checks everywhere are expensive and -- for repeated accesses -- unnecessary.

There should be `unchecked` variants of init, get and set value that do `eventLoop.assertInEventLoop()` instead. Together with `@inlinable` that should become completely free in release mode (and still check in debug).

natikgadzhi:
## Motivation
- This commit adds unchecked version of init and value get+set of `NIOLoopBound` and `NIOLoopBoundBox`.
- Closes #2506.

## Modifications
- Added `init` and `public var uncheckedValue` properties to both `NIOLoopBound` and `NIOLoopBoundBox`.

### Review
- [x] Tests pass locally.
- [ ] How should we test that `NIOLoopBound.init(uncheckedEventLoop:)` calls `assertInEventLoop`? `EmbeddedEventLoop` seems to be final — how do I mock the assert function on it elegantly, without copy and pasting all the protocol conformance stuff?

@weissi, sorry, I found a few really nice rabbit holes in swift-syntax and got distracted :eyes:

### DIFF
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -30,10 +30,18 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
     @usableFromInline
     /* private */ var _value: Value
 
-    /// Initialise a ``NIOLoopBound`` to `value` with the precondition that the code is running on `eventLoop`.
+    /// Initialize a ``NIOLoopBound`` to `value` with the precondition that the code is running on `eventLoop`.
     @inlinable
     public init(_ value: Value, eventLoop: EventLoop) {
         eventLoop.preconditionInEventLoop()
+        self.init(value, uncheckedEventLoop: eventLoop)
+    }
+
+    @inlinable
+    /// Initialize a ``NIOLoopBound`` to `value` with _an assertion_ that the code is running on `uncheckedEventLoop`.
+    /// Unlike a precondition check,  ``EventLoop/assertInEventLoop(file:line:)`` only performs the check in debug configuration, so the check is free in release configuration.
+    public init(_ value: Value, uncheckedEventLoop eventLoop: EventLoop) {
+        eventLoop.assertInEventLoop()
         self._eventLoop = eventLoop
         self._value = value
     }
@@ -49,6 +57,23 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
         }
         set {
             self._eventLoop.preconditionInEventLoop()
+            self._value = newValue
+        }
+    }
+
+    /// Access the `value` with the assertion that the code is running on `eventLoop`.
+    ///
+    /// Unlike ``NIOLoopBound/value``, this performs the assertion in debug configuration only, so it's
+    /// cheaper, and still performs the precondition check in debug mode.
+    /// - note: ``NIOLoopBound`` itself is value-typed, so any writes will only affect the current value.
+    @inlinable
+    public var uncheckedValue: Value {
+        get {
+            self._eventLoop.assertInEventLoop()
+            return self._value
+        }
+        set {
+            self._eventLoop.assertInEventLoop()
             self._value = newValue
         }
     }
@@ -77,19 +102,29 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
     @usableFromInline
     /* private */var _value: Value
 
+    /// Initialize a ``NIOLoopBoundBox`` to `value` with the an assertion that the code is running on `eventLoop`.
+    ///
+    /// - note: Unlike ``NIOLoopBoundBox/init(_:eventLoop:)``, this performs ``EventLoop/assertInEventLoop(file:line:)`` instead of a precondition check, which is free in release mode.
     @inlinable
-    internal init(_value value: Value, uncheckedEventLoop eventLoop: EventLoop) {
+    public init(_ value: Value, uncheckedEventLoop eventLoop: EventLoop) {
+        eventLoop.assertInEventLoop()
         self._eventLoop = eventLoop
         self._value = value
     }
 
-    /// Initialise a ``NIOLoopBoundBox`` to `value` with the precondition that the code is running on `eventLoop`.
+    @inlinable
+    internal init(_ value: Value, notVerifyingEventLoop eventLoop: EventLoop) {
+        self._eventLoop = eventLoop
+        self._value = value
+    }
+
+    /// Initialize a ``NIOLoopBoundBox`` to `value` with the precondition that the code is running on `eventLoop`.
     @inlinable
     public convenience init(_ value: Value, eventLoop: EventLoop) {
         // This precondition is absolutely required. If not, it were possible to take a non-Sendable `Value` from
         // _off_ the ``EventLoop`` and transport it _to_ the ``EventLoop``. That would be illegal.
         eventLoop.preconditionInEventLoop()
-        self.init(_value: value, uncheckedEventLoop: eventLoop)
+        self.init(value, uncheckedEventLoop: eventLoop)
     }
 
     /// Initialise a ``NIOLoopBoundBox`` that is empty (contains `nil`), this does _not_ require you to be running on `eventLoop`.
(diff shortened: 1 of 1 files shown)


Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -30,10 +30,18 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
     @usableFromInline
     /* private */ var _value: Value
 
-    /// Initialise a ``NIOLoopBound`` to `value` with the precondition that the code is running on `eventLoop`.
+    /// Initialize a ``NIOLoopBound`` to `value` with the precondition that the code is running on `eventLoop`.
```
natikgadzhi, line 33: Tomato tomato, but thought there are more `initializes` in our codebase than `initialises`.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -30,10 +30,18 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
...
     @inlinable
     public init(_ value: Value, eventLoop: EventLoop) {
         eventLoop.preconditionInEventLoop()
+        self.init(value, uncheckedEventLoop: eventLoop)
+    }
+
+    @inlinable
+    /// Initialize a ``NIOLoopBound`` to `value` with _an assertion_ that the code is running on `uncheckedEventLoop`.
+    /// Unlike a precondition check,  ``EventLoop/assertInEventLoop(file:line:)`` only performs the check in debug configuration, so the check is free in release configuration.
+    public init(_ value: Value, uncheckedEventLoop eventLoop: EventLoop) {
```
natikgadzhi, line 43: A few hours after I pushed this up, I realized that having an unchecked `init` does not necessarily win the user that much time — you init once. Perhaps if you have to init a lot of objects that makes sense, but `uncheckedValue` seems more useful, in comparison.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -121,5 +156,22 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
...
+            self._eventLoop.assertInEventLoop()
+            return self._value
+        }
+        set {
+            self._eventLoop.assertInEventLoop()
+            self._value = newValue
+        }
+    }
+
+
```
natikgadzhi, line 175: Nit: Run `swiftformat`, come on.

Sources/NIOCore/NIOLoopBound.swift
FranzBusch, line 43: I think unchecked isn't quite the right name for this init but I can't come up with a better one right now. It is not completely unchecked but rather just asserts than preconditions.

FranzBusch: `uncheckedValue` is basically `UnsafeTransfer` which just puts the trust into the users hand that they now where things are going and that it is safe to do so.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -52,6 +60,23 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
...
+    public var uncheckedValue: Value {
+        get {
+            self._eventLoop.assertInEventLoop()
+            return self._value
+        }
+        set {
+            self._eventLoop.assertInEventLoop()
+            self._value = newValue
+        }
+    }
```
FranzBusch, line 79: What if we just provide a second type that does asserts or preconditions. Alternatively what about just making the whole precondition/assert behaviour configurable on init?

Sources/NIOCore/NIOLoopBound.swift
Lukasa: Unchecked is fine for that: [it's what `Range` uses, for example](https://github.com/apple/swift/blob/b5ddffdb3d095e4a57abaac3f8c1e327d64ebea1/stdlib/public/core/Range.swift#L177).

Lukasa: I think this is worth keeping: while you only `init` once, you typically shouldn't _store_ one of these: instead, you should use it to effect a transfer.

Sources/NIOCore/NIOLoopBound.swift
Lukasa: I think this pattern is better: I don't see a good reason to make this a type or a runtime distinction. It's good to be able to see this in the code, as it's fundamentally a safety-breaking optimization.

dnadoba: TBH I'm also thinking a type would be better. It's how the rest of Swifts unsafe/unchecked APIs usually work e.g. `UncheckedContinuation`, `Unsafe*Pointer`, `@unchecked Sendable`.

It may also enable us to drop the stored `EventLoop` property in release modes, once `DEBUG` is properly propagated in SwiftPM (haven't checked recently), making this actually a zero cost abstraction.

weissi: One type is better here: I often want mixed use: First time in a function: checked, follow-ups: unchecked. Different type would be pain and would likely push me to use unchecked everywhere.

dnadoba: > `First time in a function: checked, follow-ups: unchecked`

It would be better to get the value checked out of the loop bound and just store it in a local `let` for future accesses.

Lukasa: > It's how the rest of Swifts unsafe/unchecked APIs usually work

This is not true: the number of _safe_ types with _unsafe_ functions vastly dwarfs the number of unsafe types.

dnadoba: > the number of safe types with unsafe functions vastly dwarfs the number of unsafe types

All the `withUnsafe*` functions I can think off vend you types that have `Unsafe` or `Unchecked` in the name. About which functions do you think?
I'm only aware of `[Closed]Range(uncheckedBounds:)` which result in a type that doesn't have `Unsafe` in the name.

dnadoba: Concretely, I'm proposing this API:
```swift
public struct NIOUncheckedLoopBound<Value> {
    public init(_ value: Value, eventLoop: EventLoop) {}
    public init(_ value: NIOLoopBound<Value>) {}
    public init(_ value: NIOLoopBoundBox<Value>) {}
    
    public var value: Value {
        get {}
    }
}

public final class NIOUncheckedLoopBoundBox<Value> {
    public init(_ value: Value, eventLoop: EventLoop) {}
    public init(_ value: NIOLoopBound<Value>) {}
    public init(_ value: NIOLoopBoundBox<Value>) {}
    
    public var value: Value {
        get {}
        set {}
    }
}
```

If `NIOLoopBound.value` is to expensive you can still wrap it in `NIOUncheckedLoopBound(myLoopBoundValue).value`. Its more verbose but that's a good thing.

natikgadzhi: Whoa, thank you, @dnadoba, @Lukasa, and @weissi for the discussion! I'm learning quite a lot from you here.

I'm not attached to the current implementation — as I learn more and if we decide to change it, I'm happy to rewrite it.

> You can't say in isolation if loopBound.uncheckedValue is safe or not. You need to go and look at the creation of the NIOLoopBound and trace back on which EventLoop it was created. 

I agree, and I agree it's different from the `Range` example.

> We are proposing a set of duplicated APIs which make all operations on NIOLoopBound unsafe/unchecked.
> This is more in line with the UnsafePointer APIs where almost all operation are unsafe.

> This design also gives the wrong impression that NIOLoopBound.value is always safe and only uncheckedValue is unsafe. With this PR value becomes unsafe as well as you might have initialised it with NIOLoopBound(_:uncheckedEventLoop).

If we choose to stick with the current implementation, I think I should at the very least comment and document on that behavior.

> Such a type is borderline useless, because to support people using it you have to litter it all over your codebase.

@Lukasa, with the `Range` example, I'd fully agree. For `NIOLoopBound`, though — we're not passing `LoopBound` around in the NIO codebase right now, right? So in theory, sure, we could do two separate structures. The weight of supporting two different structures would then be on the user. Perhaps we could provide a `NIOLoopBound` as a protocol, and make a safe and unchecked implementations of that protocol, so that users could accept `some NIOLoopBound` and deal with them that way?


Again, I feel that I'm out of my depth in the API design discussion, but I'm delighted to be part of it and listen in. Feel free to "well actually" me on any points I'm bringing up.

PR #2517, opened 2023-09-13
Introduce new typed `HTTPServerUpgrader` and `WebSocketServerUpgrader`

FranzBusch:
# Motivation
With our new `NIOAsyncChannel` and typed bootstrap APIs we want to be able to let users spell out their pipeline in a typed way. Pipelines can contain handlers that have to make a forking decision such as HTTP upgrading. Our current `HTTPServerUpgradeHandler` is one of those handlers but it lacks strict typing. To interact nicely with our new typed APIs we need to have a new variant of the `HTTPServerUpgradeHandler` that can carry type information.

# Modification
This PR adds a few things:
1. A new `NIOTypedHTTPServerUpgradeHandler` + `NIOTypedHTTPServeProtocolUpgrader`. I also moved the state handling logic to a separate state machine. I thought about unifying the state machines of the _old_ handler and the new one but they differ in behaviour which makes the state machine more complicated.
2. A new `NIOTypedWebSocketServerUpgrader` that conforms to `NIOTypedHTTPServerProtocolUpgrader`
3. An overhauled WebSocket server example that fully uses Concurrency.

# Result
We now have a way to fully type the server side of HTTP protocol upgrading.


FranzBusch:

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -0,0 +1,360 @@
...
+        case fireChannelReadCompleteAndRemoveHandler
+    }
+
+    @inlinable
+    mutating func unbuffer() -> UnbufferAction {
+        switch self.state {
+        case .initial, .awaitingUpgrader, .upgraderReady, .upgrading, .finished:
+            preconditionFailure("Invalid state \(self.state)")
+
+        case .unbuffering(var buffer):
```
glbrntt, line 343: We should enter the `modifying` state here, no?

glbrntt (commented):
I'm only part way through reviewing this but dropping my comments so far. 

Could you punt the example client and server into a follow-up PR to make this one a bit smaller?

Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
```diff
@@ -224,62 +224,177 @@ public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader, @unch
...
+    guard version == "13" else {
+        return channel.eventLoop.makeFailedFuture(NIOWebSocketUpgradeError.invalidUpgradeHeader)
+    }
+
+    return shouldUpgrade(channel, upgradeRequest).flatMapThrowing { extraHeaders in
+        guard let extraHeaders = extraHeaders else {
+            throw NIOWebSocketUpgradeError.unsupportedWebSocketTarget
         }
+        return extraHeaders
+    }.map { (extraHeaders: HTTPHeaders) in
```
glbrntt, line 357: This map seems redundant; can't we just do all of this work in the `flatMapThrowing`?

Sources/NIOWebSocket/NIOWebSocketServerUpgrader.swift
```diff
@@ -224,62 +224,177 @@ public final class NIOWebSocketServerUpgrader: HTTPServerProtocolUpgrader, @unch
...
+    upgradeRequest: HTTPRequestHead,
+    maxFrameSize: Int,
+    automaticErrorHandling: Bool,
+    upgradePipelineHandler: @escaping @Sendable (Channel, HTTPRequestHead) -> EventLoopFuture<UpgradeResult>
+) -> EventLoopFuture<UpgradeResult> {
+    /// We never use the automatic error handling feature of the WebSocketFrameDecoder: we always use the separate channel
+    /// handler.
+    var upgradeFuture = channel.pipeline.addHandler(WebSocketFrameEncoder()).flatMap {
+        channel.pipeline.addHandler(ByteToMessageHandler(WebSocketFrameDecoder(maxFrameSize: maxFrameSize)))
+    }
```
glbrntt, line 389: can we `addHandlers` here instead?

glbrntt: or even have a sync version for adding all these handlers and then do the EL dance

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -0,0 +1,360 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5: This seems unlikely to be right.

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -0,0 +1,360 @@
...
+        /// We are either running the upgrading handler.
+        case upgrading(
+            buffer: Deque<NIOAny>
+        )
+
+        case unbuffering(buffer: Deque<NIOAny>)
+
+        case finished
+
+        case modifying
```
Lukasa, line 48: We shouldn't need this state: have you validated that we do?

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    private var state = State.initial
+
+    @usableFromInline
+    enum HandlerRemovedAction {
+        case failUpgradePromise
+    }
+
+    @inlinable
```
Lukasa, line 58: We're very inconsistent with the `@inlinable`s here, despite all the actions being `@usableFromInline`. Why is that?

Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
```diff
@@ -0,0 +1,370 @@
...
+    /// All the header fields the protocol needs in the request to successfully upgrade. These header fields
+    /// will be provided to the handler when it is asked to handle the upgrade. They will also be validated
+    /// against the inbound request's `Connection` header field.
+    var requiredUpgradeHeaders: [String] { get }
+
+    /// Builds the upgrade response headers. Should return any headers that need to be supplied to the client
+    /// in the 101 Switching Protocols response. If upgrade cannot proceed for any reason, this function should
+    /// return a failed future.
+    func buildUpgradeResponse(
+        channel: Channel,
```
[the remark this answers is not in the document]
FranzBusch: It is mostly used to grab the channel's EL and create a future.

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
FranzBusch: So I was believing that this isn't necessary anymore but from @glbrntt and my recent benchmarking we saw that the compiler can still miss correctly avoiding those CoWs even if everything is inlined. So I am being defensive again until we get `switch inout`.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -374,6 +374,95 @@ extension ChannelPipeline {
...
+    /// - returns: An `EventLoopFuture` that will fire when the pipeline is configured.
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    @_spi(AsyncChannel)
+    public func configureTypedHTTPServerPipeline<UpgradeResult: Sendable>(
+        position: ChannelPipeline.Position = .last,
+        withPipeliningAssistance pipelining: Bool = true,
+        withServerUpgrade upgrade: NIOTypedHTTPServerUpgradeConfiguration<UpgradeResult>,
+        withErrorHandling errorHandling: Bool = true,
+        withOutboundHeaderValidation headerValidation: Bool = true,
+        withEncoderConfiguration encoderConfiguration: HTTPResponseEncoder.Configuration = .init()
```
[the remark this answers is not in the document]
glbrntt: I think we should. While there is value in consistency I think diverging from it to make APIs more idiomatic is justified.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -389,69 +389,63 @@ extension ChannelPipeline {
...
     @_spi(AsyncChannel)
     public func configureTypedHTTPServerPipeline<UpgradeResult: Sendable>(
-        withPipeliningAssistance pipelining: Bool = true,
-        withServerUpgrade upgrade: NIOTypedHTTPServerUpgradeConfiguration<UpgradeResult>,
-        withErrorHandling errorHandling: Bool = true,
-        withOutboundHeaderValidation headerValidation: Bool = true,
-        withEncoderConfiguration encoderConfiguration: HTTPResponseEncoder.Configuration = .init()
+        isPipeliningEnabled: Bool = true,
+        isErrorHandlingEnabled: Bool = true,
+        isResponseHeaderValidationEnabled: Bool = true,
```
glbrntt, line 407: I think `is` isn't right here, because it reads like a question. I think `enablePipelining`, `enableErrorHandling`, and `enableResponseHeaderValidation` are more appropriate

adam-fowler:
I've only had a quick look at this. The websocket upgrade code seems to require that every packet be tested for whether it is an http packet, or an upgraded packet. Can we not do the test on the channel once instead, ie is this channel producing http or websocket packets.

I was hoping the websocket server code could be separate from the http server code but this implementation means the two are intertwined and hard to separate into two modules.

FranzBusch:
> I've only had a quick look at this. The websocket upgrade code seems to require that every packet be tested for whether it is an http packet, or an upgraded packet. Can we not do the test on the channel once instead, ie is this channel producing http or websocket packets.
> 
> I was hoping the websocket server code could be separate from the http server code but this implementation means the two are intertwined and hard to separate into two modules.

The two are not really intertwined. The `NIOTypedHTTPServerUpgrader` is only present in the pipeline until the first request comes in. Once the first request is received, we check if it was an upgrade request and if we can handle it. Once, it is either upgraded or not the `NIOTypedHTTPServerUpgrader` removes itself from the pipeline. Hence, we only really check if the incoming reads are HTTP parts until we have received a complete HTTP request from the peer.

FranzBusch:
@Lukasa @glbrntt This PR is now ready to merge. In my last commit, I renamed the methods to `configureUpgradableHTTPServerPipeline` which aligns closer to what we are doing here. Furthermore, I added a new `NIOUpgradableHTTPServerPipelineConfiguration` which allows us to extend this configuration method in the future with new defaults without requiring new APIs.

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Whether to validate outbound response headers to confirm that they are
+    /// spec compliant. Defaults to `true`.
+    public var enableResponseHeaderValidation = true
+
+    /// The configuration for the ``HTTPResponseEncoder``.
+    public var httpResponseEncoderConfiguration = HTTPResponseEncoder.Configuration()
+
+    /// The configuration for the ``NIOTypedHTTPServerUpgradeHandler``.
+    public var httpServerUpgradeConfiguration: NIOTypedHTTPServerUpgradeConfiguration<UpgradeResult>
```
glbrntt, line 38: do we need the `http` prefixes here? I think the context from type name should be enough. Can probably drop the `Server` from the `httpServerUpgradeConfiguration` as well.

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,147 @@
...
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    @_spi(AsyncChannel)
+    public func configureUpgradableHTTPServerPipeline<UpgradeResult: Sendable>(
+        configuration: NIOUpgradableHTTPServerPipelineConfiguration<UpgradeResult>
+    ) -> EventLoopFuture<EventLoopFuture<UpgradeResult>> {
+        self._configureUpgradableHTTPServerPipeline(
+            configuration: configuration
+        )
+    }
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
```
glbrntt, line 73: nit:

```suggestion
    }

    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
```

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,147 @@
...
+    @_spi(AsyncChannel)
+    public func _configureUpgradableHTTPServerPipeline<UpgradeResult: Sendable>(
```
glbrntt, line 75: Does this need to be SPI at all? Can't it just be `internal`/`private`?

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,147 @@
...
+    @_spi(AsyncChannel)
+    public func configureUpgradableHTTPServerPipeline<UpgradeResult: Sendable>(
+        configuration: NIOUpgradableHTTPServerPipelineConfiguration<UpgradeResult>
+    ) throws -> EventLoopFuture<UpgradeResult> {
+        self.eventLoop.assertInEventLoop()
+
+        let responseEncoder = HTTPResponseEncoder(configuration: configuration.httpResponseEncoderConfiguration)
+        let requestDecoder = ByteToMessageHandler(HTTPRequestDecoder(leftOverBytesStrategy: .forwardBytes))
+
+        var extraHTTPHandlers: [RemovableChannelHandler] = [requestDecoder]
```
glbrntt, line 115: Can we reserve capacity for this? Potentially saves us a few allocs.

glbrntt (commented):
Couple of nits, looks good otherwise.

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,148 @@
...
+        extraHTTPHandlers.reserveCapacity(3)
```
glbrntt, line 116: I think you'll need to reserve _then_ add the decoder

Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
```diff
@@ -0,0 +1,148 @@
...
+    /// Whether to provide assistance handling protocol errors (e.g. failure to parse the HTTP
+    /// request) by sending 400 errors. Defaults to `true`.
+    public var enableErrorHandling = true
+
+    /// Whether to validate outbound response headers to confirm that they are
+    /// spec compliant. Defaults to `true`.
+    public var enableResponseHeaderValidation = true
+
+    /// The configuration for the ``HTTPResponseEncoder``.
+    public var httpResponseEncoderConfiguration = HTTPResponseEncoder.Configuration()
```
glbrntt, line 35: You missed this one, we can drop `http` here too

FranzBusch: Will fix both in my follow up client PR. Thanks!

PR #2518, opened 2023-09-15
Fix failed c++ compile of CNIOAtomics.h

saintstack:
Motivation:

    On c++ compile, we get 'error: unknown type name '_Bool' DECLARE_ATOMIC_OPERATIONS(_Bool, _Bool)'

    The c++ compile is happening when SwiftPM builds a module
    with swiftSettings: [.interoperabilityMode(.Cxx)] set and
    where nio is a dependency.

Modifications:

* Sources/CNIOAtomics/include/CNIOAtomics.h

 Conditionally, if __cplusplus, use bool type instead of
 _Bool (a c-type only) for booleans.

Result:

 Compiles where previous it error'd.

After your change, what will change.

 No change other than now it compiles under c++.

### Motivation:

### Modifications:

### Result:


Lukasa:

Lukasa (approved):
Nice one, thanks @saintstack!

PR #2519, opened 2023-09-20
Remove `ProtocolNegotiationHandler` protocol

FranzBusch:
# Motivation
When we created the first round of new async bootstrap APIs we added a `ProtocolNegotiationHandler` protocol to identify handlers that are doing protocol negotiation. Since then, we have changed the way how the async bootstraps work and we no longer need this protocol.

# Modification
Remove the `ProtocolNegotiationHandler` protocol


Lukasa (approved):

FranzBusch:
API break is expected. @Lukasa you mind merging?

PR #2523, opened 2023-09-26
Fix `CNIOSHA1.h` to support C++ interoperability

yaglo:
Fixes C++ interoperability by removing `static` from the array definition when compiling with C++ compiler.

### Motivation:

To be able to compile packages with C++ interoperability enabled.

### Modifications:

- Removes `static` when `__cplusplus` is defined.
- Adds `extern "C"` just in case.

### Result:

NIO fully compiles with C++ interoperability enabled.

### DIFF
--- a/Sources/CNIOSHA1/include/CNIOSHA1.h
+++ b/Sources/CNIOSHA1/include/CNIOSHA1.h
@@ -49,6 +49,10 @@
 #include <stdint.h>
 #include <stddef.h>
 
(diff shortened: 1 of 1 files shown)


Lukasa:

Lukasa (approved):
Thanks for this! ✨

Lukasa:
Merging over the nightly failure, which was not caused by this change.

PR #2531, opened 2023-10-03
Tolerate empty HTTP response body parts

dnadoba:
### Motivation

Empty HTTP response body parts currently eagerly end a response. This is unexpected and against NIOs documented behaviour.

### Modification

- add test which sends an empty response body part and would previously fail
- skip empty response body parts

### Result

Users can send empty body parts without ending the response.

### DIFF
--- a/Sources/NIOHTTP1/HTTPEncoder.swift
+++ b/Sources/NIOHTTP1/HTTPEncoder.swift
@@ -268,6 +268,10 @@ public final class HTTPResponseEncoder: ChannelOutboundHandler, RemovableChannel
                 buffer.write(response: response)
             }, context: context, headers: response.headers, promise: promise)
         case .body(let bodyPart):
+            if case let .byteBuffer(byteBuffer) = bodyPart, byteBuffer.readableBytes == 0 {
+                // empty body parts will end the response which we don't want
+                break
+            }
             writeChunk(wrapOutboundOut: self.wrapOutboundOut, context: context, isChunked: self.isChunked, chunk: bodyPart, promise: promise)
         case .end(let trailers):
             writeTrailers(wrapOutboundOut: self.wrapOutboundOut, context: context, isChunked: self.isChunked, trailers: trailers, promise: promise)
--- a/Sources/NIOHTTP1/HTTPEncoder.swift
+++ b/Sources/NIOHTTP1/HTTPEncoder.swift
@@ -15,26 +15,24 @@
 import NIOCore
 
 private func writeChunk(wrapOutboundOut: (IOData) -> NIOAny, context: ChannelHandlerContext, isChunked: Bool, chunk: IOData, promise: EventLoopPromise<Void>?) {
-    let (mW1, mW2, mW3): (EventLoopPromise<Void>?, EventLoopPromise<Void>?, EventLoopPromise<Void>?)
-
-    switch (isChunked, promise) {
-    case (true, .some(let p)):
-        /* chunked encoding and the user's interested: we need three promises and need to cascade into the users promise */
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -268,6 +268,10 @@ public final class HTTPResponseEncoder: ChannelOutboundHandler, RemovableChannel
                 buffer.write(response: response)
             }, context: context, headers: response.headers, promise: promise)
         case .body(let bodyPart):
+            if case let .byteBuffer(byteBuffer) = bodyPart, byteBuffer.readableBytes == 0 {
+                // empty body parts will end the response which we don't want
+                break
+            }
```
FranzBusch, line 274: Is this the right place to make this decision? I was expecting that we change this in the place where we currently send the `end` when we write an empty body part instead of holding back the write here.

dnadoba: This happens in llhttp which is C code transpiled from TypeScript which we don't want to change.

dnadoba: @FranzBusch done in https://github.com/apple/swift-nio/pull/2531/commits/abe6ae2c22454fc01b03667cb1ac84ffed9e5c2f

Sources/NIOHTTP1/HTTPEncoder.swift
```diff
@@ -268,6 +268,12 @@ public final class HTTPResponseEncoder: ChannelOutboundHandler, RemovableChannel
                 buffer.write(response: response)
             }, context: context, headers: response.headers, promise: promise)
         case .body(let bodyPart):
+            if case let .byteBuffer(byteBuffer) = bodyPart, byteBuffer.readableBytes == 0 {
+                // empty body parts will end the response which we don't want
+                // this happens in llhttp which is a vendored C library transpiled from TypeScript
+                // we don't want to change the upstream library and therefore handle this special case here
+                break
```
Lukasa, line 275: This fix is incorrect and the analysis isn't right either.

Firstly, the analysis. Empty body parts _only_ end the response if we're using chunked encoding. That's because the empty body chunk signals the end of the body, and so is what we send in `writeTrailers`. This has nothing to do with llhttp, it's spec-behaviour.

Secondly, the fix. This leaks a promise on the outbound path. We must pass on a zero-length write down the pipeline to ensure the promise is preserved.

Lukasa (changes requested):

Sources/NIOHTTP1/HTTPEncoder.swift
dnadoba: AFAICT `writeTrailers` is only called on `.end` message, not on `.body` and I couldn't find any special logic that says that an empty body ends the response.

In addition this would mean our documentation is wrong an `.body` can also end a response. 
https://github.com/apple/swift-nio/blob/44a74872464ae2b8ab7722455b679d63a511b183/Sources/NIOHTTP1/HTTPTypes.swift#L128-L134
IMO this is rather unexpected and I don't think our state machines (e.g. the pipelining state machine) handle this edge case either. 

Note that I'm aware that the wire protocol ends a chunked encoding with a zero length chunk. What I'm trying to say is that llhttp is sending out the empty chunk which in turn terminates the response body. I think our swift-nio contract is different though and we only want to end a body with an explicit `.end` HTTP part, not implicitly through an empty body part. 

If we decided we want to do that we need to clearly document this and I think also audit some of our existing code. The nio code that I have written at leased assumed that `.end` must be written on a successful HTTP response.

dnadoba: Oh I see, do you agree with the rest though?

Lukasa (approved):
Nice one, this LGTM.

FranzBusch (approved):

PR #2534, opened 2023-10-06
Adopt `package-benchmark`

FranzBusch:
# Motivation
We want to migrate our allocation and later on also our performance tests to use the `package-benchmark` plugin. This plugin makes writing benchmarks way easier than our current setup. Furthermore, debugging benchmarks is also possible from within Xcode now.

# Modification
This PR adds the setup for the benchmarking infrastructure and connects it with out

# Result
Allocations tests are more accessible and easier to iterate.

### DIFF
--- a/Benchmarks/.gitignore
+++ b/Benchmarks/.gitignore
@@ -1,0 +1,9 @@
+.DS_Store
+/.build
+/Packages
+xcuserdata/
+DerivedData/
+.swiftpm/configuration/registries.json
+.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+.netrc
+.benchmarkBaselines/
--- a/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
+++ b/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
@@ -1,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+
+let benchmarks = {
+    let defaultMetrics: [BenchmarkMetric] = [
+        .mallocCountTotal,
+    ]
+
+    Benchmark(
+        "TCPEcho",
+        configuration: .init(
+            metrics: defaultMetrics,
+            timeUnits: .milliseconds,
+            scalingFactor: .mega
+        )
+    ) { benchmark in
+        try runTCPEcho(numberOfWrites: benchmark.scaledIterations.upperBound)
+    }
+}
--- a/Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
+++ b/Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
@@ -1,0 +1,91 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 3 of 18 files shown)


Benchmarks/.gitignore
```diff
@@ -0,0 +1,9 @@
+.DS_Store
```
glbrntt, line 1: Can we nest `Benchmarks` in `IntegrationTests` instead of the root of the repo?

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import Foundation
```
glbrntt, line 16:
```suggestion
```

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        let buffer = self.unwrapInboundIn(data)
+        self.receivedData += buffer.readableBytes
+
+        if self.receivedData == self.numberOfWrites * self.bufferSize {
+            self.readsCompletePromise.succeed()
+            return
```
glbrntt, line 54: This is redundant
```suggestion
```

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+}
+
+func runTCPEcho(numberOfWrites: Int) throws {
+    let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
glbrntt, line 60: Should we just use the singleton?

Benchmarks/.gitignore
FranzBusch: I intentionally did not nest it since I expect us to migrate all of them at some point and having it at the top level is way nicer. This also follows what we do in other packages.

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
FranzBusch: I wanted to constrain this to a single thread.

Benchmarks/.gitignore
glbrntt: I don't think that's a good reason given `IntegrationTests` contains more than just allocation tests.

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEcho.swift
glbrntt: I think the singleton does that although this is more explicit.

FranzBusch: The singleton is by default constrained to number of threads/cgroups AFAIK

Benchmarks/.gitignore
FranzBusch: I still would like to have them at the top level. Similar to how we should move the performance tests out of the `NIOPerformanceTester` and into `Benchmarks` once we migrate them. @Lukasa @dnadoba WDYT?

Lukasa: I'm ok with having them at the top level.

Lukasa (approved):

glbrntt (approved):

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    Benchmark(
+        "TCPEcho",
+        configuration: .init(
+            metrics: defaultMetrics,
+            timeUnits: .milliseconds,
+            scalingFactor: .mega
+        )
+    ) { benchmark in
+        try runTCPEcho(numberOfWrites: 10_000)
```
hassila, line 30: Scalingfactor 1M but writing 10K? Perhaps consider numberOfWrites to be scalingfactor if you want to make sense of output from --scale later. Otherwise probably skip setting scalingfactor here, as it's really intended for inner loops in practice.

FranzBusch: So I set the scaling factor because with the default the p90 was to flaky for allocations. With `.mega` scaling factor they seem really stable. I just pushed a commit that uses the `scaledIterations.upperBound` for the number of writes

PR #2535, opened 2023-10-08
Fix missing whitespace in `README.md`

MaxDesiatov:
This missing whitespace causes the sentence to be rendered incorrectly in a much less readable way.

"It also provides[`EmbeddedChannel`][ec]" -> "It also provides [`EmbeddedChannel`][ec]"


Lukasa (approved):
Thanks Max!

PR #2536, opened 2023-10-08
Add `NIOAsyncChannel` benchmark

FranzBusch:
# Motivation
We want to benchmark our `NIOAsyncChannel` to see how it compares to the synchronous implementation with `ChannelHandler`s

# Modification
This PR adds a new `TCPEchoAsyncChannel` benchmark that mimics the `TCPEcho` benchmark but uses our new async bridges. Since Swift Concurrency, is normally using a global executor this benchmark would have quite high variation. To reduce this variant I introduced code to hook the global executor and set an `EventLoop` as the executor. In the future, if we get task executors we can change the code to us them instead.

# Result
New baseline benchmarks for the `NIOAsyncChannel`.

### DIFF
--- a/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
+++ b/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
@@ -13,6 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 import Benchmark
+import NIOPosix
+
+private let eventLoop = MultiThreadedEventLoopGroup(numberOfThreads: 1).next()
 
 let benchmarks = {
     let defaultMetrics: [BenchmarkMetric] = [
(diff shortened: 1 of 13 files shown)


Benchmarks/Thresholds/5.9/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
```diff
@@ -0,0 +1,3 @@
+{
+  "mallocCountTotal" : 5636901
```
FranzBusch, line 2: This is pretty bad since we are allocating for every single `enqueue` now. I will raise one or more follow up PRs to reduce this significantly. I was able to remove all allocations except one in the runtime.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -13,6 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 import Benchmark
+import NIOPosix
+
+private let eventLoop = MultiThreadedEventLoopGroup(numberOfThreads: 1).next()
```
glbrntt, line 18: Should we just grab the `next()` from the shared singleton?

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEchoAsyncChannel.swift
```diff
@@ -0,0 +1,89 @@
...
+                    synchronouslyWrapping: channel,
+                    configuration: .init(
+                        inboundType: ByteBuffer.self,
+                        outboundType: ByteBuffer.self
+                    )
+                )
+            }
+        }
+
+    let bufferSize = 10000
```
glbrntt, line 51: nit: naming threw me for a sec as I expected a buffer, not a `ByteBuffer`. something like `messageSize` is a more obvious

glbrntt (approved):

FranzBusch:
Thanks for the review. Going to address them in my next PR that I am going to open in a few minutes.

PR #2538, opened 2023-10-09
Add customization point for scheduling `ExecutorJob`s on `EventLoop`s

FranzBusch:
# Motivation
Currently, the NIO's EventLoop conformance to the `SerialExecutor` protocol always uses `execute` to schedule the actual job. However, the closure for `execute` has to close over the job and the `EventLoop` itself; hence, it always allocates. Since jobs are a very fine grained object in Concurrency that are created a lot this lead to millions of allocations in even small benchmarks.

# Modification
This PR provides a customization point for `EventLoop`s to execute `ExecutorJob`s directly. For `SelectableEventLoop` we store a type erased `UnownedJob` in our `ScheduledTask` and just run it right away.

# Result
No more allocations when NIO's EL is used as a `SerialExecutor`.

### DIFF
--- a/Sources/NIOCore/EventLoop+SerialExecutor.swift
+++ b/Sources/NIOCore/EventLoop+SerialExecutor.swift
@@ -25,6 +25,9 @@ public protocol NIOSerialEventLoopExecutor: EventLoop, SerialExecutor { }
 extension NIOSerialEventLoopExecutor {
     @inlinable
     public func enqueue(_ job: consuming ExecutorJob) {
+        // By default we are just going to use execute to run the job
+        // this is quite heavy since it allocates the closure for
+        // every single job.
         let unownedJob = UnownedJob(job)
         self.execute {
             unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
@@ -62,10 +65,7 @@ final class NIODefaultSerialEventLoopExecutor {
 extension NIODefaultSerialEventLoopExecutor: SerialExecutor {
     @inlinable
     public func enqueue(_ job: consuming ExecutorJob) {
-        let unownedJob = UnownedJob(job)
-        self.loop.execute {
-            unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
-        }
+        self.loop.enqueue(job)
     }
 
     @inlinable
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -317,6 +317,10 @@ public protocol EventLoop: EventLoopGroup {
     /// implementation returns a ``NIODefaultSerialEventLoopExecutor`` instead, which provides suboptimal performance.
     @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
     var executor: any SerialExecutor { get }
+
+    /// Submit a job to be executed by the `EventLoop`
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    func enqueue(_ job: consuming ExecutorJob)
     #endif
 
     /// Must crash if it is not safe to call `wait()` on an `EventLoopFuture`.
@@ -380,6 +384,18 @@ extension EventLoop {
     @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
     public var executor: any SerialExecutor {
         NIODefaultSerialEventLoopExecutor(self)
+    }
+
+    @inlinable
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    public func enqueue(_ job: consuming ExecutorJob) {
+        // By default we are just going to use execute to run the job
+        // this is quite heavy since it allocates the closure for
+        // every single job.
+        let unownedJob = UnownedJob(job)
+        self.execute {
+            unownedJob.runSynchronously(on: self.executor.asUnownedSerialExecutor())
+        }
     }
     #endif
 }
--- a/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
+++ b/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
@@ -402,8 +402,36 @@ extension MultiThreadedEventLoopGroup: CustomStringConvertible {
     }
 }
 
+#if swift(>=5.9)
+@usableFromInline
+struct ErasedUnownedJob {
+    @usableFromInline
+    let erasedJob: Any
+
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    init(job: UnownedJob) {
+        self.erasedJob = job
+    }
+
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    @inlinable
+    var unownedJob: UnownedJob {
+        // This force-cast is safe since we only store an UnownedJob
+        self.erasedJob as! UnownedJob
+    }
+}
+#endif
+
 @usableFromInline
 internal struct ScheduledTask {
+    @usableFromInline
+    enum UnderlyingTask {
+        case function(() -> Void)
+        #if swift(>=5.9)
+        case unownedJob(ErasedUnownedJob)
+        #endif
+    }
+
     /// The id of the scheduled task.
     ///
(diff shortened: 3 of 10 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -317,6 +317,10 @@ public protocol EventLoop: EventLoopGroup {
     /// implementation returns a ``NIODefaultSerialEventLoopExecutor`` instead, which provides suboptimal performance.
     @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
     var executor: any SerialExecutor { get }
+
+    /// Submit a job to be executed by the `EventLoop`
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    func execute(_ job: consuming ExecutorJob)
```
glbrntt, line 323: Do we really want to call this `execute`? Slightly worried about overloading `execute`...

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -402,30 +402,65 @@ extension MultiThreadedEventLoopGroup: CustomStringConvertible {
...
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    init(job: UnownedJob) {
+        self.erasedJob = job
+    }
+
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    @inlinable
+    var unownedJob: UnownedJob {
+        // This force-cast is safe since we only store an UnownedJob
+        self.erasedJob as! UnownedJob
```
glbrntt, line 419: This isn't guaranteed as `erasedJob` is mutable, can we make it a `let`?

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -402,30 +402,65 @@ extension MultiThreadedEventLoopGroup: CustomStringConvertible {
...
+}
+
 @usableFromInline
 internal struct ScheduledTask {
+    @usableFromInline
+    enum UnderlyingTask {
+        case function(() -> Void)
+        case unownedJob(ErasedUnownedJob)
+    }
     /// The id of the scheduled task.
```
glbrntt, line 430:
```suggestion
    }

    /// The id of the scheduled task.
```

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -515,7 +528,22 @@ Further information:
...
+                        #if swift(>=5.9)
+                        case .unownedJob(let erasedUnownedJob):
+                            if #available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *) {
+                                erasedUnownedJob.unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
+                            } else {
+                                fatalError("Tried to run an UnownedJob before we allow to schedule them")
+                            }
+                        #else
+                        case .unownedJob:
+                            fatalError("Tried to run an UnownedJob before we allow to schedule them")
```
glbrntt, line 544: Might make sense to make the `unknownJob` case conditional on Swift version

Sources/NIOCore/EventLoop.swift
FranzBusch: Good point! Let's rename this to `enqueue(_ job: consuming ExecutorJob)` to follow what `SerialExecutor` has

Sources/NIOCore/EventLoop+SerialExecutor.swift
```diff
@@ -25,10 +25,7 @@ public protocol NIOSerialEventLoopExecutor: EventLoop, SerialExecutor { }
 extension NIOSerialEventLoopExecutor {
     @inlinable
     public func enqueue(_ job: consuming ExecutorJob) {
-        let unownedJob = UnownedJob(job)
-        self.execute {
-            unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
-        }
+        self.enqueue(job)
```
Lukasa, line 28: How confident are you that this method isn't recursive?

Sources/NIOCore/EventLoop.swift
```diff
@@ -317,6 +317,10 @@ public protocol EventLoop: EventLoopGroup {
     /// implementation returns a ``NIODefaultSerialEventLoopExecutor`` instead, which provides suboptimal performance.
     @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
     var executor: any SerialExecutor { get }
+
+    /// Submit a job to be executed by the `EventLoop`
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    func enqueue(_ job: consuming ExecutorJob)
```
Lukasa, line 323: I don't think we can add the consuming keyword here without compiler guards.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -299,6 +299,19 @@ Further information:
...
+    #if swift(>=5.9)
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    @usableFromInline
+    func enqueue(_ job: consuming ExecutorJob) {
+        let scheduledTask = ScheduledTask(
+            id: self.scheduledTaskCounter.loadThenWrappingIncrement(ordering: .relaxed),
+            job: job,
+            readyTime: .now()
+        )
+        try? self._schedule0(scheduledTask)
```
Lukasa, line 311: Can you reproduce the comment above to express why we don't catch an error here.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -515,7 +528,19 @@ Further information:
...
+                        switch task.task {
+                        case .function(let function):
+                            function()
+
+                        #if swift(>=5.9)
+                        case .unownedJob(let erasedUnownedJob):
+                            if #available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *) {
+                                erasedUnownedJob.unownedJob.runSynchronously(on: self.asUnownedSerialExecutor())
+                            } else {
+                                fatalError("Tried to run an UnownedJob before we allow to schedule them")
```
Lukasa, line 540:
```suggestion
                                fatalError("Tried to run an UnownedJob without runtime support")
```

Sources/NIOCore/EventLoop.swift
FranzBusch: This is compiler guarded

Sources/NIOCore/EventLoop+SerialExecutor.swift
```diff
@@ -23,11 +23,12 @@ public protocol NIOSerialEventLoopExecutor: EventLoop, SerialExecutor { }
 
 @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
 extension NIOSerialEventLoopExecutor {
-    @inlinable
-    public func enqueue(_ job: consuming ExecutorJob) {
```
glbrntt, line 27: Why did this get removed?

FranzBusch: I thought I had to drop down to `UnownedJob`s but I think I made it work now.

Sources/NIOCore/EventLoop+SerialExecutor.swift
```diff
@@ -25,6 +25,9 @@ public protocol NIOSerialEventLoopExecutor: EventLoop, SerialExecutor { }
 extension NIOSerialEventLoopExecutor {
     @inlinable
     public func enqueue(_ job: consuming ExecutorJob) {
+        // By default we are just going to use execute to run the job
+        // this is quite heavy since it allocates the closure for
+        // every single job.
```
glbrntt, line 30: can't we just `self.enqueue(job)` now?

glbrntt (approved):

Sources/NIOCore/EventLoop+SerialExecutor.swift
FranzBusch: We can't do that because that would recursively call the method itself. I also tried removing this method here all together but that ends up in ambiguous look-up during compile time. So I think for now we just have to keep it here as well.

Lukasa (approved):

PR #2539, opened 2023-10-09
Improve performance of `NIOAsyncChannel`

FranzBusch:
### Motivation:

There were a couple of easy performance wins in `NIOAsyncChannel` which cut the allocations in half again. Those performance wins are mostly around the underlying `NIOAsyncWriter` and `NIOAsyncSequenceProducer`.

### Modifications:

This PR contains a few changes which split up into separate commits to make reviewing easier:
1. I added a fast path to the `NIOAsyncChannelInboundStreamChannelHandlerProducerDelegate` which allows it to skip the thread hop when we are already on the correct `EventLoop`.
2. I copied over the `_TinyArray` implementation from `swift-certificates` to optimise the fast path of having just a single producer suspended in the `NIOAsyncWriter`.
3. I implemented the `NIOAsyncWriterSinkDelegate` customization point for single element writes in the `NIOAsyncChannelOutboundWriterHandler`. This avoids allocating a `Deque` for single element writes.
4. I made sure that we call the single element customization point more often in `NIOAsyncWriter` and I moved the storage of the `suspendedYield`s to the `_TinyArray` which optimises for the single producer case.

### Result:

Halved the allocations for the `NIOAsyncChannel` benchmark again.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -15,6 +15,7 @@
 import Atomics
 import DequeModule
 import NIOConcurrencyHelpers
+import _NIODataStructures
 
 /// The delegate of the ``NIOAsyncWriter``. It is the consumer of the yielded writes to the ``NIOAsyncWriter``.
 /// Furthermore, the delegate gets informed when the ``NIOAsyncWriter`` terminated.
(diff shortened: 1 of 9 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -454,7 +455,7 @@ extension NIOAsyncWriter {
                 let action = self._stateMachine.setWritability(to: writability)
 
                 switch action {
-                case .callDidYieldAndResumeContinuations(let delegate, let elements, let suspendedYields):
+                case .callDidYieldAndResumeContinuations(let delegate, var elements, let suspendedYields):
```
glbrntt: Why is this now `var`?

glbrntt (approved):

FranzBusch:

PR #2540, opened 2023-10-09

FranzBusch:
# Motivation

# Modification
This moves all of the delegate calls outside of the locks.

# Result
Less allocations.

### DIFF
--- a/Benchmarks/Thresholds/5.10/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
+++ b/Benchmarks/Thresholds/5.10/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
@@ -1,3 +1,3 @@
 {
-  "mallocCountTotal" : 1317015
+  "mallocCountTotal" : 164419
 }
--- a/Benchmarks/Thresholds/5.9/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
+++ b/Benchmarks/Thresholds/5.9/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
@@ -1,3 +1,3 @@
 {
-  "mallocCountTotal" : 1317022
+  "mallocCountTotal" : 164426,
 }
--- a/Benchmarks/Thresholds/main/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
+++ b/Benchmarks/Thresholds/main/NIOPosixBenchmarks.TCPEchoAsyncChannel.p90.json
@@ -1,3 +1,3 @@
 {
-  "mallocCountTotal" : 1317015
+  "mallocCountTotal" : 164419,
 }
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelOutboundWriterHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelOutboundWriterHandler.swift
@@ -170,25 +170,34 @@ extension NIOAsyncChannelOutboundWriterHandler {
 
         @inlinable
         func didYield(contentsOf sequence: Deque<OutboundOut>) {
-            // This always called from an async context, so we must loop-hop.
-            self.eventLoop.execute {
+            if self.eventLoop.inEventLoop {
                 self.handler._didYield(sequence: sequence)
+            } else {
+                self.eventLoop.execute {
+                    self.handler._didYield(sequence: sequence)
+                }
             }
         }
 
         @inlinable
         func didYield(_ element: OutboundOut) {
-            // This always called from an async context, so we must loop-hop.
-            self.eventLoop.execute {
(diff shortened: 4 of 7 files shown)


FranzBusch:
@Lukasa @glbrntt So I am not 100% convinced this is yet correct because in my previous docs I called out we held the lock over the delegate calls to avoid reentrancy issues but the more that I am thinking about this the less I am convinced that there are actually reentrancy issues. My previous main concern was that when we call `didYield` of the delegate that we very likely reentrantly call `setWritability` of the writer's sink. This however should always be `writability = false` in well behaving implementations. Setting the writability to `false` is not going to trigger more `didYield` calls so everything should be fine.
I would appreciate if you can also think about this as well and see if you can come up with any reentrancy problems here.

FranzBusch:
> > held the lock over the delegate calls
> 
> That's pretty poor style and should ~never be done. You'll struggle with re-entrancy but also unpredicatable performance. Users may decide to log in the delegate but you definitely don't want to run the logging delegate (which is more unknown code) under a lock. So please never ever lock and call unknowable code inside of the lock, that's fraught with peril.

This was a decision that we took when we implemented the writer but I can’t recall why we took that decision. I added comments that this was done to protect against reentrant but I cannot come up with a scenario where reentrancy might be a problem right now.
I do agree though that one should never hold a lock across 3rd party code but there was something that led us there because we normally never do this.

glbrntt (approved):
Code-wise looks good. Still thinking about re-entrancy though.

> @Lukasa @glbrntt So I am not 100% convinced this is yet correct because in my previous docs I called out we held the lock over the delegate calls to avoid reentrancy issues but the more that I am thinking about this the less I am convinced that there are actually reentrancy issues. My previous main concern was that when we call didYield of the delegate that we very likely reentrantly call setWritability of the writer's sink. This however should always be writability = false in well behaving implementations. Setting the writability to false is not going to trigger more didYield calls so everything should be fine.
I would appreciate if you can also think about this as well and see if you can come up with any reentrancy problems here.

Was the concern that `didYield` could cause the stack to be blown by re-entrantly calling `didYield`?

FranzBusch:
> Was the concern that `didYield` could cause the stack to be blown by re-entrantly calling `didYield`?

I just pushed up a commit that adds some docs for this. The problem that can potentially happen is the following scenario:
1. We have two tasks that write into the `NIOAsyncWriter`
2. One of those task is on the `EventLoop`. Let's call it task `A`
3. Task `A` is writing N writes to the `NIOAsyncWriter`
4. The underlying delegate is synchronously calling `context.write` for every single write
5. The pipeline then triggers a `setWritability(to: false)` while writing out the writes from `A`
6. Now the second task does a write which gets suspended and queue in the `NIOAsyncWriter`
7. While still writing out the writes from `A` the pipeline calls `setWritability(to: true)`
8. This causes another call to `didYield` which must not be protected against reentrancy otherwise the order of the writes might interleave

All of this is very edge-casey and only really happens when one of the producers is on the same EL as the writer's delegate and even then it is unclear if a multi-producer pattern should ever guarantee ordering here but I added a simple reentrancy guard that makes sure we are doing the write thing.

FranzBusch:
After having paged back in everything that we discussed when I implemented this. The reason we held the lock across the delegate call was to avoid reordering of the various delegate call outs. The most important one being reordering of `didYield` and `didTerminate` where the terminate now happens before the yield because of a potential suspension by the scheduler.

However, this can be solved by just implementing a reentrancy protection in the writer itself. We can store a flag that the we are doing an outcall and buffer any event until that outcall is done. Once the outcall is then finished we can unbuffer the rest of the events.

weissi:
@FranzBusch why close this PR? That means it's still calling out under a lock, right?

FranzBusch:
@weissi I opened a new PR: https://github.com/apple/swift-nio/pull/2546

I decided for a new one because the implementation significantly changed and I wanted to have a new context for it. The new PR is calling out without holding the lock.

weissi:
> @weissi I opened a new PR: #2546
> 
> I decided for a new one because the implementation significantly changed and I wanted to have a new context for it. The new PR is calling out without holding the lock.

I'd recommend using this PR because the history and context of this is important. Why obfuscate it with a new PR? You can retitle the PR if you like

FranzBusch:
> I'd recommend using this PR because the history and context of this is important. Why obfuscate it with a new PR? You can retitle the PR if you like

Well I just tried that but GitHub is sadly not allowing this anymore and the `Reopen` button is greyed out.

weissi:
Screenshot 2023-10-11 at 3 52 10 pm

yeah, you can't reopen without the branch being there I think

FranzBusch:
> Weird, did you delete the branch maybe? Might want to push the original branch name again and see if it can reopen. But of course not critical

I deleted the branch when I closed the PR and then pushed a new branch up with the same commit and then GitHub got sad and didn't let me reopen.

Issue #2505, opened 2023-08-17
schedule(Repeated)(Async)Task (and friends): jitter support


weissi:
For many stable systems -- especially distributed systems -- jitter is very important to reduce synchronising things that shouldn't be synchronised. For example, say you trigger a large number of health checks once every 10 seconds, you probably don't want them to go out exactly every 10 seconds. That would cause load spikes in your and potentially the other system.

A NIO user can of course calculate jitter manually by doing `scheduleRepeatedTask(..., .seconds(10) + .microseconds(Int64.random(in: 0..<10_000))` or so but that's kinda tedious.

It would be much better if once could schedule tasks with a maximum allowed jitter amount and NIO would do the jitter "calculation". For example

```
.scheduleRepeatedTask(initialDelay: .milliseconds(0), delay: .seconds(10), maximumAllowableJitter: .milliseconds(10) { ... }
```

which means that a random `.zero ..< .milliseconds(10)` worth of jitter gets applied every iteration (different random values on each iteration).

But even for single-shot scheduled tasks, we should support jitter.

natikgadzhi:
Would that mean that every `schedule*` func gets a new overload with `maximumAllowableJitter:` as a time interval?

Lukasa:
It would indeed. This should also ideally be done without requiring a new protocol witness, though that may not be possible.

PR #2542, opened 2023-10-09
Add jitter support to recurring tasks

LorenzoFritzsch:
Add jitter support to functions managing recurring tasks.

### Motivation:

Closes https://github.com/apple/swift-nio/issues/2505.

### Modifications:

Overloaded the following `EventLoop` functions that might need jitter support:
- `flatScheduleTask`
- `scheduleRepeatedTask`
- `scheduleRepeatedAsyncTask`

with the following new parameter: `maximumAllowableJitter: TimeAmount`, which is the exclusive upper bound of the jitter range that will be added to the `delay` and `initialDelay` parameters.

### Result:

If jitter is needed when scheduling a repeated task with one of the functions listed above, it will be possible to add it by using the `maximumAllowableJitter` parameter.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -765,6 +765,31 @@ extension EventLoop {
     ) -> Scheduled<T> {
         self._flatScheduleTask(in: delay, file: file, line: line, task)
     }
+    
+    /// Schedule a `task` that is executed by this `EventLoop` after the given amount of time.
+    ///
+    /// - parameters:
+    ///     - delay: The delay between the end of one task and the start of the next.
+    ///     - maximumAllowableJitter: Upper exlusive bound of jitter range added to the `delay` paramether.
+    ///     - task: The asynchronous task to run. As everything that runs on the `EventLoop`, it must not block.
+    /// - returns: A `Scheduled` object which may be used to cancel the task if it has not yet run, or to wait
+    ///            on the full execution of the task, including its returned `EventLoopFuture`.
+    ///
+    /// - note: You can only cancel a task before it has started executing.
+    @discardableResult
+    @inlinable
+    @preconcurrency
+    public func flatScheduleTask<T>(
+        in delay: TimeAmount,
+        in maximumAllowableJitter: TimeAmount,
+        file: StaticString = #fileID,
+        line: UInt = #line,
+        _ task: @escaping @Sendable () throws -> EventLoopFuture<T>
+    ) -> Scheduled<T> {
+        let jitteredDelay = self._getJitteredDelay(delay: delay, maximumAllowableJitter: maximumAllowableJitter)
+        return self.flatScheduleTask(in: jitteredDelay, task)
+    }
+    
     @usableFromInline typealias FlatScheduleTaskDelayCallback<T> = @Sendable () throws -> EventLoopFuture<T>
 
     @inlinable
@@ -878,7 +903,8 @@ extension EventLoop {
         self._scheduleRepeatedTask(initialDelay: initialDelay, delay: delay, notifying: promise, task)
     }
     
-    /// Overload `scheduleRepeatedTask` function with jitter support.
+    /// Schedule a repeated task to be executed by the `EventLoop` with a fixed delay between the end and start of each
+    /// task.
     ///
     /// - parameters:
     ///     - initialDelay: The delay after which the first task is executed.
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -781,6 +781,31 @@ extension EventLoop {
...
+    ///     - maximumAllowableJitter: Exclusive upper bound of jitter range added to the `delay` parameter.
+    ///     - task: The asynchronous task to run. As everything that runs on the `EventLoop`, it must not block.
+    /// - returns: A `Scheduled` object which may be used to cancel the task if it has not yet run, or to wait
+    ///            on the full execution of the task, including its returned `EventLoopFuture`.
+    ///
+    /// - note: You can only cancel a task before it has started executing.
+    @discardableResult
+    @inlinable
+    @preconcurrency
+    public func flatScheduleTask<T>(
```
Lukasa, line 798: Can we also add an extension that offers a variation of `scheduleTask` with `jitter`?

Sources/NIOCore/EventLoop.swift
```diff
@@ -781,6 +781,31 @@ extension EventLoop {
     ) -> Scheduled<T> {
         self._flatScheduleTask(in: delay, file: file, line: line, task)
     }
+    
+    /// Schedule a `task` that is executed by this `EventLoop` after the given amount of time.
+    ///
+    /// - parameters:
+    ///     - delay: The delay between the end of one task and the start of the next.
```
Lukasa, line 788: This seems wrong, as `flatScheduleTask` doesn't reschedule anything.

Sources/NIOCore/EventLoop.swift
```diff
@@ -781,6 +781,31 @@ extension EventLoop {
...
+    ///     - maximumAllowableJitter: Exclusive upper bound of jitter range added to the `delay` parameter.
+    ///     - task: The asynchronous task to run. As everything that runs on the `EventLoop`, it must not block.
+    /// - returns: A `Scheduled` object which may be used to cancel the task if it has not yet run, or to wait
+    ///            on the full execution of the task, including its returned `EventLoopFuture`.
+    ///
+    /// - note: You can only cancel a task before it has started executing.
+    @discardableResult
+    @inlinable
+    @preconcurrency
```
Lukasa, line 797: This is not `@preconcurrency`.

Sources/NIOCore/EventLoop.swift
```diff
@@ -893,6 +918,29 @@ extension EventLoop {
...
+    public func scheduleRepeatedTask(
+        initialDelay: TimeAmount,
+        delay: TimeAmount,
+        maximumAllowableJitter: TimeAmount,
+        notifying promise: EventLoopPromise<Void>? = nil,
+        _ task: @escaping @Sendable (RepeatedTask) throws -> Void
+    ) -> RepeatedTask {
+        let jitteredDelay = self._getJitteredDelay(delay: delay, maximumAllowableJitter: maximumAllowableJitter)
+        return self.scheduleRepeatedTask(initialDelay: initialDelay, delay: jitteredDelay, notifying: promise, task)
```
Lukasa, line 942: Should we jitter the initial delay as well?

Sources/NIOCore/EventLoop.swift
```diff
@@ -950,7 +1028,15 @@ extension EventLoop {
         repeated.begin(in: initialDelay)
         return repeated
     }
-    
+
+    @inlinable
+    func _getJitteredDelay(
```
Lukasa, line 1033: This function should be `static`.

Lukasa (changes requested):
Thanks, this code looks broadly correct! Can we add some unit tests to validate their correct behaviour? `EmbeddedEventLoop` will be useful for this as it has a fake clock.

Sources/NIOCore/EventLoop.swift
LorenzoFritzsch: Yes will do, jittering the initial delay will avoid a first load spike, which is part of what this PR should fix

Sources/NIOCore/EventLoop.swift
LorenzoFritzsch: I misunderstood the nature of this function, I will remove this overload as it's unnecessary.

Sources/NIOCore/EventLoop.swift
LorenzoFritzsch: A function that schedules a task after a given amount of time, with that amount being jittered? If this is the case I don't see the difference with this overloaded function, and as you pointed out as there's no rescheduling jittering should be unnecessary.

Sources/NIOCore/EventLoop.swift
Lukasa: Sure, if we remove this overload then that's fine.

Lukasa (commented):
Looking great, just a quick note.

Lukasa (approved):
Nice one, LGTM.

Lukasa:

Lukasa:
Couple of build errors need to be cleaned up I'm afraid:

```
18:28:18 /swift-nio/Tests/NIOPosixTests/EventLoopTest.swift:394:13: error: mutation of captured var 'sentinel' in concurrently-executing code
18:28:18             sentinel += 1
18:28:18             ^
18:28:18 /swift-nio/Tests/NIOPosixTests/EventLoopTest.swift:846:13: error: mutation of captured var 'sentinel' in concurrently-executing code
18:28:18             sentinel += 1
18:28:18             ^
```

LorenzoFritzsch:

LorenzoFritzsch:

Lukasa:

Issue #2532, opened 2023-10-05
TimeAmount.hours(.max) and probably others crash because of overflow


weissi:
This may even be classified as a low severity security issue because someone may accept an `hours` number from a user and then do `TimeAmount.hours(userProvidedValue)`. Not a good idea but you never know what users do.

### Expected behavior

It should cap the amount to the maximum representable amount.

### Actual behavior

My program crashes

### Steps to reproduce

`TimeAmount.hours(.max)`

### SwiftNIO version/commit hash

2.58.0

### System & version information

macOS

LorenzoFritzsch:
This issue was affecting all `TimeAmount` functions besides `nanoseconds`, as it was the only function that didn't do any arithmetical operations on the input amount. The other functions, on the contrary, converted the input amount to nanoseconds, and this led to overflowing in case the input amount was `.max`. 
In order to avoid this `TimeAmount(.max)` is returned, in case the amount in nanoseconds overflows.

Here's an example from the PR:
```
@inlinable
public static func hours(_ amount: Int64) -> TimeAmount {
    let nanoseconds = amount &* (1000 * 1000 * 1000 * 60 * 60)
    return nanoseconds >= 0 ? TimeAmount(nanoseconds) : TimeAmount(.max)
}
```

weissi:
> This issue was affecting all `TimeAmount` functions besides `nanoseconds`, as it was the only function that didn't do any arithmetical operations on the input amount. The other functions, on the other hand, converted the input amount to nanoseconds, and this led to overflowing in case the input amount was `.max`. In order to avoid this `TimeAmount(.max)` is returned, in case the amount in nanoseconds overflows.
> 
> Here's an example from the PR:
> 
> ```
> @inlinable
> public static func hours(_ amount: Int64) -> TimeAmount {
>     let nanoseconds = amount &* (1000 * 1000 * 1000 * 60 * 60)
>     return nanoseconds >= 0 ? TimeAmount(nanoseconds) : TimeAmount(.max)
> }
> ```

Thanks for checking! Instead of using `&*` and a `>=` we should use `multiplyingReportingOverflow` which just tells us if it overflowed.

PR #2543, opened 2023-10-09
Fix overflow

LorenzoFritzsch:
Fixes overflow when initializing any `TimeAmount` other than `TimeAmount.nanoseconds` to `Int64.max`.

### Motivation:

Closes https://github.com/apple/swift-nio/issues/2532

### Modifications:

Added a check to the amount calculated in nanoseconds. If it overflows, `TimeAmount(.max)` is returned, otherwise `TimeAmount(nanoseconds)`.

### Result:
Program doesn't crash when setting a `TimeAmount` to be `Int64.max`. Instead, `TimeAmount(.max)` is returned.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -420,9 +420,12 @@ public struct TimeAmount: Hashable, Sendable {
     /// - parameters:
     ///     - amount: the amount of microseconds this `TimeAmount` represents.
     /// - returns: the `TimeAmount` for the given amount.
+    ///
+    /// - note: returns `TimeAmount(.max)` if the amount, once converted to nanoseconds, exceeds `Int64.max`.
     @inlinable
     public static func microseconds(_ amount: Int64) -> TimeAmount {
-        return TimeAmount(amount * 1000)
+        let nanoseconds = amount &* 1000
+        return nanoseconds >= 0 ? TimeAmount(nanoseconds) : TimeAmount(.max)
     }
 
     /// Creates a new `TimeAmount` for the given amount of milliseconds.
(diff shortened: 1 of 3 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -436,49 +436,64 @@ public struct TimeAmount: Hashable, Sendable {
     /// - parameters:
     ///     - amount: the amount of microseconds this `TimeAmount` represents.
     /// - returns: the `TimeAmount` for the given amount.
+    ///
+    /// - note: returns `TimeAmount(.max)` if the amount, once converted to nanoseconds, exceeds `Int64.max`.
     @inlinable
     public static func microseconds(_ amount: Int64) -> TimeAmount {
-        return TimeAmount(amount * 1000)
+        let nanoseconds = amount &* 1000
+        return nanoseconds >= 0 ? TimeAmount(nanoseconds) : TimeAmount(.max)
```
glbrntt, line 444: This isn't quite right: `TimeAmount` can represent a negative duration so `.microseconds(-1000)` would be incorrectly converted to `.max`. We'd also need to have unit tests for the extreme (overflow and underflow) cases to verify this behaves as expected.

glbrntt (changes requested):

Lukasa (changes requested):
As noted by @weissi 's comment in the linked issue, it would also be better to replace the unchecked operations by `reportingOverflow` methods.

Tests/NIOPosixTests/EventLoopTest.swift
```diff
@@ -1608,6 +1608,24 @@ public final class EventLoopTest : XCTestCase {
...
+    }
+    
+    func testTimeAmountCappedUnderflow() {
+        let underflowCap = TimeAmount(.min)
+        XCTAssertEqual(TimeAmount.microseconds(.min), underflowCap)
+        XCTAssertEqual(TimeAmount.milliseconds(.min), underflowCap)
+        XCTAssertEqual(TimeAmount.seconds(.min), underflowCap)
+        XCTAssertEqual(TimeAmount.minutes(.min), underflowCap)
+        XCTAssertEqual(TimeAmount.hours(.min), underflowCap)
+    }
```
glbrntt, line 1628: Sorry to be a pain, but could you move these to https://github.com/apple/swift-nio/blob/main/Tests/NIOCoreTests/TimeAmountTests.swift?

glbrntt:

glbrntt (approved):

Lukasa (approved):
Nice, this looks really good to me. Let's get the tests running.

Lukasa:

PR #2544, opened 2023-10-10
perf tests: reset BB indices after every iteration

weissi:
### Motivation:

The performance tests are assumed to be run with the same number of iterations, so we didn't reset indices. That however means if one were to increase the iterations by a few orders of magnitude, the perf tests may crash.

### Modifications:

Reset the indices every iteration

### Result:

Robustness against more iterations.


weissi:

Lukasa (approved):

weissi:
Screenshot 2023-10-10 at 4 17 32 pm

what's going in here in the lock benchmarks?

Lukasa:
Unclear: there's no obvious different in the logs.

PR #2545, opened 2023-10-10
measureRunTime use DispatchTime

rnro:
### Motivation:

Using `Date()` in benchmarks leaves us open to the possibility that a clock update could occur between the start and end times affecting the measured interval.

### Modifications:

Switch `measureRunTime` and related functions to use `DispatchTime` uptime in nanoseconds to calculate intervals

### Result:

Tests should no longer be susceptible to clock updates.

### DIFF
--- a/Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
+++ b/Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
@@ -17,16 +17,16 @@ import Foundation
 @testable import NIOPosix
 #endif
 
-public func measureRunTime(_ body: () throws -> Int) rethrows -> UInt64 {
-    func measureOne(_ body: () throws -> Int) rethrows -> UInt64 {
+public func measureRunTime(_ body: () throws -> Int) rethrows -> TimeInterval {
+    func measureOne(_ body: () throws -> Int) rethrows -> TimeInterval {
         let start = DispatchTime.now().uptimeNanoseconds
         _ = try body()
         let end = DispatchTime.now().uptimeNanoseconds
-        return end - start
+        return Double(end - start)/1_000_000
     }
 
     _ = try measureOne(body)
-    var measurements = Array(repeating: UInt64(0), count: 10)
+    var measurements = Array(repeating: 0.0, count: 10)
     for i in 0..<10 {
         measurements[i] = try measureOne(body)
     }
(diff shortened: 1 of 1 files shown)


Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
```diff
@@ -37,7 +37,7 @@ public func measureRunTime(_ body: () throws -> Int) rethrows -> TimeInterval {
 
 public func measureRunTimeAndPrint(desc: String, body: () throws -> Int) rethrows -> Void {
     print("measuring: \(desc)")
-    print("\(try measureRunTime(body))s")
+    print("\(try measureRunTime(body)/1_000_000)s")
```
glbrntt, line 40: We should probably do floating point arithmetic here

Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
```diff
@@ -119,7 +119,8 @@ func runSystemCallWrapperPerformanceTest(testAssertFunction: (@autoclosure () ->
         precondition(isDebugMode)
         print("WARNING: Syscall wrapper test: Over 100% overhead allowed. Running in debug assert configuration which allows \(allowedOverheadPercent)% overhead :(. Consider running in Release mode.")
     }
-    testAssertFunction(directCallTime * (1.0 + Double(allowedOverheadPercent)/100) > withSystemCallWrappersTime,
+    let upper = Double(directCallTime) * (1.0 + Double(allowedOverheadPercent)/100)
+    testAssertFunction( upper > Double(withSystemCallWrappersTime),
```
glbrntt, line 123:
```suggestion
    testAssertFunction(upper > Double(withSystemCallWrappersTime),
```

Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
Lukasa: Indeed, `TimeInterval` does.

Tests/NIOPosixTests/SystemCallWrapperHelpers.swift
```diff
@@ -17,16 +17,16 @@ import Foundation
 @testable import NIOPosix
 #endif
 
-public func measureRunTime(_ body: () throws -> Int) rethrows -> TimeInterval {
-    func measureOne(_ body: () throws -> Int) rethrows -> TimeInterval {
-        let start = Date()
+public func measureRunTime(_ body: () throws -> Int) rethrows -> UInt64 {
```
Lukasa, line 20: It will probably make your life easier to keep the function signature the same, and just compute the `TimeInterval` internal to `measureOne`.

glbrntt (approved):

PR #2546, opened 2023-10-11

FranzBusch:
# Motivation

# Modification
This moves all of the delegate calls outside of the locks and adds protection against reentrancy into the state machine.

# Result
Less allocations.


FranzBusch:
@Lukasa @glbrntt This is the overhauled PR that now implements reentrancy protections in the state machine. I opened it as a new PR because it has significantly more changes in the state machine now.

PR #2547, opened 2023-10-11

FranzBusch:
# Motivation

# Modification
This moves all of the delegate calls outside of the locks and adds protection against reentrancy into the state machine.

# Result
Less allocations.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -20,9 +20,8 @@ import _NIODataStructures
 /// The delegate of the ``NIOAsyncWriter``. It is the consumer of the yielded writes to the ``NIOAsyncWriter``.
 /// Furthermore, the delegate gets informed when the ``NIOAsyncWriter`` terminated.
 ///
-/// - Important: The methods on the delegate are called while a lock inside of the ``NIOAsyncWriter`` is held. This is done to
-/// guarantee the ordering of the writes. However, this means you **MUST NOT** call ``NIOAsyncWriter/Sink/setWritability(to:)``
-/// from within ``NIOAsyncWriterSinkDelegate/didYield(contentsOf:)`` or ``NIOAsyncWriterSinkDelegate/didTerminate(error:)``.
+/// - Important: The methods on the delegate might be called on arbitrary threads and the implementation must ensure
+/// that proper synchronization is in place.
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 public protocol NIOAsyncWriterSinkDelegate: Sendable {
     /// The `Element` type of the delegate and the writer.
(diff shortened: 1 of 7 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -966,8 +978,41 @@ extension NIOAsyncWriter {
                     }
                 } else {
                     // Yield hasn't been marked as cancelled.
-                    // This means we can either call the delegate or suspend
-                    return .init(isWritable: isWritable, delegate: delegate)
+
+                    if isWritable && !inDelegateOutcall {
```
glbrntt, line 982: FWIW I find switches really helpful as truth tables in these scenarios.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -966,8 +978,41 @@ extension NIOAsyncWriter {
...
+                        self._state = .streaming(
+                            isWritable: isWritable,
+                            inDelegateOutcall: true, // We are now making a call to the delegate
+                            cancelledYields: cancelledYields,
+                            suspendedYields: suspendedYields,
+                            elements: elements,
+                            delegate: delegate
+                        )
+
+                        return .callDidYield(delegate)
```
glbrntt, line 992: What happens to `sequence` in this branch? It looks like we just ignore it

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -1192,5 +1266,108 @@ extension NIOAsyncWriter {
...
+                        inDelegateOutcall: inDelegateOutcall,
+                        cancelledYields: cancelledYields,
+                        suspendedYields: suspendedYields,
+                        elements: .init(),
+                        delegate: delegate
+                    )
+
+                    return .callDidYield(delegate, elements)
+
+                } else {
```
glbrntt, line 1311: You'll never hit this branch because the condition above is `>= 1` but should be `> 1`

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -1192,5 +1266,108 @@ extension NIOAsyncWriter {
...
+                    return .callDidYieldElement(delegate, element)
+                }
+
+            case .writerFinished(var elements, let delegate, let error):
+                if elements.count == 0 {
+                    // We have returned the last buffered elements and have to
+                    // call didTerminate now.
+                    self._state = .finished(sinkError: nil)
+                    return .callDidTerminate(delegate, error)
+                } else if elements.count >= 1 {
```
glbrntt, line 1338: same bug here: should be `> 1` if you want to hit the `else`

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I see where the confusion is coming from. We are just yielding it to the delegate right away when we return `callDidYield`. The caller of this method still has the sequence.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Thanks. Good catch!

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
glbrntt: Ah, gotcha, thanks for clarifying.

glbrntt (approved):

Lukasa (approved):
LGTM, nice improvement.

PR #2548, opened 2023-10-11
Remove SPI from `NIOAsyncChannel` and new bootstrap methods

FranzBusch:
# Motivation
Over the past months, we have been working on new async bridges to make using NIO's `Channel` from Swift Concurrency possible. Since this work was far reaching we have opted to land all of it as SPI. Now the time has come and we feel confident enough to make the SPI official API. This comes after testing the new APIs in various scenarios such as HTTP 1&2, HTTP upgrades, protocol negotiation and in benchmarks.

# Modification
This PR removes the SPI from the `NIOAsyncChannel` and the bootstrap methods. The other SPI usages will be removed in follow up PRs to keep the scope of the PRs relatively small.

# Result
Everyone can use the `NIOAsyncChannel` 🚀

### DIFF
--- a/Tests/NIOCoreTests/AsyncChannel/AsyncChannelTests.swift
+++ b/Tests/NIOCoreTests/AsyncChannel/AsyncChannelTests.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 import Atomics
 import NIOConcurrencyHelpers
-@_spi(AsyncChannel) @testable import NIOCore
+@testable import NIOCore
 import NIOEmbedded
 import XCTest
 
@@ -41,7 +41,7 @@ final class AsyncChannelTests: XCTestCase {
         let thirdRead = try await iterator.next()
         XCTAssertNil(thirdRead)
 
-        try await channel.close()
+        try await channel.closeFuture.get()
     }
 
     func testAsyncChannelBasicWrites() async throws {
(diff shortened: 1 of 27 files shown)


Lukasa (approved):

Tests/NIOCoreTests/AsyncChannel/AsyncChannelTests.swift
```diff
@@ -41,7 +41,7 @@ final class AsyncChannelTests: XCTestCase {
         let thirdRead = try await iterator.next()
         XCTAssertNil(thirdRead)
 
-        try await channel.close()
+        try await channel.closeFuture.get()
```
FranzBusch, line 44: This change is interesting. It took me a second to figure out what was going on but it looks like the compiler resolves to a different method when SPI was in play and it incorrectly resolved to the default init instead of the one overloaded with `Outbound == Never`. Removing the SPI changed that and the test correctly failed with `Channel already closed`

FranzBusch:
@Lukasa @glbrntt This PR now includes the removal of all SPI in this repo. I decided against splitting it because diffs are really small always (just removing the SPI bits) and you can't review the actual API changes in this PR anyhow. That's what my previous PR #2549 was for.

glbrntt (approved):
Nice!

PR #2549, opened 2023-10-11
Add docs for the async NIO APIs

FranzBusch:
# Motivation
To ease reviewing of all the new APIs and to document some of the thought process behind them, I threw together a document that collects all of it.

We can use this document to holistcally review the APIs now before removing the SPI.

### DIFF
--- a/docs/public-async-nio-apis.md
+++ b/docs/public-async-nio-apis.md
@@ -42,7 +42,7 @@ applicable.
 
 ## Proposed APIs
 
-This section contains all the new APIs that we are adding and gives us a holstic
+This section contains all the new APIs that we are adding and gives us a holistic
 overview to review them.
 
 ### `NIOAsyncChannel`
@@ -922,10 +922,6 @@ public struct NIOProtocolNegotiationResult<NegotiationResult> where NegotiationR
     ///
     /// - Parameter deferredResult: The deferred result.
     public init(deferredResult: NIOCore.EventLoopFuture<NIOCore.NIOProtocolNegotiationResult<NegotiationResult>>)
-
-    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
-    @available(*, deprecated, renamed: "getResult")
-    public func waitForFinalResult() async throws -> NegotiationResult
 }
 
 extension EventLoopFuture {
--- a/docs/public-async-nio-apis.md
+++ b/docs/public-async-nio-apis.md
@@ -1,37 +1,38 @@
 # Async NIO bridges
 
-This is a summary of all the new APIs we had to introduce to make NIO and the
-various network protocols work with new async interfaces. The intention of this
-document is to quickly outline what we had to tackle and then show a holistic
-view over all the new APIs.
+This is a summary of all the new APIs we introduced to make NIO and the various
+network protocols work with new async interfaces. The intention of this document
+is to quickly outline the incompatibilities in the original API and then show a
+holistic view over all the new APIs.
 
 ## What APIs are needed?
 
 The first problem that we had to tackle was bridging NIO's `Channel` to Swift
 Concurrency. To do so we introduced two new foundational types - the
 `NIOAsyncSequenceProducer` and the `NIOAsyncWriter`. Those allow us to bridge
-the read and the write side of the `Channel` upholding the back-pressure across
-the bridge.
+the read and the write side of the `Channel` while propagating the back-pressure
+across the bridge.
 
-On top of those two types, we built out the `NIOAsyncChannel` which allows to
-bridge a `Channel`. To do this it inserts two channel handlers which bridge the
-read and write side using the before mentioned types.
+On top of those two types, we built out the `NIOAsyncChannel` which allows users
+to bridge a `Channel` into Swift Concurrency. To do this it inserts two channel
+handlers which bridge the read and write side using the
+`NIOAsyncSequenceProducer` and the `NIOAsyncWriter`.
 
 Next up we had to look at the bootstraps. Here the import part is that the
 `Channel`s **must** be wrapped at the correct timing otherwise there is the
 potential that reads might be dropped. This is not problematic for most of the
 bootstrap since they call their various `channelInitializer`s and
 `childChannelInitializer`s at the right time. However, there was one tricky
-bootstrap - `ServerBootstrap`. The `ServerBootstrap` is multiplexing the
-incoming connections and we have to make sure that the wrapping of the child
-channels happens at the correct time. Additionally, the new bootstrap APIs
-**must** be able to relay the type information of the configured channels to the
+bootstrap - `ServerBootstrap`. The `ServerBootstrap` multiplexes the incoming
+connections and we have to make sure that the wrapping of the child channels
+happens at the correct time. Additionally, the new bootstrap APIs **must** be
+able to relay the type information of the configured channels to the
 `bind`/`connect` methods.
 
 The next thing we had to tackle was networking protocols that dynamically
 re-configure the `ChannelPipeline`. The two examples that we provide
-implementations for is HTTP/1 protocol upgrades and
-Application-Protocol-Negotiation (ALPN) via TLS. Similar to the bootstraps we
+implementations for are HTTP/1 protocol upgrades and
+Application Protocol Negotiation (ALPN) via TLS. Similar to the bootstraps we
 have to ensure that the type information is upheld so that users can correctly
 identify which reconfiguration path has been taken.
 
@@ -1087,7 +1088,6 @@ extension Channel {
     public func configureAsyncHTTP2Pipeline<Output: Sendable>(
         mode: NIOHTTP2Handler.ParserMode,
         configuration: NIOHTTP2Handler.Configuration = .init(),
-        position: ChannelPipeline.Position = .last,
         inboundStreamInitializer: @escaping NIOChannelInitializerWithOutput<Output>
     ) -> EventLoopFuture<NIOHTTP2Handler.AsyncStreamMultiplexer<Output>>
 
--- a/docs/public-async-nio-apis.md
+++ b/docs/public-async-nio-apis.md
@@ -1079,7 +1079,6 @@ extension Channel {
     /// - Parameters:
     ///   - mode: The mode this pipeline will operate in, server or client.
     ///   - configuration: The settings that will be used when establishing the connection and new streams.
-    ///   - position: The position in the pipeline into which to insert this handler.
     ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
     ///     The output of this closure is the element type of the returned multiplexer
     /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can
@@ -1134,7 +1133,6 @@ extension ChannelPipeline.SynchronousOperations {
     /// - Parameters:
     ///   - mode: The mode this pipeline will operate in, server or client.
     ///   - configuration: The settings that will be used when establishing the connection and new streams.
-    ///   - position: The position in the pipeline into which to insert this handler.
     ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
     ///     The output of this closure is the element type of the returned multiplexer
     /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can
@@ -1143,7 +1141,6 @@ extension ChannelPipeline.SynchronousOperations {
     public func configureAsyncHTTP2Pipeline<Output: Sendable>(
         mode: NIOHTTP2Handler.ParserMode,
         configuration: NIOHTTP2Handler.Configuration = .init(),
-        position: ChannelPipeline.Position = .last,
         inboundStreamInitializer: @escaping NIOChannelInitializerWithOutput<Output>
     ) throws -> NIOHTTP2Handler.AsyncStreamMultiplexer<Output>
 }
--- a/docs/public-async-nio-apis.md
+++ b/docs/public-async-nio-apis.md
@@ -231,7 +231,9 @@ public struct NIOAsyncChannelOutboundWriter<OutboundOut> : Sendable where Outbou
     /// This method suspends if the underlying channel is not writable and will resume once the it becomes writable again.
     @inlinable public func write<Writes>(contentsOf sequence: Writes) async throws where OutboundOut == Writes.Element, Writes : Sequence
 
-    /// Send a sequence of writes into the ``ChannelPipeline`` and flush them right away.
+    /// Send an asynchronous sequence of writes into the ``ChannelPipeline``.
+    ///
+    /// This will flush after every write.
     ///
     /// This method suspends if the underlying channel is not writable and will resume once the it becomes writable again.
     @inlinable public func write<Writes>(contentsOf sequence: Writes) async throws where OutboundOut == Writes.Element, Writes : AsyncSequence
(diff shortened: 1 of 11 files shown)


docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1244 @@
...
+    ///   - position: The position in the pipeline into which to insert this handler.
+    ///   - inboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    ///     The output of this closure is the element type of the returned multiplexer
+    /// - Returns: An `EventLoopFuture` containing the `AsyncStreamMultiplexer` inserted into this pipeline, which can
+    ///     be used to initiate new streams and iterate over inbound HTTP/2 stream channels.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    public func configureAsyncHTTP2Pipeline<Output: Sendable>(
+        mode: NIOHTTP2Handler.ParserMode,
+        configuration: NIOHTTP2Handler.Configuration = .init(),
+        position: ChannelPipeline.Position = .last,
```
FranzBusch, line 1147: I think we should drop the position here because it is a type we cannot make `Sendable`

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1244 @@
...
+    /// - Parameters:
+    ///   - http1ConnectionInitializer: A callback that will be invoked if HTTP/1.1 has been explicitly
+    ///         negotiated, or if no protocol was negotiated. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    ///   - http2ConnectionInitializer: A callback that will be invoked if HTTP/2 has been negotiated, and that
+    ///         should configure the channel for HTTP/2 use. Must return a future that completes when the
+    ///         channel has been fully mutated.
+    /// - Returns: An `EventLoopFuture` of an `EventLoopFuture` containing the `NIOProtocolNegotiationResult` that completes when the channel
+    ///     is ready to negotiate.
+    internal func configureHTTP2AsyncSecureUpgrade<HTTP1Output: Sendable, HTTP2Output: Sendable>(
```
FranzBusch, line 1177: This is not part of the public API

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1244 @@
...
+    ///   - http2ConnectionInitializer: An optional callback that will be invoked only when the negotiated protocol
+    ///     is HTTP/2 to configure the connection channel.
+    ///   - http2InboundStreamInitializer: A closure that will be called whenever the remote peer initiates a new stream.
+    ///     The output of this closure is the element type of the returned multiplexer
+    /// - Returns: An `EventLoopFuture` containing a ``NIOTypedApplicationProtocolNegotiationHandler`` that completes when the channel
+    ///     is ready to negotiate. This can then be used to access the ``NIOProtocolNegotiationResult`` which may itself
+    ///     be waited on to retrieve the result of the negotiation.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    public func configureAsyncHTTPServerPipeline<HTTP1ConnectionOutput: Sendable, HTTP2ConnectionOutput: Sendable, HTTP2StreamOutput: Sendable>(
+        http2Configuration: NIOHTTP2Handler.Configuration = .init(),
```
FranzBusch, line 1203: What about the HTTP2 parser mode here @rnro?

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+    /// on ``HTTP2Frame/FramePayload`` objects as their base communication
+    /// atom, as opposed to the regular NIO `SelectableChannel` objects which use `ByteBuffer`
+    /// and `IOData`.
+    ///
+    /// Outbound stream channel objects are initialized upon creation using the supplied `streamStateInitializer` which returns a type
+    /// `Output`. This type may be `HTTP2Frame` or changed to any other type.
+    @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+    public struct AsyncStreamMultiplexer<InboundStreamOutput> {
+        /// Create a stream channel initialized with the provided closure
+        public func createStreamChannel<Output: Sendable>(_ initializer: @escaping NIOChannelInitializerWithOutput<Output>) async throws -> Output
```
FranzBusch, line 1057: I am still wondering if we should call this `createOutboundStream` this would provide a good balance to the `InboundStreamOutput` e.g. cc @rnro 
```suggestion
        public func createOutboundStream<Output: Sendable>(_ initializer: @escaping NIOChannelInitializerWithOutput<Output>) async throws -> Output
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+    /// Called when some data has been read from the remote peer.
+    ///
+    /// This should call `context.fireChannelRead` to forward the operation to the next `_ChannelInboundHandler` in the `ChannelPipeline` if you want to allow the next handler to also handle the event.
+    ///
+    /// - parameters:
+    ///     - context: The `ChannelHandlerContext` which this `ChannelHandler` belongs to.
+    ///     - data: The data read from the remote peer, wrapped in a `NIOAny`.
+    public func channelRead(context: NIOCore.ChannelHandlerContext, data: NIOCore.NIOAny)
+
+    public func channelRead(context: NIOCore.ChannelHandlerContext, responsePart: NIOHTTP1.HTTPClientResponsePart)
```
glbrntt, line 718: This presumably shouldn't be public API

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+    public init(result: NegotiationResult)
+
+    /// Intializes a new ``NIOProtocolNegotiationResult`` with a deferred result.
+    ///
+    /// - Parameter deferredResult: The deferred result.
+    public init(deferredResult: NIOCore.EventLoopFuture<NIOCore.NIOProtocolNegotiationResult<NegotiationResult>>)
+
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @available(*, deprecated, renamed: "getResult")
+    public func waitForFinalResult() async throws -> NegotiationResult
```
glbrntt, line 928: If we're moving this to public API should we delete the deprecated code?

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+
+    /// Send a write into the ``ChannelPipeline`` and flush it right away.
+    ///
+    /// This method suspends if the underlying channel is not writable and will resume once the it becomes writable again.
+    @inlinable public func write(_ data: OutboundOut) async throws
+
+    /// Send a sequence of writes into the ``ChannelPipeline`` and flush them right away.
+    ///
+    /// This method suspends if the underlying channel is not writable and will resume once the it becomes writable again.
+    @inlinable public func write<Writes>(contentsOf sequence: Writes) async throws where OutboundOut == Writes.Element, Writes : Sequence
```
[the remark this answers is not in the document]
FranzBusch: But this case is super important for performance sensitive cases otherwise we lose the batch writing API. FWIW, Rust has the same on their [Writer](https://doc.rust-lang.org/std/io/trait.Write.html) training

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+
+    /// Specify the `address` to connect to for the TCP `Channel` that will be established.
+    ///
+    /// - Parameters:
+    ///   - address: The address to connect to.
+    ///   - channelInitializer: A closure to initialize the channel. The return value of this closure is returned from the `connect`
+    ///   method.
+    /// - Returns: The result of the channel initializer.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    public func connect<Output>(to address: NIOCore.SocketAddress, channelInitializer: @escaping @Sendable (NIOCore.Channel) -> NIOCore.EventLoopFuture<Output>) async throws -> Output where Output : Sendable
```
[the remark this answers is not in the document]
FranzBusch: Good point. VSock might have happened while we did this. I will follow up in a separate PR on this.

docs/public-async-nio-apis.md
FranzBusch: Those must be public API since they are on a public type that implements a public protocol.

docs/public-async-nio-apis.md
FranzBusch: Ah good catch this should be removed. Thanks!

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+    /// The underlying channel being wrapped by this ``NIOAsyncChannel``.
+    public let channel: NIOCore.Channel
+
+    /// The stream of inbound messages.
+    ///
+    /// - Important: The `inboundStream` is a unicast `AsyncSequence` and only one iterator can be created.
+    public let inboundStream: NIOCore.NIOAsyncChannelInboundStream<Inbound>
+
+    /// The writer for writing outbound messages.
+    public let outboundWriter: NIOCore.NIOAsyncChannelOutboundWriter<Outbound>
```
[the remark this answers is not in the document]
FranzBusch: I am not yet convinced. It feels a bit magical that users should know that inbound implies stream and outbound implies writer. Could be the other way around as well.

docs/public-async-nio-apis.md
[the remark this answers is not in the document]
glbrntt: I'm absolutely not saying we shouldn't have this – I agree it's really important. 

I'm just noting that it's a shame we have to special case this, especially as on the read side we have to suspend for every message.

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1162 @@
...
+
+    /// Initializes a new ``NIOAsyncChannel`` wrapping a ``Channel``.
+    ///
+    /// - Important: This **must** be called on the channel's event loop otherwise this init will crash. This is necessary because
+    /// we must install the handlers before any other event in the pipeline happens otherwise we might drop reads.
+    ///
+    /// - Parameters:
+    ///   - channel: The ``Channel`` to wrap.
+    ///   - configuration: The ``NIOAsyncChannel``s configuration.
+    @inlinable public init(synchronouslyWrapping channel: NIOCore.Channel, configuration: NIOCore.NIOAsyncChannel<Inbound, Outbound>.Configuration = .init()) throws
```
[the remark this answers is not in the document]
FranzBusch: I think the reason why we choose the initializer is twofold:
1. It has kind of "consuming" semantics of the pipeline and acts like an end point
2. We couldn't expose it on `ChannelPipeline` but rather on the sync options and that might be hard to discover

docs/public-async-nio-apis.md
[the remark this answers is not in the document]
glbrntt: It's not difficult to discover though, both the type and the documentation are present to help users know what to do with them.

I have a pretty strong aversion to including type names in property names... and I don't think there are many places where you'd be happy with `widgetsArray` as a name vs `widgets`.

docs/public-async-nio-apis.md
glbrntt: Yeah but _this_ function isn't a `ChannelHandler` method

docs/public-async-nio-apis.md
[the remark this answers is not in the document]
glbrntt: > It has kind of "consuming" semantics of the pipeline and acts like an end point

You're right, the key is in the name: "wrapping".

glbrntt (approved):

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
+# Async NIO bridges
+
+This is a summary of all the new APIs we had to introduce to make NIO and the
```
Lukasa, line 3:
```suggestion
This is a summary of all the new APIs we introduced to make NIO and the
```

"We had to introduce" makes it sound like a complaint.

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+various network protocols work with new async interfaces. The intention of this
+document is to quickly outline what we had to tackle and then show a holistic
```
Lukasa, line 5:
```suggestion
document is to quickly outline the incompatibilities in the original API and then show a holistic
```

Again, this wasn't a burden, this was an opportunity!

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+view over all the new APIs.
+
+## What APIs are needed?
+
+The first problem that we had to tackle was bridging NIO's `Channel` to Swift
+Concurrency. To do so we introduced two new foundational types - the
+`NIOAsyncSequenceProducer` and the `NIOAsyncWriter`. Those allow us to bridge
+the read and the write side of the `Channel` upholding the back-pressure across
```
Lukasa, line 13:
```suggestion
the read and the write side of the `Channel` while propagating back-pressure across
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+the bridge.
+
+On top of those two types, we built out the `NIOAsyncChannel` which allows to
```
Lukasa, line 16:
```suggestion
On top of those two types, we built out the `NIOAsyncChannel` which allows users to
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+bridge a `Channel`. To do this it inserts two channel handlers which bridge the
```
Lukasa, line 17:
```suggestion
bridge a `Channel` into Swift Concurrency. To do this it inserts two channel handlers which bridge the
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+read and write side using the before mentioned types.
```
Lukasa, line 18: Mention the types again. It doesn't hurt.

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+
+Next up we had to look at the bootstraps. Here the import part is that the
+`Channel`s **must** be wrapped at the correct timing otherwise there is the
+potential that reads might be dropped. This is not problematic for most of the
+bootstrap since they call their various `channelInitializer`s and
+`childChannelInitializer`s at the right time. However, there was one tricky
+bootstrap - `ServerBootstrap`. The `ServerBootstrap` is multiplexing the
```
Lukasa, line 25:
```suggestion
bootstrap - `ServerBootstrap`. The `ServerBootstrap` multiplexes the
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+incoming connections and we have to make sure that the wrapping of the child
+channels happens at the correct time. Additionally, the new bootstrap APIs
+**must** be able to relay the type information of the configured channels to the
+`bind`/`connect` methods.
+
+The next thing we had to tackle was networking protocols that dynamically
+re-configure the `ChannelPipeline`. The two examples that we provide
+implementations for is HTTP/1 protocol upgrades and
```
Lukasa, line 33:
```suggestion
implementations for are HTTP/1 protocol upgrades and
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1156 @@
...
+Application-Protocol-Negotiation (ALPN) via TLS. Similar to the bootstraps we
```
Lukasa, line 34:
```suggestion
Application Layer Protocol Negotiation (ALPN) via TLS. Similar to the bootstraps we
```

docs/public-async-nio-apis.md
Lukasa: We almost certainly have to drop it: semantically the _only_ place we can do this is last. What does it mean to add your async channel handlers not last?

docs/public-async-nio-apis.md
rnro: This is the function for configuring a server pipeline so the parser mode is hardcoded to `.server`

docs/public-async-nio-apis.md
rnro: If we are dropping consistency with the old multiplexer then I think `createOutboundStream` is an improvement, but I'm now actually wondering if we should just go for `createStream`, do we need to disambiguate against somehow creating an inbound stream? cc @glbrntt

docs/public-async-nio-apis.md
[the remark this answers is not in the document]
rnro: I think I'm with George FWIW

docs/public-async-nio-apis.md
rnro: There's still something which feels a bit vague about how we call `createStream`/`createOutboundStream` but then hand back the caller basically anything. Our assumption is that they wrapped the channel in the initializer and returned that, which might be fair enough but I'm not sure that the word `initializer` conveys all that (new to async-land) responsibility.

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1153 @@
...
+`NIOAsyncSequenceProducer` and the `NIOAsyncWriter`. Those allow us to bridge
+the read and the write side of the `Channel` while propagating the back-pressure
+across the bridge.
+
+On top of those two types, we built out the `NIOAsyncChannel` which allows users
+to bridge a `Channel` into Swift Concurrency. To do this it inserts two channel
+handlers which bridge the read and write side using the
+`NIOAsyncSequenceProducer` and the `NIOAsyncWriter`.
+
+Next up we had to look at the bootstraps. Here the import part is that the
```
rnro, line 21:
```suggestion
Next we had to look at the bootstraps. Here the important part was recognizing that the
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1153 @@
...
+connections and we have to make sure that the wrapping of the child channels
+happens at the correct time. Additionally, the new bootstrap APIs **must** be
+able to relay the type information of the configured channels to the
+`bind`/`connect` methods.
+
+The next thing we had to tackle was networking protocols that dynamically
+re-configure the `ChannelPipeline`. The two examples that we provide
+implementations for are HTTP/1 protocol upgrades and
+Application Protocol Negotiation (ALPN) via TLS. Similar to the bootstraps we
+have to ensure that the type information is upheld so that users can correctly
```
rnro, line 36: Upheld is okay here but retained maybe makes more sense?
```suggestion
have to ensure that the type information is retained so that users can correctly
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1153 @@
...
+identify which reconfiguration path has been taken.
+
+Lastly, we had to look at how to handle protocols that multiplex like HTTP/2.
```
rnro, line 39:
```suggestion
Lastly, we had to look at how to handle protocols that multiplex, like HTTP/2.
```

docs/public-async-nio-apis.md
```diff
@@ -0,0 +1,1153 @@
...
+Those protocols need to expose a typed async interface to consume new inbound
```
rnro, line 40:
```suggestion
Multiplexing protocols need to expose a typed async interface to consume new inbound
```

docs/public-async-nio-apis.md
glbrntt: - `createStreamChannel` is slightly odd because we don't return a `Channel`
- `createOutboundStream` is better, but it's not possible to create inbound streams so I don't think the "outbound" distinction is important
- `createStream` is better still

> There's still something which feels a bit vague about how we call createStream/createOutboundStream but then hand back the caller basically anything. Our assumption is that they wrapped the channel in the initializer and returned that, which might be fair enough but I'm not sure that the word initializer conveys all that (new to async-land) responsibility.

This is true but I don't know if it's really an issue; regardless of what the user does we're still creating a stream.

rnro: @FranzBusch I can see your point but I personally think it's okay because `createStream` requires that you pass in an initializer (maybe an argument for keeping that name) so it's clear that it won't use the `inboundStreamInitializer`.

rnro: @glbrntt 
> regardless of what the user does we're still creating a stream

Good point

glbrntt: We're assuming `inboundStreamIntializer` is the right name, which isn't necessarily the case (although it matters much less IMO because trailing closure syntax). Maybe it should also just be `initializer`.

FranzBusch: IMO naming it just `initializer` is a step back. We just had recent feedback in the forums that a users was confused about what that initializer does. So become more vague feels like a step in the wrong direction.

glbrntt: > We just had recent feedback in the forums that a users was confused about what that initializer does. 

Can you share a link to this?

Neither `inboundStreamIntializer` nor `initializer` will tell you _what_ an initialiser does. `inboundStream` tells you what the initialiser applies to, without `inboundStream` you assume from the context what it applies to.

FranzBusch: The comment was in the Swift Open Source Slack actually and not in the forums. This was the users feedback

> update: i finally got an HTTP/2 request working, the piece i was missing was that you still need to add your application stream handler in NIOHTTP2Handler.StreamMultiplexer.createStreamChannel(_:), even if you already provided a callback when setting up the handlers with Channel.configureHTTP2Pipeline. which was very counterintuitive given how the docs are worded:
https://swiftinit.org/docs/swift-nio-http2/niohttp2/niohttp2handler.multiplexer
it gave the impression that if you provide the callback during outer setup then the multiplexer is “ready to go” and you can just return $0.eventLoop.createSucceededVoidFuture(). please update these docs :slightly_smiling_face:

To be fair it is more about the documentation explaining the different of the two closures and why both are needed.

Personally, I think there is still value in having outbound in the method name.

glbrntt: Yeah that's absolutely about the docs and not about the naming. The linked docs do call out that it's an outbound stream which wasn't sufficient so I'm not sure keeping it in the name makes a difference. That said, this isn't a hill I'll die on!

PR #2550, opened 2023-10-13
Avoid terminating when a precondition is not met in HTTPServerPipelineHandler

hamzahrmalik:
### Motivation:

`HTTPServerPipelineHandler` has a few preconditionFailures in its state machine. This isn’t really ideal, as while these _are_ important invariants, it is fairly easy for users to accidentally break them.

### Modifications:

1. The first time a user hits an “illegal” state transition, we fire an error down the pipeline
2. The second time they hit one, we forcibly close the pipeline (and send another error)

We continue to crash (by calling assertionFailure) when in debug mode, to make it obvious to users that they have violated an invariant
However, an internal flag allows us to disable this behaviour for testing

### Result:

Making an illegal state change in `HTTPServerPipelineHandler` will no longer crash

### DIFF
--- a/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
+++ b/Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
@@ -298,7 +298,7 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
     }
 
     private func deliverOneMessage(context: ChannelHandlerContext, data: NIOAny) -> ConnectionStateAction {
-        self.checkPrecondition(self.lifecycleState != .quiescingLastRequestEndReceived &&
+        self.checkAssertion(self.lifecycleState != .quiescingLastRequestEndReceived &&
                                self.lifecycleState != .quiescingCompleted,
                                "deliverOneMessage called in lifecycle illegal state \(self.lifecycleState)")
         let msg = self.unwrapInboundIn(data)
@@ -306,12 +306,12 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
         debugOnly {
             switch msg {
             case .head:
-                self.checkPrecondition(self.nextExpectedInboundMessage == .head)
+                self.checkAssertion(self.nextExpectedInboundMessage == .head)
                 self.nextExpectedInboundMessage = .bodyOrEnd
             case .body:
-                self.checkPrecondition(self.nextExpectedInboundMessage == .bodyOrEnd)
+                self.checkAssertion(self.nextExpectedInboundMessage == .bodyOrEnd)
             case .end:
-                self.checkPrecondition(self.nextExpectedInboundMessage == .bodyOrEnd)
+                self.checkAssertion(self.nextExpectedInboundMessage == .bodyOrEnd)
                 self.nextExpectedInboundMessage = .head
             }
         }
@@ -354,7 +354,7 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
     public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
         switch event {
         case is ChannelShouldQuiesceEvent:
-            self.checkPrecondition(self.lifecycleState == .acceptingEvents,
+            self.checkAssertion(self.lifecycleState == .acceptingEvents,
                                    "unexpected lifecycle state when receiving ChannelShouldQuiesceEvent: \(self.lifecycleState)")
             switch self.state {
             case .responseEndPending:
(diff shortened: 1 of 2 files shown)


Lukasa:

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -477,6 +527,31 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+        case .warnPreconditionViolated(let message):
+            self.preconditionFailed(message)
+            let error = ConnectionStateError.preconditionViolated(message: message)
+            self.deliverOneError(context: context, error: error)
+            promise?.fail(error)
+            return true
+        case .forceCloseConnection:
+            let message = "The connection has been forcefully closed because further IO was attempted after a precondition was violated"
+            let error = ConnectionStateError.preconditionViolated(message: message)
+            self.deliverOneError(context: context, error: error)
```
Lukasa, line 546: We must not deliver this error: it risks an infinite recursion. It's fine for us to just silently close.

Tests/NIOHTTP1Tests/HTTPServerPipelineHandlerTest.swift
```diff
@@ -1181,4 +1181,73 @@ class HTTPServerPipelineHandlerTest: XCTestCase {
...
+
+    func testServerRequestEndFirstIsError() throws {
+        self.pipelineHandler.failOnPreconditions = false
+        // End sending a request which was never startedq
+        XCTAssertThrowsError(try self.channel.writeInbound(HTTPServerRequestPart.end(nil))) { error in
+            XCTAssertEqual(error as? HTTPServerPipelineHandler.ConnectionStateError, .preconditionViolated(message: "Received second request"))
+        }
+    }
+
+    func testForcefulShutdownWhenViolatedPrecondition() async throws {
```
Lukasa, line 1217: We cannot use `EmbeddedChannel` in an async test, but we don't need to. EmbeddedChannel runs strictly synchronously, so tests that use it are 100% reliable. This test should either pass or fail.

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -73,15 +73,44 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+                    return message
+                }
+            }
+        }
+
+        var base: Base
+        var file: String
+        var line: Int
+
+        init(base: Base, file: String, line: Int) {
```
Lukasa, line 93: This, and the various `var`s, may as well be `private`.

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -66,6 +61,59 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+
+    public struct ConnectionStateError: Error, CustomStringConvertible, Hashable {
+        enum Base: Hashable, CustomStringConvertible {
+            /// A precondition was violated
+            case preconditionViolated(message: String)
+
+            var description: String {
+                switch self {
+                case .preconditionViolated(let message):
+                    return message
```
FranzBusch, line 84: We should add a `"precondition violated"` to this description as well

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -92,53 +140,76 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
                 // response.
                 self = .responseEndPending
+                return .none
             case .sentCloseOutputRequestEndPending:
                 // Got the request end we were waiting for.
                 self = .sentCloseOutput
+                return .none
             case .responseEndPending, .idle, .sentCloseOutput:
-                preconditionFailure("Received second request")
+                let message = "Received second request"
```
FranzBusch, line 203: Also here in the `responseEndPending` case the peer violated the protocol. Should we really send an error down the pipeline first?

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
```diff
@@ -92,53 +140,76 @@ public final class HTTPServerPipelineHandler: ChannelDuplexHandler, RemovableCha
...
+                return .forceCloseConnection
             case .idle:
                 self = .requestAndResponseEndPending
+                return .none
             case .requestAndResponseEndPending, .responseEndPending, .requestEndPending,
                     .sentCloseOutputRequestEndPending, .sentCloseOutput:
-                preconditionFailure("received request head in state \(self)")
+                let message = "received request head in state \(self)"
+                self = .preconditionFailed
+                return .warnPreconditionViolated(message: message)
```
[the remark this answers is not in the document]
Lukasa: We expect the user to close the channel, and we will end up closing it if they do not (because they will trigger another such error). However, it is better to do this in a way that hits the user's error handling code, as that will help surface it to them. They ought to have an error handler in the pipeline, and that error handler will log. If they do not, if they have an async channel, then the inbound stream will throw. All of these are ideal behaviours.

We should only close _for_ the user if it is clear that they are not recovering from the situation themselves.

Sources/NIOHTTP1/HTTPServerPipelineHandler.swift
Lukasa: The peer didn't violate the protocol because they can't. These messages are structured: they come from the parser, which cannot emit them in the wrong order. This error is a user error: it means they reordered the messages.

Lukasa (commented):
Looks really good, just one very minor nit.

Lukasa (approved):
Cool, this LGTM. Nice one!

PR #2551, opened 2023-10-13
Disable `testTaskCancel_whenStreaming_andNotSuspended` test

FranzBusch:
# Motivation
This test relies heavily on timing and there are scenarios where the scheduling of the various tasks and threads can make this test fail. Currently, there is not enough control in Swift Concurrency to write a non-flaky test.

# Modification
This PR disables the test until we get better control over the executor.

# Result
One disabled test but also less flakiness.

### DIFF
--- a/Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
+++ b/Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
@@ -494,25 +494,27 @@ final class NIOThrowingAsyncSequenceProducerTests: XCTestCase {
     }
 
     func testTaskCancel_whenStreaming_andNotSuspended() async throws {
+        throw XCTSkip("Disabled until we get better executor control")
+
         // We are registering our demand and sleeping a bit to make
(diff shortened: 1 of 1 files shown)


Lukasa (commented):
Do we need to disable this? Why can't we send a signal from one thread to the other to ensure they move in roughly lockstep?

FranzBusch:
> Do we need to disable this? Why can't we send a signal from one thread to the other to ensure they move in roughly lockstep?

Yes sadly we have to. The intention of the test is to trigger cancellation once the consuming task is suspended on the `next` call; however, there is currently no way to observe that observation. That's why this test had a couple of sleeps to try to get the timing right and wait for the task to suspended but this can sadly fail and we cancel before we get to suspend.

Lukasa:
Is that assessment true? The comments seem to imply that cancelling the other task in the sleep is fine too.

FranzBusch:
Okay so actually the report of the flaky test is a bit odd since it is on a quite old NIO version and the line where the assert is trigged doesn't match with what the NIO version in play is. I will close this PR for now.

Issue #2434, opened 2023-05-26
ClientBootstrap.resolver is hard to use for single-use resolvers


weissi:
Some resolvers (like NIO's internal GAIResolver) are use-exactly-once objects. They work only if used once (not 0 times, not twice).

ClientBootstrap.resolver is hard to use with those objects because one might want to pass around a pre-configured ClientBootstrap and then not use it or use it multiple times.

Maybe ClientBootstrap could offer a resolverFactory method that gets invoked exactly when we actually do the resolution.

The other option would that NIO could provide a Resolver type that wraps use-once resolvers and managed their lifetime (creates a fresh one when a resolution actually happens).

Lukasa:
I think the latter idea is the best one. This seems like a good contributor patch.

LorenzoFritzsch:
I created a draft PR for the following reason:
This implementation uses a new instance of the resolver each time a query is made. 
However, in the case of `GetaddrinfoResolver`, when it's initialized it creates 2 promises, that are resolved by the `initiateAQuery` and `initiateAAAAQuery` functions, so if they aren't both called the promises won't be fully fulfilled and this will result in a error. I'm not sure if this is a specific behavior of `GetaddrinfoResolver` or not, so this implementation always calls both of those functions each time a query is made, this should assure no promises are leaked.

Being a newbie here I'm not really sure if there's a better way to do this, do you have any insight on this?

Lukasa:
The implementation should always call both of those functions, that's a hard requirement.

PR #2553, opened 2023-10-15
Implement wrapper for use-once resolvers

LorenzoFritzsch:
Implement a use-once resolvers wrapper to obscure the wrapped resolver's single-use nature.

### Motivation:

Closes https://github.com/apple/swift-nio/issues/2434

### Modifications:
Implemented a new `DynamicResolver` class that wraps a use-exactly-once resolver and manages its lifecycle, creating a new instance of the resolver each time a new resolution is performed.

### Result:
It's now possible to wrap a single-use resolver in a `DynamicResolver` to use it in situations where it might be used never or more than once.

### DIFF
--- a/Sources/NIOPosix/DynamicResolver.swift
+++ b/Sources/NIOPosix/DynamicResolver.swift
@@ -1,0 +1,61 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import NIOCore
+
+/// A protocol that covers an object that does DNS A and AAAA queries via a single method.
+protocol DNSResolver {
+    associatedtype T
+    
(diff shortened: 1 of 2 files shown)


Sources/NIOPosix/DynamicResolver.swift
```diff
@@ -0,0 +1,64 @@
...
+    
+    func cancelQueries() {
+        resolver.cancelQueries()
+    }
+    
+    func getResolver() -> T {
+        if resolverStatus.load(ordering: .acquiring) == 2 {
+            resolver = resolverFactory()
+            resolverStatus.store(0, ordering: .releasing)
+        }
```
Lukasa, line 61: This implementation isn't correct. Prefer using a `NIOLockedValueBox`.

Sources/NIOPosix/DynamicResolver.swift
```diff
@@ -0,0 +1,64 @@
...
+    // * 2: Resolver has initiated the second and last query, time to refresh it.
+    private let resolverStatus: ManagedAtomic<Int>
+    
+    init(resolverFactory: @escaping () -> T) {
+        self.resolverFactory = resolverFactory
+        resolver = resolverFactory()
+        resolverStatus = ManagedAtomic(0)
+    }
+    
+    func initiateAQuery(host: String, port: Int) -> NIOCore.EventLoopFuture<[NIOCore.SocketAddress]> {
```
Lukasa, line 37: I don't think these implementations are correct in the face of multiple concurrent invocations. We probably want to defend against that. We may actually need a new protocol to pull that off.

PR #2558, opened 2023-10-17
Remove continuation resumption inside locks

rnro:
### Motivation:

Whilst investigating deadlocking code elsewhere we discovered that resuming a continuation inside a lock can call deadlocks.

The reason is that
- `withTaskCancellationHandler` takes a sequence's underlying runtime lock then executes the cancellation handler.
- The task cancellation handler then does work which requires taking the NIO-level lock
- If a NIO method on a separate thread then takes the NIO-level lock and within it resumes a continuation e.g. in `yield` we will deadlock because the resumption attempts to obtain the underlying runtime lock.

### Modifications:

Do not resume continuations within locks.

### Result:

Fewer deadlock opportunities

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -450,6 +450,9 @@ extension NIOAsyncWriter {
 
         @inlinable
         /* fileprivate */ internal func setWritability(to writability: Bool) {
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):
Thanks for fixing this!

johnfairh:
Nice one fixing this - I started sporadically hitting this today and was just figuring out which locks were involved and dreading writing up the issue.

PR #2561, opened 2023-10-18
Add support for async VSock bootstrap methods

FranzBusch:
# Motivation
We are about to release our new async bootstrap methods that retain the type information of the various initializers. While developing those we also added support for VSock in NIO; however, we missed adding support for the async bootstrap methods.

# Modification
This PR adds new async bootstrap methods that take a `VSockAddress` and a test for it.

# Result
Support for async VSock bootstrap.

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -556,6 +556,10 @@ extension ServerBootstrap {
     ///
     /// - Parameters:
     ///   - vsockAddress: The VSOCK socket address to bind on.
+    ///   - serverBackPressureStrategy: The back pressure strategy used by the server socket channel.
+    ///   - channelInitializer: A closure to initialize the channel. The return value of this closure is returned from the `connect`
+    ///   method.
+    /// - Returns: The result of the channel initializer.
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
     public func bind<Output: Sendable>(
         to vsockAddress: VsockAddress,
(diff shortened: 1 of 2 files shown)


simonjbeaumont (approved):

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -552,6 +552,36 @@ extension ServerBootstrap {
         )
     }
 
+    /// Bind the `ServerSocketChannel` to a VSOCK socket.
+    ///
+    /// - Parameters:
+    ///   - vsockAddress: The VSOCK socket address to bind on.
```
glbrntt, line 558: nit: missing documentation for the other parameters

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1123,6 +1154,40 @@ extension ClientBootstrap {
         )
     }
 
+    /// Specify the VSOCK address to connect to for the `Channel`.
+    ///
+    /// - parameters:
+    ///     - address: The VSOCK address to connect to.
```
glbrntt, line 1160: nit: missing some documentation here too

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -1123,6 +1154,40 @@ extension ClientBootstrap {
...
+    /// - returns: An `EventLoopFuture<Channel>` for when the `Channel` is connected.
```
glbrntt, line 1161: The documentation for `returns` isn't right either

glbrntt (approved):
The docs need fixing up but LGTM otherwise

PR #2562, opened 2023-10-19
NonBlockingFileIO: tolerate chunk handlers from other ELs

weissi:
### Motivation:

Anywhere in SwiftNIO a user supplies a future, NIO should tolerate futures from any EL.
That wasn't the case in `NonBlockingFileIO.readChunked`.

### Modifications:

Hop the `chunkHandler`'s future to the correct EL.

### Result:

More user-friendly and correct


Lukasa (approved):
Nice one, thanks!

PR #2563, opened 2023-10-20
Support disabling body aggregation in NIOHTTP1TestServer

simonjbeaumont:
### Motivation:

NIOTestUtils provides NIOHTTP1TestServer which is useful for testing client-server interactions. Currently it uses an inbound channel handler to aggregate the entire body, which means it cannot be used to test streaming requests.

### Modifications:

- Add another initializer, which takes an `aggregateBody` boolean parameter.
- Only add `AggregateBodyHandler` to the channel pipeline if `aggregateBody` is true.

### Result:

Adopters can use `NIOHTTP1TestServer` to test streaming request patterns.


Lukasa (approved):
Love it, very good.

Issue #2074, opened 2022-04-08
Think about adding a poll-loop mode where Selectable EventLoop can be driven from other threads without taking them over


weissi:
Recently, two projects ([PureSwift/Socket](https://github.com/PureSwift/Socket) and [swhitty/FlyingFox](https://github.com/swhitty/FlyingFox)) have caught my attention that both run BSD sockets code inside Swift Concurrency's threads.

Let's first establish what exactly these libraries are doing and why: Without blocking one (or more) of Swift Concurrency's threads forever (which should be avoided at all cost) at the moment you cannot do any I/O in there unless you implement a "poll loop". A poll loop constitutes scheduling a "poll" to quickly check all sockets for available events every so often. For example it could schedule a poll of the sockets every 10 milliseconds. If there are available events (such as a socket is readable/writable/...) these events get handled and control is returned back to the Swift Concurrency default executor.

So far, this sounds kinda good: We can share a thread with Swift Concurrency, we can do I/O and we don't have to switch threads. But there are a number of issues:

- Poll loops _always_ consume energy (battery) even if nothing's happening (bad for anything with a battery)
- The _minimum_ latency is the frequency of the poll timer

In other words: The better the latency is the worse the idle energy consumption is.

Personally, I have not come across a scenario where implementing such a poll loop makes sense which is why I never really thought that SwiftNIO should add a mode for running in a poll loop (which wouldn't be hard to implement).

Why do I think that this doesn't really make sense? SwiftNIO interops just fine with Swift Concurrency if given its own thread pool where it does I/O. Yes, to interact with Swift Concurrency code it will need to thread switch from and to the Swift Concurrency threads (from NIO's I/O threads, usually called the `EventLoop`). And yes, switching threads adds latency. But: As discussed above, the poll loops _also_ add latency and additionally burn energy when idle. To be competitive in latency with a traditional SwiftNIO setup where NIO has its own threads and you switch back and forth you'd need to schedule the poll loop timer so fast that you'd be burning a lot of energy for no good reason.

So why am I even filing this issue? Maybe I missed something and there is in fact a scenario where such a pool loop makes sense. If that's the case, please comment. If convinced I think SwiftNIO should just add a mode where it can run as a pool loop scheduled in somebody else's thread, the implementation shouldn't be hard at all.

hassila:
Just a related thought (not advocating for a poll loop mode though)

[...]

PR #2564, opened 2023-10-20
allow setting MTELG.singleton as Swift Concurrency executor

weissi:
### Motivation:

There are situations where the currently necessary thread switches between Swift Concurrency's default global executor and a place that can do I/O (such as Swift NIO's EventLoops) can be very detrimental to the overall system performance.

The right solution is something akin to upcoming [the Task Executor preference proposal](https://github.com/apple/swift-evolution/pull/2187/files) but for now (Swift 5.9) there is another solution. Instead of separating the threads of Swift Concurrency & SwiftNIO entirely we can replace Swift Concurrency's default global executor by a more powerful executor: SwiftNIO's EventLoopGroups.

The result is that almost all non-`MainActor` `async` functions will then run on SwiftNIO's EventLoops. That means if those have to perform I/O they don't necessarily have to switch to a separate I/O thread (such as NIO's EventLoops) anymore because they might be already there.

Be mindful however: This approach might not work well for all workloads and is a _global_ setting.

Some more information about this can be found in:
- https://forums.swift.org/t/future-of-swift-nio-in-light-of-concurrency-roadmap/41633/6
- #2074 

### Modifications:

Provide a `NIOSingletons.unsafeTryInstallSingletonPosixEventLoopGroupAsConcurrencyGlobalExecutor()` method that installs `MultiThreadedEventLoopGroup.singleton` as Swift Concurrency's global executor.

### Result:

Faster I/O on Swift Concurrency.


Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
```diff
@@ -0,0 +1,91 @@
...
+
+private let _haveWeTakenOverTheConcurrencyPool = ManagedAtomic(false)
+extension NIOSingletons {
+    /// Install ``MultiThreadedEventLoopGroup/singleton`` as Swift Concurrency's global executor.
+    ///
+    /// This allows to use Swift Concurrency and retain high-performance I/O alleviating the otherwise necessary thread switches between
+    /// Swift Concurrency's own global pool and a place (like an `EventLoop`) that allows to perform I/O
+    ///
+    /// This method uses an atomic compare and exchange to install the hook (and makes sure it's not already set). This unilateral atomic memory
+    /// operation doesn't guarantee anything becuase another piece of code could have done the same without using atomic operations. But we
```
fabianfett, line 12:
```suggestion
    /// operation doesn't guarantee anything because another piece of code could have done the same without using atomic operations. But we
```

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
```diff
@@ -0,0 +1,107 @@
+import Atomics
+import NIOCore
+
+private protocol SilenceWarning {
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    func enqueue(_ job: UnownedJob)
```
ktoso, line 7: Yeah huh fun workaround. Okey

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// Swift Concurrency's own global pool and a place (like an `EventLoop`) that allows to perform I/O
+    ///
+    /// This method uses an atomic compare and exchange to install the hook (and makes sure it's not already set). This unilateral atomic memory
+    /// operation doesn't guarantee anything because another piece of code could have done the same without using atomic operations. But we
+    /// do our best.
+    ///
+    /// - warning: You may only call this method from the main thread.
+    /// - warning: You may only call this method once.
+    @discardableResult
+    public static func unsafeTryInstallSingletonPosixEventLoopGroupAsConcurrencyGlobalExecutor() -> Bool {
```
ktoso, line 27: Thx for prefixing with unsafe

ktoso (approved):
Thanks for all the checks before allowing the hooking. This looks as okey as it can be heh 😉 

let’s get task executors in soon! :-)

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
weissi: Yes thought `unsafe` + a loooong name is okay, can easily make it disappear. Also it's a `*Try*` function. Nothing wrong to just turn this into a `return false` in a few years.

FranzBusch: What if we hide it behind an SPI? We have been recently discussing using an SPI for exactly those scenarios where we want users to make a very very conscious choice about it

weissi:
> let’s get task executors in soon! :-)

Yes please

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
weissi: Hmm, this is not a complex API (just one function) so I don't think we need to use SPI for that. If we deprecate that, it doesn't leave a huge wrinkle on the NIO API surface.

PeterAdams-A: Should we document when not to do this?  Should we underscore the name?  Presumably it's a poor trade for most people unless they really know they want it.  I'm guessing likely disastrous for anyone running a UI app.

Lukasa: I don't know if unsafe is strong enough. It fails to communicate to users that this relies on implementation details of the Swift runtime (`dlsym`ing symbols) and unsupported language features (`@convention(thin)`) as well as a bunch of assumptions about the layout of the runtime's non-public types.

It's not just that calling this might do weird junk, it's that calling it might explode in new Swift releases, and it would very much be our fault.

I'm inclined to say that we also want to underscore this, and expand on the doc comment to say that this function cannot guarantee that it will safely work on newer Swift compilers or any non-stable Swift release.

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
```diff
@@ -0,0 +1,123 @@
...
+
+        typealias ConcurrencyEnqueueGlobalHook = @convention(thin) (
+            UnownedJob, @convention(thin) (UnownedJob) -> Void
+        ) -> Void
+
+        guard
+            _haveWeTakenOverTheConcurrencyPool.compareExchange(
+                expected: false,
+                desired: true,
+                ordering: .relaxed
```
Lukasa, line 55:
```suggestion
                ordering: .acquiringAndReleasing
```

This is probably safe without using an acquiring load here due to the `dlsym` call that follows it, but semantically this should be an acquiring load. Logically then it also has to be a releasing store because we're synchronising with ourselves!

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
```diff
@@ -0,0 +1,123 @@
...
+            return enqueueOnNIOPtr.baseAddress!.withMemoryRebound(
+                to: UnsafeRawPointer.self,
+                capacity: 1
+            ) { enqueueOnNIOPtr in
+                // Unsafe 4: We just pretend that we're the only ones in the world pulling this trick (or at least
+                // that the others also use a `compareExchange`)...
+                guard concurrencyEnqueueGlobalHookAtomic.compareExchange(
+                    expected: nil,
+                    desired: enqueueOnNIOPtr.pointee,
+                    ordering: .relaxed
```
Lukasa, line 110: Similarly this should probably be a `releasing` ordering. Again, `.enqueue` should be an appropriate compiler barrier so arguably unnecessary, but I'd rather we did this right.

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
weissi: @Lukasa I think you're mixing up the two variables. This is just the guard, no other memory reads/writes are ordered with this guy. This is correct not because of the `dlsym` but because of the `guard`.

You need non-`.relaxed` if you care about the order of _other_ memory reads/writes but we don't

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
weissi: What about leaving it `unsafe*` (no underscore) but making it only perform work on `#if compiler(>=5.9) && compiler(<5.10)`. Then we can gradually relax it as new Swift versions get released? And maybe at some point we can stop this continuous version-upping process once Swift implements enough to do this properly.

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
weissi: Actually, we don't even need the `guard`. So yeah, any order of memory operations is completely fine here.

Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift
FranzBusch: FWIW, I don't like the compiler guards because it will continuously break and we need to play catch up with the compiler version. IMO `_` or `SPI` works for me here and indicates that this is really not something we want people to actively do.

weissi: But nothing breaks here, it'll just return `false` and be slower. That's better than having `_badStuff` that people use anyway which then leads to crashes.

IMHO we can have either of two arguments:

1. Future Swift versions will likely break us here -> we should disable on future Swifts
2. Future Swift versions will likely not break us here -> `unsafe*`'s enough

I don't think offering useful functionality behind a `_` that may crash is ideal. If we're not sure if something will continue to work we should disable it for anything but known-good compilers, no? We'd then just force the `#if compiler(...)` into our users' codebases.

weissi: I'm happy with that.

weissi: @Lukasa implemented

FranzBusch (approved):
LGTM!

Issue #2565, opened 2023-10-21
NonBlockingFileIO has no async APIs 


adam-fowler:
It appears that the best option for file access is still `NonBlockingFileIO`. Currently to use `NonBlockingFileIO` in a Swift concurrency context you are required to use the `EventLoop` based API and `EventLoopFuture.get`. This is unsatisfactory because 

1) You need an EventLoop to call any of the functions
2) Waiting on a process running on the NIOThreadPool requires two hops: one to the EventLoop and then one to your async task.

Are there any plans to produce async `NonBlockingFileIO` APIs?

Lukasa:
We certainly plan to produce _a_ solution in this area. However, in our ideal world this wouldn't be NIO-based at all, as the solution should be more general purpose than that.

adam-fowler:
Would you be adverse to me producing a temporary solution that still uses NIOThreadPool but skips the EventLoopPromise and uses a Continuation instead?

This is basically the last barrier to removing EventLoops from Soto.

Lukasa:
I'd be happy for you to do that work.

JawadHF:
> We certainly plan to produce _a_ solution in this area. However, in our ideal world this wouldn't be NIO-based at all, as the solution should be more general purpose than that.

Just to understand, would that be something like [Swift System](https://github.com/apple/swift-system) ?

PR #2566, opened 2023-10-23
Add async version of NIOThreadPool.runIfActive

adam-fowler:
Add async version of NIOThreadPool.runIfActive

### Motivation:

Currently when using NIOThreadPool in a Swift concurrency context you still need an EventLoop and the call will make two hops, one to the EventLoop and then one back to your Task.

### Modifications:

Added `NIOThreadPool.runIfActive(_) async throws`

### Result:

You don't need an EventLoop to use `NIOThreadPool`

This is the first stage of completing https://github.com/apple/swift-nio/issues/2565

### DIFF
--- a/Sources/NIOPosix/NIOThreadPool.swift
+++ b/Sources/NIOPosix/NIOThreadPool.swift
@@ -292,6 +292,48 @@ extension NIOThreadPool {
     }
 }
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOThreadPool {
+    #if swift(>=5.7)
+    /// Runs the submitted closure if the thread pool is still active, otherwise throw an error.
+    /// The closure will be run on the thread pool so can do blocking work.
+    ///
+    /// - parameters:
+    ///     - body: The closure which performs some blocking work to be done on the thread pool.
+    /// - returns: result of the passed closure.
+    @preconcurrency
+    public func runIfActive<T>(_ body: @escaping @Sendable () throws -> T) async throws -> T {
+        try await self._runIfActive(body)
+    }
+    #else
+    /// Runs the submitted closure if the thread pool is still active, otherwise throw an error.
+    /// The closure will be run on the thread pool so can do blocking work.
+    ///
+    /// - parameters:
+    ///     - body: The closure which performs some blocking work to be done on the thread pool.
+    /// - returns: result of the passed closure.
+    public func runIfActive<T>(_ body: @escaping () throws -> T) async throws -> T {
+        try await self._runIfActive(body)
+    }
+    #endif
+    
+    private func _runIfActive<T>(_ body: @escaping () throws -> T) async throws -> T {
+        try await withCheckedThrowingContinuation { (cont: CheckedContinuation<T, Error>) in
+            self.submit { shouldRun in
+                guard case shouldRun = NIOThreadPool.WorkItemState.active else {
+                    cont.resume(throwing: NIOThreadPoolError.ThreadPoolInactive())
+                    return
+                }
+                do {
+                    try cont.resume(returning: body())
+                } catch {
+                    cont.resume(throwing: error)
+                }
+            }
+        }
+    }
+}
+
 extension NIOThreadPool {
     @preconcurrency
     public func shutdownGracefully(_ callback: @escaping @Sendable (Error?) -> Void) {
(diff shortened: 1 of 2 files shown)


Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -292,6 +292,48 @@ extension NIOThreadPool {
     }
 }
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOThreadPool {
+    #if swift(>=5.7)
```
Lukasa, line 297: We no longer support older versions of Swift than 5.7, so we can drop this conditional compilation step.

Tests/NIOPosixTests/NIOThreadPoolTest.swift
```diff
@@ -110,6 +111,21 @@ class NIOThreadPoolTest: XCTestCase {
...
+        pool.start()
+        do {
+            let hitCount = ManagedAtomic(false)
+            try await pool.runIfActive {
+                hitCount.store(true, ordering: .relaxed)
+            }
+            XCTAssertEqual(hitCount.load(ordering: .relaxed), true)
+        } catch {}
+        try await pool.shutdownGracefully()
+    }
```
Lukasa, line 127: Can we test the two other paths through the code as well?

Lukasa (changes requested):
Thanks, a couple of quick notes here.

Tests/NIOPosixTests/NIOThreadPoolTest.swift
adam-fowler: Do you mean throwing errors from the closure, and the thread pool not being active

Lukasa: I do, yeah. 😄

Sources/NIOPosix/NIOThreadPool.swift
adam-fowler: ok, I've removed the older version

Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -292,6 +292,36 @@ extension NIOThreadPool {
...
+    ///
+    /// - parameters:
+    ///     - body: The closure which performs some blocking work to be done on the thread pool.
+    /// - returns: result of the passed closure.
+    @preconcurrency
+    public func runIfActive<T>(_ body: @escaping @Sendable () throws -> T) async throws -> T {
+        try await self._runIfActive(body)
+    }
+    
+    private func _runIfActive<T>(_ body: @escaping () throws -> T) async throws -> T {
```
Lukasa, line 308: I don't think we need this inner function now, and so we can avoid dropping the `@Sendable` annotation here.

Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -292,6 +292,36 @@ extension NIOThreadPool {
...
 
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NIOThreadPool {
+    /// Runs the submitted closure if the thread pool is still active, otherwise throw an error.
+    /// The closure will be run on the thread pool so can do blocking work.
+    ///
+    /// - parameters:
+    ///     - body: The closure which performs some blocking work to be done on the thread pool.
+    /// - returns: result of the passed closure.
+    @preconcurrency
```
Lukasa, line 303: We can drop `preconcurrency`, this method doesn't predate concurrency.

Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -292,6 +292,36 @@ extension NIOThreadPool {
...
+    public func runIfActive<T>(_ body: @escaping @Sendable () throws -> T) async throws -> T {
```
Lukasa, line 304: `T: Sendable`

Lukasa (commented):
Generally looking good, a few quick notes on the API.

Sources/NIOPosix/NIOThreadPool.swift
adam-fowler: done

Sources/NIOPosix/NIOThreadPool.swift
adam-fowler: done

Sources/NIOPosix/NIOThreadPool.swift
adam-fowler: done.

Lukasa (approved):
Nice one, thanks!

PR #2569, opened 2023-10-23
Mark retroactive conformances appropriately.

Lukasa:
Motivation:

In nightly Swift, we now need to mark retroactive conformances when they are intentional. These conformances are safe for us, so we can safely suppress the warnings.

Modifications:

- Mark NIOFoundationCompat retroactive conformances.

Result:

Nightly builds work again


dnadoba (approved):

dnadoba (approved):

weissi (approved):
thank you!

PR #2570, opened 2023-10-23
WIP: HTTPDecoder: Reenable Parsing After Failed Upgrade

jshier:
Reenable parsing in `HTTPDecoder` after an upgrade is attempted but failed by the response.

### Motivation:

Currently, `HTTPDecoder` sets `stopParsing = true` whenever a message is parsed and an update is detected (`isUpgrade == true`). Unfortunately, this means that if the upgrade wasn't successful, the decoder silently blocks any attempt to reuse the connection, essentially hanging the client until they timeout.

### Modifications:

I've attempted to generalize `HTTPDecoder`'s conformance to `WriteObservingByteToMessageDecoder` so that it can reenable parsing based on the outgoing `HTTPServerResponsePart`.

### Result:

A properly reusable decoder.

---

WIP:

I need a few things to finish this PR.

- Proper way to generalize the `write(data:)` implementation. Right now I just check the decoder's full metatype. What's the preferred pattern here?
- Proper condition for reenabling parsing. Right now I ensure that the connection will be kept alive (if it's being killed who cares), that we attempted an upgrade (just realized that `isUpgrade` is reset as soon as the message is decoded, so I need a different condition there), and that the response is NOT an upgrade.
- Proper test. I see the `HTTPDecoderTest` but it's not clear what kind of test would be best here, or how to simulate this scenario exactly.
- Proper way to integrate this local branch with my Vapor server to verify it fixes my observed issue. Simply adding SwiftNIO as a dependency doesn't replace Vapor's expected dependency on it. Do I need to fork Vapor locally too?

### DIFF
--- a/Sources/NIOHTTP1/HTTPDecoder.swift
+++ b/Sources/NIOHTTP1/HTTPDecoder.swift
@@ -473,7 +473,7 @@ public typealias HTTPResponseDecoder = HTTPDecoder<HTTPClientResponsePart, HTTPC
 ///
 /// While the `HTTPRequestDecoder` does not currently have a specific ordering requirement in the
 /// `ChannelPipeline` (unlike `HTTPResponseDecoder`), it is possible that it will develop one. For
-/// that reason, applications should try to ensure that the `HTTPRequestDecoder` *later* in the
+/// that reason, applications should try to ensure that the `HTTPRequestDecoder` is *later* in the
 /// `ChannelPipeline` than the `HTTPResponseEncoder`.
 ///
 /// Rather than set this up manually, consider using `ChannelPipeline.configureHTTPServerPipeline`.
@@ -484,12 +484,18 @@ public enum HTTPDecoderKind: Sendable {
     case response
 }
 
-extension HTTPDecoder: WriteObservingByteToMessageDecoder where In == HTTPClientResponsePart, Out == HTTPClientRequestPart {
+extension HTTPDecoder: WriteObservingByteToMessageDecoder {
     public typealias OutboundIn = Out
 
-    public func write(data: HTTPClientRequestPart) {
-        if case .head(let head) = data {
+    public func write(data: Out) {
+        if Self.self == HTTPResponseDecoder.self,
+            case .head(let head) = (data as? HTTPClientRequestPart) {
             self.parser.requestHeads.append(head)
+        } else if Self.self == HTTPRequestDecoder.self, 
+                    case let .head(head) = data as? HTTPServerResponsePart {
+            if head.isKeepAlive, self.isUpgrade == true, head.status != .switchingProtocols {
+                self.stopParsing = false
+            }
         }
     }
 }
@@ -567,7 +573,6 @@ public final class HTTPDecoder<In, Out>: ByteToMessageDecoder, HTTPDecoderDelega
         case .response:
             self.context!.fireChannelRead(NIOAny(HTTPClientResponsePart.body(self.buffer!.readSlice(length: bytes.count)!)))
         }
-
     }
 
     func didReceiveHeaderName(_ bytes: UnsafeRawBufferPointer) throws {


jshier:
I believe I've verified that my fix works (local server can be used to pass the Alamofire test that previously hung), but I still need guidance around the proper conditions and unit test.

Lukasa:
> Proper way to generalize the write(data:) implementation. Right now I just check the decoder's full metatype. What's the preferred pattern here?

You can use `HTTPDecoder.kind`, which tells you which kind of decoder you are.

> Proper condition for reenabling parsing. Right now I ensure that the connection will be kept alive (if it's being killed who cares), that the response is NOT an upgrade, and that it has stopped parsing.

I think for now this is appropriate.

> Proper test. I see the HTTPDecoderTest but it's not clear what kind of test would be best here, or how to simulate this scenario exactly.

I think you want a few scenarios:

1. Send an upgrade request, do not accept the upgrade, and then send another request. Confirm both get through.
2. Pipeline a pair of requests. This is generally unsound and we don't really support it, but if the upgrade fails then we should resume parsing the second one. (I have an inkling this test will fail, it's possible that we need to trigger ourselves to start parsing again in this case which will be a fun wrinkle).

To achieve "do not accept the upgrade" you don't actually need an encoder, you can just write the response head into the pipeline. `EmbeddedChannel` doesn't care what data you pass through it.

weissi:
@swift-nio-bot test performance please

weissi:
@swift-nio-bot performance test please

Lukasa:
@swift-server-bot test performance please

Lukasa:

Lukasa:

weissi:
@Lukasa / @jshier  ugh, this is rough

```
no-net_http1_1k_reqs_1_conn | 0.013450951 | 0.011678896 | previous | 15%
```

15% slower with this change :|. And that includes client (unchanged) & servers (changed) meaning that servers will likely be affected even more gravely.

Lukasa:
The change as written does runtime type checks. My review feedback above would replace it with a jump based on an enum value, which should perform a lot better.

weissi:
> The change as written does runtime type checks. My review feedback above would replace it with a jump based on an enum value, which should perform a lot better.

Missed that, sorry. That's great!

jshier:
I've pushed the kind check up, feel free to test that, but I think there will still be an impact due to the `data` cast. I'll try to finish up the local tests soon.

weissi:

weissi:

jshier:
I've added two tests, one for successful upgrade and one for failed upgrade.

@Lukasa What does a pipelined request pair look like? Do I just shove both through in one buffer?

Lukasa:
> What does a pipelined request pair look like? Do I just shove both through in one buffer?

Yeah, that's the easiest way to do it.

jshier:
@Lukasa I can write a pipelined test for the current behavior just fine, as parsing is stopped as soon as the first request is parsed, but I'm not clear what behavior I should be seeing in the unsuccessful upgrade case. As far as I can tell, the data just disappears, and the channel is empty. I'm guessing that with parsing stopped, the additional data is just lost, and the unsuccessful upgrade turns parsing back on for a decoder that doesn't receive any additional data. To support the pipeline case I think I'd need to buffer any additional requests until we can tell whether the upgrade was successful, or some other change in behavior.

jshier:
@Lukasa I've investigated how to restart parsing in `HTTPDecoder` and there doesn't seem to be any way to do so simply. `cannel.writeInbound` triggers a codepath the decoder can't reach, and since the decoder `nil`s out the context when reading finishes, there's no way to start reading again internally. Any ideas what a proper fix might be here? We could make the `nil`ing behavior dependent on the outgoing response, but I don't know if that's appropriate here. Any ideas?

PR #2576, opened 2023-10-26
Async apis for NonBlockingFileIO

adam-fowler:
Provide async versions of most NonBlockingFileIO APIs. This PR provides versions of all functions apart from the `readChunked` function which should probably be replaced by an AsyncSequence.

Some of the code running on the NIOThreadPool that would have been duplicated has been broken out into separate functions so it can be used by both ELF and async functions.

I've also provided async versions of the relevant tests


adam-fowler:
I think it is complaining because I've reduced the number of allocations for the readChunked test.

adam-fowler:
I just saw #2578 which may clash with this PR as. I am passing `NIOFileHandle` across concurrency contexts. How should I proceed? I have used an `@unchecked Sendable` type to pass a `FileRegion` from the `openFile` operation. In this situation is it considered ok?

FranzBusch:
`NIOFileHandle` can definitely not become `Sendable` since they must be owned by a single isolation region at any given time. However, with the new [region based isolation pitch](https://forums.swift.org/t/pitch-region-based-isolation/67888) a lot of the cases where you currently want `NIOFileHandle` to be `Sendable` should be solved.

adam-fowler:
Oh I know it shouldn't be Sendable. But what about the situation this PR covers where it is created on a NIOThreadPool and immediately passed back to a Task.

dnadoba:
@adam-fowler You will need to use [`UnsafeTransfer`](https://github.com/apple/swift-nio/blob/9497e442486aab515c8486ef8153a506f93a5032/Sources/NIOCore/NIOSendable.swift#L27) in your implementation for this until we get region based isolation. You will need to copy the implementation over to NIOPosix as it is not public (and we don't want it to be public).

adam-fowler:
Any chance of a review here?

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +726,329 @@ public struct NIODirectoryEntry: Hashable {
...
+    ///
+    /// This function will return (a future) of the `NIOFileHandle` associated with the file opened and a `FileRegion`
+    /// comprising of the whole file. The caller must close the returned `NIOFileHandle` when it's no longer needed.
+    ///
+    /// - note: The reason this returns the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as the querying of its size are blocking.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be opened for reading.
+    /// - returns: the `NIOFileHandle` and the `FileRegion` comprising the whole file.
+    public func openFile(path: String) async throws -> (NIOFileHandle, FileRegion) {
```
FranzBusch, line 901: We should probably spell this as `func withOpenFile<Result>(path: String, _ body: (NIOFileHandle, FileRegion) async throws -> Result) async throws -> Result`. This allows us to keep the file handle open for the duration of the `with` body and we can close at the end providing the user an even simpler API where they don't have to take care of the closing explicitly.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +726,329 @@ public struct NIODirectoryEntry: Hashable {
...
+    ///
+    /// This function will return (a future) of the `NIOFileHandle` associated with the file opened.
+    /// The caller must close the returned `NIOFileHandle` when it's no longer needed.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be opened for writing.
+    ///     - mode: File access mode.
+    ///     - flags: Additional POSIX flags.
+    /// - returns: NIOFileHandle`.
+    public func openFile(path: String, mode: NIOFileHandle.Mode, flags: NIOFileHandle.Flags = .default) async throws -> NIOFileHandle {
```
FranzBusch, line 925: Same here

FranzBusch (commented):
In general this looks good to me. I would like @glbrntt to also take a look since he spent some time with file systems recently.

Sources/NIOPosix/NonBlockingFileIO.swift
glbrntt: We also need to be careful to make sure we close _even if_ the task has been cancelled.

glbrntt: Also while `NIOFileHandle` isn't `Sendable` I'm wondering if we should at least make it thread safe. At the moment a double close could result in closing someone else's fd, we can and I think should, avoid doing that.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +726,329 @@ public struct NIODirectoryEntry: Hashable {
     }
 }
 #endif
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+extension NonBlockingFileIO {
+    /// Read a `FileRegion` in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// The returned `ByteBuffer` will not have less than `fileRegion.readableBytes` unless we hit end-of-file in which
+    /// case the `ByteBuffer` will contain the bytes available to read.
```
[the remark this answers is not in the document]
adam-fowler: OK this is just copied from the ELF version. I can fix up both though

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: `withOpenFile` is obviously different from the ELF version of this function. But I'm open to making this change.

glbrntt: True, I think it's okay to diverge here given how easy it is to not `close()` given the lack of an `async` `defer`.

adam-fowler: I was just caught out by this right now 🤣

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +726,329 @@ public struct NIODirectoryEntry: Hashable {
...
+                         allocator: allocator)
+    }
+
+    /// Read `byteCount` bytes from `fileHandle` in `NonBlockingFileIO`'s private thread pool which is separate from any `EventLoop` thread.
+    ///
+    /// The returned `ByteBuffer` will not have less than `byteCount` bytes unless we hit end-of-file in which
+    /// case the `ByteBuffer` will contain the bytes available to read.
+    ///
+    /// - note: Only use this function for small enough `byteCount`s as it will need to allocate enough memory to hold `byteCount` bytes.
+    /// - note: `read(fileRegion:allocator:eventLoop:)` should be preferred as it uses `FileRegion` object instead of
```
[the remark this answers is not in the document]
adam-fowler: You mean the double back tick?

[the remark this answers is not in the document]
glbrntt: yup!

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: done

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: done

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,346 @@ public struct NIODirectoryEntry: Hashable {
...
+            do {
+                let fr = try FileRegion(fileHandle: fh)
+                return UnsafeTransfer((handle: fh, region: fr))
+            } catch {
+                _ = try? fh.close()
+                throw error
+            }
+        }
+        defer {
+            try? file.wrappedValue.handle.close()
```
adam-fowler, line 923: Currently this is a `try?`, would you prefer to throw an error from the close?

glbrntt: We should throw the error; there's not a lot the user can do about `EBADF` and presumably we handle `EINTR` but it may be `EIO` which the caller may want to know about.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -884,54 +894,67 @@ extension NonBlockingFileIO {
...
+    /// This function will return the result of the operation.
     ///
     /// - parameters:
     ///     - path: The path of the file to be opened for writing.
     ///     - mode: File access mode.
     ///     - flags: Additional POSIX flags.
     /// - returns: NIOFileHandle`.
-    public func openFile(path: String, mode: NIOFileHandle.Mode, flags: NIOFileHandle.Flags = .default) async throws -> NIOFileHandle {
-        let result = try await self.threadPool.runIfActive {
+    public func withOpenFile<Result>(
```
glbrntt, line 940: Bikeshedding... I wonder if we need the "Open" here? It seems somewhat redundant.

glbrntt (commented):
This looks great, left couple of comments.

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: `withFile` doesn't make much sense to me, maybe `withFileHandle`? And should we keep the other function named `withOpenFile`

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,356 @@ public struct NIODirectoryEntry: Hashable {
     }
 }
 #endif
+
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
```
FranzBusch, line 734: Super nit: But can we move those to the individual methods I always feel like it is easier to discover the availability then

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,356 @@ public struct NIODirectoryEntry: Hashable {
...
+extension NonBlockingFileIO {
+    /// Read a `FileRegion` in ``NonBlockingFileIO``'s private thread pool which is separate from Task pool.
```
FranzBusch, line 736: I find the `which is separate from Task pool` part to be confusing in all of those doc comments. Can we just drop this?

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,356 @@ public struct NIODirectoryEntry: Hashable {
...
+    /// The open file operation runs on a private thread pool which is separate from Task pool.
+    ///    
+    /// - note: The reason this provides the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as 
+    /// the querying of its size are blocking.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be opened for reading.
+    ///     - body: operation to run with file handle and region
+    /// - returns: return value of operation
+    public func withOpenFile<Result>(path: String, _ body: (NIOFileHandle, FileRegion) async throws -> Result) async throws -> Result {
```
FranzBusch, line 912:
```suggestion
    public func withOpenFile<Result>(path: String, _ body: (_ fileHandle: NIOFileHandle, _ fileRegion: FileRegion) async throws -> Result) async throws -> Result {
```

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,356 @@ public struct NIODirectoryEntry: Hashable {
...
+                _ = try? fh.close()
+                throw error
+            }
+        }
+        do {
+            let result = try await body(file.wrappedValue.handle, file.wrappedValue.region)
+            try file.wrappedValue.handle.close()
+            return result
+        } catch {
+            try file.wrappedValue.handle.close()
```
FranzBusch, line 928: Do we want to throw the close error in this case or the bodies error? Also this might try to double close if the previous close threw.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,356 @@ public struct NIODirectoryEntry: Hashable {
...
+        mode: NIOFileHandle.Mode, 
+        flags: NIOFileHandle.Flags = .default, 
+        _ body: (NIOFileHandle) async throws -> Result
+    ) async throws -> Result {
+        let fileHandle = try await self.threadPool.runIfActive {
+            return try UnsafeTransfer(NIOFileHandle(path: path, mode: mode, flags: flags))
+        }
+        do {
+            let result = try await body(fileHandle.wrappedValue)
+            try fileHandle.wrappedValue.close()
```
FranzBusch, line 955: Same here

FranzBusch (commented):
Some last comments but looks really good

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: Yeah I'll fix the double throw. 

I don't know which is better to report the error from the body or the file close. The user might be using errors as a way to report state from the body and then will never get an error from the file close.

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: Yeah it was originally about running outside of EventLoops. I'm not happy with the wording either.

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: Fixed double close. Left throwing error though

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -320,40 +320,49 @@ public struct NonBlockingFileIO: Sendable {
...
-                                                  size: byteCount - bytesRead)
-                        }
-                    }
-                    switch res {
-                    case .processed(let n):
-                        assert(n >= 0, "read claims to have read a negative number of bytes \(n)")
-                        return n
-                    case .wouldBlock:
-                        throw Error.descriptorSetToNonBlocking
+            try readSync(fileHandle: fileHandle, fromOffset: fromOffset, byteCount: byteCount, allocator: allocator)
```
glbrntt, line 324:
```suggestion
            try self.readSync(fileHandle: fileHandle, fromOffset: fromOffset, byteCount: byteCount, allocator: allocator)
```

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,379 @@ public struct NIODirectoryEntry: Hashable {
...
+            try self.writeSync(fileHandle: fileHandle, byteCount: byteCount, toOffset: toOffset, buffer: buffer)
+        }
+    }
+
+    /// Open file at `path` on a private thread pool, run an operation given the file handle and region and then close the file handle.
+    ///
+    /// The open file operation runs on a private thread pool.
+    ///    
+    /// - note: The reason this provides the `NIOFileHandle` and the `FileRegion` is that both the opening of a file as well as 
+    /// the querying of its size are blocking.
```
glbrntt, line 914: Do we actually need to provide both in the closure? `FileRegion` has a `fileHandle` property which is `public` so it's available if users need it. `withFileRegion` would be a more direct analogue to `withFileHandle` than `withOpenFile` as well.

Sources/NIOPosix/NonBlockingFileIO.swift
```diff
@@ -712,3 +730,379 @@ public struct NIODirectoryEntry: Hashable {
...
+    ///
+    /// The open file operation runs on a private thread pool.
+    ///    
+    /// This function will return the result of the operation.
+    ///
+    /// - parameters:
+    ///     - path: The path of the file to be opened for writing.
+    ///     - mode: File access mode.
+    ///     - flags: Additional POSIX flags.
+    /// - returns: NIOFileHandle`.
```
glbrntt, line 956: The `returns:` docs need updating

glbrntt (changes requested):
This is looks great, I left a couple of nits though.

Sources/NIOPosix/NonBlockingFileIO.swift
adam-fowler: That's a nice idea. This is now implemented

dnadoba (changes requested):
We are planning to cut a release in the next couple hours. Let's wait til we merge this until then.

dnadoba:
If you feel fancy you could also change the `byteCount` parameter types from `Int` to `Int64` and temporarily throw an error if the value doesn't fit exactly into an `Int` (i.e. [`Int(exactly: byteCount)`](https://developer.apple.com/documentation/swift/int/init(exactly:)-b1dy) returns nil` until we convert the underlying APIs to also take `Int64`. More infos in this issue comment: https://github.com/apple/swift-nio/issues/2591#issuecomment-1810267230
It would be quite easy as we don't need to deprecate any API as the async API is new. 

Note that I'm not suggesting to fix the 4GB Buffer overflow issue, just the type change from `Int` to `Int64`.
This can also be done in a separate PR and doesn't block this PR from being merged.

dnadoba (approved):
[NIO 2.62.0 released ](https://github.com/apple/swift-nio/releases/tag/2.62.0)

adam-fowler:
> If you feel fancy you could also change the `byteCount` parameter types from `Int` to `Int64` and temporarily throw an error if the value doesn't fit exactly into an `Int` (i.e. [`Int(exactly: byteCount)`](https://developer.apple.com/documentation/swift/int/init(exactly:)-b1dy) returns nil) until we convert the underlying APIs to also take `Int64`. More infos in this issue comment: [#2591 (comment)](https://github.com/apple/swift-nio/issues/2591#issuecomment-1810267230)
> It would be quite easy as we don't need to deprecate any API as the async API is new.

@dnadoba I can add this. One minor issue though `NonBlockingFileIO.Error` is an `enum` which I can't extend. I'd have to add another error type. It might be better to deal with this in a separate PR which doesn't require the error because 64 bit read sizes will be supported.

dnadoba:
Sure, we can just use a specific error that we later deprecate as this should only be a temporary issue I guess.

PR #2578, opened 2023-10-27
Add explicit `Sendable` unavailability and add missing `Sendable` conformances

dnadoba:
Found with `-require-explicit-sendable` compiler flag. We may want to enable this flag in CI but it currently warns also for internal types that are @usableFromInline but we could mark them explicitly as well if we wanted.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -147,7 +147,7 @@ public struct NIOAsyncWriter<
     /// This struct contains two properties:
     /// 1. The ``sink`` which should be retained by the consumer and is used to set the writability.
     /// 2. The ``writer`` which is the actual ``NIOAsyncWriter`` and should be passed to the producer.
-    public struct NewWriter {
+    public struct NewWriter: @unchecked Sendable {
         /// The ``sink`` which **MUST** be retained by the consumer and is used to set the writability.
         public let sink: Sink
         /// The ``writer`` which is the actual ``NIOAsyncWriter`` and should be passed to the producer.
@@ -375,6 +375,9 @@ extension NIOAsyncWriter {
         }
     }
 }
+
+@available(*, unavailable)
+extension NIOAsyncWriter.Sink: Sendable { }
 
 @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
 extension NIOAsyncWriter {
--- a/Sources/NIOCore/BSDSocketAPI.swift
+++ b/Sources/NIOCore/BSDSocketAPI.swift
@@ -93,7 +93,7 @@ let SO_TIMESTAMP = CNIOLinux_SO_TIMESTAMP
 let SO_RCVTIMEO = CNIOLinux_SO_RCVTIMEO
 #endif
 
-public enum NIOBSDSocket {
+public enum NIOBSDSocket: Sendable{
 #if os(Windows)
     public typealias Handle = SOCKET
 #else
--- a/Sources/NIOCore/Channel.swift
+++ b/Sources/NIOCore/Channel.swift
@@ -378,7 +378,7 @@ extension ChannelError: Equatable { }
 /// The removal of a `ChannelHandler` using `ChannelPipeline.removeHandler` has been attempted more than once.
 public struct NIOAttemptedToRemoveHandlerMultipleTimesError: Error {}
 
-public enum DatagramChannelError {
+public enum DatagramChannelError: Sendable {
     public struct WriteOnUnconnectedSocketWithoutAddress: Error {
         public init() {}
     }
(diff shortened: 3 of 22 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -147,7 +147,7 @@ public struct NIOAsyncWriter<
     /// This struct contains two properties:
     /// 1. The ``sink`` which should be retained by the consumer and is used to set the writability.
     /// 2. The ``writer`` which is the actual ``NIOAsyncWriter`` and should be passed to the producer.
-    public struct NewWriter {
+    public struct NewWriter: @unchecked Sendable {
```
FranzBusch, line 150: Unsure about that one the `Sink` is not `Sendable` right now because we expect it to be only held in one isolation domain at any given time even though it is fully thread-safe. We certainly can't make this type `unchecked Sendable` then. So we either make the `Sink` `Sendable` as well or don't make this one `Sendable`

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -93,7 +93,7 @@ let SO_TIMESTAMP = CNIOLinux_SO_TIMESTAMP
 let SO_RCVTIMEO = CNIOLinux_SO_RCVTIMEO
 #endif
 
-public enum NIOBSDSocket {
+public enum NIOBSDSocket: Sendable{
```
FranzBusch, line 96: I don't think we can make a socket `Sendable` it is the same as `FileHandle`

Sources/NIOCore/Channel.swift
```diff
@@ -378,7 +378,7 @@ extension ChannelError: Equatable { }
 /// The removal of a `ChannelHandler` using `ChannelPipeline.removeHandler` has been attempted more than once.
 public struct NIOAttemptedToRemoveHandlerMultipleTimesError: Error {}
 
-public enum DatagramChannelError {
+public enum DatagramChannelError: Sendable {
```
FranzBusch, line 381: This is just a namespace right? We probably shouldn't make it `Sendable` right now

Sources/NIOCore/ChannelOption.swift
```diff
@@ -64,7 +64,7 @@ public typealias ConnectTimeoutOption = ChannelOptions.Types.ConnectTimeoutOptio
 public typealias AllowRemoteHalfClosureOption = ChannelOptions.Types.AllowRemoteHalfClosureOption
 
 extension ChannelOptions {
-    public enum Types {
+    public enum Types: Sendable {
```
FranzBusch, line 67: Same here. This is just a namespace right?

FranzBusch (changes requested):
Overall, before we mark more things as explicitly not `Sendable` we should check how that interacts with region based isolation. Can an explicitly not `Sendable` value still be send across regions?

Sources/NIOCore/BSDSocketAPI.swift
Lukasa: This is just a namespace.

FranzBusch: Ah good catch. Then my comments from below apply. Should we really make namespaces `Sendable`?

Lukasa: I can't see any reason to do so.

dnadoba:
@FranzBusch eventually everything is either Sendable or not Sendable. We have this grey zone right now where the  diagnostics severity is influenced by explicit or implicit marking but in Swift 6 mode this should no longer be the case.

Sources/NIOCore/Channel.swift
dnadoba: Okay, I will mark it then as non-sendable. I don't think it really matters though.

FranzBusch:
> @FranzBusch eventually everything is either Sendable or not Sendable.

I am unsure if that's true. In my head there are three distinct constraints `Sendable` not-`Sendable` and unavailable `Sendable`. The latter has some special treatment in the compiler right now and I am wondering if the region based isolation diagnostic also treats it differently or not. If it doesn't treat it differently then I am fine with marking everything as either `Sendable` or unavailable `Sendable`

Lukasa:
> @FranzBusch eventually everything is either Sendable or not Sendable.

I think this is fundamentally not true. Uninhabited types are neither Sendable nor not Sendable. They can certainly _conform_, but as a practical semantic matter they aren't Sendable or not Sendable because no values of them exist.

Similarly, `Never` conforms to `Sendable`, but you cannot send a `Never`.

dnadoba:
> Uninhabited types are neither Sendable nor not Sendable
I can't follow. You say yourself that `Never` is `Sendable` and `Never` is an uninhabited type which contradicts your statement above.

The only way a Type can "maybe" conform to a Protocol is through retroactive conformance. However, this isn't relevant as Sendable is a marker protocol and is checked at compile time only. Generics are another possible case where the concrete type might conform to Sendable but as Sendable is only checked at compile time this doesn't matter either.

Lukasa:
I'm not saying uninhabited types can't conform to `Sendable`, I'm saying they can't _be_ `Sendable`. That is, they can never satisfy the semantic requirement of the protocol. There are no values of uninhabited types, so you can never Send one.

Similarly, `Never` conforms to `Equatable` but you can never equate them.

This is why it's simply not worth arguing about whether they conform or not: their conformance is meaningless, and no correct code could possibly rely on it.

Lukasa:
We’re working through fixing this now across the whole package.

Issue #2574, opened 2023-10-25
Version 2.60.0 causing app to crash when launched on iOS 15.5 simulator


MatthewTHFisher:
### Expected behavior

When running our iOS app on an iOS 15.5 simulator with the swift-nio package version `2.60.0`, our app launches and runs as expected.

### Actual behavior

When running on a iOS 15.5 simulator while using swift-nio 2.60.0, the app launches and then immediately crashes with `Thread 1: signal SIGABRT`

When run on a iOS 16.4 simulator the app runs as expected.

### Workaround

Downgrade to swift-nio version `2.59.0` and the app runs fine on iOS 15.5 simulators.

### System & version information

OS System - MacOS 13.6 (22G120)
Processer type - Attempted on both Intel and Apple-silicon
Xcode - 14.3.1 (14E300c)
Xcode simulators - iPhone 12 | iOS 15.5 & iOS 16.4

dnadoba:
Could you provide us with a crash report? [Here is some information how to gather crash reports](https://developer.apple.com/documentation/xcode/acquiring-crash-reports-and-diagnostic-logs).

dnadoba:
hm... this doesn't include a backtrace/stacktrace. Do have this as well? The crashing thread is the most important thing but from the other running threads would be good as well.

weissi:
@dnadoba this has a stack trace

```
$ osatool legacy /tmp/crash.ips 

Incident Identifier: 1086025F-7F64-42EC-8CAF-D9724CBAB527
CrashReporter Key:   40D309A8-4B1A-C8AB-A1EF-CB71C1CC050A
Hardware Model:      MacBookPro14,3
Process:             APP_NAME [82754]
Path:                /Users/USER/Library/Developer/CoreSimulator/Devices/2EF156A4-64B6-40FC-AC71-03A0E58CE698/data/Containers/Bundle/Application/0A73DFE0-D246-4BB8-AFC9-1D6ED0F5AA61/APP_NAME.app/APP_NAME
Identifier:          com.APP_NAME.app
Version:             1.0 (1)
Code Type:           X86-64 (Native)
Role:                Foreground
Parent Process:      Exited process [82767]
Coalition:           com.apple.CoreSimulator.SimDevice.2EF156A4-64B6-40FC-AC71-03A0E58CE698 [3117]
Responsible Process: SimulatorTrampoline [6243]

Date/Time:           2023-10-25 13:32:11.6325 +0100
[...]

PR #2579, opened 2023-10-27
Back out new typed HTTP protocol upgrader

FranzBusch:
# Motivation
We got reports in https://github.com/apple/swift-nio/issues/2574 that our new typed HTTP upgrader are hitting a Swift compiler bug which manifests in a runtime crash on older iOS/macOS/etc.

# Modification
This PR backs out the new typed HTTP protocol upgrader APIs so that we can unblock our users until the Swift compiler bug is fixed.

# Result
No more crashes for our users.


FranzBusch:

Lukasa (approved):

FranzBusch:
API breakage expected

PR #2580, opened 2023-10-27
Fix exclusive access violation in `NIOAsyncChannelOutboundWriterHandler`

FranzBusch:
# Motivation
We were setting `self.sink = nil` in the `NIOAsyncChannelOutboundWriterHandler` twice in the same call stack which is an exclusivity violation. This happens because the first `self.sink = nil` triggers the `didTerminate` delegate call which again triggered `self.sink = nil`.

# Modification
This PR changes the code to only call `self.sink?.finish()` and only sets the `sink` to `nil` in the `didTerminate` implementation. This follows what we do for the inbound handler implementation. I also added a test that triggers this exclusivity violation.

# Result
No more exclusivity violations in our code.


finagolfin:
I can confirm this gets the NIO HTTP/2 tests passing again on linux x86_64.

glbrntt (approved):

PR #2581, opened 2023-11-01
use feature-specific guard for @retroactive

rnro:
### Motivation:

We should use the more granular guard for uses of `@retroactive` rather than coarse swift versions to guard against corner-cases

### Modifications:

Switch `#if compiler(>=5.11)` for `#if hasFeature(RetroactiveAttribute)`

### Result:

No change in most cases, more protected against corner-cases.


dnadoba:
`hasFeature` is only a thing since Swift 5.8

rnro:
> `hasFeature` is only a thing since Swift 5.8

I spotted that, thanks! I pushed a revised commit which should work

dnadoba (approved):

rnro:
For future reference, unfortunately the short-circuiting version doesn't work (on 5.7.3 at least)
```swift
#if swift(>=5.8) && hasFeature(RetroactiveAttribute)
```

PR #2587, opened 2023-11-06
Fix reordering/reentrancy bug in `NIOAsyncWriter` + `NIOAsyncChannel`

FranzBusch:
# Motivation
While testing the latest async interfaces we found a potential reordering/reentrancy bug in the `NIOAsyncWriter`. This was caused due to our latest performance changes where we fast-pathed calls in `didYield` to not hop. The problem was in the following flow:

1. Task 1: Calls `outbound.write()` -> which led an `EventLoop` enqueue with `didYield`
2. Task 1: Calls `outbound.write()` -> which led an `EventLoop` enqueue with `didYield`
3. EventLoop: While processing the write from 1. the channel became **not** writable
4. Task 1: Calls `outbound.write()` -> which lead to buffering the write in the writer's state machine since we are **not** writable
5. EventLoop: While still processing the write from 1. the channel became writable again -> We informed the `NIOAsyncWriter` about this which unbuffered the write in 4. that was stored in the state machine and call `didYield`. Since, we are on the EventLoop already we processed the write right away

The above flow show-cases a flow where we reordered the write in 2. and 4.

# Modification
This PR fixes the above issue while upholding a few constraints:

1. Produce as few context switches as possible
2. Minimize allocations

I tried different approaches but in the end decided to do the following:
1. Make sure to never call `didYield/didTerminate` from calls on the `NIOAsyncWriter.Sink`
2. Don't coalesce the elements of different writes in the `NIOAsyncWriter` but rather use the suspended tasks to retry a write after they were suspended. I choose to do this since I wanted to avoid any allocation (remember writers are `some Sequence`) and because we assume that continuous contention in a multi producer pattern is low.
3. Make sure that `Sink.finish()` is terminal and does not lead to a `didTerminate` event. This is in line with 1.

One important thing to call out, our `writer.finish()` method is not `async` we have to buffer the finish event and deliver it with the yield that got buffered before we transitioned to `writerFinished`.

# Result
No more reordering/reentrancy problems in `NIOAsyncWriter` or `NIOAsyncChannel`.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
@@ -31,26 +31,20 @@ public protocol NIOAsyncWriterSinkDelegate: Sendable {
     ///
     /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
     /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
-    /// until the ``NIOAsyncWriter`` becomes writable again. All buffered writes, while the ``NIOAsyncWriter`` is not writable,
-    /// will be coalesced into a single sequence.
+    /// until the ``NIOAsyncWriter`` becomes writable again.
     ///
     /// The delegate might reentrantly call ``NIOAsyncWriter/Sink/setWritability(to:)`` while still processing writes.
-    /// This might trigger more calls to one of the `didYield` methods and it is up to the delegate to make sure that this reentrancy is
-    /// correctly guarded against.
     func didYield(contentsOf sequence: Deque<Element>)
 
     /// This method is called once a single element was yielded to the ``NIOAsyncWriter``.
     ///
     /// If the ``NIOAsyncWriter`` was writable when the sequence was yielded, the sequence will be forwarded
     /// right away to the delegate. If the ``NIOAsyncWriter`` was _NOT_ writable then the sequence will be buffered
-    /// until the ``NIOAsyncWriter`` becomes writable again. All buffered writes, while the ``NIOAsyncWriter`` is not writable,
-    /// will be coalesced into a single sequence.
+    /// until the ``NIOAsyncWriter`` becomes writable again.
     ///
     /// - Note: This a fast path that you can optionally implement. By default this will just call ``NIOAsyncWriterSinkDelegate/didYield(contentsOf:)``.
     ///
     /// The delegate might reentrantly call ``NIOAsyncWriter/Sink/setWritability(to:)`` while still processing writes.
-    /// This might trigger more calls to one of the `didYield` methods and it is up to the delegate to make sure that this reentrancy is
-    /// correctly guarded against. 
     func didYield(_ element: Element)
 
     /// This method is called once the ``NIOAsyncWriter`` is terminated.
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -692,15 +679,18 @@ extension NIOAsyncWriter {
...
+            /// The state once the writer finished and there are still task that need to write. This can happen if:
             /// 1. The ``NIOAsyncWriter`` was deinited
             /// 2. ``NIOAsyncWriter/finish(completion:)`` was called.
             case writerFinished(
-                elements: Deque<Element>,
+                isWritable: Bool,
+                inDelegateOutcall: Bool,
+                suspendedYields: _TinyArray<SuspendedYield>,
+                cancelledYields: [YieldID],
+                knownYieldIDs: _TinyArray<YieldID>,
```
glbrntt, line 693: What are the known yield IDs?

Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift
```diff
@@ -76,7 +76,7 @@ final class NIOAsyncWriterTests: XCTestCase {
 
     func testMultipleConcurrentWrites() async throws {
         var elements = 0
-        self.delegate.didYieldHandler = { elements += $0.count }
+        self.delegate.didYieldHandler = { print("DIDYIELD"); elements += $0.count }
```
glbrntt, line 79: nit: leftover print

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -59,13 +53,11 @@ public protocol NIOAsyncWriterSinkDelegate: Sendable {
     /// - The ``NIOAsyncWriter`` is deinited and all yielded elements have been delivered to the delegate.
     /// - ``NIOAsyncWriter/finish()`` is called and all yielded elements have been delivered to the delegate.
     /// - ``NIOAsyncWriter/finish(error:)`` is called and all yielded elements have been delivered to the delegate.
-    /// - ``NIOAsyncWriter/Sink/finish()`` or ``NIOAsyncWriter/Sink/finish(error:)`` is called.
     ///
-    /// - Note: This is guaranteed to be called _exactly_ once.
+    /// - Note: This is guaranteed to be called _at most_ once.
```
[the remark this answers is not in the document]
FranzBusch: So the behaviour here changed that we only call `didTerminate` when somebody called `writer.finished` and not on `sink.finished` anymore. The logic before we quite weird and it caused problems with overlapping exclusive access that I fixed in previous PRs. Now the logic is that a call to `didTerminate` should result in the underlying resource to trigger termination and then a call to `sink.didFinish` indicates that this is done.
I am on the fence with this change so happy to discuss this more.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I added a comment to clarify this but they are the buffered yields before we got the finish.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -515,6 +483,16 @@ extension NIOAsyncWriter {
...
                         self._lock.unlock()
                     }
+
+                    switch yieldResult {
+                    case .yielded:
+                        return
+
+                    case .retry:
+                        // The yield has to be retried. We are recursively calling yield here
+                        // but don't expect to blow the stack.
```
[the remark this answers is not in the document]
FranzBusch: I changed this to a while loop now

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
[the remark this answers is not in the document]
glbrntt: Do you think anyone outside of NIO actually implements this? Might be worth having a quick search.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -483,85 +451,42 @@ extension NIOAsyncWriter {
...
+            return try await withTaskCancellationHandler {
                 // We are manually locking here to hold the lock across the withCheckedContinuation call
                 self._lock.lock()
 
-                let action = self._stateMachine.yield(contentsOf: CollectionOfOne(element), yieldID: yieldID)
+                let action = self._stateMachine.yield(contentsOf: sequence, yieldID: yieldID)
 
                 switch action {
                 case .callDidYield(let delegate):
+                    // We are allocating a new Deque for every write here
```
glbrntt, line 476: Ouch. I didn't realise we did this.

glbrntt (approved):

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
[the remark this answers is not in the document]
FranzBusch: I just did a quick GH search and couldn't find anyone besides `grpc-swift` using this and there we are only relying on the `didTerminate` being called from `writer.finish` and not from `sink.finish` which is the intention of this PR.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: Yeah this one of the early design choices because users can yield us any kind of sequence and we have to normalise this in the `Deque` case. Right now we have provided a fast path for a single element we ought to be able to do the same for `Deque`/`Array` and more.

PR #2582, opened 2023-11-01
try fix deinit abuse

weissi:

PR #2589, opened 2023-11-08
Add `withInboundOutboud` to `NIOAsyncChannel` and deprecate deinit ba…

FranzBusch:
…sed cleanup

# Motivation
We just released our new async NIO APIs and have already gotten quite a bunch of feedback from adopters. One of the feedback was that the deinit based closing that we have added to the `NIOAsyncChannel` has caused problems since it leads to unexpected closure of their `Channel`. Furthermore, it makes it impossible to determine how many open sockets a program has at any given time since deinit based clean up relies on the optimizer and can happen at random times.

# Modifications
This PR adds new inits to `NIOAsyncSequenceProducer` and `NIOAsyncWriter` which disable the `deinit` based clean up and instead replace them with an assertion. This allows developers to still catch these issues at debug time. Furthermore, I added a new `withInboundOutbound` scoped access to `NIOAsyncChannel` which will close the channel at the end of the scope. This still gives users a nice API while not having to care much about closing themselves.

# Result
We are no longer using deinit based clean up and bring back one of the core principles of NIO which is deterministic resource usage.


Sources/NIOCore/AsyncChannel/AsyncChannelOutboundWriterHandler.swift
```diff
@@ -96,15 +95,8 @@ internal final class NIOAsyncChannelOutboundWriterHandler<OutboundOut: Sendable>
     func _didTerminate(error: Error?) {
         self.eventLoop.preconditionInEventLoop()
 
-        switch self.closeRatchet.closeWrite() {
-        case .nothing:
-            break
-
-        case .closeOutput:
+        if self.isOutboundHalfClosureEnabled {
```
FranzBusch, line 98: Call out for reviewers: We are still closing the output side if users are calling `NIOAsyncWriter.finish` and outbound half closure is enabled. This is not triggered by deinit based cleanup

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -229,12 +264,6 @@ extension NIOAsyncSequenceProducer {
             init(throwingSource: ThrowingSource) {
                 self._throwingSource = throwingSource
             }
-
-            @inlinable
-            deinit {
-                // We need to call finish here to resume any suspended continuation.
-                self._throwingSource.finish()
-            }
```
glbrntt, line 237: In the throwing producer we call through to `sourceDeiniitialized` here -- why are they different?

glbrntt (changes requested):

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -164,6 +176,52 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
             outboundWriter: outboundWriter
         )
     }
+
+    /// Provides scoped access to the inbound and outbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter body: A closure that gets scoped access to the inbound and outbound.
+    public func withInboundOutbound<Result>(
```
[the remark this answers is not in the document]
FranzBusch: Hm why? The only type called stream is the `InboundStream`. I was going between `withInboundStreamAndOutboundWriter` before but decided to follow the same logic where we renamed them to `inbound` and `outbound`

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -164,6 +176,52 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter body: A closure that gets scoped access to the inbound.
+    public func withInbound<Result>(
+        _ body: (NIOAsyncChannelInboundStream<Inbound>) async throws -> Result
+    ) async throws -> Result where Outbound == Never{
+        try await self.withInboundOutbound { inbound, _ in
+            try await body(inbound)
+        }
+    }
```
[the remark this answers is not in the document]
FranzBusch: I first only had the method with both but in the case of the `ServerBootstrap` you have to spell `withInboundOutbound { _, outbound in` a lot and I personally found this very annoying.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
FranzBusch: Because we store the throwing producer in here. So if we get deinited we will implicitly deinit the throwing one which will execute `sourceDeinitialized`.

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -791,6 +872,63 @@ extension NIOThrowingAsyncSequenceProducer {
...
+                    // We have a continuation, this means our buffer must be empty
+                    // Furthermore, we can now transition to finished
+                    // and resume the continuation with the failure
+                    precondition(buffer.isEmpty, "Expected an empty buffer")
+
+                    self._state = .finished(iteratorInitialized: iteratorInitialized)
+
+                    return .resumeContinuationWithFailureAndCallDidTerminate(continuation, nil)
+                } else {
+                    assertionFailure("Deinited NIOAsyncSequenceProducer.Source without finishing it first")
```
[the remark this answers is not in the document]
FranzBusch: Good point. We probably should `precondition` here.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -164,6 +176,52 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+
+    /// Provides scoped access to the inbound and outbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter body: A closure that gets scoped access to the inbound and outbound.
+    public func withInboundOutbound<Result>(
+        _ body: (NIOAsyncChannelInboundStream<Inbound>, NIOAsyncChannelOutboundWriter<Outbound>) async throws -> Result
+    ) async throws -> Result {
+        let result: Result
```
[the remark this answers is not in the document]
FranzBusch: Oh yeah that's a nice pattern but it does require a few extensions on `Result` and in my opinion for this single method it is not worth bringing in.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -139,17 +139,52 @@ public struct NIOAsyncSequenceProducer<
     /// - Parameters:
     ///   - elementType: The element type of the sequence.
     ///   - backPressureStrategy: The back-pressure strategy of the sequence.
+    ///   - finishOnDeinit: Indicates if ``NIOAsyncSequenceProducerDelegate/didTerminate()`` should be called on deinit. We do not recommend to rely on
```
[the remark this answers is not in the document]
FranzBusch: Right, I reworded the documentation. `DidTerminate` is still called at the same time just `source.finish` is not called anymore on `deinit`

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -164,6 +176,52 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
         )
     }
+
+    /// Provides scoped access to the inbound and outbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter body: A closure that gets scoped access to the inbound and outbound.
+    public func withInboundOutbound<Result>(
+        _ body: (NIOAsyncChannelInboundStream<Inbound>, NIOAsyncChannelOutboundWriter<Outbound>) async throws -> Result
```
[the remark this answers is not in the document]
glbrntt: I didn't ask for external label ;)

The instructions in the API guidelines say should have internal labels because they act as documentation and act as hints in auto completion etc.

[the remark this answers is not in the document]
FranzBusch: Oh sorry, I misunderstood you then. Pushed a commit with internal labels.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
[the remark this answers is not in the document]
glbrntt: I just don't think the name is particularly clear. 

I'm not sure `withStreams` is the right name either. I suggested "streams" because it's how I'd refer to `inbound` and `outbound` collectively.

Maybe the naming should indicate the semantics of what the function offers rather than what the closure provides to the caller? Thinking along the lines of "with auto closing streams", "with lifecycle managed streams" (I don't like these _either_ but throwing them out anyway...)

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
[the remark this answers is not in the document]
glbrntt: Yup no need to bring them in.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
glbrntt: Gotcha, thanks for explaining. As we no longer need `deinit` can this become a `struct` then?

FranzBusch:
@glbrntt Just renamed the `withInboundOutbound` to `executeThenCloseChannel`.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -409,9 +444,14 @@ extension NIOAsyncWriter {
         @inlinable
         /* fileprivate */ internal init(
```
weissi, line 445: There are still two `deinit`s in `NIOAsyncWriter` that haven't been fixed here: https://github.com/apple/swift-nio/pull/2582/files#diff-7914f7c4968a26adfb4f6be410d7a1c12a72a35ed7bfd1c3a62f08e08f4b37f1R178-R376

FranzBusch: Oh I actually forgot to call the new method in one of those. The other one was already correct. Both will unconditionally call through to the state machine and we make the decision what happens in the state machine.

weissi: Can we not flip this around to make this reviewable and very clear: Unless the user tells us to, we're not gonna do anything (but assertions) in `deinit`. If `deinit` seems to do a thing (like calling something else) then you lose local reasoning and have to follow all possible trails to check if that code is actually `deinit`-safe Swift code. If it's just

```
deinit {
    if self.yoloMode {
       self.stateMachine.doStuff()
    }
}
```

then it's super clear that nothing bad is going to happen

weissi: `deinit` shouldn't drive state machines, it shouldn't do anything if at all possible.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -682,24 +742,32 @@ extension NIOAsyncWriter {
...
+                    // the writer.
+                    precondition(suspendedYields.isEmpty, "We have outstanding suspended yields")
+
+                    // We can transition to finished directly
+                    self._state = .finished(sinkError: nil)
 
-                return .callDidTerminate(delegate)
+                    return .callDidTerminate(delegate)
+                } else {
+                    preconditionFailure("Deinited NIOAsyncWriter without finishing")
```
weissi, line 769: Users are going to see this message so I'd probably add some text like which exact method they're supposed to call when.

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
FranzBusch: I disagree with this. I prefer to have all reasoning in the state machine and only have a single place that takes decisions. Furthermore, we must involve the state machine anyhow because we need to understand if we are already finished. I would prefer to keep it like I implemented here.

weissi: I do understand what you're saying but `deinit` is a very special case. Code that's in there runs under special rules (must not publish `self` or else UB or crash). If you want anybody to understand this code you have to have it in `deinit`.

Imagine this deinit

```
deinit {
    self.stateMachine.finish()
}
```

or 

```
deinit {
    self.finish()
}
```

looks nice and simple. If a future PR changes `func finish()` in the state machine or on `self`, then the reviewer of that PR will have _no clue_ that this code may be called from `deinit`. Normally, the rules are simple: No code is called from deinit that isn't. If we bend that rule it's really important that this is visible in every diff that touches it. Hiding it in the state machine might have dire consequences that can be ~impossible to fix. If Swift had a mandatory `@mayBeCalledFromDeinit` attribute this were different but we don't

FranzBusch: That's why we are not calling `finish` in the state machine but specific methods called `writerDeinitialized` or `sinkDenitialized`. This makes it very explicit to understand where this call comes from. IMO that's better than calling out to the state machine in deinit to check for a specific state and then make the decision what to do outside of the state machine. That's just blurs the line where state logic happens.

Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducer.swift
```diff
@@ -139,17 +139,52 @@ public struct NIOAsyncSequenceProducer<
     /// - Parameters:
     ///   - elementType: The element type of the sequence.
     ///   - backPressureStrategy: The back-pressure strategy of the sequence.
+    ///   - finishOnDeinit: Indicates if ``NIOAsyncSequenceProducer/Source/finish()`` should be called on deinit of the.
```
glbrntt, line 142: Incomplete sentence: "on deinit of the" ...?

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -161,14 +161,23 @@ public struct NIOAsyncWriter<
...
+        init(storage: Storage, finishOnDeinit: Bool) {
             self._storage = storage
+            self._finishOnDeinit = finishOnDeinit
         }
 
         @inlinable
         deinit {
-            _storage.writerDeinitialized()
+            if !self._finishOnDeinit && !self._storage.isWriterFinished {
+                preconditionFailure("Deinited NIOAsyncWriter without calling finish()")
```
weissi, line 176: beautiful

weissi (approved):
thanks so much

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEchoAsyncChannel.swift
```diff
@@ -53,37 +53,43 @@ func runTCPEchoAsyncChannel(numberOfWrites: Int, eventLoop: EventLoop) async thr
     try await withThrowingTaskGroup(of: Void.self) { group in
         // This child task is echoing back the data on the server.
         group.addTask {
-            for try await connectionChannel in serverChannel.inbound {
-                for try await inboundData in connectionChannel.inbound {
-                    try await connectionChannel.outbound.write(inboundData)
+            try await serverChannel.withInbound { serverChannelInbound in
```
glbrntt, line 56: `executeThenClose`

Benchmarks/Benchmarks/NIOPosixBenchmarks/TCPEchoAsyncChannel.swift
```diff
@@ -53,37 +53,43 @@ func runTCPEchoAsyncChannel(numberOfWrites: Int, eventLoop: EventLoop) async thr
...
+                for try await connectionChannel in serverChannelInbound {
+                    try await connectionChannel.withInboundOutbound { connectionChannelInbound, connectionChannelOutbound in
```
glbrntt, line 58: `executeThenClose` (and elsewhere)

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -164,6 +176,52 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+    }
+
+    /// Provides scoped access to the inbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter body: A closure that gets scoped access to the inbound.
+    public func executeThenClose<Result>(
+        _ body: (_ inbound: NIOAsyncChannelInboundStream<Inbound>) async throws -> Result
+    ) async throws -> Result where Outbound == Never{
```
glbrntt, line 220:
```suggestion
    ) async throws -> Result where Outbound == Never {
```

glbrntt (approved):

PR #2592, opened 2023-11-14
Add `closeOnDeinit` to the `NIOAsyncChannel` init

FranzBusch:
# Motivation
In my previous PR, I already did the work to add `finishOnDeinit` configuration to the `NIOAsyncWriter` and `NIOAsyncSequenceProducer`. This PR also automatically migrated the `NIOAsyncChanell` to set the `finishOnDeinit = false`. This was intentional since we really want users to not use the deinit based cleanup; however, it also broke all current adopters of this API semantically and they might now run into the preconditions.

# Modification
This PR reverts the change in `NIOAsyncChannel` and does the usual deprecate + new init dance to provide users to configure this behaviour while still nudging them to check that this is really what they want.

# Result
Easier migration without semantically breaking current adopters of `NIOAsyncChannel`.

### DIFF
--- a/Sources/NIOCore/AsyncChannel/AsyncChannel.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannel.swift
@@ -99,20 +99,17 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
     /// - Parameters:
     ///   - channel: The ``Channel`` to wrap.
     ///   - configuration: The ``NIOAsyncChannel``s configuration.
-    ///   - closeOnDeinit: Indicates if the underlying channel should be closed once the `inbound` and `outbound` have been deinited. We do not recommend to rely on
-    ///   deinit based resource tear down.
     @inlinable
     public init(
-        synchronouslyWrapping channel: Channel,
-        configuration: Configuration = .init(),
-        closeOnDeinit: Bool
+        wrappingChannelSynchronously channel: Channel,
+        configuration: Configuration = .init()
     ) throws {
         channel.eventLoop.preconditionInEventLoop()
         self.channel = channel
         (self._inbound, self._outbound) = try channel._syncAddAsyncHandlers(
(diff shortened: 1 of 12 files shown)


Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -99,6 +99,61 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
     /// - Parameters:
     ///   - channel: The ``Channel`` to wrap.
     ///   - configuration: The ``NIOAsyncChannel``s configuration.
+    ///   - closeOnDeinit: Indicates if the underlying channel should be closed once the `inbound` and `outbound` have been deinited. We do not recommend to rely on
+    ///   deinit based resource tear down.
+    @inlinable
+    public init(
+        synchronouslyWrapping channel: Channel,
+        configuration: Configuration = .init(),
+        closeOnDeinit: Bool
```
glbrntt, line 108: It's a shame we have to rely on users setting this to get the right behaviour because they can still set this to be `true` and get the old behaviour without a deprecation warning.

Ideally users shouldn't have to think about this and we would have two differently spelled `init`s, the deprecated old one and the new no-close-on-deinit one. That'd make it obvious which one to call. 


I'm not sure what to suggest for naming here, however. A couple of ideas (don't love either):
- a `static` helper `static func synchronouslyWrapping(_:configuration:) -> Self`
- `init(wrappingChannelSynchronously:configuration:)`

FranzBusch: I agree with those points. I had it quickly prototyped with a factory method but disliked it. Maybe just having a `wrappingChannelSynchronously` is the way out :D

yim-lee:

glbrntt (approved):
LGTM aside from the now redundant docs

FranzBusch:

PR #2593, opened 2023-11-15
Revert "Back out new typed HTTP protocol upgrader (#2579)"

FranzBusch:
# Motivation
We have reverted the typed HTTP protocol upgrader pieces since adopters were running into a compiler bug (https://github.com/apple/swift/pull/69459) that caused the compiler to emit strong references to `swift_getExtendedExistentialTypeMetadata`. The problem is that `swift_getExtendedExistentialTypeMetadata` is not available on older runtimes before constrained existentials have been introduced. This caused adopters to run into runtime crashes when loading any library compiled with this NIO code.

# Modifications
This PR reverts the revert and guard all new code in a compiler guard that checks that we are either on non-Darwin platforms or on a new enough Swift compiler that contains the fix.

# Result
We can offer the typed HTTP upgrade code to our adopters again.

### DIFF
--- a/Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
+++ b/Sources/NIOHTTP1/HTTPTypedPipelineSetup.swift
@@ -1,0 +1,248 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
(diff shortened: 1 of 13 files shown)


FranzBusch:
@glbrntt The only changes in here are adapting to the new `NIOAsyncChannel.executeThenClose` APIs and adding this compiler guard `#if !canImport(Darwin) || (canImport(Darwin) && swift(>=5.10))` in a few places

FranzBusch:

glbrntt:
@FranzBusch is it much bother for you to split this into two commits? I.e. the revert (even if it doesn't build) then the fixups? Much easier to review then...

FranzBusch:
@glbrntt This is now separated in the revert and the fix-ups

glbrntt (approved):

PR #2595, opened 2023-11-16
Changes to support building with Musl

al45tair:
Various changes to allow building NIO for a Linux system that is using Musl instead of Glibc.

### Motivation:

We would like to be able to build NIO on top of Musl, as well as the usual Glibc.

### Modifications:

Define `_GNU_SOURCE` in the `Package.swift` rather than in `shim.c` (this is required because `_GNU_SOURCE` affects modular headers).

Add an import for Musl to IO.swift.

Add code to disable `SIGPIPE` to `SocketProtocols.swift`.

Remove types from a pile of functions in `System.swift`; Swift will use the correct type automatically (except in cases where there are multiple versions of a function, e.g. `ioctl()`, in which case we need to be explicit which one we mean).

### Result:

NIO should build against Musl.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -115,7 +115,10 @@ let package = Package(
         ),
         .target(
             name: "CNIOLinux",
-            dependencies: []
+            dependencies: [],
+            cSettings: [
+                .define("_GNU_SOURCE"),
+            ]
         ),
         .target(
             name: "CNIODarwin",
--- a/IntegrationTests/tests_02_syscall_wrappers/defines.sh
+++ b/IntegrationTests/tests_02_syscall_wrappers/defines.sh
@@ -36,7 +36,10 @@ let package = Package(
             dependencies: ["CNIOLinux", "CNIODarwin", "NIOCore"]),
         .target(
             name: "CNIOLinux",
-            dependencies: []),
+            dependencies: [],
+            cSettings: [
+                .define("_GNU_SOURCE")
+            ]),
         .target(
             name: "CNIODarwin",
             dependencies: []),
--- a/IntegrationTests/tests_05_assertions/defines.sh
+++ b/IntegrationTests/tests_05_assertions/defines.sh
@@ -31,7 +31,10 @@ let package = Package(
             dependencies: ["CNIOLinux", "CNIODarwin", "NIOCore"]),
         .target(
             name: "CNIOLinux",
-            dependencies: []),
+            dependencies: [],
+            cSettings: [
+                .define("_GNU_SOURCE")
+            ]),
         .target(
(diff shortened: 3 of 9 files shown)


dnadoba (commented):
LGTM but would like someone else to take another look as well

dnadoba:

FranzBusch (approved):
LGTM!

Package.swift
```diff
@@ -115,7 +115,10 @@ let package = Package(
         ),
         .target(
             name: "CNIOLinux",
-            dependencies: []
+            dependencies: [],
+            cSettings: [
+                .define("_GNU_SOURCE"),
```
Lukasa, line 120: Are we confident we're allowed to do this? Previously this has appeared to conflict with Swift's glibc module, which was declared without this definition.

al45tair: Hmmm. Well, we're definitely _not_ safe just `#define`ing it in a source file, because with modular includes you have to define it on the command line (or it won't affect the modularized headers). If it doesn't work with the Glibc module, that would be… bad.

FranzBusch:
The CI seems to be failing with

```
17:09:30 /tmp/.swift-nio-syscall-wrappers-sh-test_6C5iKP/syscallwrapper/Sources/CNIOLinux/shim.c:22:2: error: You must define _GNU_SOURCE
17:09:30 #error You must define _GNU_SOURCE
```

This might just be a problem for the syscall wrapper tests. @al45tair you mind taking a look?

al45tair:
Will do. I also need to have another look at the `System.swift` stuff; I think maybe we _do_ need the types there (I've just had a compiler crash that looks suspiciously like it might be related to that code; obviously it shouldn't crash, but I started thinking that _maybe_ the types are there to force the compiler to pick the underlying system functions rather than ones in the respective overlay).

al45tair:
Update: the crash wasn't that problem (it was another problem, in a different dependency), and everything seems to work without the type information being mentioned explicitly.

al45tair:
> The CI seems to be failing with
> 
> ```
> 17:09:30 /tmp/.swift-nio-syscall-wrappers-sh-test_6C5iKP/syscallwrapper/Sources/CNIOLinux/shim.c:22:2: error: You must define _GNU_SOURCE
> 17:09:30 #error You must define _GNU_SOURCE
> ```
> 
> This might just be a problem for the syscall wrapper tests. @al45tair you mind taking a look?

I think I have a fix for that. Seems there are some extra `Package.swift`s hiding in the integration tests' shell scripts.

al45tair:

Lukasa (approved):

PR #2597, opened 2023-11-24
RFC/DNM: Add NIOAsyncWriterSinkDelegate._didSuspend hook for testing

simonjbeaumont:
### Motivation:

The tests for the `NIOAsyncWriter` currently rely on a number of `Task.sleep` calls to wait for tasks to be suspended or cancelled before performing some other operation. These have been shown to be flakey in CI. We could just increase all the sleeps, but instead we can make use of `XCTestExpectation` to wait for things to happen. The events that we are waiting on are these:

1. Waiting within a task for the task to be cancelled before we yield.
2. Waiting outside a task that is performing a yield for that task to be suspended.

This didn't require any additional changes to the writer for (1) and can just wait for an expectation inside the task that we fulfil after the cancelling the task from outside.

For (2) we need to have a hook that we can use when the task has performed the state transition and is parked that we can use to fulfil the expectation.

### Modifications:

- Add `_didSuspend` function to `protocol NIOAsyncWriterSinkDelegate`.
- Provide default, empty implementation for `NIOAsyncWriterSinkDelegate._didSuspend`.
- Replace all calls to `Task.sleep(nanoseconds:)` in tests with `await fulfillment(of:[timeout:])`.

### Result:

Tests should be more deterministic.


Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import DequeModule
-import NIOCore
+@testable import NIOCore
```
FranzBusch, line 16: I prefer if we don't use `@testable`. It is a horrible compiler hack and I would love to slowly get away from it by either using `SPI` or the new `package` modifier. For now let's use `@_spi(Testing)` and we can migrate that in the future.

Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift
```diff
@@ -81,6 +94,18 @@ final class NIOAsyncWriterTests: XCTestCase {
         super.tearDown()
     }
 
+    func assert(
+        numSuspends: Int,
+        numYields: Int,
+        numTerminates: Int,
```
FranzBusch, line 100: NIT:
```suggestion
        suspendCallCount: Int,
        yieldCallCount: Int,
        terminateCallCount: Int,
```

FranzBusch (commented):
I hate it but it solves the flakiness, so let's do it! We should probably do the same for the `NIO[Throwing]AsyncSequenceProducer` tests. Do you mind doing the same there? Can even put it into this PR if you want!

Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift
simonjbeaumont: Done, throughout.

simonjbeaumont:
> I hate it but it solves the flakiness, so let's do it! We should probably do the same for the `NIO[Throwing]AsyncSequenceProducer` tests. Do you mind doing the same there? Can even put it into this PR if you want!

Added commit to give the `NIOThrowingAsyncSequenceProducerTests` the same treatment.

Tests/NIOCoreTests/AsyncSequences/NIOThrowingAsyncSequenceTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
+@testable import NIOCore
```
FranzBusch, line 15: Same nit about `@testable`

FranzBusch (approved):

PR #2598, opened 2023-11-25
Make NIOFileDescriptor/FileRegion/IOData Sendable & soft-deprecated

weissi:
Motivation:

`IOData` is a legacy but alas also core type that needs to be `Sendable`. Before this PR however it can't be `Sendable` because it holds a `FileRegion` which holds a `NIOFileDescriptor`. So let's make all of these `Sendable` but let's also start the deprecation journey for the following types:

- `IOData`, now soft-deprecated (no warnings) because on its reliance on `FileRegion`
- `FileRegion`, now soft-deprecated (no warnings) because on its reliance on `NIOFileHandle`
- `NIOFileHandle`, now soft-deprecated (warnings on the `NIOFileHandle(descriptor:)` constructor but with a `NIOFileHandle(_deprecatedTakingOwnershipOfDescriptor:)` alternative
- `NonBlockingFileIO`, now soft-deprecated (warnings on the `openFile` functions (but with `_deprecated` alternatives) because of their reliance on `NIOFileHandle)

Modification:

- Make `NIOFileDescriptor`, `FileRegion` and `IOData` `Sendable` by tracking the fd number and the usage state in an atomic
- Enforce singular access by making the `withFileDescriptor { fd ... }` function atomically exchange the fd number for a "I'm busy" sentinel value
- Start deprecating `IOData`, `NIOFileHandle`, `NonBlockingFileIO`, `FileRegion`

Result:
- `NIOFileDescriptor`, `FileRegion` and `IOData` can be `Sendable`

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2024 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -2407,8 +2407,8 @@ extension NIOPipeBootstrap {
         let channelOptions = self._channelOptions
 
         let channel: PipeChannel
-        let inputFileHandle: NIOFileHandle?
-        let outputFileHandle: NIOFileHandle?
+        let pipeChannelInput: SelectablePipeHandle?
+        let pipeChannelOutput: SelectablePipeHandle?
         do {
             if let input = input {
                 try self.validateFileDescriptorIsNotAFile(input)
@@ -2417,18 +2417,18 @@ extension NIOPipeBootstrap {
                 try self.validateFileDescriptorIsNotAFile(output)
             }
 
-            inputFileHandle = input.flatMap { NIOFileHandle(descriptor: $0) }
-            outputFileHandle = output.flatMap { NIOFileHandle(descriptor: $0) }
+            pipeChannelInput = input.flatMap { SelectablePipeHandle(takingOwnershipOfDescriptor: $0) }
+            pipeChannelOutput = output.flatMap { SelectablePipeHandle(takingOwnershipOfDescriptor: $0) }
             do {
                 channel = try self.hooks.makePipeChannel(
                     eventLoop: eventLoop as! SelectableEventLoop,
-                    inputPipe: inputFileHandle,
-                    outputPipe: outputFileHandle
+                    input: pipeChannelInput,
+                    output: pipeChannelOutput
                 )
             } catch {
                 // Release file handles back to the caller in case of failure.
-                _ = try? inputFileHandle?.takeDescriptorOwnership()
-                _ = try? outputFileHandle?.takeDescriptorOwnership()
+                _ = try? pipeChannelInput?.takeDescriptorOwnership()
+                _ = try? pipeChannelOutput?.takeDescriptorOwnership()
                 throw error
             }
         } catch {
@@ -2447,10 +2447,10 @@ extension NIOPipeBootstrap {
                 channel.registerAlreadyConfigured0(promise: promise)
                 return promise.futureResult.map { result }
             }.flatMap { result -> EventLoopFuture<ChannelInitializerResult> in
-                if inputFileHandle == nil {
+                if pipeChannelInput == nil {
                     return channel.close(mode: .input).map { result }
                 }
-                if outputFileHandle == nil {
+                if pipeChannelOutput == nil {
                     return channel.close(mode: .output).map { result }
                 }
                 return channel.selectableEventLoop.makeSucceededFuture(result)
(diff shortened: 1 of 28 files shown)


FranzBusch:
Thanks for opening the discussion here. So far we have stayed away from making those changes and debated if we should just drop the whole `IOData` support in a potential NIO 3. Though we haven't come to any agreement on that matter yet. This would definitely solve the immediate pain of the `Sendable` problem that we are facing now.

FranzBusch:
After having looked at the strict checking this is one of the last remaining issues in `NIOCore`. The PR itself looks good and IMO we should move forward with this.

weissi:
@Lukasa I think this one's ready to review now, making `IOData` `Sendable` is really important and making `NIOFileDescriptor` `Sendable` is a precursor for that.

Sources/NIOPosix/PipePair.swift
```diff
@@ -13,55 +13,92 @@
...
+        for fileHandle in [self.inputSPH, self.outputSPH].compactMap({ $0 }) {
             try fileHandle.withUnsafeHandle {
                 try PipePair.ignoreSIGPIPE(descriptor: $0)
             }
         }
     }
 
     var description: String {
-        "PipePair { in=\(String(describing: inputFD)), out=\(String(describing: inputFD)) }"
+        "PipePair { in=\(String(describing: self.inputSPH)), out=\(String(describing: self.outputSPH)) }"
```
weissi, line 101: there was actually a bug here (it previously printed the input fd twice, whoops)

weissi:
@FranzBusch / @Lukasa Did all the remaining steps which also take `NIOFileHandle` etc out of the fast path (it used to be in there for `PipeChannel`, completely unnecessarily).

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -434,9 +434,10 @@ final class PendingStreamWritesManager: PendingWritesManager {
         case .fileRegion(let file):
             let readerIndex = file.readerIndex
             let endIndex = file.endIndex
-            return try file.fileHandle.withUnsafeFileDescriptor { fd in
-                self.didWrite(itemCount: 1, result: try operation(fd, readerIndex, endIndex))
+            let writeResult = try file.fileHandle.withUnsafeFileDescriptor { fd in
+                try operation(fd, readerIndex, endIndex)
             }
+            return self.didWrite(itemCount: 1, result: writeResult)
```
weissi, line 440: This was always a bug: Should've never called out whilst in `withUnsafeFileDescriptor`...

weissi:
> I'm extremely unsure about this. The changes to `FileHandle` make it lightly terrifying: it appears to claim in its API surface that it's safe to use from multiple threads, but it absolutely isn't: you have zero chance of writing a correct program if you do that.

@Lukasa How do you mean? You can't use a `NIO.FileHandle` multiple times at the same time (not before, not after this change), _however_ you can easily hand it from one thread to the next which is now completely safe. Previously, we _pretended_ (in NBFIO) that it's safe but documented that it's not. `NonBlockingFileIO` was technically never correctly usable with `NIOFileHandle`, it just happened to work -- now it actually works if you stick to the rules and sequentialise access.

There's only one piece of user-space state in `NIO.FileHandle` which is the fd number which remains constant until we `close` it. But the `close` mutation must be correctly synchronised. Previously, that wasn't the case (we just looked the other way), now it is.

And FWIW, from a system's PoV, you can absolutely use a single file descriptor at the same time safely from multiple threads, `pwrite`/`pread` are pretty much designed for that. But `NIO.FileHandle` did never officially support this, not before or after this PR.

Finally, `IOData` really has to become `Sendable` and there's no reason that it shouldn't be.

Lukasa:
The difference between before and after is that `NIOFileHandle` doesn't currently pretend that the use is safe: marking the type `Sendable` does. But as noted, right now the code that was actually safe becomes randomly buggy, as you can no longer get the FD concurrently from multiple threads. There's already a spelling in Swift 6 for moving the value between threads, which we can rely on to express that semantic.

My proposal here is not to save `NIOFileHandle`, but to deprecate it. Similarly for `IOData`, which only meaningfully lives on in the HTTP types. A better solution, IMO, is not to offer this weird footgun of a type for much longer.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -2417,8 +2417,8 @@ extension NIOPipeBootstrap {
                 try self.validateFileDescriptorIsNotAFile(output)
             }
 
-            inputFileHandle = input.flatMap { NIOFileHandle(descriptor: $0) }
-            outputFileHandle = output.flatMap { NIOFileHandle(descriptor: $0) }
+            inputFileHandle = input.flatMap { NIOFileHandle(_deprecatedTakingOwnershipOfDescriptor: $0) }
+            outputFileHandle = output.flatMap { NIOFileHandle(_deprecatedTakingOwnershipOfDescriptor: $0) }
```
Lukasa, line 2421: These make it look like we're gonna need a way to have an internal non-deprecated flow that can be used for the few internal use-cases we have that aren't totally FUBAR. And then a GH issue to get us to transition away from it. Alternatively, we need a different spelling here.

weissi: @Lukasa good catch. Actually, I don't think we need to go via NIOFileHandle at all here.  gimme a sec

weissi: @Lukasa yes, fixed, gone now

Lukasa (approved):

PR #2600, opened 2023-11-26
Build for Android with NDK 26, by accounting for the new nullability annotations

finagolfin:
### Motivation:

Fix build for the latest LTS NDK 26

### Modifications:

- Update C declarations
- Add force unwraps where needed

### Result:

Everything works on Android with NDK 26b

This is needed because [Bionic recently added a bunch of these annotations](https://android.googlesource.com/platform/bionic/+/1cda74daba3834a819a58e4b4aae909a8ede5a7e%5E%21/#F0). I made sure this pull doesn't break anything by testing it on linux x86_64, and the force unwraps with the previous NDK 25c too. I used this patch with others to build the Swift toolchain and this package for my Android CI, finagolfin/swift-android-sdk#122.

### DIFF
--- a/Sources/NIOCore/BSDSocketAPI.swift
+++ b/Sources/NIOCore/BSDSocketAPI.swift
@@ -68,8 +68,13 @@ import Musl
 #endif
 import CNIOLinux
 
+#if os(Android)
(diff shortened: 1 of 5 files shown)


Lukasa (approved):
Thanks for the cleanup!

PR #2601, opened 2023-11-27
Add tests to validate the behaviour when requests/response content-length headers are wrong in HTTP1

hamzahrmalik:
### Motivation:
To better understand what happens when a request or response specifies a content-length different to the actual length of the content

### Modifications:

Add some tests

### Result:

We can conclude that when the content is longer than the header suggests, the first part of the content is used and the rest is leftover and treated as part of the next request. 
When the content is shorter than the header suggests, we wait for more bytes

### DIFF
--- a/Tests/NIOHTTP1Tests/ContentLengthTests.swift
+++ b/Tests/NIOHTTP1Tests/ContentLengthTests.swift
@@ -1,0 +1,131 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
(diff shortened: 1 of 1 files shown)


Tests/NIOHTTP1Tests/ContentLengthTests.swift
```diff
@@ -0,0 +1,131 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIOCore
+import NIOEmbedded
+import NIOHTTP1
+
+class ContentLengthTests: XCTestCase {
```
FranzBusch, line 20:
```suggestion
final class ContentLengthTests: XCTestCase {
```

Tests/NIOHTTP1Tests/ContentLengthTests.swift
```diff
@@ -0,0 +1,131 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
```
FranzBusch, line 5:
```suggestion
// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
```

FranzBusch (commented):
LGTM!

Lukasa (approved):

PR #2608, opened 2023-12-13
Add delegate for collecting eventloop tick metrics

hamzahrmalik:
Add delegate for collecting eventloop tick metrics

### Motivation:

Users need a way to monitor how eventloops are running, to detect problems such as the eventloop being blocked

### Modifications:

Add a delegate to SelectableEventLoop and multithreadedEventLoop which is called on every tick. The delegate is given information about the tick


Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -30,6 +30,27 @@ internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+    public var startTime: NIODeadline
+
+    internal init(numberOfTasks: Int, startTime: NIODeadline) {
+        self.numberOfTasks = numberOfTasks
+        self.startTime = startTime
+    }
+}
+
+/// Implement this delegate to receive information about the EventLoop, such as each tick
+public protocol EventLoopDelegate {
```
weissi, line 47: - needs a `NIO` prefix to satisfy the public API guarantees NIO has
- I'd call this `NIOEventLoopMetricsDelegate` or something

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -76,6 +76,7 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
                                    canEventLoopBeShutdownIndividually: Bool,
                                    selectorFactory: @escaping () throws -> NIOPosix.Selector<NIORegistration>,
                                    initializer: @escaping ThreadInitializer,
+                                   delegate: EventLoopDelegate?,
```
weissi, line 79: this this should be `metricsDelegate` or similar

weissi: What's also odd is that we seem to have one delegate which gets metrics from multiple EventLoops in that EventLoopGroup.

To make this useful to users we should aggregate the metrics from all loops in a group, no? If we don't give the user to stats from all loops collectively, then the user would have to perform the aggregation themselves -- likely with locks across all loops -- which isn't ideal.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -30,6 +30,27 @@ internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
 #endif
 }
 
+/// Information about an EventLoop tick
+public struct EventLoopTickInfo {
```
weissi, line 34: - needs `NIO` prefix
- lacks identifier for EventLoop
- as said above, I think we should probably periodically give the user one value which contains information from _all_ loops in a group together. Something like


```
public struct NIOEventLoopMetrics {
    public struct LoopMetric {
        var loopID: ...
        var numberOfTasks: Int
        var startTime: NIODeadline
    }
    
    public var loopMetrics: [LoopMetric]
}
```

I know this raises a lot of questions but I think we should think through how a user would use this and then settle on what NIO should provide. Just calling the delegate _on each loop_ for every tick is easy but requires a lot of heavy-lifting on the user's part which seems bound to go wrong, no?

Lukasa: IMO the trade-off falls the other way: we should offer the lowest level thing, which lets expert users get what they need, first. We can then do the engineering to produce something better. But frankly we've spent years not exposing these metrics because we wanted to do it right, and I'm a bit sick of stumbling around in the dark. So if we can get something good enough cheaply, we should.

FranzBusch: I agree with Cory here. Let's do the simplest approach and just expose a delegate that gets called every tick and users have to do the aggregation. Over time we will see what kind of aggregation users are doing and provide pre-backed delegates that give them an even simpler interface

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -146,41 +168,48 @@ public final class MultiThreadedEventLoopGroup: EventLoopGroup {
...
                               canBeShutDown: Bool,
                               threadNamePrefix: String,
+                              metricsDelegate: NIOEventLoopMetricsDelegate?,
                               selectorFactory: @escaping () throws -> NIOPosix.Selector<NIORegistration>) {
         precondition(numberOfThreads > 0, "numberOfThreads must be positive")
         let initializers: [ThreadInitializer] = Array(repeating: { _ in }, count: numberOfThreads)
         self.init(threadInitializers: initializers,
                   canBeShutDown: canBeShutDown,
                   threadNamePrefix: threadNamePrefix,
+                   metricsDelegate: metricsDelegate,
```
Lukasa, line 193: Nit: indentation.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -30,6 +30,30 @@ internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
 #endif
 }
 
+/// Information about an EventLoop tick
+public struct NIOEventLoopTickInfo {
```
Lukasa, line 34: Let's make this `Sendable` at the very least. Maybe `Hashable` too.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -30,6 +30,30 @@ internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+
+    internal init(eventLoopID: ObjectIdentifier, numberOfTasks: Int, startTime: NIODeadline) {
+        self.eventLoopID = eventLoopID
+        self.numberOfTasks = numberOfTasks
+        self.startTime = startTime
+    }
+}
+
+/// Implement this delegate to receive information about the EventLoop, such as each tick
+public protocol NIOEventLoopMetricsDelegate {
```
Lukasa, line 50: This type needs to be `Sendable`.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -526,6 +557,7 @@ Further information:
                 }
 
                 // Execute all the tasks that were submitted
+                tasksProcessedInTick += self.tasksCopy.count
```
Lukasa, line 560: We should be a little careful here. It's not really practically possible, but conceptually this addition can trap. We probably want to replace this with an `addingReportingOverflow` that, if we overflow, saturates instead.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -547,6 +579,8 @@ Further information:
                 // Drop everything (but keep the capacity) so we can fill it again on the next iteration.
                 self.tasksCopy.removeAll(keepingCapacity: true)
             }
+            let tickInfo = NIOEventLoopTickInfo(eventLoopID: ObjectIdentifier(self), numberOfTasks: tasksProcessedInTick, startTime: tickStartTime)
```
Lukasa, line 582: We can optimize this and create the loop ID only once. The optimizer might do this for us, but it's nicer not to rely on it.

Lukasa (commented):
I'm generally happy with this approach @hamzahrmalik. I've left a few notes in the diff but I think it's a good start. Now we just need some tests!

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -31,6 +31,30 @@ internal func withAutoReleasePool<T>(_ execute: () throws -> T) rethrows -> T {
...
+        self.eventLoopID = eventLoopID
+        self.numberOfTasks = numberOfTasks
+        self.startTime = startTime
+    }
+}
+
+/// Implement this delegate to receive information about the EventLoop, such as each tick
+public protocol NIOEventLoopMetricsDelegate: Sendable {
+    /// Called after a tick has run
+    /// This function is called after every tick - avoid long-running tasks here
```
Lukasa, line 53: Let's be really clear in this API, add a big warning block that says something like: "This function is called after every event loop tick and on the event loop thread. Any non-trivial work in this function will block the event loop and cause latency increases and performance degradation."

Lukasa (commented):
Basically good, just a doc note.

Lukasa (approved):

Lukasa (changes requested):
Looks like you have a flaky test here:

```

Lukasa (approved):

PR #2613, opened 2023-12-19
Set `SWIFT_VERSION` environment variable to resolve to the correct benchmarks thresholds path

dnadoba:
CI started to fail because `package-benchmark` has started to fail if it doesn't find the given thresholds path folder. This is a good thing because this means we actually never enforced the thresholds on CI.

The PR sets the `SWIFT_VERSION` environment variable to the same value we use as well in `update-benchmark-baseline`.
This will then be used to create the correct thresholds folder path.


gjcairo (approved):

dnadoba:

dnadoba:
`TCPEchoAsyncChannel` is flaky on CI. I have disabled it for now. Needs further investigation.

Issue #1890, opened 2021-07-06
Refactor Channels and MultiThreadedEventLoopGroup for io_uring


Lukasa:
cc @hassila @weissi 

As part of the io_uring PR (#1788) we had a number of discussions about how we might need to evolve the NIO codebase to fully support io_uring, instead of using it in the "glorified epoll" mode we do today. I've thought on this some more and want to lay out a proposal for how we might evolve the current implementations to make it much easier to implement an io_uring based event loop, ideally without any API change at all and with minimal changes to the Channels themselves.

### Problem Statement

io_uring operates, at a fundamental level, by making system calls asynchronous. Each system call is enqueued in the submission queue in the form of an SQE. Some time later, the kernel will either poll for an entry or we will call `io_uring_enter` and the kernel will begin processing the submitted system calls. Finally, at yet another time in the future, we will read from the completion queue, which will notify us of some of the completed events.

The first and most obvious problem we have here is that this is not how the current NIO Channels operate. In NIO today, the Channels are responsible for making their own I/O system calls: they call `read`/`readmsg`/`readmmsg`/`write`/`writev`/`sendmsg`/`sendmmsg` themselves, in response to both user activity and the notifications from the selectable event loop. Additionally, they assume all system calls are synchronous and non-blocking, meaning that they can issue these system calls directly without blocking the event loop and will synchronously find out whether there is space in the buffer.

In io_uring, system calls are not synchronous, meaning they do not return right away. Indeed, we aren't making system calls at all, we're enqueuing I/O operations onto the submission queue. From time-to-time the event loop will want to make a system call, but those times are not directly correlated with when the Channel wants to make its system call. Additionally, they work on top of a limited shared resource: the submission queue. The submission queue is not arbitrarily large, and when it fills up we cannot submit further I/O work to it until the kernel has processed it. This means we need an efficient way to know what I/O operations have been "enqueued" without actually making it to the kernel, so we can process them as space appears in the submission queue.

### Proposed Solution

We had originally discussed implementing new Channels and new Event Loops in order to support the io_uring model of operation. I think I have a better option: let's make Event Loops do I/O, instead of Channels.

This change requires moving a bunch of work out of the Channel and into the EventLoop. We cannot simply move the I/O operations (i.e. change the `PendingWritesManager`s to call EventLoop functions instead of system calls), because the current pending writes manager backpressure implementation relies on having _synchronous_ feedback about write completion.

The result of this is that we need to perform multiple steps, roughly as follows:

[...]

Issue #1805, opened 2021-04-15
Refactor I/O memory ownership for asynchronous backends


hassila:
Trying to start a discussion and capture requirements / next steps, based on the experience from #1761 and related discussions, a prerequisite for supporting truly asynchronous backends like io_uring fully (i.e. using async send/recv) requires a refactoring.

To try to recap, the root issue is that memory ownership of the IO operations is currently owned by the channels, but this does not fit the model of e.g. io_uring where (simplified) we would register a set of memory buffers with the kernel, issue a bunch of read I/O directly for (e.g.) sockets, and when receiving the read result eventually, the kernel will tell us which buffer it put the result in. We then would re-register this buffer for reuse after we have processed the result. 

This means that the ownership of the read memory buffers would preferably be managed on e.g. an EL basis instead of per channel. 

Similarly, the lifespan and ownership for write buffers needs to be adjusted, as the kernel won't have copied the user payload until we get a asynchronous completion event (unlike now, when we know synchronously that the write has completed) - so a decoupling of the write operation and the reclaiming of that memory buffer is needed.

What we then ideally would have, is a new truly async EL, which will schedule all I/O directly for all channels depending on expressed interest (no more epoll/kqeue, no more changing of registrations externally towards the kernel - we can just choose when to schedule the actual read/write/accept). The EL would also need to notify channels when their I/O has completed for e.g write buffer management.

Lukasa:
This seems like an appropriate design. Right now this is already done in some cases. For example, we have a number of [buffers for vector operations](https://github.com/apple/swift-nio/blob/5b6b7f02d7a2caf12175c8734d5bd1a21e6f85c7/Sources/NIO/SelectableEventLoop.swift#L83-L96) stored on the selectable event loop.

The biggest limiting factor here is that channels currently have some degree of control over how they allocate memory for reads by way of the `RecvByteBufferAllocator`. Working out whether we can persist this design is going to be important, as it would be valuable if we can do it.

As for writes, we thankfully can do that already. `ByteBuffer` has `withVeryUnsafeBytesWithStorageManagement` that allows us to hold an opaque object that will manage the lifetime of the buffer. This will allow us to "pin" the buffer in memory until the write completes. This interacts nicely with CoW and every other property of `ByteBuffer`, so we can use that effectively.

hassila:
[...]

PR #2615, opened 2023-12-21
Add NIOFilesystem

glbrntt:
Motivation:

I/O is typically blocking, but options for interacting with the
filesystem using `async` APIs are limited. This change adds a new module
to NIO, `NIOFileSystem`. At the moment this is exposed as the
`_NIOFileSystem` product while users get a chance to try it out and
provide feedback.

During this time the API of `NIOFileSystem` is subject to change.
Breaking changes will come in _minor_ NIO releases and additive changes
may come in _patch_ releases of NIO.

Modifications:

- Add `NIOFileSystem`, `NIOFileSystemFoundationCompat`,
  `NIOFileSystemTests`, `NIOFileSystemIntegrationTests`, and
  `NIOFileSystemFoundationCompatTests`.
- Update docs
- Update CNIOLinux and CNIODarwin shims

Result:

File system can be accessed via `async` APIs


glbrntt:
This change is split into two commits:
- Update C shims, docs, etc: https://github.com/apple/swift-nio/pull/2615/commits/6e9ed64f6c2284d2c38f3630fb992909b7e96ccd
- Add filesystem: https://github.com/apple/swift-nio/pull/2615/commits/589c5c450359a557757380067647cde1878b854d

adam-fowler (commented):
From an initial look this is really cool. So far I've used this to read, write files and read chunks and write buffered files. I've included a few comments below.

I did have one crash. `deinit` was run on a lock that still had a count. I cannot be sure if it was my code or the file system code though. It seemed to be related to a task using a readChunks AsyncSequence being cancelled. I cannot replicate it though

hassila:
Very cool. Haven't had time to dig in looking at the implementation yet, but just wanted to check if the design has a memory ownership model suitable for (and is considering...) true async backends (ie. Io_uring) - would make sense to use that if available (even if not part of the initial implementation, it would be nice to be able to add it in later without too much trouble).

adam-fowler:
One thing I've noticed is GH actions running on a macOS 13 runner fail to compile because they can't find the FTS symbols

Sources/NIOFileSystem/Docs.docc/NIOFileSystem.md
```diff
@@ -0,0 +1,174 @@
...
+}
+
+// Let's find out what's in that file.
+do {
+  // Reading a whole file requires a limit. If the file is larger than the limit
+  // then an error is thrown. This avoids accidentally consuming too much memory
+  // if the file is larger than expected.
+  let plan = try await Array<UInt8>(
+    contentsOf: "/Users/hal9000/demise-of-dave.txt",
+    maximumBytesAllowed: 1024 * 1024
```
FranzBusch, line 47: Could we use the `ByteCount` type here instead to showcase it?

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -0,0 +1,1442 @@
...
+/// Errors thrown by ``FileSystem`` will be of type:
+/// - ``FileSystemError`` if it wasn't possible to complete the operation, or
+/// - `CancellationError` if the `Task` was cancelled.
+///
+/// ``FileSystemError`` implements `CustomStringConvertible`. The output from the `description`
+/// contains basic information including the error code, message and underlying error. You can get
+/// more information about the error by calling ``FileSystemError/detailedDescription()`` which
+/// returns a structured multi-line string containing information about the error.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public final class FileSystem: Sendable, FileSystemProtocol {
```
FranzBusch, line 48: Does this actually need to be a `class`?

FranzBusch (commented):
Great stuff! I think this PR is already using `swift-format` right? Should we setup format checking for just the new modules as a way to get us started with adopting `swift-format` and avoiding format divergence in those new modules?

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -0,0 +1,1442 @@
...
+    /// setting the `NIO_SINGLETON_FILESYSTEM_THREAD_COUNT` environment variable.
+    public static var fileSystem: FileSystem { globalFileSystem }
+}
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension FileSystemProtocol where Self == FileSystem {
+    /// A global shared instance of ``FileSystem``.
+    ///
+    /// See also ``FileSystem/shared``.
+    public static var shared: FileSystem {
```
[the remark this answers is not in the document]
glbrntt: It seems the compiler isn't able to see the `@discardableResult` in some contexts and drop the result in favour of `Void`.

Given:

```swift
@discardableResult
func discardableInt() -> Int { 42 }

func withResultOfClosure<R>(_ fn: () throws -> R) rethrows -> R {
    try fn()
}

func withVoidResult(_ fn: () throws -> Void) rethrows {
    try fn()
}
```

This warns:

```swift
func fn() {
    withResultOfClosure { // WARN: Result of call to 'withResultOfClosure' is unused
        discardableInt()
    }
}
```

Adding an explicit `return` suppresses the warning:

```swift
func fn() {
    return withResultOfClosure {
        discardableInt()
    }

    // Also suppresses the warning:
    //
    // withResultOfClosure {
    //     discardableInt()
    // } as Void
}
```

Wrapping the closure in a closure which must return `Void` (more or less what you've run into) emits an error:

```swift
func fn() {
    withVoidResult { // ERROR: Type of expression is ambiguous without a type annotation
        withResultOfClosure {
            discardableInt()
        }
    }
}
```

Adding the explicit `return` 'fixes' it:
```swift
func fn() {
    withVoidResult {
        return withResultOfClosure {
            discardableInt()
        }
    }
}
```

So I agree this isn't a great situation _but_ I think this is on the compiler to improve. Having the `shared` instance is the right thing to do as is returning the number of bytes written as a discardable result. I'm open to workarounds if you have suggestions though!

Sources/NIOFileSystem/ByteBuffer+Convenience.swift
```diff
@@ -0,0 +1,53 @@
...
+        contentsOf path: FilePath,
+        maximumSizeAllowed: ByteCount
+    ) async throws {
+        self = try await Self(
+            contentsOf: path,
+            maximumSizeAllowed: maximumSizeAllowed,
+            fileSystem: .shared
+        )
+    }
+}
```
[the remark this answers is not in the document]
glbrntt: Agreed. There are extensions on `Sequence<UInt8>` and by extension `ByteBufferView` but I think it's worth adding shims on `ByteBuffer` which call these.

glbrntt:
> One thing I've noticed is GH actions running on a macOS 13 runner fail to compile because they can't find the FTS symbols

Thanks for highlighting this, will investigate.

glbrntt:
> Great stuff! I think this PR is already using `swift-format` right? Should we setup format checking for just the new modules as a way to get us started with adopting `swift-format` and avoiding format divergence in those new modules?

It was, yes. We can but if we do that it should be in a separate PR.

Sources/NIOFileSystem/Docs.docc/NIOFileSystem.md
glbrntt: Yeah good catch, this API no longer exists.

adam-fowler:
> > I did have one crash. `deinit` was run on a lock that still had a count. I cannot be sure if it was my code or the file system code though. It seemed to be related to a task using a readChunks AsyncSequence being cancelled. I cannot replicate it though
> 
> Interesting, could you share what the calling code looking like?

Err that was last year 😬 . It's in one of the Soto multipart S3 upload/download functions. They are pretty complex. I just had a quick look through the code and I found an unstructured Task in one of the download functions (That section was written in the swift 5.5 anything goes days) it isn't exactly in the file saving part but means the code is probably untrustworthy. I'll rewrite it, I was meaning to anyway, and see if I can replicate after that.

glbrntt:
> Very cool. Haven't had time to dig in looking at the implementation yet, but just wanted to check if the design has a memory ownership model suitable for (and is considering...) true async backends (ie. Io_uring) - would make sense to use that if available (even if not part of the initial implementation, it would be nice to be able to add it in later without too much trouble).

There is API space for different backend implementations but I don't have enough experience with io_uring to know if or how well it'd fit. It'd be great to have support but it isn't a priority for us and is unlikely to be one any time soon.

hassila:
> > Very cool. Haven't had time to dig in looking at the implementation yet, but just wanted to check if the design has a memory ownership model suitable for (and is considering...) true async backends (ie. Io_uring) - would make sense to use that if available (even if not part of the initial implementation, it would be nice to be able to add it in later without too much trouble).
> 
> There is API space for different backend implementations but I don't have enough experience with io_uring to know if or how well it'd fit. It'd be great to have support but it isn't a priority for us and is unlikely to be one any time soon.

It is a true async syscall interface with basically two ring buffers on in each direction, we can pray that we will get something similar on Darwin in the future too as it's really a great step forward on many fronts - we are working on getting backend support for io_uring for the networking part in https://github.com/apple/swift-nio/pull/2357 and there were some discussions in https://github.com/apple/swift-nio/issues/1890 and https://github.com/apple/swift-nio/issues/1805 with regard to memory ownership in that context. 

The major issue in general is typically memory ownership of buffers and the assumption that syscalls are synchronous in nature, rather than allowing for a completion event to finish the syscall - designing api surfaces so that it's possible to write a backend which is async in nature would be nice future proofing - but now, I still haven't had time to go through the API in detail, just wanted to share the links above as they might be relevant.

glbrntt:
> One thing I've noticed is GH actions running on a macOS 13 runner fail to compile because they can't find the FTS symbols

@adam-fowler FYI this should be fixed in https://github.com/apple/swift-nio/pull/2615/commits/52a0091d9f51467cf174c0d5ed48cca93db9361b

adam-fowler:
> @adam-fowler FYI this should be fixed in https://github.com/apple/swift-nio/pull/2615/commits/52a0091d9f51467cf174c0d5ed48cca93db9361b

I reran the failing job and it worked 👍

Sources/NIOFileSystem/FileSystem.swift
FranzBusch: @glbrntt you never replied here

glbrntt: Oh sorry, I missed this one. Just changed it.

FranzBusch (approved):
LGTM!

glbrntt:
Credit also goes to @stefanadranca and @gjcairo who contributed to this work.

Sources/CNIOLinux/shim.c
```diff
@@ -188,4 +188,17 @@ int CNIOLinux_system_info(struct utsname* uname_data) {
 
 const unsigned long CNIOLinux_IOCTL_VM_SOCKETS_GET_LOCAL_CID = IOCTL_VM_SOCKETS_GET_LOCAL_CID;
 
+const char* CNIOLinux_dirent_dname(struct dirent* ent) {
+  return ent->d_name;
+}
+
+int CNIOLinux_renameat2(int oldfd, const char* old, int newfd, const char* newName, unsigned int flags) {
+  return renameat2(oldfd, old, newfd, newName, flags);
```
finagolfin, line 196: This [broke my Android CI today](https://github.com/finagolfin/swift-android-sdk/actions/runs/7553358697/job/20564040487#step:13:71) because Android didn't add this to [its Bionic libc till API 30 a couple years ago](https://android.googlesource.com/platform/bionic/+/9c06d16ca3d230059ccf3b7f270574bc5e8fcfcb/libc/include/stdio.h#199), whereas I build against the older API 24.

Since this approach is still experimental, can we just disable `rename()` alone on Android or something like that?

glbrntt: This is also used during file creation so we can't just disable it on Android. I think we can work around this though as this also isn't available on Darwin so I think we can just do what we're doing on Darwin.

glbrntt: See https://github.com/apple/swift-nio/pull/2627

PR #2616, opened 2023-12-25
Handle bind failures when running NIO in a tight sandbox 

### LINKED ISSUE #2465: BaseSocketChannel: assert(self.lifecycleManager.isActive) hit when running NIO tests in sandbox
I hit this assertion

https://github.com/apple/swift-nio/blob/324bc65a28323660fad8a36a7a37f0c2c78eeb9a/Sources/NIOPosix/BaseSocketChannel.swift#L1077

when running the NIO tests in a tight sandbox. Output

```
     Test Suite 'All tests' started at 2023-07-11 05:46:11.149.
     Test Suite 'swift-nioPackageTests.xctest' started at 2023-07-11 05:46:11.153.
     Test Suite 'AcceptBackoffHandlerTest' started at 2023-07-11 05:46:11.154.
     Test Case '-[NIOPosixTests.AcceptBackoffHandlerTest testChannelInactiveCancelScheduled]' started.
     swift-nio/Tests/NIOPosixTests/AcceptBackoffHandlerTest.swift:268: error: -[NIOPosixTests.AcceptBackoffHandlerTest testChannelInactiveCancelScheduled] : XCTAssertNoThrow failed: threw error "bind(descriptor:ptr:bytes:): Operation not permitted (errno: 1)"
     NIOPosix/BaseSocketChannel.swift:1077: Assertion failed
```

The sandbox disallows `bind` and other network related stuff. So it's expected that the tests fail but I wouldn't expect an assertion to be hit.

This can probably be tested with the SAL (syscall abstraction layer) because that should make it easy to make `bind` fail with `EPERM` (== `1`).

dkz2:
_Added error handling for failed bind operations in SwiftNIO's `SocketChannel`_

### Motivation:

_When running SwiftNIO tests in a restricted environment, an assertion error was encountered when bind and other network related operations were disallowed. The tests were expected to fail in such environment, but hitting an assertion was not expected. This change aims to handle such failure gracefully. Notably, the `testChannelInactiveCancelScheduled` in `AcceptBackoffHandlerTest` was failing as expected, but it was also hitting an assertion, which was not the intended behavior. This PR closes #2465._

### Modifications:

* _Modified `bind0` in `SocketChannel` to catch and handle errors when trying to bind._
* _Added a new error case `bindFailed` to `ChannelError` in `Channel` to represent a failed bind operation._
* _Created a new test in `testBindFailureClosesChannel` in `SALChannelTests` to simulate and test the behavior when a bind operation fails._
* _Added a new function `assertBindFailure` in `SyscallAbstractionLayer` to assert that a bind operation fails as expected._

### Result:

_After these changes, SwiftNIO will handle bind failures more gracefully, if a bind operation fails, it will not cause an assertion to be hit, but instead throw a ChannelError.bindFailed error. This change improves the robustness of the code and makes the error handling more informative._

### DIFF
--- a/Sources/NIOPosix/SocketChannel.swift
+++ b/Sources/NIOPosix/SocketChannel.swift
@@ -266,14 +266,18 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
             promise?.fail(error)
         }
         executeAndComplete(p) {
-            switch target {
-            case .socketAddress(let address):
-                try socket.bind(to: address)
-            case .vsockAddress(let address):
-                try socket.bind(to: address)
+            do {
+                switch target {
+                case .socketAddress(let address):
+                    try socket.bind(to: address)
+                case .vsockAddress(let address):
+                    try socket.bind(to: address)
+                }
+                self.updateCachedAddressesFromSocket(updateRemote: false)
+                try self.socket.listen(backlog: backlog)
+            } catch {
+                promise?.fail(ChannelError.bindFailed)
             }
-            self.updateCachedAddressesFromSocket(updateRemote: false)
-            try self.socket.listen(backlog: backlog)
         }
     }
 
--- a/Tests/NIOHTTPTests/HTTPDecoderLengthTest.swift
+++ b/Tests/NIOHTTPTests/HTTPDecoderLengthTest.swift
@@ -1,0 +1,556 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIOCore
+import NIOEmbedded
+import NIOHTTP1
+
+private class MessageEndHandler<Head: Equatable, Body: Equatable>: ChannelInboundHandler {
+    typealias InboundIn = HTTPPart<Head, Body>
+
+    var seenEnd = false
+    var seenBody = false
+    var seenHead = false
+
+    func channelRead(context: ChannelHandlerContext, data: NIOAny) {
+        switch self.unwrapInboundIn(data) {
+        case .head:
+            XCTAssertFalse(self.seenHead)
+            self.seenHead = true
+        case .body:
+            XCTAssertFalse(self.seenBody)
+            self.seenBody = true
+        case .end:
+            XCTAssertFalse(self.seenEnd)
+            self.seenEnd = true
+        }
+    }
+}
+
+/// Tests for the HTTP decoder's handling of message body framing.
+///
+/// Mostly tests assertions in [RFC 7230 § 3.3.3](https://tools.ietf.org/html/rfc7230#section-3.3.3).
+class HTTPDecoderLengthTest: XCTestCase {
+    private var channel: EmbeddedChannel!
+    private var loop: EmbeddedEventLoop {
+        return self.channel.embeddedEventLoop
+    }
+
+    override func setUp() {
+        self.channel = EmbeddedChannel()
+    }
+
+    override func tearDown() {
+        XCTAssertNoThrow(try self.channel?.finish(acceptAlreadyClosed: true))
+        self.channel = nil
+    }
+
+    /// The mechanism by which EOF is being sent.
+    enum EOFMechanism {
+        case channelInactive
+        case halfClosure
+    }
+
+    /// The various header fields that can be used to frame a response.
+    enum FramingField {
+        case contentLength
+        case transferEncoding
+        case neither
+    }
+
+    private func assertSemanticEOFOnChannelInactiveResponse(version: HTTPVersion, how eofMechanism: EOFMechanism) throws {
+        print("yes")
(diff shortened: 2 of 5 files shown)


Sources/NIOPosix/SocketChannel.swift
```diff
@@ -266,14 +266,18 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
...
+                switch target {
+                case .socketAddress(let address):
+                    try socket.bind(to: address)
+                case .vsockAddress(let address):
+                    try socket.bind(to: address)
+                }
+                self.updateCachedAddressesFromSocket(updateRemote: false)
+                try self.socket.listen(backlog: backlog)
+            } catch {
+                promise?.fail(ChannelError.bindFailed)
```
FranzBusch, line 279: We should definitely still fail the promise with the error produced by the `socket.bind` class. Otherwise it is hard to tell what went wrong. Could you explain why this approach works around the assert? I would like to understand how we actually ended up in the assert in the first place since something must have caused the call to `readable` and we just have to avoid that.

dkz2: after re thinking my approach the way I set up my sandbox to test this was not correct, thus nor is this solution. In the actual `testChannelInactiveCancelScheduled` I set up a class `MockServerSocket: ServerSocket {}` that essentially always simulate bind failure. However I did this by overriding bind but by doing that I am not hitting all of NIOs binding process. 

Now I am overriding the standard bind sys call thats called by NIO by creating a bind function that always returns `EPERM`

```
int bind(int sockfd, const struct sockaddr *addr, socklen_t addrlen) {
    int (*original_bind)(int, const struct sockaddr *, socklen_t);
    original_bind = dlsym(RTLD_NEXT, "bind");
    errno = EPERM;
    return -1;
}
```

compile it down to a shared object and preload it and run the test. Going to use this approach to create a new solution here. 

In the actual `testChannelInactiveCancelScheduled` we call `readable` since we want to test whether scheduled reads are closed when the channel is closed. The issue here is `readable` checks if the `lifecycleManager` is `active`, however at this stage it's only transitioned from `fresh` to `preRegistered` thus triggering the assert.

Tests/NIOHTTPTests/HTTPDecoderLengthTest.swift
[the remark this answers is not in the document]
dkz2: correct not related.

Sources/NIOPosix/SocketChannel.swift
```diff
@@ -263,6 +263,7 @@ final class ServerSocketChannel: BaseSocketChannel<ServerSocket> {
             // It's important to call the methods before we actually notify the original promise for ordering reasons.
             self.becomeActive0(promise: promise)
         }.whenFailure{ error in
+            self.close0(error: error, mode: .all, promise: nil)
```
dkz2, line 266: Upon bind failing, the `lifecycleManager` seems to remain in the `preRegistered` state. thus we are transiting to the `closed` state here.

Tests/NIOPosixTests/SyscallAbstractionLayer.swift
```diff
@@ -884,7 +884,18 @@ extension SALTest {
             }
         }
     }
-
+    
+    func assertBindFailure(expectedAddress: SocketAddress, file: StaticString = #filePath, line: UInt = #line) throws {
```
Lukasa, line 888: Can we fold this into the existing `assertBind` call, with an optional `errorReturn` parameter that defaults to `nil`? That's a little closer to how we typically structure things in the SAL.

Tests/NIOHTTPTests/HTTPDecoderLengthTest.swift
```diff
@@ -0,0 +1,556 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import NIOCore
+import NIOEmbedded
+import NIOHTTP1
+
+private class MessageEndHandler<Head: Equatable, Body: Equatable>: ChannelInboundHandler {
```
Lukasa, line 20: Presumably you don't actually want to add this.

dkz2: removed!

Tests/NIOPosixTests/SyscallAbstractionLayer.swift
dkz2: done!

FranzBusch (approved):
LGTM! I would like @Lukasa to have one final looks as well though

Tests/NIOPosixTests/SALChannelTests.swift
```diff
@@ -322,6 +322,48 @@ final class SALChannelTest: XCTestCase, SALTest {
                 }
         }.salWait())
     }
+    
+    func testBindFailureClosesChannel() {
```
Lukasa, line 326: This test isn't specific enough: it continues to pass even with the new `close0` line removed. It looks like this is because the failed `bind` promise is caught by the `ClientBootstrap` in the `connect` flow and triggers a `close0`.

The same code appears to be present in the server bootstrap. Can we try to nail down where the `bind` issue came from in your original reproducer? It doesn't currently look like `bind0` needs to close where it does.

Lukasa (changes requested):
See comment

Tests/NIOPosixTests/SALChannelTests.swift
dkz2: I see, I will be taking a closer look at this over the weekend!

weissi:
@dkz2 are you still interested in pushing this over the line?

Issue #2223, opened 2022-07-06
Xcode 14 Thread Performance Checker Warning


Sherlouk:
### Expected behavior

When I create a new `MultiThreadedEventLoopGroup`, I should not receive a performance warning from Xcode.

### Actual behavior

When I create a new `MultiThreadedEventLoopGroup`, Xcode's new Thread Performance Checker throws the warning below. It does not appear to impede functionality. The thread checker can be turned off to silence this warning but I feel an issue here is best served to look to address the potential underlying issue.

```
Thread Performance Checker: Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_DEFAULT. Investigate ways to avoid priority inversions
```

<details>
  <summary>Full Backtrace</summary>
  
```
Thread Performance Checker: Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_DEFAULT. Investigate ways to avoid priority inversions
PID: 71975, TID: 24582681
Backtrace
=================================================================
3   Application                           0x0000000104b9df5a $s8NIOPosix27MultiThreadedEventLoopGroupC014setupThreadAnddE033_C2B1528F4FBA68A3DBFA89DBAEBE9D4DLL4name06parentF015selectorFactory11initializerAA010SelectabledE0CSS_AcA8SelectorCyAA15NIORegistrationVGyKcyAA9NIOThreadCctFZ + 586
4   Application                           0x0000000104b9ecba $s8NIOPosix27MultiThreadedEventLoopGroupC18threadInitializers15selectorFactoryACSayyAA9NIOThreadCcG_AA8SelectorCyAA15NIORegistrationVGyKctcfcAA010SelectabledE0CyAGcXEfU_ + 650
5   Application                           0x0000000104b9ed47 $s8NIOPosix27MultiThreadedEventLoopGroupC18threadInitializers15selectorFactoryACSayyAA9NIOThreadCcG_AA8SelectorCyAA15NIORegistrationVGyKctcfcAA010SelectabledE0CyAGcXEfU_TA + 39
6   Application                           0x0000000104b9edbe $s8NIOPosix9NIOThreadCIegg_AA19SelectableEventLoopCs5Error_pIggozo_xq_r0_lyACytIsegnr_AEsAF_pIegnrzo_TR + 110
[...]

PR #2618, opened 2024-01-02
Support asynchronous MTELG construction

Lukasa:
Motivation

Currently when a MultithreadedEventLoopGroup is constructed we block the calling thread until all the background threads launch. Generally this is ok (we don't expect it to take long), but it can cause issues when the launching thread is high priority, as we will wait for those low priority threads to get time to execute.

Additionally, this causes nuisance warnings for users: see https://github.com/apple/swift-nio/issues/2223 for more.

To make this safe, we need to make it possible to construct a SelectableEventLoop instance before the owning thread has launched. This requires us to make the existing strong stored `thread` property into something that can be lazily initialized.

We need this property in order to correctly implement `inEventLoop`. The good news is that so long as the first thing the background thread does is initialize the property, we can draw a confident signal from this value being `nil`: we know that we aren't on the EL thread. After all, if we were, we would have already initialized that field!

This calls for your friend and mine, `ManagedAtomicLazyReference`.

Modifications

- Move the stored Thread object for SelectableEventLoop and Selector into a `ManagedAtomicLazyReference`.
- Provide convenience accessors that force-unwrap the EL.
- Fix a few specific use-cases that may actually run before the EL is provided.
- Update a couple of tests that can observe this change.

Result

We can now construct MTLEG without waiting for the background threads to spawn, and everything is fully functional.

### DIFF
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -68,7 +68,21 @@ internal final class SelectableEventLoop: EventLoop {
     }
 
     /* private but tests */ internal let _selector: NIOPosix.Selector<NIORegistration>
+
+    /// The thread on which this EL is running.
+    ///
+    /// This is a lazy reference because we may create the SelectableEventLoop before we have
+    /// started the thread on which it runs.
+    ///
+    /// This will be initialized when the EL thread starts (by a call to `launchedThread`).
+    /// It is always initialized by the EL thread, as the first thing the thread does before it
+    /// starts actually running the loop.
+    ///
+    /// This is acceptable because we only use this to detect whether we are "on" the EL. As we
+    /// can only be on the EL when we are on the thread, we know that if this value hasn't yet
+    /// been initialized then we cannot possibly be on the EL thread.
     private let thread: ManagedAtomicLazyReference<NIOThread>
+
     @usableFromInline
     // _pendingTaskPop is set to `true` if the event loop is about to pop tasks off the task queue.
     // This may only be read/written while holding the _tasksLock.
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -68,7 +68,7 @@ internal final class SelectableEventLoop: EventLoop {
     }
 
     /* private but tests */ internal let _selector: NIOPosix.Selector<NIORegistration>
-    private let thread: NIOThread
+    private let thread: ManagedAtomicLazyReference<NIOThread>
     @usableFromInline
     // _pendingTaskPop is set to `true` if the event loop is about to pop tasks off the task queue.
     // This may only be read/written while holding the _tasksLock.
@@ -82,11 +82,7 @@ internal final class SelectableEventLoop: EventLoop {
     // for every appended closure. https://bugs.swift.org/browse/SR-15872
     private var tasksCopy = ContiguousArray<ScheduledTask>()
     @usableFromInline
-    internal var _succeededVoidFuture: Optional<EventLoopFuture<Void>> = nil {
-        didSet {
-            self.assertInEventLoop()
-        }
-    }
+    internal var _succeededVoidPromise: Optional<EventLoopPromise<Void>> = nil
 
     private let canBeShutdownIndividually: Bool
     @usableFromInline
@@ -173,13 +169,12 @@ Further information:
         }
     }
 
-    internal init(thread: NIOThread,
-                  parentGroup: MultiThreadedEventLoopGroup?, /* nil iff thread take-over */
+    internal init(parentGroup: MultiThreadedEventLoopGroup?, /* nil iff thread take-over */
                   selector: NIOPosix.Selector<NIORegistration>,
                   canBeShutdownIndividually: Bool) {
         self._parentGroup = parentGroup
         self._selector = selector
-        self.thread = thread
+        self.thread = .init()
         self.bufferPool = Pool<PooledBuffer>(maxSize: 16)
         self.msgBufferPool = Pool<PooledMsgBuffer>(maxSize: 16)
         // We will process 4096 tasks per while loop.
@@ -188,9 +183,17 @@ Further information:
         // note: We are creating a reference cycle here that we'll break when shutting the SelectableEventLoop down.
         // note: We have to create the promise and complete it because otherwise we'll hit a loop in `makeSucceededFuture`. This is
         //       fairly dumb, but it's the only option we have.
-        let voidPromise = self.makePromise(of: Void.self)
-        voidPromise.succeed(())
-        self._succeededVoidFuture = voidPromise.futureResult
+        self._succeededVoidPromise = self.makePromise(of: Void.self)
+    }
+
+    internal func launchedThread(_ thread: NIOThread) {
+        precondition(NIOThread.current == thread)
+        let original = self.thread.storeIfNilThenLoad(thread)
+        precondition(original == thread)
+        self._selector.threadLaunched(thread)
+
+        // Force-unwrap is safe, can only be nil if we have called deinit
+        self._succeededVoidPromise!.succeed()
     }
 
     deinit {
@@ -246,7 +249,7 @@ Further information:
     /// - see: `EventLoop.inEventLoop`
     @usableFromInline
     internal var inEventLoop: Bool {
-        return thread.isCurrent
+        return self.thread.load()?.isCurrent ?? false
     }
 
     /// - see: `EventLoop.scheduleTask(deadline:_:)`
@@ -553,7 +556,7 @@ Further information:
         try self._selector.close()
 
         // This breaks the retain cycle created in `init`.
-        self._succeededVoidFuture = nil
+        self._succeededVoidPromise = nil
     }
 
     internal func initiateClose(queue: DispatchQueue, completionHandler: @escaping (Result<Void, Error>) -> Void) {
@@ -626,7 +629,8 @@ Further information:
             return
         }
         if joinThread {
-            self.thread.join()
+            // Force-unwrap is safe: we cannot join the thread until the thread has started, at which point this cannot be nil.
+            self.thread.load()!.join()
         }
         self.externalStateLock.withLock {
             precondition(self.externalState == .reclaimingResources)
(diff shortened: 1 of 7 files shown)


Lukasa:

Lukasa:
@swift-server-bot test performance please

Lukasa:

Lukasa:

Lukasa:
Those results don't show much of a change in EL-bound benchmarks. The biggest difference is in the task scheduling benchmarks, which makes some sense as they hit the now slightly slower path. The I/O benchmarks are unaffected. Looks like a worst-case 2% slowdown (the 8% extra slowdown for actually executing the tasks seems unlikely to be relevant).

I've also run this locally, as we won't see any extra pain from the cache coherency algorithm in the above benchmark numbers. This is because the above benchmark runs on Intel, and on that platform acquiring loads aren't different to regular loads. My Apple Silicon mac gives me these results:

| name | current | previous | winner | diff |
|--|--|--|--|--|
| write_http_headers | 0.006830917 | 0.006834458 | current | 0% |
| http_headers_canonical_form | 0.06431075 | 0.0653365 | current | -1% |
| http_headers_canonical_form_trimming_whitespace | 0.011751084 | 0.011801125 | current | 0% |
| http_headers_canonical_form_trimming_whitespace_from_short_string | 0.01089575 | 0.01097325 | current | 0% |
| http_headers_canonical_form_trimming_whitespace_from_long_string | 0.016938458 | 0.017022208 | current | 0% |
| bytebuffer_write_12MB_short_string_literals | 0.06002675 | 0.060073334 | current | 0% |
| bytebuffer_write_12MB_short_calculated_strings | 0.044949875 | 0.045041292 | current | 0% |
| bytebuffer_write_12MB_medium_string_literals | 0.368481459 | 0.36521975 | previous | 0% |
| bytebuffer_write_12MB_medium_calculated_strings | 0.045406459 | 0.045337125 | previous | 0% |
| bytebuffer_write_12MB_large_calculated_strings | 0.044483334 | 0.044534666 | current | 0% |
| bytebuffer_lots_of_rw | 0.022767833 | 0.022531792 | previous | 1% |
| bytebuffer_write_http_response_ascii_only_as_string | 0.016077458 | 0.016036625 | previous | 0% |
| bytebuffer_write_http_response_ascii_only_as_staticstring | 0.015089708 | 0.015048541 | previous | 0% |
| bytebuffer_write_http_response_some_nonascii_as_string | 0.016044292 | 0.016035541 | previous | 0% |
| bytebuffer_write_http_response_some_nonascii_as_staticstring | 0.01500325 | 0.015001625 | previous | 0% |
| no-net_http1_1k_reqs_1_conn | 0.006101541 | 0.006090625 | previous | 0% |
| http1_1k_reqs_1_conn | 0.031084709 | 0.030956 | previous | 0% |
| http1_1k_reqs_100_conns | 0.045975542 | 0.045737666 | previous | 0% |
| future_whenallsucceed_100k_immediately_succeeded_off_loop | 0.054005792 | 0.053302834 | previous | 1% |
| future_whenallsucceed_100k_immediately_succeeded_on_loop | 0.063516917 | 0.063262708 | previous | 0% |
| future_whenallsucceed_10k_deferred_off_loop | 0.019840292 | 0.019439458 | previous | 2% |
| future_whenallsucceed_10k_deferred_on_loop | 0.009708 | 0.009349708 | previous | 3% |
| future_whenallcomplete_100k_immediately_succeeded_off_loop | 0.013893292 | 0.0130675 | previous | 6% |
| future_whenallcomplete_100k_immediately_succeeded_on_loop | 0.023964833 | 0.023060333 | previous | 3% |
| future_whenallcomplete_10k_deferred_off_loop | 0.01273325 | 0.0128225 | current | 0% |
| future_whenallcomplete_100k_deferred_on_loop | 0.033297333 | 0.031608875 | previous | 5% |
| future_reduce_10k_futures | 0.009584083 | 0.009297333 | previous | 3% |
| future_reduce_into_10k_futures | 0.008435417 | 0.008121959 | previous | 3% |
| channel_pipeline_1m_events | 0.042228292 | 0.04227225 | current | 0% |
| websocket_encode_50b_space_at_front_100k_frames_cow | 0.018007667 | 0.018061208 | current | 0% |
| websocket_encode_50b_space_at_front_1m_frames_cow_masking | 0.343775875 | 0.344408042 | current | 0% |
| websocket_encode_1kb_space_at_front_1m_frames_cow | 0.190924542 | 0.19154225 | current | 0% |
| websocket_encode_50b_no_space_at_front_100k_frames_cow | 0.018032833 | 0.018092542 | current | 0% |
| websocket_encode_1kb_no_space_at_front_100k_frames_cow | 0.019067958 | 0.019155083 | current | 0% |
| websocket_encode_50b_space_at_front_100k_frames | 0.029710791 | 0.031144208 | current | -4% |
| websocket_encode_50b_space_at_front_10k_frames_masking | 0.004548 | 0.004681958 | current | -2% |
| websocket_encode_1kb_space_at_front_10k_frames | 0.00359175 | 0.003707833 | current | -3% |
| websocket_encode_50b_no_space_at_front_100k_frames | 0.03074775 | 0.031406333 | current | -2% |
| websocket_encode_1kb_no_space_at_front_10k_frames | 0.003477708 | 0.00371075 | current | -6% |
| websocket_decode_125b_10k_frames | 0.006691459 | 0.00666475 | previous | 0% |
| websocket_decode_125b_with_a_masking_key_10k_frames | 0.006872708 | 0.006823125 | previous | 0% |
| websocket_decode_64kb_10k_frames | 0.006888583 | 0.006823416 | previous | 0% |
| websocket_decode_64kb_with_a_masking_key_10k_frames | 0.006961542 | 0.006998959 | current | 0% |
| websocket_decode_64kb_+1_10k_frames | 0.006845792 | 0.006830583 | previous | 0% |
| websocket_decode_64kb_+1_with_a_masking_key_10k_frames | 0.007061917 | 0.007022959 | previous | 0% |
| circular_buffer_into_byte_buffer_1kb | 0.016782875 | 0.016785125 | current | 0% |
| circular_buffer_into_byte_buffer_1mb | 0.032569084 | 0.032592125 | current | 0% |
| byte_buffer_view_iterator_1mb | 0.008687167 | 0.008728875 | current | 0% |
| byte_buffer_view_contains_12mb | 0.019686958 | 0.019703791 | current | 0% |
| byte_to_message_decoder_decode_many_small | 0.023601625 | 0.023338666 | previous | 1% |
| generate_10k_random_request_keys | 0.004252625 | 0.004171584 | previous | 1% |
| bytebuffer_rw_10_uint32s | 0.020738417 | 0.020741958 | current | 0% |
| bytebuffer_multi_rw_10_uint32s | 0.040607292 | 0.040121625 | previous | 1% |
| lock_1_thread_10M_ops | 0.085068291 | 0.085016041 | previous | 0% |
| lock_2_threads_10M_ops | 0.171361792 | 0.17232825 | current | 0% |
| lock_4_threads_10M_ops | 0.204662292 | 0.138865333 | previous | 47% |
| lock_8_threads_10M_ops | 0.224987459 | 0.213312708 | previous | 5% |
| schedule_100k_tasks | 0.027195583 | 0.0257005 | previous | 5% |
| schedule_and_run_100k_tasks | 0.175171375 | 0.173897875 | previous | 0% |
| execute_100k_tasks | 0.1243835 | 0.125171667 | current | 0% |
| bytebufferview_copy_to_array_100k_times_1kb | 0.005714041 | 0.005829834 | current | -1% |
| circularbuffer_copy_to_array_10k_times_1kb | 0.010163875 | 0.010160917 | previous | 0% |
| deadline_now_1M_times | 0.011881958 | 0.011854792 | previous | 0% |
| asyncwriter_single_writes_1M_times | 1.196198958 | 1.194668 | previous | 0% |
| asyncsequenceproducer_consume_1M_times | 0.697478583 | 0.700970667 | current | 0% |
| udp_10k_writes | 0.196309291 | 0.180699333 | previous | 8% |
| udp_10k_vector_writes | 0.085784584 | 0.086185125 | current | 0% |
| udp_10k_vector_reads | 0.193827584 | 0.180634875 | previous | 7% |
| udp_10k_vector_reads_and_writes | 0.086080375 | 0.08631675 | current | 0% |
| tcp_100k_messages_throughput | 0.259064833 | 0.258545416 | previous | 0% |


The numbers here are noisy, but mostly we seem to be in the 0% range, except for the occasional jump to 5% or so. Not something I'm very nervous about either.

Lukasa:
`14:04:05 Test Case 'EventLoopTest.testScheduledRepeatedAsyncTask' started at 2024-01-03 13:44Build timed out (after 20 minutes). Marking the build as failed.`

Not really sure how that test could possibly have hung. I'll dig in.

Lukasa:
Nope, absolutely nothing. Suspect the real hang was obfuscated by buffered test output. Re-testing.

Lukasa:

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -68,7 +68,7 @@ internal final class SelectableEventLoop: EventLoop {
     }
 
     /* private but tests */ internal let _selector: NIOPosix.Selector<NIORegistration>
-    private let thread: NIOThread
+    private let thread: ManagedAtomicLazyReference<NIOThread>
```
FranzBusch, line 71: Can we add a comment here why this is a `ManagedAtomicLazyReference` and how the threading is working.

Lukasa: Done, please re-review.

Lukasa:
Hmm, some of the EL tests are still hanging intermittently. Let me try to dig into this a bit.

Lukasa:
Yeah if I run the tests in a loop I can reproduce the hang. I have theories, I'll investigate.

Lukasa:
Oh that's fantastic: I inverted the logic of `wakeup0`. A writeup for the interested:

Here's the current (buggy) code (for both epoll and kqueue):

```swift
    func wakeup0() throws {
        guard let thread = self._myThread.load() else {
            // No need to wake up, the thread hasn't started yet. It'll run anyway.
            return
        }

        assert(NIOThread.current != thread)
        try self.externalSelectorFDLock.withLock {
                guard self.eventFD >= 0 else {
                    throw EventLoopError.shutdown
                }
                _ = try EventFd.eventfd_write(fd: self.eventFD, value: 1)
        }
    }
```

In this context, we expect that `wakeup0` is only called from off the EL thread. This assumption is encoded in the assertion.

Typically, in cases where the thread hasn't yet been initialized, I've taken that as saying that we are not on the EL thread. That would be totally appropriate here. However, in this exact case I tried to be too clever, and took the absence of the message to mean that the run loop hadn't started yet, and so we didn't need to wake the selector.

This is too clever by half, and forgot the most important rule of concurrent programming: only reason under the worst-case behaviour of the scheduler. In this case, that worst-case is that the thread calling `wakeup0` does the acquiring load of the thread ID and sees `nil`, then the scheduler immediately sleeps it. During that time the EL thread begins, and starts looping, getting to the bottom of the EL loop and sleeping in the selector. Then our first thread wakes and decides not to signal the selector.

This should be very easy to fix: just get the logic to match the rest of the code in these locations too.

FranzBusch (approved):
Nice!

Lukasa:
For now we're going to defer doing this: while it's arguably more architecturally appropriate, it's also riskier and has more performance implications than the cheeky workaround that I landed in #2620. For now I'll close this, but we can always revisit.

PR #2622, opened 2024-01-09
Fix `NIOAsyncChannel` allocation benchmarks

FranzBusch:
# Motivation
We had to disable the benchmarks since they regressed without us noticing and they appear to be flaky.

# Modification
This PR fixes the allocation regression and tries to re-enable them.

### DIFF
--- a/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
+++ b/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
@@ -39,28 +39,29 @@ let benchmarks = {
     // This benchmark is only available above 5.9 since our EL conformance
     // to serial executor is also gated behind 5.9.
     #if compiler(>=5.9)
-// In addition this benchmark currently doesn't produce deterministic results on our CI
-// and therefore is currently disabled
-//    Benchmark(
-//        "TCPEchoAsyncChannel",
-//        configuration: .init(
-//            metrics: defaultMetrics,
-//            timeUnits: .milliseconds,
-//            scalingFactor: .mega,
-//            setup: {
-//                swiftTaskEnqueueGlobalHook = { job, _ in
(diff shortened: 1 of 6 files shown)


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -293,7 +293,7 @@ public struct NIOAsyncWriter<
     /// - Parameter element: The element to yield.
     @inlinable
     public func yield(_ element: Element) async throws {
-        try await self._storage.yield(contentsOf: CollectionOfOne(element))
+        try await self._storage.yield(element: element)
```
FranzBusch, line 296: The fix here is quite simple. The problem was that we lost the fast path for single element writes in some refactoring. This PR duplicates the write path and specialises one of them for single element writes to call the optimised delegate method of the writer.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -39,28 +39,26 @@ let benchmarks = {
...
-//            numberOfWrites: benchmark.scaledIterations.upperBound,
-//            eventLoop: eventLoop
-//        )
-//    }
+    Benchmark(
+        "TCPEchoAsyncChannel",
+        configuration: .init(
+            metrics: defaultMetrics,
+            timeUnits: .milliseconds,
+            scalingFactor: .mega,
```
hassila, line 47: If you update to Benchmark `1.22.0` you can now add some slack to the p90 check for malloc counts, this should do the trick (allowing for a +-100 mallocs);

```suggestion
            scalingFactor: .mega,
            thresholds: [.mallocCountTotal : .init(absolute: [.p90 : 100])],
```

dnadoba (approved):

PR #2627, opened 2024-01-17
Avoid renameat2 on Android

glbrntt:
Motivation:

Until very recently Android didn't support the `renameat2` syscall, this means that on most systems NIO will fail to compile. `renameat2` is used to delay the materialization of files on Linux. On Darwin a different code path is used which we can make use of on Android too.

Modifications:

- Update a few if-defs to force Android onto the alternative path

Result:

NIO builds on Android

### DIFF
--- a/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
+++ b/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
@@ -308,6 +308,7 @@ public enum Libc {
         }
     }
 
+    #if !os(Android)
     static func constr(_ name: CInt) -> Result<String, Errno> {
         var buffer = [CInterop.PlatformChar](repeating: 0, count: 128)
 
@@ -331,6 +332,7 @@ public enum Libc {
             }
         } while true
     }
+    #endif
 
     static func ftsOpen(_ path: FilePath, options: FTSOpenOptions) -> Result<CInterop.FTSPointer, Errno> {
         // 'fts_open' needs an unsafe mutable pointer to the C-string, `FilePath` doesn't offer this
--- a/Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift
+++ b/Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift
@@ -384,6 +384,7 @@ internal func libc_getcwd(
 }
 
 /// confstr(3)
+#if !os(Android)
 internal func libc_confstr(
     _ name: CInt,
     _ buffer: UnsafeMutablePointer<CInterop.PlatformChar>,
@@ -391,6 +392,7 @@ internal func libc_confstr(
 ) -> Int {
     return confstr(name, buffer, size)
 }
+#endif
 
 /// fts(3)
 internal func libc_fts_open(


glbrntt:
@finagolfin could you verify this fixes your build?

finagolfin:
I'll apply it locally and let you know in a little bit, thanks.

finagolfin:
This isn't building either, some issues with optionals, will try modifying this patch and let you know within an hour.

glbrntt:
> This isn't building either, some issues with optionals, will try modifying this patch and let you know within an hour.

Suspect you'll need this commit too: https://github.com/apple/swift-nio/pull/2627/commits/600a21069a773d7401f830b9b6ef92f50caccec9

finagolfin:
Another problem is that Bionic doesn't have `confstr()` at all:
```
/home/swift-nio/Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift:389:12: error: cannot find 'confstr' in scope
    return confstr(name, buffer, size)                                                                                                                                        ^~~~~~~
```
I think this pull works to fix the `renameat2` issue, but then other compilation errors from your large pull pop up on Android, looking into them now.

glbrntt:
> Another problem is that Bionic doesn't have `confstr()` at all:
> 
> ```
> /home/swift-nio/Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift:389:12: error: cannot find 'confstr' in scope
>     return confstr(name, buffer, size)                                                                                                                                        ^~~~~~~
> ```

We can workaround `confstr` too as it looks like we're only using it on Darwin.

> I think this pull works to fix the `renameat2` issue, but then other compilation errors from your large pull pop up on Android, looking into them now.

Great, thanks for looking!

finagolfin:
I don't see other issues related to `renameat2`, so I think this pull works. If you can ifdef `confstr()` out on Android too, please go ahead and merge these changes.

I will work through the remaining compilation issues related to optionals, because the latest Android NDK added nullability annotations all over Bionic that your pull will have to be slightly modified for, and submit another pull for that later.

glbrntt:
> I don't see other issues related to `renameat2`, so I think this pull works. If you can ifdef `confstr()` out on Android too, please go ahead and merge these changes.

Great, will do!

> I will work through the remaining compilation issues related to optionals, because the latest Android NDK added nullability annotations all over Bionic that your pull will have to be slightly modified for, and submit another pull for that later.

Do you have any timeline on this? We'd like to tag a release and would rather not break you!

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -753,6 +747,12 @@ extension SystemFileHandle.SendableView {
                     relativeTo: .currentWorkingDirectory,
                     flags: materialization.exclusive ? [.exclusive] : []
                 )
+                #else
+                renameResult = Syscall.rename(
+                    from: createdPath,
+                    to: desiredPath,
+                    options: materialization.exclusive ? [.exclusive] : []
+                )
```
finagolfin, line 755: Turns out this doesn't compile on Android, because `rename(from:to:options:)` is only defined on Darwin.

finagolfin:
> Do you have any timeline on this?

Most likely later today.

> We'd like to tag a release and would rather not break you!

I mostly build trunk right now, so shouldn't be an issue.

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
glbrntt: Oops, yes, this variant is Darwin only as it has the `.exclusive` flag. I think we can avoid this by not supporting "atomic" file creation on Android when exclusive create is set (when it isn't set we can do a normal rename without the exclusive flag).

finagolfin:
Applied this locally and all the remaining errors appear to be type and optional mismatches related to slightly different Android declarations, so you can go ahead and merge this when you like.

FranzBusch (approved):

Lukasa (changes requested):
I just want to sync up here a bit: musl also lacks `renameat2` as a libc function, but it _does_ have it as a system call. I'm thinking that we'll want to omit the call through libc and instead make the system call directly. @finagolfin do we think that's gonna cause any problems on android?

finagolfin:
No, I think that should work, as I don't think that syscall is filtered out on Android.

Lukasa:
See #2628

finagolfin:
Tried that Musl syscall pull out, seems to work well after adding my patch for the type/optional mismatches and the `confstr()` commit from this pull, except I see several tests failing because they call `linkat()`, not sure where that's coming from.

finagolfin:
I will look into the `linkat()` testing issues on Android in the next couple days. Meanwhile, if @glbrntt would reduce this pull to the `confstr()` commit and merge, that would be helpful.

finagolfin (approved):
Thanks

Lukasa (approved):

PR #2642, opened 2024-02-08
Pass initial offset to BufferedReader.init

yim-lee:
Motivation:
There is a bug in `ReadableFileHandleProtocol.bufferedReader(startingAtAbsoluteOffset:capacity:)` where we are hardcoding initial offset to 0 instead of passing the actual param value.

Modifications:
Pass the actual param value.


glbrntt (approved):
Thanks Yim!

PR #2643, opened 2024-02-09
fix: conditionally define RENAME_* macros

Austinpayne:
When using `-Xcc -Werror` flags redefining macros results in a compilation error.

### Motivation:

Fixes regression in https://github.com/apple/swift-nio/pull/2628.

### Modifications:

Check if `RENAME_*` macros are already defined before before defining for the musl case.

### Result:

NIO now compiles when using `-Werror`.


glbrntt (approved):
Thank you!

Issue #1909, opened 2021-07-22
EventLoop(Future): clarify memory ordering / memory consistency


weissi:
Now that Swift has a [memory consistency model](https://github.com/apple/swift-evolution/blob/main/proposals/0282-atomics.md), we should clarify how `EventLoop`s and `EventLoopFuture`s tie into that.

For example, is the following program correct?

```swift
let group = MultiThreadedEventLoopGroup(numberOfThreads: 3)
defer { try! group.syncShutdownGracefully() }

let el1 = group.next()
let el2 = group.next()
let el3 = group.next()

// on main thread
var counter = 0
counter += 1

el1.makeSucceededFuture(()).flatMap {
    // on EL1's thread
    counter += 1

    return el2.makeSucceededFuture(()).flatMap {
        // on EL1's thread
        counter += 1
        return el3.makeSucceededFuture(())
    }
}.flatMap {
    // on EL1's thread
    counter += 1
}.wait()

// on main thread
assert(counter == 4)
```

Clearly it reads and writes `counter` from about 3 different threads without any explicit synchronisation.
However, the EventLoopFutures establish a clear before/after memory ordering and in the current implementations of `SelectableEventLoop` (because if the locks which are full mem barriers) and `NIOTSEventLoop` (because of `DispatchQueue`'s guarantees) this is actually totally safe. But it's never stated anywhere that this is _in fact_ safe. So should the programmer assume this is unsafe and resort to a `NIOAtomic<Int>` or a `Lock`? I don't think so.

I think we should add to the docs of `EventLoop` and `EventLoopFuture` a few guarantees regarding memory orders. Most importantly:
- futures: call the callbacks one after the other, with full barriers
- ELs: calls to `.execute`/`.schedule`/`.submit`/... have full barriers

Lukasa:
Hmm, so while I agree with you that we should document things, I think `EventLoopFuture` does not actually express any "happens before" relationship w.r.t its callbacks. `EventLoopFuture`'s happens-before relationship is derived solely from the `EventLoop`. This can be observed with `EmbeddedEventLoop`, and indeed if any of the above event loops were replaced with an `EmbeddedEventLoop` then the safety guarantees of the program evaporate.

To this end, I think we need to explore two _related_ concepts:

1. What the memory order is for a _thread safe_ `EventLoop` implementation: that is, the appropriate barriers and edges that an implementation must provide. Concretely, I believe there are only three actual guarantees required here:

[...]

PR #2645, opened 2024-02-12
Track `execute()` and `enqueue()` tasks separately from scheduled tasks.

gmilos:
Motivation:

`SelectableEventLoop` uses `ScheduledTask` to track tasks scheduled for some specific point in time, but also tasks that are supposed to be executed immediately.

This casuse performance issues where the number of immediate tasks grows, as they need to be kept sorted in the `_scheduledTasks` `PriorityQueue`. Additionally, that datastructure is protected by lock(s) which casues lock contention from other threads.

Finally, immediate tasks differ from future scheduled tasks in that they don't have a `failFn` to handle destruction at the end of `SelectableEventLoop.run()` execution.

Modifications:

* Modify `ScheduledTask` to only track closure tasks and require `failFn` to be provided.
* Retain `UnderlyingTask` for tracking closure or `ErasedUnownedJob` for async-await tasks.
* Add `SelectableEventLoop._immediateTasks` to track immediate tasks (collection of `UnderlyingTask`-s).
* Split production / consumption of the tasks to match the datastructure changes described above.
* Modify the `defer` block of `SelectableEventLoop.run()` to execute immediate tasks, instead of dropping them silently.

Result:

Performance of workloads relying on queuing lots of tasks gets better.


gmilos:
The test failure is ... because the # of allocations for less than expected. By about 1k. That's a good thing.

```
17:35:51  ++ assert_greater_than 155002 156050
17:35:51  ++ [[ ! 155002 -gt 156050 ]]
```

And ... I see the fix just got merged: https://github.com/apple/swift-nio/commit/e4102aecb5e7da46b07d8048a2171614d7cb4cb9#diff-660361849e6074a975e1787c5836a6138eef55096f2b80fb76d5da0af73b48d5R40

gmilos:
@weissi this is the change we discussed in person.

gmilos:

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,46 +427,80 @@ Further information:
...
                     self._pendingTaskPop = true
 
+                    // During the shutdown we don't need to be that careful about reallocs
+                    var scheduledTasksCopy = ContiguousArray<ScheduledTask>()
                     // reserve the correct capacity so we don't need to realloc later on.
                     scheduledTasksCopy.reserveCapacity(self._scheduledTasks.count)
                     while let sched = self._scheduledTasks.pop() {
                         scheduledTasksCopy.append(sched)
                     }
+                    var immediateTasksCopy = ContiguousArray<UnderlyingTask>()
```
Lukasa, line 467: We should avoid allocating new arrays here on each iteration. Prefer to have persistent ones we reuse. This is a defer block, but the loop may still execute a hefty number of times.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,46 +427,80 @@ Further information:
...
+                        immediateTasksCopy.append(immediate)
+                    }
+                    return (scheduledTasksCopy, immediateTasksCopy)
                 }
 
+                // Run all the immediate tasks. They're all "expired" and don't have failFn,
+                // therefore the best course of action is to run them.
+                for task in immediateTasksCopy {
+                    self.run(task)
+                }
```
Lukasa, line 479: This is a fairly substantial behavioural change. Previously these tasks got dropped. Normally I’d say the logic here is good but we should prepare ourselves for a few subtle bugs popping out of this.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -495,15 +535,20 @@ Further information:
             while true {
                 // TODO: Better locking
                 self._tasksLock.withLock { () -> Void in
+                    if !self._immediateTasks.isEmpty {
+                        while self.tasksCopy.count < self.tasksCopy.capacity, let task = self._immediateTasks.popFirst() {
+                            self.tasksCopy.append(task)
+                        }
+                    } else
                     if !self._scheduledTasks.isEmpty {
```
Lukasa, line 543: Can you clean this indentation up?

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -495,15 +535,20 @@ Further information:
             while true {
                 // TODO: Better locking
                 self._tasksLock.withLock { () -> Void in
+                    if !self._immediateTasks.isEmpty {
```
Lukasa, line 538: So this worries me a little bit. In particular, it affects some of the guarantees we discussed in #1909. In particular, we now have two issues:

1. Immediate tasks enqueued after scheduled ones may execute before them, even when the scheduled task could have executed or was scheduled for now.
2. Immediate tasks can starve scheduled ones altogether.

I think issue 2 is simply a bug that has to be fixed. Issue 1 is more interesting. We could argue we never promised this ordering, and we’d probably be safe, but we have historically suggested that scheduling tasks for now and calling execute weren’t meaningfully different. 

I think if we fix 2 the blast radius of this change isn’t that bad, but if we don’t then it’s really awful.

Lukasa (changes requested):
Comment in the diff

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Yes, true. But without this, the application would have their `Task`-s/functions not executed at all. I don't think any behaviour can be worse than that. We haven't been noticing, because before `Tasks` this wasn't frequently exercised path. But all Tasks are submitted without `failFn`. But yes, there could be some unintended consequences.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: 👍 I can fix 2, e.g. zipping immediate and scheduled tasks to guarantee non-starvation.

I could give a stronger ordering guarantee, by e.g. using the same `id` space `ScheduledTask`s have to `UnderlyingTask`s and sort on it before execution. Is that worth the complexity (and performance) cost? I'm sure it's clear, but the semantics would still be complex, in that tasks that get scheduled for later would need to slot in the stream of `UnderlyingTask`s out of `id` order.

weissi:

weissi:
@swift-nio-bot perf test please

weissi:
@gmilos nice!

<!DOCTYPE html>


asyncwriter_single_writes_1M_times | 1.464787299 | 1.596832264 | current | -8%


<br class="Apple-interchange-newline">

8% better in one of the perf tests that actually uses `async` stuff. That's expected to be one of the few ones because most perf tests won't use `execute`. But that one will!

Sources/NIOPosix/SelectableEventLoop.swift
weissi: Agreed it's a substantial change but it's definitely a good one. I think this will actually make tearing down an EventLoop much easier than it was before this patch. Previously we just couldn't implement this sensibly, now we can.

But of course, it'll likely trigger a few subtle changes here or there but I can't see this changing anything that was reliably working fine for the worse.

gmilos:
@weissi how much signal is in the performance tests? Looks like there are a few that did regress, including:

```
lock_8_threads_10M_ops | 0.957658591 | 0.873560162 | previous | 9%
```

gmilos:
And another:

```
future_whenallcomplete_100k_deferred_on_loop | 0.084619949 | 0.08085791 | previous | 4%
```

There seems to be a pattern where the tests with large queues degraded in performance. I'll take a look.

gmilos:
@swift-nio-bot perf test please

gmilos:
I reviewed the test responsible for:

```
lock_8_threads_10M_ops | 0.957658591 | 0.873560162 | previous | 9%
```

This doesn't exercise my code change in any meaningful way. So the degradation must be due to an unrelated difference. I'm reviewing the test @weissi pointed out:

```
schedule_and_run_100k_tasks | 0.252803345 | 0.233980813 | previous | 8%
```

This may be down to me.

gmilos:
I reviewed:

```
schedule_and_run_100k_tasks | 0.252803345 | 0.233980813 | previous | 8%
```

In the current version of the PR (which isn't the same as what was perf tested before), I see ~2x improvement. I had to raise the number of repeats by 100x to make it human scale. I also recorded sample. Which point to better arc performance with the simpler task type stored in `ScheduledTask`.

gmilos:

weissi:

weissi:
@swift-server-bot perf test please

weissi:

weissi:
@Lukasa the perf tests are broken: Both of the results are from 3rd Jan (run id 156). Don't know why it still posts them as if they were current

gmilos:

gmilos:

gmilos:

weissi:
> I did a manual performance test run on my desktop. I run perf tests on main & my branch. They show major wins:
> 
> ```
> future_whenallsucceed_10k_deferred_off_loop	0.011824291	0.0222875	current	-46%
> future_whenallcomplete_10k_deferred_off_loop	0.005605708	0.016276791	current	-65%
> schedule_and_run_100k_tasks	0.108089916	0.200684584	current	-46%
> execute_100k_tasks	0.014859583	0.153192833	current	-90%

nice!!


> ```
> 
> There are couple of regressions also: 1.
> 
> ```
> lock_8_threads_10M_ops	0.215040375	0.18947725	previous	13%
> ```
> 
> but it looks like the lock tests are noisy (some of the others go faster, so it's probably just natural variability in the OS locking)

yes, ignore the locks one for this. They depend on the contention reached.

> ```
> schedule_100k_tasks	0.037446041	0.033186875	previous	12%
> ```
> 
> When I started looking into the problem, I realised the test is badly implemented. The test prewarms the heap with 100k tasks, runs them to completion, and hands over for the measured test run. But then, the test loop is run 10x, thus placing 10x as many tasks into the heap than what the prewarm did. We're therefore testing the efficiency of heap doubling and not just the scheduling. I was able to prove this is indeed responsible for the test instability by running it with much bigger set of tasks:
> 
> ```
> measuring: schedule_10000k_tasks: 3.708622625, 4.334766542, 3.842835292, 4.574104792, 7.119135958, 4.038103833, 4.324497875, 4.464178167, 4.49384075, 4.640537542,
> ```
> 
> The runs range from 3.7 to 7.1.
> 
> I'll try to fix this test separately from this PR.

awesome, thank you!!

weissi:

weissi:
@swift-server-bot perf test please

gmilos:
The `SchedulingBenchmark` fix is here: https://github.com/apple/swift-nio/pull/2650

gmilos:
I rerun the perf tests manually with the `SchedulingBenchmark` fix proposed above. The results all look good now. Attaching them below:

| name | current | previous | winner | diff |
|---|---|---|---|---|
| `write_http_headers` | 0.006981583 | 0.007003667 | current | 0% |
| `http_headers_canonical_form` | 0.065201959 | 0.06490125 | previous | 0% |
| `http_headers_canonical_form_trimming_whitespace` | 0.011664375 | 0.01182325 | current | -1% |
| `http_headers_canonical_form_trimming_whitespace_from_short_string` | 0.010846792 | 0.010806459 | previous | 0% |
| `http_headers_canonical_form_trimming_whitespace_from_long_string` | 0.016952292 | 0.016881875 | previous | 0% |
| `bytebuffer_write_12MB_short_string_literals` | 0.061121709 | 0.06131825 | current | 0% |
| `bytebuffer_write_12MB_short_calculated_strings` | 0.04662225 | 0.04578075 | previous | 1% |
| `bytebuffer_write_12MB_medium_string_literals` | 0.375843375 | 0.371900542 | previous | 1% |
| `bytebuffer_write_12MB_medium_calculated_strings` | 0.046074583 | 0.04596325 | previous | 0% |
| `bytebuffer_write_12MB_large_calculated_strings` | 0.039829125 | 0.0395385 | previous | 0% |
| `bytebuffer_lots_of_rw` | 0.02256425 | 0.022103792 | previous | 2% |
| `bytebuffer_write_http_response_ascii_only_as_string` | 0.016512291 | 0.015939625 | previous | 3% |
| `bytebuffer_write_http_response_ascii_only_as_staticstring` | 0.015740875 | 0.014887708 | previous | 5% |
| `bytebuffer_write_http_response_some_nonascii_as_string` | 0.016609083 | 0.015876542 | previous | 4% |
| `bytebuffer_write_http_response_some_nonascii_as_staticstring` | 0.015478166 | 0.014938584 | previous | 3% |
| `no-net_http1_1k_reqs_1_conn` | 0.006108959 | 0.006180042 | current | -1% |
| `http1_1k_reqs_1_conn` | 0.033368541 | 0.033116917 | previous | 0% |
| `http1_1k_reqs_100_conns` | 0.054673291 | 0.054437 | previous | 0% |
| `future_whenallsucceed_100k_immediately_succeeded_off_loop` | 0.069128083 | 0.06880025 | previous | 0% |
| `future_whenallsucceed_100k_immediately_succeeded_on_loop` | 0.068877625 | 0.069294458 | current | 0% |
| `future_whenallsucceed_10k_deferred_off_loop` | 0.012099125 | 0.022666375 | current | -46% |
| `future_whenallsucceed_10k_deferred_on_loop` | 0.011050083 | 0.011109875 | current | 0% |
| `future_whenallcomplete_100k_immediately_succeeded_off_loop` | 0.030047792 | 0.0300765 | current | 0% |
| `future_whenallcomplete_100k_immediately_succeeded_on_loop` | 0.029346708 | 0.030066708 | current | -2% |
| `future_whenallcomplete_10k_deferred_off_loop` | 0.005671625 | 0.015987834 | current | -64% |
| `future_whenallcomplete_100k_deferred_on_loop` | 0.04997975 | 0.049469292 | previous | 1% |
| `future_reduce_10k_futures` | 0.011866709 | 0.011785208 | previous | 0% |
| `future_reduce_into_10k_futures` | 0.009929375 | 0.009837 | previous | 0% |
| `channel_pipeline_1m_events` | 0.042976584 | 0.043107 | current | 0% |
| `websocket_encode_50b_space_at_front_100k_frames_cow` | 0.018315833 | 0.018359458 | current | 0% |
| `websocket_encode_50b_space_at_front_1m_frames_cow_masking` | 0.3579905 | 0.354862875 | previous | 0% |
| `websocket_encode_1kb_space_at_front_1m_frames_cow` | 0.196264417 | 0.194618708 | previous | 0% |
| `websocket_encode_50b_no_space_at_front_100k_frames_cow` | 0.018282459 | 0.0181675 | previous | 0% |
| `websocket_encode_1kb_no_space_at_front_100k_frames_cow` | 0.019736083 | 0.019344667 | previous | 2% |
| `websocket_encode_50b_space_at_front_100k_frames` | 0.030248458 | 0.030699083 | current | -1% |
| `websocket_encode_50b_space_at_front_10k_frames_masking` | 0.004702875 | 0.004675792 | previous | 0% |
| `websocket_encode_1kb_space_at_front_10k_frames` | 0.003431167 | 0.003457959 | current | 0% |
| `websocket_encode_50b_no_space_at_front_100k_frames` | 0.030517875 | 0.030633041 | current | 0% |
| `websocket_encode_1kb_no_space_at_front_10k_frames` | 0.003432833 | 0.003474583 | current | -1% |
| `websocket_decode_125b_10k_frames` | 0.006401 | 0.006419208 | current | 0% |
| `websocket_decode_125b_with_a_masking_key_10k_frames` | 0.006506875 | 0.0067095 | current | -3% |
| `websocket_decode_64kb_10k_frames` | 0.006525792 | 0.006700417 | current | -2% |
| `websocket_decode_64kb_with_a_masking_key_10k_frames` | 0.006675708 | 0.006802083 | current | -1% |
| `websocket_decode_64kb_+1_10k_frames` | 0.006505042 | 0.006692208 | current | -2% |
| `websocket_decode_64kb_+1_with_a_masking_key_10k_frames` | 0.006727583 | 0.006812292 | current | -1% |
| `circular_buffer_into_byte_buffer_1kb` | 0.020241209 | 0.020156042 | previous | 0% |
| `circular_buffer_into_byte_buffer_1mb` | 0.039579667 | 0.039712958 | current | 0% |
| `byte_buffer_view_iterator_1mb` | 0.00040675 | 0.0004065 | previous | 0% |
| `byte_buffer_view_contains_12mb` | 0.019846375 | 0.020035916 | current | 0% |
| `byte_to_message_decoder_decode_many_small` | 0.020560792 | 0.0209705 | current | -1% |
| `generate_10k_random_request_keys` | 0.004089042 | 0.00415975 | current | -1% |
| `bytebuffer_rw_10_uint32s` | 0.021125083 | 0.020929542 | previous | 0% |
| `bytebuffer_multi_rw_10_uint32s` | 0.039448042 | 0.039168375 | previous | 0% |
| `lock_1_thread_10M_ops` | 0.085574208 | 0.086346459 | current | 0% |
| `lock_2_threads_10M_ops` | 0.152125875 | 0.15997275 | current | -4% |
| `lock_4_threads_10M_ops` | 0.209404916 | 0.219198125 | current | -4% |
| `lock_8_threads_10M_ops` | 0.194281334 | 0.193939542 | previous | 0% |
| `schedule_100k_tasks` | 0.030432792 | 0.031422375 | current | -3% |
| `schedule_and_run_100k_tasks` | 0.108168959 | 0.202768583 | current | -46% |
| `execute_100k_tasks` | 0.014813375 | 0.152129334 | current | -90% |
| `bytebufferview_copy_to_array_100k_times_1kb` | 0.005596 | 0.005377667 | previous | 4% |
| `circularbuffer_copy_to_array_10k_times_1kb` | 0.010262417 | 0.010123291 | previous | 1% |
| `deadline_now_1M_times` | 0.012010584 | 0.0120285 | current | 0% |
| `asyncwriter_single_writes_1M_times` | 1.0667385 | 1.067504709 | current | 0% |
| `asyncsequenceproducer_consume_1M_times` | 0.632515333 | 0.611443667 | previous | 3% |
| `udp_10k_writes` | 0.209090416 | 0.209700042 | current | 0% |
| `udp_10k_vector_writes` | 0.089144167 | 0.091539292 | current | -2% |
| `udp_10k_vector_reads` | 0.209473375 | 0.208093917 | previous | 0% |
| `udp_10k_vector_reads_and_writes` | 0.089285833 | 0.091710708 | current | -2% |

(minor fixups were needed in the markdown output generation logic, I filed this as: https://github.com/apple/swift-nio/pull/2651)

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -77,10 +77,12 @@ internal final class SelectableEventLoop: EventLoop {
     internal var scheduledTaskCounter = ManagedAtomic<UInt64>(0)
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
+    @usableFromInline
+    internal var _immediateTasks = CircularBuffer<UnderlyingTask>()
 
     // We only need the ScheduledTask's task closure. However, an `Array<() -> Void>` allocates
```
weissi, line 83: nit: remove the comment, no longer relevant as you're now holding `UnderlyingTask` anyway

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,16 +427,101 @@ Further information:
...
+                    
+                    // Reset nextReadyDeadline to nil which means we will do a blocking select.
+                    return nil
+                }
+                
+                // We only fetch the time one time as this may be expensive and is generally good enough as if we miss anything we will just do a non-blocking select again anyway.
+                let now: NIODeadline = .now()
+                var nextScheduledTaskDeadline = now
+
+                while self.tasksCopy.count < self.tasksCopy.capacity &&
```
weissi, line 468: @gmilos why do we consider `self.tasksCopy.capacity`? That's essentially a random value

weissi (changes requested):
Looks mostly great, but one question regarding the usage of `.capacity` seems important enough such that I'm requesting changes for now.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: This is how this code always worked (except I span off this block to a separate function), see:
https://github.com/apple/swift-nio/pull/2645/files#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faL503

The value isn't random, it's set to `4096`, and never changed:
https://github.com/apple/swift-nio/pull/2645/files#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR188

I infer we wanted to get batch of tasks out, under the lock. And use `taskCopy` as completely "static" memory block to shuffle tasks out from the `_scheduledTasks` heap (and now also `_immediateTasks` circular buffer).

I would appreciate review of the logic here though. It's a bit fiddly. Especially around when is the `_pendingTaskPop` meant to be set and what value should `nextScheduledTaskDeadline` assume.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: done.

Sources/NIOPosix/SelectableEventLoop.swift
weissi: oh dear, I had no idea :). That's super duper fragile, capacity is _at least_ what `reserveCapacity` did, unless we ever call `removeAll` forgetting to pass `keepingCapacity: true`. Would you mind fixing this long-standing bug and just comparing it to a constant? As in `self.tasksCopy.count < Self.maximumEventLoopTaskExecutionBatchSize` or something?

gmilos: fixed here: https://github.com/apple/swift-nio/pull/2645/commits/f76e65bb71f92bd35b532bbeb477d8ec1953a328

I happen to have picked a different name for the constant (before reading your comment here). Same idea though.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,16 +430,101 @@ Further information:
...
+            }
+            
+            // all pending tasks are set to occur in the future, so we can stop looping.
+            if self.tasksCopy.isEmpty {
+                return nextReadyDeadline
+            }
+
+            // Execute all the tasks that were submitted
+            for task in self.tasksCopy {
+                run(task)
```
weissi, line 510:
```suggestion
                self.run(task)
```

gmilos:
@weissi that's all the feedback handled. lmk if there is any more.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Done, with the "zipping" approach. Happy to do the (merge) sorting if we need stricter ordering guarantee.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Done.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Done.

gmilos:

gmilos:

weissi (approved):
Thank you! That looks good to me.

Lukasa:
From a "strict semver" this is a patch change, but I'm going to take some editorial discretion and mark it "minor". We're making a subtle but observable behavioural change, and while it will _probably_ be fine I want to make it possible for users to ensure that they don't have to deal with both variants by calling this a minor instead of a patch.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -77,10 +77,15 @@ internal final class SelectableEventLoop: EventLoop {
     internal var scheduledTaskCounter = ManagedAtomic<UInt64>(0)
     @usableFromInline
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
+    @usableFromInline
+    internal var _immediateTasks = CircularBuffer<UnderlyingTask>()
```
Lukasa, line 81: While we're here, per @dnadoba's feedback on another PR, let's use Deque here.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -77,10 +77,15 @@ internal final class SelectableEventLoop: EventLoop {
...
+
+    // Historical note: an `Array<() -> Void>` allocates for every appended closure. https://bugs.swift.org/browse/SR-15872
+    // we're now appending an enum type, one member of which is such closure. We're
+    // therefore side stepping the issue. It may become relevant again if we reorganised the code.
+    private var tasksCopy = ContiguousArray<UnderlyingTask>()
+    static private let taskCopyBatchSize = 4096
```
Lukasa, line 87: Silly nit, but let's make this a computed var so we don't store this in memory.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,16 +430,169 @@ Further information:
...
+                    if nextScheduledTaskDeadline == now {
+                        return true
+                    }
+
+                    guard let topMostScheduledTask = self._scheduledTasks.peek() else {
+                        return false
+                    }
+
+                    return topMostScheduledTask.readyTime == nextScheduledTaskDeadline
+                }
```
Lukasa, line 560: Having all these assertions is great, but I feel like they're obscuring the algorithm itself, and the fact that they're inline closures gives them the ability to mutate the state (though they don't).

What about factoring these out to a `private static func` that runs all of these, and takes the state it validates as input (that is, does not pass `self`)? That would help make it really easy for future maintenance to see that these do not modify the state, and avoids these landing smack in the middle of the algorithm, but helps keep all their value.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,16 +430,169 @@ Further information:
...
+                        } else {
+                            nextScheduledTaskDeadline = task.readyTime
+                            moreScheduledTasksToConsider = false
+                        }
+                    } else {
+                        moreScheduledTasksToConsider = false
+                    }
+                }
+
+                assert(self.tasksCopy.count <= Self.taskCopyBatchSize)
```
Lukasa, line 497: This assertion seems like it is straightforward to break. We can add two tasks in each loop iterator above, but we only check the length once. I think the loop above needs an extra length check.

Sources/NIOPosix/SelectableEventLoop.swift
Lukasa: Actually, maybe an even more drastic refactor is available. A quick check of the current `main` suggests that `tasksCopy` being a stored property on `self` is entirely unnecessary: it's only ever accessed at `init` and in this method.

That suggests to me that we can refactor the entire `zip` logic into a `private static` helper function that takes the tasksCopy array `inout` and populates it, including testing its invariants. That would be a really nice refactor.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -439,28 +601,44 @@ Further information:
                     while let sched = self._scheduledTasks.pop() {
                         scheduledTasksCopy.append(sched)
                     }
+                    immediateTasksCopy.reserveCapacity(self._immediateTasks.count)
+                    while let immediate = self._immediateTasks.popFirst() {
+                        immediateTasksCopy.append(immediate)
+                    }
```
Lukasa, line 607: Instead of this `while let`, can we use `swap`?

Lukasa (commented):
Generally looking really nice. I have a few small refactoring suggestions, as we're in the space and you've already started with refactors ;)

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Good catch. Reorganising the loop to check the length of the `taskCopy` array at each point we're mutating (adding to) it.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: This is removing tasks from `self._immediateTasks` and putting them to a different array `immediateTasksCopy`. So `swap` isn't applicable, unless there is some other `swap` I'm not aware of.

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Heh, I tried to write a test for this, and realised the code before was fine, just. It's because:
* if `_scheduledTasks` and `_immediateTasks` are non-empty, we'd be adding 2 items to the `tasksCopy` each time round the loop. And `taskCopyBatchSize` is even, so one test every 2 items is sufficient
* if either `_scheduledTasks` or `_immediateTasks` run out of tasks, we'd be testing the size of `tasksCopy` once per each addition. This applies even if we just run out when we're terminating the batch.

This wasn't intentional micro-optimisation, so I'm adding the additional length check back to make the code safer against future refactoring (including making the size of the batch odd).

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Done, I introduced:
* [`_popTasksLocked`](https://github.com/apple/swift-nio/pull/2645/files#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR518): static function that takes all the state it needs to operate on `inout`
* [`_popTasksLockedAssertInvariants`](https://github.com/apple/swift-nio/pull/2645/files#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR444): static function that takes all the state after popping took place, and validates the invariants. The state is passed immutably.

Sources/NIOPosix/SelectableEventLoop.swift
Lukasa: Yeah my concern was very much about future refactors :wink:

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: Ah, this [`swap`](https://developer.apple.com/documentation/swift/swap(_:_:)). Adopted here: https://github.com/apple/swift-nio/pull/2645/commits/8b818e6af83c1ba0808f4c4778eda6be09c3f98b#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR632

Sources/NIOPosix/SelectableEventLoop.swift
gmilos: FYI @Lukasa, I made the  `tasksCopy` a local var in `runLoop`, but then made it a member variable back again in the [latest commit](https://github.com/apple/swift-nio/pull/2645/commits/c7a6eced58510463c2a1ce76288fed131227ba43). This is because otherwise allocation counter regresses (and shows up for tests that churn through lots of `SelectableEventLoop` instances).

gmilos:
Ah, `Deque` has `1.5` growth factor: https://github.com/apple/swift-collections/blob/ca8b4ab855f4b8075c1fd29eb50db756b1688e61/Sources/DequeModule/Deque._Storage.swift#L161, it doesn't appear to be configurable.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -421,16 +431,195 @@ Further information:
...
+                                        tasksCopy: tasksCopy,
+                                        tasksCopyBatchSize: tasksCopyBatchSize,
+                                        now: now,
+                                        nextScheduledTaskDeadline: nextScheduledTaskDeadline,
+                                        moreImmediateTasksToConsider: moreImmediateTasksToConsider, moreScheduledTasksToConsider: moreScheduledTasksToConsider)
+        }
+
+
+        // nextScheduledTaskDeadline is the overall next deadline, but iff there are no more immediate tasks left.
+        return moreImmediateTasksToConsider ? now : nextScheduledTaskDeadline
```
Lukasa, line 582: Will we ever hit `now`? We can only get here if both `moreImmediateTasksToConsider` _and_ `moreScheduledTasksToConsider` are both false, which should imply we never hit `now`.

gmilos: Ah, yes. This changed because the batch size checks got moved to each of the `if`-s, and not the loop overall. I think the statement is harmless, but can be simplified. Checking it carefully.

gmilos: The statement was indeed harmless, because while it'd always return the deadline for next scheduled task, this deadline is only innacted _iff_ there are no more tasks to run. In all other cases we have to recheck, because more tasks may get queued up. Still, this deserved a cleanup, with [`now` returned if there are any more immediate tasks left](https://github.com/apple/swift-nio/pull/2645/commits/7555cc5fae698da0eb2f32d5efad956ff186746e?diff=split&w=0#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR549) and an [associated assert](https://github.com/apple/swift-nio/pull/2645/commits/7555cc5fae698da0eb2f32d5efad956ff186746e?diff=split&w=0#diff-565b354476ebe678da3be159a85e33e0b7630b1e45b62b20efe506210c1d62faR482-R488). Also, I added a test that tries to get [immediate tasks stuck](https://github.com/apple/swift-nio/pull/2645/commits/7555cc5fae698da0eb2f32d5efad956ff186746e?diff=split&w=0#diff-ee2a5074023719eacac1b5187aeb224fb9afb04f9e010b85809c3f3eb9a2b3b4R1801-R1836) (which never reproduced a problem, but useful defence for the future).

Lukasa (approved):

PR #2654, opened 2024-02-16
Fix `EventLoopFuture` and `EventLoopPromise` under strict concurrency checking

FranzBusch:
# Motivation

We need to tackle the remaining strict concurrency checking related `Sendable` warnings in NIO. The first place to start is making sure that `EventLoopFuture` and `EventLoopPromise` are properly annotated.

# Modification

In a previous https://github.com/apple/swift-nio/pull/2496, @weissi changed the `@unchecked Sendable` conformances of `EventLoopFuture/Promise` to be conditional on the sendability of the generic `Value` type. After having looked at all the APIs on the future and promise types as well as reading the latest Concurrency evolution proposals, specifically the [Region based Isolation](https://github.com/apple/swift-evolution/blob/main/proposals/0414-region-based-isolation.md), I came to the conclusion that the previous `@unchecked Sendable` annotations were correct. The reasoning for this is:

1. An `EventLoopPromise` and `EventLoopFuture` pair are tied to a specific `EventLoop`
2. An `EventLoop` represents an isolation region and values tied to its isolation are not allowed to be shared outside of it unless they are disconnected from the region
3. The `value` used to succeed a promise often come from outside the isolation domain of the `EventLoop` hence they must be transferred into the promise.
4. The isolation region of the event loop is enforced through `@Sendable` annotations on all closures that receive the value in some kind of transformation e.g. `map()` or `whenComplete()`
5. Any method on `EventLoopFuture` that combines itself with another future must require `Sendable` of the other futures `Value` since we cannot statically enforce that futures are bound to the same event loop i.e. to the same isolation domain

Due to the above rules, this PR adds back the `@unchecked Sendable` conformances to both types. Furthermore, this PR revisits every single method on `EventLoopPromise/Future` and adds missing `Sendable` and `@Sendable` annotation where necessary to uphold the above rules. A few important things to call out:

- Since `transferring` is currently not available this PR requires a `Sendable` conformance for some methods on `EventLoopPromise/Future` that should rather take a `transffering` argument
- To enable the common case where a value from the same event loop is used to succeed a promise I added two additional methods that take a `eventLoopBoundResult` and enforce dynamic isolation checking. We might have to do this for more methods once we adopt those changes in other targets/packages.

# Result

After this PR has landed our lowest level building block should be inline with what the rest of the language enforces in Concurrency. The `EventLoopFuture.swift` produces no more warnings under strict concurrency checking on the latest 5.10 snapshots.


Sources/NIOCore/EventLoop.swift
```diff
@@ -742,18 +736,15 @@ extension EventLoop {
...
     @inlinable
     @preconcurrency
-    public func flatSubmit<T>(_ task: @escaping @Sendable () -> EventLoopFuture<T>) -> EventLoopFuture<T> {
-        self._flatSubmit(task)
-    }
-    @usableFromInline typealias FlatSubmitCallback<T> = @Sendable () -> EventLoopFuture<T>
-
-    @inlinable
-    func _flatSubmit<T>(_ task: @escaping FlatSubmitCallback<T>) -> EventLoopFuture<T> {
+    public func flatSubmit<T: Sendable>(_ task: @escaping @Sendable () -> EventLoopFuture<T>) -> EventLoopFuture<T> { // TODO: This should take a closure that returns fresh
```
glbrntt, line 739: ... that returns fresh?

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -178,30 +178,52 @@ public struct EventLoopPromise<Value> {
...
+    /// Deliver a successful result to the associated `EventLoopFuture<Value>` object.
+    ///
+    /// - Note: The call to this method must happen on the same event loop as this promise was created from.
+    ///
+    /// - parameters:
+    ///     - eventLoopBoundValue: The successful result of the operation.
+    @inlinable
+    public func succeed(eventLoopBoundValue: Value) {
+        self._resolve(eventLoopBoundResult: .success(eventLoopBoundValue))
+    }
```
glbrntt, line 196: I'm not sure this is the right spelling; the pattern we have for functions which must be called on the right EL with runtime assertions are synchronous views (e.g. `ChannelPipeline.SynchronousOperations` and `Channel.syncOptions`). 

It also skirts around the slight oddity with this API in that you could call this function with a `Sendable` i.e. not EL bound value with no ill side effect.

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1334,6 +1388,67 @@ extension EventLoopFuture {
...
+                if case .failure = result {
+                    return  // Once the promise is failed, future results do not need to be processed.
+                }
+            } else {
+                // We have to map to `Void` here to avoid sharing the potentially non-Sendable
+                // value across event loops.
+                future
+                    .map { _ in () }
+                    .hop(to: eventLoop)
+                    .whenComplete { result in processResult(index, result) }
```
glbrntt, line 1448: Can't we do something like the following to avoid the extra ELFs?

```swift
future.whenComplete { result in
  let voidResult = result.map { _ in }
  future.eventLoop.exeute {
    processResult(index, voidResult)
  }
}
```

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1334,6 +1388,67 @@ extension EventLoopFuture {
...
+                // Fast-track already-fulfilled results without the overhead of calling `whenComplete`. This can yield a
+                // ~20% performance improvement in the case of large arrays where all elements are already fulfilled.
+                switch result {
+                case .success:
+                    processResult(index, .success(()))
+                case .failure(let error):
+                    processResult(index, .failure(error))
+                }
+                if case .failure = result {
+                    return  // Once the promise is failed, future results do not need to be processed.
```
glbrntt, line 1440: Should this be moved into the `failure` case above?

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1484,6 +1616,58 @@ extension EventLoopFuture {
...
+                case .failure(let error):
+                    processResult(index, .failure(error))
+                }
+            } else {
+                // We have to map to `Void` here to avoid sharing the potentially non-Sendable
+                // value across event loops.
+                future
+                    .map { _ in () }
+                    .hop(to: eventLoop)
+                    .whenComplete { result in processResult(index, result) }
```
glbrntt, line 1667: Same as above re: `whenComplete`

glbrntt (changes requested):
I think the explanation makes sense and this all ties together nicely. I think the spelling of the new APIs is not quite right though and we should align with patterns we've used elsewhere in NIO (i.e. sync views)

Sources/NIOCore/EventLoopFuture.swift
FranzBusch: I agree. I removed those `loopBound` APIs from this PR since they are really part of the `AssumeIsolated` PR

glbrntt (commented):
LGTM now. Purposefully not selecting 'Approve' as I'd like at least one other set of eyes to approve this too, but consider it approved by me.

Lukasa (approved):

glbrntt (approved):

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1825,13 +1949,19 @@ public struct _NIOEventLoopFutureIdentifier: Hashable, Sendable {
...
+//
+// 1. Receiving the value of the future is always done on the EventLoop of the future, hence
+// the value is never transferred out of the event loops isolation domain. It only gets transferred
+// by certain methods such as `hop()` and those methods are annotated with requiring the Value to be
+// Sendable
+// 2. The promise is `Sendable` but fulfilling the promise with a value requires the user to
+// transfer the value to the promise. This ensures that the value is now isolated to the event loops
+// isolation domain. Note: Sendable values can always be transferred
+
+extension EventLoopPromise: @unchecked Sendable { }
```
dnadoba, line 1962: Non need for unchecked. `EventLoopFuture` is already `Sendable` and that is the only store property.
```suggestion
extension EventLoopPromise: Sendable { }
```

FranzBusch: I think this still applies.

Lukasa:
I've modified this further, and also added some narrative documentation explaining the thinking behind the type. Please re-review.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+
+In Swift NIO the most common mechanism to arrange a series of asynchronous work items is
+_not_ to queue up a series of ``EventLoop/execute(_:)`` calls. Instead, users typically
+use ``EventLoopFuture``.
+
+``EventLoopFuture`` has some extensive semantics documented in its API documentation. The
+most important principal for this discussion is that all callbacks added to an
+``EventLoopFuture`` will execute on the ``EventLoop`` to which that ``EventLoopFuture`` is
+bound. By extension, then, all callbacks added to an ``EventLoopFuture`` execute on the same
+executor (the ``EventLoop``) in the same isolation domain.
```
FranzBusch, line 78: I would love if we could highlight that part more. I am always thinking of our ELs as both an executor but also as an isolation domain.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+
+The analogy to an actor here is hopefully fairly clear. Conceptually, an ``EventLoopFuture``
+could be modelled as an actor. That means all the callbacks have the same logical semantics:
+the ``EventLoopFuture`` defines an isolation domain, and all the callbacks are `sent` into the
```
FranzBusch, line 82: IMO the EL defines an isolation domain and not the ELF

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1012,16 +1040,18 @@ extension EventLoopFuture {
     ///
     /// This is also forbidden in async contexts: prefer ``EventLoopFuture/get()``.
     ///
+    /// - Note: The `Value` must be `Sendable` since it is shared outside of the isolation domain of the event loop.
+    ///
     /// - returns: The value of the `EventLoopFuture` when it completes.
     /// - throws: The error value of the `EventLoopFuture` if it errors.
     @available(*, noasync, message: "wait() can block indefinitely, prefer get()", renamed: "get()")
-    @inlinable
```
FranzBusch, line 1018: Did we drop the `inalienable` intentionally here?

FranzBusch (commented):
LGTM. Thanks for taking the time to write the document. I can't approve the PR since I am the author but consider this comment as an approval. I only had three small comments.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
Lukasa: Line 54 has this earlier in the doc, is there a step you think I should use to punch it up?

FranzBusch: Oh damn I somehow read right over the second part of that sentence. Ignore me.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+    are `Sendable`. This is not a long-term position, but reflects the need to continue
+    to support Swift 5 code which requires this stricter standard. In a future release of
+    SwiftNIO we expect to relax this constraint: if you need this relaxed constraint
+    then please file an issue.
+
+There are also a few ways to extract a value, such as ``EventLoopFuture/wait(file:line:)``
+and ``EventLoopFuture/get()``. These APIs can only safely be called when the ``EventLoopFuture``
+is carrying a `Sendable` value. This is because ``EventLoopFuture``s hold on to their value and
+can give it to other closures or other callers of `get` and `wait`. Thus, `sending` is not
+sufficient.
```
ktoso, line 112: I think that's right, yes.

ktoso (approved):
I didn;t notice anything in the .md that would trigger spidey senses, It looks right I think.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -62,8 +63,11 @@ extension EventLoopPromise {
     /// - returns: A `Task` which was created to `await` the `body`.
     @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
     @discardableResult
+    @preconcurrency
     @inlinable
-    public func completeWithTask(_ body: @escaping @Sendable () async throws -> Value) -> Task<Void, Never> {
+    public func completeWithTask(
+        _ body: @escaping @Sendable () async throws -> Value
```
weissi, line 69: Pretty sure this is incorrect. We're running `body` in one place and it can return a non-Sendable `Value` which we then transfer onto the EL. Nothing restricts `Value` from still being accessed on the previous place.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -457,10 +457,10 @@ public final class ChannelPipeline: ChannelInvoker {
         let promise = self.eventLoop.makePromise(of: ChannelHandlerContext.self)
 
         if self.eventLoop.inEventLoop {
-            promise.completeWith(self.contextSync(handler: handler))
+            promise.assumeIsolated().completeWith(self.contextSync(handler: handler))
```
weissi, line 460: `assumeIsolated()` is a noop in `release`, right?

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+
+NIO predates the Swift concurrency model. As a result, several of NIO's concepts are not perfect matches to
+the concepts that Swift uses, or have overlapping responsibilities.
+
+## Isolation domains and executors
+
+First, a quick recap. The core of Swift 6's data-race safety protection is the concept of an "isolation
+domain". Some valuable reading regarding the concept can be found in
+[SE-0414 (Region based isolation)](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0414-region-based-isolation.md)
+but at a high level an isolation domain can be understood to be a region within which there cannot be
```
weissi, line 15: I wouldn't use the word "region" here because that's easily confusable with "isolation region" which is a completely different thing

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+multiple executors executing code at the same time.
+
+In standard Swift Concurrency, the main boundaries of isolation regions are actors and tasks. Each actor,
```
weissi, line 18:
```suggestion
In standard Swift Concurrency, the main boundaries of isolation domains are actors and tasks. Each actor,
```

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+including global actors, defines an isolation domain. Additionally, for functions and methods that are
+not isolated to an actor, the `Task` within which that code executes defines an isolation domain. Passing
+values between these isolation domains requires that these values are either `Sendable` (safe to hold in
+multiple domains), or that the `sending` keyword is used to force the value to be passed from one domain
+to another.
+
+A related concept to an "isolation domain" is an "executor". Again, useful reading can be found in
+[SE-0392 (Custom actor executors)](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0392-custom-actor-executors.md).
+At a high level, an executor is simply an object that is capable of executing Swift `Task`s. Executors can be
```
weissi, line 27: Probably a nit but I'm not sure if `Task` is 100% correct here. Like there's no way to retrieve a `let theTask: Task` from a child task. But then, child tasks still ask `Task.isCancelled` so maybe it is correct :).

I would've said that they can run Jobs but then that's not really terminology that's actually used. Probably fine

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
```diff
@@ -0,0 +1,176 @@
...
+concurrent, or they can be serial. Serial executors are the most common, as they can be used to back an
+actor.
+
+## Event Loops
+
+NIO's core execution primitive is the ``EventLoop``. An ``EventLoop`` is fundamentally nothing more than
+a Swift Concurrency Serial Executor that can also perform I/O operations directly. Indeed, NIO's
+``EventLoop``s can be exposed as serial executors, using ``EventLoop/executor``. This provides a mechanism
+to protect actor-isolated state using a NIO event-loop. With [the introduction of task executors](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0417-task-executor-preference.md),
+future versions of SwiftNIO will also be able to offer their event loops for individual `Task`s to execute
```
weissi, line 37: that's already done today, no?

Sources/NIOCore/EventLoopFuture+Deprecated.swift
```diff
@@ -58,44 +60,48 @@ extension EventLoopFuture {
...
     }
 
+    @preconcurrency
     @inlinable
     @available(*, deprecated, message: "Please don't pass file:line:, there's no point.")
     public func flatMapResult<NewValue, SomeError: Error>(
         file: StaticString = #fileID,
         line: UInt = #line,
-        _ body: @escaping (Value) -> Result<NewValue, SomeError>
+        _ body: @escaping @Sendable (Value) -> Result<NewValue, SomeError>
```
weissi, line 74: This requires `NewValue` to be `Sendable` but it isn't

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1012,16 +1040,19 @@ extension EventLoopFuture {
...
+    @preconcurrency
     @inlinable
-    public func wait(file: StaticString = #file, line: UInt = #line) throws -> Value {
+    public func wait(file: StaticString = #file, line: UInt = #line) throws -> Value where Value: Sendable {
         try self._wait(file: file, line: line)
     }
 
     @inlinable
-    func _wait(file: StaticString, line: UInt) throws -> Value {
+    func _wait(file: StaticString = #file, line: UInt = #line) throws -> Value where Value: Sendable {
```
weissi, line 1055: don't think the default args are a good idea. They're either provided or wrong I think

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1481,34 +1596,33 @@ extension EventLoopFuture {
...
 
-        var remainingCount = futures.count
-
-        if remainingCount == 0 {
+        if futures.count == 0 {
             promise.succeed(())
             return
         }
 
+        let remainingCount = NIOLoopBoundBox(_value: futures.count, uncheckedEventLoop: eventLoop)
```
weissi, line 1617: atomic?

Sources/NIOCore/ChannelPipeline.swift
FranzBusch: Right now it is but it needs to change. It uses `assert`s to check but we need to upgrade this `preconditions` since check at creation time is not enough similar to loop bound. Here is an example where just a `precondition` on creation of the `assumeIsolated` would fail

```
// On the EL
let assumeIsolated = someFuture.assumeIsolated()
assumeIsolated.wait() // Fine without a compiler error since it is no longer required to be Sendable. Also no runtime crash

await withTaskExecutorPreference(someRandomExec) {
  assumeIsolated.wait() // No compiler error but runtime crash now
}
```

Sources/NIOCore/EventLoopFuture+Deprecated.swift
FranzBusch: I don't think so. This is a simple transformation of `Value` to `Result<NewValue, SomeError>`. There is no EL hop or anything here possible since the body closure does **not** return a ELF

Sources/NIOCore/ChannelPipeline.swift
weissi: @FranzBusch the line above is `if self.eventLoop.inEventLoop()`, so the `assumeIsolated()` is duplicated.

FranzBusch: Ah sorry, I thought you were asking in a more general term. In this particular case I don't know if the compiler is able to optimise the precondition away.

weissi: No, it doesn't do that sort of thing. If we're not sure that this is a noop, we should remove it here. Already not super cheap to check it

weissi: Or do `assertIsolated` not `preconditionIsolated`

Sources/NIOCore/EventLoopFuture+Deprecated.swift
Lukasa: Right, as the closure is `@Sendable` the `NewValue` can only have come from somewhere else if it is `Sendable`, or is produced by the closure itself. That closure executes on the EL isolation domain, so no sending is occurring here.

weissi: Oh sorry I missed that this is synchronous :). Ignore me, completely fine as it'll be running on the EL

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
Lukasa: Nope: #2732.

Sources/NIOCore/EventLoopFuture.swift
Lukasa: Agreed, addressed.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
Lukasa: I tried for a while to rewrite this and couldn't come up with anything I liked. If we think of anything better we can change it.

Sources/NIOCore/Docs.docc/loops-futures-concurrency.md
Lukasa: Reworded.

Sources/NIOCore/AsyncAwaitSupport.swift
Lukasa: I agree with Franz, this seems fine. So far as I can see, we can only close over `Sendable` state in the closure, so anything we closed over must be safe to send. Assuming that's true, any value we return was either closed over, or produced on the EL (as that's where the body executes), so I think this is fine.

weissi: I missed the `where Value: Sendable`, so this seems fine. I assumed that it can return a non-Sendable Value because I missed it but that's not the case.

PR #2657, opened 2024-02-19
Introduce `assumeIsolated()` methods on `EventLoop`, `EventLoopPromise` and `EventLoopFuture`

FranzBusch:
> All methods/types are currently `internal` so we don't have to bikeshed just yet but we can move forward to get `NIOCore` warning free under strict concurrency

# Motivation

Methods on the above types are often called from the same event loop; however, we cannot prove to the compiler that this is true so we had to mark many methods on those types with `@Sendable` or require the generic type to be `Sendable`. This leads to unnecessary usage of `NIOLoopBound` when instead we should just dynamically assert that we are on the event loop. @dnadoba opened a very similar PR https://github.com/apple/swift-nio/pull/2228.

# Modification

This PR provides a method called `assumeIsolated()` on the three types that returns a type which re-declaration of all methods of the wrapped typed that have `Sendable` annotations. This new type is asserting at runtime that we are on the right event loop; hence, we don't need a `Sendable` value.

# Result

This PR makes it easier for our adopters to avoid newly introduced `Sendable` warnings

### DIFF
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
@@ -135,20 +135,19 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
         case .transformation(let channelReadTransformation):
             // The unsafe transfers here are required because we need to use self in whenComplete
             // We are making sure to be on our event loop so we can safely use self in whenComplete
-            let unsafeSelf = NIOLoopBound(self, eventLoop: context.eventLoop)
-            let unsafeContext = NIOLoopBound(context, eventLoop: context.eventLoop)
             channelReadTransformation(unwrapped)
                 .hop(to: context.eventLoop)
+                .assumeIsolated()
                 .map { result -> ProducerElement in
-                    context.eventLoop.preconditionInEventLoop()
                     // We have to fire through the original data now. Since our channelReadTransformation
                     // is the channel initializer. Once that's done we need to fire the channel as a read
                     // so that it hits channelRead0 in the base socket channel.
                     context.fireChannelRead(data)
                     return result
                 }
+                .assumeIsolated()
                 .whenComplete { result in
-                    unsafeSelf.value._transformationCompleted(context: unsafeContext.value, result: result)
+                    self._transformationCompleted(context: context, result: result)
                 }
         }
     }
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
@@ -138,14 +138,16 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
             channelReadTransformation(unwrapped)
                 .hop(to: context.eventLoop)
                 .assumeIsolated()
-                .map { result -> ProducerElement in
-                    // We have to fire through the original data now. Since our channelReadTransformation
-                    // is the channel initializer. Once that's done we need to fire the channel as a read
-                    // so that it hits channelRead0 in the base socket channel.
-                    context.fireChannelRead(data)
-                    return result
-                }
                 .whenComplete { result in
+                    switch result {
+                    case .success:
+                        // We have to fire through the original data now. Since our channelReadTransformation
+                        // is the channel initializer. Once that's done we need to fire the channel as a read
+                        // so that it hits channelRead0 in the base socket channel.
+                        context.fireChannelRead(data)
+                    case .failure:
+                        break
+                    }
                     self._transformationCompleted(context: context, result: result)
                 }
         }
--- a/Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
+++ b/Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
@@ -1,0 +1,510 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A struct wrapping an ``EventLoop`` that ensures all calls to any method on this struct
+/// are coming from the event loop.
+@usableFromInline
+struct AssumeIsolatedEventLoop: Sendable {
+    @usableFromInline
+    let wrapped: EventLoop
+
+    @inlinable
+    internal init(_ eventLoop: EventLoop) {
+        self.wrapped = eventLoop
+    }
+
+    /// Submit a given task to be executed by the `EventLoop`
+    @inlinable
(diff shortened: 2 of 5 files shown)


FranzBusch:

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,510 @@
...
+            try unsafeTransfer.wrappedValue()
+        }
+    }
+}
+extension EventLoop {
+    /// Assumes the calling context is isolated to the event loop.
+    @usableFromInline
+    func assumeIsolated() -> AssumeIsolatedEventLoop {
+        AssumeIsolatedEventLoop(self)
+    }
```
glbrntt, line 125: I think the naming/behaviour here is a bit surprising. From reading the function name I would assume that _this_ function checks we are isolated to the EL, not the individual functions in `AssumeIsolatedEventLoop`. This is made more confusing by `AssumeIsolatedEventLoop` being `Sendable` too.

I know the pattern is the roughly the same as our sync views, however, the sync views predate `Sendable` (and aren't marked `Sendable`).  Perhaps we should change the pattern for the sync views such that they assert on EL when created and are explicitly not `Sendable`?

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,510 @@
...
+}
+
+extension EventLoopFuture {
+    /// A struct wrapping an ``EventLoopFuture`` that ensures all calls to any method on this struct
+    /// are coming from the event loop of the future.
+    @usableFromInline
+    struct AssumeIsolated: @unchecked Sendable {
```
glbrntt, line 132: I'm a bit surprised that none of the methods on here return the caller back an `AssumeIsolated` future? This would make chaining much easier. Also, similar to above, it feels strange that we can move the `AssumeIsolated` future across isolation domains because it's `Sendable`. I think we want to make the assumption when we create it and have it not be `Sendable`.

glbrntt (changes requested):

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: I agree that those types shouldn't be `Sendable`. Pushed the changes up.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,510 @@
...
+        return AssumeIsolated(wrapped: self)
+    }
+}
+
+
+extension EventLoopPromise {
+    /// A struct wrapping an ``EventLoopPromise`` that ensures all calls to any method on this struct
+    /// are coming from the event loop of the promise.
+    @usableFromInline
+    struct AssumeIsolated {
```
glbrntt, line 467: > I intentionally did not return the AssumeIsolated from the methods but rather the regular types since those types contain way more methods and I don't want to duplicate everything.

In that case why don't we have an `unwrap` API or similar which returns the non-isolated future? That makes the API symmetric when the isolation changes: i.e. you explicitly have to do so.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,510 @@
...
+    func assumeIsolated() -> AssumeIsolatedEventLoop {
+        AssumeIsolatedEventLoop(self)
+    }
+}
+
+extension EventLoopFuture {
+    /// A struct wrapping an ``EventLoopFuture`` that ensures all calls to any method on this struct
+    /// are coming from the event loop of the future.
+    @usableFromInline
+    struct AssumeIsolated {
```
glbrntt, line 138: I think we should drop the `Assume` from the name here. None of the other sync views have assume/assert in their name.

Sources/NIOCore/AsyncChannel/AsyncChannelInboundStreamChannelHandler.swift
```diff
@@ -135,20 +135,19 @@ internal final class NIOAsyncChannelInboundStreamChannelHandler<InboundIn: Senda
...
+                .assumeIsolated()
                 .map { result -> ProducerElement in
-                    context.eventLoop.preconditionInEventLoop()
                     // We have to fire through the original data now. Since our channelReadTransformation
                     // is the channel initializer. Once that's done we need to fire the channel as a read
                     // so that it hits channelRead0 in the base socket channel.
                     context.fireChannelRead(data)
                     return result
                 }
+                .assumeIsolated()
```
glbrntt: FWIW in this case we could just ditch the `map` and switch on the result in `whenComplete`

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,528 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A struct wrapping an ``EventLoop`` that ensures all calls to any method on this struct
+/// are coming from the event loop.
+@usableFromInline
+struct AssumeIsolatedEventLoop {
```
glbrntt, line 18: Can we drop `Assume` from the name?

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,528 @@
...
+extension EventLoop {
+    /// Assumes the calling context is isolated to the event loop.
+    @usableFromInline
+    func assumeIsolated() -> AssumeIsolatedEventLoop {
+        AssumeIsolatedEventLoop(self)
+    }
+}
+
+extension EventLoopFuture {
+    /// A struct wrapping an ``EventLoopFuture`` that ensures all calls to any method on this struct
```
glbrntt, line 135: nit: the docs on the methods should be updated too

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,528 @@
...
+    /// are coming from the event loop of the future.
+    @usableFromInline
+    struct AssumeIsolated {
+        @usableFromInline
+        let wrapped: EventLoopFuture<Value>
```
glbrntt, line 140: given `nonisolated()` exists now, should we prefix this with an underscore to avoid accidental usage?

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -0,0 +1,528 @@
...
+        @usableFromInline
+        let wrapped: EventLoopPromise<Value>
+
+
+        /// Deliver a successful result to the associated `EventLoopFuture<Value>` object.
+        ///
+        /// - parameters:
+        ///     - value: The successful result of the operation.
+        @inlinable
+        func succeed(_ value: Value) {
```
glbrntt, line 489: No `fail` counterpart?

glbrntt (commented):
A few nits but otherwise I think this looks pretty good

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: We don't need `fail` counter part since that doesn't require `Sendable` values. I really wanted to avoid replicating all methods on the new types.

glbrntt (approved):
LFTM module the `fail` on the promise. Not a hill I'll die on though.

PR #2658, opened 2024-02-19
Remove large temp file after FS tests

glbrntt:
Motivation:

One of the file system tests copies a large file but didn't clear up after itself.

Modifications:

- Remove the files after the test

Result:

Less disk space used

### DIFF
--- a/Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
+++ b/Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
@@ -562,6 +562,11 @@ final class FileSystemTests: XCTestCase {
     func testCopyLargeFile() async throws {
         let sourcePath = try await self.fs.temporaryFilePath()
         let destPath = try await self.fs.temporaryFilePath()
+        self.addTeardownBlock {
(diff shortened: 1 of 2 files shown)


FranzBusch:
@yim-lee it seems like most of our nodes are out of disk space. Could you help us identify what is using the disk space and clean up the nodes?

glbrntt:
> @yim-lee it seems like most of our nodes are out of disk space. Could you help us identify what is using the disk space and clean up the nodes?

I _think_ this should be fixed now, the docker-compose was temporarily modified to blast away `/tmp` which we think was getting filled up by large leftover files from this test.

Is there a way we could have `/tmp` cleared periodically or just have the nodes restarted periodically?

yim-lee:
I just checked, the lowest disk space right now that one agent has is 10GB, the rest has at least 30GB.

> Is there a way we could have /tmp cleared periodically or just have the nodes restarted periodically?

I think the optimal thing to do is what this PR does, i.e. for tests to clean up after themselves. We might do a one-time cleanup of `/tmp` to reclaim more disk space though. cc @tomerd

PR #2660, opened 2024-02-20
Build the new FileSystem module for Android

finagolfin:
### Motivation:

Get the new module [building and tested on my Android CI again](https://github.com/finagolfin/swift-android-sdk/actions/runs/7970885103) (I had to make a few extra modifications there to get this working with the older Android API 24 that I test on my CI)

### Modifications:

- Add force unwraps where needed
- Update C types and account for nullability annotations in NDK 26
- Use /data/local/tmp for the `temporaryDirectory` and invoke it in the tests, instead of /tmp directly
- Don't copy over extended attributes
- Android doesn't support hard-linking on most partitions, so don't use it for materializing temporary files

### Result:

Everything works again on Android

@glbrntt, please review.

### DIFF
--- a/Sources/NIOFileSystem/FileSystem.swift
+++ b/Sources/NIOFileSystem/FileSystem.swift
@@ -625,6 +625,8 @@ public struct FileSystem: Sendable, FileSystemProtocol {
                     )
                 }.get()
             }
+            #elseif os(Android)
+            return "/data/local/tmp"
             #else
             return "/tmp"
             #endif
@@ -959,6 +961,7 @@ extension FileSystem {
                 permissions: info.permissions
             )
 
+            #if !os(Android)
             // Copy over extended attributes, if any exist.
             do {
                 let attributes = try await dir.attributeNames()
@@ -979,6 +982,7 @@ extension FileSystem {
                 // that is the case.
                 ()
             }
+            #endif
 
             // Build a list of directories to copy over. Do this after closing the current
             // directory to avoid using too many descriptors.
--- a/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
+++ b/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
@@ -239,7 +239,7 @@ public enum Syscall {
         size: Int
     ) -> Result<Int, Errno> {
         valueOrErrno(retryOnInterrupt: false) {
-            system_sendfile(output.rawValue, input.rawValue, offset, size)
+            system_sendfile(output.rawValue, input.rawValue, off_t(offset), size)
         }
     }
     #endif
@@ -347,7 +347,11 @@ public enum Libc {
         return valueOrErrno {
             pathBytes.withUnsafeMutableBufferPointer { pointer in
                 // The array must be terminated with a nil.
+                #if os(Android)
+                libc_fts_open([pointer.baseAddress!, unsafeBitCast(0, to: UnsafeMutablePointer<CInterop.PlatformChar>.self)], options.rawValue)
+                #else
                 libc_fts_open([pointer.baseAddress, nil], options.rawValue)
+                #endif
             }
         }
     }
--- a/Sources/NIOFileSystem/Internal/SystemFileHandle.swift
+++ b/Sources/NIOFileSystem/Internal/SystemFileHandle.swift
@@ -1116,16 +1116,22 @@ extension SystemFileHandle {
         if delayMaterialization {
             // When opening in this mode we can more "atomically" create the file, that is, by not
             // leaving the user with a half written file should e.g. the system crash or throw an
-            // error while writing. On Linux we do this by opening the directory for the path
-            // with `O_TMPFILE` and creating a hard link when closing the file. On other platforms
-            // we generate a dot file with a randomised suffix name and rename it to the
+            // error while writing. On non-Android Linux we do this by opening the directory for
+            // the path with `O_TMPFILE` and creating a hard link when closing the file. On other
+            // platforms we generate a dot file with a randomised suffix name and rename it to the
             // destination.
+            #if os(Android)
+            let temporaryHardLink = false
+            #else
+            let temporaryHardLink = true
+            #endif
             return Self.syncOpenWithMaterialization(
                 atPath: path,
                 mode: mode,
                 options: options,
                 permissions: permissions,
-                executor: executor
+                executor: executor,
+                useTemporaryFileIfPossible: temporaryHardLink
             )
         } else {
             return Self.syncOpen(
--- a/Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
+++ b/Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
@@ -34,7 +34,7 @@ final class FileHandleTests: XCTestCase {
         autoClose: Bool = true,
         _ execute: @Sendable (SystemFileHandle) async throws -> Void
     ) async throws {
-        let path = FilePath("/tmp/\(Self.temporaryFileName())")
+        let path = try await FilePath("\(FileSystem.shared.temporaryDirectory)/\(Self.temporaryFileName())")
         defer {
             // Remove the file when we're done.
             XCTAssertNoThrow(try Libc.remove(path).get())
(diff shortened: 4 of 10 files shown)


Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -1116,17 +1116,28 @@ extension SystemFileHandle {
...
+            // platforms we generate a dot file with a randomised suffix name and rename it to the
             // destination.
+            #if os(Android)
+            return Self.syncOpenWithMaterialization(
+                atPath: path,
+                mode: mode,
+                options: options,
+                permissions: permissions,
+                executor: executor,
+                useTemporaryFileIfPossible: false
```
finagolfin, line 1130: [No hard links on most Android partitions](https://stackoverflow.com/questions/32365690/is-android-m-not-allowing-hard-links).

finagolfin:
5.8 test run timed out for some reason, while nightly CI has been broken for awhile.

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -17,6 +17,7 @@
 import Atomics
 import NIOCore
 @preconcurrency import SystemPackage
+import class Foundation.FileManager
```
FranzBusch, line 20: The `NIOFilsystem` module is intentionally not depending on `Foundation`. Is there a way to solve this without using `Foundation` here?

finagolfin: I didn't know that, will revert this.

finagolfin: Done

FranzBusch (approved):
LGTM but I would like @glbrntt to also take a quick look before we merge this

Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
```diff
@@ -347,7 +347,11 @@ public enum Libc {
         return valueOrErrno {
             pathBytes.withUnsafeMutableBufferPointer { pointer in
                 // The array must be terminated with a nil.
+                #if os(Android)
+                libc_fts_open([pointer.baseAddress!, unsafeBitCast(0, to: UnsafeMutablePointer<CInterop.PlatformChar>.self)], options.rawValue)
```
glbrntt, line 351: Is `unsafeBitCast(0, to: UnsafeMutablePointer<CInterop.PlatformChar>.self)` any different to `UnsafeMutablePointer(bitPattern: 0)`?

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -1116,17 +1116,28 @@ extension SystemFileHandle {
...
             // leaving the user with a half written file should e.g. the system crash or throw an
-            // error while writing. On Linux we do this by opening the directory for the path
-            // with `O_TMPFILE` and creating a hard link when closing the file. On other platforms
-            // we generate a dot file with a randomised suffix name and rename it to the
+            // error while writing. On non-Android Linux we do this by opening the directory for
+            // the path with `O_TMPFILE` and creating a hard link when closing the file. On other
+            // platforms we generate a dot file with a randomised suffix name and rename it to the
             // destination.
+            #if os(Android)
+            return Self.syncOpenWithMaterialization(
```
glbrntt, line 1124: Rather than duping the whole call, can we conditionalise just `useTemporaryFileIfPossible`? i.e.

```swift
#if os(Android)
let useTemporaryFileIfPossible = false
#else
let useTemporaryFileIfPossible = true
#endif

Makes it a little bit easier to see what the differences are.
```

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
```diff
@@ -34,7 +34,7 @@ final class FileHandleTests: XCTestCase {
         autoClose: Bool = true,
         _ execute: @Sendable (SystemFileHandle) async throws -> Void
     ) async throws {
-        let path = FilePath("/tmp/\(Self.temporaryFileName())")
+        let path = FilePath("\(FileManager.default.temporaryDirectory.path)/\(Self.temporaryFileName())")
```
glbrntt, line 37: Can we  use `FileSystem.shared.temporaryDirectory` instead of `FileManager`?

glbrntt (changes requested):
Looks good, left a couple of nits.

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
finagolfin: Unfortunately not, as that uses a hard-coded `/tmp`, which doesn't exist on Android. I had [originally modified to use Foundation also](https://github.com/apple/swift-nio/commit/dc83c6bc54c9658b63c1dfbc4568c3d0ee41160d#diff-c0a9b635bb72d291323e5d46ff5cebff0a1bd2ed979612d2df50ab90cbef9696R630), but [removed it when asked above](https://github.com/apple/swift-nio/pull/2660#discussion_r1495500012). I figured it was fine here in the tests, since they're already using Foundation, which is why I didn't have to add an import here.

glbrntt: So `FileSystem.shared.temporaryDirectory` returns a bogus path on Android, we should fix that instead of resorting on `FileManager` here if possible. How does Foundation get the right tmp path on Android?

Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
finagolfin: The Swift compiler gave me the same recommendation weeks ago:
```
swift-nio/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift:351:54: warning: 'unsafeBitCast' from 'Int' to 'UnsafeMutablePointer<CInterop.PlatformChar>' (aka 'UnsafeMutablePointer<Int8>') can be replaced with 'bitPattern:' initializer on 'UnsafeMutablePointer<CInterop.PlatformChar>' (aka 'UnsafeMutablePointer<Int8>')
                libc_fts_open([pointer.baseAddress!, unsafeBitCast(0, to: UnsafeMutablePointer<CInterop.PlatformChar>.self)], options.rawValue)
                                                     ^~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                     UnsafeMutablePointer<CInterop.PlatformChar>(bitPattern:  )
```
so I tried it, replacing this line with
```
libc_fts_open([pointer.baseAddress!,  UnsafeMutablePointer<CInterop.PlatformChar>(bitPattern: 0)], options.rawValue)
```
That was a no go for compilation:
```
swift-nio/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift:351:55: error: cannot convert value of type 'UnsafeMutablePointer<CInterop.PlatformChar>?' (aka 'Optional<UnsafeMutablePointer<Int8>>') to expected element type 'UnsafeMutablePointer<CInterop.PlatformChar>' (aka 'UnsafeMutablePointer<Int8>')
                libc_fts_open([pointer.baseAddress!,  UnsafeMutablePointer<CInterop.PlatformChar>(bitPattern: 0)], options.rawValue)
                                                      ^
```
OK, so let's try force-unwrapping instead with this line and see what happens:
```
libc_fts_open([pointer.baseAddress!,  UnsafeMutablePointer<CInterop.PlatformChar>(bitPattern: 0)!], options.rawValue)
```
As might be expected, tests now fail at runtime:
```
NIOFileSystem/Syscall.swift:351: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```
That's why I used this bitcast instead.

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
glbrntt: Can we just default to "/data/local/tmp" on Android then? The comment suggests TMPDIR is rarely set, so this would at least produce a less wrong answer on Android than "/tmp".

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
finagolfin: Done

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
finagolfin: Done

glbrntt (approved):
Thanks!

glbrntt:

PR #2661, opened 2024-02-20
Migrate to `syncOperations` in more places

FranzBusch:
# Motivation

Adding a handler via the regular `pipeline.addHandler` APIs can lead to transferring the handler over isolation regions. To avoid running into `Sendable` warnings here we can use the `syncOperations` of the pipeline to avoid transferring the handlers.

# Modification

This PR is mostly changing code to avoid transferring handlers to and from the event loops in tests. This should be a no-op mostly.


Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1082,6 +1082,23 @@ extension ChannelPipeline {
             try self._pipeline.addHandlersSync(handlers, position: position).get()
         }
 
+        /// Remove a `ChannelHandler` from the `ChannelPipeline`.
+        ///
+        /// - parameters:
+        ///     - handler: the `ChannelHandler` to remove.
+        /// - returns: the `EventLoopFuture` which will be notified once the `ChannelHandler` was removed.
+        @preconcurrency
+        public func removeHandler(_ handler: RemovableChannelHandler) -> EventLoopFuture<Void> {
```
FranzBusch, line 1091: This is the most relevant change in this PR since it adds a new `removeHandler` method to the `syncOperations`. Importantly this still returns a future since removing is asynchronous but it requires the call to  this method to be from the event loop of the pipeline.

Tests/NIOPosixTests/BootstrapTest.swift
```diff
@@ -448,7 +448,7 @@ class BootstrapTest: XCTestCase {
         }
 
         // But now, it should be there.
-        XCTAssertNoThrow(_ = try server.pipeline.context(name: "AcceptHandler").wait())
+        XCTAssertNoThrow(_ = try server.pipeline.context(name: "AcceptHandler").map { _ in () }.wait())
```
gjcairo, line 451: Why is this `map` needed?

gjcairo (approved):
LGTM, just left a small question

Tests/NIOPosixTests/BootstrapTest.swift
FranzBusch: Because the handler is not `Sendable` so we get a warning when we `wait` for it. So mapping to Void avoids that :)

Sources/NIOAsyncAwaitDemo/main.swift
```diff
@@ -18,10 +18,16 @@ import Dispatch
...
-    try await channel.pipeline.addHTTPClientHandlers().get()
-    try await channel.pipeline.addHandler(NIOHTTPClientResponseAggregator(maxContentLength: 1_000_000))
-    try await channel.pipeline.addHandler(MakeFullRequestHandler())
+    let channel = try await ClientBootstrap(group: group)
+        .channelInitializer { channel in
+            channel.eventLoop.makeCompletedFuture {
+                try channel.pipeline.syncOperations.addHTTPClientHandlers()
+                try channel.pipeline.syncOperations.addHandler(NIOHTTPClientResponseAggregator(maxContentLength: 1_000_000))
+                try channel.pipeline.syncOperations.addHandler(MakeFullRequestHandler())
+            }
```
Lukasa, line 27: The closure here might well allocate. Probably more complex than just returning a succeeded void future.

Sources/NIOChatServer/main.swift
```diff
@@ -125,9 +125,9 @@ let bootstrap = ServerBootstrap(group: group)
     // Set the handlers that are applied to the accepted Channels
     .childChannelInitializer { channel in
         // Add handler that will buffer data until a \n is received
-        channel.pipeline.addHandler(ByteToMessageHandler(LineDelimiterCodec())).flatMap { v in
-            // It's important we use the same handler for all accepted channels. The ChatHandler is thread-safe!
-            channel.pipeline.addHandler(chatHandler)
+        channel.eventLoop.makeCompletedFuture {
```
Lukasa, line 128: Same note here.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -456,6 +456,29 @@ extension EventLoopFuture {
...
+        /// and non-blocking code.
+        ///
+        /// This is also forbidden in async contexts: prefer ``EventLoopFuture/get()``.
+        ///
+        /// - Note: The `Value` must be `Sendable` since it is shared outside of the isolation domain of the event loop.
+        ///
+        /// - returns: The value of the `EventLoopFuture` when it completes.
+        /// - throws: The error value of the `EventLoopFuture` if it errors.
+        @available(*, noasync, message: "wait() can block indefinitely, prefer get()", renamed: "get()")
+        @preconcurrency
```
Lukasa, line 475: I don't think we want `@preconcurrency` here.

Sources/NIOCrashTester/OutputGrepper.swift
```diff
@@ -30,8 +30,10 @@ internal struct OutputGrepper {
         let channelFuture = NIOPipeBootstrap(group: group)
             .channelOption(ChannelOptions.allowRemoteHalfClosure, value: true)
             .channelInitializer { channel in
-                channel.pipeline.addHandlers([ByteToMessageHandler(NewlineFramer()),
-                                              GrepHandler(promise: outputPromise)])
+                channel.eventLoop.makeCompletedFuture {
```
Lukasa, line 33: Same note here on makeCompletedFuture.

Sources/NIOEchoServer/main.swift
```diff
@@ -46,8 +46,9 @@ let bootstrap = ServerBootstrap(group: group)
     // Set the handlers that are appled to the accepted Channels
     .childChannelInitializer { channel in
         // Ensure we don't read faster than we can write by adding the BackPressureHandler into the pipeline.
-        channel.pipeline.addHandler(BackPressureHandler()).flatMap { v in
-            channel.pipeline.addHandler(EchoHandler())
+        channel.eventLoop.makeCompletedFuture {
```
Lukasa, line 49: Same note here on makeCompletedFuture.

Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -119,8 +119,9 @@ final class TCPThroughputBenchmark: Benchmark {
 
         self.clientChannel = try ClientBootstrap(group: group)
             .channelInitializer { channel in
-                channel.pipeline.addHandler(ByteToMessageHandler(StreamDecoder())).flatMap { _ in
-                    channel.pipeline.addHandler(ClientHandler())
+                channel.eventLoop.makeCompletedFuture {
```
Lukasa, line 122: Same note here on `makeCompletedFuture`.

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
```diff
@@ -38,7 +40,7 @@ extension ChannelPipeline {
     fileprivate func assertContainsUpgrader() {
         #if !canImport(Darwin) || swift(>=5.10)
         do {
-            _ = try self.context(handlerType: NIOTypedHTTPServerUpgradeHandler<Bool>.self).wait()
+            _ = try self.context(handlerType: NIOTypedHTTPServerUpgradeHandler<Bool>.self).map { _ in () }.wait()
```
Lukasa, line 43: This pattern is common enough that you might get more mileage to by factoring this out to a `ChannelPipeline.contains` function.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -456,6 +456,29 @@ extension EventLoopFuture {
...
+        @inlinable
+        public func wait(file: StaticString = #file, line: UInt = #line) throws -> Value {
```
Lukasa, line 477: This shouldn't be public.

Lukasa, line 477: Also, this method is a bit of a code smell to me. In almost all cases this is a deterministic crash, because the only place you are _not_ allowed to call `wait` from is the event loop to which the future belongs. The only time it's safe is when you're on `EmbeddedEventLoop`.

A better fix moves this to a helper method on `EmbeddedEventLoop`.

Tests/NIOPosixTests/IdleStateHandlerTest.swift
```diff
@@ -79,8 +79,9 @@ class IdleStateHandlerTest : XCTestCase {
         let serverChannel = try assertNoThrowWithValue(ServerBootstrap(group: group)
             .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
             .childChannelInitializer { channel in
-                channel.pipeline.addHandler(handler).flatMap { f in
-                    channel.pipeline.addHandler(TestWriteHandler(writeToChannel, assertEventFn))
+                channel.eventLoop.makeCompletedFuture {
```
Lukasa, line 82: Same makeCompletedFuture note here.

Lukasa (changes requested):

FranzBusch:
@Lukasa Addressed all your comments except the `makeCompletedFuture` once since the work around here is either a manual `do/catch` block or we can overload the `makeCompletedFuture` method for the `Void` case to optimise the allocation to only happen in the error path.

FranzBusch:
@Lukasa I updated the allocation numbers for the nighty. Those have been wrong since the nightly was crashing for a long time, hence, I didn't update them last time. The breaking API is expected and not really breaking from what I can see. Can you merge this PR once the CI has passed?

```
10:56:30   💔 API breakage: func ChannelPipeline.addHandler(_:name:position:) is now with @preconcurrency
10:56:30   💔 API breakage: func ChannelPipeline.addHandler(_:name:position:) has parameter 0 type change from NIOCore.ChannelHandler to NIOCore.ChannelHandler & Swift.Sendable
10:56:30   💔 API breakage: func ChannelPipeline.removeHandler(_:) is now with @preconcurrency
10:56:30   💔 API breakage: func ChannelPipeline.removeHandler(_:) has parameter 0 type change from NIOCore.RemovableChannelHandler to NIOCore.RemovableChannelHandler & Swift.Sendable
10:56:30   💔 API breakage: func ChannelPipeline.context(handler:) is now with @preconcurrency
10:56:30   💔 API breakage: func ChannelPipeline.context(handler:) has parameter 0 type change from NIOCore.ChannelHandler to NIOCore.ChannelHandler & Swift.Sendable
10:56:30   💔 API breakage: func ChannelPipeline.addHandlers(_:position:) is now with @preconcurrency
10:56:30   💔 API breakage: func ChannelPipeline.addHandlers(_:position:) has parameter 0 type change from [NIOCore.ChannelHandler] to [NIOCore.ChannelHandler & Swift.Sendable]
10:56:30   💔 API breakage: func ChannelPipeline.addHandlers(_:position:) is now with @preconcurrency
10:56:30   💔 API breakage: func ChannelPipeline.addHandlers(_:position:) has parameter 0 type change from NIOCore.ChannelHandler... to (NIOCore.ChannelHandler & Swift.Sendable)...
```

Lukasa (approved):

PR #2669, opened 2024-02-28
Fix CoW performance bug in `NIOThreadPool` work queue

gmilos:
Fix CoW performance bug in `NIOThreadPool` work queue.

### Motivation:

`NIOThreadPool` uses an enum to track it's state. If the pool is running, it'll be in `State.running(CircularBuffer<WorkItem>)`. However this hard to mutate in place reliably. The current code certainly doesn't achieve that, which leads to CoW on each enqueue and dequeue.

### Modifications:

* I added a test to measure the behaviour before and after. It's called `RunIfActiveBenchmark`. It's run in 1 & 8 threads variants, with 100k tasks.
* I encapsulated the `CircularBuffer<WorkItem>` in a class, thus providing a layer of indirection during mutations, which avoid CoW
* I wrapped `WorkItem` which is a closure in a struct wrapped to avoid the cost of allocations discussed in: https://bugs.swift.org/browse/SR-15872

### Result:

Performance is orders of magnitudes greater. The "after" (measured for [`Deque` version](https://github.com/apple/swift-nio/pull/2669/commits/9380a6b1cdc394b9003ca447c193e9c4ccf1198a), on my own laptop [given current perf testing CI infra breakage]), is:
```
measuring: runIfActive_1_thread_100k_tasks: 0.251138375, 0.247212625, 0.257556917, 0.2396585, 0.245091042, 0.240477125, 0.246961292, 0.244519042, 0.240759375, 0.240233208,
measuring: runIfActive_8_threads_100k_tasks: 0.26803725, 0.262704083, 0.263101083, 0.266763792, 0.269062791, 0.267278875, 0.265546208, 0.266732625, 0.265480375, 0.271898916,
```
The before takes way too long to run (hours).

### DIFF
--- a/Sources/NIOPosix/NIOThreadPool.swift
+++ b/Sources/NIOPosix/NIOThreadPool.swift
@@ -57,14 +57,12 @@ public final class NIOThreadPool {
 
     /// The work that should be done by the `NIOThreadPool`.
     public typealias WorkItem = @Sendable (WorkItemState) -> Void
-
     private enum State {
         /// The `NIOThreadPool` is already stopped.
         case stopped
         /// The `NIOThreadPool` is shutting down, the array has one boolean entry for each thread indicating if it has shut down already.
         case shuttingDown([Bool])
-        /// The `NIOThreadPool` is up and running, `WorkItems` contains the yet unprocessed
-        /// `WorkItems`.
+        /// The `NIOThreadPool` is up and running, the `CircularBuffer` containing the yet unprocessed `WorkItems`.
         case running(CircularBuffer<WorkItem>)
         /// Temporary state used when mutating the .running(items). Used to avoid CoW copies.
         /// It should never be "leaked" outside of the lock block.
--- a/Sources/NIOPosix/NIOThreadPool.swift
+++ b/Sources/NIOPosix/NIOThreadPool.swift
@@ -13,23 +13,22 @@
 //===----------------------------------------------------------------------===//
 
 import Dispatch
-import NIOCore
 import NIOConcurrencyHelpers
+import NIOCore
 
 /// Errors that may be thrown when executing work on a `NIOThreadPool`
 public enum NIOThreadPoolError {
 
     /// The `NIOThreadPool` was not active.
-    public struct ThreadPoolInactive: Error { 
+    public struct ThreadPoolInactive: Error {
         public init() {}
     }
 
     /// The `NIOThreadPool` operation is unsupported (e.g. shutdown of a perpetual pool).
-    public struct UnsupportedOperation: Error { 
+    public struct UnsupportedOperation: Error {
         public init() {}
     }
 }
-
 
 /// A thread pool that should be used if some (kernel thread) blocking work
 /// needs to be performed for which no non-blocking API exists.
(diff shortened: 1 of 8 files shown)


weissi (approved):
Thank you! LGTM, letting @Lukasa merge this if he's happy

Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -58,17 +57,46 @@ public final class NIOThreadPool {
 
     /// The work that should be done by the `NIOThreadPool`.
     public typealias WorkItem = @Sendable (WorkItemState) -> Void
+
+    /// `WorkItem`-s encapsulated in a reference type in order to avoid CoW when mutating the queue.
+    /// Since `WorkItems` are embedded in `State` enum, it's hard to reliably avoid CoW otherwise.
+    ///
+    /// Also the closures are wrapped in a struct, to avoid the cost of allocation as discussed on https://bugs.swift.org/browse/SR-15872
+    final class WorkItems {
```
FranzBusch, line 65: Instead of boxing the buffer here with a class we should be able to avoid the CoW by introducing a new state called `modifying` into the `State` enum and before mutating the queue we transition to that state and then back to the running state. We have used that approach in most other places recently and it avoids this class entirely. 

The other upside of this pattern is that once we language gets `mutating/inout` switches in the future we can easily transition over to that and remove the `modifying` case again.

FranzBusch (changes requested):

Sources/NIOPosix/NIOThreadPool.swift
weissi: good catch! shoot, missed the `class`, yes, let's do this properly

gmilos: The `.modifying` solution is 60% slower than the `class`. @FranzBusch @weissi do you still want me to do that?

FranzBusch: Where does the slow down come from and is sprinkling some `@inlinable` + `@usableFromInline` making it go faster?

gmilos: Ah, this confused me also.

Turns out I had a different testing environment. When I was developing the original solution, and performance testing it. I was sitting on top of #2645, which also has an effect, as ELFs are used to notify completions.

In the "this is 60% slower" result, I was sitting on top of now rebased branch from this PR.

My bad. I'll update the results in the PR description to give the numbers for this branch alone (as they should have been). And then we'll see the 60% perf bump on once #2645 goes in.

gmilos: @dnadoba I checked `Deque` and it didn't have any noticeable performance difference. Is it supposed to? And/or are we attempting to move away from `CircularBuffer`? Pushing `Deque`-s version, so that we have all the different paints on the PR :).

dnadoba: We are moving away from `CircularBuffer`. We are using `Deque` in all new code e.g. in our new custom AsyncSequences. 
Both are heavily optimised but `Deque` is the way forward. This isn't urgent and we will keep `CircularBuffer` around but it isn't likely to be updated. All new performance optimisations, if any are possible, should be done on top of `Deque` as it is independent of NIO and therefore more generally useful.

FranzBusch (approved):

Sources/NIOPosix/NIOThreadPool.swift
gmilos: @dnadoba got it, thanks. Will keep this in mind for the future. This code could be trivially switched to `Deque` so it's already in.

PR #2671, opened 2024-03-04
Add support for WASILibc

MaxDesiatov:
Dispatch is not supported on WASI, and only Unix domain sockets are supported, which means we have to exclude those APIs on this platform.

There's work in progress to enable tests for this on CI, but nothing I can provide for this PR at the current moment.


Sources/NIOCore/FileHandle.swift
```diff
@@ -138,13 +142,15 @@ extension NIOFileHandle {
         public static let defaultPermissions = S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH
 #endif
 
+#if !os(WASI)
```
kateinoigakukun, line 145: wasi-libc has `mode_t` and `O_CREAT` but `O_CREAT` cannot be imported through ClangImporter for the same reason as `CLOCK_MONOTONIC`. Can we make a shim for it too?

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -191,6 +193,11 @@ extension NIOBSDSocket.AddressFamily {
 
 // Protocol Family
 extension NIOBSDSocket.ProtocolFamily {
+#if os(WASI)
+    /// UNIX local to the host.
+    public static let unix: NIOBSDSocket.ProtocolFamily =
+            NIOBSDSocket.ProtocolFamily(rawValue: 1)
```
kateinoigakukun, line 199: `PF_INET` and `PF_INET6` are available but `PF_UNIX` isn't. 1 is bound to `PF_INET`. https://github.com/WebAssembly/wasi-libc/blob/09683b3623ec49f79899fc1d847f29e26af80fd1/libc-bottom-half/headers/public/__header_sys_socket.h#L29-L30

Sources/NIOCore/Interfaces.swift
```diff
@@ -62,6 +64,7 @@ private extension ifaddrs {
 }
 #endif
 
+#if !os(WASI)
```
kateinoigakukun, line 67: Can we keep the type itself available but mark `init` unavailable to reduce subsequent ifdef?

Sources/NIOCore/SocketAddresses.swift
```diff
@@ -279,6 +295,7 @@ public enum SocketAddress: CustomStringConvertible, Sendable {
         self = .v6(.init(address: addr, host: host))
     }
 
+#if !os(WASI)
```
kateinoigakukun, line 298: `sockaddr_in` and `sockaddr_in6` are available in recent wasi-libc.

Package.swift
```diff
@@ -15,6 +15,7 @@
 
 import PackageDescription
 
+
```
Lukasa: Nit: remove the newline.

Package.swift
```diff
@@ -151,7 +160,7 @@ let package = Package(
         .target(
             name: "NIOHTTP1",
             dependencies: [
-                "NIO",
+                .target(name: "NIO", condition: .when(platforms: historicalNIOPosixDependencyRequired)),
```
Lukasa, line 162: This condition should probably be persisted across almost any library target that depends on `NIO` which also depends on `NIOCore`.

`_NIOConcurrency`, `NIOFoundationCompat`, `NIOWebSocket`, `NIOTLS` appear to be the other places.

Package.swift
MaxDesiatov, line 18:
```suggestion
```

Package.swift
```diff
@@ -42,7 +46,7 @@ let package = Package(
         .library(name: "_NIOFileSystemFoundationCompat", targets: ["NIOFileSystemFoundationCompat"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
+        .package(url: "https://github.com/apple/swift-atomics.git", revision: "1f2007ef4165432f59f28615c473eb79a844a2af"),
```
MaxDesiatov, line 49:
```suggestion
        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
```

Lukasa:

Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -45,7 +47,7 @@ public final class Lock {
     public init() {
 #if os(Windows)
         InitializeSRWLock(self.mutex)
-#else
+#elseif !os(WASI)
```
Lukasa, line 50: I don't think we can merge a PR that silently disables these primitives on WASI without being _extremely_ confident that we'll never need them. Is there any way we can replace this with a feature-detection on threading?

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -199,12 +204,14 @@ extension NIOBSDSocket.ProtocolFamily {
     public static let inet6: NIOBSDSocket.ProtocolFamily =
             NIOBSDSocket.ProtocolFamily(rawValue: PF_INET6)
 
+#if !os(WASI)
```
Lukasa, line 207: My general preference here would be that the APIs here are _present_, they just end up failing when actually used in WASI. That would mean hardcoding the syscall value on WASI, or deliberately using a spoiler value. That note applies to the rest of this file.

Sources/NIOCore/EventLoop.swift
```diff
@@ -13,7 +13,12 @@
 //===----------------------------------------------------------------------===//
 
 import NIOConcurrencyHelpers
+#if canImport(Dispatch)
 import Dispatch
+#elseif canImport(WASILibc)
```
Lukasa, line 18: The `elseif` implies a relationship between these two ideas, but I think they're fundamentally unrelated.

Sources/NIOCore/FileHandle.swift
```diff
@@ -54,6 +57,7 @@ public final class NIOFileHandle: FileDescriptor {
         assert(!self.isOpen, "leaked open NIOFileHandle(descriptor: \(self.descriptor)). Call `close()` to close or `takeDescriptorOwnership()` to take ownership and close by some other means.")
     }
 
+#if !os(WASI)
```
Lukasa, line 60: Same note here: generally it's preferable to have APIs not simply be absent, but to misbehave when not present. In this case, `duplicate` will presumably _become_ available, and it'd be nicer if this just threw instead. The only time that isn't my preference is when an API literally cannot be spelled, e.g. the return or argument types are missing.

Relatedly, it's best if we put the `#if os` at the _lowest_ possible level. In this case, we should arrange for `dup` to throw or error out, and let the rest of the code path handle that appropriately, which minimises the OS customisation we have to do.

Sources/NIOCore/SocketAddresses.swift
```diff
@@ -445,6 +456,7 @@ public enum SocketAddress: CustomStringConvertible, Sendable {
         self = .v6(.init(address: ipv6Addr, host: ""))
     }
 
+#if !os(WASI)
```
kateinoigakukun, line 459: Could you make `makeAddressResolvingHost` just throw exception instead of guarding out the API itself?

Sources/NIOCore/SocketOptionProvider.swift
```diff
@@ -140,6 +142,7 @@ public protocol SocketOptionProvider: _NIOPreconcurrencySendable {
 //
 // You are welcome to add more helper methods here, but each helper method you add must be tested.
 extension SocketOptionProvider {
+#if !os(WASI)
```
kateinoigakukun, line 145: How about adding a typealias like the following instead of guarding out the whole method definition?

```swift
#if canImport(WASILibc)
typealias NIOLinger = Never
#else
typealias NIOLinger = linger
#endif

weissi:
@MaxDesiatov / @Lukasa This looks like a PR that has bit rotting potential because it needs to touch a lot of pieces, most of the issues seem pretty tractable to me so I'd love to see this go in.

w.r.t. disabling locks on WASI: Honestly I think this can go in as is: Today this is fine and if threading really happens this will blow up very obviously and can be fixed.

Sources/NIOConcurrencyHelpers/lock.swift
MaxDesiatov: Swift `main` snapshots support `#if _runtime(_multithreaded)` checks, which I'm now using here.

Sources/NIOCore/BSDSocketAPI.swift
MaxDesiatov: What would be the benefit of that as compared to compile-time warnings for absent symbols? My worry is that someone ends up adopting these APIs when targeting WASI and then discovering that they fail at the most inconvenient moment.

If you insist on runtime errors, would you prefer a `fatalError` or a certain sentinel value? In the latter case, how do I pick such value or do we just change `NIOBSDSocket.Option.rawValue` to have `CInt?` type instead of `CInt`?

Sources/NIOCore/FileHandle.swift
MaxDesiatov: Ditto, what's the benefit of preferring obscured runtime errors when we can easily make these compile-time errors and make unavailability obvious to a library user?

Sources/NIOCore/SocketAddresses.swift
MaxDesiatov: I'm unsure why we'd prefer runtime checks as opposed to compile-time checks, but this particular change seems easy, so I pushed it for now.

Sources/NIOCore/BSDSocketAPI.swift
MaxDesiatov: One other argument for compile-time checks is consistency: Windows did it that way, so why WASI shouldn't follow that precedent?

Sources/NIOCore/SocketOptionProvider.swift
MaxDesiatov: Same comment as before, unsure why we'd prefer runtime failures, but this one seemed trivial so I'd considered it addressed

weissi:
> Currently blocked on [swiftlang/swift#75618](https://github.com/swiftlang/swift/pull/75618), which would make `PF_INET` and `PF_INET6` available.

You can just hardcode the values for the time being, right?

MaxDesiatov:
> > Currently blocked on [swiftlang/swift#75618](https://github.com/swiftlang/swift/pull/75618), which would make `PF_INET` and `PF_INET6` available.
> 
> You can just hardcode the values for the time being, right? They are `#define`s and therefore ABI, so they can't change.

If that version bump is easy and we get it done quickly, I'd rather have that one merged and make this PR cleaner at the same time.

MaxDesiatov:
With changes I've pushed the build is now successful with latest development snapshots when building `NIOCore` and `NIOHTTP1` targets for WASI.

MaxDesiatov:
Pushed more changes so that it uses `wasi_pthread` when multi-threading is available with `wasm32-unknown-wasip1-threads` triples.

MaxDesiatov:
> w.r.t. disabling locks on WASI: Honestly I think this can go in as is: Today this is fine and if threading really happens this will blow up very obviously and can be fixed.

@weissi In this PR we now conditionally enable locks when those are available

weissi:
CC @FranzBusch & @Lukasa

FranzBusch (approved):
This LGTM. @Lukasa do you want to review as well?

weissi (approved):
LGTM, thanks @MaxDesiatov !

PR #2677, opened 2024-03-11
Prevent BufferedWriter from producing empty files

stefanadranca:
_Preventing BufferedWriter from producing empty files_

### Motivation:

_In the current implementation, if the users forget to call 'flush()' on the writer when creating a new file with 'withFileHandle(forWritingAt:options:execute)' the newly created file is empty._

### Modifications:

_Implemented the 'withBufferedWriter()' method for the 'WriteFileHandle'. The method calls flush on the bufferedWriter after executing the closure that writes the contents to the buffer as indicated by the user._

### Result:

_Users can now use the 'withBufferedWriter' method instead of instantiating the bufferedWriter themselves, reducing the possibility of omitting 'flush()'._


glbrntt:

Sources/NIOFileSystem/BufferedWriter.swift
```diff
@@ -186,6 +186,25 @@ extension WritableFileHandleProtocol {
             capacity: Int(capacity.bytes)
         )
     }
+    
+    /// Convenience function that creates a buffered reader, executes
+    /// the closure that writes the contents into the buffer and calls 'flush()'.
+    ///
+    /// - Parameters:
+    ///   - body: The closure that writes the contents to the buffer created in this method.
+    ///   - Returns: The result of the executed closure.
```
glbrntt, line 195: `Returns:` shouldn't be indented here, otherwise it's treated as documentation for a parameter

Sources/NIOFileSystem/BufferedWriter.swift
```diff
@@ -186,6 +186,25 @@ extension WritableFileHandleProtocol {
             capacity: Int(capacity.bytes)
         )
     }
+    
+    /// Convenience function that creates a buffered reader, executes
+    /// the closure that writes the contents into the buffer and calls 'flush()'.
+    ///
+    /// - Parameters:
+    ///   - body: The closure that writes the contents to the buffer created in this method.
```
glbrntt, line 194: We're missing docs for the other parameters

glbrntt (changes requested):
Couple of doc nits but looks good otherwise!

glbrntt (approved):

PR #2681, opened 2024-03-13
Add "empty" privacy manifests

glbrntt:
- Add "empty" privacy manifests to all library targets that are depended on (directly or transitively) by library products
- Add resources in Package.swift


simonjbeaumont:
@glbrntt is there someway to use one file, or at least have these all symlink to an e.g. `EmptyPrivacyInfo.xcprivacy` so we're not duplicating this same file everywhere?

Lukasa (approved):

PR #2682, opened 2024-03-13
Handle 'atomically' created files more gracefully when detached

glbrntt:
Motivation:

When a file is created 'atomically' and the descriptor is detached from its handle then the user is provided with the descriptor for a temporary file, and the desired file is never created.

Modifications:

Where supported, atomically created files use a temporary unnamed file which is then linked on close. Elsewhere a temporary named file is created which is renamed on closed.

Temporary unnamed files are linked in detach but not closed. Temporary named files are linked to the desired path before being unlinked.

- Break out the logic for materializing a file created using a temporary unnamed file
- Add syscalls and error mappings for link(2) and unlink(2)

Result:

Atomically created files behave as expected when detached

### DIFF
--- a/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
+++ b/Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
@@ -195,6 +195,29 @@ public enum Syscall {
     #endif
 
     @_spi(Testing)
+    public static func link(
+        from source: FilePath,
+        to destination: FilePath
+    ) -> Result<Void, Errno> {
+        return nothingOrErrno(retryOnInterrupt: false) {
+            source.withPlatformString { src in
+                destination.withPlatformString { dst in
+                    system_link(src, dst)
+                }
+            }
+        }
+    }
+
+    @_spi(Testing)
+    public static func unlink(path: FilePath) -> Result<Void, Errno> {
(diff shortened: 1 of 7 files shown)


Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
```diff
@@ -194,6 +194,29 @@ public enum Syscall {
     }
     #endif
 
+    @_spi(Testing)
+    public static func link(
+        from source: FilePath,
+        to destination: FilePath
+    ) -> Result<Void, Errno> {
+        return nothingOrErrno(retryOnInterrupt: false) {
```
Lukasa, line 202: How sure are we that not retrying here is appropriate?

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -271,7 +271,51 @@ extension SystemFileHandle: FileHandleProtocol {
             switch lifecycle {
             case let .open(descriptor):
                 lifecycle = .detached
-                return descriptor
+
+                // We need to be careful handling files which have delayed materialiszation to avoid
```
glbrntt, line 275:
```suggestion
                // We need to be careful handling files which have delayed materialization to avoid
```

PR #2684, opened 2024-03-18
Bug fix. Thread running at User-interactive quality-of-service class …

trevor-sonic:
Adjust Thread Priority to prevent an error message. 

### Motivation:

_It was displaying this serious error on terminal, I wanted to resolve because I need to use Vapor framework and this dependancy should have been fixed._

### Modifications:

_I changed the Thread Priority_

### Result:

_A massive error message and potential crash is prevented from happening._

### DIFF
--- a/Sources/NIOPosix/NIOThreadPool.swift
+++ b/Sources/NIOPosix/NIOThreadPool.swift
@@ -254,7 +254,25 @@ public final class NIOThreadPool {
             }
         }
 
-        group.wait()
+        // *** BUG FIX ***
+        // This prevents the crash with error message:
+        /*
+         Thread Performance Checker: Thread running at User-interactive quality-of-service class waiting on a lower QoS thread running at Default quality-of-service class. Investigate ways to avoid priority inversions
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
Thanks for submitting a PR! However, this fix is not safe to apply. It’s also not a potential or actual crash: this is merely a thread performance checker warning. Can you file an issue so we can discuss the best way to resolve it?

trevor-sonic:
> Thanks for submitting a PR! However, this fix is not safe to apply. It’s also not a potential or actual crash: this is merely a thread performance checker warning. Can you file an issue so we can discuss the best way to resolve it?

I can file the issue, please let me know how to do it.
I'm using the library in iOS app, it cause a purple warning and to me it is quite serious when I see the message in the log one page long.
Thanks

Lukasa:
I totally understand the feeling of urgency. If you use the “Issues” tab in GitHub it will let you file an issue. Please describe your problem there and we can work out a good approach to fixing it.

PR #2689, opened 2024-03-22
Add a NIOFileSystem shim

glbrntt:
Motivation:

Renaming the `NIOFileSystem` module to `_NIOFileSystem` is a breaking change (which is okay, it's documented to be unstable API). However, users will be presented with a wall of build errors on updating to the next version of NIO. We can make the transition slightly easier for them.

Modifications:

Add back a `NIOFileSystem` module which re-exports `_NIOFileSystem` and add it as a target to the `_NIOFileSystem` product. This makes both modules available to users. We can then remove `NIOFileSystem` after a grace period.

Result:

Users can still build using `NIOFileSystem` while transitioning to `_NIOFileSystem`.

### DIFF
--- a/Sources/_NIOFileSystemExported/Exports.swift
+++ b/Sources/_NIOFileSystemExported/Exports.swift
@@ -1,0 +1,15 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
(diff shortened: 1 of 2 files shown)


simonjbeaumont (approved):

Sources/_NIOFileSystemExported/Exports.swift
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
```
weissi, line 14: can we add a `#warning(...)` here?

weissi: ah no sorry, that won't work

PR #2691, opened 2024-03-25
Clarify Client/Server column headers

davedelong:
Clarify the Client/Server column headers in the README

### Motivation:

As I was reading the README, I was confused about what "Client" and "Server" meant in the tables. For example, `async-http-client` has an ❌ in the "server" column. I thought that meant it couldn't be used in server-side applications, which seemed at odds with what I'd heard about the project.

### Modifications:

Added more descriptive text to the column headers

### Result:

The README is clearer.


Lukasa (approved):
Thanks for the cleanup!

PR #2692, opened 2024-03-25
Use NIOThreadPool in NIOFileSystem

glbrntt:
Motivation:

There are two thread pools: one in NIOFileSystem and one in NIOPosix; we should avoid the duplication.

Modifications:

- Switch to using NIOThreadPool
- Remove the old thread pool and associated code
- Add an init to FileSystem which takes a thread pool

Result:

Less duplication


glbrntt:
API breakage is expected, we removed public-but-spi types from NIOFileSystem:

```
16:11:14   💔 API breakage: class IOExecutor has been removed
16:11:14   💔 API breakage: constructor SystemFileHandle.init(takingOwnershipOf:path:materialization:executor:) has been removed
```

PR #2693, opened 2024-03-25
Added file and line to NIOAsyncWriterError description

stefanadranca:
_Added file and line to NIOAsyncWriterError description and made them `public`_

### Motivation:

_Even though NIOAsyncWriterError captures the file and line of the error they are internal and not printed in the description._

### Modifications:

- made the 2 properties public
- added them in the description

### Result:

_Users will be able to access and see the line number and file where a NIOAsyncWriterError occurred_


Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
```diff
@@ -111,7 +109,7 @@ public struct NIOAsyncWriterError: Error, Hashable, CustomStringConvertible {
 
     @inlinable
     public var description: String {
-        "NIOAsyncWriterError.\(self._code.rawValue)"
+      "\(self.file): \(self.line) NIOAsyncWriterError.\(self._code.rawValue)"
```
glbrntt, line 112: We normally stick file and line at the end: as "<file>:<line>". Also, four spaces of indentation please

glbrntt (changes requested):

Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift
stefanadranca: Just to be sure, do you mean the format should be like this "NIOAsyncWriterError.\(self._code.rawValue) \(self.file):\(self.line)"?

glbrntt (approved):

Lukasa (approved):

PR #2695, opened 2024-04-08
Add privacy manifest

glbrntt:
Motivation:

NIOPosix and NIOFileSystem use stat(2) (or some variation of it). These are "required reason APIs" meaning that their usage and reason must be declared in a privacy manifest in order to be submitted to the App Store.

Modifications:

- Add a privacy manifest to NIOPosix and NIOFileSystem

Result:

Apps using NIO won't be rejected from App Store submission for missing privacy manifests.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -26,10 +26,10 @@ let swiftSystem: PackageDescription.Target.Dependency = .product(
 // This doesn't work when cross-compiling: the privacy manifest will be included in the Bundle and
 // Foundation will be linked. This is, however, strictly better than unconditionally adding the
 // resource.
-#if os(Linux) || os(Android)
-let includePrivacyManifest = false
-#else
+#if canImport(Darwin)
 let includePrivacyManifest = true
+#else
+let includePrivacyManifest = false
 #endif
 
 let package = Package(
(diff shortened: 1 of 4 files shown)


Lukasa (approved):

glbrntt:

Package.swift
```diff
@@ -23,6 +23,14 @@ let swiftSystem: PackageDescription.Target.Dependency = .product(
   condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .linux, .android])
 )
 
+// This doesn't work when cross-compiling: the privacy manifest will be included in the Bundle and
+// Foundation will be linked. This is, however, strictly better than unconditionally adding the
+// resource.
+#if canImport(Darwin)
```
weissi, line 29: This is evaluated on the wrong platform when cross compiling. 

CC @MaxDesiatov , is there a `.when(...)` platform thing?

glbrntt: Correct, and as noted in the comment, if we don't do this the alternative is unconditionally including the resource which is worse.

There's no `.when` condition for resources but other solutions gladly accepted!

MaxDesiatov: Yes, this will break when cross-compiling, but apparently `.when` is not flexible enough either 😞

PR #2699, opened 2024-04-10
Add support for `SWIFTCI_USE_LOCAL_DEPS` convention

kateinoigakukun:
Support ci.swift.org convention

### Motivation:

This package is used by https://github.com/apple/swift-sdk-generator and it needs to be built during build-script pipeline, so we need to make its dependencies follow the CI convention. 

### Modifications:

Changed `Package.swift` to respect `SWIFTCI_USE_LOCAL_DEPS` environment variable and use locally checked out packages if defined.

### Result:

We can use `swift-nio` [2.41.0](https://github.com/apple/swift-nio/releases/tag/2.41.0), which is the first version with dependencies, and later in utils/build-script pipeline.

CC: @MaxDesiatov

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,12 +41,6 @@ let package = Package(
         .library(name: "_NIOFileSystem", targets: ["_NIOFileSystem", "NIOFileSystem"]),
         .library(name: "_NIOFileSystemFoundationCompat", targets: ["_NIOFileSystemFoundationCompat"]),
     ],
-    dependencies: [
-        .package(url: "https://github.com/apple/swift-atomics.git", from: "1.1.0"),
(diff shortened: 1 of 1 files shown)


Lukasa (approved):
SGTM, thanks!

Lukasa:

kateinoigakukun:
@Lukasa Would you mind having a new tag to pin the version including this fix in update-checkout-config.json if possible? 🙏

Lukasa:
Hi @kateinoigakukun: yes, we can do a release, but we don't typically do releases this late in the week, so it won't happen until early next week.

kateinoigakukun:
Sure! Thank you so much.

PR #2707, opened 2024-04-26
Add `ByteBuffer` support to `BufferedWriter`

### LINKED ISSUE #2662: NIOFileSystem's BufferedWriter wants Sequence<UInt8>s and doesn't accept ByteBuffers
NIOFileSystem's BufferedWriter wants Sequence<UInt8>s and doesn't accept ByteBuffers.

Kinda weird I have to do

```
try await writer.write(contentsOf: buffer.readableBytesView)
```

glbrntt:
Motivation:

The `BufferedWriter` in `NIOFileSystem` has API in terms of `some Sequence<UInt8>`, meaning that you need to pass `byteBuffer.readableBytesView` which is a bit inconvenient.

Modifications:

- Add overloads which allow you to pass `ByteBuffer` directly
- Remove redundant availability guards; they are already applied to the type

Result:

- Better API for users
- Resolves #2662

### DIFF
--- a/Sources/NIOFileSystem/BufferedWriter.swift
+++ b/Sources/NIOFileSystem/BufferedWriter.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(Linux) || os(Android)
+import NIOCore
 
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice one.

PR #2708, opened 2024-04-26
Add 'withTemporaryDirectory'

### LINKED ISSUE #2664: NIOFileSystem: Should offer withTemporaryDirectory
NIOFileSystem already offers `createTemporaryDirectory`. But usually the only thing I do with that is remove it at the end. So it'd be ideal to have a `withTemporaryDirectory` which auto-removes it.

glbrntt:
Motivation:

NIOFileSystem can create the path of a temporary directory but doesn't offer any API to create and then remove a directory.

Modifications:

- Add `withTemporaryDirectory`

Result:

- Users can get scoped access to a temporary directory which is subsequently removed for them
- Resolves #2664

### DIFF
--- a/Sources/NIOFileSystem/FileSystemProtocol.swift
+++ b/Sources/NIOFileSystem/FileSystemProtocol.swift
@@ -490,11 +490,11 @@ extension FileSystemProtocol {
     ///   - options: Options used to create the directory.
     ///   - execute: A closure which provides access to the directory and its path.
     /// - Returns: The result of `execute`.
-    public func withTemporaryDirectory<R>(
+    public func withTemporaryDirectory<ReturnType>(
         prefix: FilePath? = nil,
         options: OpenOptions.Directory = OpenOptions.Directory(),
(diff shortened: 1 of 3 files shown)


Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -476,6 +476,42 @@ extension FileSystemProtocol {
...
+    /// the prefix.
+    ///
+    /// The temporary directory, and all of its contents, is removed once `execute` returns.
+    ///
+    /// - Parameters:
+    ///   - prefix: The prefix to use for the path of the temporary directory.
+    ///   - options: Options used to create the directory.
+    ///   - execute: A closure which provides access to the directory and its path.
+    /// - Returns: The result of `execute`.
+    public func withTemporaryDirectory<R>(
```
Lukasa, line 493: Prefer longer type variable names. `ReturnType` is usually better than `R`.

Lukasa (approved):
Generally good, a quick note on generic type parameters if you want to take it.

PR #2709, opened 2024-04-26
Replace 'R' with 'ReturnType'

glbrntt:
Motivation:

Single letter names are discouraged but are used in a handful of places in NIOFileSystem.

Modifications:

- Replace 'R' with 'ReturnType' where appropriate

Result:

Clearer APIs

### DIFF
--- a/Sources/NIOFileSystem/BufferedWriter.swift
+++ b/Sources/NIOFileSystem/BufferedWriter.swift
@@ -231,11 +231,11 @@ extension WritableFileHandleProtocol {
     ///     buffer to the file system when it exceeds this capacity. Defaults to 512 KiB.
     ///   - body: The closure that writes the contents to the buffer created in this method.
     /// - Returns: The result of the executed closure.
-    public func withBufferedWriter<R>(
+    public func withBufferedWriter<Result>(
         startingAtAbsoluteOffset initialOffset: Int64 = 0,
         capacity: ByteCount = .kibibytes(512),
-        execute body: (inout BufferedWriter<Self>) async throws -> R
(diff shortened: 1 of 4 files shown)


FranzBusch (commented):
One nit. In other places such as the `NIOAsyncChannel` or the `NonBlockingFileIO` we use `Result` for the generic type instead of `ReturnType`. If we want to be consistent across the project we should use `Result`

glbrntt:
> One nit. In other places such as the `NIOAsyncChannel` or the `NonBlockingFileIO` we use `Result` for the generic type instead of `ReturnType`. If we want to be consistent across the project we should use `Result`.
> 
> The only public API for `ReturnType` was `AsyncTestingEventLoop` from what I found.

Not a fan of `Result`, too similar to `Swift.Result`. I picked `ReturnType` as Cory suggested it in the previous PR. I don't mind too much though, can change to `Result` if you prefer.

FranzBusch:
I just wanted to point out that `Result` is the spelling we used most across the project. Personally I am not a fan of seeing `Type` in generic names. But this is not a hill I want to die on since it is barely user visible.

FranzBusch:
One more point of reference the `Mutex` proposal also uses `Result`

```
  public borrowing func withLock<Result: ~Copyable, E: Error>(
    _ body: (transferring inout State) throws(E) -> transferring Result
  ) throws(E) -> transferring Result
```

glbrntt:
Sounds like you prefer `Result`. Will change to that.

FranzBusch (approved):

glbrntt:
API breaks are expected

PR #2711, opened 2024-05-01
improvement: pass down file/line in makeFutureWithTask

Austinpayne:
In some circumstances it can be difficult to know which specific call to `makeFutureWithTask` leaked a promise. Passes through file/line to help aid debugging such situations.


Lukasa (changes requested):
Thanks for this! This is a good change, however we can't accept it as-is because it's API breaking. We need to keep the old function around, which also means we need a slight name change for the new one, along with a deprecation for the old one and a good fixit.

How about `makeFutureWithResultOfTask`?

Austinpayne:
@Lukasa Thanks, forgot about that detail! I've deprecated the old version as well, let me know if you'd like that removed.

PR #2713, opened 2024-05-03
Changed occurrences of two "no non-blocking" to "blocking" in comments

ladiesman218:
Seems like "blocking" is easier to understand compare to the original "no non-blocking".
Feel free to reject & close this PR though.


Lukasa (changes requested):
Thanks for this patch! The wrinkle here is that this change actually modifies the meaning of the sentence. We specifically should use this API only when there isn't a non-blocking API available. If there _is_, it doesn't matter whether there's a blocking one, users should call the non-blocking one.

PR #2714, opened 2024-05-03
Make EventLoopPromise conform to Equatable

gjcairo:
This PR conforms `EventLoopPromise` to the `Equatable` protocol.

### Motivation:

`EventLoopPromise` is not currently `Equatable`, whereas the underlying `EventLoopFuture` is.
For consistency, it makes sense for `EventLoopPromise` to conform to `Equatable` too.

### Modifications:

Conform `EventLoopPromise` to `Equatable`. This is trivial, as the only property is the underlying `EventLoopFuture`, which is already `Equatable`.

### Result:

`EventLoopPromise` is now `Equatable`.


Lukasa (changes requested):
Mind adding some tests here Gus?

dnadoba (changes requested):
This is a bad idea. Promise can change at any time, violating all requirements of Equatable, including the most basic one:
a == a is always true (Reflexivity)

we should deprecate the existing conformance.

dnadoba (approved):
Never mind, it is based on identity, not on the actual value of the promise which is fine.

Lukasa (approved):

PR #2715, opened 2024-05-07
Release file handles back to caller on failure to take ownership

PeterAdams-A:
Motivation:

When taking ownership of input and output file descriptors in a bootstrap the ownership of the file handles remains with the caller in the function taking ownership fails.  This is currently where _takingOwnershipOfDescriptors uses NIOFileHandle to take ownership of descriptors but then experiences a later failure.

Modifications:

If an exception is throw in _takingOwnershipOfDescriptors release file descriptors from NIOFileHandle.

Result:

No crash on failure to close file handles before end of scoped usage.


PeterAdams-A:
I can't think of a good way to unit test this.  You can force a failure in the right place by passing in an already closed file handle as follows.

`
let sock = socket(AF_LOCAL, SOCK_STREAM, 0)
close(sock)

let channelFuture = NIOPipeBootstrap(group: NIOSingletons.posixEventLoopGroup).takingOwnershipOfDescriptor(inputOutput: sock)
`

Using a closed file descriptor of course is almost certain to result in a flaky test, especially if run in parallel.
Using numbers of file descriptors which are out of range or have ever been used causes failure earlier in the call stack.

PeterAdams-A:

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -2281,11 +2281,18 @@ extension NIOPipeBootstrap {
...
-            )
+            do {
+                channel = try PipeChannel(
+                    eventLoop: eventLoop as! SelectableEventLoop,
+                    inputPipe: inputFileHandle,
+                    outputPipe: outputFileHandle
+                )
+            } catch {
+                // Release file handles back to the caller in case of failure.
+                _ = try? inputFileHandle?.takeDescriptorOwnership()
```
weissi, line 2292: This is doing the opposite of the comment. It takes the ownership instead of leaving it with the caller.

If this method throws it means that NIO has _not_ taken ownership, so the user keeps it.

This change doesn't seem right, the old code seems correct. What's the issue with it?

Lukasa: We take ownership of the handle when we wrap it in `NIOFileHandle` on lines 2282 and 2283. If the `PipeChannel` constructor throws, it will not close the `NIOFileHandle`s (as it isn't fully constructed at the time). That will cause them to hit their `deinit` (as they go out of scope here) and crash.

Put another we: we _took_ ownership at 2282: if we throw here, we have to "un-take" it, and this is how you spell that on `NIOFileHandle`.

PeterAdams-A:
@Lukasa Let me know how you feel about the test I've tried adding.  I looked into hooking sys calls but I believe the current sys call hooking doesn't touch this area at all yet.  In order to hook the call made directly from PipeChannel I think I'd need to add a hooking mechanism to that which would involved changing the construction of that.  Given the failure is from the construction of PipeChannel feels simpler to just throw the exception directly from a hooking factory for that.

Interested to hear your thoughts.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -2327,3 +2348,17 @@ extension NIOPipeBootstrap {
 
 @available(*, unavailable)
 extension NIOPipeBootstrap: Sendable {}
+
+protocol NIOPipeBootstrapHooks {
+    func makePipeChannel(eventLoop: SelectableEventLoop,
+                         inputPipe: NIOFileHandle?,
+                         outputPipe: NIOFileHandle?) throws -> PipeChannel;
```
Lukasa, line 2355:
```suggestion
                         outputPipe: NIOFileHandle?) throws -> PipeChannel
```

Tests/NIOPosixTests/BootstrapTest.swift
```diff
@@ -680,6 +680,30 @@ class BootstrapTest: XCTestCase {
...
+        defer {
+            close(sock)
+        }
+        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            try! elg.syncShutdownGracefully()
+        }
+
+        let bootstrap = NIOPipeBootstrap(validatingGroup: elg, hooks: NIOPipeBootstrapHooksChannelFail())
+            XCTAssertNotNil(bootstrap)
```
Lukasa, line 702:
```suggestion
        XCTAssertNotNil(bootstrap)
```

Tests/NIOPosixTests/BootstrapTest.swift
```diff
@@ -680,6 +680,30 @@ class BootstrapTest: XCTestCase {
...
+        struct NIOPipeBootstrapHooksChannelFail: NIOPipeBootstrapHooks {
+            func makePipeChannel(eventLoop: NIOPosix.SelectableEventLoop, inputPipe: NIOCore.NIOFileHandle?, outputPipe: NIOCore.NIOFileHandle?) throws -> NIOPosix.PipeChannel {
+                throw IOError(errnoCode: EBADF, reason: "testing")
+            }
+        }
+
+        let sock = socket(AF_LOCAL, SOCK_STREAM, 0)
+        defer {
+            close(sock)
+        }
```
Lukasa, line 695: I don't think we're using this socket anywhere.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -2327,3 +2348,17 @@ extension NIOPipeBootstrap {
...
+}
+
+fileprivate struct DefaultNIOPipeBootstrapHooks : NIOPipeBootstrapHooks {
```
Lukasa, line 2358:
```suggestion
fileprivate struct DefaultNIOPipeBootstrapHooks: NIOPipeBootstrapHooks {
```

Lukasa (commented):
I don't hate this testing strategy, but it does need a little cleanup.

Tests/NIOPosixTests/BootstrapTest.swift
PeterAdams-A: Used on line 704 - `let channelFuture = bootstrap?.takingOwnershipOfDescriptor(inputOutput: **sock**)`

Lukasa (approved):

PR #2717, opened 2024-05-07
No longer need test discovery command line.

PeterAdams-A:
Motivation:

Command line argument has not been needed for several swift version.

Modifications:

This causes warnings in recent swift versions.

Result:

No more warnings about swift command line when testing


PeterAdams-A:

Lukasa (approved):

Issue #2649, opened 2024-02-14
TCPThroughputBenchmark has a crashing race


gmilos:
### Expected behavior
Test runs reliably.

### Actual behavior
Test crashes occasionally with `Unexpectedly found nil while implicitly unwrapping an Optional value`

### Steps to reproduce

1. Build `NIOPerformanceTester` in release mode
2. Run, potentially multiple times, observe the crash[1]

### If possible, minimal yet complete reproducer code (or URL to code)

I can do one better, and explain the crash:

https://github.com/apple/swift-nio/blob/82fd942745b11ccebbf0db3e9e4bf150b60e5e44/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift#L49

is invoked before the channel is active (and sets the implicitly unwrapping optional `context`)

https://github.com/apple/swift-nio/blob/82fd942745b11ccebbf0db3e9e4bf150b60e5e44/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift#L43-L45

### SwiftNIO version/commit hash

https://github.com/apple/swift-nio/tree/82fd942745b11ccebbf0db3e9e4bf150b60e5e44

### System & version information

An internal 23E macOS build, with an internal 15A Xcode / toolchain build.

### Crash details

```
$ lldb ./NIOPerformanceTester tcp_100k_messages_throughput
(lldb) target create "./NIOPerformanceTester"
Current executable set to '/Users/gmilos/Library/Developer/Xcode/DerivedData/swift-nio-fjkfolkpoxtsaocgwcamnzdfgyrw/Build/Products/Release/NIOPerformanceTester' (arm64).
(lldb) settings set -- target.run-args  "tcp_100k_messages_throughput"
(lldb) run
Process 92876 launched: '/Users/gmilos/Library/Developer/Xcode/DerivedData/swift-nio-fjkfolkpoxtsaocgwcamnzdfgyrw/Build/Products/Release/NIOPerformanceTester' (arm64)
skipping 'write_http_headers', limit set = ["tcp_100k_messages_throughput"]
skipping 'http_headers_canonical_form', limit set = ["tcp_100k_messages_throughput"]
skipping 'http_headers_canonical_form_trimming_whitespace', limit set = ["tcp_100k_messages_throughput"]
...
skipping 'udp_10k_vector_reads', limit set = ["tcp_100k_messages_throughput"]
[...]

PR #2724, opened 2024-05-20
Fix race in TCPThroughputBenchmark

ser-0xff:
Fix race in TCPThroughputBenchmark

### Motivation:

Fix an https://github.com/apple/swift-nio/issues/2649

### Modifications:

Start test after server handler got `channelActive` to avoid race.
I was not able to reproduce the crash locally but obviously there is a problem and the PR should fix it.

### Result:

Performance test will not crash.


Lukasa (approved):
This is a lovely fix, thanks!

PR #2725, opened 2024-05-22
testSimpleMPTCP should not fail for ENOPROTOOPT

rnro:
### Motivation:

MPTCP implementations are permitted to return `ENOPROTOOPT` if MPTCP has been disabled. This should not report as a test failure.

Source: https://www.mptcp.dev/implementation.html

### Modifications:

Instead of failing the test, log the permitted error and return.

### Result:

No more misleading testSimpleMPTCP test failures


Lukasa (changes requested):
I think you got the conditional a bit screwy here.

rnro:
> I think you got the conditional a bit screwy here.

Don't code and dash kids. It was very screwy. I have amended it and run all NIO tests locally in a Swift 5.10 Docker container.

Lukasa (approved):

rnro:

PR #2731, opened 2024-05-29
Improve rename error

glbrntt:
Motivation:

The name of the syscall used in errors for all rename calls is "rename". However, this isn't always correct, in some cases "renamex_np" is used, and in others "renameat2" is used.

Modifications:

- Allow the name of the syscall to be provided to the rename error
- Use the correct syscall name

Result:

Better errors

### DIFF
--- a/Sources/NIOFileSystem/FileSystem.swift
+++ b/Sources/NIOFileSystem/FileSystem.swift
@@ -1216,6 +1216,7 @@ extension FileSystem {
 
         case let .failure(errno):
             let error = FileSystemError.rename(
+                "rename",
                 errno: errno,
                 oldName: sourcePath,
                 newName: destinationPath,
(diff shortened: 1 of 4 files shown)


glbrntt:
The breaking changes are okay as they are all SPI:

```
13:34:58 4 breaking changes detected in _NIOFileSystem:
13:34:58   💔 API breakage: func FileSystemError.rename(errno:oldName:newName:location:) has parameter 0 type change from SystemPackage.Errno to Swift.String
13:34:58   💔 API breakage: func FileSystemError.rename(errno:oldName:newName:location:) has parameter 1 type change from SystemPackage.FilePath to SystemPackage.Errno
13:34:58   💔 API breakage: func FileSystemError.rename(errno:oldName:newName:location:) has parameter 3 type change from _NIOFileSystem.FileSystemError.SourceLocation to SystemPackage.FilePath
13:34:58   💔 API breakage: func FileSystemError.rename(errno:oldName:newName:location:) has been renamed to func rename(_:errno:oldName:newName:location:)
```

PR #2733, opened 2024-05-30
Add a fallback path if renameat2 fails

glbrntt:
Motivation:

'renameat2' can fail with EINVAL if the RENAME_NOREPLACE flag isn't supported. However, not all kernel versions support this flag which can result in an error when creating a file 'transactionally'.

Using 'renameat2' only happens on Linux as a fallback when O_TMPFILE isn't available.

Modifications:

- On Linux, fallback to a combination of 'stat' and 'rename' if 'renameat2' fails with EINVAL.
- Add a couple of tests

Result:

Files can still be created exclusively

### DIFF
--- a/Sources/NIOFileSystem/Internal/SystemFileHandle.swift
+++ b/Sources/NIOFileSystem/Internal/SystemFileHandle.swift
@@ -650,7 +650,7 @@ extension SystemFileHandle.SendableView {
     }
 
     @_spi(Testing)
-    public func _close(materialize: Bool) -> Result<Void, FileSystemError> {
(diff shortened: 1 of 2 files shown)


Lukasa (approved):

glbrntt:
API break is a false-positive as it's SPI:
```
16:20:09 1 breaking change detected in _NIOFileSystem:
16:20:09   💔 API breakage: func SystemFileHandle.SendableView._close(materialize:) has been renamed to func _close(materialize:failRenameat2WithEINVAL:)
```

Issue #2712, opened 2024-05-02
NIOFileSystem: Provide API to set last-modified timestamp (ie: "touch")


sliemeobn:
### Expected behavior
I can set the last-modified timestamp on a file via a simple `NIOFileSystem` API.

### Actual behavior
No APIs exist to set timestamps on files or perform a simple "touch".

---
I am not sure how much control over the various file timestamps one would realistically need, but I find that setting the current time as last-modified is a common thing on server systems.

My concrete use case was a tiny service that periodically checks local certificate files (for TLS), renews them if required, and triggers a "config-reload" of a long-running [traefik](https://traefik.io/) instance. The simplest (maybe only?) way to tell traefik to hot-reload its certificates is by touching the config file.

I ended up sys-calling into `utime` myself, but it would be a great fit for the NIOFileSystem APIs.

FranzBusch:
Thanks for opening this up. Going to tag @glbrntt here since he build most of the `NIOFileSystem` stuff.

glbrntt:
Thanks for opening this issue, I think this is a reasonable API to add.

We likely want to add a new API to the `WritableFileHandleProtocol` which can be used to update the access and modification time of an open file. Under the hood this would use `futimens`.

From that base API we can then extend `WritableFileHandleProtocol` to have a 'touch' API which sets both access and modification times to the current time.

FranzBusch:
Implemented in https://github.com/apple/swift-nio/pull/2735

PR #2735, opened 2024-06-05
Add API for setting last accessed and last modified file times

gjcairo:
This PR adds API to set a file's last accessed and last modified times, as well as a `touch()` shorthand that updates both to the current time.

### Motivation:

Feature request: https://github.com/apple/swift-nio/issues/2712

### Modifications:

This PR adds API to set a file's last accessed and last modified times, as well as a `touch()` shorthand that updates both to the current time.

### Result:

Files' last accessed and last modified times can be updated.

### DIFF
--- a/Sources/NIOFileSystem/FileHandleProtocol.swift
+++ b/Sources/NIOFileSystem/FileHandleProtocol.swift
@@ -507,7 +507,30 @@ extension WritableFileHandleProtocol {
     ) async throws -> Int64 {
         try await self.write(contentsOf: buffer.readableBytesView, toAbsoluteOffset: offset)
     }
-    
+}
+
+// MARK: - File times modifiers
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension FileHandleProtocol {
+    /// Sets the file's last access time to the given time.
+    /// 
+    /// - Parameter time: The time to which the file's last access time should be set.
+    ///
+    /// - Throws: If there's an error updating the time. If this happens, the original value won't be modified.
+    public func setLastAccessTime(to time: FileInfo.Timespec) async throws {
+        try await self.setTimes(lastAccess: time, lastDataModification: nil)
+    }
+
+    /// Sets the file's last data modification time to the given time.
+    ///
+    /// - Parameter time: The time to which the file's last data modification time should be set.
+    ///
+    /// - Throws: If there's an error updating the time. If this happens, the original value won't be modified.
+    public func setLastDataModificationTime(to time: FileInfo.Timespec) async throws {
+        try await self.setTimes(lastAccess: nil, lastDataModification: time)
+    }
+
     /// Sets the file's last access and last data modification times to the current time.
     ///
     /// - Throws: If there's an error updating the times. If this happens, the original values won't be modified.
--- a/Sources/NIOFileSystem/FileHandleProtocol.swift
+++ b/Sources/NIOFileSystem/FileHandleProtocol.swift
@@ -15,15 +15,6 @@
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(Linux) || os(Android)
 import NIOCore
 import SystemPackage
-#if canImport(Darwin)
-import Darwin
-#elseif canImport(Glibc)
-import Glibc
-#elseif canImport(Musl)
-import Musl
-#else
-#error("The File Handle Protocol module was unable to identify your C library.")
-#endif
 
 /// A handle for a file system object.
 ///
@@ -152,6 +143,26 @@ public protocol FileHandleProtocol {
     ///
     /// After closing the handle calls to other functions will throw an appropriate error.
     func close() async throws
+
+    /// Sets the file's last access and last data modification times to the given values.
+    ///
+    /// If **either** time is `nil`, the current value will not be changed.
+    /// If **both** times are `nil`, then both times will be set to the current time.
+    ///
+    /// > Important: Times are only considered valid if their nanoseconds components are one of the following:
+    /// > - `UTIME_NOW` (you can use ``FileInfo/Timespec/now`` to get a Timespec set to this value),
+    /// > - `UTIME_OMIT` (you can use ``FileInfo/Timespec/omit`` to get a Timespec set to this value),,
+    /// > - Greater than zero and no larger than 1000 million
+    ///
+    /// - Parameters:
+    ///   - lastAccessTime: The new value of the file's last access time, as time elapsed since the Epoch.
+    ///   - lastDataModificationTime: The new value of the file's last data modification time, as time elapsed since the Epoch.
+    ///
+    /// - Throws: If there's an error updating the times. If this happens, the original values won't be modified.
+    func setTimes(
+        lastAccess: FileInfo.Timespec?,
+        lastDataModification: FileInfo.Timespec?
+    ) async throws
 }
 
 // MARK: - Readable
@@ -472,26 +483,6 @@ public protocol WritableFileHandleProtocol: FileHandleProtocol {
     ///       filesystem with the expected name, otherwise no file will be created or the original
     ///       file won't be modified (if one existed).
     func close(makeChangesVisible: Bool) async throws
-    
-    /// Sets the file's last access and last data modification times to the given values.
-    ///
-    /// If **either** time is `nil`, the current value will not be changed.
-    /// If **both** times are `nil`, then both times will be set to the current time.
-    ///
-    /// > Important: Times are only considered valid if their nanoseconds components are one of the following:
-    /// > - `UTIME_NOW`,
-    /// > - `UTIME_OMIT`,
-    /// > - Greater than zero and no larger than 1000 million
-    ///
-    /// - Parameters:
-    ///   - lastAccessTime: The new value of the file's last access time, as time elapsed since the Epoch.
-    ///   - lastDataModificationTime: The new value of the file's last data modification time, as time elapsed since the Epoch.
-    ///
-    /// - Throws: If there's an error updating the times. If this happens, the original values won't be modified.
-    func setTimes(
-        lastAccessTime: FileInfo.Timespec?,
-        lastDataModificationTime: FileInfo.Timespec?
-    ) async throws
 }
 
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
@@ -520,7 +511,7 @@ extension WritableFileHandleProtocol {
     ///
     /// - Throws: If there's an error updating the times. If this happens, the original values won't be modified.
     public func touch() async throws {
-        try await self.setTimes(lastAccessTime: nil, lastDataModificationTime: nil)
+        try await self.setTimes(lastAccess: nil, lastDataModification: nil)
     }
 }
 
--- a/Sources/NIOFileSystem/FileHandleProtocol.swift
+++ b/Sources/NIOFileSystem/FileHandleProtocol.swift
@@ -151,8 +151,9 @@ public protocol FileHandleProtocol {
     ///
     /// > Important: Times are only considered valid if their nanoseconds components are one of the following:
     /// > - `UTIME_NOW` (you can use ``FileInfo/Timespec/now`` to get a Timespec set to this value),
-    /// > - `UTIME_OMIT` (you can use ``FileInfo/Timespec/omit`` to get a Timespec set to this value),,
-    /// > - Greater than zero and no larger than 1000 million
+    /// > - `UTIME_OMIT` (you can use ``FileInfo/Timespec/omit`` to get a Timespec set to this value),
+    /// > - Greater than zero and no larger than 1000 million: if outside of this range, the value will be clamped to the closest valid value.
+    /// > The seconds component must also be positive: if it's not, zero will be used as the value instead.
     ///
     /// - Parameters:
     ///   - lastAccessTime: The new value of the file's last access time, as time elapsed since the Epoch.
--- a/Sources/NIOFileSystem/FileHandleProtocol.swift
+++ b/Sources/NIOFileSystem/FileHandleProtocol.swift
@@ -15,6 +15,15 @@
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(Linux) || os(Android)
 import NIOCore
 import SystemPackage
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#else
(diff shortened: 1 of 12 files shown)


Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -1046,6 +1046,70 @@ extension SystemFileHandle: WritableFileHandleProtocol {
...
+        lastAccessTime: FileInfo.Timespec?,
+        lastDataModificationTime: FileInfo.Timespec?
+    ) async throws {
+        try await self.threadPool.runIfActive { [sendableView] in
+            try sendableView._withUnsafeDescriptor { descriptor in
+                if lastAccessTime == nil, lastDataModificationTime == nil {
+                    // If the timespec array is nil, as per the `futimens` docs,
+                    // both the last accessed and last modification times
+                    // will be set to now.
+                    futimens(descriptor.rawValue, nil)
```
glbrntt, line 1060: Can you move this to `Syscalls.swift`? The pattern there is to wrap the syscall so we can more easily deal with platform differences and allows us to mock the sys calls for testing.

The function should then be wrapped up in a static function on `Syscall` returning a `Result<Void, Errno>`.

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -1046,6 +1046,70 @@ extension SystemFileHandle: WritableFileHandleProtocol {
...
+                } else {
+                    let lastAccessTimespec: timespec
+                    if let lastAccessTime = lastAccessTime {
+                        lastAccessTimespec = timespec(
```
glbrntt, line 1064: Should we just add an extension to `timespec` to init it from `Timespec`? We do it a handful of times so probably worth doing.

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -1046,6 +1046,70 @@ extension SystemFileHandle: WritableFileHandleProtocol {
...
+                        // Don't modify the last access time.
+                        // Note: tv_sec will be ignored.
+                        lastAccessTimespec = timespec(
+                            tv_sec: 0,
+                            tv_nsec: Int(UTIME_OMIT)
+                        )
+                    }
+
+                    let lastDataModificationTimespec: timespec
+                    if let lastDataModificationTime = lastDataModificationTime {
```
glbrntt, line 1078: FWIW:  rather than all the if-elses we can switch over both values to create a `timespec`:

```swift
let ts: timespec

switch (lastAccessTime, lastModificationTime) {
case (.some(let lastAccess), .some(let lastModification)):
  ts = timspec(...)
case (.some(let lastAccess), .none):
  // ...
case (.none, .some(let lastModification)):
  // ...
case (.none, .none):
  // ...
}
```

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
```diff
@@ -1178,6 +1178,121 @@ final class FileHandleTests: XCTestCase {
             }
         }
     }
+
```
glbrntt, line 1181: Can you also add tests to:
- `FileSystemErrorTests` to check the mapping of `errno`s to error codes?
- `SyscallTests` after you've added the appropriate function to `Syscall`

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -15,6 +15,15 @@
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS) || os(Linux) || os(Android)
 import NIOCore
 import SystemPackage
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#else
```
glbrntt, line 24: Why do we now need these imports?

glbrntt (changes requested):
Great start, I think we need to change a couple of minor things though to fit in with how syscalls are done in the rest of the module.

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -463,6 +472,26 @@ public protocol WritableFileHandleProtocol: FileHandleProtocol {
...
+    /// > - Greater than zero and no larger than 1000 million
+    ///
+    /// - Parameters:
+    ///   - lastAccessTime: The new value of the file's last access time, as time elapsed since the Epoch.
+    ///   - lastDataModificationTime: The new value of the file's last data modification time, as time elapsed since the Epoch.
+    ///
+    /// - Throws: If there's an error updating the times. If this happens, the original values won't be modified.
+    func setTimes(
+        lastAccessTime: FileInfo.Timespec?,
+        lastDataModificationTime: FileInfo.Timespec?
```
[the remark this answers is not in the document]
gjcairo: You're right about the param names, I'll change them. 
WRT to `Times` vs Time` - I don't really have a strong preference, I can make the change. I'll only point out that you can't set one time only unless we default them to nil - you think we should do that?

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -463,6 +472,26 @@ public protocol WritableFileHandleProtocol: FileHandleProtocol {
     ///       filesystem with the expected name, otherwise no file will be created or the original
     ///       file won't be modified (if one existed).
     func close(makeChangesVisible: Bool) async throws
+    
+    /// Sets the file's last access and last data modification times to the given values.
```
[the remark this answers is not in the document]
gjcairo: Yes you're right, this should be on `FileHandleProtocol`

Sources/NIOFileSystem/FileSystemError+Syscall.swift
```diff
@@ -1066,6 +1066,50 @@ extension FileSystemError {
...
+        switch errno {
+        case .permissionDenied, .notPermitted:
+            code = .permissionDenied
+            message = "Not permited to change last access or last data modification times for \(path)."
+
+        case .invalidArgument:
+            code = .invalidArgument
+            message = """
+                Invalid times value: nanoseconds for both timespecs must be UTIME_NOW, UTIME_OMIT, or a number between 0 and 1000 million
+                (last access time: \(String(describing: lastAccessTime)), last data modification time: \(String(describing: lastDataModificationTime))).
```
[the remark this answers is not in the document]
gjcairo: We can - do we still want to handle this error explicitly though even if we never hit it, given it's documented in the syscall's man page?

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -143,6 +143,26 @@ public protocol FileHandleProtocol {
...
+
+    /// Sets the file's last access and last data modification times to the given values.
+    ///
+    /// If **either** time is `nil`, the current value will not be changed.
+    /// If **both** times are `nil`, then both times will be set to the current time.
+    ///
+    /// > Important: Times are only considered valid if their nanoseconds components are one of the following:
+    /// > - `UTIME_NOW` (you can use ``FileInfo/Timespec/now`` to get a Timespec set to this value),
+    /// > - `UTIME_OMIT` (you can use ``FileInfo/Timespec/omit`` to get a Timespec set to this value),,
+    /// > - Greater than zero and no larger than 1000 million
```
glbrntt, line 155: Should we remove this line as we clamp the value?

Sources/NIOFileSystem/FileInfo.swift
```diff
@@ -145,6 +147,29 @@ extension FileInfo {
...
+        private static let UTIME_OMIT_INT = Int(UTIME_OMIT)
+        private static let UTIME_NOW_INT = Int(UTIME_NOW)
+        #elseif canImport(Glibc) || canImport(Musl)
+        private static let UTIME_OMIT_INT = Int(CNIOLinux_UTIME_OMIT)
+        private static let UTIME_NOW_INT = Int(CNIOLinux_UTIME_NOW)
+        #endif
+
+        /// A timespec where the seconds are set to zero and the nanoseconds set to `UTIME_OMIT`.
+        /// In syscalls such as `futimens`, this means the time component set to this value will be ignored.
+        public static var omit = Self.init(
```
glbrntt, line 160: 🚨 `public static var` alert! 🚨

This should either be a `let` or a computed property. Same goes for `now`.

Sources/NIOFileSystem/FileInfo.swift
```diff
@@ -145,6 +147,29 @@ extension FileInfo {
 
     /// A time interval consisting of whole seconds and nanoseconds.
     public struct Timespec: Hashable, Sendable {
+        #if canImport(Darwin)
+        private static let UTIME_OMIT_INT = Int(UTIME_OMIT)
+        private static let UTIME_NOW_INT = Int(UTIME_NOW)
+        #elseif canImport(Glibc) || canImport(Musl)
+        private static let UTIME_OMIT_INT = Int(CNIOLinux_UTIME_OMIT)
+        private static let UTIME_NOW_INT = Int(CNIOLinux_UTIME_NOW)
```
glbrntt, line 155: Can we use swifty casing: `utimeOmit`/`utimeNow`?

Sources/NIOFileSystem/FileInfo.swift
glbrntt: Also let's avoid `Self.init(...)` and just use `Self(...)`

Sources/NIOFileSystem/FileSystemError+Syscall.swift
```diff
@@ -1066,6 +1066,50 @@ extension FileSystemError {
...
+        lastDataModificationTime: FileInfo.Timespec?,
+        location: SourceLocation
+    ) -> FileSystemError {
+        let code: FileSystemError.Code
+        let message: String
+
+        switch errno {
+        case .permissionDenied, .notPermitted:
+            code = .permissionDenied
+            message = "Not permited to change last access or last data modification times for \(path)."
```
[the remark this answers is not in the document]
glbrntt: ^ this one didn't get fixed

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -909,6 +923,84 @@ extension SystemFileHandle.SendableView {
...
+                message: "Couldn't modify file dates, the file '\(self.path)' is closed.",
+                cause: nil,
+                location: .here()
+            )
+        }
+    }
+}
+
+fileprivate extension timespec {
+    init(_ fileinfoTimespec: FileInfo.Timespec) {
```
glbrntt, line 985: nit: don't put access modifiers on `extension`s, it's too easy to accidentally give a method the wrong access (this is something we apply to all of our projects)

```suggestion
extension timespec {
    fileprivate init(_ fileinfoTimespec: FileInfo.Timespec) {
```

Tests/NIOFileSystemTests/FileSystemErrorTests.swift
```diff
@@ -558,6 +558,19 @@ final class FileSystemErrorTests: XCTestCase {
...
+    func testErrnoMapping_futimens() {
+        self.testErrnoToErrorCode(
+            expected: [
+                .permissionDenied: .permissionDenied,
+                .notPermitted: .permissionDenied,
+                .readOnlyFileSystem: .unsupported,
+                .badFileDescriptor: .closed,
+            ]
+        ) { errno in
+                .futimens(errno: errno, path: "", lastAccessTime: nil, lastDataModificationTime: nil, location: .fixed)
```
glbrntt, line 570: nit: intendtation

```suggestion
            .futimens(errno: errno, path: "", lastAccessTime: nil, lastDataModificationTime: nil, location: .fixed)
```

glbrntt (changes requested):
A few small things need fixing otherwise this looks great!

Sources/NIOFileSystem/FileHandleProtocol.swift
gjcairo: I thought it was better to explain that we will clamp/update the value to the closest valid value instead. Otherwise if people are willingly using an invalid value they could be confused as to why the times are being set to the "wrong" thing. Let me know if you don't agree.

glbrntt: Oh yeah I'm happy with that.

Sources/NIOFileSystem/FileInfo.swift
glbrntt: Easily done 🙂

glbrntt (approved):
Looks great, thanks Gus!

glbrntt:
API breakage is expected, we're adding requirements to a protocol.

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -152,6 +143,26 @@ public protocol FileHandleProtocol {
     ///
     /// After closing the handle calls to other functions will throw an appropriate error.
     func close() async throws
+
+    /// Sets the file's last access and last data modification times to the given values.
+    ///
+    /// If **either** time is `nil`, the current value will not be changed.
+    /// If **both** times are `nil`, then both times will be set to the current time.
```
MahdiBM, line 150: IMO ["clamp/update the value to the closest valid value"](https://github.com/apple/swift-nio/pull/2735/files/8222158a110c16d705f767fd3e53caeca910aa56..d3cf9efe57bff7a541f0edf24dcda48dd0d4962a#r1631408774) is good/fine, but the API design still bothers me.

I would have preferred these 2 lines to have been obvious in the API design.
Perhaps using an (structified) enum as the single parameter that the functions accepts. Or anything better. Maybe splitting up the responsibility into 2/3 different functions.

glbrntt: I think it's clear from `setTimes(lastAccess: t)` that `lastDataModification` isn't modified. 

It's fair to say though that setting both to `nil` _isn't_ an obvious API: `setTimes(lastAccess: nil, lastDataModification: nil)` but that's why we also have the `touch()` wrapper which many developers _are_ familiar with.

glbrntt:

Sources/NIOFileSystem/FileHandleProtocol.swift
gjcairo: If we had separate methods for each, we'd unavoidably have to make two separate sys calls, which is expensive. We'd also not be setting the times to the exact same value, as the current time will have changed if we use the `UTIME_NOW` `timespec`.

However, I've added two new convenience methods that default one of the parameters to `nil` to have nicer API: `setLastAccessTime(to:)` and `setLastDataModificationTime(to:)`.

glbrntt (approved):

PR #2739, opened 2024-06-11
Update availability guard on tests

gjcairo:
Some tests are currently broken because of unavailable APIs

### Motivation:

We want tests to build.

### Modifications:

Update availability guard

### Result:

Tests build on all platforms

### DIFF
--- a/Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
+++ b/Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
@@ -19,7 +19,7 @@ import NIOPosix
 import NIOFoundationCompat
 import XCTest
 
-@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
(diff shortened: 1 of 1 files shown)


Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
```diff
@@ -19,7 +19,7 @@ import NIOPosix
 import NIOFoundationCompat
 import XCTest
 
-@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+@available(macOS 10.15, iOS 15.0, watchOS 6.0, tvOS 13.0, *)
```
simonjbeaumont, line 22: Is this the right bump, is it just iOS affected?

gjcairo: You're right, I was missing bumping the other platforms.

glbrntt (approved):

glbrntt:

Issue #2734, opened 2024-05-30
MessageToByteHandler never shrinks its buffer (it should!)


weissi:
https://github.com/apple/swift-nio/blob/e2aef20375a9626b4ce3a32786a6626678e9f19c/Sources/NIOCore/Codec.swift#L748

NIO's `MessageToByteHandler` tries to reuse its existing `self.buffer` to save allocations. It doesn't however ever shrink that buffer. So if the user ever sends something super large, the buffer will forever be super large. I reckon even potential CoW copies will just reallocate a huge buffer.

Anyway, M2BH should shrink the buffer just like B2MD does if it's too large.

(This is not a security issue though because M2BH is in the _outbound_ pipeline, so the data is user controlled and not network controlled.)

ali-ahsan-ali:
Hi, trying to take a look at this one 👍 . "M2BH should shrink the buffer just like B2MD does." This suggests that there is a solution in mind. 

I have taken some time to look at the code, wondering if said solution is to copy the implementation for discardReadBytes but instead shrink the buffer size to the writeIndex after the context.write operation has been completed?

Any pointers would be appreciated. Thanks for your time 😄 .

ali-ahsan-ali:
I think this can be closed

Issue #2729, opened 2024-05-28
ByteBuffer: Investigate optimal allocation sizes (probably just under a power of two)


weissi:
Currently, ByteBuffer always allocates in powers of 2 which makes sense. But many allocators have a small amount of overhead which means the actual space reservation might be even bigger, possibly another power of two, often unnecessarily.

For example the following could happen:
- The user needs to write 9,000 bytes and asks for `allocator.buffer(capacity: 9_000)`
- ByteBuffer gets a `malloc(16_384)` (power of 2)
- libc's allocator has (for the example's sake) 16 bytes of overhead storing some metadata
- This means we now require 16384 (for the malloc) + 16 bytes (allocator overhead) = 16394 which isn't a power of two

This could yield to an extra page being mapped for nothing (we only needed 9000 bytes after all). NIO's ByteBuffer should probably allocate just under the next power of two to leave some slack for the allocator in case it needs it. Needs investigation.

This becomes even more important if/when ByteBuffer is able to use tail allocations.

hassila:
See some note to consider in the related case:

https://github.com/apple/swift-nio/issues/2723#issuecomment-2135885509

weissi:
Similar to #2770

hassila:
Maybe we could be allowed to optionally provide a malloc size  strategy that could default to the current power-of-two behavior but let one provide the result from the Darwin function (or some other heuristic of we know we want an exact size as in the other case I linked to, or if we know optimal sizes for the current allocator in use eg. Jemalloc)?

This would give us the ability to influence overhead. The size returned by such a closure would need to be bigger or equal to the requested size of course.

Lukasa:
I think this wraps into a broader question around custom allocator patterns. There's been some interest in us actually making `ByteBufferAllocator` into a more meaningful type, and some of this will intersect with that.

PR #2745, opened 2024-06-19
Issue-2734 - Clamp buffer to maximum upon large write operation

ali-ahsan-ali:
Fix Issue 2734 by clamping storage size after large write operation
### Motivation:

Fix Issue 2734
### Modifications:
- Added a function to clamp storage by copying bytes and setting new capacity of storage
- Adding a function to clamp the capacity of ByteBuffer
- Added the ability to specify he maxBufferCapacity to MessageToByteHandler
### Result:

Once a write message that is larger than the capacity of the MessageToByteHandler's maxBufferCapacity, it clamps the byteBuffer down.


Sources/NIOCore/Codec.swift
```diff
@@ -768,9 +768,11 @@ public final class MessageToByteHandler<Encoder: MessageToByteEncoder>: ChannelO
     private var state: State = .notInChannelYet
     private let encoder: Encoder
     private var buffer: ByteBuffer? = nil
+    private let maxBufferCapacity: Int
```
ali-ahsan-ali, line 771: I toyed around the idea of having the ability to switch between different strategies, for example a median of the last N number of write operations and using that as the size to clamp around. I think if there is consistently large requests, this clamping would be more pain than it would be useful. But I don't fully understand the context of the repo so I let it be. LMK your thoughts.

ali-ahsan-ali:
Addressing https://github.com/apple/swift-nio/issues/2734

Sources/NIOCore/Codec.swift
FranzBusch: I agree that constantly clamping the buffer for requests that are even slightly over the size would have a negative performance impact here. I think we might want to become a bit smarter here to avoid flip flopping the size of the buffer.

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -825,6 +831,24 @@ public struct ByteBuffer {
         }
         return true
     }
+    
+    /// Reset capacity of the buffer to the specified amount.
+    ///
+    /// - returns: `true` if one or more bytes have been discarded, `false` if there are no bytes to discard.
+    @inlinable
+    @discardableResult public mutating func clampBufferCapacity(to maxRetainedCapacity: Int) -> Bool {
```
Lukasa, line 839: This method is pretty scary to me, as it behaves very strangely if there are more readable bytes in the buffer than the requested capacity. This behaviour is also out-of-line with the comment: the comment says it returns true if bytes have been discarded, but it returns true even if those bytes are not in the readable slice.

I think it might be better to have this refuse to shrink if the target capacity is _smaller_ than the number of readable bytes.

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1837,6 +1837,43 @@ class ByteBufferTest: XCTestCase {
...
+        XCTAssertTrue(buffer.clampBufferCapacity(to: maxBufferCapacityToBeRetained))
+        XCTAssertEqual(buffer.capacity, maxBufferCapacityToBeRetained)
+        print(buffer._storage.dumpBytes(slice: buffer._slice, offset: 0, length: buffer.capacity))
+
+        // For a small item it should not clamp buffer capacity and keep original buffer capacity before write
+        buffer.clear()
+        buffer.writeString("Small item")
+        XCTAssertFalse(buffer.clampBufferCapacity(to: maxBufferCapacityToBeRetained))
+        XCTAssertEqual(buffer.capacity, maxBufferCapacityToBeRetained) // Keeps original capacity of the buffer
+        print(buffer._storage.dumpBytes(slice: buffer._slice, offset: 0, length: buffer.capacity))
```
Lukasa, line 1874: Can we remove the prints?

Sources/NIOCore/Codec.swift
Lukasa: The easiest strategy is probably to add a power-of-two slack.

Sources/NIOCore/Codec.swift
```diff
@@ -768,9 +768,11 @@ public final class MessageToByteHandler<Encoder: MessageToByteEncoder>: ChannelO
...
 
-    public init(_ encoder: Encoder) {
+    public init(_ encoder: Encoder, maxBufferCapacity: Int = 2048) {
```
Lukasa, line 773: We can't add a new parameter to this initializer, it's an API break. We need to add a new initializer.

Lukasa (changes requested):
Thanks for this! I've left a few notes in the diff.

Sources/NIOCore/Codec.swift
ali-ahsan-ali: Is this the suggestion? I don't see how this is a better alternative to the above. Or are we making a breaking change and forcing consumers to specify this field as part of their implementation?

Screenshot 2024-06-22 at 2 25 41 pm

ali-ahsan-ali: The non-breaking method is a new initialiser with a default initialiser, but at that point, why not keep he original implementation. Is there some principle of API creation I am violating?

Tests/NIOCoreTests/ByteBufferTest.swift
ali-ahsan-ali: 😅

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: "Refuse to shrink if the target capacity is smaller than the number of readable byes"

Wouldn't the target capacity ideally be smaller than the number of readable bytes in the situation that we want to shrink the size of the buffer? 

I think I need some extra help understanding this comment or what the resulting guard statement should look like.

FranzBusch: The problem is if you would shrink below the number of readable bytes you would basically discard data which is lost. This isn't great and can lead to pretty exciting bugs.
Going back to the original issue. The problem is that the `MessageToByteHandler` might have to write a single large message into the buffer. This buffer is then sent to the pipeline to be written out. This will result in the written bytes of the buffer to be "read" by the pipeline and the `readableBytes` will become `0`. The problem we are trying to solve here is that the buffer is then re-used by the `MessageToByteHandler` for only small messages going forward but it will forever reserve the capacity needed for the single large message.

The flow is basically like this
- Encode large message in `MessageToByteHandler` to `ByteBuffer` `b1`
- Write `b1` in channel pipeline -> `b1.readableBytes` becomes `0`
- Encode small message in `MessageToByteHandler` to `b1` (`b1` is still sized for the previous large message)

What we need here is a clamp that does `max(self.readableBytes, desiredCapacity)`

ali-ahsan-ali: Thank you. This helps me quite a bit. Will continue and push a PR once my second question below is answered about the API initialiser design.

Sources/NIOCore/Codec.swift
ali-ahsan-ali: I have pushed something, take a look at LMK

Lukasa: The new API shape is right.

Just to clarify, it is always an API break to add a new parameter to a method, even if that parameter has a default value. The _reason_ it's a break is that you can refer to a method by its "fully qualified name": `MessageToByteEncoder.init(_:)`. If you add a new parameter with a default value that method no longer exists. Having two methods as you've written here is the right spelling.

Sources/NIOCore/ByteBuffer-core.swift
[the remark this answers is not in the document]
ali-ahsan-ali: I am pretty sure in my implementation it is not clamping to the amount of readable bytes, only to the desired capacity if it is larger than or equal to the number of readable bytes. 

"I agree that constantly clamping the buffer for requests that are even slightly over the size would have a negative performance impact here" - I took this to mean that constantly clamping the requests to a size other than the one specified by the consumer would be a negative performance impact. Is that the case?

Let me know and I will push a commit tonight (your morning) with the edited comment and the new logic for a "max" function of the two values.

Sources/NIOCore/Codec.swift
ali-ahsan-ali: Thanks. Had to sleep on that one to understand it.

Sources/NIOCore/ByteBuffer-core.swift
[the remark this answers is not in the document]
ali-ahsan-ali: Pushed change with above comment and implementation following your comment.

Sources/NIOCore/Codec.swift
```diff
@@ -768,9 +768,16 @@ public final class MessageToByteHandler<Encoder: MessageToByteEncoder>: ChannelO
     private var state: State = .notInChannelYet
     private let encoder: Encoder
     private var buffer: ByteBuffer? = nil
+    private let maxBufferCapacity: Int?
 
+    public init(_ encoder: Encoder, maxBufferCapacity: Int) {
```
FranzBusch, line 773: This is no longer a `maxBufferCapacity` but rather the desired capacity to keep the buffer at.
```suggestion
    public init(_ encoder: Encoder, desiredBufferCapacity: Int) {
```

Sources/NIOCore/Codec.swift
```diff
@@ -811,6 +818,9 @@ extension MessageToByteHandler {
             self.buffer!.clear()
             try self.encoder.encode(data: data, out: &self.buffer!)
             context.write(self.wrapOutboundOut(self.buffer!), promise: promise)
+            if let maxBufferCapacity {
```
FranzBusch, line 821: I am still a bit worried of doing this on every single write. Instead of adding some heuristic to only do It after a period of small writes.

Lukasa: Easiest heuristic is to do it on `flush`.

Sources/NIOCore/Codec.swift
ali-ahsan-ali: Committed this change.

Sources/NIOCore/Codec.swift
ali-ahsan-ali: I have pushed a change, can you please take a look?

ali-ahsan-ali: Sorry, @FranzBusch  and @Lukasa wondering if I have hit the mark by clamping the buffer by using 

```
public func flush(context: ChannelHandlerContext) {
        context.flush()
        if let desiredBufferCapacity {
            self.buffer?.clampBufferCapacity(to: desiredBufferCapacity)
        }
    }
````

Let me know what you think when you find some time.

File: Sources/NIOCore/ByteBuffer-core.swift [hunk 7]
```diff
@@ -825,6 +831,25 @@ public struct ByteBuffer {
         }
         return true
     }
+    
+    /// Clamps the buffers capacity to either the maximum of the `desiredCapacity` or the buffer's `readableBytes`.
+    ///
+    /// - returns: A boolean indicating if the capacity of the buffer has changed.
+    @inlinable
+    @discardableResult public mutating func clampBufferCapacity(to desiredCapacity: Int) -> Bool {
```
FranzBusch, line 839: We should document he parameter here

FranzBusch (commented):
Except the doc nit this look good to me now. I would like @Lukasa to also take a look before merging.

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -324,6 +324,12 @@ public struct ByteBuffer {
             self.bytes = ptr
             self.capacity = newCapacity
         }
+        
+        @inlinable
+        func clampStorage(to capacity: Int) {
+            self.bytes.copyMemory(from: self.bytes, byteCount: capacity)
+            self.capacity = _toCapacity(capacity)
```
Lukasa, line 331: This doesn't effectively do what we want: it shrinks what we _report_ the capacity to be, but not what it _is_. Clamping the storage will almost always require a new allocation.

As we currently use a power-of-two capacity policy, we should probably only actually do the reallocation when the `desiredCapacity` is smaller than the current capacity _after_ applying the next power of two math. At that point, we should likely just call `reallocSlice` on the storage.

ali-ahsan-ali: I have pushed a commit. I think it addresses what you have said here. Please take a look. I removed the _copyStorageAndRebase optimisation since it seems to be doing exactly the same thing, which was reallocSlice.

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -825,6 +825,23 @@ public struct ByteBuffer {
         }
         return true
     }
+    
+    /// Shrinks the buffers capacity to a `desiredCapacity` without altering readably bytes.
```
Lukasa, line 829: Let's elaborate on this by explaining the actual algorithm, just to save people from looking at the code. Something like:

```
/// The `ByteBuffer` will successfully be shrunk if the requested capacity is less than the current capacity, and the requested capacity is more than the number of readable bytes in the buffer. If either condition is not true, the buffer will not be shrunk.
```

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -825,6 +825,23 @@ public struct ByteBuffer {
...
+    ///
+    /// - Parameter desiredCapacity: The desired capacity for the buffers capacity to be shrunken to
+    /// - Returns: Bool indicating whether the buffer capacity has been shrunk to the desiredCapacity.
+    @inlinable
+    @discardableResult public mutating func shrinkBufferCapacity(to desiredCapacity: Int) -> Bool {
+        guard desiredCapacity < capacity, desiredCapacity > readableBytes else {
```
Lukasa, line 835: We should compare to the effective requested capacity, which will be `capacity == 0 ? 0 : capacity.nextPowerOf2ClampedToMax()`.

Now is also probably a good time to factor that computation out to a `static func` on `ByteBuffer`, as we're probably going to want to tweak it (see #2729).

ali-ahsan-ali: Sorry, I don't understand this. 

If desired capacity is 1023 and the current capacity is also 512, would comparing with capacity.nextPowerOf2 not mean that the condition passes and this operation actually reallocates the buffers capacity by expanding it? 
This is assuming the below implementation method which is what I got from your comment. 
`guard desiredCapacity < capacity. nextPowerOf2(), desiredCapacity > readableBytes`

Also, since Capacity it an integer, I think it can only call nextPowerOf2 and not nextPowerOf2ClampedToMax. Meaning we do not need the `capacity == 0 ? 0` as it is already done in `nextPowerOf2` by a 
`
guard self != 0 else {
            return 1
        }
`

Lukasa: The check is to move `desiredCapacity` to `nextPowerOf2`, not `currentCapacity`. In that case, the desired capacity would be 1024, which is larger than current capacity at 512.

ali-ahsan-ali: Hey @Lukasa. I think I have addressed this in the latest commit. Please take a look when you find some time.

ali-ahsan-ali: Bump

Lukasa (approved):
Great, I'm happy with this if @weissi is.

Sources/NIOCore/Codec.swift
```diff
@@ -844,4 +851,11 @@ extension MessageToByteHandler {
             context.fireErrorCaught(error)
         }
     }
+    
+    public func flush(context: ChannelHandlerContext) {
+        context.flush()
+        if let desiredBufferCapacity {
```
weissi, line 857:
```suggestion
        if let self.desiredBufferCapacity {
```

Sources/NIOCore/Codec.swift
```diff
@@ -844,4 +851,11 @@ extension MessageToByteHandler {
             context.fireErrorCaught(error)
         }
     }
+    
+    public func flush(context: ChannelHandlerContext) {
+        context.flush()
```
weissi, line 856: This should happen _after_ reconciling the state. `context.flush()` calls other code which may re-enter this method

weissi: as in: just switch the lines or do `defer { context.flush() }`

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -884,6 +884,34 @@ public struct ByteBuffer {
...
+    /// The `ByteBuffer` will successfully be shrunk if the requested capacity is less than the current capacity, 
+    /// and the requested capacity is more than the number of readable bytes in the buffer.
+    /// If either condition is not true, the buffer will not be shrunk.
+    ///
+    /// - Parameter desiredCapacity: The desired capacity for the buffers capacity to be shrunken to
+    /// - Returns: Bool indicating whether the buffer capacity has been shrunk to the desiredCapacity.
+    @inlinable
+    @discardableResult public mutating func shrinkBufferCapacity(to desiredCapacity: Int) -> Bool {
+        let desiredCapacity = ByteBuffer.addPaddingTo(desiredCapacity)
+        guard desiredCapacity < capacity, desiredCapacity > readableBytes else {
```
weissi, line 897:
```suggestion
        guard desiredCapacity < self.capacity, desiredCapacity > self.readableBytes else {
```

missing `self.`s make this very difficult to review for correctness. Mind adding those?

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -884,6 +884,34 @@ public struct ByteBuffer {
...
+            return false
+        }
+        
+        let shrunkenCapacity = _toCapacity(desiredCapacity)
+        self._storage = self._storage.reallocSlice(self._slice.lowerBound..<self._slice.lowerBound + shrunkenCapacity, capacity: shrunkenCapacity)
```
weissi, line 902: I think this reads potentially unwritten bytes which is UB.

Let's say we have an old ByteBuffer with `capacity: 8` in this state

```
slice.lowerBound (0)
| readerIndex (2)
| | writerIndex (4)
| | |   slice.upperBound (8, one past the end of the slice)
| | |   |
v v v   v
uuWWuuuu
```

If we now call `shrinkBufferCapacity(to: 4)` which will work, it's more than `self.readableBytes` but less than `self.capacity`, then we'll allocate a new storage of size 4 and copy `uuWW` in using `memcpy` in `reallocSlice`, that's UB.

where:

- `W`: is a byte written to (and readable)
- `u`: is a byte never written

We could create such a buffer as

```
var buf = ByteBuffer(capacity: 8)
buf.setWriterIndex(to: 2)
buf.setReaderIndex(to: 2)
buf.writeString("WW")
buf.shrinkBufferCapacity(to: 4)
XCTAssertEqual("WW", String(buffer: buf))
```

if this is run through Valgrind, I would expect valgrind to complain that we read bytes that were never written.

weissi: @ali-ahsan-ali regardless of the potential UB, would you mind adding the above test case? The existing tests don't test anything with `readerIndex != 0`

Also worth testing:

```
var buf = ByteBuffer(capacity: 16384)
buf.setWriterIndex(to: 16000)
buf.setReaderIndex(to: 16000)
buf.writeString("WW")
buf.shrinkBufferCapacity(to: 4)
XCTAssertEqual("WW", String(buffer: buf))
```

weissi: Why aren't we using `_copyStorageAndRebase(capacity: _Capacity, resetIndices: Bool = false)` actually?

weissi (changes requested):
Hey, thanks for the patch. I _think_ I found a UB issue and I'm not sure if the reader/writerIndices are handled correctly either

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: Noted.

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: Added tests and using _copyStorageAndRebase with resettingIndices == true to avoid copying unwritten bytes. LMK if that addresses this comment

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1891,7 +1891,24 @@ class ByteBufferTest: XCTestCase {
...
+        buffer.shrinkBufferCapacity(to: 4)
+        XCTAssertEqual("WW", String(buffer: buffer))
+
+        // If readable bytes is exactly the same as buffer capacity shrunken to
+        buffer = self.allocator.buffer(capacity: 16)
+        buffer.moveWriterIndex(to: 8)
+        buffer.moveReaderIndex(to: 8)
+        buffer.writeString("WWWWWWWW")  // 8 bytes written
+        buffer.shrinkBufferCapacity(to: 4)  // Invisible padding makes this 8 bytes
+        XCTAssertEqual("WWWWWWWW", String(buffer: buffer))  // All 8 bytes are returned!
```
ali-ahsan-ali, line 1909: Can we very specifically look at the 2nd test in `testShrinkBufferCapacityWithLeadingUnwrittenBytes` and the comments and make sure it is what is expected. Or if I am testing implementation detail that should not fail if the function changes?

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: @weissi Please take a look when you find the time.

ali-ahsan-ali:
Hey @weissi LMK when you find time to take a look at this. No rush!

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -885,6 +886,30 @@ public struct ByteBuffer {
...
+    /// The `ByteBuffer` will successfully be shrunk if the requested capacity is less than the current capacity,
+    /// and the requested capacity is more than the number of readable bytes in the buffer.
+    /// If either condition is not true, the buffer will not be shrunk.
+    ///
+    /// - Parameter desiredCapacity: The desired capacity for the buffers capacity to be shrunken to
+    /// - Returns: Bool indicating whether the buffer capacity has been shrunk to the desiredCapacity.
+    @inlinable
+    @discardableResult public mutating func shrinkBufferCapacity(to desiredCapacity: Int) -> Bool {
+        let desiredCapacity = ByteBuffer.addPaddingTo(desiredCapacity)
+        guard desiredCapacity < self.capacity, desiredCapacity > self.readableBytes else {
```
weissi, line 898:
```suggestion
        guard desiredCapacity < self.capacity, desiredCapacity >= self.readableBytes else {
```

shouldn't this be a `>=` ?

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -885,6 +886,30 @@ public struct ByteBuffer {
...
+        self._copyStorageAndRebase(capacity: _toCapacity(desiredCapacity), resetIndices: true)
+        return true
+    }
+
+    /// Returns size of capacity with optimal padding
+    /// - Parameter initialCapacity: Capacity that needs expansion with padding
+    /// - Returns: Capacity with calculated padding
+    @inlinable
+    static func addPaddingTo(_ initialCapacity: Int) -> Int {
+        initialCapacity == 0 ? 0 : initialCapacity.nextPowerOf2()
```
weissi, line 910:
```suggestion
        initialCapacity <= 0 ? 0 : initialCapacity.nextPowerOf2()
```

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -885,6 +886,30 @@ public struct ByteBuffer {
...
+        }
+        self._copyStorageAndRebase(capacity: _toCapacity(desiredCapacity), resetIndices: true)
+        return true
+    }
+
+    /// Returns size of capacity with optimal padding
+    /// - Parameter initialCapacity: Capacity that needs expansion with padding
+    /// - Returns: Capacity with calculated padding
+    @inlinable
+    static func addPaddingTo(_ initialCapacity: Int) -> Int {
```
weissi, line 909:
```suggestion
    static func roundUpToUsableCapacity(_ initialCapacity: Int) -> Int {
```

I think we should be more specific here. You can also use this function elsewhere.

weissi:
> Hey @weissi LMK when you find time to take a look at this. No rush!

Commented, looks 99% complete, just three small things

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: I have commited this change.

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: I have not commited this change but changed the function signature to be UInt32. This would make it usable across the repo and not just my use case of `Int`. LMK What you think.

Sources/NIOCore/ByteBuffer-core.swift
ali-ahsan-ali: Commited this change. And I can raise a follow up to use this across the repo.

weissi (approved):
lgtm, thank you!

ali-ahsan-ali:
Will cry tears of joy once this gets merged 👍

Lukasa (approved):
Very nice, thank you very much for your hard work and persistence with this!

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -887,6 +888,30 @@ public struct ByteBuffer {
         return true
     }
 
+    /// The `ByteBuffer` will successfully be shrunk if the requested capacity is less than the current capacity,
+    /// and the requested capacity is more than the number of readable bytes in the buffer.
```
ali-ahsan-ali, line 892: greater than or equal to 

Needs a change

PR #2753, opened 2024-06-26
NIOSendableBox: allow off-loop initialisation iff Value is Sendable

weissi:
### Motivation:

`NIOLoopBound` and `NIOLoopBoundBox` are important tools for making mutable types `Sendable`.
The power of `NIOLoopBoundBox` is that it allows correct mutation, whilst remaining `Sendable` without locks and without having to use `@unchecked` if the mutable class makes sure to have all accesses (reading & writing) run on one particular `EventLoop`. This is safe as `EventLoop`s guarantee sequentially consistent memory order across executions of different work items/events. Typically `EventLoop`s achieve this by just being thread-bound.

These types are well used already but there's a small and common pattern which is safe but unsupported as of yet:

Initialise a `NIOLoopBoundBox` with a `Sendable` value _off_ the loop. All further accesses (reads & writes) however happen _on_ the loop. That's safe because of Swift's Definitive Initialisation (DI) but `NIOLoopBoundBox` didn't support this pattern (apart from `makeEmptyBox` which always initialises with `nil`). 

### Modifications:

- Allow `Sendable` values to be provided whilst creating the type off loop.

### Result:

- Even more types can be safely & correctly made `Sendable` without using `@unchecked.

### DIFF
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -107,6 +107,26 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         return .init(_value: nil, uncheckedEventLoop: eventLoop)
     }
 
+    /// Initialise a ``NIOLoopBoundBox`` by sending a ``Sendable`` value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because we know that `value` is ``Sendable``.
+    /// So we don't need to protect `value` itself, we just need to protect the ``NIOLoopBoundBox`` against mutations which we do because the ``value``
+    /// accessors are checking that we're on `eventLoop`.
+    public static func makeBoxSendingValue(
+        _ value: Value,
+        as: Value.Type = Value.self,
+        eventLoop: EventLoop
+    ) -> NIOLoopBoundBox<Value> where Value: Sendable {
+        // Here, we -- possibly surprisingly -- do not precondition being on the EventLoop. This is okay for a few
+        // reasons:
+        // - This function only works with `Sendable` values, so we don't need to worry about somebody
+        //   still holding a reference to this.
+        // - Because of Swift's Definitive Initialisation (DI), we know that we did write `self._value` before `init`
+        //   returns.
+        // - The only way to ever write (or read indeed) `self._value` is by proving to be inside the `EventLoop`.
+        return .init(_value: value, uncheckedEventLoop: eventLoop)
+    }
+
     /// Access the `value` with the precondition that the code is running on `eventLoop`.
     ///
     /// - note: ``NIOLoopBoundBox`` itself is reference-typed, so any writes will affect anybody sharing this reference.
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -107,9 +107,9 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         return .init(_value: nil, uncheckedEventLoop: eventLoop)
     }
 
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -107,6 +107,26 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         return .init(_value: nil, uncheckedEventLoop: eventLoop)
     }
 
+    /// Initialise a ``NIOLoopBoundBox`` with a ``Sendable`` value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because we know that `value` is ``Sendable``.
+    /// So we don't need to pretect `value` itself, we just need to protect the ``NIOLoopBoundBox`` against mutations which we do because the ``value``
```
Lukasa, line 113:
```suggestion
    /// So we don't need to protect `value` itself, we just need to protect the ``NIOLoopBoundBox`` against mutations which we do because the ``value``
```

weissi: thanks!

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -107,6 +107,26 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         return .init(_value: nil, uncheckedEventLoop: eventLoop)
     }
 
+    /// Initialise a ``NIOLoopBoundBox`` by sending a ``Sendable`` value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because we know that `value` is ``Sendable``.
+    /// So we don't need to protect `value` itself, we just need to protect the ``NIOLoopBoundBox`` against mutations which we do because the ``value``
+    /// accessors are checking that we're on `eventLoop`.
+    public static func makeBoxSendingValue(
```
FranzBusch, line 115: I find it weird to use a factory method here personally. I would prefer using an init here. I also see little value in having this constrained to `Sendable` since `Sendable` types can freely be shared between threads.

I expected a spelling like this

```
init(value: sending Value, eventLoop: any EventLoop) {}
```

This would allow moving an non-Sendable value into an EL's isolation region which is dynamically enforced by the LoopBoundBox.

weissi: - factory is to get "more words", the `init`s are unconstrained but have to run on `eventLoop`
- `sending` isn't ready yet (#2754)
- there's a huge value for `Sendable` values. It's not about sending the _values_ around, `NIOLoopBoundBox` is about _mutating_ the values. Typically this is for thread-safe network clients/servers that have a state machine which only gets accessed on a particular EL. So current: `var state: State // only access on self.eventLoop`; new: `let state: NIOLoopBound<State>`

weissi: That's not accurate.

- `class Box<T> { var value: T }` cannot be Sendable (without synchronisation), you get _mutation_ and _sharing_.
- `class NIOLoopBoundBox<T> { var value: T}` is always `Sendable` because it checks that all accesses are on the same EL. So you get _mutation_ but no _sharing_

Lukasa: I've got no objection to the PR in its current state, but I'm not going to approve because I'd like to see @FranzBusch and @weissi get aligned here.

FranzBusch (approved):

weissi:
```
19:46:31 Test Case 'NIOConcurrencyHelpersTests.testLoadAndStoreHammeBuild timed out (after 20 minutes). Marking the build as failed.
```

weissi:
@swift-nio-bot test failed

weissi:

PR #2756, opened 2024-06-27
Close channel during upgrade, if client closes input

### LINKED ISSUE #2742: NIOTypedWebSocketServerUpgrader hangs when connection is opened and closed without sending any data
### Expected behavior
Given a server setup with WebSocket upgrade using `NIOTypedWebSocketServerUpgrader` and `allowRemoteHalfClosure` set to true on the childChannel. If I open a connection to this server and close it immediately without sending any data the negotiation future returned from `configureUpgradableHTTPServerPipeline` on the server should throw an connection closed error.

### Actual behavior

The negotiation future never completes leaving the child channel open.

If `allowRemoteHalfClosure` is set to false on the server then the negotiation future throws a `inappropriateOperationForState` error which isn't ideal but at least I can catch that the client has closed the connection 

### Steps to reproduce

1. Using the NIOWebSocketServer.swift example
2. Add `.childChannelOption(ChannelOptions.allowRemoteHalfClosure, value: true)` before `bind` call in `run`
3.  Use a client to connect and close connection immediately
```swift
try await ClientBootstrap(group: MultiThreadedEventLoopGroup.singleton)
    .connect(host: "localhost", port: 8888)
    .flatMap { $0.close() }
    .get()
```
4. Note that negotiation future never completes

e5a216ba89deba84356bad9d4c2eab99071c745b

swift-driver version: 1.90.11.1 Apple Swift version 5.10 (swiftlang-5.10.0.13 clang-1500.3.9.4)
Target: x86_64-apple-macosx14.0

adam-fowler:
Check for half closure during server upgrade and close channel if client closes the channel

### Motivation:

This is to fix #2742 

### Modifications:

Add `userInboundEventTriggered` function to `NIOTypedHTTPServerProtocolUpgrader` which checks for `ChannelEvent.inputClosed`

### Result:

Negotiation future now errors when client closes the connection instead of never completing

### DIFF
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -154,6 +154,17 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
         }
     }
 
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        switch event {
+        case let evt as ChannelEvent where evt == ChannelEvent.inputClosed:
+            // The remote peer half-closed the channel during the upgrade so we should close
+            context.close(promise: nil)
+
+        default:
+            context.fireUserInboundEventTriggered(event)
+        }
+    }
+
     private func channelRead(context: ChannelHandlerContext, requestPart: HTTPServerRequestPart) {
         switch self.stateMachine.channelReadRequestPart(requestPart) {
         case .failUpgradePromise(let error):
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -160,8 +160,8 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
             // The remote peer half-closed the channel during the upgrade. Should we close the other side
             switch self.stateMachine.inputClosed() {
             case .close:
-                self.upgradeResultPromise.fail(ChannelError.inputClosed)
                 context.close(promise: nil)
+                self.upgradeResultPromise.fail(ChannelError.inputClosed)
             case .continue:
                 break
             case .fireInputClosedEvent:
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -157,8 +157,14 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
     public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
         switch event {
         case let evt as ChannelEvent where evt == ChannelEvent.inputClosed:
-            // The remote peer half-closed the channel during the upgrade so we should close
-            context.close(promise: nil)
+            // The remote peer half-closed the channel during the upgrade. Should we close the other side
+            switch self.stateMachine.closeInbound() {
+            case .close:
+                self.upgradeResultPromise.fail(ChannelError.ioOnClosedChannel)
+                context.close(promise: nil)
+            case .continue:
+                break
+            }
 
         default:
             context.fireUserInboundEventTriggered(event)
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -158,12 +158,14 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
         switch event {
         case let evt as ChannelEvent where evt == ChannelEvent.inputClosed:
             // The remote peer half-closed the channel during the upgrade. Should we close the other side
-            switch self.stateMachine.closeInbound() {
+            switch self.stateMachine.inputClosed() {
             case .close:
                 self.upgradeResultPromise.fail(ChannelError.inputClosed)
                 context.close(promise: nil)
             case .continue:
                 break
+            case .fireInputClosedEvent:
+                context.fireUserInboundEventTriggered(event)
             }
 
         default:
@@ -418,6 +420,7 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
             switch self.stateMachine.unbuffer() {
             case .close:
                 context.close(promise: nil)
+
             case .fireChannelRead(let data):
                 context.fireChannelRead(data)
 
@@ -425,6 +428,9 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
                 context.fireChannelReadComplete()
                 context.pipeline.removeHandler(self, promise: nil)
                 return
+
+            case .fireInputClosedEvent:
+                context.fireUserInboundEventTriggered(ChannelEvent.inputClosed)
             }
         }
     }
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -416,6 +416,8 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
     private func unbuffer(context: ChannelHandlerContext) {
         while true {
             switch self.stateMachine.unbuffer() {
+            case .close:
+                context.close(promise: nil)
             case .fireChannelRead(let data):
                 context.fireChannelRead(data)
 
--- a/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
@@ -160,7 +160,7 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
             // The remote peer half-closed the channel during the upgrade. Should we close the other side
             switch self.stateMachine.closeInbound() {
             case .close:
-                self.upgradeResultPromise.fail(ChannelError.ioOnClosedChannel)
+                self.upgradeResultPromise.fail(ChannelError.inputClosed)
                 context.close(promise: nil)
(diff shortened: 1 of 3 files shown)


Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
```diff
@@ -152,6 +152,17 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
         }
     }
 
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        switch event {
+        case let evt as ChannelEvent where evt == ChannelEvent.inputClosed:
+            // The remote peer half-closed the channel during the upgrade so we should close
+            context.close(promise: nil)
```
FranzBusch, line 159: Thanks for working on this!

Can we please go through the state machine here. In particular, if we are in the state unbuffering we should not close here but rather just hold the event until the unbuffering is done.

Can we also write a test for this and look if we need to do the same for the client upgrade handler?

Lukasa: We should also consider only closing if we haven't received a full upgrade request, and otherwise we should buffer this.

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -382,5 +382,23 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
...
+    @usableFromInline
+    enum CloseInboundAction {
+        case close
+        case `continue`
+    }
+
+    @inlinable
+    mutating func closeInbound() -> CloseInboundAction {
+        switch self.state {
+        case .initial, .awaitingUpgrader:
```
FranzBusch, line 394: We should immediately close in the state `initial`. In the state `awaitingUpgrader` we have to check if we have `seenFirstRequest == true`. If have seen that we should buffer the `inputClosed` and send it after the upgrading is finished. In the state `upgraderReady` we need to buffer the `inputClosed` indefinitely.

The reason why this is so complicated is that the client could potentially send a half closure after it send the initial upgrade request + some data on the upgraded protocol. We can't just unconditionally close here but we rather need to buffer the `inputClose` and unbuffer it with the data. I would recommend changing the `var buffer: Deque<NIOAny>` in the various states to `var buffer: Deque<enum(NIOAny | inputClosed)>`.

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -382,5 +382,23 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
...
+            self.state = .finished
+            return .close
+
+        default:
```
FranzBusch, line 398: Can we please spell out all cases explicitly so if we ever add one we get a compiler error here

Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
FranzBusch: It should be possible to emulate half closure with the embedded channel by just sending the event to the channel.

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
adam-fowler: Done!

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -382,5 +392,49 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
...
+            return .close
+
+        case .awaitingUpgrader(var awaitingUpgrader):
+            if awaitingUpgrader.seenFirstRequest {
+                // We should buffer the input close since we have seen the full request.
+                awaitingUpgrader.buffer.append(.inputClosed)
+                self.state = .awaitingUpgrader(awaitingUpgrader)
+                return .continue
+            } else {
+                // We shouldn't buffer. This means we are still expecting HTTP parts.
```
FranzBusch, line 415: NIT
```suggestion
                // We shouldn't buffer. This means we were still expecting HTTP parts.
```

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
```diff
@@ -2066,5 +2073,31 @@ final class TypedHTTPServerUpgradeTestCase: HTTPServerUpgradeTestCase {
         // We also want to confirm that the upgrade handler is no longer in the pipeline.
         try connectedServer.pipeline.waitForUpgraderToBeRemoved()
     }
+
+    func testHalfClosure() throws {
```
FranzBusch, line 2077: Can we add one or two more tests cases here:
1. That sends the request head then input close
2. That sends a full request head & end and then input close to check that we continue the upgrade

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
```diff
@@ -2066,5 +2073,31 @@ final class TypedHTTPServerUpgradeTestCase: HTTPServerUpgradeTestCase {
...
+                case ChannelError.inputClosed:
+                    errorCaught.wrappedValue = true
+                default:
+                    break
+                }
+            }
+        ) { (context) in
+        }
+
+        try client.close().wait()
```
FranzBusch, line 2097: Is this really testing the half closure here?

Tests/NIOHTTP1Tests/HTTPServerUpgradeTests.swift
adam-fowler: @FranzBusch 
I did 2

I'm not sure what the expected result is for 1
First I had to send a request head that included a content-length so a `.head` would be passed down the pipeline, but not an`.end` eg
```
OPTIONS * HTTP/1.1\r\nHost: localhost\r\ncontent-length: 10\r\nUpgrade: myproto\r\nKafkaesque: yup\r\nConnection: upgrade\r\nConnection: kafkaesque\r\n\r\n
```
With this it gets stuck in the state `.upgradeReady` because nothing else happens after inputClosed to forward the state machine onwards and the connection stays open. If I return `.close` from the state machine everything works but you explicitly said I shouldn't do that.

adam-fowler: @FranzBusch any thoughts

weissi:
ping @FranzBusch and @Lukasa

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -393,5 +403,49 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
...
     }
 
+    @usableFromInline
+    enum CloseInboundAction {
+        case close
+        case `continue`
+    }
+
+    @inlinable
+    mutating func closeInbound() -> CloseInboundAction {
```
FranzBusch, line 413: Naming NIT: Can we call this `inputClosed` instead please

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -393,5 +403,49 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
...
+            self.state = .upgraderReady(upgraderReady)
+            return .continue
+
+        case .unbuffering(var unbuffering):
+            unbuffering.buffer.append(.inputClosed)
+            self.state = .unbuffering(unbuffering)
+            return .continue
+
+        case .modifying, .finished:
+            return .continue
```
FranzBusch, line 446: In finished we should forward the `inputClosed` event. In `modifying` we should `fatalError` as we do in the other methods.

FranzBusch (commented):
Sorry for the delay. We are really close here. Just a few last comments

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
adam-fowler: done

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
adam-fowler: done

Sources/NIOHTTP1/NIOTypedHTTPServerUpgraderStateMachine.swift
```diff
@@ -379,8 +385,12 @@ struct NIOTypedHTTPServerUpgraderStateMachine<UpgradeResult> {
 
             if let element = unbuffering.buffer.popFirst() {
                 self.state = .unbuffering(unbuffering)
-
-                return .fireChannelRead(element)
+                switch element {
+                case .data(let data):
+                    return .fireChannelRead(data)
+                case .inputClosed:
+                    return .close
```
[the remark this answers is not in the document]
adam-fowler: done

Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
```diff
@@ -154,6 +154,25 @@ public final class NIOTypedHTTPServerUpgradeHandler<UpgradeResult: Sendable>: Ch
...
     }
 
+    public func userInboundEventTriggered(context: ChannelHandlerContext, event: Any) {
+        switch event {
+        case let evt as ChannelEvent where evt == ChannelEvent.inputClosed:
+            // The remote peer half-closed the channel during the upgrade. Should we close the other side
+            switch self.stateMachine.inputClosed() {
+            case .close:
+                self.upgradeResultPromise.fail(ChannelError.inputClosed)
+                context.close(promise: nil)
```
FranzBusch, line 164: Can we switch these two lines around so we first call close and then fulfil the promise. This should avoid potential bugs where the promise callbacks interact with the channel.

FranzBusch (commented):
This looks great. One last small comment!

Sources/NIOHTTP1/NIOTypedHTTPServerUpgradeHandler.swift
adam-fowler: done

FranzBusch (approved):
Great! Thanks for working on this.

PR #2759, opened 2024-06-28
Add EventLoop APIs for simpler scheduling of callbacks

simonjbeaumont:
## Motivation

The current `scheduleTask` APIs make use of _both_ callbacks and promises, which leads to confusing semantics. For example, on cancellation, users are notified in two ways: once via the promise and once via the callback. Additionally the way the API is structured results in unavoidable allocations—for the closures and the promise—which could be avoided if we structured the API differently.

## Modifications

This PR introduces new protocol requirements on `EventLoop`:

```swift
protocol EventLoop {
    // ...
    @discardableResult
    func scheduleCallback(at deadline: NIODeadline, handler: some NIOScheduledCallbackHandler) throws -> NIOScheduledCallback

    @discardableResult
    func scheduleCallback(in amount: TimeAmount, handler: some NIOScheduledCallbackHandler) throws -> NIOScheduledCallback

    func cancelScheduledCallback(_ scheduledCallback: NIOScheduledCallback)
}
```

Default implementations have been provided that call through to `EventLoop.scheduleTask(in:_:)` to not break existing `EventLoop` implementations, although this implementation will be (at least) as slow as using `scheduleTask(in:_:)` directly.

The API is structured to allow for `EventLoop` implementations to provide a custom implementation, as an optimization point and this PR provides a custom implementation for `SelectableEventLoop`, so that `MultiThreadedEventLoopGroup` can benefit from a faster implementation.

Finally, this PR adds benchmarks to measure the performance of setting a simple timer using both `scheduleTask(in:_:)` and `scheduleCallback(in:_:)` APIs using a `MultiThreadedEventLoopGroup`.

## Result

A simpler and more coherent API surface.

There is also a small performance benefit for heavy users of this API, e.g. protocols that make extensive use of timers: when using MTELG to repeatedly set a timer with the same handler, switching from `scheduleTask(in:_:)` to `scheduleCallback(in:_:)` reduces almost all allocations (and amortizes to zero allocations) and is ~twice as fast.

```
MTELG.scheduleCallback(in:_:)
╒═══════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *      │       0 │       0 │       0 │       0 │       0 │       0 │       0 │    1109 │
╘═══════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛

MTELG.scheduleTask(in:_:)
╒═══════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *      │       4 │       4 │       4 │       4 │       4 │       4 │       4 │     576 │
╘═══════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```


simonjbeaumont:

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,14 @@ public protocol EventLoop: EventLoopGroup {
...
     /// however, both of them should be implemented.
     func _promiseCompleted(futureIdentifier: _NIOEventLoopFutureIdentifier) -> (file: StaticString, line: UInt)?
+
+    /// Set a timer that will call a handler at the given time.
+    @discardableResult
+    func setTimer(for deadline: NIODeadline, _ handler: any NIOTimerHandler) -> NIOTimer
+
+    /// Set a timer that will call a handler after a given amount of time.
+    @discardableResult
+    func setTimer(for duration: TimeAmount, _ handler: any NIOTimerHandler) -> NIOTimer
```
glbrntt, line 366: Do we need this as a protocol requirement? Or should we just rely on the extension defined in terms of the deadline?

Sources/NIOCore/NIOTimer.swift
```diff
@@ -0,0 +1,94 @@
...
+public protocol NIOTimerHandler {
+    func timerFired(loop: any EventLoop)
+}
+
+/// An opaque handle that can be used to cancel a timer.
+///
+/// Users cannot create an instance of this type; it is returned by ``EventLoop/setTimer(for:_:)-5e37g``.
+///
+/// - Seealso: ``EventLoop/setTimer(for:_:)-5e37g``.
+public struct NIOTimer {
```
glbrntt, line 27: Can you mark this explicitly not Sendable?

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,14 @@ public protocol EventLoop: EventLoopGroup {
     /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
     /// however, both of them should be implemented.
     func _promiseCompleted(futureIdentifier: _NIOEventLoopFutureIdentifier) -> (file: StaticString, line: UInt)?
+
+    /// Set a timer that will call a handler at the given time.
+    @discardableResult
+    func setTimer(for deadline: NIODeadline, _ handler: any NIOTimerHandler) -> NIOTimer
```
glbrntt, line 362: I'm not sure about hiding the label of the `handler` here, I think the API would be clearer if users have to explicitly set `handler` (`_ handler: ...` is typically a closure)

Sources/NIOCore/NIOTimer.swift
simonjbeaumont: I'm sure there's a good reason but can you educate me? :)

I just looked at this afresh and thought that it could actually be annotated `Sendable`. Indeed doing so satisfied the compiler. 

This type is a value type that wraps an enum that holds sendable types.

Sources/NIOCore/NIOTimer.swift
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @inlinable
+    init(_ eventLoop: any NIOCustomTimerImplementation, id: UInt64) {
+        self.backing = .custom(eventLoop: eventLoop, id: id)
+    }
+
+    /// Cancel the timer associated with this handle.
+    @inlinable
+    public func cancel() {
```
simonjbeaumont, line 50: @glbrntt Thinking we could make this one safe to call from anywhere cheaply, since it doesn't return a value, we won't have to deal with the ELF? We can just make `NIOCustomTimerImplementation.cancelTimer(_:)` do event loop hop, if necessary?

Sources/NIOCore/EventLoop.swift
simonjbeaumont, line 362: Starting a separate thread to bikeshed how we name or document the requirements of this API.

Specifically, this API only tolerates being called on the event loop. Failure to do so should crash (I need to add the precondition in).

We discussed briefly off-PR naming these `unsafe*` but we concluded that this is a misuse of the "unsafe" terminology, which is reserved for undefined behaviour. Furthermore, these APIs are very safe because they will crash if used incorrectly. This is more like checked vs. unchecked (like with continuations and arithmetic) but we don't really want to provide an unchecked variant.

So we have some options (and I'm open to others). In all cases we will document the requirements:
1. Keep `setTimer(for:_:)` and rely only on documentation for the caller requirements.
2. `setCheckedTimer(for:_:)`.
3. `setTimerSync(for:_:)`.
4. `setTimer(for:_:iAmCallingThisFromTheEventLoop: Bool)`

My thoughts:

(1) seems in keeping with other Swift APIs that do runtime checks (e.g. arithmetic).
(2) would be in keeping with other Swift APIs that do runtime checks but only make sense if we were providing an unchecked counterpart, which we don't want to do.
(3) There's _some_ (not much) president for this "sync" terminology in NIO API but it predates `async` and so is arguably confusing now. This goes double for a timer API and could be misunderstood as a call that blocks until the timer fires.
(4) This just creates noise when used properly and still fatal errors when used incorrectly. So lose-lose.

My current conclusion is to keep the name as-is, but I'm happy to discuss it further.

Sources/NIOCore/EventLoop.swift
glbrntt: Makes sense, thanks for clarifying.

Sources/NIOCore/EventLoop.swift
FranzBusch: NIT: I am wondering if we should retain the `schedule` naming here and make this `scheduleTimer` instead of `setTimer`.

Sources/NIOCore/NIOTimer.swift
FranzBusch: This type should be marked as `Sendable` and we do expect this to be fully thread safe. Both the `Scheduled.cancel` and the new `EL.cancelTimer` method are fully thread safe

Sources/NIOCore/NIOTimer.swift
```diff
@@ -0,0 +1,94 @@
...
+/// Users cannot create an instance of this type; it is returned by ``EventLoop/setTimer(for:_:)-5e37g``.
+///
+/// - Seealso: ``EventLoop/setTimer(for:_:)-5e37g``.
+public struct NIOTimer {
+    @usableFromInline
+    enum Backing {
+        /// A task created using `EventLoop.scheduleTask(deadline:_:)`, used by default for `EventLoop` implementations.
+        case scheduledTask(Scheduled<Void>)
+        /// An identifier for a timer, used by `EventLoop` implementations that conform to `CustomTimerImplementation`.
+        case custom(eventLoop: any NIOCustomTimerImplementation, id: UInt64)
```
FranzBusch, line 33: Instead of storing the event loop here I am wondering if we can just store a `cancel` closure. I think I tried that with `ScheduledTask` once but I am unsure if I got it down to 0 allocations. That might allow us to avoid the whole `UInt64` completely

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -883,3 +894,20 @@ internal func assertExpression(_ body: () -> Bool) {
         return body()
     }())
 }
+
+extension SelectableEventLoop: NIOCustomTimerImplementation {
+    @inlinable
+    internal func setTimer(for deadline: NIODeadline, _ handler: any NIOTimerHandler) -> UInt64 {
+        let taskId = self.scheduledTaskCounter.loadThenWrappingIncrement(ordering: .relaxed)
```
FranzBusch, line 901: We now have two places here that create a task id and then a `ScheduledTask` can we de-dupe the code into a shared method?

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -883,3 +894,20 @@ internal func assertExpression(_ body: () -> Bool) {
...
+        let task = ScheduledTask(id: taskId, handler, deadline)
+        try! self._schedule0(.scheduled(task))
+        return taskId
+    }
+
+    @inlinable
+    internal func cancelTimer(_ id: UInt64) {
+        self._tasksLock.withLock {
+            self._scheduledTasks.removeFirst(where: { $0.id == id })
+        }
```
FranzBusch, line 911: Similar here can we de-dupe this code

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,45 @@ let benchmarks = {
...
+        let loop = group.next()
+
+        final class Timer: NIOTimerHandler {
+            func timerFired(loop: any EventLoop) {}
+        }
+        let timer = Timer()
+
+        benchmark.startMeasurement()
+        for _ in benchmark.scaledIterations {
+            let handle = loop.setTimer(for: .hours(1), timer)
```
FranzBusch, line 105: This benchmark shows that the `setTimer` method are actually thread safe otherwise this would blow up in your face

FranzBusch (changes requested):

Sources/NIOCore/NIOTimer.swift
```diff
@@ -0,0 +1,94 @@
...
+extension EventLoop {
+    @discardableResult
+    @inlinable
+    public func setTimer(for duration: TimeAmount, _ handler: any NIOTimerHandler) -> NIOTimer {
+        self.setTimer(for: .now() + duration, handler)
+    }
+}
+
+/// Extension point for `EventLoop` implementations implement a custom timer.
+public protocol NIOCustomTimerImplementation: EventLoop {
```
Lukasa, line 78: Is it necessary for us to add a new protocol? Any reason not to have `EventLoop` provide this directly?

Sources/NIOCore/EventLoop.swift
Lukasa, line 362: Let's add another bikeshed: is there any reason not to make this method generic?

Sources/NIOCore/NIOTimer.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A type that handles timer callbacks scheduled with ``EventLoop/setTimer(for:_:)-5e37g``.
+///
+/// - Seealso: ``EventLoop/setTimer(for:_:)-5e37g``.
+public protocol NIOTimerHandler {
+    func timerFired(loop: any EventLoop)
```
Lukasa, line 19: Another question here is whether we should make this protocol witness generic instead of handling an existential.

Sources/NIOCore/EventLoop.swift
Lukasa, line 366: While I'm here, can I suggest that we should use different labels instead of `for` in both? That will help with tab-completion a bit.

Sources/NIOCore/NIOTimer.swift
Lukasa, line 27: We might want to give this a name that implies it's more of a handle.

simonjbeaumont: How strongly do you feel about this? I did try this, but personally, I didn't like the term handle for a few reasons.

Mainly because "handle" is such a generic term in an API. If you have a value of a type, isn't that always a "handle" of sorts? Along these lines, this has one method called `cancel()` and a method is a verb and the subject of the verb is the thing you have. I want to "cancel the timer" not "cancel the timer handle". If we want to use the term handle, then the method should probably called `cancelTimer()`.

Secondarily, `NIOTimerHandler` and `NIOTimerHandle` were just very similar and reading the API surface took a lot more care.

Sources/NIOCore/NIOTimer.swift
simonjbeaumont: I will try this again, but my intuition is that this will not be as performant. Is avoiding the `UInt64` a goal?

Sources/NIOCore/NIOTimer.swift
simonjbeaumont: Yes, I tried to outline this in the PR description, but the goal here is to have a simple default implementation and have this as an opt-in extension point for implementations that want to do something different (e.g. more performant).

If we elevated these to the `EventLoop` protocol how would the default implementation look? Additionally, the protocol refinement here allows the type system to enforce that an implementation that's opting in here implements _both_ halves of the customisation point.

Sources/NIOPosix/SelectableEventLoop.swift
simonjbeaumont: I wanted to keep the `CustomTimerImplementation` conformance in its own extension here for clarity but, yes, let me take a look and see what refactoring I can do in the `SelectableEventLoop` to provide a reusable method.

Sources/NIOCore/NIOTimer.swift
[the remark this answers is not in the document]
simonjbeaumont: Absolutely! I was holding off polishing all the docs until we get the API nailed down. I'll keep this conversation unresolved until it's done.

I'll rename the parameter though! I think I saw `loop:` being used but now I checked again, this was mostly in test helpers, and I can see now that most of the API surface deals with `eventLoop:` labels 👍

simonjbeaumont: I can certainly take a look at this. Do we have any examples of API we provide that's generic over an event loop? Everything I've seen so far (`Channel`, `NIOLoopBound`) hold an existential IIUC.

Sources/NIOCore/EventLoop.swift
simonjbeaumont: Hm, it felt OK to me. Would you prefer the following?

- `setTimer(for: TimeAmount, ...)`
- `setTimer(forDeadline: NIODeadline, ...)`

I'm thinking of how I set timers verbally on my phone I normally always use "set a timer _for_" regardless of whether I then say a duration or an absolute time.

Open to suggestions here, though :)

Sources/NIOCore/EventLoop.swift
simonjbeaumont: I don't think you _schedule_ a timer because it can imply that the timer doesn't start immediately.

If it's more important to keep the _schedule_ verb, then we can have something like `scheduleCallback`, and if you really want to mix the _schedule_ verb with the `_timer_` noun, it would need to be something like `scheduleTimerToFinish(in:)` and `scheduleTimerToFinish(at:)`.

simonjbeaumont: OK, so that means we don't concern ourselves with any augmented API symbols to communicate the contract. I'll add some documentation to `NIOCustomTimerImplementation` methods that implementations _should_ tolerate being called from off the EL.

glbrntt: FWIW I don't like the "set" here, it sounds like we're providing a timer and storing it on the event loop, rather than having the event loop create a timer for us.

I'm okay with `scheduleCallback` – I think that actually best describes what the function does.

Sources/NIOCore/EventLoop.swift
Lukasa: `setTimer(for:)` and `setTimer(at:)` is probably the easiest spelling distinction.

Sources/NIOCore/NIOTimer.swift
Lukasa: Over an event loop is a good question. Probably we should avoid it here, but only because we'll inevitably want to store this in an existential anyway, so we just invite a translation. So let's leave it here.

Sources/NIOCore/NIOTimer.swift
Lukasa: Not very strongly.

Sources/NIOCore/NIOTimer.swift
Lukasa: I don't think we _ever_ want the default implementation: it's almost always worse than doing something bespoke. We have to provide it to avoid the API break, but we don't really care about it.

> If we elevated these to the EventLoop protocol how would the default implementation look?

More or less exactly as it does now. `cancel` is straightforward to implement, as you already have.

> Additionally, the protocol refinement here allows the type system to enforce that an implementation that's opting in here implements both halves of the customisation point.

This is handy, but doesn't necessarily outweigh the noise of the extra protocol.

More broadly, because we have added a new protocol, we make all event loops publicly offer these two methods:

```
    func setTimer(for deadline: NIODeadline, _ handler: any NIOTimerHandler) -> UInt64
    func cancelTimer(_ id: UInt64)
```

But we don't _want_ these two methods, we want the two that return typed handles. Nonetheless, users will see and be able to call these two on event loops. I think that's a strong enough API argument to reject this spelling.

simonjbeaumont: > More broadly, because we have added a new protocol, we make all event loops publicly offer these two methods:
> 
> ```
>     func setTimer(for deadline: NIODeadline, _ handler: any NIOTimerHandler) -> UInt64
>     func cancelTimer(_ id: UInt64)
> ```
> 
> But we don't _want_ these two methods, we want the two that return typed handles. Nonetheless, users will see and be able to call these two on event loops. I think that's a strong enough API argument to reject this spelling.

Presumably it would be only event loops that conform to this protocol that would publicly expose these methods and only when they are being held as a concrete type, rather than an `any EventLoop`?

However, I do concede that these symbols end up being public when ideally they should be internal.

I did this because I thought that was what the design called for: a slow default and the ability to provide something faster.

Maybe I've been thinking about exposing the wrong half of this. In the current PR I'm exposing the new methods (as a new protocol with new methods) and keeping the typed handle very opaque. Are you suggesting that I make the typed `NIOTimer` less opaque. Specifically that users can construct it as either a `scheduledTask` or a `custom` (will think on a better name if it's going to be public) which would allow event loop implementations to still opt in to provide a more performant implementation?

Sources/NIOCore/EventLoop.swift
simonjbeaumont: OK, the summary of this thread is two fold:
1. We want to have the `handler` label (i.e. no underscore).
2. We didn't like the _set_ and/or _timer_ language in the API.

The relevant targeted changes for these pieces of feedback are contained in their own commits:
- (1) is addressed by 950db0c84.
- (2) is addressed by ceabc7b28.

Sources/NIOCore/EventLoop.swift
simonjbeaumont: https://github.com/apple/swift-nio/pull/2759/files/df3cdb1c1f1f3ce8fb66fd81cdfa0cef3318c984#r1664313726 resulted in new API spellings. But these do use different prepositions for the `NIODeadline` and `TimeAmount` variants: `scheduleCallback(at:handler:)` and `scheduleCallback(in:handler:)`, respectively.

Sources/NIOCore/NIOTimer.swift
simonjbeaumont: I tried this and it resulted in an allocation. I have kept the commits in the history so you can see the attempt and the revert:


- b3e49036c: `feedback(attempted): Store a closure instead of UInt64`
- fb5e83536: `feedback(reverted, allocates): Store a closure instead of UInt64`

Sources/NIOCore/EventLoop.swift
simonjbeaumont: Done in 06a4ce76a. Although I didn't see much improvement from it.

Sources/NIOCore/NIOTimer.swift
simonjbeaumont: OK, so I've gone ahead and made the patch that I believe is more in line with this feedback.

We no longer have another protocol, but instead we now have to be careful about ensuring we provide the right overloads for _both_ the schedule and cancel functions and need a runtime crash if the contract has been violated.

IMO, this was a place we could have had the type system help us and any external authors of event loops.

I appreciate the argument that it would result in the exposing of lower level functionality to end users, but they don't have to call that, and, if they did it would be safe still.

Possibly a moot point now, but would you have considered a layered approach like that initially proposed if we made use of `@_spi` to expose it only to event loop implementors?

The targeted relevant changes for this feedback are in 89c57ec15.

simonjbeaumont (commented):
Thanks for the feedback so far folks. I've addressed it all and attempted to do so in a series of targeted commits that can be squashed on merge.

//cc @Lukasa @glbrntt @FranzBusch

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,29 @@ public protocol EventLoop: EventLoopGroup {
...
+    /// - NOTE: Event loops that provide a custom scheduled callback implementation **must** implement _both_
+    ///         `sheduleCallback(at deadline:handler:)` _and_ `cancelScheduledCallback(_:)`. Failure to do so will
+    ///         result in a runtime error.
+    @discardableResult
+    func scheduleCallback(in amount: TimeAmount, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback
+
+    /// Cancel a scheduled callback.
+    ///
+    /// - NOTE: Event loops that provide a custom scheduled callback implementation **must** implement _both_
+    ///         `sheduleCallback(at deadline:handler:)` _and_ `cancelScheduledCallback(_:)`. Failure to do so will
```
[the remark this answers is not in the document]
glbrntt: Similar comment applies to a few other places too.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -883,3 +894,23 @@ internal func assertExpression(_ body: () -> Bool) {
         return body()
     }())
 }
+
+extension SelectableEventLoop {
+    @inlinable
+    func scheduleCallback(at deadline: NIODeadline, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback {
+        let taskId = self.scheduledTaskCounter.loadThenWrappingIncrement(ordering: .relaxed)
+        let task = ScheduledTask(id: taskId, handler, deadline)
+        try! self._schedule0(.scheduled(task))
```
glbrntt, line 903: Can you drop a comment here as to why the `try!` is safe?

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -883,3 +894,23 @@ internal func assertExpression(_ body: () -> Bool) {
         return body()
     }())
 }
+
+extension SelectableEventLoop {
+    @inlinable
+    func scheduleCallback(at deadline: NIODeadline, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback {
+        let taskId = self.scheduledTaskCounter.loadThenWrappingIncrement(ordering: .relaxed)
```
glbrntt, line 901: nit: s/Id/ID

Tests/NIOPosixTests/NIOScheduledCallbackTests.swift
```diff
@@ -0,0 +1,151 @@
...
+        try await group.shutdownGracefully()
+
+        await fulfillment(of: [handler.timerDidFire], timeout: 0.01)
+        XCTAssertEqual(handler.firedCount, 0)
+    }
+}
+
+fileprivate final class MockTimerHandler: NIOScheduledCallbackHandler {
+    var firedCount = 0
+    var timerDidFire = XCTestExpectation(description: "Timer fired")
```
glbrntt, line 133: Maybe @FranzBusch remembers, but isn't there an issue with expectations and sendability? Should we just rely on a promise being completed instead?

glbrntt (changes requested):
Not sure about the method name for `NIOScheduledCallbackHandler` but beyond that this looks good aside from some nits.

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,29 @@ public protocol EventLoop: EventLoopGroup {
...
+    ///         result in a runtime error.
+    @discardableResult
+    func scheduleCallback(at deadline: NIODeadline, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback
+
+    /// Schedule a callback after given time.
+    ///
+    /// - NOTE: Event loops that provide a custom scheduled callback implementation **must** implement _both_
+    ///         `sheduleCallback(at deadline:handler:)` _and_ `cancelScheduledCallback(_:)`. Failure to do so will
```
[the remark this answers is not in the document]
simonjbeaumont: Now that we're converging on the naming I've made these proper DocC references instead in 5d6ac17a8.

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,29 @@ public protocol EventLoop: EventLoopGroup {
     /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
     /// however, both of them should be implemented.
     func _promiseCompleted(futureIdentifier: _NIOEventLoopFutureIdentifier) -> (file: StaticString, line: UInt)?
+
+    /// Schedule a callback at a given time.
+    ///
+    /// - NOTE: Event loops that provide a custom scheduled callback implementation **must** implement _both_
+    ///         `sheduleCallback(at deadline:handler:)` _and_ `cancelScheduledCallback(_:)`. Failure to do so will
```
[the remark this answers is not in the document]
simonjbeaumont: Now that we're converging on the naming I've made these proper DocC references instead in 5d6ac17a8.

Sources/NIOCore/EventLoop.swift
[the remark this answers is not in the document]
simonjbeaumont: Now that we're converging on the naming I've made these proper DocC references instead in 5d6ac17.

Sources/NIOPosix/SelectableEventLoop.swift
simonjbeaumont: I copy-pasted this code from elsewhere in the file, but sure, I fixed it up to use `taskID` now instead of `taskId` in 80d91f68d.

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,107 @@
...
+//===----------------------------------------------------------------------===//
+
+/// A type that handles callbacks scheduled with `EventLoop.scheduleCallback(at:handler:)`.
+///
+/// - Seealso: `EventLoop.scheduleCallback(at:handler:)`.
+public protocol NIOScheduledCallbackHandler {
+    /// This function is called at the scheduled time, unless the scheduled callback is cancelled.
+    ///
+    /// - Parameter eventLoop: The event loop on which the callback was scheduled.
+    func onSchedule(eventLoop: some EventLoop)
```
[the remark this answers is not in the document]
simonjbeaumont: I've updated the name to `handleScheduledCallback` in 2ec55431f.

Question: do we need to be more considerate about name clashes since we're expecting folks to conform their types to this protocol; e.g. should this be something like `handleNIOScheduledCallback`? 

This case isn't quite covered by https://github.com/apple/swift-nio/blob/main/docs/public-api.md, but it seems similar in nature.

WDYT?

simonjbeaumont (commented):
Thanks @glbrntt for your latest round of review. I have addressed most of the feedback. I wanted some additional thoughts on these two though:

- https://github.com/apple/swift-nio/pull/2759/files#r1670150749
- https://github.com/apple/swift-nio/pull/2759/files/ceabc7b28c80578dbdfece27df7acba5188bbcbe#r1668904283

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,45 @@ let benchmarks = {
...
+    ) { benchmark in
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { try! group.syncShutdownGracefully() }
+        let loop = group.next()
+
+        benchmark.startMeasurement()
+        for _ in benchmark.scaledIterations {
+            loop.scheduleTask(in: .hours(1), {})
+        }
+        benchmark.stopMeasurement()
```
glbrntt, line 84: FWIW: `stopMeasurement()` is redundant at the end of a benchmark (it's automatically called)

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,107 @@
...
+    public func cancel() {
+        self.eventLoop.cancelScheduledCallback(self)
+    }
+
+    /// The callback identifier, if the event loop uses a custom scheduled callback implementation; nil otherwise.
+    ///
+    /// - NOTE: This property is for event loop implementors only.
+    @inlinable
+    public var customCallbackID: UInt64? {
+        guard case .custom(let id) = backing else { return nil }
```
glbrntt, line 73:
```suggestion
        guard case .custom(let id) = self.backing else { return nil }
```

Sources/NIOPosix/SelectableEventLoop.swift
glbrntt: I think (2) is fine

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -460,21 +460,32 @@ internal struct ScheduledTask {
...
+    @usableFromInline
+    enum Kind {
+        case task(task: () -> Void, failFn: (Error) -> Void)
+        case callback(any NIOScheduledCallbackHandler)
+    }
+
+    @usableFromInline
+    let kind: Kind
+
+    // TODO: Should these be .init() or should they be static functions?
```
glbrntt, line 476: I think `init` is fine, it's `internal` API anyway so doesn't matter too much.

glbrntt (commented):
The `try!` needs fixing but otherwise I'm happy with this!

Sources/NIOPosix/SelectableEventLoop.swift
simonjbeaumont: After some offline discussion it turns out that I was mistaken. Scheduling _can_ throw in other circumstances, not just when the EL is shutdown.

Made `scheduleCallback` functions throwing in 9bf65f610.

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
simonjbeaumont: Yeah, it doesn't really matter I suppose, I've removed the TODO in 21fd1cca5.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
simonjbeaumont: I think this was a hangover from when I was playing around with a lower sample size and FWIW, it did make a minor difference having it called inside the closure that benchmark takes.

In any event, I've now removed them in bbf8f9f81.

Sources/NIOCore/NIOScheduledCallback.swift
simonjbeaumont: Done in dbba9e3dc.

Sources/NIOCore/NIOTimer.swift
Lukasa: > Possibly a moot point now, but would you have considered a layered approach like that initially proposed if we made use of `@_spi` to expose it only to event loop implementors?

I'm not convinced this would actually have worked. It's not immediately clear to me what the semantics of conformance to `@_spi` protocols actually _is_, so it's not clear to me that this would have prevented the protocol conformance being essentially public _or_ that it would have successfully populated the witness table.

If we assume that it would work as intended, though, I'm still not sure I'd have gone for it. I feel less strongly: the extra API noise is removed. But it would be a bit challenging to communicate to implementors how they should provide this implementation. Ultimately, we'd have to document it: otherwise it won't occur to implementors that they should implement this other protocol to implement these two functions. We have a lot of protocols, and it'll be easy to miss.

Once we're relying on documentation for correct implementation, I don't know that it's much worse to simply document that implementors must provide both methods. Yes, the type system will help correct implementations, but incorrect implementations will fail almost immediately the moment anyone tries to use them, including the implementor. So I'm just fundamentally not convinced that the downside risk is worth the extra abstractions.

Sources/NIOEmbedded/AsyncTestingEventLoop.swift
```diff
@@ -171,6 +171,15 @@ public final class NIOAsyncTestingEventLoop: EventLoop, @unchecked Sendable {
         return self.scheduleTask(deadline: self.now + `in`, task)
     }
 
+    @discardableResult
+    public func scheduleCallback(in amount: TimeAmount, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback {
+        // TODO: docs
+        /// Even though this type does not conform to `CustomTimerImplemenation`, it has a manual clock so we cannot
```
Lukasa, line 177: Nit: the comment here and in `EmbeddedEventLoop` refer to a protocol that no longer exists.

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,107 @@
...
+//===----------------------------------------------------------------------===//
+
+/// A type that handles callbacks scheduled with `EventLoop.scheduleCallback(at:handler:)`.
+///
+/// - Seealso: `EventLoop.scheduleCallback(at:handler:)`.
+public protocol NIOScheduledCallbackHandler {
+    /// This function is called at the scheduled time, unless the scheduled callback is cancelled.
+    ///
+    /// - Parameter eventLoop: The event loop on which the callback was scheduled.
+    func handleScheduledCallback(eventLoop: some EventLoop)
```
Lukasa, line 22: Should there be an optional method to implement when the scheduled callback is cancelled?

simonjbeaumont: Not sure really. I think we'd need to see how people use this a bit more to form an opinion on whether it's useful.

Also, if we were to provide it, would it be cleaner to add a non-optional requirement with a default empty implementation?

Sources/NIOCore/EventLoop.swift
```diff
@@ -356,6 +356,25 @@ public protocol EventLoop: EventLoopGroup {
     /// It is valid for an `EventLoop` not to implement any of the two `_promise` functions. If either of them are implemented,
     /// however, both of them should be implemented.
     func _promiseCompleted(futureIdentifier: _NIOEventLoopFutureIdentifier) -> (file: StaticString, line: UInt)?
+
+    /// Schedule a callback at a given time.
+    ///
+    /// - NOTE: Event loops that provide a custom scheduled callback implementation **must** also implement
+    ///         `cancelScheduledCallback`. Failure to do so will result in a runtime error.
+    @discardableResult
+    func scheduleCallback(at deadline: NIODeadline, handler: some NIOScheduledCallbackHandler) throws -> NIOScheduledCallback
```
FranzBusch, line 365: NIT: I personally find `callback` an overloaded term here since it is mostly used for meaning `closure` in Swift/ObjC. I liked the previous timer naming.

Sources/NIOCore/NIOScheduledCallback.swift
FranzBusch: non-optional requirement + default implementation is Swift's equivalent of optional requirement.

FWIW I think having an onCancel callback is useful. It allows you to centralise the state keeping into a single place.

Sources/NIOEmbedded/AsyncTestingEventLoop.swift
```diff
@@ -171,6 +171,14 @@ public final class NIOAsyncTestingEventLoop: EventLoop, @unchecked Sendable {
         return self.scheduleTask(deadline: self.now + `in`, task)
     }
 
+    @discardableResult
+    public func scheduleCallback(in amount: TimeAmount, handler: some NIOScheduledCallbackHandler) -> NIOScheduledCallback {
+        /// Even though this type does not implement a custom `scheduleCallback(at:handler)`, it has a manual clock so
+        /// we cannot rely on the default implemntation of `setTimer(for:handler:)`, which computes the deadline as an
```
glbrntt, line 177: nit: docs reference `setTimer(for:handler:)`

glbrntt (approved):
LGTM modulo a couple of nits.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,45 @@ let benchmarks = {
         )
     }
     #endif
+
+    Benchmark(
+        "MTELG.scheduleTask(in:_:)",
+        configuration: Benchmark.Configuration(
+            metrics: [.mallocCountTotal, .cpuTotal],
```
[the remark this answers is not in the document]
simonjbeaumont: OK, I switched to `.instructions`; is that what we want here, or should I just only measure allocations?

Sources/NIOCore/EventLoop.swift
simonjbeaumont: > I liked the previous timer naming.
😭 😰 

IMO @glbrntt is right here. This _is_ a callback in the truest sense of what is being _scheduled_.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,45 @@ let benchmarks = {
...
+            scalingFactor: .kilo
+        )
+    ) { benchmark in
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer { try! group.syncShutdownGracefully() }
```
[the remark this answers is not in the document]
FranzBusch: We can reuse ELs across benchmarks so I was thinking it might just be better if we create one EL that we share. I would just define a `let eventLoop = MultithreadedEventLoopGroup.singleton.any()` at the top.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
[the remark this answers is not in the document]
FranzBusch: I would like to see if `instructions` work. We are running inside docker and that might not work actually.

Sources/NIOCore/EventLoop.swift
FranzBusch: Maybe. It was just a nit and I am fine with both

Sources/NIOCore/NIOScheduledCallback.swift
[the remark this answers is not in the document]
simonjbeaumont: @glbrntt did you have any thoughts regarding the namespacing or is it fine the way it is?

The function takes a NIO type, so at worst it could cause an overload? Maybe it's OK the way it is now.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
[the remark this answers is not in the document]
simonjbeaumont: OK, let's keep the `.instructions` for now then and we can always remove it if it causes problems.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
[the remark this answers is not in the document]
simonjbeaumont: I'm not sure that's great, I think I'd rather each benchmark do it's own setup and teardown to ensure its unpolluted by the side effects of running other benchmarks.

Where we can, I'm happy to try and use the `setup`/`teardown` closure style, if you prefer it, and where we cannot, use this.

It also has the benefit for local reasoning of what's being benchmarked. It's _all_ contained in the call to `Benchmark { ... }`.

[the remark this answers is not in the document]
FranzBusch: I disagree and I just looked at the code in the benchmark again and we are already defining an EL for exactly this purpose. There is a static EL defined in this file called `eventLoop`.

The reason why I think it's not important to keep is that it should be irrelevant for the benchmark here. The important part of the benchmark is the scheduling and not how the EL is constructed and shut down so it keeps the benchmark more concise.

IMO we should definitely only have one style here and not mix static EL with one EL per benchmark.

simonjbeaumont:
In order to support cancellation, I'm gonna wait to land https://github.com/apple/swift-nio/pull/2800.

simonjbeaumont:
OK, cancellation support added. FYI @glbrntt @Lukasa @FranzBusch

simonjbeaumont:
> > ## Motivation
> > Scheduling a timer is currently an expensive operation because this is done using `EventLoop.scheduleTask(in:_:)`, which results in several allocations. A simple timer use case does not typically need all the functionality of this API and use cases that set many timers (e.g. for timeouts) pay this cost repeatedly.
> 
> @simonjbeaumont Could you add information here on _why_ `EventLoop.scheduleTask` results in more allocations than this? It shouldn't, maybe that's fixable? `EventLoop.scheduledTask` is meant to kinda be the simplest timer API but the PR contrasts it with "simple timers" which suggests that I'm missing something here.

Great questions, and sorry if the write up in the PR description wasn't detailed enough. 

IIUC, `scheduleTask` will allocate a bunch because it takes closures and needs promises. For cases where you just want a callback to an object you own, e.g. for setting repeated timers, this can be wasteful. 

> Also how many allocations are we talking?

I have added benchmarks and their results to the PR description. We're talking 4 allocations for using `scheduleTask` and none for `scheduleCallback` when using `SelectableEventLoop`.

Note that the default implementation for event loops of `scheduleCallback` is backed by `scheduleTask` so there won't be a win here, but it provides an opportunity for event loops to provide a custom implementation, which this PR has for `SelectableEventLoop`.

> And do you have something real-worldish where that difference is observable?

A real-world use case would be something like QUIC, where IIUC it makes use of lots of timers in the protocol. I also think that grpc-swift is likely to benefit from this.

One thing this discussion has prompted me to do is to benchmark _repeated_ `scheduleCallback` vs using `scheduleRepeatedTask` to see if there is some amortised benefit to the latter that I've overlooked.

FranzBusch:
> @simonjbeaumont Could you add information here on _why_ `EventLoop.scheduleTask` results in more allocations than this? It shouldn't, maybe that's fixable? `EventLoop.scheduledTask` is meant to kinda be the simplest timer API but the PR contrasts it with "simple timers" which suggests that I'm missing something here.

FWIW, I tried to make `EL.scheduleTask` as cheap as possible in the past already and tried to minimise allocations. IRCC I hit a point where I couldn't remove any more of the allocations because that would have had public API impact. One of them is for example the future that we allocate which we can completely avoid with this new API.

Lukasa:
Candidly, there's almost no way to make `scheduleTask` fast in the current API. `scheduleTask` is required to return a `Scheduled<T>`, which looks like this:

https://github.com/apple/swift-nio/blob/28f9cae996fd25c157251ced7205f399bd38cc0b/Sources/NIOCore/EventLoop.swift#L26-L53

This has the minimum cost of 1 promise, which is 1 allocation. However, in SelectableEventLoop it requires _two_ allocations, as our `cancellationTask` closes over `self` and a `taskId`, so we have to heap-allocate the closure context:

https://github.com/apple/swift-nio/blob/28f9cae996fd25c157251ced7205f399bd38cc0b/Sources/NIOPosix/SelectableEventLoop.swift#L322-L331

But it's worse! The internal datastructure that `SelectableEventLoop` uses to store the scheduled task is a `ScheduledTask` (natch). That looks like this:

https://github.com/apple/swift-nio/blob/28f9cae996fd25c157251ced7205f399bd38cc0b/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift#L510-L534

Note that we have two more closures. How many of them heap allocate? At least one:

https://github.com/apple/swift-nio/blob/28f9cae996fd25c157251ced7205f399bd38cc0b/Sources/NIOPosix/SelectableEventLoop.swift#L306-L312

As all closures over closures must allocate.

All this scaffolding means that this timer cannot help but cause 3 allocations. None of these can go away without losing the ability to handle cancellation, or breaking the existing API surface: the promise is clearly necessary, the cancel function is API and can't be removed (though we could do a deprecation cycle to remove it, probably), and the mere existence of the promise in alloc 1 forces alloc 3.

weissi:
> Candidly, there's almost no way to make `scheduleTask` fast in the current API.

Would it be worth thinking about improving the API with something that is very similar but doesn't suffer all these issues?



> I have added benchmarks and their results to the PR description. We're talking 4 allocations for using `scheduleTask` and none for `scheduleCallback` when using `SelectableEventLoop`.

Right, in a naive benchmark, 4 mallocs & 4 frees of allocations that are small and don't last very long is 59 nanoseconds on my machine. Please don't interpret that as saving allocations is not worthwhile, quite the opposite! But inventing duplicate APIs should be well motivated in my opinion.

<details>

```
// run as swiftc -O test.swift && time ./test
import Darwin

@inline(never)
@_optimize(none)
func blackhole(
    _ a: UnsafeMutableRawPointer,
    _ b: UnsafeMutableRawPointer,
    _ c: UnsafeMutableRawPointer,
    _ d: UnsafeMutableRawPointer
) {
    free(a)
    free(b)
    free(c)
    free(d)
}

for _ in 0..<1_000_000_000 {
    guard let a = malloc(24),
    let b = malloc(32),
    let c = malloc(48),
    let d = malloc(48) else {
        preconditionFailure()
    }

    blackhole(a, b, c, d)
}
```

</details>


> A real-world use case would be something like QUIC, where IIUC it makes use of lots of timers in the protocol. I also think that grpc-swift is likely to benefit from this.

I meant a real world use case where the ScheduledTask overhead is large enough that it's actually observable. I understand that QUIC, gRPC and many others use a lot of scheduled tasks, but 59 ns is very very little. So we need to be confident that any processing we would do in the timers doesn't totally dwarf the allocation time.

Lukasa:
Why are we adding only one order of magnitude? Each of these timers is kept per-connection, so the correct order of multiplicand is "number of connections / number of cores".

This API _can_ be made very similar to `scheduleTask`, by having it take a closure instead. In that case, it's an almost perfect replacement, except that it drops the problematic promise. This does open the user up to the sharp edge of the closure context, but expert users can avoid that sharp edge.

But I'd argue this makes your objection worse, not better. That API is much closer to the existing one than the current proposal, which increases confusion instead of decreasing it. We cannot deprecate the existing API, because it's a protocol requirement: event loops _must_ implement it, and indeed it must be used as the fallback implementation for the new API.

Lukasa:
As a related sidebar, these timers are typically set on the event loop, so they are a source of latency across the loop as a whole: we can't do I/O for any time where we're allocating memory for these timers.

weissi:
> Why are we adding only one order of magnitude? Each of these timers is kept per-connection, so the correct order of multiplicand is "number of connections / number of cores".

I meant adding an order of magnitude for _each_. In a best-case scenario 4 allocs&frees are 59ns. I'm suggesting to assume they 4 allocs & 4 frees come in at 500ns.


> This API _can_ be made very similar to `scheduleTask`, by having it take a closure instead. In that case, it's an almost perfect replacement, except that it drops the problematic promise. This does open the user up to the sharp edge of the closure context, but expert users can avoid that sharp edge.

Yeah, that's what I mean, why not add an API that's the same as the current one except for no `ScheduledTask` return (and other niche things).

 
> But I'd argue this makes your objection worse, not better. That API is much closer to the existing one than the current proposal, which increases confusion instead of decreasing it. We cannot deprecate the existing API, because it's a protocol requirement: event loops _must_ implement it, and indeed it must be used as the fallback implementation for the new API.

Just so I get it right, your argument is that having a new `scheduleTaskFireAndForget(in: TimeAmount, body: @escaping () -> Void) -> Void` (name tbd) is confusing? I can sympathise with that but I think we can find a resolution.

My thinking was/is: Let's start with a use case that makes an overhead of 50 to 500ns per schedule actually observable, then let's see what that use case needs. Maybe it needs cancellation support, maybe it doesn't, who knows. It makes it easier to design a new API if we actually now precisely what's necessary.

For example: I mostly _do_ need the returned `ScheduledTask` such that I'm able to clean up. And I also need the `.futureResult` usually such that I can schedule the next time (if not expressible by repeated schedule task).

The protocol requirement isn't an issue IMHO. The current API is quite expressive and supports a lot of use cases. But I won't deny that there are use cases where we don't need all features. (I have had those too)


> As a related sidebar, these timers are typically set on the event loop, so they are a source of latency across the loop as a whole: we can't do I/O for any time where we're allocating memory for these timers.

Completely. But there's much more overhead than just allocations. Literally the only thing I'm suggesting is a motivation that shows this as an actual win. I know it's a theoretical win but just saving the allocations might be small enough that it's not obversable.

Lukasa:
I'm struggling a bit with this feedback, because when I read it it seems like you're asking for two separate things, without clarifying whether they're an AND or an OR.

> Yeah, that's what I mean, why not add an API that's the same as the current one except for no ScheduledTask return (and other niche things).

The API is almost identical in function to the current one. It's not fire-and-forget. It's not even that we don't return a `ScheduledTask`: this API supports cancellation by returning a token, it notifies the callee about that cancellation so we can handle EL quiescing, so it's just as capable of being cancelled as the existing API. The only thing it doesn't have is a `Promise`.

This is because the `Promise` on the current one is _weird_. When we complete or cancel a `Scheduled` right now, we don't tell you once, we tell you twice: once via `Promise` and once via callback. That's a very confused idea: why are there two ways? These are much more naturally two APIs: one that fires a callback, and one that takes a promise:

```swift
protocol EventLoop {
    func scheduleTask(in: TimeAmount, @Sendable @escaping () -> Void, onCancel: @Sendable @escaping () -> Void) -> TaskHandle

    func scheduleTask(in: TimeAmount, notifying promise: EventLoopPromise<Void>) -> TaskHandle
}

struct TaskHandle {
    func cancel()
}
```

Of course, each of these APIs could easily be implemented in terms of the other.

But the current API is weirdly _both_ of these at the same time. You pass a callback, but you also get a promise. The two are both notified in the same way. That's weird, and hard to justify.

The new API says "screw it, just do the callback". But it acknowledges that we already have an API and it can do quite a lot, so in an attempt to differentiate itself it becomes far more restrictive. Essentially, it forces the user into a pattern that will allow amortized zero-allocation timers, by ensuring that you always give us something to call rather than us just invoking a closure.

But as you know very well, there is no difference between a delegate and a pair of closures, so we could take the pair of closures instead. The downside is that this API gets very close to obviating `ScheduledTask`, and it gets quite hard for us to tell users when they should use one or the other. I honestly thought the better thing was to provide a _more_ differentiated API, not a less differentiated one.

While we're here, the 59ns argument isn't interesting because Si has already written a benchmark, which is available in his original post but I will reproduce here:

```
MTELG.scheduleTask(in:_:)
╒═════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                  │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *        │       4 │       4 │       4 │       4 │       4 │       4 │       4 │     520 │
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (ns) * │     447 │     719 │     773 │     834 │     908 │    1098 │    1348 │     520 │
╘═════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

```
MTELG.setTimer(for:_:) without NIOCustomTimerImplementation conformance
╒═════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                  │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *        │       5 │       5 │       5 │       5 │       5 │       5 │       5 │     476 │
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (ns) * │     482 │     760 │     823 │     905 │     966 │    1119 │    1364 │     476 │
╘═════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

```
MTELG.setTimer(for:_:) with NIOCustomTimerImplementation conformance
╒═════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                  │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *        │       0 │       0 │       0 │       0 │       0 │       0 │       0 │    1071 │
├─────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (ns) * │     178 │     278 │     327 │     383 │     434 │     535 │     664 │    1071 │
╘═════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

P50 saves 446ns from a 773ns operation, p90 saves 474ns from a 908ns operation, and p99 saves 563ns from a 1.1us operation.

All of these are things that you can get away with doing millions of times in a second, but all of these are things that are likely to be done many thousands of times per second. As an example, QUIC connections will typically set at least two per-packet timers per connection (idle timer and probe timer). Using `ScheduledTask` in this context gives little room to move, and the EL offers no other timer solution. If we need a cheaper one, we can work out how to do that, but I don't think we have many ways to get cheaper than the implementation offered here.

weissi:
Thanks @Lukasa for writing this. I think this _is_ the missing motivation. I don't think the real motivation are the savings:

> P50 saves 446ns from a 773ns operation, p90 saves 474ns from a 908ns operation, and p99 saves 563ns from a 1.1us operation.

That's nice and exactly in line with the prediction of 59ns just for allocs which we upped by 10x (to account for worse cases and the other overhead like initialising the classes in those allocs) to 500ns. Of course that's fantastic to save this but without a use case where that makes even a 1% difference I didn't (and don't) think it's worth to invent new API _today_ before that use case actually is there.

But your post _does_ give the real reason (which is much much better than the performance): The current API is weird and not as performant as it could be. I still think a real world use case motivating doing the change _now_ would be nice (as opposed to when the use case is there which would provide more information). But I think your reasoning is good. My feedback to @simonjbeaumont is then: Copy Cory's post as the motivation (weird API) with the nice side effect: Your programs may also run a tiny bit faster if you schedule loads of timers.


> But as you know very well, there is no difference between a delegate and a pair of closures, so we could take the pair of closures instead. The downside is that this API gets very close to obviating `ScheduledTask`, and it gets quite hard for us to tell users when they should use one or the other. I honestly thought the better thing was to provide a _more_ differentiated API, not a less differentiated one.

I would've thought making it more aligned is better as it makes it easier for people to migrate from weird API to better API but I can see different arguments here.

Last question: Do we think we need to keep the old API (apart for SemVer reasons) or can we make the new API a complete replacement?

weissi:
> OK, @weissi; I updated the PR description to emphasise the API clarity and demote the performance win as a nice side-effect for the use cases for which it would matter.

Awesome, that makes much more sense to me (maybe adjust the title too).

simonjbeaumont:
Looks like @weissi is happy now with the motivation (not sure if that included a code review), and @glbrntt has approved.

@Lukasa, @FranzBusch: this looks like it's waiting for a re-review from one/both of you.

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,43 @@ let benchmarks = {
         )
     }
     #endif
+
+    Benchmark(
+        "MTELG.scheduleTask(in:_:)",
+        configuration: Benchmark.Configuration(
+            metrics: [.mallocCountTotal, .instructions],
```
FranzBusch, line 72: Can we use our `defaultMetrics` here? We don't have instruction based benchmarks yet

Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
```diff
@@ -64,4 +65,43 @@ let benchmarks = {
...
+            scalingFactor: .kilo
```
FranzBusch, line 73: Can we align the maximum duration/iterations with https://github.com/apple/swift-nio/pull/2839

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,166 @@
...
+
+    /// This function is called if the scheduled callback is cancelled.
+    ///
+    /// The callback could be cancelled explictily, by the user calling ``NIOScheduledCallback/cancel()``, or
+    /// implicitly, if it was still pending when the event loop was shut down.
+    ///
+    /// - Parameter eventLoop: The event loop on which the callback was scheduled.
+    func onCancelScheduledCallback(eventLoop: some EventLoop)
```
FranzBusch, line 30: Ultra naming nit: Should this be `didCancelScheduledCallback`? This follows how Swift APIs are often called when something did happen. `on` is more used when it is about to happen or in the process of happening. e.g. `viewDidLoad` in UIKit vs `onAppear` in SwiftUI

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,166 @@
...
+//===----------------------------------------------------------------------===//
+
+/// A type that handles callbacks scheduled with `EventLoop.scheduleCallback(at:handler:)`.
+///
+/// - Seealso: `EventLoop.scheduleCallback(at:handler:)`.
+public protocol NIOScheduledCallbackHandler {
+    /// This function is called at the scheduled time, unless the scheduled callback is cancelled.
+    ///
+    /// - Parameter eventLoop: The event loop on which the callback was scheduled.
+    func handleScheduledCallback(eventLoop: some EventLoop)
```
FranzBusch, line 22: Should these methods really take a generic EL? We usually never do that.

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -0,0 +1,166 @@
...
+    @inlinable
+    public func cancel() {
+        self.eventLoop.cancelScheduledCallback(self)
+    }
+
+    /// The callback identifier, if the event loop uses a custom scheduled callback implementation; nil otherwise.
+    ///
+    /// - NOTE: This property is for event loop implementors only.
+    @inlinable
+    public var customCallbackID: UInt64? {
```
FranzBusch, line 85: NIT: Do we need the `custom` here in the naming?

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -917,3 +929,33 @@ internal func assertExpression(_ body: () -> Bool) {
...
     )
 }
+
+extension SelectableEventLoop {
+    @inlinable
+    func scheduleCallback(
+        at deadline: NIODeadline,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback {
+        let taskID = self.scheduledTaskCounter.loadThenWrappingIncrement(ordering: .relaxed)
```
FranzBusch, line 939: Something for the future potentially. With this new fast implementation to schedule callbacks it might be worth looking at the normal `scheduleTask` APIs again to see if we can revamp their internal implementation to avoid some of the allocations that they currently have. We can't get rid of all of them i.e. the returned ELP/ELF but potentially the internal allocs where we close over state.

Sources/NIOCore/NIOScheduledCallback.swift
simonjbeaumont: I think that `on` is also used for when something has happened but is more ambiguous about whether it's after. IIRC it's `will` that implies before. E.g. `willSet` and `didSet`.

A quick `git-grep` in the NIO repo shows a use of both `on` and `did` quite a bit.

However, it's fair that the only `public` API uses `did`:

```
Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducerStrategies.swift:37:        public mutating func didYield(bufferDepth: Int) -> Bool {
Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducerStrategies.swift:42:        public mutating func didConsume(bufferDepth: Int) -> Bool {
Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift:70:    public func didYield(_ element: Element) {
Sources/NIOPosix/PendingDatagramWritesManager.swift:249:    public mutating func didWrite(
Sources/NIOPosix/PendingWritesManager.swift:201:    public mutating func didWrite(
```

I can change to using `did` 😅

Sources/NIOCore/NIOScheduledCallback.swift
simonjbeaumont: IMO it adds value when glancing at it as the property name implies that it's only relevant for custom implementations. How strongly do you feel about it. It's public API so if there's a consensus that this needs a different name I'll suck it up 😄

Sources/NIOPosix/SelectableEventLoop.swift
simonjbeaumont: Sure, I think @Lukasa's tome comment here will provide good content for someone wanting to see if we can improve it: https://github.com/apple/swift-nio/pull/2759#issuecomment-2252500183.

But as to whether we _should_, it sounded like the back and forth on the motivation was that the older APIs are a little confusing to hold. Without wanting to open another can of worms, I guess a bigger question is: are there real use cases for `scheduleTask` that cannot be implemented with `scheduleCallback`? If the answer is, _no_, then should we consider deprecating the old one given the discussion about its confusing shape?

Sources/NIOCore/NIOScheduledCallback.swift
simonjbeaumont: This was an ask in this PR. I _think_ it stemmed from a review comment from @Lukasa. We should converge on whether we do or don't want this generic I suppose.

simonjbeaumont (commented):
@FranzBusch addressed your feedback, thanks!

simonjbeaumont:
Gentle <sub>ping.</sub> 🥺

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -508,28 +508,38 @@ struct ErasedUnownedJob {
 
 @usableFromInline
 internal struct ScheduledTask {
+    @usableFromInline
+    enum Kind {
+        case task(task: () -> Void, failFn: (Error) -> Void)
+        case callback(any NIOScheduledCallbackHandler)
```
FranzBusch, line 514: One performance thought here. Currently we are storing an existential callback handler. However, we could just store the two closures itself which we can get while we are in the generic method. This way we would avoid calling through an existential on every scheduled callback task.

FranzBusch (approved):
This LGTM me now. I left one perf suggestion which I think should be noticeable on a micro level. If @Lukasa is also happy with I am happy to merge this

Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
simonjbeaumont: The requirements for `protocol NIOScheduledCallbackHandler` are _generic_ functions. Is there a way I can store these as closures in the enum associated value, without making the `Kind` generic?

FranzBusch: I forgot to add the second half of this. Yes this is why I think we should go back to `any EventLoop`. I assume the perf benefit outweighs this. This is an assumption but the scheduling and running of tasks is probably hotter than whatever we do in their callback with the passed EL. @Lukasa WDYT?

Lukasa: The functions being generic don't really matter here: we can store generic closures in the `Kind` type without promoting the generic to the `Kind` type itself. We won't get specialization, but that's fine.

So TL;DR: yes, changing `ScheduledEventLoop`'s representation to a pair of closures is probably the right thing to do.

simonjbeaumont: I spent some time working this and couldn't come up with a representation of this form that resulted in the amortized zero allocations we were looking for. Shall we shelve the suggestion on this thread? IIUC this is all internal anyway so we're not painting ourselves into a corner AFAICT?

Lukasa (approved):

simonjbeaumont:
Have we formally dropped 5.8 support now? This CI job is failing when it wasn't before.

Lukasa:
Nope, those are required CI checks

simonjbeaumont:
No worries, the `AsyncStream.makeStream` backfill for older platforms was made `fileprivate` in another test. Fixed that up. Should be good now.

simonjbeaumont:
Thanks for all the help getting this one over line @glbrntt @FranzBusch @Lukasa

PR #2760, opened 2024-07-02
[GHA] Introduce the first GitHub Action

FranzBusch:
# Motivation

We want to move our CI over to use GitHub Actions. We need to cover multiple different things like unit tests, benchmarks, soundness, API breaks, doc validation etc.

# Modification

This PR starts our journey to GitHub Actions by introducing the PR workflow. Right now this workflow only runs the Swift tests on our repository without any additional compiler flags like `-warnings-as-errors`. This aims to get us started and we will continue to modify them in subsequent PRs.

# Result

We got our first GH workflows running

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -1,0 +1,31 @@
+name: Pull Request
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.ref }}
+  cancel-in-progress: true
+ 
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        swift:
+        - image: swift5.8-jammy
+        - image: swift5.9-jammy
+        - image: swift5.10-noble
+        - image: swiftlang/swift:nightly-6.0-jammy
+        - image: swiftlang/swift:nightly-main-jammy
+    container:
+      image: ${{ matrix.swift.image }}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run tests
+      run: swift test
+      timeout-minutes: 20


PeterAdams-A (approved):

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        swift:
+        - image: swift5.8.1-jammy
+        - image: swift5.9.2-jammy
+        - image: swift5.10.1-noble
```
MahdiBM, line 21: Is this a private repository or did you mean:
```suggestion
        - image: swift:5.8-jammy
        - image: swift:5.9-jammy
        - image: swift:5.10-noble
```
This will automatically get the latest minor version of each image.

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,31 @@
...
+ 
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        swift:
+        - image: swift5.8.1-jammy
+        - image: swift5.9.2-jammy
```
Lukasa, line 20: Can we use less specific tags so we automatically take updates?

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,31 @@
+name: Pull Request
+
+on:
+  pull_request:
+    branches: [ main ]
+
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
MahdiBM, line 9: Doesn't really matter in this CI, but you can use `${{ github.ref }}` across all kinds of CI files.

[Vapor has been using](https://github.com/vapor/vapor/blob/ebbe71c89aa1b76a0920277760b12be7a2ec7c70/.github/workflows/test.yml#L3) the same thing for a long while and it's been working without fail:

```suggestion
  group: ${{ github.workflow }}-${{ github.ref }}
```

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,31 @@
+name: Pull Request
+
+on:
+  pull_request:
+    branches: [ main ]
```
MahdiBM, line 5: Might be a good idea to keep running tests on the main branch as well, that'll also enable the repo to have those "CI badge"s.
In that case you'll need to rename this file and change the name of the CI.

I'm also not sure if this'll run in any PRs since no PR is based on the "main" branch.
As you can see it's not running in this PR either, while it should.

[In Vapor](https://github.com/vapor/vapor/blob/ebbe71c89aa1b76a0920277760b12be7a2ec7c70/.github/workflows/test.yml#L6), it looks like this:

```suggestion
  pull_request: { types: [opened, reopened, synchronize, ready_for_review] }
  push: { branches: [ main ] }
```
The `pull_request` part will avoid running tests in e.g. draft pull requests.
Not sure if that's desirable, but Vapor has been using this.

.github/workflows/pull_requests.yml
MahdiBM: Ah I see you did mention this should use Apple's repository.

.github/workflows/pull_requests.yml
FranzBusch: Yeah I did that intentionally but you are right we should just pick up the latest patch updates

.github/workflows/pull_requests.yml
FranzBusch: I intentionally only focused on pull request based triggers for now. This is the first of many PRs to setup or GH action but I did not want to add all at once. Thanks for the trigger suggestion. Let's try it out!

.github/workflows/pull_requests.yml
```diff
@@ -0,0 +1,31 @@
+name: Pull Request
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.ref }}
```
MahdiBM, line 9: `${{ github.event.ref }}` is not quite right. It should be `${{ github.ref }}`.
`github.event` gives you access to the full webhook event which can be anything and doesn't necessarily contain a `ref`. As a matter of fact it doesn't contain a `ref` for pull-request events.

`${{ github.ref }}` on the other hand is another GitHub-injected value, just like `${{ github.event }}`, and is guaranteed to exist.

PR #2762, opened 2024-07-02
[GHA] Fix docker images

FranzBusch:


PeterAdams-A (approved):
Looks good.  I've at least checked I can pull these so should work.

PR #2763, opened 2024-07-04
[GHA] Add API breaking check

FranzBusch:
# Motivation

The next GH action pipeline that we need is to add our API breaking change checker.

# Modification

This PR adds an API breaking change job to our pull request workflow.

### DIFF
--- a/.github/workflows/pull_requests.yml
+++ b/.github/workflows/pull_requests.yml
@@ -6,7 +6,7 @@ on:
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
  
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_requests.yml
```diff
@@ -23,9 +23,26 @@ jobs:
...
+    timeout-minutes: 20
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
     - name: Run tests
       run: swift test
-      timeout-minutes: 20
\ No newline at end of file
+
+  api-breakage:
```
glbrntt, line 33: Small UI thing, should we give each job a more human readable `name`? E.g. "Unit tests", "Check for API breaking changes". This would be displayed in the checks on the PR rather than "unit-tests" or "api-breakage".

FranzBusch: Don't think that's possible since the following restriction for job ids apply

> The <job_id> must start with a letter or _ and contain only alphanumeric characters, -, or _.

.github/workflows/pull_requests.yml
```diff
@@ -23,9 +23,26 @@ jobs:
...
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        # We need to fetch everything otherwise only the head commit will be fetched.
+        fetch-depth: 0
+    - name: Mark the workspace as safe
+      # https://github.com/actions/checkout/issues/766
+      run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
+    - name: Run API breakage check
+      run: swift package diagnose-api-breaking-changes origin/main
```
PeterAdams-A, line 48: Should we be using GITHUB_BASE_REF ?

.github/workflows/pull_requests.yml
glbrntt: Yup, but you can add a `name:` key as well which will be used as the "pretty" job name, e.g. https://github.com/grpc/grpc-swift/blob/43aa501c505b0447291755fd350914c1df4e94bc/.github/workflows/ci.yaml#L9

.github/workflows/pull_requests.yml
FranzBusch: Not sure if that's a good idea. Do we really care about breaking changes between non-release branches?

.github/workflows/pull_requests.yml
FranzBusch: Ah amazing. Updated!

glbrntt (approved):

PeterAdams-A (approved):

PR #2764, opened 2024-07-04
[GHA] Documentation check

FranzBusch:
# Motivation

Next up replacing the documentation check.

# Modification

This moves the current approach of the script into a GH action. I tried just running the documentation check against all targets but that also results in warnings from downstream deps in their docs to show up.

I also fixed up some new errors that appeared since our current pipeline was running on 5.8.

# Result

Next CI job migrated to GHA

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -314,7 +314,7 @@ public protocol EventLoop: EventLoopGroup {
     /// This executor can be used to isolate an actor to a given ``EventLoop``. Implementers are encouraged to customise
     /// this implementation by conforming their ``EventLoop`` to ``NIOSerialEventLoopExecutor`` which will provide an
     /// optimised implementation of this method, and will conform their type to `SerialExecutor`. The default
-    /// implementation returns a ``NIODefaultSerialEventLoopExecutor`` instead, which provides suboptimal performance.
+    /// implementation provides suboptimal performance.
     @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
     var executor: any SerialExecutor { get }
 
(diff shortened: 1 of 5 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -314,7 +314,7 @@ public protocol EventLoop: EventLoopGroup {
     /// This executor can be used to isolate an actor to a given ``EventLoop``. Implementers are encouraged to customise
     /// this implementation by conforming their ``EventLoop`` to ``NIOSerialEventLoopExecutor`` which will provide an
     /// optimised implementation of this method, and will conform their type to `SerialExecutor`. The default
-    /// implementation returns a ``NIODefaultSerialEventLoopExecutor`` instead, which provides suboptimal performance.
+    /// implementation provides suboptimal performance.
```
PeterAdams-A, line 317: Feels like we're losing info here.  Perhaps that information would be better in a regular comment more than a doc comment though.

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -656,7 +656,7 @@ public struct FileSystem: Sendable, FileSystemProtocol {
 extension NIOSingletons {
     /// A suggestion of how many threads the global singleton ``FileSystem`` uses for blocking I/O.
     ///
-    /// The thread count is ``System/coreCount`` unless the environment variable
+    /// The thread count is the systems core count unless the environment variable
```
PeterAdams-A, line 659: Core count is slightly more complex than this as cgroup aware etc.

PeterAdams-A: Perhaps not super important to drop this distinction though.  Maybe available cores would be a better way of phrasing it.

Sources/NIOCore/EventLoop.swift
FranzBusch: I think we leaked implementation details here before. `NIODefault...Executor` is `internal` and we shouldn't talk about it in public docs.

PeterAdams-A (approved):

glbrntt (approved):

PR #2765, opened 2024-07-04
Pre-box some errors to reduce allocations

glbrntt:
Motivation:

Existential errors (`any Error`) are unconditionally boxed. In NIO we often create errors to fail promises. One place we do this is in the head channel handler when the pipeline is closed. This error is created regardless of whether there are actually any promises to fail. The result is an unnecessary allocation when every channel is closed. This is particuarly egregious when multiplexed protocols such as HTTP/2 are used.

As the majority of errors in NIO carry no information beyond their type/name we can declare these statically and pay the cost of each error just once rather than once per use.

Modifications:

- Add internal static constants for `ChannelError` and `EventLoopError` and use them throughout NIOCore and NIOPosix where they would otherwise be boxed.

Result:

Fewer allocations

### DIFF
--- a/Sources/NIOCore/Channel.swift
+++ b/Sources/NIOCore/Channel.swift
@@ -377,6 +377,7 @@ extension ChannelError {
     // 'any Error' is unconditionally boxed, avoid allocating per use by statically boxing them.
     static let _alreadyClosed: any Error = ChannelError.alreadyClosed
     static let _inputClosed: any Error = ChannelError.inputClosed
+    @usableFromInline
     static let _ioOnClosedChannel: any Error = ChannelError.ioOnClosedChannel
     static let _operationUnsupported: any Error = ChannelError.operationUnsupported
     static let _outputClosed: any Error = ChannelError.outputClosed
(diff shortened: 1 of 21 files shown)


Sources/NIOCore/Channel.swift
```diff
@@ -373,6 +373,16 @@ public enum ChannelError: Error {
     case unremovableHandler
 }
 
+extension ChannelError {
+    // 'any Error' is unconditionally boxed, avoid allocating per use by statically boxing them.
+    static let _alreadyClosed: any Error = ChannelError.alreadyClosed
```
FranzBusch, line 378: I am wondering if should file a compiler bug here and this is something that the compiler ought to be able to optimise. The error enum is in the same module and it has zero associated values. So it might as well store the instance in a global memory region

glbrntt: I don't think it's about the underlying error type; holding a `ChannelError` doesn't allocate. It's holding an `any Error` which always allocates.

Lukasa (approved):

PR #2766, opened 2024-07-04
[GHA] Unacceptable language check

FranzBusch:
# Motivation

Next up replacing part of our soundness script that checked for unacceptable language.

# Modification

This PR adds a new GH action that checks for unacceptable language.

# Result

One more script replaced.


.github/workflows/pull_requests.yml
```diff
@@ -68,4 +68,33 @@ jobs:
...
+        log "Checking for unacceptable language..."
+        PATHS_WITH_UNACCEPTABLE_LANGUAGE=$(git grep \
+          -l -F -w -E ${UNACCEPTABLE_WORDS} \
+        ) || true | /usr/bin/paste -s -d " " -
+
+        if [ -n "${PATHS_WITH_UNACCEPTABLE_LANGUAGE}" ]; then
+          fatal "❌ Found unacceptable language in files: ${PATHS_WITH_UNACCEPTABLE_LANGUAGE}."
+        fi
+
+        log "✅ Found no unacceptable language."
```
0xTim, line 100: Very minor nit - no newline at the end of the file adds a tiny bit of noise to the diff

.github/workflows/pull_requests.yml
```diff
@@ -68,4 +68,33 @@ jobs:
...
+        UNACCEPTABLE_WORDS="blacklist|whitelist|slave|master|sane|sanity|insane|insanity|kill|killed|killing|hang|hung|hanged|hanging"
+
+        log "Checking for unacceptable language..."
+        PATHS_WITH_UNACCEPTABLE_LANGUAGE=$(git grep \
+          -l -F -w -E ${UNACCEPTABLE_WORDS} \
+        ) || true | /usr/bin/paste -s -d " " -
+
+        if [ -n "${PATHS_WITH_UNACCEPTABLE_LANGUAGE}" ]; then
+          fatal "❌ Found unacceptable language in files: ${PATHS_WITH_UNACCEPTABLE_LANGUAGE}."
+        fi
```
simonjbeaumont, line 98: We can pull out the configuration into a required environment variable which will make it easier to factor this into its own action later. It also uses space separated, rather than forcing the configuration to use regex syntax.

Additionally, this uses some different `grep` flags which make sure we see the filename, line, column, and unacceptable term (not the whole line), and case-insensitive matching.

```suggestion
       test -n "${UNACCEPTABLE_LANGUAGE_PATTERNS:-}" || fatal "UNACCEPTABLE_LANGUAGE_PATTERNS unset"
       
       log "Checking for unacceptable language..."
       UNACCEPTABLE_LANGUAGE_LINES=$(git -C "${REPO_ROOT}" grep \
         -w -i -I -o \
         -H -n --column \
         -E "${UNACCEPTABLE_LANGUAGE_PATTERNS// /|}" \
       ) || true | /usr/bin/paste -s -d " " -
       
       if [ -n "${UNACCEPTABLE_LANGUAGE_LINES}" ]; then
         fatal " ❌ Found unacceptable language:
       ${UNACCEPTABLE_LANGUAGE_LINES}
       "
       fi
```

This should result in the following (tested as a standalone script in a project I have with some manual examples added). It will give one output line for each violation and in a form that could later be consumed to comment back to the PR.

```console
❯ env UNACCEPTABLE_LANGUAGE_PATTERNS="kill hanging" bash scripts/check-for-unacceptable-language.sh
** Checking for unacceptable language...
** ERROR:  ❌ Found unacceptable language:
Sources/PetstoreConsumerTestCore/Common.swift:179:1:Kill
Sources/_OpenAPIGeneratorCore/FeatureFlags.swift:35:17:hanging
Sources/_OpenAPIGeneratorCore/FeatureFlags.swift:36:11:kill
Sources/swift-openapi-generator/FilterCommand.swift:97:1:kill
```

.github/workflows/pull_requests.yml
```diff
@@ -68,4 +68,33 @@ jobs:
...
+        done
+
+  acceptable-language-check:
+    name: Acceptable language check
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run acceptable language check
```
simonjbeaumont, line 80:
```suggestion
    - name: Set unacceptable language variable
      run: echo 'UNACCEPTABLE_LANGUAGE_PATTERNS=blacklist whitelist slave master sane sanity insane insanity kill killed killing hang hung hanged hanging"' >> $GITHUB_ENV

    - name: Run acceptable language check
```

.github/workflows/reusable_pull_request.yml
```diff
@@ -83,4 +91,39 @@ jobs:
...
+        set -euo pipefail
+
+        log() { printf -- "** %s\n" "$*" >&2; }
+        error() { printf -- "** ERROR: %s\n" "$*" >&2; }
+        fatal() { error "$@"; exit 1; }
+
+        test -n "${UNACCEPTABLE_LANGUAGE_PATTERNS:-}" || fatal "UNACCEPTABLE_LANGUAGE_PATTERNS unset"
+      
+        log "Checking for unacceptable language..."
+        UNACCEPTABLE_LANGUAGE_LINES=$(git -C . grep \
```
weissi, line 118: variables in shell scrips ought to be lower case. ALL_CAPS variable names are
- for reserved shell variables (such as `PATH`, `EDITOR`, `SHELL`, `IFS`, ...)
- for adjusting _other_ programs' behaviour

> The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.

c.f. https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

.github/workflows/reusable_pull_request.yml
```diff
@@ -83,4 +91,39 @@ jobs:
...
+    timeout-minutes: 1
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run acceptable language check
+      shell: bash
+      env:
+        UNACCEPTABLE_LANGUAGE_PATTERNS: ${{ inputs.acceptable_language_check_word_list}}
+      run: |
+        set -euo pipefail
```
weissi, line 109: could we put scripts in actual files and run `shellcheck` over them? It's extra error-prone to program in a `.yml` file. If we just have a `dev/action_check_unacceptable_language.sh` or something, then you can just invoke that here.

That makes it easier to run `shellcheck` too

FranzBusch: The reason why these script are not in a file is that we want to move this reusable workflow out of the NIO repo and share it across all of our package repos for easier maintenance. Shared workflows can't pull in any files from their repo sadly. So we have to make the scripts inline in the yaml. That's not great but IMO worth the trade of copying scripts around

simonjbeaumont: I also initially wanted to have the scripts locally but I think the goal here is that we don't have to copy them into every repo long term. And yes, I use shellcheck and, in the OpenAPI Generator project, we use `shellcheck` in CI on the CI scripts themselves.

weissi: I get the issue that we want shared resources without having to copy stuff around. I don't think embedding them in `.yml` is a great solution for this issue.

What about creating a `swift-nio-ci-helpers` or something repo that we just `git clone` or have available too?  I know it's not great either but there has to be a better way than `yml` embedding. That makes testing the scripts locally hard (which leads to bugs), that makes shellcheck'ing them hard (leading to more bugs) and requires users to learn the weird yml syntax (which leads to frustration -> more bugs still)

FranzBusch: We already have a solution for testing locally which I have been using and it works great. It's called [act](https://nektosact.com/introduction.html) and allows you run the full workflows locally. It reduces the overhead of scripts a lot and contributors can just do `act pull_request` and run their whole CI checks

glbrntt: Definitely an increase in faff, but we could separate the scripts out and generate the yaml files with another script (basically just to inline the scripts into the yaml). We could also shellcheck the scripts in CI + check the yaml is up-to-date.

FranzBusch: I agree though that having the scripts inline isn't great and I want to change that in the future once we have everything in place and move the reusable workflow into its final place actually

weissi: Please, this requires users to install more software. That adds friction and frustration.

Let's take a step back: Embedding multiple files into one `.yml` file just because we can't figure out how to share code across a set of repos seems really wrong.

weissi: > having the scripts inline isn't great and I want to change that

Fab, let's do that. Clone `https://github.com/apple/swift-nio` if you need a crutch for now?

weissi: > Definitely an increase in faff, but we could separate the scripts out and generate the yaml files with another script (basically just to inline the scripts into the yaml). We could also shellcheck the scripts in CI + check the yaml is up-to-date.

Not a bad idea. Generating the yml from scripts isn't too bad either

0xTim: You don't even need to clone, you can just download a file from the tree, e.g. https://github.com/vapor/api-docs/blob/main/.github/workflows/build-and-deploy-docs-workflow.yml#L59-L64

NOTICE.txt
```diff
@@ -68,14 +68,14 @@ This product contains "sha1.c" and "sha1.h" from FreeBSD (Copyright (C) 1995, 19
...
   * HOMEPAGE:
-    * https://github.com/freebsd/freebsd/tree/master/sys/crypto
+    * https://github.com/freebsd/freebsd-src/tree/adf1fc7dd99ad514455f3dd17d289cc59cbae217/sys/crypto
 
 ---
 
 This product contains a derivation of Fabian Fett's 'Base64.swift'.
 
   * LICENSE (Apache License 2.0):
-    * https://github.com/fabianfett/swift-base64-kit/blob/master/LICENSE
```
glbrntt, line 78: also switched to `main`

Sources/CNIOSHA1/update_and_patch_sha1.sh
```diff
@@ -35,7 +35,7 @@ for f in sha1.c sha1.h; do
       echo "    - use welcoming language (soundness check)"
       echo "    - ensure BYTE_ORDER is defined"
       echo "*/"
-      curl -Ls "https://raw.githubusercontent.com/freebsd/freebsd/master/sys/crypto/$f"
+      curl -Ls "https://raw.githubusercontent.com/freebsd/freebsd/master/sys/crypto/$f" # ignore-unacceptable-language
```
glbrntt, line 38: This switched to `main`

.github/workflows/reusable_pull_request.yml
```diff
@@ -3,18 +3,26 @@ name: Pull Request
...
         type: boolean
         description: "Boolean to enable the docs check job. Defaults to true."
         default: true
+      acceptable_language_check_enabled:
+        type: boolean
+        description: "Boolean to enable the acceptable language check job. Defaults to true."
+        default: true
+      acceptable_language_check_word_list:
+        type: string
+        description: "List of unacceptable words. Defaults to a sensitive list of words."
```
glbrntt, line 24: ... a sensitive list of words? I'm not sure the list of words has feelings.

glbrntt (changes requested):

NOTICE.txt
```diff
@@ -41,14 +41,14 @@ This product contains a derivation of the Tony Stone's 'process_test_files.rb'.
...
-    * https://github.com/tonystone/build-tools/blob/master/source/xctest_tool.rb
+    * https://github.com/tonystone/build-tools/blob/cf3440f43bde2053430285b4ed0709c865892eb5/source/xctest_tool.rb
 
 ---
 
 This product contains NodeJS's llhttp.
 
   * LICENSE (MIT):
-    * https://github.com/nodejs/llhttp/blob/master/LICENSE-MIT
+    * https://github.com/nodejs/llhttp/blob/1e1c5b43326494e97cf8244ff57475eb72a1b62c/LICENSE-MIT
```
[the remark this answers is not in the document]
FranzBusch: I disagree. I think using exact sha's is better here since `main/master` can remove the files.

.github/workflows/reusable_pull_request.yml
simonjbeaumont: Yeah, maybe just remove this, the default is just below. Presumably any tooling that's showing the action can render the default if it wants from the YAML.

.github/workflows/reusable_pull_request.yml
```diff
@@ -83,4 +91,17 @@ jobs:
 
         for target in "${targets[@]}"; do
           swift package plugin generate-documentation --target "$target" --warnings-as-errors --analyze --level detailed
-        done
\ No newline at end of file
+        done
+
+  acceptable-language-check:
```
simonjbeaumont: nit: I see this changed from a check for unacceptable language, to a check for acceptable language. IMO that's a stretch...

scripts/acceptable_language_check.sh
```diff
@@ -0,0 +1,37 @@
...
+##
+##===----------------------------------------------------------------------===##
+
+set -euo pipefail
+
+log() { printf -- "** %s\n" "$*" >&2; }
+error() { printf -- "** ERROR: %s\n" "$*" >&2; }
+fatal() { error "$@"; exit 1; }
+
+test -n "${UNACCEPTABLE_LANGUAGE_PATTERNS:-}" || fatal "UNACCEPTABLE_LANGUAGE_PATTERNS unset"
```
simonjbeaumont: I think this would be clearer as an input to the script if this was `UNACCEPTABLE_LANGUAGE_WORDS` or similar. The fact that it becomes _patterns_ below is an implementation detail of the fact we're using grep. It would also make the actions yaml clearer IMO.

simonjbeaumont (commented):
Looking really nice. Agree with George's comments and left a couple of nits for consideration.

LGTM though so I'm not gonna mark it as needs changes from my end.

NOTICE.txt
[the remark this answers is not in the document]
glbrntt: For license I agree as long as long it aligns with the copy we vendored. For homepage I think it's odd to reference a specific commit, IMO makes more sense to just link to the repo.

glbrntt (approved):

PR #2767, opened 2024-07-05
[GHA] Create reusable workflow

FranzBusch:
# Motivation

In the future, we want to share the workflow definition across multiple repos to avoid duplication.

# Modification

This PR extracts the current workflow definition into a reusable workflow and calls the reusable workflow from a new workflow that is triggered on PR events. The new reusable workflow starts off with a few simple inputs to enable/disable the three current jobs. By default all jobs are enabled.

# Result

First step into reusability of our workflow.

### DIFF
--- a/.github/workflows/reusable_pull_request.yml
+++ b/.github/workflows/reusable_pull_request.yml
@@ -1,8 +1,20 @@
 name: Pull Request
 
 on:
-  pull_request:
-    types: [opened, reopened, synchronize, ready_for_review]
+  workflow_call:
+    inputs:
+      enable_unit_tests:
+        type: boolean
+        description: "Boolean to enable the unit tests job. Defaults to true."
+        default: true
+      enable_api_breakage_check:
+        type: boolean
+        description: "Boolean to enable the API breakage check job. Defaults to true."
+        default: true
+      enable_docs_check:
+        type: boolean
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable_pull_request.yml
```diff
@@ -1,8 +1,23 @@
...
 on:
-  pull_request:
-    types: [opened, reopened, synchronize, ready_for_review]
+  workflow_call:
+    inputs:
+      enable_unit_tests:
+        type: boolean
+        description: "Boolean to enable the unit tests job. Defaults to true."
+        required: true
+        default: true
```
glbrntt, line 10: Seems like an odd mix, if there's a default is the input actually required?

FranzBusch: Yeah the idea is to have each job be configurable and by default all jobs are enabled. So for most repos they just call this workflow and have to pass no inputs. Now some repos might want to turn on/off certain jobs and they would pass inputs.

glbrntt: Right, so inputs aren't `required` then?

FranzBusch: I see what you mean. Yeah removed the `required`

glbrntt (approved):

PR #2771, opened 2024-07-07
Allow in-place mutation of `NIOLoopBoundBox.value`

dnadoba:
### Motivation:
`NIOLoopBound` and `NIOLoopBoundBox` should be lightweight wrappers and support in-place mutation of CoW (Copy on Write) types. 

### Modifications:

- replace `set` accessor with `_modify` accessor
- add test that verifies in-place mutation of CoW types

### Result:

Less copying

### DIFF
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -47,9 +47,9 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
             self._eventLoop.preconditionInEventLoop()
             return self._value
         }
-        set {
+        _modify {
             self._eventLoop.preconditionInEventLoop()
-            self._value = newValue
+            yield &self._value
         }
     }
 }
@@ -136,9 +136,9 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
             self._eventLoop.preconditionInEventLoop()
             return self._value
         }
-        set {
+        _modify {
             self._eventLoop.preconditionInEventLoop()
-            self._value = newValue
+            yield &self._value
         }
     }
 }
--- a/Tests/NIOPosixTests/CoWValue.swift
+++ b/Tests/NIOPosixTests/CoWValue.swift
@@ -1,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A Copy on Write (CoW) type that can be used in tests to assert in-place mutation
+struct CoWValue: @unchecked Sendable {
+    private final class UniquenessIndicator {}
+
+    /// This reference is "copied" if not uniquely referenced
+    private var uniquenessIndicator = UniquenessIndicator()
+
+    /// mutates `self` and returns a boolean weather in was mutated in place
+    /// - Returns: true if mutations happened in-place, false if Copy on Write (CoW) was triggered
+    mutating func mutateInPlace() -> Bool {
+        guard isKnownUniquelyReferenced(&self.uniquenessIndicator) else {
+            self.uniquenessIndicator = UniquenessIndicator()
+            return false
+        }
+        return true
+    }
+}
--- a/Tests/NIOPosixTests/CoWValue.swift
+++ b/Tests/NIOPosixTests/CoWValue.swift
@@ -19,8 +19,8 @@ struct CoWValue: @unchecked Sendable {
     /// This reference is "copied" if not uniquely referenced
     private var uniquenessIndicator = UniquenessIndicator()
 
-    /// mutates `self` and returns a boolean weather in was mutated in place
-    /// - Returns: true if mutations happened in-place, false if Copy on Write (CoW) was triggered
+    /// mutates `self` and returns a boolean whether it was mutated in place or not
+    /// - Returns: true if mutation happened in-place, false if Copy on Write (CoW) was triggered
     mutating func mutateInPlace() -> Bool {
         guard isKnownUniquelyReferenced(&self.uniquenessIndicator) else {
             self.uniquenessIndicator = UniquenessIndicator()
--- a/Tests/NIOPosixTests/NIOLoopBoundTests.swift
+++ b/Tests/NIOPosixTests/NIOLoopBoundTests.swift
@@ -68,6 +68,14 @@ final class NIOLoopBoundTests: XCTestCase {
         }.wait())
     }
 
+    func testInPlaceMutation() {
+        var loopBound = NIOLoopBound(CoWValue(), eventLoop: loop)
+        XCTAssertTrue(loopBound.value.mutateInPlace())
+
+        let loopBoundBox = NIOLoopBoundBox(CoWValue(), eventLoop: loop)
+        XCTAssertTrue(loopBoundBox.value.mutateInPlace())
+    }
+
     // MARK: - Helpers
     func sendableBlackhole<S: Sendable>(_ sendableThing: S) {}
 


Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -47,9 +47,9 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
             self._eventLoop.preconditionInEventLoop()
             return self._value
         }
-        set {
+        _modify {
```
weissi, line 50: @Lukasa What's the current position on `_modify`, is that now allowed? Alternative is obvs to make a `withMutating { inout ... }`?

FranzBusch: We have used `_modify` and `_read` in multiple places now for performance. I think at this point it _sadly_ is allowed to be used.

FranzBusch (approved):

Sources/NIOCore/NIOLoopBound.swift
weissi: fine by me

weissi (approved):

Sources/NIOCore/NIOLoopBound.swift
dnadoba: I have completely switched to use `_modify` by default if the property type can be a CoW type.
Combining this with `consume self` if the property isn't simply a store property and I can implement efficient computed properties without worrying about triggering CoW. The `CoWValue` test helper then makes sure this actually checks out and stays like this.

weissi: Hmm, I don't think that's a good development. There are still open questions about `_modify` which is why it's underscored. And `_modify` is probably less benign than other underscored stuff like `@inline(__always)` or `@_cdecl`. The precise semantics haven't been settled and things may change.

Anybody aware of a post-2022 effort of getting these accessors specified?

dnadoba: AFAIK they will need to get specified for non-copyable types sooner rather than later. I hope they will then also be defined for copyable types although this hasn't been the case for the partial consumption of aggregates proposal which only works for non-copyable types.

FranzBusch: Yes the latest `Span` proposal calls out that these two modifiers will get properly pitched so we should have a solution for them soon.

Lukasa: My strong preference would be to combine inlinable with set if at all possible, in preference to read and modify with underscores.

dnadoba: I highly doubt that would be possible because of the dynamic call that cannot be fully devirtualize (_eventLoop.preconditionInEventLoop()) in between the get and set of `_value`. `NIOLoopBoundBox` is a class and `_eventLoop.preconditionInEventLoop()` could read or modify `_value` from it. Therefore the optimizer can't do this for us as it would be an exclusivity violation. We would need more performance annotations on `preconditionInEventLoop()` that promise that this isn't the case but these annotations are a lot more dangerous as they are not checked by the compiler. I'm not even sure these annotations work on protocol requirements.

Lukasa: I highly doubt isn’t “can’t”. Do we have a benchmark that demonstrated any improvement from this change?

weissi: I agree with this. Without a benchmark and an according Swift bug to fix this instance I don't think we should be using `Unsafe*` or `_stuff`. Exceptions apply (calling C, "known safe" `_stuff`)

dnadoba: I'm not aware of an existing benchmark but this PR has something better: a unit test that gives us clear answer that we are now mutating in-place.

In-place mutation is necessary for performance guarantees of most mutating operations on CoW types to be valid. Otherwise they are all moot, easily turning constant operations into linear operations and linear operations into quadratic operations. 

Remember that this is a public type used throughout the ecosystem and replaces a simple var or nonisolated(unsafe) var. We should encourage users to use the safer alternative without them needing to worry about in-place mutations and not require them to benchmark and report back to NIO. 

With that being said, here is a benchmark to double check that the compiler isn't messing anything up in release mode: https://github.com/apple/swift-nio/pull/2778
```
----------------------------------------------------------------------------------------------------------------------------
NIOLoopBoundBox.value metrics
----------------------------------------------------------------------------------------------------------------------------

╒══════════════════════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│             Malloc (total) *             │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞══════════════════════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│                 v2.68.0                  │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │    6725 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│               Current_run                │       1 │       1 │       1 │       1 │       1 │       1 │       1 │   10000 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│                    Δ                     │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │    3275 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│              Improvement %               │     100 │     100 │     100 │     100 │     100 │     100 │     100 │    3275 │
╘══════════════════════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

@weissi I don’t see a swift bug here as I have explained above. Please open a bug if you think otherwise.

Lukasa: I remain very uncomfortable about the fact we merged this. We have simple, non-underscored alternatives available.

While I’m here, @FranzBusch 

> Yes the latest Span proposal calls out that these two modifiers will get properly pitched so we should have a solution for them soon.

Does it? Where? The only reference I can see is this text from Future Directions

> This proposal includes some _read accessors, the coroutine version of the get accessor. _read accessors are not an official part of the Swift language, but are necessary for some types to be able to provide borrowing access to their internal storage. When a stable replacement for _read accessors is proposed and accepted, the implementation of Span will be adapted to the new syntax.

This text explicitly _disclaims_ the idea that a stable replacement is being proposed any time soon, and only covers `_read`.

Lukasa: In fact the absence of a stable `_read` is why no ContiguousStorage protocol is being added, which continues to suggest to me that we should expect semantic changes in the future.

dnadoba: >I remain very uncomfortable about the fact we merged this. We have simple, non-underscored alternatives available.

@Lukasa which do you have in mind?

Lukasa: Adding a `func withMutating` or `func modify` as we have done elsewhere.

dnadoba: I think that would only be a simple alternative if we remove the existing `value` property.
This is not only be a performance problem but also a security problem if the number of modifications to this property is attacker controlled.

Lukasa: I’m open to taking that nuclear option, but I don’t actually think it’s necessary. That line of argument extends to almost all Swift code. Leaving a tool out that enables accidentally quadratic code to be written is unfortunate, but does not rise to the level of a security issue; and it certainly doesn’t override the desire to avoid using unspecified language features.

Lukasa: After all, even the modify accessor doesn’t protect against code that splits the access from the mutation.

dnadoba: Keep in mind that `NIOLoopBoundBox` replaces a simple `var` and has been adopted throughout the ecosystem. If you [search for `NIOLoopBoundBox`](https://github.com/search?q=NIOLoopBoundBox&type=code) and look at the first non-nio owned project that uses `NIOLoopBoundBox` you will see that this is already a performance and potentially security problem where [Vapor is buffering an incoming request in memory](https://github.com/vapor/vapor/blob/a823735db57b46100b0c61cdfc5a08525b1e7cad/Sources/Vapor/Request/Request%2BBodyStream.swift#L93) which will result in exponentially time complexity instead of linear without `_modify`.

Next non-nio project, [Hummingbird, is effected as well in the same way](https://github.com/hummingbird-project/hummingbird-core/blob/2a16f079f78b2a746ad57b3931c366cd14961927/Sources/HummingbirdCore/Request/ByteBufferStreamer.swift#L276-L299).

Lukasa: All sound like good reasons to deprecate the setter. If we don’t think we can provide a suitable API on this type without `_modify` we may want to consider whether this type is a good type to offer at all. The problem would not exist with a user-defined Box type inside a NIOLoopBound, or indeed with any other reference type. It is only our choice to write the combo API that produced the issue.

weissi: The concern is that it's an underscored (i.e. not ready for prime time) feature whose semantics aren't fully specified.

The standard library is developed and shipped together with the compiler itself which makes it different. Example: Before Swift had a defined memory model, the standard library pretended there to be a defined memory.

Regarding swift-collections: That's up to the Swift collection authors.

Concerns regarding `_modify`: It's very easy to go wrong. For example if you have this code

```
_modify {
   print("A")
   yield &self._something
   print("B")
}
```

then you will not always see `B` after you saw an `A`. It depends if there was a throw in the mutation. This can leave your type in an inconsistent (possibly memory-unsafe) state and is arguably hard to see (there's no `try` or place to `catch`). Sure, if the `yield` is the last line in a `_modify` then it's probably fine.

But fundamentally, we can just offer a `func withMutableValue { inout Value ... }` which can achieve the same as `_modify` (worse syntax ofc).

---

This is just a decision for the NIO team. They maintain the code base, so they need to be comfortable with the amount of undefined, underscored stuff that's in the code base at present. New compiler versions could adjust things for underscored features.

dnadoba: That is fair and matches my understanding of the risk. It's a trade off, similar to the use of unsafe or unchecked swift features.  The [pitch](https://forums.swift.org/t/modify-accessors/31872) explains this as well. Are there any other known issues today? 

> possibly memory-unsafe

@weissi is that possible without additional use of unsafe swift? 

---

Agree, thats up to the NIO team.

weissi: > > possibly memory-unsafe
> @weissi is that possible without additional use of unsafe swift?

I would assume so because compiler-emitted code could be skipped too (unless it protects against that). The key point is: Without an `_modify` implementations checked against the current compiler invocations we don't really know.

FranzBusch: I am personally okay with using `_modify` since it is used by the compiler and standard library developers outside the stdlib e.g. `swift-collections`. It makes the code that developers write fast without them having to discover a `modify` method. Additionally, we are using more underscored attributes such as `@_exported` or `@_alwaysEmitIntoClient`.

Lukasa: If we're using `@_alwaysEmitIntoClient` we should definitely rethink that too.

`@_exported` is underscored, but its behaviour is well understood and safe. `_modify` is definitely not that.

As you know, the compiler and standard library developers explicitly do _not_ make the same compatibility promises we do. That allows them to freely use these annotations, as they can always resolve issues by dropping all prior Swift versions. We can't do that, so if `_modify` introduces issues, we'll be forced to fork or revert the change.

PR #2772, opened 2024-07-09
Made chunk reading explicit when using read or pread

rpecka:
Resolved an issue where reading a file in chunks using an unbounded range would read from the current file pointer even for regular files.

### Motivation:

When a `FileChunks` object is initialized, it checks if the range is set to `0..<Int.max`. If it is, then instead of reading using a series of offsets and the `pread` sys call, it uses the `read` syscall repeatedly. This is fine if the read is the first ever of this type for a file, but if we read this way twice, then the second call will be affected by the side file pointer effect from the first call.

For example:
```swift
// Read the file the first time. This will repeatedly call `read` until EOF.
var firstRead = ByteBuffer()
for try await chunk in handle.readChunks(in: 0..<Int.max, chunkLength: .bytes(128)) {
  XCTAssertLessThanOrEqual(chunk.readableBytes, 128)
  firstRead.writeImmutableBuffer(chunk)
}
// Read the file again using `read` until EOF. This will read zero bytes since the previous call moved the file pointer to the end of the file without resetting it.
var secondRead = ByteBuffer()
for try await chunk in handle.readChunks(in: 0..<Int.max, chunkLength: .bytes(128)) {
  XCTAssertLessThanOrEqual(chunk.readableBytes, 128)
  secondRead.writeImmutableBuffer(chunk)
}
```

The main issue is that the `read` syscall affects the file pointer while the `pread` syscall does not.

### Modifications:

- Add a `readChunksFromFilePointer` to `ReadableFileHandleProtocol` to explicitly read from the current file pointer instead of relying on the magic `0..<Int.max` range.
- Rename the `ChunkRange` cases to make what they are doing clearer.

### Result:

Reading `0..<Int.max` over and over again will have the same result each time.


Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -22,8 +22,8 @@ import NIOPosix
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
 public struct FileChunks: AsyncSequence {
     enum ChunkRange {
-        case entireFile
-        case partial(Range<Int64>)
+        case filePointerToEnd
+        case range(Range<Int64>)
```
glbrntt, line 26: I don't think there's any reason to change these names

Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -202,6 +202,13 @@ public protocol ReadableFileHandleProtocol: FileHandleProtocol {
     ///   - chunkLength: The maximum length of the chunk to read as a ``ByteCount``.
     /// - Returns: A sequence of chunks read from the file.
     func readChunks(in range: Range<Int64>, chunkLength: ByteCount) -> FileChunks
+
+    /// Returns an asynchronous sequence of chunks read from the file starting from the current file pointer.
+    ///
+    /// - Parameters:
+    ///   - size: The maximum length of the chunk to read as a ``ByteCount``.
+    /// - Returns: A sequence of chunks read from the file.
+    func readChunksFromFilePointer(chunkLength size: ByteCount) -> FileChunks
```
glbrntt, line 211: This isn't quite what I had in mind. Rather than adding new API I think we should use the type of the file to determine how to do the read inside `FileChunks`. Once we know the type of the file we can determine whether the range passed in is acceptable and then call the appropriate read function.

glbrntt (changes requested):

Sources/NIOFileSystem/FileHandleProtocol.swift
rpecka: This would make the checks that happen inside of `.readToEnd` redundant. Should we keep those or remove them?

Sources/NIOFileSystem/FileChunks.swift
rpecka: Isn't `entireFile` misleading because it read from the current file pointer, not from the beginning of the file?

Sources/NIOFileSystem/FileHandleProtocol.swift
rpecka: This also means that calling readToEnd will stat the file twice.

I’m trying out your recommended solution and it also causes problems because if we check the file type in the FileChunks initializer, then that means the function has to be async throws. But the readChunks function from `ReadableFileHandleProtocol` is neither async or throws so that would be an API change.

PR #2775, opened 2024-07-09
HTTPPart Documentation Clarification

dimitribouniol:
Improved documentation for HTTP Parts to clarify how often each part is received.

### Motivation:

It was previously ambiguous if an HTTP head or end part could be send more than once to encode streamed header information, which was not the case, so after clarifying on the [Swift Open Source Slack](https://swift-open-source.slack.com/archives/C9MMT6VGB/p1720516004684889), I propose this new wording.

### Modifications:

Modified the documentation for `HTTPPart`, and added documentation for each of its cases.

### Result:

Less confusion for the next person to explore NIO!


Sources/NIOHTTP1/HTTPTypes.swift
```diff
@@ -125,15 +125,30 @@ public struct HTTPRequestHead: Equatable {
...
-/// which may also contain any trailers that make up the message.
+/// The parts of a complete HTTP message, representing either a request or a response.
+/// 
+/// An HTTP message is made up of:
+/// - a request or status line with several headers, encoded by a single ``HTTPPart/head(_:)`` part,
+/// - zero or more ``HTTPPart/body(_:)`` parts,
+/// - and some optional trailers (represented as headers) in a single ``HTTPPart/end(_:)`` part.
+/// 
+/// To indicate that a complete HTTP message has been sent or received,
+/// an ``HTTPPart/end(_:)`` part should always be used, even when no trailers are included.
```
glbrntt, line 136: "should" is a suggestion, "must" is a requirement so let's use that here:

```suggestion
/// an ``HTTPPart/end(_:)`` part must be used, even when no trailers are included.
```

glbrntt (commented):
Thank you, this looks good, I left one small suggestion though

Sources/NIOHTTP1/HTTPTypes.swift
dimitribouniol: Good callout, thanks!

glbrntt (approved):

dimitribouniol:
@glbrntt Oops, didn't realizing force pushing to rebase on main and get failed tests to re-run would cancel the auto merge 😅

PR #2778, opened 2024-07-10
Dn loop bound box benchmark

dnadoba:
### Motivation:

@Lukasa and @weissi asked for a benchmark

### Modifications:

add benchmark

### Result:

```
----------------------------------------------------------------------------------------------------------------------------
NIOLoopBoundBox.value metrics
----------------------------------------------------------------------------------------------------------------------------

╒══════════════════════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│             Malloc (total) *             │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞══════════════════════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│                 v2.68.0                  │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │    6725 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│               Current_run                │       1 │       1 │       1 │       1 │       1 │       1 │       1 │   10000 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│                    Δ                     │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │   -1000 │    3275 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│              Improvement %               │     100 │     100 │     100 │     100 │     100 │     100 │     100 │    3275 │
╘══════════════════════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```


FranzBusch (approved):

Benchmarks/Benchmarks/NIOCoreBenchmarks/NIOLoopBoundBox.swift
```diff
@@ -0,0 +1,29 @@
...
+import NIOCore
+import NIOEmbedded
+
+func runNIOLoopBoundBoxInPlaceMutation() {
+    let embeddedEventLoop = EmbeddedEventLoop()
+    let boundBox = NIOLoopBoundBox([Int](), eventLoop: embeddedEventLoop)
+    boundBox.value.reserveCapacity(1000)
+
+    for i in 0..<1000 {
+        boundBox.value.append(i)
```
Lukasa, line 24: Can we run this multiple times, for the scaling factor? Ideally we’d only count allocations in the loop body, as we’re otherwise very sensitive to the runtime.

dnadoba: Done. This breaks the benchmarking comparison tool as it now thinks we have improved by 0% instead of 100% 😆
```
$ swift package benchmark baseline compare v2.68.0 --filter="NIOLoopBoundBox.value"  
[...]
----------------------------------------------------------------------------------------------------------------------------
NIOLoopBoundBox.value metrics
----------------------------------------------------------------------------------------------------------------------------

╒══════════════════════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│             Malloc (total) *             │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞══════════════════════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│                 v2.68.0                  │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │    1001 │       8 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│               Current_run                │       0 │       0 │       0 │       0 │       0 │       0 │       0 │     469 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│                    Δ                     │   -1001 │   -1001 │   -1001 │   -1001 │   -1001 │   -1001 │   -1001 │     461 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│              Improvement %               │       0 │       0 │       0 │       0 │       0 │       0 │       0 │     461 │
╘══════════════════════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```
@hassila is this a known issue?

hassila: Hmm, it seems it is normalizing to the slower benchmark instead of the reverse - looks like a plain bug but don't think we've got a case for it - out of office until August so will be a while before I can have a proper look. If you can, file a case would be great - otherwise I'll pick it up when back. 

I would suggest using a scaling factor for your bench mark though and pass it on instead to the function - that way you can get normalized numbers per subiteration

Benchmarks/Benchmarks/NIOCoreBenchmarks/NIOLoopBoundBox.swift
```diff
@@ -0,0 +1,37 @@
...
+func runNIOLoopBoundBoxInPlaceMutation(benchmark: Benchmark) {
+    let embeddedEventLoop = EmbeddedEventLoop()
+    let boundBox = NIOLoopBoundBox([Int](), eventLoop: embeddedEventLoop)
+    boundBox.value.reserveCapacity(1000)
+
+    benchmark.startMeasurement()
+
+    for _ in benchmark.scaledIterations {
+        boundBox.value.removeAll(keepingCapacity: true)
+        for i in 0..<1000 {
```
Lukasa, line 28: We should only have the scaled iterations loop, not this inner one.

dnadoba: How does that protect against runtime allocations?

dnadoba: Also, this is now no longer testing what I wanted to test: the performance of repeated mutation of value. CoW types like Array generally can have one off allocations but that isn't the real problem. The problem is if it happens on every iteration in a loop.

Lukasa: For the future, the reason looping `scaledIterations` times helps is that the malloc values output are divided by `scaledIterations`. This means any allocation that doesn't happen _every_ loop iteration is folded away, so spurious runtime allocations that may slightly differ between runs don't affect our thresholds.

Lukasa (approved):
I'm generally happy with this benchmark.

PR #2779, opened 2024-07-10
Combine the two NIOAsyncChannel channel handlers

glbrntt:
Motivation:

The NIOAsyncChannel allocates 12 times on init. 4 of these allocations come from creating two channel handlers and two channel handler contexts. There's no inherent reason that these channel handlers can't be combined to eliminate two allocations (one handler and one context).

Modifications:

- Combine `NIOAsyncChannelInboundStreamChannelHandler` and `NIOAsyncChannelOutboundWriterHandler` into a single `NIOAsyncChannelHandler`. Most of this was straightforward as only a few handler operations were duplicated across both.
- Add a 'NIOAsyncChannelHandlerWriterDelegate' in place of the 'NIOAsyncChannelOutboundWriterHandler.Delegate'. One knock on from this is that the new delegate stores callbacks rather than the concrete type of the handler. This is necessary to prevent the generics from the new channel handler bubbling up to the outbound writer (which would break API and be somewhat odd).

Result:

Fewer allocations


FranzBusch (approved):
Nice!

PR #2780, opened 2024-07-10
[GHA] Benchmark job

FranzBusch:
# Motivation

We need to run our `benchmark-package` based benchmarks on GH action as well. This is another part that we can make reusable for other packages.

# Modification

Add a new `benchmarks-linux` job to the reusable workflow.

# Result

One more CI flow migrated.


FranzBusch:

glbrntt (approved):
Looks good but just wondering about the approach here as each job implicitly picks up all Swift versions to run on which isn't that flexible. One example which comes to mind is a new package starting with support only for the latest version of Swift. It seems like the step should be the reusable component (I'm not familiar with GH workflows so don't know how feasible this is) or that each job should take platforms as a parameter.

FranzBusch:
> Looks good but just wondering about the approach here as each job implicitly picks up all Swift versions to run on which isn't that flexible. One example which comes to mind is a new package starting with support only for the latest version of Swift. It seems like the step should be the reusable component (I'm not familiar with GH workflows so don't know how feasible this is) or that each job should take platforms as a parameter.

Making the minimum Swift version configurable is on the list. I was thinking about having a general input of minimum version and then skipping both unit and benchmarks depending on that input.

glbrntt:
What about other platforms?

FranzBusch:

FranzBusch:
> What about other platforms?

I think we can use a similar concept. It depends on well toolchain installation is going to work on the macOS runners.

FranzBusch:

PR #2781, opened 2024-07-11
[GHA] Add license header check

FranzBusch:
# Motivation

We need to make sure all code files have the appropriate license headers in place. This is currently part of the soundness check and we need to replace this with a GH action.

# Modification

Adds a new job to the reusable workflow the check all files for license headers. Since some files are ignored every repo can specify a `.licenseignore` file.

# Result

Last part of the soundness script migrated.


.github/workflows/reusable_pull_request.yml
```diff
@@ -135,4 +143,17 @@ jobs:
...
+    name: License headers check
+    if: ${{ inputs.license_header_check_enabled }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run license header check
+      env:
+        PROJECT_NAME: ${{ inputs.license_header_check_project_name}}
```
glbrntt, line 158:
```suggestion
        PROJECT_NAME: ${{ inputs.license_header_check_project_name }}
```

.licenseignore
```diff
@@ -0,0 +1,16 @@
+Sources/CNIOAtomics/src/cpp_magic.h
```
glbrntt, line 1: What's the format for this file? Presumably paths relative to the root of the repo, but are wildcards supported? Can whole directories be ignored (e.g. vendored in dependencies)?

scripts/check-license-header.sh
```diff
@@ -0,0 +1,105 @@
...
+  ":(exclude).mailfilter" \
+  ":(exclude).mailmap" \
+  ":(exclude).spi.yml" \
+  ":(exclude).swift-format" \
+  ":(exclude).github/*" \
+  ":(exclude)*.md" \
+  ":(exclude)*.txt" \
+  ":(exclude)*.yaml" \
+  ":(exclude)*.json" \
+  ":(exclude)Package.swift" \
```
glbrntt, line 53: Can you add `Package@swift-*.swift` as well?

scripts/check-license-header.sh
```diff
@@ -0,0 +1,105 @@
...
+
+while IFS= read -r file_path; do
+  file_basename=$(basename -- "${file_path}")
+  file_extension="${file_basename##*.}"
+
+  case "${file_extension}" in
+    swift) expected_file_header=$(sed -e 's|@@|//|g' <<<"${expected_file_header_template}") ;;
+    h) expected_file_header=$(sed -e 's|@@|//|g' <<<"${expected_file_header_template}") ;;
+    c) expected_file_header=$(sed -e 's|@@|//|g' <<<"${expected_file_header_template}") ;;
+    yml) expected_file_header=$(sed -e 's|@@|##|g' <<<"${expected_file_header_template}") ;;
```
glbrntt, line 79: Can you add `yaml` as well?

scripts/check-license-header.sh
```diff
@@ -0,0 +1,105 @@
...
+    sh) expected_file_header=$(cat <(echo '#!/bin/bash') <(sed -e 's|@@|##|g' <<<"${expected_file_header_template}")) ;;
+    py) expected_file_header=$(cat <(echo '#!/usr/bin/env python3') <(sed -e 's|@@|##|g' <<<"${expected_file_header_template}")) ;;
+    rb) expected_file_header=$(cat <(echo '#!/usr/bin/env ruby') <(sed -e 's|@@|##|g' <<<"${expected_file_header_template}")) ;;
+    *) fatal "Unsupported file extension for file (exclude or update this script): ${file_path}" ;;
```
glbrntt, line 83: Should we ignore all json since it can't have a header as is likely used for config/input for a bunch of things?

.licenseignore
FranzBusch: The format is almost `.gitignore` like. So glowing and everything is supported i.e. `**/Package.swift` works.

scripts/check-license-header.sh
FranzBusch: JSON is ignored via the standard excludes so we don't have to handle it.

scripts/check-license-header.sh
FranzBusch: We are actually ignoring all `yml` and `yaml` files so we don't have to check them.

scripts/check-license-header.sh
glbrntt: Ah, missed that one!

glbrntt (approved):
Nice!

scripts/check-license-header.sh
simonjbeaumont: Let’s not have standard excludes. Too easy for things to accidentally creep through.

simonjbeaumont (changes requested):
I really don’t think the script should bake in the excludes.

simonjbeaumont (approved):
Awesome work @FranzBusch 👏

.licenseignore
```diff
@@ -1,3 +1,34 @@
...
+*.modulemap"
+**/*.modulemap"
+**/*.docc/*"
+*.xcprivacy"
+**/*.xcprivacy"
+*.symlink"
+**/*.symlink"
+Dockerfile"
+**/Dockerfile"
+Snippets/*"
```
glbrntt, line 31: These all have trailing `"`

glbrntt (changes requested):

glbrntt (approved):

PR #2782, opened 2024-07-11
Add an `AcceptBackoffHandler` to the async server bootstraps

FranzBusch:
# Motivation

When encountering a file descriptor limit we are currently closing the server socket when using the async bootstraps since the error caught is delivered via the `inbound` sequence of the server socket async sequence.

# Modification

This PR adds the `AcceptBackoffHandler` in the async bootstrap case to the pipeline to avoid us closing the server socket channel when we hit the file descriptor limit.

# Result

We are now gracefully handling file descriptor limits.

### DIFF
--- a/Sources/NIOCore/ChannelHandlers.swift
+++ b/Sources/NIOCore/ChannelHandlers.swift
@@ -23,6 +23,7 @@ public final class AcceptBackoffHandler: ChannelDuplexHandler, RemovableChannelH
 
     private var nextReadDeadlineNS: Optional<NIODeadline>
     private let backoffProvider: (IOError) -> TimeAmount?
+    private let shouldForwardIOErrorCaught: Bool
     private var scheduledRead: Optional<Scheduled<Void>>
 
     /// Default implementation used as `backoffProvider` which delays accept by 1 second.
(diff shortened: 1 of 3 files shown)


FranzBusch:
@Lukasa @glbrntt Would love to get your feedback on this approach before I add tests to it.

glbrntt:
This seems reasonable for running out of descriptors, what other `IOError`s do we expect, and is it reasonable to backoff for those rather than closing?

Sources/NIOCore/ChannelHandlers.swift
```diff
@@ -68,8 +68,9 @@ public final class AcceptBackoffHandler: ChannelDuplexHandler, RemovableChannelH
                     scheduleRead(at: self.nextReadDeadlineNS!, context: context)
                 }
             }
+        } else {
```
Lukasa, line 71: I'm inclined to make this behaviour configurable. Users may be relying on the existing reporting, which makes sense as a default behaviour, but for the use-case we have it may well be sensible to suppress the error when it is "handled".

Lukasa (approved):

PR #2783, opened 2024-07-11
Fix benchmark thresholds update script

FranzBusch:
# Motivation

We added new CI to run both nightly main and nightly next but forgot to update the benchmark thresholds script

# Modification

This PR updates the benchmark thresholds script and adds a nightly next docker compose image.

# Result

We can use the update benchmark thresholds script again.


dnadoba (approved):
Works! Although it seems some of the existing tests are a bit flaky. We will see it in the next run of https://github.com/apple/swift-nio/pull/2778

PR #2784, opened 2024-07-11
Use the new Android overlay and Bionic module from Swift 6

finagolfin:
### Motivation:

Get this repo building again for Android with the new overlay

### Modifications:

- Import the new module or overlay wherever `Glibc` is used
- Keep this repo building with Swift 5 by duplicating some declarations

### Result:

All the same tests keep passing on my Android CI, finagolfin/swift-android-sdk#158

I imported the smaller Bionic module where possible, and only used `Android` when files invoked C functions not in the new Bionic module, swiftlang/swift#72161.

### DIFF
--- a/Sources/NIOCore/BSDSocketAPI.swift
+++ b/Sources/NIOCore/BSDSocketAPI.swift
@@ -65,6 +65,8 @@ internal typealias socklen_t = ucrt.size_t
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Android)
+import Android
 #endif
 import CNIOLinux
 
@@ -90,9 +92,15 @@ private let sysInet_pton: @convention(c) (CInt, UnsafePointer<CChar>?, UnsafeMut
 #endif
 
 #if os(Android)
+#if compiler(>=6.0)
+let IFF_BROADCAST: CUnsignedInt = numericCast(Android.IFF_BROADCAST.rawValue)
+let IFF_POINTOPOINT: CUnsignedInt = numericCast(Android.IFF_POINTOPOINT.rawValue)
+let IFF_MULTICAST: CUnsignedInt = numericCast(Android.IFF_MULTICAST.rawValue)
+#else
 let IFF_BROADCAST: CUnsignedInt = numericCast(SwiftGlibc.IFF_BROADCAST.rawValue)
 let IFF_POINTOPOINT: CUnsignedInt = numericCast(SwiftGlibc.IFF_POINTOPOINT.rawValue)
 let IFF_MULTICAST: CUnsignedInt = numericCast(SwiftGlibc.IFF_MULTICAST.rawValue)
+#endif
 #if arch(arm)
 let SO_RCVTIMEO = SO_RCVTIMEO_OLD
 let SO_TIMESTAMP = SO_TIMESTAMP_OLD
--- a/Sources/NIOPosix/Linux.swift
+++ b/Sources/NIOPosix/Linux.swift
@@ -80,20 +80,24 @@ internal enum Epoll {
     internal static let EPOLL_CTL_MOD: CInt = numericCast(CNIOLinux.EPOLL_CTL_MOD)
     internal static let EPOLL_CTL_DEL: CInt = numericCast(CNIOLinux.EPOLL_CTL_DEL)
 
-    #if os(Android)
+    #if canImport(Android) || canImport(Musl)
+    internal static let EPOLLIN: CUnsignedInt = numericCast(CNIOLinux.EPOLLIN)
+    internal static let EPOLLOUT: CUnsignedInt = numericCast(CNIOLinux.EPOLLOUT)
+    internal static let EPOLLERR: CUnsignedInt = numericCast(CNIOLinux.EPOLLERR)
+    internal static let EPOLLRDHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLRDHUP)
+    internal static let EPOLLHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLHUP)
+    #if canImport(Android)
+    internal static let EPOLLET: CUnsignedInt = 2_147_483_648  // C macro not imported by ClangImporter
+    #else
+    internal static let EPOLLET: CUnsignedInt = numericCast(CNIOLinux.EPOLLET)
+    #endif
+    #elseif os(Android)
     internal static let EPOLLIN: CUnsignedInt = 1  //numericCast(CNIOLinux.EPOLLIN)
     internal static let EPOLLOUT: CUnsignedInt = 4  //numericCast(CNIOLinux.EPOLLOUT)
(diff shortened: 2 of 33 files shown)


Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -85,9 +87,15 @@ private let sysInet_pton: @convention(c) (CInt, UnsafePointer<CChar>?, UnsafeMut
 #endif
 
 #if os(Android)
+#if canImport(Android)
+let IFF_BROADCAST: CUnsignedInt = numericCast(Android.IFF_BROADCAST.rawValue)
+let IFF_POINTOPOINT: CUnsignedInt = numericCast(Android.IFF_POINTOPOINT.rawValue)
+let IFF_MULTICAST: CUnsignedInt = numericCast(Android.IFF_MULTICAST.rawValue)
+#else
 let IFF_BROADCAST: CUnsignedInt = numericCast(SwiftGlibc.IFF_BROADCAST.rawValue)
```
finagolfin, line 95: Swift 5 has `canImport(Glibc)` and `SwiftGlibc` on Android, while Swift 6 no longer uses those but has `canImport(Bionic)` and `canImport(Android)` instead. Both have `os(Android)`.

To keep this code compiling for both Swift 5 and 6, I duplicated these C declarations. Whenever NIO stops supporting Swift 5, I will submit a pull to rip these out.

Sources/NIOPosix/Linux.swift
```diff
@@ -75,20 +75,24 @@ internal enum Epoll {
...
-    internal static let EPOLLET: CUnsignedInt = 2147483648 //numericCast(CNIOLinux.EPOLLET)
-    #elseif canImport(Musl)
+    #if canImport(Android) || canImport(Musl)
     internal static let EPOLLIN: CUnsignedInt = numericCast(CNIOLinux.EPOLLIN)
     internal static let EPOLLOUT: CUnsignedInt = numericCast(CNIOLinux.EPOLLOUT)
     internal static let EPOLLERR: CUnsignedInt = numericCast(CNIOLinux.EPOLLERR)
     internal static let EPOLLRDHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLRDHUP)
     internal static let EPOLLHUP: CUnsignedInt = numericCast(CNIOLinux.EPOLLHUP)
+    #if canImport(Android)
+    internal static let EPOLLET: CUnsignedInt = 2147483648
```
finagolfin, line 85: Previously, all these constants couldn't be automatically set by the ClangImporter from the C headers, but now [only this one constant couldn't, because it's the only one that's a C macro](https://android.googlesource.com/platform/bionic/+/7f4c837186fc09821cfebf94388a645ead246142/libc/kernel/uapi/linux/eventpoll.h#32).

Sources/NIOCore/BSDSocketAPI.swift
```diff
@@ -85,9 +87,15 @@ private let sysInet_pton: @convention(c) (CInt, UnsafePointer<CChar>?, UnsafeMut
 #endif
 
 #if os(Android)
+#if canImport(Android)
```
FranzBusch, line 90: Shouldn't we just use the same conditionals as above? Instead of checking for the os we just check for `canImport(Glibc)` and `canImport(Android)`?

FranzBusch (commented):
@finagolfin Thanks for the PR. Relatedly I have been wondering if we could setup GH action based Android testing in this repo in the long term?

I would interested to hear what your thoughts are and how we can make developing for Android as easy as installing a cross compilation SDK.

Sources/NIOCore/BSDSocketAPI.swift
finagolfin: The OS check works when building for Android on both Swift 5 and 6 and also includes the Android armv7 definitions on line 100 below. It cannot be replaced because then `canImport(Glibc)` would be true and entered into on linux too, and the `#elseif os(Linux)` on line 103 would be skipped on most linux distros that use glibc.

finagolfin:
Thanks for the quick review, a github action CI workflow for Android on this repo would be great. 😄

I plan to start distributing an Android SDK bundle, which would obsolete [the additional local configuration required for my current Android SDK](https://github.com/finagolfin/swift-android-sdk#cross-compiling-and-testing-swift-packages-with-the-android-sdk), and then look into putting together a custom github build action that any Swift package can use with that new SDK bundle for their own CI, finagolfin/swift-android-sdk#127.

If you would like, I can submit an Android CI workflow for this repo when that SDK bundle is ready. However, my daily Android CI already tests this repo and I quickly upstream most changes needed, so I'm fine with that cadence too. It really just depends if you want to know about Android breaks with every pull, and given that Android is not an officially supported platform, I'm fine with not imposing that additional requirement on NIO pull authors for now.

FranzBusch:
> If you would like, I can submit an Android CI workflow for this repo when that SDK bundle is ready. However, my daily Android CI already tests this repo and I quickly upstream most changes needed, so I'm fine with that cadence too. It really just depends if you want to know about Android breaks with every pull, and given that Android is not an officially supported platform, I'm fine with not imposing that additional requirement on NIO pull authors for now.

In the fullness of time I would like to have Android support in all the repos as well. This would require a few things:
- The cross compilation SDK bundle
- Pitching Android as an officially supported language for Swift.

Once we tackle the latter we might be able to distribute this cross compilation bundle on swift.org which would unlock us testing this in our GitHub Actions.

finagolfin:
> Pitching Android as an officially supported language for Swift

[Already done and I think provisionally agreed to](https://forums.swift.org/t/cross-compilation-sdks-for-android/49101/13) by the core team members I talked to. I will know for sure once I submit it as part of the official CI, which has only been delayed by me.

> Once we tackle the latter we might be able to distribute this cross compilation bundle on swift.org which would unlock us testing this in our GitHub Actions.

Yes, once I'm distributing my own Android SDK bundle and then if and when Android becomes officially supported, distributing an official Android SDK bundle like the current official linux static SDK bundle would be the next step.

We can try setting up an Android CI action for this and other Swift repos then, let's see.

finagolfin:
Rebased and fixed merge conflicts, anything holding up review here?

Sources/NIOCore/BSDSocketAPI.swift
FranzBusch: Can we just guard this code in a compiler directive like `#if compiler(<=5.10)`. Since we always search for these when dropping swift versions

finagolfin: OK, replaced this `canImport(Android)` and three others in this pull with `compiler(>=6.0)` instead. Whenever Swift 5 support is removed, should be easy to keep those declarations and remove these for Swift 5.

Lukasa (approved):

FranzBusch (approved):

FranzBusch:
@finagolfin can you just fix up the formatting then we can merge it

finagolfin:
Will do.

finagolfin:
Ran `swift format` like on the CI and updated this pull.

PR #2786, opened 2024-07-12
Add manual control to NIOLockedValueBox

glbrntt:
Motivation:

NIOLockedValueBox has a 'safer' API than NIOLock as it only provides scoped access to its boxed value. NIOLock requires users to only access protected state while the lock is acquired. As such NIOLockedValueBox should be preferred where possible. However, there are cases where manual control must be used (such as storing a continuation) and users must use a NIOLock for this.

There are two downsides to this:

1. All other access to the protected state must use the NIOLock API putting the onus on the developer to only access the protected state while the lock is held.
2. NIOLock can't store its protected state inline which typically results in users storing it on a class.

Modifications:

- Add an 'unsafe' view to NIOLockedValueBox which allows users to manually control the lock and access its protected state
- Update NIOAsyncWriter and NIOThrowingAsyncSequenceProducer to use NIOLockedValueBox

Result:

- Safer locking API is used in more places
- Fewer allocations

### DIFF
--- a/Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
+++ b/Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
@@ -37,6 +37,43 @@ public struct NIOLockedValueBox<Value> {
     public func withLockedValue<T>(_ mutate: (inout Value) throws -> T) rethrows -> T {
         return try self._storage.withLockedValue(mutate)
     }
+
+    /// Provides an unsafe view over the lock and its value.
+    ///
+    /// This can be beneficial when you require fine grained control over the lock in some
+    /// situations but don't want lose the benefits of ``withLockedValue(_:)`` in others by
+    /// switching to ``NIOLock``.
+    public var unsafe: Unsafe {
+        Unsafe(_storage: self._storage)
+    }
+
+    /// Provides an unsafe view over the lock and its value.
+    public struct Unsafe {
+        @usableFromInline
+        let _storage: LockStorage<Value>
+
+        /// Manually acquire the lock.
+        @inlinable
+        public func lock() {
+            self._storage.lock()
+        }
+
+        /// Manually release the lock.
+        @inlinable
+        public func unlock() {
+            self._storage.unlock()
+        }
+
+        /// Mutate the value, assuming the lock has been acquired manually.
+        @inlinable
+        public func withValueAssumingLockIsAcquired<T>(
+            _ mutate: (inout Value) throws -> T
+        ) rethrows -> T {
+            return try self._storage.withUnsafeMutablePointerToHeader { value in
+                try mutate(&value.pointee)
+            }
+        }
+    }
 }
 
 extension NIOLockedValueBox: Sendable where Value: Sendable {}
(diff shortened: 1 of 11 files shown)


Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
```diff
@@ -37,6 +37,43 @@ public struct NIOLockedValueBox<Value> {
...
+
+        /// Manually release the lock.
+        @inlinable
+        public func unlock() {
+            self._storage.unlock()
+        }
+
+        /// Mutate the value, assuming the lock has been acquired manually.
+        @inlinable
+        public func withValueAssumingLockIsAcquired<T>(
```
Lukasa, line 69: The spelling for this on `OSAllocatedUnfairLock` is `withLockUnchecked`. It might be worth us aligning with that.

While we're here, let's avoid using single-letter generic type parameter names. `ReturnValue` is  probably better.

Sources/NIOCore/AsyncSequences/NIOThrowingAsyncSequenceProducer.swift
```diff
@@ -388,43 +388,64 @@ extension NIOThrowingAsyncSequenceProducer {
...
-        @usableFromInline
-        /* private */ internal var _stateMachine: StateMachine
-        /// The delegate.
+        struct State: Sendable {
+            @usableFromInline
+            var stateMachine: StateMachine
+            @usableFromInline
+            var delegate: Delegate?
+            @usableFromInline
+            var didSuspend: (@Sendable () -> Void)?
```
Lukasa, line 399: This value was never mutated, so I think we can make it `let` and remove it from the lock.

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
glbrntt: Do you think `Unchecked` is also a more appropriate name for the type as well? Wasn't so sure on the naming...

FranzBusch: I think `Unchecked` is better. We normally use `Unsafe` when it causes undefined behaviour which manual locking shouldn't

glbrntt: IIUC `withLockUnchecked` has different semantics: it still acquires the lock but does not check the sendability of the closure passed in.

This is different: it assumes that the caller has already called `lock()` and that they will subsequently call `unlock()`.

glbrntt: re: `Unsafe`, I think depending on what is stored in the locks state, this _could_ be unsafe if you didn't acquire the lock first.

Lukasa: Ok, let's leave the name and fix the generic type parameters.

And FWIW, getting manual locking wrong can absolutely cause UB.

FranzBusch (approved):
LGTM!

PR #2787, opened 2024-07-12
[GHA] Broken symlink and format check

FranzBusch:
# Motivation

The next two reusable checks that we need to create are for broken symlinks and formatting. The latter is disabled for this repo for now since we need to discuss the formatting rules separately.

# Modification

This PR adds two new checks - broken symlinks and formatting.

# Result

Closer to having everything on GitHub actions.

### DIFF
--- a/.github/workflows/reusable_pull_request.yml
+++ b/.github/workflows/reusable_pull_request.yml
@@ -39,6 +39,14 @@ on:
         type: string
         description: "Name of the project called out in the license header."
         required: true
+      broken_symlink_check_enabled:
+        type: boolean
+        description: "Boolean to enable the broken symlink check job. Defaults to true."
+        default: true
+      format_check_enabled:
+        type: boolean
+        description: "Boolean to enable the format check job. Defaults to true."
+        default: true
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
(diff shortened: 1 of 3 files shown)


FranzBusch:
Broken symlink check failing is excepted since the script isn't merged yet. It passes locally

.github/workflows/reusable_pull_request.yml
```diff
@@ -159,3 +167,27 @@ jobs:
...
+    if: ${{ inputs.format_check_enabled }}
+    runs-on: ubuntu-latest
+    container:
+      image: swiftlang/swift:nightly-6.0-jammy
+    timeout-minutes: 5
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run format check
+      run: swift format format --parallel --recursive --in-place
```
glbrntt, line 193: is 'format format' expected?

.github/workflows/reusable_pull_request.yml
```diff
@@ -159,3 +167,27 @@ jobs:
...
+    if: ${{ inputs.broken_symlink_check_enabled }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 1
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run broken symlinks check
+      run: curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/check-broken-symlinks.sh | bash
+
+  formatk-check:
```
glbrntt, line 182:
```suggestion
  format-check:
```

.github/workflows/reusable_pull_request.yml
FranzBusch: Yes it is. `swift-format` now ships with the toolchain and you can reach it by calling `swift format`. The second `format` is the argument to trigger the formatting action. Now that you say this we should probably call `swift format lint` though

glbrntt (approved):

PR #2788, opened 2024-07-12
[GHA] Cxx interoperability compatibility check

FranzBusch:
# Motivation

Another reusable check is to make sure that all library products of a package are successfully building when consumed from a module that has Cxx interoperability enabled.

# Modification

This PR adds a new check to the reusable workflow to check for Cxx interoperability compatibility.

# Result

This should be one of the last reusable workflow checks.

### DIFF
--- a/.github/workflows/reusable_pull_request.yml
+++ b/.github/workflows/reusable_pull_request.yml
@@ -47,6 +47,10 @@ on:
         type: boolean
         description: "Boolean to enable the format check job. Defaults to true."
         default: true
+      cxx_interoperability_check_enabled:
+        type: boolean
+        description: "Boolean to enable the Cxx interoperability check job. Defaults to true."
+        default: true
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_pull_request.yml
```diff
@@ -191,3 +195,18 @@ jobs:
       uses: actions/checkout@v4
     - name: Run format check
       run: swift format lint --parallel --recursive --strict
+
+  cxx-interoperability-check:
```
glbrntt, line 199: minor usability: should we call this `cxx-interop-check`?

.github/workflows/reusable_pull_request.yml
```diff
@@ -191,3 +195,18 @@ jobs:
...
+    name: Cxx interoperability check
+    if: ${{ inputs.cxx_interoperability_check_enabled }}
+    runs-on: ubuntu-latest
+    container:
+      image: swift:5.10-noble
```
glbrntt, line 204: We currently run this on 5.9 and 5.10, what coverage do we want moving forward?

FranzBusch: I just realised that as well. I am currently extracting the matrix based build into a separate workflow so we can easily add more matrix based jobs to it.

PR #2789, opened 2024-07-15
[GHA] Introduce reusable matrix workflow

FranzBusch:
# Motivation

We have a few checks that we want to run across all supported Swift versions e.g. unit tests and benchmarks. Right now we already added two matrix based jobs to the reusable workflow. This leads to a lot of duplication and harder maintenance.

# Modification

This PR splits off the current reusable workflow to only contain soundness related checks and introduces a new reusable matrix based workflow which can be customized to execute different commands.

# Result

It is now easy to setup multiple matrix based workflows.

### DIFF
--- a/.github/workflows/pull_request_soundness.yml
+++ b/.github/workflows/pull_request_soundness.yml
@@ -3,18 +3,6 @@ name: Pull Request
 on:
   workflow_call:
     inputs:
-      unit_tests_linux_enabled:
-        type: boolean
-        description: "Boolean to enable the unit tests linux job. Defaults to true."
-        default: true
-      benchmarks_linux_enabled:
-        type: boolean
-        description: "Boolean to enable the benchmarks linux job. Defaults to true."
-        default: true
-      benchmarks_linux_package_path:
-        type: string
-        description: "Path to the package containing the benchmarks. Defaults to the repository root."
-        default: "."
       api_breakage_check_enabled:
         type: boolean
         description: "Boolean to enable the API breakage check job. Defaults to true."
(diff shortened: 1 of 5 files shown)


FranzBusch:
Benchmarks and Unit tests are expected to fail since the script isn't on main yet. The previous run showed the expected failures of the `nightly-next` and `nightly-main` unit test check due to warnings in the build.

FranzBusch:

FranzBusch:
Our Jenkins CI seems to have some problems today

.github/workflows/pull_request_swift_matrix.yml
```diff
@@ -0,0 +1,84 @@
...
+        type: string
+        description: "The name of the workflow used for the concurrency group."
+        required: true
+      matrix_linux_command:
+        type: string
+        description: "The command of the current Swift version linux matrix job to execute."
+        required: true
+      matrix_linux_5_8_enabled:
+        type: string
+        description: "The command of the 5.8 Swift version linux matrix job to execute."
```
glbrntt, line 16: This and all the other `_enabled`s don't seem right, they should surely be booleans?

.github/workflows/pull_request_swift_matrix.yml
```diff
@@ -0,0 +1,84 @@
...
+      matrix:
+        swift:
+        - image: swift:5.8-jammy
+          swift_version: "5.8"
+        - image: swift:5.9-jammy
+          swift_version: "5.9"
+        - image: swift:5.10-jammy
+          swift_version: "5.10"
+        - image: swiftlang/swift:nightly-6.0-jammy
+          swift_version: "nightly-next"
```
glbrntt, line 65: Should the version be more specific? e.g. "nightly-6.0"?

.github/workflows/pull_request_swift_matrix.yml
glbrntt: actually, are at the `_enabled`s used at all?

glbrntt (changes requested):

.github/workflows/pull_request_soundness.yml
```diff
@@ -50,64 +38,10 @@ on:
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-soundness
```
MahdiBM, line 41: I ... don't think this addition is needed? Did you run into a problem and this solved it?

.github/workflows/pull_request_swift_matrix.yml
```diff
@@ -0,0 +1,84 @@
...
+        description: "The command of the nightly next Swift version linux matrix job to execute."
+      matrix_linux_nightly_next_command_override:
+        type: string
+        description: "The command of the nightly next Swift version linux matrix job to execute."
+      matrix_linux_nightly_main_enabled:
+        type: string
+        description: "The command of the nightly main Swift version linux matrix job to execute."
+      matrix_linux_nightly_main_command_override:
+        type: string
+        description: "The command of the nightly main Swift version linux matrix job to execute."
```
MahdiBM, line 43: Seems redundant. You could accept key-value pairs in a JSON.
```
fromJson(inputs.matrix_linux_command_override).510
```
assuming we have a 
```
matrix_linux_command_override:
  type: string
```
I haven't tried but it think using brackets should also work:
```
fromJson(inputs.matrix_linux_command_override)[matrix.swift_version]
```

.github/workflows/pull_request_swift_matrix.yml
FranzBusch: The next version number is often decided quite close to the release so I wanted to keep it general with `next`

.github/workflows/pull_request_swift_matrix.yml
FranzBusch: Thanks. Good catch!

.github/workflows/pull_request_swift_matrix.yml
FranzBusch: As an example the next release was called swift `5.11` for a long time until it got renamed to swift 6.

FranzBusch: I just changed it to `nightly-6.0` now.

glbrntt: But there was never a 5.11 branch or docker image, it only appeared in `swiftc --version`. At this point, where we're selecting a docker image we _must_ know the version (apart from `main` builds)

glbrntt (approved):

.github/workflows/pull_request_swift_matrix.yml
MahdiBM: hmm but do you really _want_ to be backward incompatible? I think that's not a great idea to be honest.
Imagine doing a PR and noticing that CI is failing due to unrelated reasons.
I guess the maintainers can be advised to look out for these kinds of changes: "if CI failed and was complaining about an invalid value in the inputs, that's because we've probably removed it". But is that a good idea?

Lukasa:
Merging over flaky benchmark

PR #2791, opened 2024-07-17
ChannelHandler: provide static (un)wrap(In|Out)bound(In|Out)

weissi:
### Motivation:

Since version 1.0.0, SwiftNIO has provided the 'type-assisted' ChannelHandler concept with the `self.(un)wrap(In|Out)bound(In|Out)` helpers. Up until we got `Self` to work in all situations (I believe Swift 5.3 or so), this was the best syntax.

Unfortunately, this syntax does come at a cost:
- If `self` is only retained to wrap/unwrap something in an `@escaping` closure (such as a future callback), there is needless ARC
- If `self` is the only thing and only retained to wrap/unwrap there might even be an unnecessary allocation
- Under strict concurrency, capturing `self: ChannelHandler` actually leads to the user needlessly having to `NIOLoopBound(self, eventLoop: context.eventLoop)` or similar 

### Modifications:

Provide static versions of `(un)wrap(In|Out)bound(In|Out)` in `ChannelHandler`s such that a user can write for example `Self.unwrapInboundOut(foo)`.

(Mostly done automatically using `git grep 'self\.\(un\)\?wrap\(In\|Out\)bound\(In\|Out\)' | cut -d: -f1 | sort -u | while read -r line; do sed -ri '' 's/self\.(un)?wrap(In|Out)bound(In|Out)/Self.\1wrap\2bound\3/g' "$line"; done`)

### Result:

- Less ARC traffic
- Fewer allocations
- Strict Concurrency becomes easier
- Easier job for the optimiser

### DIFF
--- a/Sources/NIOCore/TypeAssistedChannelHandler.swift
+++ b/Sources/NIOCore/TypeAssistedChannelHandler.swift
@@ -30,6 +30,11 @@ extension _EmittingChannelHandler {
     public func wrapOutboundOut(_ value: OutboundOut) -> NIOAny {
         return NIOAny(value)
     }
+
+    @inlinable
+    public static func wrapOutboundOut(_ value: OutboundOut) -> NIOAny {
+        return NIOAny(value)
+    }
 }
 
 ///  `ChannelHandler` which handles inbound I/O events for a `Channel`.
@@ -62,6 +67,16 @@ extension ChannelInboundHandler {
     public func wrapInboundOut(_ value: InboundOut) -> NIOAny {
         return NIOAny(value)
     }
+
+    @inlinable
+    public static func unwrapInboundIn(_ value: NIOAny) -> InboundIn {
+        return value.forceAs()
+    }
+
+    @inlinable
+    public static func wrapInboundOut(_ value: InboundOut) -> NIOAny {
+        return NIOAny(value)
+    }
 }
 
 /// `ChannelHandler` which handles outbound I/O events or intercept an outbound I/O operation for a `Channel`.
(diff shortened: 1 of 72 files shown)


Sources/NIOCore/TypeAssistedChannelHandler.swift
```diff
@@ -30,6 +30,11 @@ extension _EmittingChannelHandler {
     public func wrapOutboundOut(_ value: OutboundOut) -> NIOAny {
         return NIOAny(value)
     }
+
+    @inlinable
+    public static func wrapOutboundOut(_ value: OutboundOut) -> NIOAny {
```
Lukasa, line 35: Can we think of a reason not to deprecate the non-static versions?

Lukasa:

Sources/NIOCore/TypeAssistedChannelHandler.swift
weissi: I think it would be annoying to users having to change all of those :|.

What we should definitely remove though is the customisation point on those. There is only one correct implementation so I'd argue we could even remove those without a SemVer major break.

Lukasa:
Can't say I'm surprised that that's brokne.

Sources/NIOCore/TypeAssistedChannelHandler.swift
Lukasa: While we can, I don't think it helps us much, and will cause mayhem for those few users who are using library evolution mode even though we told them not to.

It's definitely annoying to users, but otherwise the odds that anyone will get the perf win is pretty low. `self.` is going to be muscle memory for most, and already written for the rest.

weissi: Fair

weissi:

FranzBusch (approved):
The code change looks good to me! We could potentially write an ARC benchmark for this but our experience with ARC benchmarks were pretty flaky so far.

weissi:
> The code change looks good to me! We could potentially write an ARC benchmark for this but our experience with ARC benchmarks were pretty flaky so far.

Indeed. Also, the risk here is 0. We all know that `Self.something` is better or equal to `self.something`, in any place.

PR #2793, opened 2024-07-18
Disable warnings as errors on Swift 6 and main

glbrntt:
Motivation:

The Swift 6 compiler emits a number of warnings about sendability which we haven't dealt with yet. The CI failing on warnings may mask other compilation issues or test failures.

Modifications:

- Disable warnings as errors on Swift 6 and main

Result:

Swift 6 and main CI should pass


FranzBusch (approved):

FranzBusch (approved):

.github/workflows/pull_request.yml
```diff
@@ -2,8 +2,8 @@ name: Pull Request
 
 on:
   pull_request:
-    types: [opened, reopened, synchronize, ready_for_review]
```
MahdiBM, line 5: This change will prevent running tests when a PR changes status from `draft` to `ready_for_review` I think.
Unless `synchronize` already covers that, which I don't think so.

Currently the tests won't be run if the PR is a `draft`.

glbrntt: @MahdiBM they do run for draft PRs: https://github.com/apple/swift-nio/pull/2798

MahdiBM: hmmm that was not my experience with Vapor's CI which had this exact same setting.
I might have gotten something wrong then.
Thanks for confirming.

glbrntt: I _think_ "opened" covers it. Draft / Ready for Review seems orthogonal to whether the PR is open or closed.

MahdiBM: Maybe then the case where it won't work is when you open the PR as draft, then when the CI is over, you add a commit, then after that commit, you mark the PR as `ready_for_review`.
_maybe_, not 100% sure.

PR #2794, opened 2024-07-18
Adopt swift-format

FranzBusch:
### Motivation:

We have been discussing the adoption of a formatting tool in our repositories for a long time and it turns out the time has come now with the addition of our new GH action pipelines we also want to setup formatting for our repos.

### Modifications:

This PR configures `swift-format` very similar to how `swift-certificates` and `swift-openapi-generator` configured there's with only minimal differences. The goal is to land on one `swift-format` config that we can then use in all repos. The differences to the existing configurations are as follows:

`NoLabelsInCasePatterns=true` OpenAPI: false; Certs: false
`OmitExplicitReturns=true` OpenAPI: true, Certs: false
`OrderedImports=true` OpenAPI: false, Certs: false
`ReplaceForEachWithForLoop=true` OpenAPI: true, Certs: false
`UseEarlyExits=false` OpenAPI: false, Certs: true

These options are also different but only for the reason that they would cause a lot of time to adopt. New projects should start with them and we should enable them in the fullness of time.
`ValidateDocumentationComments`
`AllPublicDeclarationsHaveDocumentation`

### Result:

We have a formatter configured and check for it in our CI.


Lukasa (approved):

czechboy0 (approved):

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -36,7 +36,7 @@ extension ByteBuffer {
             // this is not technically correct because we shouldn't just bind
             // the memory to `UInt8` but it's not a real issue either and we
             // need to work around https://bugs.swift.org/browse/SR-9604
-            Array<UInt8>(UnsafeRawBufferPointer(fastRebase: ptr[range]).bindMemory(to: UInt8.self))
+            [UInt8](UnsafeRawBufferPointer(fastRebase: ptr[range]).bindMemory(to: UInt8.self))
```
weissi, line 39: @FranzBusch this is against the Swift style guides

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -21,8 +21,8 @@ public struct ByteBufferView: RandomAccessCollection, Sendable {
     public typealias Index = Int
     public typealias SubSequence = ByteBufferView
 
-    /* private but usableFromInline */ @usableFromInline var _buffer: ByteBuffer
```
weissi, line 24: @FranzBusch why are we losing all these comments

Sources/NIOCore/ByteBuffer-views.swift
```diff
@@ -34,37 +34,41 @@ public struct ByteBufferView: RandomAccessCollection, Sendable {
...
 
     @inlinable
     public var endIndex: Index {
-        return self._range.upperBound
+        self._range.upperBound
     }
 
     @inlinable
     public func index(after i: Index) -> Index {
-        return i + 1
```
weissi, line 60: why are we losing the `return`s? That makes debugging so much harder because now you can't just add a `print` above :(

Sources/NIOCore/ByteBuffer-aux.swift
FranzBusch: Which style guide?

Sources/NIOCore/ByteBuffer-views.swift
FranzBusch: That was on me and I can bring those back. We have a formatting rule that disallows block comments. I removed some of those `private but...` comments since we are super inconsistent with them throughout the project and all the properties are `_`. We have been consistent with the `_` but not so much with the `private but...` comment.

PR #2797, opened 2024-07-19
[GHA] Only format Swift files that are in Git index

simonjbeaumont:
### Motivation:

Sometimes you want to be able to run the same commands used in the Github action locally, e.g. before pushing. The format Github action has a couple of issues that make running locally problematic. The most important of these is the fact that it uses `--recursive .` which formats _everything_, including ignored files.

### Modifications:

- Only format `.swift` files in the Git index.
- Don't pollute global Git config.
- Only update local Git config if running in CI.

### Result:

- Commands are better suited to be used locally.

### DIFF
--- a/.github/workflows/pull_request_soundness.yml
+++ b/.github/workflows/pull_request_soundness.yml
@@ -126,9 +126,9 @@ jobs:
     - name: Mark the workspace as safe
     # https://github.com/actions/checkout/issues/766
       run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
-    - name: Run format lint check
-      run:  swift format lint --strict --recursive --parallel .
-    - name: Run format and check for modified files
-      run: |
-        swift format format --parallel --recursive --in-place .
-        git diff-index --quiet HEAD
+    - name: Format Swift files
+      run: git ls-files -z '*.swift' | xargs -0 swift format format --parallel --in-place
+    - name: Lint Swift files
+      run:  git ls-files -z '*.swift' | xargs -0 swift format lint --strict --parallel
+    - name: Check format produced no diff
+      run: GIT_PAGER= git diff --exit-code '*.swift'


.github/workflows/pull_request_soundness.yml
```diff
@@ -125,10 +125,10 @@ jobs:
       uses: actions/checkout@v4
     - name: Mark the workspace as safe
     # https://github.com/actions/checkout/issues/766
-      run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
+      run: if [ -n "${GITHUB_WORKSPACE:-}" ]; then git config --global --add safe.directory ${GITHUB_WORKSPACE}; fi
```
FranzBusch, line 128: Can we put the if check into a separate job `if:` condition instead of into the script

simonjbeaumont: `if:` is only on the `job` not the `steps`. Are you hung up on this one, if so I'll just remove this and update my `yq` hack to filter it out (which I'd rather not, but don't want to block progress).

FranzBusch: That's not correct. `steps` support `if` conditions https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

FranzBusch: I am happy to add this but can we also add it to the API breaking job that does the same

FranzBusch (approved):

.github/workflows/pull_request_soundness.yml
simonjbeaumont: You were right about `steps[*]` supporting `if`, sorry about that.

Done, and updated the API check job to do the same, as requested.

simonjbeaumont: Hm, I used the exact syntax from the example docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets.

```yaml
    steps:
      - if: ${{ env.super_secret != '' }}
        run: echo 'This step will only run if the secret has a value set.'
      - if: ${{ env.super_secret == '' }}
        run: echo 'This step will only run if the secret does not have a value set.'
```

But it's failing.

.github/workflows/pull_request_soundness.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         fetch-depth: 0
     - name: Mark the workspace as safe
       # https://github.com/actions/checkout/issues/766
+      if: ${{ env.GITHUB_WORKSPACE != '' }}
```
FranzBusch, line 60: I think we might have to make this `act` specific. They recommend doing this

```
if: ${{ !github.event.act }}
```

.github/workflows/pull_request_soundness.yml
simonjbeaumont: OK, I ditched this and also fixed the diff command while there to exit non-zero with changes and actually show what they are.

.github/workflows/pull_request_soundness.yml
simonjbeaumont: We can leave this how it is now, I'm happy to run in `act` and not try to scrape the commands to run locally.

FranzBusch (approved):

PR #2801, opened 2024-07-22
[GHA] Make unit tests, benchmarks and cxx interop easier to reuse

FranzBusch:
# Motivation

Currently the `pull_request_swift_matrix` workflow provided an easy way to run a command across many Swift versions. We do have a few common commands that we want to run across the repos. Namely, unit tests, benchmarks and cxx interop.

# Modification

This PR creates new workflows for the above common matrix commands so it's easier to reuse them from other repos.

# Result

Even easier to share these workflows.


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,38 @@
...
+            description: "Boolean to enable the Linux nightly main Swift version matrix job. Defaults to true."
+            default: true
+ 
+jobs:
+    benchmarks:
+        name: Benchmarks
+        uses: ./.github/workflows/pull_request_swift_matrix.yml
+        with:
+          name: "Benchmarks"
+          matrix_linux_command: "apt-get update -y -q && apt-get install -y -q libjemalloc-dev && swift package --package-path Benchmarks/ --disable-sandbox benchmark baseline check --check-absolute-path Benchmarks/Thresholds/${SWIFT_VERSION}/"
```
simonjbeaumont, line 33: Should we make the thresholds root directory an input with default?

simonjbeaumont: Is it OK to have `--disable-sandbox` be default?

.github/workflows/cxx_interop.yml
```diff
@@ -0,0 +1,38 @@
...
+            description: "Boolean to enable the Linux nightly 6.0 Swift version matrix job. Defaults to true."
+            default: true
+        linux_nightly_main_enabled:
+            type: boolean
+            description: "Boolean to enable the Linux nightly main Swift version matrix job. Defaults to true."
+            default: true
+ 
+jobs:
+    cxx-interop:
+        name: Cxx interop
```
simonjbeaumont, line 29: Nit: Any reason to not spell this as "C++" when we have a string label?

.github/workflows/cxx_interop.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: ./.github/workflows/pull_request_swift_matrix.yml
+        with:
+          name: "Cxx interop"
+          matrix_linux_command: "apt-get update -y -q && apt-get install -y -q jq && ./scripts/check-cxx-interop-compatibility.sh"
```
simonjbeaumont, line 33: Does this rely on the adopter having this script in their repo? If so, should we make this pipeline opt-in rather than opt-out?

.github/workflows/pull_request.yml
```diff
@@ -1,43 +1,34 @@
...
+        uses: ./.github/workflows/cxx_interop.yml
 
-  call-pull-request-integration-tests-workflow:
-    name: Integration tests
-    uses: ./.github/workflows/pull_request_swift_matrix.yml
-    with:
-      name: "Integration tests"
-      matrix_linux_command: "apt-get update -y -q && apt-get install -y -q lsof dnsutils netcat-openbsd net-tools curl jq && ./scripts/integration_tests.sh"
+    integration-tests:
+        name: Checks
```
simonjbeaumont, line 30: A bunch of these all share the same `name: Checks`, what is the result of that?

.github/workflows/pull_request.yml
```diff
@@ -1,43 +1,34 @@
...
+        uses: ./.github/workflows/pull_request_swift_matrix.yml
+        with:
+          name: "Integration tests"
+          matrix_linux_command: "apt-get update -y -q && apt-get install -y -q lsof dnsutils netcat-openbsd net-tools curl jq && ./scripts/integration_tests.sh"
```
simonjbeaumont, line 34: Nice. Is this an example of how a project could use the matrix for an arbitrary job that they can name and provide a command for, and have that job be run for all the Swift versions?

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,63 @@
...
+            description: "The arguments passed to swift test in the Linux nightly main Swift version matrix job."
+            default: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
+ 
+jobs:
+    unit-tests:
+        name: Unit tests
+        uses: ./.github/workflows/pull_request_swift_matrix.yml
+        with:
+          name: "Unit tests"
+          matrix_linux_command: "swift test -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
```
simonjbeaumont, line 53: IIUC these flags shouldn't be here, since they are provided in the override in the other YAML file?

Wondering if the most acceptable thing to do would be for the default to command to just be `swift test` and allow folks to provide their override when they adopt, opting into whatever additional level of validation they want for their project.

.github/workflows/pull_request.yml
FranzBusch: That is intentional to get the naming of the checks somewhat aligned. Otherwise it would be `Pull Request / Benchmarks / Benchmarks / 5.9`

.github/workflows/pull_request.yml
FranzBusch: Yes exactly. This still allows individual project to configure their own jobs.

.github/workflows/unit_tests.yml
FranzBusch: We need to pick a default here since this input is required. In practice this input will never actually be used because we provide an override for each version specific job. I do think we should have strong defaults though. In particular those two seem like net wins and I would rather people have to disable warnings as errors than the other way around.

.github/workflows/cxx_interop.yml
FranzBusch: Kinda stayed consistent with the script name and the flag that's set in the `Package.swift`. Also AFAIK all of C++/Cpp/Cxx is valid to use.

.github/workflows/benchmarks.yml
FranzBusch: Removed the `--disable-sandbox` we don't need it on the CI.

.github/workflows/cxx_interop.yml
simonjbeaumont: Sure, they're all valid and I'm fine with keeping it. I always assumed Cxx and Cpp were there for when we couldn't use non-ascii characters, like in command line flags and file extensions.

.github/workflows/pull_request.yml
simonjbeaumont: OK, fine.

.github/workflows/unit_tests.yml
simonjbeaumont: If you feel super strongly about it, then fine, but I would have had a slight lean the other way...

I think if we're offering this as a reusable building block it shouldn't be so opinionated so would prefer it for this to be free of anything that isn't the default, i.e. just `swift test`. I think that will be the least surprising behaviour. It would be annoying for folks to adopt the pipeline when their project currently builds fine (without additional strict flags, and then scratch their heads a little).

simonjbeaumont (approved):
Looks good. Still not sure I like the defaults but if you really want them, then OK.

.github/workflows/unit_tests.yml
FranzBusch: You convinced me and I removed the default args now.

PR #2804, opened 2024-07-24
Fix flaky test: testInstantTCPConnectionResetThrowsError

simonjbeaumont:
### Motivation:

Running this test in a loop will fail. It's understandable because the test itself is trying to trigger a race condition. We wouldn't see this on our PR CI because it only runs on Darwin. 

### Modifications:

Extend the test server used in the test with another promise that allows the test to wait for the channel to become active before continuing with the rest of the test.

### Result:

Can now run this test repeatedly without failure.


simonjbeaumont:
Actually, hold up. This didn't actually make it go away, just a lot more rare. Used to fail ~1:100, now it's ~1:3000.

PR #2806, opened 2024-07-24
FileSystem.copyItem can parallelise directory copy

UncleMattHope:
FileSystem.copyItem defaults to perform directory copies in parallel

### Motivation:

Using some limited parallelism can improved the speed of the copy for very limited increase in cpu use.

In circumstances where the source and/or destination are on a high latency device, such as a network file system,
the resulting speed up can be even more significant.
Providing conservative defaults, while allowing users the ability to control it directly if desired, provides for a faster default and access to user targeted gains in performance. 
The use of parallel by default is considered in keeping with the swift-nio model of concurrent capabilities. 

The parallelism used is restricted because the underlying limited resource for this is highly likely to be limited system file descriptors, rather than the limits of the underlying executor itself, especially on embedded devices with significantly more limited concurrent descriptors.

### Modifications:

All copyItem overloads except ```copyItem(sourcePath:destinationPath:shouldProceedAfterError:shouldCopyFile)```
can, and likely will, use concurrent copies when copying directories.

A Newly type ```CopyStrategy``` controls the behaviour, allowing:
* ```parallel(maxDescriptors: Int)```
   * Parallel limited to a maximum number of concurrently open file descriptors.
* ```sequential```
   * The previous behaviour, still asynchronous but only one operation at a time
* ```platformDefault``` 
    * Whatever the API determines the default should be from the prior options

The cancellation semantics of copyItem were previously unclear/undocumented. 
These are now stated and cancellation is tested.

There were some aspects of naming/types on this API which were inconsistent/unclear. Those have been altered to aid clarity.
* Less duplication with the protocol docs
* Consistent use of `item` not `file`
* Use DirectoryEntry for the source in both ```shouldXxx``` callbacks

NIOFileSystem package status remains unchanged, it is still _NIOFileSystem with the semantics that entails.

### Result:

For any use of the following functions on FileSystemProtocol, when backed by FileSystem:

#### ```copyItem(*)```:

The shouldCopyItem check happens before the internal FileSystem "do I know how to copy this sort of FileType" check. This makes the code a little simpler, but the desire is to allow consuming code to filter out anything unexpected if they like without needing to trap errors.

#### ```copyItem(sourcePath:destinationPath)```:

Depending on the platform this will become parallel, users on MacOS, Linux and Windows will have up to 4 concurrent file system operations at once.
Empirical testing on a MacBook pro with a significant dense directory hierarchy of 12 GiB had performance improvements of ~50% reduction in elapsed time for CPU usage increases below the noise floor of the testing

#### ```copyItem(sourcePath:destinationPath:shouldProceedAfterError:shouldCopyFile)```:
When migrating from this to ```copyItem(sourcePath:destinationPath:copyStrategy:shouldProceedAfterError:shouldCopyItem)```care should be taken to ensure the should callbacks are safe to use in different concurrency domains, or that CopyStrategy.sequential is explicitly stated.
The backward compatibility shim provided does this by default.


Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -917,72 +923,228 @@ extension FileSystem {
...
+        ) async throws -> Void,
+        shouldCopyItem: @escaping @Sendable (
+            _ source: DirectoryEntry,
+            _ destination: FilePath
+        ) async -> Bool
+    ) async throws {
+        // Strategy: find all needed items to copy/recurse into while the directory is open;
+        // defer actual copying and recursion until after the source directory has been closed
+        // to avoid consuming too many file descriptors.
+        let toCopy = try await self.prepareDirectoryForRecusiveCopy(
```
glbrntt, line 988: nit: can we call this `itemsToCopy` or something to avoid shadowing the `toCopy` enum case?

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -917,72 +923,228 @@ extension FileSystem {
...
+            _ error: Error
+        ) async throws -> Void,
+        shouldCopyItem: @escaping @Sendable (
+            _ source: DirectoryEntry,
+            _ destination: FilePath
+        ) async -> Bool
+    ) async throws {
+        switch copyStrategy.wrapped {
+        case .sequential:
+            return try await copyDirectorySequential(
```
glbrntt, line 1066: nit: explicit self

```suggestion
            return try await self.copyDirectorySequential(
```

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -176,24 +176,69 @@ public protocol FileSystemProtocol: Sendable {
...
-    ///       copying files if an error is encountered during the operation.
-    ///   - shouldCopyFile: A closure which is executed before each copy to determine whether each
-    ///       file should be copied.
+    ///       copying files if an error is encountered during the operation. See Errors section for full details.
+    ///   - shouldCopyItem: A closure which is executed before each copy to determine whether each
+    ///       item should be copied. See Filtering section for full details
+    ///
+    /// #### Errors
+    ///
+    /// No errors should be throw by implementors without first calling ``shouldProceedAfterError``,
```
glbrntt, line 193:
```suggestion
    /// No errors should be thrown by implementors without first calling ``shouldProceedAfterError``,
```

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -176,24 +176,69 @@ public protocol FileSystemProtocol: Sendable {
...
+    /// if that returns without throwing this is taken as permission to continue and the error is swallowed.
+    /// If instead the closure throws then ``copyItem`` will throw and copying will stop, though the
+    /// precise semantics of this can depend on the ``strategy``.
+    ///
+    /// if using ``CopyStrategy/parallel``
+    /// Already started work may continue for an indefinite period of time. In particular, after throwing an error
```
glbrntt, line 199:
```suggestion
    /// If using ``CopyStrategy/parallel`` already started work may continue for an indefinite period of time. In particular, after throwing an error
```

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -176,24 +176,69 @@ public protocol FileSystemProtocol: Sendable {
...
+    /// it is possible that invocations of ``shouldCopyItem`` may continue to occur!
+    ///
+    /// If using ``CopyStrategy/sequential`` only one invocation of any of the `should*` closures will occur at a time,
+    /// and an error will immediately stop further activity.
+    ///
+    /// The specific error thrown from copyItem is undefined, it does not have to be the same error thrown from
+    /// ``shouldProceedAfterError``.
+    /// In the event of any errors (ignored or otherwise) implementations are under no obbligation to
+    /// attempt to 'tidy up' after themselves. The state of the file system within ``destinationPath``
+    /// after an aborted copy should is undefined
```
glbrntt, line 209:
```suggestion
    /// after an aborted copy should is undefined.
```

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -417,22 +462,80 @@ extension FileSystemProtocol {
     ///   exists prior to the copy or its parent directory does not exist.
     ///
     /// Note that other errors may also be thrown. If any error is encountered during the copy
-    /// then the copy is aborted. You can modify the behaviour with the `shouldProceedAfterError`
-    /// parameter of ``copyItem(at:to:shouldProceedAfterError:shouldCopyFile:)``.
+    /// then the copy is aborted. You can modify the behaviour with the `shouldProceedAfterError`cop
```
glbrntt, line 465:
```suggestion
    /// then the copy is aborted. You can modify the behaviour with the `shouldProceedAfterError`
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,158 @@
...
+        ) async throws -> Void,
+        shouldCopyItem: @escaping @Sendable (
+            _ source: DirectoryEntry,
+            _ destination: FilePath
+        ) async -> Bool
+    ) async throws {
+        // Implemented with NIOAsyncSequenceProducer rather than AsyncStream.
+        // It is approximately the same speed in the best case but has significantly less variance.
+        // NIOAsyncSequenceProducer also enforces a multi producer single consumer access pattern.
+        let sequence = NIOAsyncSequenceProducer<DirCopyItem, NoBackPressureStrategy, DirCopyDelegate>.makeSequence(
```
glbrntt, line 47: Is this effectively a sequence of items to copy? If so can we give it a better name

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,158 @@
...
+            backPressureStrategy: .init(),
+            finishOnDeinit: false,
+            delegate: .init()
+        )
```
glbrntt, line 51: Can we re-spell this slightly for readability?

```suggestion
        let sequence = NIOAsyncSequenceProducer.makeSequence(
            elementType: DirCopyItem.self,
            backPressureStrategy: NoBackPressureStrategy(),
            finishOnDeinit: false,
            delegate: DirCopyDelegate()
        )
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+}
+
+/// An 'always ask for more' no  back-pressure strategy for a ``NIOAsyncSequenceProducer``.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+private struct NoBackPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategy {
+    public mutating func didYield(bufferDepth: Int) -> Bool { true }
+
+    public mutating func didConsume(bufferDepth: Int) -> Bool { true }
```
glbrntt, line 146: nit: `public` is redundant and misleading here

```suggestion
    mutating func didYield(bufferDepth: Int) -> Bool { true }

    mutating func didConsume(bufferDepth: Int) -> Bool { true }
```

Sources/NIOFileSystem/CopyStrategy.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    /// The minimum possible parallelism supported by ``parallel(maxDescriptors:)``
+    /// Using this in your non testing code paths would be an anti pattern, use ``sequential``instead.
+    ///
+    /// Less than this and you can't do a copy (without pulling an entire file into memory, which is not generally possible, or desirable)
+    /// This is is not quite the same as sequential, different code paths are used.
+    ///
+    /// This is primarily exposed for testing to ensure use of the parallel paths (which are more complex) while keeping actual
+    /// parallelism to minimal levels to make debugging simpler.
+    public static let minimalParallel: Self = .init(.parallel(minDescriptorsAllowed))
```
glbrntt, line 94: Is this useful outside of our own tests?

glbrntt (changes requested):

Sources/NIOFileSystem/CopyStrategy.swift
UncleMattHope: I found writing tests using ```.parallel(maxDescriptors:2)``` to be painful since that call declares throws, necessitating lots of cascading the throwing capability or wrapping it in an ugly try/catch.

It seemed reasonable that other people testing code using this might want to make their tests follow the parallel path via ```.platformDefault```, but have the option to switch to the limited parallelism easily without having to mess about at the CopyStrategy declaration point in their tests.

Happy to hide it and just push it into our tests if you prefer.

glbrntt: I don't feel too strongly here but my lean is to hide it from the public API, as you note using it outside of testing is an anti-pattern, and I think it's unlikely that users will also need this in their tests. If they do they can `try!` `parallel(maxDescriptors: 2)` (which should be safe per our docs).

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -917,72 +922,228 @@ extension FileSystem {
...
-            if await shouldCopyFile(entry.from, entry.to) {
-                try await self.copyDirectory(
-                    from: entry.from,
-                    to: entry.to,
-                    shouldProceedAfterError: shouldProceedAfterError,
-                    shouldCopyFile: shouldCopyFile
+    /// Building block of the parallel directory copy implementation
+    /// Each invovation of this is allowed to consume two file descriptors,
+    /// any further work (if any) should be sent to `yield` for future processing
+    @Sendable func copySelfAndEnqueueChildren(
```
FranzBusch, line 1092: We should never have to mark methods as `@Sendable`. The compiler should be able to infer that 
```suggestion
    func copySelfAndEnqueueChildren(
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,159 @@
...
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+private struct NoBackPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategy {
+    mutating func didYield(bufferDepth: Int) -> Bool { true }
+
+    mutating func didConsume(bufferDepth: Int) -> Bool { true }
+}
+
+/// We ignore back pressure, the inherent handle limiting in copyDirectoryParallel means it is unnecessary.
+@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+private final class DirCopyDelegate: NIOAsyncSequenceProducerDelegate, @unchecked Sendable {
```
FranzBusch, line 152:
```suggestion
private final struct DirCopyDelegate: NIOAsyncSequenceProducerDelegate, Sendable {
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,159 @@
...
+        let copyRequiredQueue = NIOAsyncSequenceProducer.makeSequence(
+            elementType: DirCopyItem.self,
+            backPressureStrategy: NoBackPressureStrategy(),
+            finishOnDeinit: false,
+            delegate: DirCopyDelegate()
+        )
+
+        // We ignore the result of yield in all cases because we are not implementing back pressure
+        // and cancellation is dealt with separately.
+        @Sendable func yield(_ contentsOf: [DirCopyItem]) {
```
FranzBusch, line 56:
```suggestion
         func yield(_ contentsOf: [DirCopyItem]) {
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,159 @@
...
+            _ = copyRequiredQueue.source.yield(contentsOf: contentsOf)
+        }
+
+        // Kick start the procees by enqueuing the root entry,
+        // the calling function already validated the root needed copying.
+        _ = copyRequiredQueue.source.yield(.toCopy(from: .init(path: sourcePath, type: .directory)!, to: destinationPath))
+
+        // The processing of the very first item (the root) will increment this,
+        // after than when it hits zero we've finished.
```
FranzBusch, line 65:
```suggestion
        // after then when it hits zero we've finished.
```

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            let iter = copyRequiredQueue.sequence.makeAsyncIterator()
+
+            // inProgress counts the number of tasks we have added to the task group
+            // Get up to the maximum concurrency first.
+            // We haven't started monitoring for task completion, so inProgress is 'worst case'.
+            var inProgress = 0
+            while inProgress <= maxConcurrentOperations {
+                try Task.checkCancellation()
```
FranzBusch, line 112: We are calling `iter.next()` on the next line which will check for cancellation when needed

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
```diff
@@ -0,0 +1,159 @@
...
+                    return
+                }
+            }
+
+            // Then operate one in (finish) -> one out (start another),
+            // but only for items that trigger a task.
+            while let _ = try await taskGroup.next() {
+                var keepConsuming = true
+                while keepConsuming {
+                    try Task.checkCancellation()
```
FranzBusch, line 129: Same here

Sources/NIOFileSystem/CopyStrategy.swift
UncleMattHope: cool, I'll move it to the testing

Sources/NIOFileSystem/FileSystem.swift
UncleMattHope: Ah yes, this was previously a closure and I failed to remove that when I pulled it out to an instance method thanks

Sources/NIOFileSystem/Internal/ParallelDirCopy.swift
UncleMattHope: I was attempting to remove a warning in the swift 6 language mode here. Should I not do that yet?

```Concurrently-executed local function 'yield' must be marked as '@Sendable'; this is an error in the Swift 6 language mode```
and at callsite:
```Converting non-sendable function value to '@Sendable ([FileSystem.DirCopyItem]) -> Void' may introduce data races```

Comment (UncleMattHope) [reply Sources/NIOFileSystem/Internal/ParallelDirCopy.swift:56]: There's other places we do this BTW, so for now I've left it in in case you thought this was a instance function, not a closure.

Comment (FranzBusch) [reply Sources/NIOFileSystem/Internal/ParallelDirCopy.swift:56]: Now that I think about this again I think this is safe todo in the Swift 6 compiler since that actually verified that `@Sendable` is correct. Previous compilers would just blindly trust you.

**Review by glbrntt — APPROVED** [review]
This LGTM, thanks @UncleMattHope!

File: Sources/NIOFileSystem/Internal/ParallelDirCopy.swift [hunk 17]
```diff
@@ -0,0 +1,159 @@
...
+            // We haven't started monitoring for task completion, so inProgress is 'worst case'.
+            var inProgress = 0
+            while inProgress <= maxConcurrentOperations {
+                let item = await iter.next()
+                if let item = item {
+                    if onNextItem(item) {
+                        inProgress += 1
+                    }
+                } else {
+                    // We completed things before we hit the limit
```
UncleMattHope, line 120: This was a bug, that made my cancellation test flaky if it happened to trigger during this phase.
I have a fix for that and will push shortly.

The cancellation propagates as the sequence cleanly terminating, so our task just completes early, rather than resulting an a cancellation.
simply inserting a ```try Task.checkCancellation()``` here is sufficient, but I have also streamlined the test side to be faster too

glbrntt (approved):

UncleMattHope:
Confirmation of acceptance of API break in _NIOFileSystem

>  5 breaking changes detected in _NIOFileSystem:
>   💔 API breakage: func FileSystem.copyItem(at:to:shouldProceedAfterError:shouldCopyFile:) has parameter 2 type change from (_NIOFileSystem.DirectoryEntry, any Swift.Error) async throws -> Swift.Void to _NIOFileSystem.CopyStrategy
>   💔 API breakage: func FileSystem.copyItem(at:to:shouldProceedAfterError:shouldCopyFile:) has parameter 3 type change from (SystemPackage.FilePath, SystemPackage.FilePath) async -> Swift.Bool to (_NIOFileSystem.DirectoryEntry, any Swift.Error) async throws -> Swift.Void
>   💔 API breakage: func FileSystemProtocol.copyItem(at:to:strategy:shouldProceedAfterError:shouldCopyItem:) has been added as a protocol requirement
>   💔 API breakage: func FileSystem.copyItem(at:to:shouldProceedAfterError:shouldCopyFile:) has been renamed to func copyItem(at:to:strategy:shouldProceedAfterError:shouldCopyItem:)
>   💔 API breakage: func FileSystemProtocol.copyItem(at:to:) has been renamed to func copyItem(at:to:strategy:)

Since NIOFileSystem is still in leading underscore mode this is deemed acceptable (confirmed in discussions offline)
- In most cases the backward compatibility shim provided in ```FileSystemProtocol``` will cover it. 
- In cases where a user is actively using the callbacks or direct usage of ```FileSystem``` becoming aware of the semantic changes is desirable considering the cost to implement full backward compatibility

PR #2809, opened 2024-07-25
[CI] Add `shellcheck` and fix up warnings

FranzBusch:
# Motivation

We have quite a lot of shell scripts in our repo and want to make sure that they all pass `shellcheck`.

# Modification

This PR adds a GH action workflow to the soundness script for `shellcheck` and fixes up all errors and warnings.

# Result

No more shell/bash discussions


IntegrationTests/tests_01_http/defines.sh
```diff
@@ -84,7 +84,7 @@ function start_server() {
 
     mkdir "$tmp/htdocs"
     swift build
-    "$(swift build --show-bin-path)/NIOHTTP1Server" $extra_args $maybe_nio_host "$port" "$tmp/htdocs" &
+    "$(swift build --show-bin-path)/NIOHTTP1Server" "$extra_args" $maybe_nio_host "$port" "$tmp/htdocs" &
```
weissi, line 87: 99% sure this is incorrect as it wants to pass multiple args. `extra_args` needs to become an array, unless it already is and then you can `"${extra_args[@]}"`

IntegrationTests/run-tests.sh
```diff
@@ -36,7 +36,9 @@ function plugins_do() {
     done
 }
 
+# shellcheck source=/dev/null
```
simonjbeaumont, line 39: You can configure shellcheck where to look for these.

IntegrationTests/tests_01_http/defines.sh
```diff
@@ -58,7 +58,7 @@ function do_netstat() {
 }
 
 function create_token() {
-    mktemp "$tmp/server_token_XXXXXX"
+    mktemp "${tmp:?}/server_token_XXXXXX"
```
simonjbeaumont, line 61: If you're going to use `${foo?}` can you give a helpful message on the end of the parameter expansion, e.g. `${foo?"foo is not set"}`.

scripts/check-license-header.sh
```diff
@@ -36,13 +36,15 @@ expected_file_header_template="@@===--------------------------------------------
 
 paths_with_missing_license=( )
 
-file_paths=$(git ls-files $(cat .licenseignore | xargs -I% printf ":(exclude)% "))
+excludes=$(xargs -I% printf ":(exclude)% " < .licenseignore)
```
weissi, line 39: this won't be correct with files with spaces, quotes etc

scripts/check-swift-format.sh
```diff
@@ -18,22 +18,22 @@ log() { printf -- "** %s\n" "$*" >&2; }
...
 fatal() { error "$@"; exit 1; }
 
-excluded_files=""
+excludes=""
 
-if [ -f .swiftformatignore ]; then
+if [[ -f .swiftformatignore ]]; then
     log "Found swiftformatignore file..."
-    excluded_files=$(cat .swiftformatignore | xargs -I% printf ":(exclude)% ")
+    excludes=$(xargs -I% printf ":(exclude)% " < .swiftformatignore)
```
weissi, line 25: same here, quoting not right

IntegrationTests/tests_01_http/defines.sh
```diff
@@ -125,21 +127,25 @@ function start_server() {
 }
 
 function get_htdocs() {
+    # shellcheck source=/dev/null
     source "$1"
-    echo "$token_htdocs"
+    # shellcheck disable=SC2154
+    echo "${token_htdocs:?}"
```
simonjbeaumont, line 133: Why are we supporting it not being set? I guess the right way here would be to pass it as an argument to the function. But maybe you're trying to minimise churn?

IntegrationTests/tests_01_http/defines.sh
```diff
@@ -149,6 +155,7 @@ function stop_server() {
         if ! kill -0 "$token_pid" 2> /dev/null; then # ignore-unacceptable-language
             break # good, dead
         fi
+        # shellcheck disable=SC2009
```
simonjbeaumont, line 158: Didn't fancy `pgrep` or just minimising churn?

weissi (commented):
WHOA, super nice!! Found only one thing that looks like a regression. The xargs quoting errors are preexisting. Might wanna fix

IntegrationTests/tests_01_http/test_01_get_file.sh
```diff
@@ -13,6 +13,7 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+# shellcheck source=IntegrationTests/tests_01_http/defines.sh
```
simonjbeaumont, line 16: Right... this is what you can use elsewhere.

IntegrationTests/tests_01_http/test_03_post_random_bytes.sh
```diff
@@ -13,11 +13,12 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+# shellcheck source=IntegrationTests/tests_01_http/defines.sh
 source defines.sh
 
 token=$(create_token)
 start_server "$token"
-dd if=/dev/urandom of="$tmp/random.bytes" bs=$((64*1024)) count=1
+dd if=/dev/urandom of="${tmp:?}/random.bytes" bs=$((64*1024)) count=1
```
simonjbeaumont, line 21: We're seeing this pattern a lot in this PR. Are all these things coming from `defines.sh` presumably? 

If not, why are we not explicitly setting it somewhere?

simonjbeaumont (commented):
Thanks for doing this @FranzBusch. You might consider a `.shellcheckrc` file, or at least confirm that, if one _is_ present in a repo, it will be respected?

scripts/check-license-header.sh
FranzBusch: The previous one also wasn't correct then right?

IntegrationTests/tests_01_http/defines.sh
weissi: The `kill -0 "$token_pid"` does the right thing. IIRC, the bad-looking `ps auxw | grep "$token_pid" || true` is for actual human output. The idea is: We check (correctly) with `kill -0` is a pid is dead. But if it isn't, then we present the user with the well-known `ps axuw` output so they get to see what's up.

I'm happy to also delete `ps auxw | grep "$token_pid" || true` without replacement. This hasn't failed in 6 years and if it does we can re-add debugging.

scripts/check-license-header.sh
weissi: indeed, neither of them are correct

IntegrationTests/tests_01_http/defines.sh
FranzBusch: Yes minimising churn since this didn't seem required but more a suggestion

IntegrationTests/tests_01_http/defines.sh
FranzBusch: Yeah I didn't want to rearchitect all of this

scripts/check-license-header.sh
weissi: To make it correct, you'd need something like

```
declare -a excludes=()
while IFS=$'\n' read -r -a line; do excludes+=( ":(exclude)$(printf "%q" "$line")" ); done

[...]

git ls-files "${excludes[@]}"
```

untested, but should be roughly right

IntegrationTests/tests_01_http/test_03_post_random_bytes.sh
FranzBusch: Those don't come from the `defines.sh` but rather `run-tests.sh` is setting the tmp dir and then all other scripts are executed with that from what I can tell

scripts/check-license-header.sh
simonjbeaumont: This is working when filenames have spaces in the include file.

```
eval git ls-files $(cat IGNOREFILE | tr '\n' '\0' | xargs -0 -I% printf '":(exclude)%" ')
```

simonjbeaumont:
```
❯ cat IGNOREFILE | tr '\n' '\0' | xargs -0 -I% printf '":(exclude)%" ' | xargs git ls-files
```
no eval

FranzBusch:
> You might consider a `.shellcheckrc` file, or at least confirm that, if one _is_ present in a repo, it will be respected?

My understand is that `shellcheck` automatically looks for that file and there is a `--norc` option to disable looking for it

scripts/check-license-header.sh
weissi: yeah, that one breaks on files with `"`s in it though. Not implying that it's super important but it's also not correct.

e.g.

```
$ cat ignorefile 
test.sh
foo " bar
foo bar
foobar
'
test.sh
johannes:/tmp/foo (main)
$ cat ignorefile | tr '\n' '\0' | xargs -0 -I% printf '":(exclude)%" ' | xargs git ls-files
xargs: unterminated quote
```

FranzBusch: At this point I consider this script good enough :D I am done spending my time handling all cases. We control the file names so we can just not use `"` in a filename

weissi: Sure, I did however suggest something that might actually work. But your call

simonjbeaumont (approved):

PR #2812, opened 2024-07-26
[CI] Add Swift 6 language mode workflow

FranzBusch:
# Motivation

We want to check how our packages perform in Swift 6 language mode.

# Modification

This PR adds a new reusable workflow that sets the tools version to Swift 6 and then builds with warnings as errors enabled.

Additionally, I did a bit of cleanup for the workflow names to hopefully make them look nice in every GH Action UI.

# Result

We now get a feeling for our Swift 6 language mode compatibility.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,4 +1,4 @@
-name: Pull Request
+name: PR
 
 on:
     pull_request:
@@ -12,7 +12,7 @@ jobs:
             license_header_check_project_name: "SwiftNIO"
 
     unit-tests:
-        name: Checks
+        name: Unit tests
         uses: ./.github/workflows/unit_tests.yml
         with:
             linux_5_8_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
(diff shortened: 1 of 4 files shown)


FranzBusch:
Turns out we are not warnings clean on Swift 6 mode 🙃

FranzBusch:
@Lukasa I think we gotta live with this naming for now on the checks. If you could be so kind and update the branch protection rules hopefully one last time

Lukasa:
Have you deliberately not added the repeated name to the integration tests @FranzBusch?

FranzBusch:
> Have you deliberately not added the repeated name to the integration tests @FranzBusch?

Yeah. That's the only way to make this work sadly. Since the other workflows are reusable workflows and the integration tests are just a job of our PR workflow

Lukasa:
Why aren't the integration tests reusable? We do have them in other NIO repos.

FranzBusch:
> Why aren't the integration tests reusable? We do have them in other NIO repos.

Yes more repos have a script that executes the integration tests but those are not "standard" commands like `swift test`. So every repo has to configure its own matrix reusable workflow to call the appropriate script. I see little value in adding a reusable workflow that just calls `./scripts/run-integration-tests.sh`. Furthermore, it is totally fine and expected that each repo has their own jobs. For example, the OpenAPI repo has a job to build all of their example packages.

Now coming back to the problem at hand here. The naming that GH is displaying for actions is just different in the different UIs.
- On the PR: `root workflow name / root job name / reusable workflow job name / unit tests job name / matrix job name`
- In other UI: `root job name` then a drop down `matrix job name`

Lukasa:
Ok great, branch changes made.

FranzBusch:
@Lukasa Thanks. Feel free to approve then

Lukasa (approved):

PR #2814, opened 2024-07-26
[CI] Add action and workflow to check for semver label

FranzBusch:
# Motivation

We want to make sure every PR that gets merged is appropriately labeled.

# Modification

This PR adds a new workflow and action to check for a semver label to be present.

# Result

No more labeling after a PR got merged.

### DIFF
--- a/.github/actions/pull_request_semver_label_checker.yml
+++ b/.github/actions/pull_request_semver_label_checker.yml
@@ -1,0 +1,9 @@
+name: 'Pull request semver label checker'
+description: 'Checks that at least one semver label is applied to the pull request'
+runs:
+    using: "composite"
+    steps:
+      - name: Check labels
+        run: |
+            gh pr view ${{ github.event.number }} --repo apple/swift-nio --json labels \
+            | jq -e '[.labels[].name] | any(. == "semver/major" or . == "semver/minor" or . == "semver/patch" or . == "semver/none")'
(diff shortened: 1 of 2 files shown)


.github/actions/pull_request_semver_label_checker.yml
```diff
@@ -0,0 +1,9 @@
+name: 'Pull request semver label checker'
+description: 'Checks that at least one semver label is applied to the pull request'
+runs:
+    using: "composite"
+    steps:
+      - name: Check labels
+        run: |
+            gh pr view 2809 --repo apple/swift-nio --json labels \
```
glbrntt, line 8: Don't we want to get the current PR number or is this just hardcoded for testing and to be updated in a later PR?

.github/actions/pull_request_semver_label_checker.yml
```diff
@@ -0,0 +1,9 @@
...
+            | jq -e '[.labels[].name] | any(. == "semver/major" or . == "semver/minor" or . == "semver/patch" or . == "semver/none")'
```
glbrntt, line 9: Also do we want this to be in a script?

FranzBusch: I intentionally left it inline here because it is such a small thing and scripting didn't provide any meaningful local impact

glbrntt (approved):

PR #2823, opened 2024-08-02
Clone files on Darwin rather than copying them

rnro:
### Motivation:

Copying regular files on Darwin was taking more time than it needed to because it was manually opening files and copying over bites rather than making use of the system's ability to create CoW clones.

### Modifications:

The primary change is to switch `copyItem` to be backed by Darwin's `copyfile` method rather than `fcopyfile`. `fcopyfile` ignores the flag we were passing in to indicate that the file should be cloned if possible, whereas `copyfile` is a higher-level API which takes this into account.

This change exposes a `copyfile` wrapper method as a new public function.

This change also adds a workaround to a seeming compiler bug on channel options blocking building tests on more recent Xcodes.

### Result:

Copying files on Darwin will be faster.

### DIFF
--- a/Sources/NIOFileSystem/FileSystem.swift
+++ b/Sources/NIOFileSystem/FileSystem.swift
@@ -1183,7 +1183,7 @@ extension FileSystem {
             state: nil,
             flags: flags
         ).mapError { errno in
-            FileSystemError.fcopyfile(
+            FileSystemError.copyfile(
                 errno: errno,
                 from: sourcePath,
                 to: destinationPath,
--- a/Sources/NIOFileSystem/FileSystem.swift
+++ b/Sources/NIOFileSystem/FileSystem.swift
@@ -1172,6 +1172,27 @@ extension FileSystem {
             )
         }
 
+        #if canImport(Darwin)
+        // COPYFILE_CLONE clones the file if possible and will fallback to doing a copy.
+        // COPYFILE_ALL is shorthand for:
+        //    COPYFILE_STAT | COPYFILE_ACL | COPYFILE_XATTR | COPYFILE_DATA
+        let flags = copyfile_flags_t(COPYFILE_CLONE) | copyfile_flags_t(COPYFILE_ALL)
+        return Libc.copyfile(
+            from: sourcePath,
+            to: destinationPath,
+            state: nil,
+            flags: flags
+        ).mapError { errno in
+            FileSystemError.fcopyfile(
+                errno: errno,
+                from: sourcePath,
+                to: destinationPath,
+                location: .here()
+            )
+        }
+
+        #elseif canImport(Glibc) || canImport(Musl) || canImport(Bionic)
+
         let openSourceResult = self._openFile(
             forReadingAt: sourcePath,
             options: OpenOptions.Read(followSymbolicLinks: true)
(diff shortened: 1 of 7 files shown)


Tests/NIOPosixTests/StreamChannelsTest.swift
```diff
@@ -365,9 +365,8 @@ class StreamChannelTest: XCTestCase {
             )
 
             // We will immediately send exactly the amount of data that fits in the receiver's receive buffer.
-            let receiveBufferSize = Int(
-                (try? receiver.getOption(.socketOption(.so_rcvbuf)).wait()) ?? 8192
-            )
+            let opt = try? receiver.getOption(.socketOption(.so_rcvbuf)).wait()
+            let receiveBufferSize = Int(opt ?? 8192)
```
glbrntt, line 369: Can you do this in separate PR?

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -1172,6 +1172,27 @@ extension FileSystem {
...
+        // COPYFILE_ALL is shorthand for:
+        //    COPYFILE_STAT | COPYFILE_ACL | COPYFILE_XATTR | COPYFILE_DATA
+        let flags = copyfile_flags_t(COPYFILE_CLONE) | copyfile_flags_t(COPYFILE_ALL)
+        return Libc.copyfile(
+            from: sourcePath,
+            to: destinationPath,
+            state: nil,
+            flags: flags
+        ).mapError { errno in
+            FileSystemError.fcopyfile(
```
glbrntt, line 1186: Can you update this helper? One of the things it'll do is attach a `SyscallError` to the `FileSystemError` which will include the name of the syscall, now it will incorrectly be `fcopyfile`. 

Where it doesn't make sense to duplicate the helper I've tended to just pass in the name of the syscall: `FileSystemError.copyfile("copyfile", ... )`

Sources/NIOFileSystem/FileSystemError+Syscall.swift
```diff
@@ -1016,6 +1016,25 @@ extension FileSystemError {
                 Can't copy file from '\(sourcePath)' to '\(destinationPath)', the destination \
                 path already exists.
                 """
+        case .fileExists:
```
glbrntt, line 1019: Can you update the error tests? They check that `errno`s get mapped to an appropriate file system error code.

Sources/NIOFileSystem/Internal/System Calls/Syscalls.swift
```diff
@@ -400,7 +400,7 @@ internal func libc_remove(
 }
 
 #if canImport(Darwin)
-/// copyfile(3): Copy a file from one file to another.
+/// copyfile(3): Copy a file from one file to another. (fcopyfile)
```
glbrntt, line 403: nit
```suggestion
/// fcopyfile(3): Copy a file from one file to another.
```

Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
```diff
@@ -313,6 +313,24 @@ public enum Libc {
     }
     #endif
 
+    #if canImport(Darwin)
+    @_spi(Testing)
+    public static func copyfile(
```
glbrntt, line 318: Can you add a test for this please? https://github.com/apple/swift-nio/blob/ba6245748fe1c07d72bd486fb056deced5baad21/Tests/NIOFileSystemTests/Internal/SyscallTests.swift#L376?

glbrntt (changes requested):
Looks great, thanks for fixing this Rick. The actual fix looks good but I left a couple of nitty things to fix up.

Tests/NIOPosixTests/StreamChannelsTest.swift
rnro: https://github.com/apple/swift-nio/pull/2826

glbrntt (approved):
Great patch, thanks @rnro!

PR #2826, opened 2024-08-05
StreamChannelsTest don't build on recent Xcode

rnro:
### Motivation:

This change adds a workaround to a seeming compiler bug on channel options blocking building tests on more recent Xcodes.

### Modifications:

Refactor the code around the socket option get.

### Result:

Code builds.


rnro:
https://github.com/apple/swift-nio/pull/2824

PR #2831, opened 2024-08-07
Provide documentation and context information for NIOTooManyBytesError

cmcgee1024:
### Motivation:

The NIOTooManyBytesError doesn't have any documentation for someone that encounters this error. Also, they have no idea the magnitude of the limit that was set to decide if the payload is an unreasonable size.

### Modifications:

Provide documentation that explains the situation when the error occurs, which is the upTo limit of an AsyncSequence is exceeded. Describe one potential action, which is to increase this limit.

Provide the maxBytes in the error so that the user can gauge whether the upTo limit is already reasonable and the payload size is excessive, or the limit needs to be increased to suit more situations.

### Result:

There will be documentation for the NIOTooManyBytesError so that if someone looks it up they will have a better understanding of the situation, and possible actions. The limit will be included in the error to give them an idea of the scale of the payload limit that is in place, or potentially the configuration value that they can change.


cmcgee1024:
@weissi would it preferable to provide a backward compatible init function and make maxBytes optional?

weissi:
> @weissi would it preferable to provide a backward compatible init function and make maxBytes optional?

Yes, we should not break SemVer major just for this. I'll let @Lukasa  & @FranzBusch et al comment more on what API would be good

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -240,8 +240,25 @@ extension ChannelPipeline {
...
+/// When collecting the bytes from an AsyncSequence, there is a limit up to where the content
+/// exceeds a certain threshold beyond which the content isn't matching the expected maximum size.
+/// that will be processed. This error is generally thrown when it is discovered that there are
+/// more bytes in a sequence than what was specified as the maximum. It could be that this upTo
+/// limit should be increased, or that the sequence has unexpected content in it.
+///
+/// - maxBytes: (optional) current limit on the maximum number of bytes in the sequence
 public struct NIOTooManyBytesError: Error, Hashable {
-    public init() {}
+    let maxBytes: Int?
```
dnadoba, line 253: This should be a var. no need to restrict it to a let as there are no invariants or similar that need to be verified.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -240,8 +240,25 @@ extension ChannelPipeline {
...
 
+/// An error that is thrown when the number of bytes in an AsyncSequence exceeds the limit.
+///
+/// When collecting the bytes from an AsyncSequence, there is a limit up to where the content
+/// exceeds a certain threshold beyond which the content isn't matching the expected maximum size.
+/// that will be processed. This error is generally thrown when it is discovered that there are
+/// more bytes in a sequence than what was specified as the maximum. It could be that this upTo
+/// limit should be increased, or that the sequence has unexpected content in it.
+///
+/// - maxBytes: (optional) current limit on the maximum number of bytes in the sequence
```
dnadoba, line 251: The docs for a property should be above the property.

dnadoba (commented):
We likely want to make two of these errors equal regardless of their value of the maxBytes property.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -240,8 +240,25 @@ extension ChannelPipeline {
...
 public struct NIOTooManyBytesError: Error, Hashable {
-    public init() {}
+    let maxBytes: Int?
+
+    public init() {
```
Lukasa, line 255: We probably want to deprecate this constructor.

Sources/NIOCore/AsyncAwaitSupport.swift
Lukasa: Also this should be public.

Lukasa (changes requested):
This generally looks good, but I'd like to see the tests updated to validate that this is working properly.

Sources/NIOCore/AsyncAwaitSupport.swift
cmcgee1024: If it's var and public, then what's a good way to prevent it from being mutated after the initial throw?

Lukasa: There's no reason to prevent it being mutated. And indeed, given the `init`, you can't:

```swift
extension NIOTooManyBytesError {
   mutating func getOutOfMyWayFoolsIMutateWhatIWant(maxBytes: Int?) {
       self = .init(maxBytes: maxBytes)
   }
}

Comment (cmcgee1024) [reply Sources/NIOCore/AsyncAwaitSupport.swift]: Ok, I've updated it to be public mutable.

File: Sources/NIOCore/AsyncAwaitSupport.swift (hunk 3 above)
Comment (cmcgee1024) [reply Sources/NIOCore/AsyncAwaitSupport.swift:257]: It's deprecated now.

File: Tests/NIOCoreTests/AsyncSequenceTests.swift [hunk 4]
```diff
@@ -138,14 +139,22 @@ final class AsyncSequenceCollectTests: XCTestCase {
                     file: testCase.file,
                     line: testCase.line
                 )
+                if let tooManyBytesErr = error as? NIOTooManyBytesError {
```
Lukasa, line 142: This should fail the test if the cast doesn't pass.

Tests/NIOCoreTests/AsyncSequenceTests.swift
```diff
@@ -154,6 +163,14 @@ final class AsyncSequenceCollectTests: XCTestCase {
...
```
Lukasa, line 166: This should fail the test if the cast doesn't pass.

Sources/NIOCore/AsyncAwaitSupport.swift
```diff
@@ -240,8 +240,25 @@ extension ChannelPipeline {
...
 }
 
+/// An error that is thrown when the number of bytes in an AsyncSequence exceeds the limit.
+///
+/// When collecting the bytes from an AsyncSequence, there is a limit up to where the content
+/// exceeds a certain threshold beyond which the content isn't matching an expected reasonable
+/// size to be processed. This error is generally thrown when it is discovered that there are more
+/// more bytes in a sequence than what was specified as the maximum. It could be that this upTo
+/// limit should be increased, or that the sequence has unexpected content in it.
 public struct NIOTooManyBytesError: Error, Hashable {
```
dnadoba, line 250: Can you add a custom `Equatable` and `Hashable` implementation that ignores `maxBytes`? `==` can just return true and `hash(into:)` can just hash a constant e.g. `0`. Users are usually not interested in the specific `maxBytes` value but rather just that it is a `NIOTooManyBytesError` error.

dnadoba (approved):
Thanks! LGTM

Lukasa (approved):

cmcgee1024:
After looking through the various failures I see that there's a formatter failure. How best to run the formatter for this project to reformat the code?

PR #2835, opened 2024-08-09
Adopt strict concurrency in _NIODataStructures

Lukasa:
Motivation:

We're slowly adopting strict concurrency in NIO. This is next.

Modifications:

- Make TinyArray's Iterator Sendable
- Add the stict concurrency flags

Result:

One small step for NIO, one giant leap for NIOkind.


FranzBusch (approved):
Outstanding!

Issue #2748, opened 2024-06-21
ByteBuffer should have hex bytes initialiser / write method


weissi:
This is pretty much a companion of #2447 / #2475. The aforementioned bugs/patches introduce the ability to get the hex dump from a `ByteBuffer` which is awesome.

Next, I'd like to construct/write into a `ByteBuffer` from a hex dump:
- `ByteBuffer(hexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")`
- `buffer.writeHexEncodedBytes("68656c6c6f20776f726c640a")`

I'd say it just needs to support the `.plain` format (i.e. 2 hex-digit values with whitespace ignored)

---

Super basic test:

```
var buffer = ByteBuffer(hexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")
buffer.writeHexEncodedBytes("68656c6c6f20776f726c64")
buffer.writeHexEncodedBytes("")
buffer.writeHexEncodedBytes("0a")
XCTAssertEqual(ByteBuffer(string: "hello world\nhello world\n"), buffer)
```

another one:

```
var allBytes = ByteBuffer()
for x in UInt8.min ... UInt8.max {
    buffer.writeInteger(x)
}

let allBytesHex = allBytes.hexDump(format: .plain)
let allBytesDecoded = ByteBuffer(hexEncodedBytes: allBytesHex)

XCTAssertEqual(allBytes, allBytesDecoded)
```

ali-ahsan-ali:
I have taken a look and would like to pick this up after https://github.com/apple/swift-nio/issues/2734.

weissi:
Of course this should also parse the `.compat` format suggested in #2825

weissi:
> Happy if someone else is eager to give it a go. I have started looking at it but no real progress. It feels overly ambitious to try and start working on a second PR when my first has taken quite a while.

I did re-request review on your other PR. Remember to click the  :recycle  (re-request) review once you addressed the changed suggested/requested. I'm sure @FranzBusch & @Lukasa just missed that you pushed updates to your existing PR.

Separately, if you're up for it, you can totally have multiple PRs open at the same time, shouldn't slow things down.

ali-ahsan-ali:
The "good for newcomers" in this repo are the hardest thing every 😅 (which I VERY much enjoy).

Experimenting with the below and seeing if writing that to the buffer is the right outcome for writing a hex encoded string to buffer. 

[...]

PR #2837, opened 2024-08-11
Issue-2748 - Add ByteBuffer Hex init & write

ali-ahsan-ali:
Add ability for below two functions 
`ByteBuffer(hexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")`
` buffer.writeHexEncodedBytes("68656c6c6f20776f726c640a")`

### Motivation:

https://github.com/apple/swift-nio/issues/2748

### DIFF
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -92,19 +92,14 @@ extension ByteBuffer {
     }
 
     // MARK: Hex encoded string APIs
-    /// Write `hex encoded string` into this `ByteBuffer`, moving the writer index forward appropriately.
-    ///
+    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately.
     /// - parameters:
     ///     - string: The hex encoded string to write.
     /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
     public mutating func writeHexEncodedBytes(_ string: String) -> Int {
-        let hexPlainDecodedBytes = string.hexPlainDecodedBytes
-        guard !hexPlainDecodedBytes.isEmpty else {
-            return 0
-        }
-        let written = self.setBytes(hexPlainDecodedBytes, at: self.writerIndex)
+        let written = self.setBytes(string.hexPlainDecodedBytes, at: self.writerIndex)
         self._moveWriterIndex(forwardBy: written)
         return written
     }
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -776,6 +776,18 @@ extension ByteBufferAllocator {
         return buffer
     }
 
+    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from the ASCII `hexEncodedBytes` string .
+    ///
+    /// This will allocate a new `ByteBuffer` with enough space to fit `bytes` and potentially some extra space.
+    ///
+    /// - returns: The `ByteBuffer` containing the written bytes.
+    @inlinable
+    public func buffer(hexEncodedBytes string: String) -> ByteBuffer {
+        var buffer = self.buffer(capacity: string.hexPlainDecodedBytes.underestimatedCount)
+        buffer.writeBytes(string.hexPlainDecodedBytes)
+        return buffer
+    }
+    
     /// Create a fresh `ByteBuffer` containing the bytes of the byte representation in the given `endianness` of
     /// `integer`.
     ///
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -101,9 +101,15 @@ extension ByteBuffer {
     public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
         var slice = plainHexEncodedBytes.utf8[...]
         var written = 0
+        var bytesToWrite: [UInt8] = []
 
+        // Single pass to ensure string is a valid representation of hex bytes
         while let nextByte = try slice.popNextHexByte() {
-            written += self.writeInteger(nextByte)
+            bytesToWrite.append(nextByte)
+        }
+
+        for byte in bytesToWrite {
+            written += self.writeInteger(byte)
         }
 
         return written
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -92,27 +92,26 @@ extension ByteBuffer {
     }
 
     // MARK: Hex encoded string APIs
-    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately.
+    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately. This method will throw if the string input is not of the "plain" hex encoded format.
     /// - parameters:
-    ///     - plainHexEncodedBytes: The hex encoded string to write.
+    ///     - plainHexEncodedBytes: The hex encoded string to write. Plain hex dump format is hex bytes optionally separated by spaces, i.e. `48 65 6c 6c 6f` or `48656c6c6f` for `Hello`.
+    ///     This format is compatible with `xxd` CLI utility.
     /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
     public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
         var slice = plainHexEncodedBytes.utf8[...]
-        var written = 0
-        var bytesToWrite: [UInt8] = []
+        let initialWriterIndex = self._writerIndex
 
-        // Single pass to ensure string is a valid representation of hex bytes
-        while let nextByte = try slice.popNextHexByte() {
-            bytesToWrite.append(nextByte)
+        do {
+            while let nextByte = try slice.popNextHexByte() {
+                self.writeInteger(nextByte)
+            }
+            return Int(self._writerIndex - initialWriterIndex)
+        } catch {
+            self._moveWriterIndex(to: initialWriterIndex)
+            throw error
         }
-
-        for byte in bytesToWrite {
-            written += self.writeInteger(byte)
-        }
-
-        return written
     }
 
     // MARK: String APIs
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -92,7 +92,8 @@ extension ByteBuffer {
     }
 
     // MARK: Hex encoded string APIs
-    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately. This method will throw if the string input is not of the "plain" hex encoded format.
+    /// Write ASCII hexadecimal `string` into this `ByteBuffer` as raw bytes, decoding the hexadecimal & moving the writer index forward appropriately.
+    /// This method will throw if the string input is not of the "plain" hex encoded format.
     /// - parameters:
     ///     - plainHexEncodedBytes: The hex encoded string to write. Plain hex dump format is hex bytes optionally separated by spaces, i.e. `48 65 6c 6c 6f` or `48656c6c6f` for `Hello`.
     ///     This format is compatible with `xxd` CLI utility.
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -101,15 +101,15 @@ extension ByteBuffer {
     @inlinable
     public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
         var slice = plainHexEncodedBytes.utf8[...]
-        let initialWriterIndex = self._writerIndex
+        let initialWriterIndex = self.writerIndex
 
         do {
             while let nextByte = try slice.popNextHexByte() {
                 self.writeInteger(nextByte)
             }
-            return Int(self._writerIndex - initialWriterIndex)
+            return self.writerIndex - initialWriterIndex
         } catch {
-            self._moveWriterIndex(to: initialWriterIndex)
+            self.moveWriterIndex(to: initialWriterIndex)
             throw error
         }
     }
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -94,13 +94,18 @@ extension ByteBuffer {
     // MARK: Hex encoded string APIs
     /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately.
     /// - parameters:
-    ///     - string: The hex encoded string to write.
+    ///     - plainHexEncodedBytes: The hex encoded string to write.
     /// - returns: The number of bytes written.
     @discardableResult
     @inlinable
-    public mutating func writeHexEncodedBytes(_ string: String) -> Int {
-        let written = self.setBytes(string.hexPlainDecodedBytes, at: self.writerIndex)
-        self._moveWriterIndex(forwardBy: written)
+    public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
+        var slice = plainHexEncodedBytes.utf8[...]
+        var written = 0
+
+        while let nextByte = try slice.popNextHexByte() {
+            written += self.writeInteger(nextByte)
+        }
+
         return written
     }
 
@@ -776,18 +781,18 @@ extension ByteBufferAllocator {
         return buffer
     }
 
-    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from the ASCII `hexEncodedBytes` string .
+    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from the ASCII `plainHexEncodedBytes` string .
     ///
     /// This will allocate a new `ByteBuffer` with enough space to fit `bytes` and potentially some extra space.
     ///
     /// - returns: The `ByteBuffer` containing the written bytes.
     @inlinable
-    public func buffer(hexEncodedBytes string: String) -> ByteBuffer {
-        var buffer = self.buffer(capacity: string.hexPlainDecodedBytes.underestimatedCount)
-        buffer.writeBytes(string.hexPlainDecodedBytes)
+    public func buffer(plainHexEncodedBytes string: String) throws -> ByteBuffer {
+        var buffer = self.buffer(capacity: string.utf8.count / 2)
+        try buffer.writePlainHexEncodedBytes(string)
         return buffer
     }
-    
+
     /// Create a fresh `ByteBuffer` containing the bytes of the byte representation in the given `endianness` of
     /// `integer`.
     ///
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -91,6 +91,24 @@ extension ByteBuffer {
         )
     }
 
+    // MARK: Hex encoded string APIs
+    /// Write `hex encoded string` into this `ByteBuffer`, moving the writer index forward appropriately.
+    ///
+    /// - parameters:
+    ///     - string: The hex encoded string to write.
+    /// - returns: The number of bytes written.
+    @discardableResult
+    @inlinable
+    public mutating func writeHexEncodedBytes(_ string: String) -> Int {
+        let hexPlainDecodedBytes = string.hexPlainDecodedBytes
+        guard !hexPlainDecodedBytes.isEmpty else {
+            return 0
+        }
+        let written = self.setBytes(hexPlainDecodedBytes, at: self.writerIndex)
+        self._moveWriterIndex(forwardBy: written)
+        return written
+    }
+
     // MARK: String APIs
     /// Write `string` into this `ByteBuffer` using UTF-8 encoding, moving the writer index forward appropriately.
     ///
--- a/Sources/NIOCore/ByteBuffer-hex.swift
+++ b/Sources/NIOCore/ByteBuffer-hex.swift
@@ -24,7 +24,7 @@ extension ByteBuffer {
     ///         buffer use `channel.allocator.buffer(capacity: ...)` to allocate a `ByteBuffer` of the right
     ///         size followed by a `writeHexEncodedBytes` instead of using this method. This allows SwiftNIO to do
     ///         accounting and optimisations of resources acquired for operations on a given `Channel` in the future.
-    init(plainHexEncodedBytes string: String) throws {
+    public init(plainHexEncodedBytes string: String) throws {
         self = try ByteBufferAllocator().buffer(plainHexEncodedBytes: string)
     }
 
@@ -354,21 +354,25 @@ extension ByteBuffer {
             }
         }
     }
-}
 
-enum HexDecodingError: Error {
-    case invalidHexLength
-    case invalidCharacter
+    public struct HexDecodingError: Error {
+        let kind: HexDecodingErrorKind
+
+        public enum HexDecodingErrorKind {
+            case invalidHexLength
+            case invalidCharacter
+        }
+    }
 }
 
 extension UInt8 {
     fileprivate var isASCIIWhitespace: Bool {
-        [UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\n"), UInt8(ascii: "\r"), UInt8(ascii: " ")].contains(
+        [UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\r"), UInt8(ascii: " ")].contains(
             self
         )
     }
 
-    fileprivate var asciiHexNibble: UInt8? {
+    fileprivate var asciiHexNibble: UInt8 {
         get throws {
             switch self {
             case UInt8(ascii: "0")...UInt8(ascii: "9"):
@@ -378,7 +382,7 @@ extension UInt8 {
             case UInt8(ascii: "A")...UInt8(ascii: "F"):
                 return self - UInt8(ascii: "A") + 10
             default:
-                throw HexDecodingError.invalidCharacter
+                throw ByteBuffer.HexDecodingError(kind: .invalidCharacter)
             }
         }
     }
@@ -395,12 +399,8 @@ extension Substring.UTF8View {
             return nil  // No next byte to pop
         }
 
-        while let nextByte = self.first, nextByte.isASCIIWhitespace {
-            self = self.dropFirst()
-        }
-
         guard let secondHex = try self.popFirst()?.asciiHexNibble else {
-            throw HexDecodingError.invalidHexLength
+            throw ByteBuffer.HexDecodingError(kind: .invalidHexLength)
         }
 
         return (firstHex << 4) | secondHex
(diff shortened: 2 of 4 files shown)


Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -12,8 +12,30 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Foundation
```
ali-ahsan-ali, line 15: Is this allowed in NioCore? If not, can I use the "replaceWith" that is only MacOs 13.0 and above? Or should I implement the ability to replace white spaces of a string custom in this module?

ali-ahsan-ali: Ignore, found a solve.

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -88,6 +88,24 @@ extension ByteBuffer {
         )
     }
 
+    // MARK: Hex encoded string APIs
+    /// Write `hex encoded string` into this `ByteBuffer`, moving the writer index forward appropriately.
```
Lukasa, line 92:
```suggestion
    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately.
```

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -263,3 +283,23 @@ extension ByteBuffer {
         }
     }
 }
+
+extension String {
+    /// Plain decode a string representing a hexadecimal sequence them into an UInt8 sequence
+    /// - Complexity: O(n)
+    public var hexPlainDecodedBytes: [UInt8] {
```
Lukasa, line 290: This should not be `public`.

We also don't want to round-trip it through an Array: we should be able to convert directly into `ByteBuffer`.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -14,6 +14,26 @@
...
+    /// This will allocate a new `ByteBuffer` with enough space to fit the hex decoded `bytes` and potentially some extra space
+    /// using the default allocator.
+    ///
+    /// - info: If you have access to a `Channel`, `ChannelHandlerContext`, or `ByteBufferAllocator` we
+    ///         recommend using `channel.allocator.buffer(integer:)`. Or if you want to write multiple items into the
+    ///         buffer use `channel.allocator.buffer(capacity: ...)` to allocate a `ByteBuffer` of the right
+    ///         size followed by a `writeHexEncodedBytes` instead of using this method. This allows SwiftNIO to do
+    ///         accounting and optimisations of resources acquired for operations on a given `Channel` in the future.
+    init(hexEncodedBytes string: String) {
+        let hexPlainDecodedBytes = string.hexPlainDecodedBytes
```
Lukasa, line 28: Rather than use this, we can create the buffer `var` and `writeHexEncodedString` into it.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Screenshot 2024-08-13 at 11 39 23 am

Without turning it into an array we lose 2 things 

1. The ability to count the array for the exact number of bytes. The underestimated count can give 0 for sequence length that that are really 11 and it then means either I cannot allocate a buffer of the correct length ( or return early without setting any bytes). 

So if in the initliser someone tried to enter a hex string without anything in it, it would still return an empty buffer, but after some extrannuous computation and going through the setBytes slow path.

Or again, if someone tried to write empty bytes.

2. It always uses the slow method for setting bytes in the sequence rather than the contiuous if available which is done always in an array from my testing. 

Is this extra rount trip from sequence to array worth more than the cost of the above? It is becoming a difficult question for me to answer. After your comment I am leaning towards it is not worth the extra cost, seems like there is a change it takes the slow path ANYWAYS with an array. So the fast case might be faster, but the slow case slower. 

Putting my brain dump here for challenging the outcome or the thought process.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Can I challenge this? It doesnt seem like the pattern.
Screenshot 2024-08-13 at 11 58 16 am

Would this suffice or do you want me to create that var? 
Screenshot 2024-08-13 at 11 58 32 am

ali-ahsan-ali: Is it because I am calling that extra computer variable inside the initialiser that makes this not part of the original pattern and needings something new.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Made the changes. Take a look when you find the time

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Made the changes, but I think my challenge still stands.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -14,6 +14,22 @@
...
+    ///
+    /// This will allocate a new `ByteBuffer` with enough space to fit the hex decoded `bytes` and potentially some extra space
+    /// using the default allocator.
+    ///
+    /// - info: If you have access to a `Channel`, `ChannelHandlerContext`, or `ByteBufferAllocator` we
+    ///         recommend using `channel.allocator.buffer(integer:)`. Or if you want to write multiple items into the
+    ///         buffer use `channel.allocator.buffer(capacity: ...)` to allocate a `ByteBuffer` of the right
+    ///         size followed by a `writeHexEncodedBytes` instead of using this method. This allows SwiftNIO to do
+    ///         accounting and optimisations of resources acquired for operations on a given `Channel` in the future.
+    init(hexEncodedBytes string: String) {
```
Lukasa, line 27: You right pointed out that we need to move this to `ByteBufferAllocator`, and then have this method call that one.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -263,3 +279,24 @@ extension ByteBuffer {
...
     }
 }
+
+extension String {
+    /// Plain decode a string representing a hexadecimal sequence them into an UInt8 sequence
+    /// - Complexity: O(n)
+    @inlinable
+    var hexPlainDecodedBytes: UnfoldSequence<UInt8, String> {
+        let stringWithoutWhiteSpaces = self.filter { !$0.isWhitespace }
+        return sequence(
```
Lukasa, line 289: I think this might be a bit too clever, and I wonder if we should try a simpler pattern. Instead of doing a `filter` followed by a `sequence`, all working on the `String`, can I propose we do a for loop? Something like this:

```swift
extension UInt8 {
    fileprivate var isASCIIWhitespace: Bool {
        switch self {
        case UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\n"), UInt8(ascii: "\r"):
            return true
        default: return false
        }
    }

    var asciiHexNibble: UInt8? {
        switch self {
        case UInt8(ascii: "0")...UInt8(ascii: "9"):
            return self - UInt8(ascii: "0")
        case UInt8(ascii: "a")...UInt8(ascii: "f"):
            return self - UInt8(ascii: "a")
        case UInt8(ascii: "A")...UInt8(ascii: "F"):
            return self - UInt8(ascii: "A")
        default:
            return nil
        }
    }
}

extension Substring.UTF8View {
    mutating func popNextHexByte() -> UInt8? {
        while let nextByte = self.first, nextByte.isASCIIWhitespace {
            self = self.dropFirst()
        }
        
        guard let firstHex = self.popFirst()?.asciiHexNibble, 
            let secondHex = self.popFirst()?.asciiHexNibble else {
            return nil
        }
        
        return (firstHex << 4) | secondHex
    }
}

extension ByteBuffer {
    mutating func writeHexEncodedBytes(_ hexString: String) -> Int {
        var slice = hexString.utf8[...]
        var written = 0
        while let nextByte = slice.popNextHexByte() {
            written += self.writeInteger(nextByte)
        }
        return written
    }
}
```

Lukasa: As a side note, we should probably amend the above to `throw` if the input string is invalid.

ali-ahsan-ali: I cannot wrap my head around the ascii nibble and popNextHexByte function. It seems to be decoding the letters wrong. I still feel like turning the String into Hex using Uint(string, radix: 16) is the simple and correct way to do things. 

If the sequence is over complicating it, I can implement it using a for loop. Something like the below. 

```swift
enum HexDecodingError: Error {
    case invalidHexLength
}

extension Substring {
    mutating func popNextHexByte() throws -> UInt8? {
        while let next = self.first, next.isWhitespace {
            self = self.dropFirst()
        }
        
        guard let firstHex = self.popFirst() else {
            return nil
        }
        
        while let next = self.first, next.isWhitespace {
            self = self.dropFirst()
        }
        
        guard let secondHex = self.popFirst() else {
            throw HexDecodingError.invalidHexLength
        }
        
        return UInt8(String(firstHex) + String(secondHex), radix: 16)
    }
}

extension ByteBuffer {
    @discardableResult
    public mutating func writeHexEncodedBytes(_ hexString: String) throws -> Int {
        var slice = hexString[...]
        var written = 0
        while let nextByte = try slice.popNextHexByte() {
            written += self.writeInteger(nextByte)
        }
        return written
    }
}
```

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1878,6 +1889,29 @@ class ByteBufferTest: XCTestCase {
         XCTAssertEqual(expected, actual)
     }
 
+    func testWriteHexEncodedBytes() throws {
+        var buffer = ByteBuffer(hexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")
```
weissi, line 1893: This doesn't `throw` and is non-`Optional`, what happens if the input isn't a valid hex string? Examples

- `686`
- `686hello`
- `!.=/`
- `1 2 3 4 5 6 7 8` (this one could be argued is valid as `12 34 56 78`

@Lukasa WDYT is the right shape of the API here?

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -756,6 +769,18 @@ extension ByteBufferAllocator {
         return buffer
     }
 
+    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from the ASCII `hexEncodedBytes` string .
+    ///
+    /// This will allocate a new `ByteBuffer` with enough space to fit `bytes` and potentially some extra space.
+    ///
+    /// - returns: The `ByteBuffer` containing the written bytes.
+    @inlinable
+    public func buffer(hexEncodedBytes string: String) -> ByteBuffer {
```
weissi, line 778: This is an uncommon API shape in NIO, this should be spelled as an `init`ialiser, no?

Tests/NIOCoreTests/ByteBufferTest.swift
ali-ahsan-ali: Once my above comment about how to implement it is answered. I wil add throws for these edge cases.

Sources/NIOCore/ByteBuffer-aux.swift
ali-ahsan-ali: String, substring all have the same methods in ByteBufferAllocator. I think the initialiser you are looking for is below. 
I think this is following the pattern of the repo. 

Screenshot 2024-08-19 at 1 29 41 PM

Tests/NIOCoreTests/ByteBufferTest.swift
Lukasa: I suspect we should throw.

Sources/NIOCore/ByteBuffer-aux.swift
Lukasa: Yeah, this is the standard shape for ByteBufferAllocator. While I agree with you that in hindsight this was the _wrong_ API shape, it's where we are now and the consistency does matter.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali:
```swift 
func testWriteHexEncodedBytes() throws {
        var buffer = ByteBuffer(hexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")
        XCTAssertEqual(try buffer.writeHexEncodedBytes("68656c6c6f20776f726c64"), 11)
        XCTAssertEqual(try buffer.writeHexEncodedBytes("     0a    "), 1)
        XCTAssertEqual(try buffer.writeHexEncodedBytes(""), 0)
        XCTAssertEqual(try buffer.writeHexEncodedBytes("      "), 0)
        XCTAssertThrowsError(try buffer.writeHexEncodedBytes("    1  "))
        print(buffer.hexDump(format: .plain))
        XCTAssertEqual(ByteBuffer(string: "hello world\nhello world\n"), buffer)
    }
```

Testing hello world, it seems to get caught up on anything that goes through the below case. There seems to just be something wrong with that calculation. I will try understand the calculation more and solve it. 

Screenshot 2024-08-18 at 12 48 00 PM

ali-ahsan-ali: I founda  fix. Will push something using the implementation you suggested.

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -88,6 +88,24 @@ extension ByteBuffer {
...
+    /// - parameters:
+    ///     - plainHexEncodedBytes: The hex encoded string to write.
+    /// - returns: The number of bytes written.
+    @discardableResult
+    @inlinable
+    public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
+        var slice = plainHexEncodedBytes.utf8[...]
+        var written = 0
+
+        while let nextByte = try slice.popNextHexByte() {
```
ali-ahsan-ali, line 102: Will a throw here shut down the channel and mean the inconsistent state of the byte buffer with only half the values written is OK? 

OR Should I do an initial pass and validate BEFORE then going on and writing them to the buffer to ensure a consistent state? 

I did it initially the latter way, but poking around I assumed the channel closes and that bytebuffer is no longer used.

Tests/NIOCoreTests/ByteBufferTest.swift
ali-ahsan-ali: Can you let me know if it looks OK. I have made everything throw including the initialiser and the write method.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -263,3 +277,53 @@ extension ByteBuffer {
...
 }
+
+enum HexDecodingError: Error {
+    case invalidHexLength
+}
+
+extension UInt8 {
+    fileprivate var isASCIIWhitespace: Bool {
+        switch self {
+        case UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\n"), UInt8(ascii: "\r"), UInt8(ascii: " "):
```
ali-ahsan-ali, line 288: static let?

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1878,6 +1889,35 @@ class ByteBufferTest: XCTestCase {
...
+    func testWriteHexEncodedBytes() throws {
+        var buffer = try ByteBuffer(plainHexEncodedBytes: "68 65 6c 6c 6f 20 77 6f 72 6c 64 0a")
+        XCTAssertEqual(try buffer.writePlainHexEncodedBytes("68656c6c6f20776f726c64"), 11)
+        XCTAssertEqual(try buffer.writePlainHexEncodedBytes("     0a    "), 1)
+        XCTAssertEqual(try buffer.writePlainHexEncodedBytes(""), 0)
+        XCTAssertEqual(try buffer.writePlainHexEncodedBytes("      "), 0)
+        XCTAssertThrowsError(try buffer.writePlainHexEncodedBytes("    1  "))
+        XCTAssertThrowsError(try buffer.writePlainHexEncodedBytes("    1"))
+        XCTAssertThrowsError(try buffer.writePlainHexEncodedBytes("1       "))
+        // The first byte (68 = "h") is valid, the method throws and the valid byte IS written to the ByteBuffer
```
ali-ahsan-ali, line 1901: This feels wrong

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: @Lukasa

ali-ahsan-ali:
@weissi please take a look when you find the time

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -88,6 +88,30 @@ extension ByteBuffer {
         )
     }
 
+    // MARK: Hex encoded string APIs
+    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately.
+    /// - parameters:
+    ///     - plainHexEncodedBytes: The hex encoded string to write.
+    /// - returns: The number of bytes written.
```
Lukasa, line 95: Can this doc comment be updated to have an example of what this format looks like?

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -88,6 +88,30 @@ extension ByteBuffer {
...
+        var bytesToWrite: [UInt8] = []
+
+        // Single pass to ensure string is a valid representation of hex bytes
+        while let nextByte = try slice.popNextHexByte() {
+            bytesToWrite.append(nextByte)
+        }
+
+        for byte in bytesToWrite {
+            written += self.writeInteger(byte)
+        }
```
Lukasa, line 110: We don't need the intermediate array. Instead of using `bytesToWrite` and calling `append`, just call `writeInteger` directly.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -14,6 +14,20 @@
...
+    ///
+    /// This will allocate a new `ByteBuffer` with enough space to fit the hex decoded `bytes` and potentially some extra space
+    /// using the default allocator.
+    ///
+    /// - info: If you have access to a `Channel`, `ChannelHandlerContext`, or `ByteBufferAllocator` we
+    ///         recommend using `channel.allocator.buffer(integer:)`. Or if you want to write multiple items into the
+    ///         buffer use `channel.allocator.buffer(capacity: ...)` to allocate a `ByteBuffer` of the right
+    ///         size followed by a `writeHexEncodedBytes` instead of using this method. This allows SwiftNIO to do
+    ///         accounting and optimisations of resources acquired for operations on a given `Channel` in the future.
+    init(plainHexEncodedBytes string: String) throws {
```
Lukasa, line 27: This should be `public`.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -341,3 +355,54 @@ extension ByteBuffer {
         }
     }
 }
+
+enum HexDecodingError: Error {
```
Lukasa, line 359: Can we make this `public`, and make it a `struct`?

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -341,3 +355,54 @@ extension ByteBuffer {
...
+    case invalidHexLength
+    case invalidCharacter
+}
+
+extension UInt8 {
+    fileprivate var isASCIIWhitespace: Bool {
+        [UInt8(ascii: "\n"), UInt8(ascii: "\t"), UInt8(ascii: "\n"), UInt8(ascii: "\r"), UInt8(ascii: " ")].contains(
```
Lukasa, line 366: We don't need the duplicate.

```suggestion
        [UInt8(ascii: "\t"), UInt8(ascii: "\n"), UInt8(ascii: "\r"), UInt8(ascii: " ")].contains(
```

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -341,3 +355,54 @@ extension ByteBuffer {
...
+            self
+        )
+    }
+
+    fileprivate var asciiHexNibble: UInt8? {
```
Lukasa, line 371: Never returns nil.

```suggestion
    fileprivate var asciiHexNibble: UInt8 {
```

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -341,3 +355,54 @@ extension ByteBuffer {
...
+    mutating func popNextHexByte() throws -> UInt8? {
+        while let nextByte = self.first, nextByte.isASCIIWhitespace {
+            self = self.dropFirst()
+        }
+
+        guard let firstHex = try self.popFirst()?.asciiHexNibble else {
+            return nil  // No next byte to pop
+        }
+
+        while let nextByte = self.first, nextByte.isASCIIWhitespace {
```
Lukasa, line 398: I don't think we allow whitespace between the two hex characters.

Sources/NIOCore/ByteBuffer-hex.swift
weissi: Also needs to _either_

- be nested into `ByteBuffer`
- or have a `NIO` prefix as per the public API guarantees that we don't add top-level things that aren't `NIO`-prefixed

Sources/NIOCore/ByteBuffer-aux.swift
ali-ahsan-ali: XCTAssertThrowsError(try buffer.writePlainHexEncodedBytes("68 1"))

If I write it the way you have suggested, this method will throw, BUT, the 68 will be written to the byte. Would that not leave the byte in an inconsistent or unintended state potentially leading to some confusion?

Lukasa: We can save the original writer index and restore it when we throw the error.

ali-ahsan-ali: Thanks, will do

Sources/NIOCore/ByteBuffer-aux.swift
ali-ahsan-ali: LMK if latest push meets expectation.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -340,4 +354,55 @@ extension ByteBuffer {
             }
         }
     }
+
+    public struct HexDecodingError: Error {
+        let kind: HexDecodingErrorKind
+
+        public enum HexDecodingErrorKind {
```
Lukasa, line 361: This `enum` needs to be private, or we can't safely use this type.

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -340,4 +354,55 @@ extension ByteBuffer {
             }
         }
     }
+
+    public struct HexDecodingError: Error {
```
Lukasa, line 358: Can you add a docstring here?

Lukasa (commented):
Ok awesome, this is getting a _lot_ closer. Only a few little notes here.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Are you suggesting internal? Private would mean this would be impossible (to my understanding) to initialise (atleast the way I have done so).

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -91,6 +91,29 @@ extension ByteBuffer {
...
+    @inlinable
+    public mutating func writePlainHexEncodedBytes(_ plainHexEncodedBytes: String) throws -> Int {
+        var slice = plainHexEncodedBytes.utf8[...]
+        let initialWriterIndex = self._writerIndex
+
+        do {
+            while let nextByte = try slice.popNextHexByte() {
+                self.writeInteger(nextByte)
+            }
+            return Int(self._writerIndex - initialWriterIndex)
```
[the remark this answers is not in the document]
ali-ahsan-ali: Easy.

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -91,6 +91,29 @@ extension ByteBuffer {
...
+        } catch {
+            self._moveWriterIndex(to: initialWriterIndex)
```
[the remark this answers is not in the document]
ali-ahsan-ali: Easy.

Sources/NIOCore/ByteBuffer-hex.swift
ali-ahsan-ali: Updated. Take a look when you are free.

Sources/NIOCore/ByteBuffer-hex.swift
Lukasa: Private should be fine, you'd use `public static let`s on the parent to access the fields:

```swift
public struct HexDecodingError: Error {
    private let kind: Kind

    enum Kind {
        case invalidHexLength
        case invalidCharacter
    }
    
    public static let invalidHexLength = HexDecodingError(kind: .invalidHexLength)
    public static let invalidCharacter = HexDecodingError(kind: .invalidCharacter)
}
```

ali-ahsan-ali: thanks will find time to implement.

Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -91,6 +91,29 @@ extension ByteBuffer {
         )
     }
 
+    // MARK: Hex encoded string APIs
+    /// Write `string` into this `ByteBuffer` in ASCII hexadecimal, moving the writer index forward appropriately. This method will throw if the string input is not of the "plain" hex encoded format.
```
Lukasa, line 95:
```suggestion
    /// Write ASCII hexadecimal `string` into this `ByteBuffer` as raw bytes, decoding the hexadecimal & moving the writer index forward appropriately. This method will throw if the string input is not of the "plain" hex encoded format.
```

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -14,6 +14,20 @@
 
 extension ByteBuffer {
 
+    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from  the string representation of `plainHexEncodedBytes`.
```
Lukasa, line 17:
```suggestion
    /// Create a fresh `ByteBuffer` containing the `bytes` decoded from the string representation of `plainHexEncodedBytes`.
```

Lukasa (commented):
Fantastic, this is looking really good. A pair of very minor textual tweaks and then I'm happy. Thanks for sticking with this!

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -340,4 +354,58 @@ extension ByteBuffer {
             }
         }
     }
+
+    /// An error that is thrown when an invalid hex encoded string was attempted to be written to a ByteBuffer.
+    public struct HexDecodingError: Error {
+        let kind: HexDecodingErrorKind
+
+        enum HexDecodingErrorKind {
+            /// The hex encoded string was not of the expected even length.
```
ali-ahsan-ali, line 363: Make that change here that I never made

ali-ahsan-ali:
> Fantastic, this is looking really good. A pair of very minor textual tweaks and then I'm happy. Thanks for sticking with this!

Doing a bit of travelling. Actually flying to London for the swift conference. Will find time to push the changes.

Lukasa (approved):
Awesome, I'm happy with this! Well done getting through the gauntlet again @ali-ahsan-ali, it's really good that you were able to stick with it. Much appreciated ✨ .

PR #2838, opened 2024-08-12
Adopt strict concurrency in _NIOBase64

Lukasa:
Motivation:

Another brick in the wall.

Modifications:

- Make Base64 an enum, not a struct, as it's a namespace.
- Make Base64 Sendable, which it trivially is.

Result:

Another pass, another success

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -34,13 +34,13 @@ let strictConcurrencySettings: [SwiftSetting] = {
         .enableUpcomingFeature("InferSendableFromCaptures"),
     ])
 
-#if compiler(>=6.0)
+    #if compiler(>=6.0)
     if ProcessInfo.processInfo.environment["CI"] != nil {
         // -warnings-as-errors here is a workaround so that IDE-based development can
         // get tripped up on -require-explicit-sendable.
         initialSettings.append(.unsafeFlags(["-require-explicit-sendable", "-warnings-as-errors"]))
     }
-#endif
+    #endif
 
     return initialSettings
 }()
--- a/Package.swift
+++ b/Package.swift
@@ -27,6 +27,8 @@ let swiftSystem: PackageDescription.Target.Dependency = .product(
     condition: .when(platforms: [.macOS, .iOS, .tvOS, .watchOS, .linux, .android])
 )
 
+let strictConcurrencyDevelopment = false
+
 let strictConcurrencySettings: [SwiftSetting] = {
     var initialSettings: [SwiftSetting] = []
     initialSettings.append(contentsOf: [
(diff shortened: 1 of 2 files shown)


glbrntt (approved):

Package.swift
```diff
@@ -23,16 +27,21 @@ let swiftSystem: PackageDescription.Target.Dependency = .product(
...
+let strictConcurrencySettings: [SwiftSetting] = {
+    var initialSettings: [SwiftSetting] = []
+    initialSettings.append(contentsOf: [
+        .enableUpcomingFeature("StrictConcurrency"),
+        .enableUpcomingFeature("InferSendableFromCaptures"),
+    ])
+
+#if compiler(>=6.0)
+    if ProcessInfo.processInfo.environment["CI"] != nil {
+        initialSettings.append(.unsafeFlags(["-require-explicit-sendable", "-warnings-as-errors"]))
```
FranzBusch, line 39: Why are we passing `warnings-as-errors` here? We normally pass this in the CI invocations

Lukasa: For at-desk development, is the answer.

Sources/_NIOBase64/Base64.swift
```diff
@@ -35,7 +35,7 @@ public enum Base64Error: Error {
 }
 
 @usableFromInline
-internal struct Base64 {
+internal enum Base64: Sendable {
```
FranzBusch, line 38: Do we need this to be `Sendable` even if it is just a namespace?

Package.swift
Lukasa: Specifically, `-require-explicit-sendable` doesn't appear to work in CI, so this dance gives us the opportunity to have the IDE itself police this if you launch it appropriately.

Sources/_NIOBase64/Base64.swift
Lukasa: Yes. Because it's `@usableFromInline`, `-require-explicit-sendable` applies to it, and that flag does not notice that you can't actually form a value to it.

Lukasa: More broadly, `-require-explicit-sendable` forces our hand for everything. "Need" doesn't really come into it: we have to mark anything it wants us to mark.

Package.swift
FranzBusch: Can we add this as a comment so we hopefully can remove this again in the future.

FranzBusch (approved):

Package.swift
Lukasa: Done.

PR #2848, opened 2024-08-19
[CI] Mark the workspace as safe for the matrix job

FranzBusch:
# Motivation

Some scripts that run as part of a matrix job may want to call git commands and the workspace needs to be marked as safe.

# Modification

This PR adds a step to mark the workspace as safe


simonjbeaumont (approved):
Happy for you to merge this as-is. Added a question about whether this could be tighter scoped, but it's not blocking.

PR #2849, opened 2024-08-20
Add new ChannelOption to get the amount of buffered outbound data in the Channel

johnnzhou:
Add ability to get the amount of buffered outbound data from `Channel`

### Motivation:
Right now, SwiftNIO does not have the API to answer the question "how much data is buffered in the Channel".  Applications focusing on performance may need to fine-tune the amount of outbound data that will be sent to optimize data throughput, adjust sending rate to avoid overflow, and potentially reduce latency.

SwiftNIO currently provides some backpressure mechanism. This new API will be a good addition. By knowing how much data is buffered directly, applications can make informed decision to adjust for optimal buffer sizes and send rates. 

### Modifications:
- Expose current buffer size through ChannelOptions so that users can read the value out. StreamSocketChannel,  DatagramSocketChannel, EmbeddedChannel, and AsyncTestingChannel have the same API interface.
- Various modifications to the existing tests to make sure the new API is working correctly.
- Add a new `so_sndbuf` socket option so that users can easily adjust the send buffer size.

### Result:

Users can get the amount of outbound bytes currently buffered in the `Channel` through the new `BufferedWritableBytesOption` channel option.


Sources/NIOCore/ChannelOption.swift
```diff
@@ -287,6 +287,13 @@ extension ChannelOptions {
             public typealias Value = Bool
             public init() {}
         }
+        
+        /// `BufferedWritableBytesOption` allows users to know the number of writable bytes currently buffered in the `Channel`.
+        public struct BufferedWritableBytesOption: ChannelOption, Sendable {
+            public typealias Value = Int64
```
Lukasa, line 293: The preference in Swift is to use `Int` at API boundaries such as this. Given that a buffer size of more than 2GB on a 32-bit system is going to cause A Bad Time anyway, we can safely make that change here.

Lukasa (changes requested):
Thanks for this, it generally looks good! Can you add support for the new channel option in `EmbeddedChannel` and `AsyncTestingChannel` as well? It should only behave sensibly when the buffered data is ByteBuffer, so runtime type assertions there are fine.

While we're here, it's worth calling out that this is an incomplete solution, as there may be `ChannelHandler`s that have also buffered some bytes. This is a very useful building block, but we may want to revisit this space to add more fully-featured functionality for this.

johnnzhou:
Hi @Lukasa, thanks for bringing it up. I'm planning to follow our discussion [here](https://forums.swift.org/t/getting-the-bufferedamount-in-the-channel-for-swift-nio-websocket/70423) to add a new protocol that audits the outbound buffered bytes. Anyone who's interested in is also welcome to contribute :). I will fix the comments ASAP.

johnnzhou:
Sure. Actually I'm working on the tests - should be ready soon. Thanks for the reminder!

weissi:
Super cool, thanks @johnnzhou !!

johnnzhou:
Hi @Lukasa, whenever you are available, could you take a look at the code change. Appreciate your time and help in advance!

johnnzhou:
Hi @Lukasa @weissi, thanks for the help refining my PR. Whenever you have a moment, could you please review the change? Your feedback is greatly appreciated! Thanks for your time!

Lukasa (approved):
Very nice patch @johnnzhou, thank you so much and sorry for the delay.

johnnzhou:
Thanks @Lukasa for the review! I synced the PR with the main. It requires maintainer approval to kick off the pipeline. If you are available, can you help rerun the pipeline? Thanks!

johnnzhou:
Hi @Lukasa,

Thank you for consistently monitoring this PR. I noticed that the latest pipeline failed on the formatting check again - my apologies. I’ve updated the PR based on the pipeline results. Whenever you have a moment, could you please review it and run the workflow again? I really appreciate your time and help!

Thanks so much!

Lukasa (approved):
No problem!

Issue #2700, opened 2024-04-11
Remove swift-docc-plugin dependency


weissi:
The `swift-docc-plugin` dependency isn't needed and makes everybody's builds slower as it gets pulled. Swift Package Index auto-adds the dependency.

weissi:
Fixed in #2851

PR #2854, opened 2024-08-28
Pin DocC to below 1.4

Lukasa:
Motivation:

Unfortunately, the DocC plugin no longer builds on 5.8: https://github.com/swiftlang/swift-docc-plugin/issues/94. We need to keep building.

Modifications:

- Forcefully pin the DocC plugin down.

Result:

CI works now.


simonjbeaumont (approved):

weissi:
@Lukasa I don't think we need the DocC dependency at all: #2700

FranzBusch:
@weissi We have removed the docc dependency in https://github.com/apple/swift-nio/pull/2851

Issue #2825, opened 2024-08-05
ByteBuffer should have a `.hexdump(format: .compact)`


weissi:
ByteBuffer has awesome `hexdump(format: .detailed | .plain)` formats (from #2475). They're awesome and I use them all the time. But there's one format that I think is quite useful: A compact one without spaces. So

```
XCTAssertEqual("68656c6c6f20776f726c640a", ByteBuffer(string: "hello world\n").hexdump(format: .compact))
```

Why is compact so important? Frequently, UNIX tools output formats that are easy to split using `cut -d' ' -f2` or `awk '{ print $2 }'` or `while read -r line prefix hex; do echo "$hex"; done`. But all these require that there's no whitespace in the hexdump.

So I think we should add another format which is exactly like `.plain` but doesn't emit spaces or any other whitespace. #2475 made great groundwork here and this should be very easy to add.

weissi:
Related to #2748

supersonicbyte:
Made a [PR ](https://github.com/apple/swift-nio/pull/2856) for this!

natikgadzhi:
> ByteBuffer has awesome hexdump(format: .detailed | .plain) formats (from https://github.com/apple/swift-nio/pull/2475). They're awesome and I use them all the time.

🥹 ❤️‍🩹

weissi:
This issue is actually done! Thanks @natikgadzhi & @supersonicbyte

PR #2856, opened 2024-09-01
Add .compact format to ByteBuffer's hexdump method

supersonicbyte:
Motivation:

Resolving the following issue: https://github.com/apple/swift-nio/issues/2825.

Modifications:

• Added a compact format for ByteBuffer's hexdump method

Result:

A new format which is analog to the `.plain` but without whitespaces.

### Motivation:

### Modifications:

### Result:


Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -85,6 +95,46 @@ extension ByteBuffer {
...
+
+        // Safe to force-unwrap because we just checked readableBytes is > maxBytes above.
+        let front = buffer.readSlice(length: maxBytes / 2)!
+        buffer.moveReaderIndex(to: buffer.writerIndex - maxBytes / 2)
+        let back = buffer.readSlice(length: buffer.readableBytes)!
+
+        let startHex = front.hexDumpCompact()
+        let endHex = back.hexDumpCompact()
+        return startHex + " ... " + endHex
+    }
```
FranzBusch, line 137: Both these methods are almost identical to the `plain` ones. Can we refactor them to reuse the code instead?

FranzBusch (commented):
Thanks for working on this. Great start!

Sources/NIOCore/ByteBuffer-hexdump.swift
supersonicbyte: Ofcourse! Should we factor out common code to a separate function and make the `hexDumpPlain` and `hexDumpCompact` call that function?

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -85,6 +95,46 @@ extension ByteBuffer {
...
         return startHex + " ... " + endHex
     }
+    
+    /// Return a `String` of  hexadecimal digits of the readable bytes in the buffer,
+    /// analog to `.plain` format but without whitespaces.
+    /// `hexDumpCompact()` always dumps all readable bytes, i.e. from `readerIndex` to `writerIndex`,
+    /// so you should set those indices to desired location to get the offset and length that you need to dump.
+    private func hexDumpCompact() -> String {
+        var hexString = ""
+        hexString.reserveCapacity(self.readableBytes * 3)
```
simonjbeaumont, line 105: Drive by comment. Is this meant to be `3`? 
```suggestion
        hexString.reserveCapacity(self.readableBytes * 2)
```

simonjbeaumont, line 105:
```suggestion
        hexString.reserveCapacity(self.readableBytes * 2)
```

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -85,6 +95,46 @@ extension ByteBuffer {
...
+            hexString += String(byte, radix: 16, padding: 2)
+        }
+
+        return hexString
+    }
+
+    /// Return a `String` of  hexadecimal digits of the readable bytes in the buffer,
+    /// analog to `.plain` format but without whitespaces and clips the output to the max length of `maxBytes` bytes.
+    /// If the dump contains more than the `maxBytes` bytes, this function will return the first `maxBytes/2`
+    /// and the last `maxBytes/2` of that, replacing the rest with `...`, i.e. `01 02 03 ... 09 11 12`.
```
simonjbeaumont, line 117: Should this example have whitespace?

simonjbeaumont (commented):
Thanks for working on this ❤️  I just noticed a couple of things in passing where we're not updating the code or comments to account for the compact (lack of whitespace).

Sources/NIOCore/ByteBuffer-hexdump.swift
supersonicbyte: It should not have whitespaces, good catch! I will fix that. Thanks!

Sources/NIOCore/ByteBuffer-hexdump.swift
supersonicbyte: Oh, yes, since we are not adding whitespaces we can reserve fewer capacity. Thanks!

supersonicbyte:
Updated to resolve PR review comments. Feel free to review again when you find the time! Thanks.

FranzBusch (approved):
LGTM!

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -1904,6 +1904,33 @@ class ByteBufferTest: XCTestCase {
...
+        XCTAssertEqual("6c6c6f", buf.hexDump(format: .compact))
+    }
+
+    func testHexDumpCompactWithMaxBytes() {
+        self.buf.clear()
+        for f in UInt8.min...UInt8.max {
+            self.buf.writeInteger(f)
+        }
+        let actual = self.buf.hexDump(format: .compact(maxBytes: 10))
+        let expected = "0001020304 ... fbfcfdfeff"
```
simonjbeaumont, line 1931: This is a completely subjective, style question here for the output but... this variant is _compact_ and documents itself as being "like plain, but without the whitespace". With that in mind, how do we feel about the whitespace around the `...`, when `readableBytes > maxBytes`? Should we have: 

1. `0001020304 ... fbfcfdfeff`; or
2. `0001020304...fbfcfdfeff`?

I'm just parking this here for discussion, so no need to action this right away @supersonicbyte, but given it's API we might like to make a deliberate decision on it.

//cc @weissi @FranzBusch

supersonicbyte: Anyways, just let me know for what option we will go so I can fix it accordingly. Thanks!

weissi: I think no whitespace would be great indeed. The initial reason I asked for this was for a command line too that outputs:

```
key-hex-bytes value-hex-bytes
```

and it was important that the `...-hex-bytes` could be read with

```
while read -r key value; do
   ...
done
```

so having whitespace in either key or value would break that code. Similarly, if you would use `cut -d' ' -f1` or so you wouldn't want any whitespace there.

We should probably also add to the docs that this format guarantees to never emit whitespace.

Thank you!!

supersonicbyte: Alright, as per @weissi answer I went ahead and implemented the option 2. Feel free to review and give feedback. Thanks!

Sources/NIOCore/ByteBuffer-hexdump.swift
```diff
@@ -128,7 +136,7 @@ extension ByteBuffer {
     }
 
     /// Return a `String` of  hexadecimal digits of the readable bytes in the buffer,
-    /// analog to `.plain` format but without whitespaces.
+    /// analog to `.plain` format but without whitespaces. This format guarantees not to emmit whitespaces.
```
weissi, line 139:
```suggestion
    /// analog to `.plain` format but without whitespaces. This format guarantees not to emit whitespaces.
```

weissi (approved):
Thanks very much!

weissi:
@FranzBusch I'm happy with this, do you want to auto-merge it when it's green?

simonjbeaumont (approved):

Tests/NIOCoreTests/ByteBufferTest.swift
MahdiBM: Will also be slightly nicer for all the `...`s to instead be `…`. The former is 3 unicodes (3 `.`s), the latter is the 1 unicode character, the "U+2026 HORIZONTAL ELLIPSIS".

FranzBusch (approved):

FranzBusch:
Can you fix up the failed formatting run please @supersonicbyte

supersonicbyte:
> Can you fix up the failed formatting run please @supersonicbyte

Ofcourse!

PR #2858, opened 2024-09-02
[CI] Exclude disabled jobs

FranzBusch:
# Motivation

Currently disabled matrix jobs are still shown as skipped.

# Modification

This PR excludes any disabled job which should make them not show up at all.

# Result

No more unneeded skipped jobs.


FranzBusch:
This sadly doesn't work since you can't reference variables of an object in the matrix to exclude.

PR #2859, opened 2024-09-02
Add functions for reading and writing quic variable-length encoded integers

hamzahrmalik:
### Motivation:

QUIC, as well as other protocols such as OHTTP, use variable-length integers. These are integers with their length encoded into the first few bits.

RFC: https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc

### Modifications:

Add functions for reading and writing these integers:
- `ByteBuffer.readQUICVariableLengthInteger`
- `ByteBuffer.writeQUICVariableLengthInteger`
- `ByteBuffer.writeQUICLengthPrefixed(ByteBuffer)`
- `ByteBuffer.writeQUICLengthPrefixed((inout ByterBuffer) -> Int)`

### Result:

Users can easily interact with variable length integers

### DIFF
--- a/Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
@@ -1,0 +1,171 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if os(Windows)
+import ucrt
+#elseif canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
+#else
+#error("The Byte Buffer module was unable to identify your C library.")
+#endif
+
+extension ByteBuffer {
+    /// Read a QUIC variable-length integer, moving the `readerIndex` appropriately
+    /// If there are not enough bytes, nil is returned
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    public mutating func readQUICVariableLengthInteger() -> Int? {
+        guard let firstByte = self.getInteger(at: self.readerIndex, as: UInt8.self) else {
+            return nil
+        }
+
+        // Look at the first two bits to work out the length, then read that, mask off the top two bits, and
+        // extend to integer.
+        switch firstByte & 0xC0 {
+        case 0x00:
+            // Easy case.
+            self.moveReaderIndex(forwardBy: 1)
+            return Int(firstByte & ~0xC0)
+        case 0x40:
+            // Length is two bytes long, read the next one.
+            return self.readInteger(as: UInt16.self).map { Int($0 & ~(0xC0 << 8)) }
+        case 0x80:
+            // Length is 4 bytes long.
+            return self.readInteger(as: UInt32.self).map { Int($0 & ~(0xC0 << 24)) }
+        case 0xC0:
+            // Length is 8 bytes long.
+            return self.readInteger(as: UInt64.self).map { Int($0 & ~(0xC0 << 56)) }
+        default:
+            fatalError("Unreachable")
+        }
+    }
+
+    /// Write a QUIC variable-length integer.
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    @discardableResult
+    public mutating func writeQUICVariableLengthInteger(_ value: Int) -> Int {
+        self.writeQUICVariableLengthInteger(UInt(value))
+    }
+
+    /// Write a QUIC variable-length integer.
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    ///
+    /// - Returns: The number of bytes written
+    @discardableResult
+    public mutating func writeQUICVariableLengthInteger(_ value: UInt) -> Int {
+        switch value {
+        case 0..<63:
+            // Easy, store the value. The top two bits are 0 so we don't need to do any masking.
+            return self.writeInteger(UInt8(truncatingIfNeeded: value))
+        case 0..<16383:
+            // Set the top two bit mask, then write the value.
+            let value = UInt16(truncatingIfNeeded: value) | (0x40 << 8)
+            return self.writeInteger(value)
+        case 0..<1_073_741_823:
+            // Set the top two bit mask, then write the value.
+            let value = UInt32(truncatingIfNeeded: value) | (0x80 << 24)
+            return self.writeInteger(value)
+        case 0..<4_611_686_018_427_387_903:
+            // Set the top two bit mask, then write the value.
+            let value = UInt64(truncatingIfNeeded: value) | (0xC0 << 56)
+            return self.writeInteger(value)
+        default:
+            fatalError("Could not write QUIC variable-length integer: outside of valid range")
+        }
+    }
+
+    /// Write a QUIC variable-length integer prefixed buffer.
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    ///
+    /// Write `buffer` prefixed with the length of buffer as a QUIC variable-length integer
+    /// - Parameter buffer: The buffer to be written
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the buffer itself
+    @discardableResult
+    public mutating func writeQUICLengthPrefixed(_ buffer: ByteBuffer) -> Int {
+        var written = 0
+        written += self.writeQUICVariableLengthInteger(buffer.readableBytes)
+        written += self.writeImmutableBuffer(buffer)
+        return written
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
...
+
+    /// Write a QUIC variable-length integer prefixed buffer.
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    ///
+    /// Write `buffer` prefixed with the length of buffer as a QUIC variable-length integer
+    /// - Parameter buffer: The buffer to be written
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the buffer itself
+    @discardableResult
+    public mutating func writeQUICLengthPrefixed(_ buffer: ByteBuffer) -> Int {
```
Lukasa, line 104: We should strongly consider adding variants of this for `bytes` and `string` as we do with non-length-prefixed options.

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
```

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
...
+        default:
+            fatalError("Unreachable")
+        }
+    }
+
+    /// Write a QUIC variable-length integer.
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    @discardableResult
+    public mutating func writeQUICVariableLengthInteger(_ value: Int) -> Int {
```
Lukasa, line 64: Any reason this shouldn't be generic over `FixedWidthInteger`? It would avoid the awkward need to include UInt in our API surface.

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
...
+        totalBytesWritten += self.writeBytes([0, 0, 0, 0])
+
+        let startWriterIndex = self.writerIndex
+        let messageLength = try writeMessage(&self)
+        let endWriterIndex = self.writerIndex
+
+        totalBytesWritten += messageLength
+
+        let actualBytesWritten = endWriterIndex - startWriterIndex
+        assert(
```
Lukasa, line 138: `precondition` please

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
...
+            "writeMessage returned \(messageLength) bytes, but actually \(actualBytesWritten) bytes were written, but they should be the same"
+        )
+
+        if messageLength >= 1_073_741_823 {
+            // This message length cannot fit in the 4 bytes which we reserved. We need to make more space
+            // Reserve 4 more bytes
+            totalBytesWritten += self.writeBytes([0, 0, 0, 0])
+            // Move the message forward by 4 bytes
+            self.withUnsafeMutableReadableBytes { pointer in
+                _ = memmove(
```
Lukasa, line 149: Rather than bury this in here, please factor this out into a `fileprivate` helper method.

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,171 @@
...
+                    // The new position is 4 forward from where the user written message currently begins
+                    pointer.baseAddress!.advanced(by: startWriterIndex + 4),
+                    // This is the position where the user written message currently begins
+                    pointer.baseAddress?.advanced(by: startWriterIndex),
```
Lukasa, line 153: Don't mix a force-unwrap and an optional unwrap. Force unwrap is fine in both cases here.

Tests/NIOCoreTests/ByteBufferQUICLengthPrefixTests.swift
```diff
@@ -0,0 +1,173 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2021 Apple Inc. and the SwiftNIO project authors
```
Lukasa, line 5:
```suggestion
// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
```

Tests/NIOCoreTests/ByteBufferQUICLengthPrefixTests.swift
```diff
@@ -0,0 +1,173 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import XCTest
+
+final class ByteBufferQUICLengthPrefixTests: XCTestCase {
+    private var buffer = ByteBuffer()
```
Lukasa, line 19: Did making this global save us any code?

hamzahrmalik: good point, probably not
I just shamelessly stole this from `ByteBufferLengthPrefixTests.swift`

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
hamzahrmalik: We could but it will need to be internal if we want to keep inlineable

Tests/NIOCoreTests/ByteBufferQUICLengthPrefixTests.swift
hamzahrmalik: so technically it does save us one line per test
But im happy to repeat that line in every test if you think that's better

One advantage would be that a badly written test wouldn't fail other tests

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
Lukasa: That's fine, leading underscore works great too.

Tests/NIOCoreTests/ByteBufferQUICLengthPrefixTests.swift
Lukasa: It doesn't even save us that much, as several tests immediately shadow it with a local variable. I think it's not worth the cost.

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,238 @@
...
+            return self.writeInteger(UInt8(truncatingIfNeeded: value))
+        case 0..<16383:
+            // Set the top two bit mask, then write the value.
+            let value = UInt16(truncatingIfNeeded: value) | (0x40 << 8)
+            return self.writeInteger(value)
+        case 0..<1_073_741_823:
+            // Set the top two bit mask, then write the value.
+            let value = UInt32(truncatingIfNeeded: value) | (0x80 << 24)
+            return self.writeInteger(value)
+        case 0..<4_611_686_018_427_387_903:
```
Lukasa, line 79: I don't think this works with smaller integer sizes. Can you write a test that tries to call this with a `UInt8`, validates that it doesn't work, and then fix it?

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,238 @@
...
+extension ByteBuffer {
+    /// Creates `requiredSpace` bytes of free space before `index` by moving the `length` bytes after `index` forward by `requiredSpace`
+    /// e.g. given [a, b, c, d, e, f, g, h, i, j] and calling this function with (before: 4, length: 6, requiredSpace: 2) would result in
+    /// [a, b, c, d, 0, 0, e, f, g, h, i, j]
+    /// 2 extra bytes of space were created before index 4 (the letter e)
+    /// The total bytes written will be equal to `requiredSpace`, and the writer index will be moved accordingly
+    @usableFromInline
+    mutating func _createSpace(before index: Int, length: Int, requiredSpace: Int) {
+        // Add the required number of bytes to the end first
+        self.writeBytes([UInt8](repeatElement(0, count: requiredSpace)))
```
Lukasa, line 226: Better to force this to 4 and write the integer than to make the allocation here.

Sources/NIOCore/ByteBuffer-QUICLengthPrefix.swift
```diff
@@ -0,0 +1,238 @@
...
+#else
+#error("The Byte Buffer module was unable to identify your C library.")
+#endif
+
+extension ByteBuffer {
+    /// Read a QUIC variable-length integer, moving the `readerIndex` appropriately
+    /// If there are not enough bytes, nil is returned
+    ///
+    /// [RFC 9000 § 16](https://www.rfc-editor.org/rfc/rfc9000.html#section-16)
+    public mutating func readQUICVariableLengthInteger() -> Int? {
```
weissi, line 34: What about using a `format` or something parameter to allow implementing a bunch of variable-length integer formats. For example `buffer.readVariableLengthInteger(format: .quic)` or `format: .protobuf` or so?

weissi: Maybe maybe that's overkill too, just an idea.

Lukasa: This seems like a really lovely extension point idea.

Issue #2828, opened 2024-08-06
WebSocketFrame should have a nice description


weissi:
### Expected behavior

`WebSocketFrame.description` outputs something nice.

### Actual behavior

`WebSocketFrame.description` outputs something that's not nice at all

```
WebSocketFrame(_storage: NIOWebSocket.WebSocketFrame.(unknown context at $1084150f8)._Storage, maskKey: nil, firstByte: 138)
```

### Steps to reproduce

- print any `WebSocketFrame`

supersonicbyte:
Made a [PR ](https://github.com/apple/swift-nio/pull/2862) for this.

ali-ahsan-ali:
Can this be closed?

PR #2862, opened 2024-09-03
Adding a nicer description for `WebSocketFrame`.

supersonicbyte:
###Motivation:

Resolving the following issue: https://github.com/apple/swift-nio/issues/2828

###Modifications:

Making `WebSocketFrame` conform to `CustomStringConvertible`.

###Result:

A nicer description for `WebSocketFrame`.

### DIFF
--- a/Sources/NIOWebSocket/WebSocketFrame.swift
+++ b/Sources/NIOWebSocket/WebSocketFrame.swift
@@ -363,3 +363,42 @@ extension WebSocketFrame._Storage: Equatable {
         lhs.data == rhs.data && lhs.extensionData == rhs.extensionData
     }
 }
+
+extension WebSocketFrame: CustomStringConvertible {
+    /// A `String` describing this `WebSocketFrame`. Example:
+    ///
+    ///     WebSocketFrame {
+    ///     maskKey: Optional(NIOWebSocket.WebSocketMaskingKey(_key: (187, 28, 185, 79))),
+    ///     fin: true,
+    ///     rsv1: true,
+    ///     rsv2: true,
+    ///     rsv3: true,
+    ///     opcode: WebSocketOpcode.binary,
+    ///     length: 0,
+    ///     data: ByteBuffer { readerIndex: 0, writerIndex: 0, readableBytes: 0, capacity: 0, storageCapacity: 0, slice: _ByteBufferSlice { 0..<0 }, storage: 0x00006000028246b0 (0 bytes) },
+    ///     extensionData: nil,
+    ///     unmaskedData: ByteBuffer { readerIndex: 0, writerIndex: 0, readableBytes: 0, capacity: 0, storageCapacity: 0, slice: _ByteBufferSlice { 0..<0 }, storage: 0x0000600002824800 (0 bytes) },
+    ///     unmaskedDataExtension: nil
+    ///     }
+    ///
+    /// The format of the description is not API.
+    ///
+    /// - returns: A description of this `WebSocketFrame`.
+    public var description: String {
+        """
+        WebSocketFrame {
+        maskKey: \(String(describing: self.maskKey)),
+        fin: \(self.fin),
+        rsv1: \(self.rsv1),
+        rsv2: \(self.rsv2),
+        rsv3: \(self.rsv3),
+        opcode: \(self.opcode),
+        length: \(self.length),
+        data: \(String(describing: self.data)),
+        extensionData: \(String(describing: self.extensionData)),
+        unmaskedData: \(String(describing: self.unmaskedData)),
+        unmaskedDataExtension: \(String(describing: self.unmaskedExtensionData))
+        }
+        """
+    }
+}


weissi:
Thank you so much! @Lukasa to take a look. LGTM

weissi (approved):
Looks good to me, thank you! Could do with a tiny unit test though

supersonicbyte:
> > > Looks good to me, thank you! Could do with a tiny unit test though
> > 
> > 
> > Ofcourse, should I make a new test file or put it in an existing one? Thanks for the review.
> 
> Completely up to you. In case you're interested what I would do: Given that it's such a tiny thing I'd reuse an existing file because that's slightly less work :)

Added! Thanks.

Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -363,3 +363,42 @@ extension WebSocketFrame._Storage: Equatable {
...
+    ///     data: ByteBuffer { readerIndex: 0, writerIndex: 0, readableBytes: 0, capacity: 0, storageCapacity: 0, slice: _ByteBufferSlice { 0..<0 }, storage: 0x00006000028246b0 (0 bytes) },
+    ///     extensionData: nil,
+    ///     unmaskedData: ByteBuffer { readerIndex: 0, writerIndex: 0, readableBytes: 0, capacity: 0, storageCapacity: 0, slice: _ByteBufferSlice { 0..<0 }, storage: 0x0000600002824800 (0 bytes) },
+    ///     unmaskedDataExtension: nil
+    ///     }
+    ///
+    /// The format of the description is not API.
+    ///
+    /// - returns: A description of this `WebSocketFrame`.
+    public var description: String {
```
FranzBusch, line 387: Can we please avoid adding a multiline string as a description here. `description` is often used in other interpolated strings such as log messages and adding line breaks can often break those formats. Instead can we add a single line for this.

Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -363,3 +363,42 @@ extension WebSocketFrame._Storage: Equatable {
...
+        """
+        WebSocketFrame {
```
FranzBusch, line 389: NIT: Can we use `(` instead of `{` to align it with other types.

Sources/NIOWebSocket/WebSocketFrame.swift
weissi: good point, we should add `\`s to the end of the line (such that the result is just one line). I missed that currently we'll actually print multiple lines.

supersonicbyte: @FranzBusch very good point, thanks! I will fix it.

Sources/NIOWebSocket/WebSocketFrame.swift
supersonicbyte: ofcourse!

supersonicbyte:
Updated. @FranzBusch please take a look when you find the time. Also I noticed that `ByteBuffer's` description uses curly braces `{}` so we might wanna convert those to `()` so we are unified with other types. Thanks.

Sources/NIOWebSocket/WebSocketFrame.swift
FranzBusch: This is still a multiline string from what I can see. Can you use `"` instead of `"""` unless I am missing something here

weissi: @FranzBusch no it's not, it has `\`s everywhere but the last line.

weissi:
```
  1> print(""" 
  2.     hello \ 
  3.     my \ 
  4.     beautiful \ 
  5.     world 
  6.     """)
hello my beautiful world
```

FranzBusch:
> Updated. @FranzBusch please take a look when you find the time. Also I noticed that `ByteBuffer's` description uses curly braces `{}` so we might wanna convert those to `()` so we are unified with other types. Thanks.

Actually thinking about this again. We should probably drop the `WebsocketFrame` altogether for `description`. Normally types don't print their type name in description. The only difference then between `description` and `debugDescription` is wrapping it in `()` like `String` works e.g.

```
print(frame.description) // produces masKey: foo, fin: true
print(frame.debugDescription) // produces (masKey: foo, fin: true)
```

dnadoba:
>Normally types don't print their type name in description.

By default it does e.g.:
```swift
struct WebSocketFrame {}
print(WebSocketFrame()) // WebSocketFrame()
debugPrint(WebSocketFrame()) // PlaygroundExecutable.WebSocketFrame()
```
With that being said, I'm not opposed to remove the name but we would diverge from the default.

supersonicbyte:
I dropped the names as @FranzBusch suggested. If needed otherwise I can add them back.

Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -363,3 +363,60 @@ extension WebSocketFrame._Storage: Equatable {
...
+        unmaskedDataExtension: \(String(describing: self.unmaskedExtensionData))
+        """
+    }
+}
+
+extension WebSocketFrame: CustomDebugStringConvertible {
+    public var debugDescription: String {
+        """
+        ( \
+        maskKey: \(String(describing: self.maskKey)), \
```
FranzBusch, line 408: You can just interpolate `self.description` here instead of duplicating the code. Can we also add a test for the `debugDescription` please

supersonicbyte: Yes, done!

FranzBusch (approved):

Issue #2863, opened 2024-09-03
Make ByteBuffer.description useful and pretty


weissi:
Right now, `ByteBuffer`'s `description` is ugly and not very useful (doesn't print the bytes). With #2856 being merged, I'm proposing to make `ByteBuffer`'s description the following:

```
[ hex bytes, maximum 64 or something ](length bytes)
```

so `hello world` would be

```
[68656c6c6f20776f726c64](11 bytes)
```

and 256 0 bytes would be

```
[00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000...00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000](256 bytes)
```

the empty string would be `[](0 bytes)`

FranzBusch:
Onboard with this idea. The current description is not really helpful in most cases and I always do a hexDump manually while debugging. We should also change the `debugDescription` of `ByteBuffer` while we are it.

weissi:
> Onboard with this idea. The current description is not really helpful in most cases and I always do a hexDump manually while debugging. We should also change the `debugDescription` of `ByteBuffer` while we are it.

_Normally_ I'd say that we shouldn't change `debugDescription` (because that actually has the bytes and when debugging a ByteBuffer the other stuff is interesting too). _However_, if you do `Array<ByteBuffer>.description` then that will invoke the element's `debugDescription`!?... So yeah, I guess we should just make `.description == .debugDescription`

FranzBusch:
Yes `CustromStringConvertible` and `CustomDebugStringConvertible` are a bit messy. The latter is used by collection types to print their contents. So you should always implement both. At least that's my latest understanding @dnadoba dove into this a bit a while ago.

supersonicbyte:
I can pick this one up also 👋

supersonicbyte:
@weissi since @FranzBusch mentioned [here](https://github.com/apple/swift-nio/pull/2862#discussion_r1741892580 ) that `()` are preferred, should the output of `ByteBuffer's` description be wrapped with `()` or should we stick with `[]`? IMO, `[]` feels more natural for `ByteBuffer`.

dnadoba:
[...]

PR #2864, opened 2024-09-03
Making ByteBuffer's description more useful

supersonicbyte:
### Motivation:

Resolving the following issue: https://github.com/apple/swift-nio/issues/2863

### Modifications:

Rewrote `description` and `debugDescription` on `ByteBuffer` to make it more useful. Also fixed a bug in the in `hexDumpCompact`.

### Result:

A more useful `description` and `debugDescription`.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -957,8 +957,9 @@ public struct ByteBuffer {
 }
 
 extension ByteBuffer: CustomStringConvertible, CustomDebugStringConvertible {
-    /// A `String` describing this `ByteBuffer`.  For a `ByteBuffer` initialised with `hello world`
-    /// the description would be the following:
+    /// A `String` describing this `ByteBuffer` including length and the bytes it contains (partially).
+    ///
+    /// For a `ByteBuffer` initialised with `hello world` the description would be the following:
     ///
     ///     [68656c6c6f20776f726c64](11 bytes)
     ///
@@ -970,8 +971,9 @@ extension ByteBuffer: CustomStringConvertible, CustomDebugStringConvertible {
         "[\(self.hexDump(format: .compact(maxBytes: 64)))](\(self.readableBytes) bytes)"
     }
 
-    /// A `String` describing this `ByteBuffer`.  For a `ByteBuffer` initialised with `hello world`
-    /// the description would be the following:
+    /// A `String` describing this `ByteBuffer` including length and the bytes it contains (partially).
+    ///
+    /// For a `ByteBuffer` initialised with `hello world` the description would be the following:
     ///
     ///     [68656c6c6f20776f726c64](11 bytes)
     ///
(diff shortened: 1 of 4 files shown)


weissi (approved):
LGTM thank you. I'll let the NIO team actually review this before they can merge.

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -957,37 +957,30 @@ public struct ByteBuffer {
 }
 
 extension ByteBuffer: CustomStringConvertible, CustomDebugStringConvertible {
-    /// A `String` describing this `ByteBuffer`. Example:
+    /// A `String` describing this `ByteBuffer`.  For a `ByteBuffer` initialised with `hello world`
```
weissi, line 960:
```suggestion
    /// A `String` describing this `ByteBuffer` including length and the bytes it contains (partially).
    ///
    /// For a `ByteBuffer` initialised with `hello world`
```

Nit: The first line in the doc comments is special, it's the summary. That summary gets displayed by itself in many scenarios so we should aim for it to make sense without the rest of the doc comment.

Tests/NIOCoreTests/NIOAnyDebugTest.swift
```diff
@@ -24,7 +24,7 @@ class NIOAnyDebugTest: XCTestCase {
         let bb = ByteBuffer(string: "byte buffer string")
         XCTAssertTrue(
             wrappedInNIOAnyBlock(bb).contains(
-                "NIOAny { ByteBuffer { readerIndex: 0, writerIndex: 18, readableBytes: 18, capacity: 32, storageCapacity: 32, slice: _ByteBufferSlice { 0..<32 }, storage: "
+                "NIOAny { [627974652062756666657220737472696e67](18 bytes) }"
```
weissi, line 27: Nit: I almost think with this change we should change NIOAny's `description` to include `ByteBuffer: ` when it's the `case .byteBuffer`. Previously that wasn't necessary because `ByteBuffer.description` contained the word `ByteBuffer` but that's now changed.

Tests/NIOCoreTests/ByteBufferTest.swift
```diff
@@ -3602,5 +3607,12 @@ extension ByteBufferTest {
             XCTAssertEqual(error as? Base64Error, .invalidCharacter)
         }
     }
-
+    
+    func testByteBufferDescription() {
```
weissi, line 3611: think it would be nice to add two other test cases:

- the empty ByteBuffer
- one where the hex bytes get truncated (i.e. more than 128 bytes or whatever the limit is)

Tests/NIOCoreTests/NIOAnyDebugTest.swift
supersonicbyte: Looking at `NIOAny` I would improve its description altogether to include information about the wrapped type. But I would do that in a separate issue/PR.

Tests/NIOCoreTests/ByteBufferTest.swift
supersonicbyte: Thanks, added!

FranzBusch (approved):
This LGTM. @supersonicbyte please raise another PR for adapting `NIOAny` then

PR #2865, opened 2024-09-03
Make `assumeIsolated` work with SerialExecutors that are backed by EventLoops

fabianfett:
Allow usage of `assumeIsolated` in SerialExecutors that are backed by EventLoops.

### Motivation:

We want to support all new Swift 6 features in SerialExecutors.

### Modifications:

- Implement `isSameExclusiveExecutionContext`
- Implement `checkIsolated`

### Result:

We can use `assumeIsolated` in actors that use an `EventLoop` as their `SerialExecutor`.

### DIFF
--- a/Sources/NIOCore/EventLoop+SerialExecutor.swift
+++ b/Sources/NIOCore/EventLoop+SerialExecutor.swift
@@ -50,7 +50,7 @@ extension NIOSerialEventLoopExecutor {
         other === self
     }
 
-    @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+    @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, *)
     @inlinable
     public func checkIsolated() {
         self.preconditionInEventLoop()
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):

Sources/NIOCore/EventLoop+SerialExecutor.swift
```diff
@@ -43,6 +43,18 @@ extension NIOSerialEventLoopExecutor {
...
         self
     }
+
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    @inlinable
+    public func isSameExclusiveExecutionContext(other: Self) -> Bool {
+        other === self
+    }
+
+    @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
```
FranzBusch, line 53: Should fix the compilation failures
```suggestion
    @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, *)
```

Sources/NIOCore/EventLoop+SerialExecutor.swift
```diff
@@ -43,6 +43,18 @@ extension NIOSerialEventLoopExecutor {
     public var executor: any SerialExecutor {
         self
     }
+
+    @available(macOS 14.0, iOS 17.0, watchOS 10.0, tvOS 17.0, *)
+    @inlinable
+    public func isSameExclusiveExecutionContext(other: Self) -> Bool {
```
ktoso, line 49: This is not going to be called yet.

You must adjust `asUnownedSerialExecutor` to return `UnownedSerialExecutor` with the marker flag (there's a `complexEquality:` initializer.

PR #2867, opened 2024-09-04
Add functions for reading and writing length-prefixed data with customizable encodings for the length

hamzahrmalik:
Add functions for reading and writing length-prefixed data with customizable encodings for the length, particularly for quic variable-length integers (RFC 9000)

### Motivation:

Many protocols require us to write data and then prefix that data with its length. But each protocol has a different way of encoding the length. This PR introduces general purpose functions which can be extended for different encoding strategies

### Modifications:

Create a new protocol which defines how to encode an integer
Implement this protocol for QUIC

Provide functions on bytebuffer for writing length-prefixed buffers, strings or bytes

### DIFF
--- a/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
@@ -98,7 +98,7 @@ extension ByteBuffer {
     }
 
     /// Prefixes bytes written by `writeData` with the number of bytes written.
-    /// The number of bytes written is encoded usng `strategy`
+    /// The number of bytes written is encoded using `strategy`
     ///
     /// - Note: This function works by reserving the number of bytes suggested by `strategy` before the data.
     /// It then writes the data, and then goes back to write the length.
--- a/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
@@ -79,8 +79,11 @@ extension ByteBuffer {
     /// Read a binary encoded integer, moving the `readerIndex` appropriately.
     /// If there are not enough bytes, nil is returned.
     @inlinable
-    public mutating func readEncodedInteger<Strategy: NIOBinaryIntegerEncodingStrategy>(_ strategy: Strategy) -> Int? {
-        strategy.readInteger(as: Int.self, from: &self)
+    public mutating func readEncodedInteger<Strategy: NIOBinaryIntegerEncodingStrategy, Integer: FixedWidthInteger>(
+        _ strategy: Strategy,
+        as: Integer.Type = Integer.self
+    ) -> Integer? {
+        strategy.readInteger(as: Integer.self, from: &self)
     }
 
     /// Write a binary encoded integer.
--- a/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 /// Describes a way to encode and decode an integer as bytes
+/// For more information, see <doc:ByteBuffer-lengthPrefix>
 ///
 public protocol NIOBinaryIntegerEncodingStrategy {
     /// Read an integer from a buffer.
@@ -44,10 +45,10 @@ public protocol NIOBinaryIntegerEncodingStrategy {
     var requiredBytesHint: Int { get }
 
     /// Write an integer to a buffer. Move the writer index to after the written integer.
-    /// Call this function if you have already reserved some capacity for an integer to be written.
-    /// Implementors should consider using a less efficient encoding, if possible,to fit exactly within the reserved capacity.
-    /// Otherwise, callers may need to shift bytes to reconcile the difference.
-    /// It is up to the implementor to find the balance between performance and size.
+    /// This function will be called when an integer needs to be written, and some capacity has already been reserved for it.
+    /// Implementers should consider using a less efficient encoding, if possible,to fit exactly within the reserved capacity.
+    /// Otherwise, the caller will need to shift bytes to reconcile the difference.
+    /// It is up to the implementer to find the balance between performance and size.
     /// - Parameters:
     ///   - integer: The integer to write
     ///   - reservedCapacity: The capacity already reserved for writing this integer
--- a/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
@@ -1,0 +1,313 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if os(Windows)
+import ucrt
+#elseif canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
+#else
+#error("The Byte Buffer module was unable to identify your C library.")
+#endif
+
+/// Describes a way to encode and decode an integer as bytes
+///
+public protocol BinaryIntegerEncodingStrategy {
+    /// Read an integer from a buffer. The reader index should be moved accordingly
+    /// - Returns: The integer
+    func readInteger<IntegerType: FixedWidthInteger>(
+        as: IntegerType.Type,
+        from buffer: inout ByteBuffer
+    ) -> IntegerType?
+
+    /// Write an integer to a buffer. The writer index should be moved accordingly
+    func writeInteger<IntegerType: FixedWidthInteger>(
+        _ integer: IntegerType,
+        to buffer: inout ByteBuffer
+    ) -> Int
+
+    /// When writing an integer using this strategy, how many bytes we should reserve
+    /// If the actual bytes used by the write function is more or less than this, it will be necessary to shuffle bytes
+    /// Therefore, an accurate prediction here will improve performance
+    var reservedSpaceForInteger: Int { get }
+
+    /// Write an integer to a buffer. The writer index should be moved accordingly
+    /// This function takes  a `reservedSpace` parameter which is the number of bytes we already reserved for writing this integer
+    /// If you use write more or less than this many bytes, we wll need to move bytes around to make that possible
+    /// Therefore, you may decide to encode differently to use more bytes than you actually would need, if it is possible for you to do so
+    /// That way, you waste the reserved bytes, but improve writing performance
+    func writeIntegerWithReservedSpace(
+        _ integer: Int,
+        reservedSpace: Int,
+        to buffer: inout ByteBuffer
+    ) -> Int
+}
+
+extension BinaryIntegerEncodingStrategy {
+    @inlinable
+    public var reservedSpaceForInteger: Int { 1 }
+
+    @inlinable
+    public func writeIntegerWithReservedSpace<IntegerType: FixedWidthInteger>(
+        _ integer: IntegerType,
+        reservedSpace: Int,
+        to buffer: inout ByteBuffer
+    ) -> Int {
+        self.writeInteger(integer, to: &buffer)
+    }
+}
+
+extension ByteBuffer {
+    /// Read a binary encoded integer, moving the `readerIndex` appropriately
+    /// If there are not enough bytes, nil is returned
+    @inlinable
+    public mutating func readEncodedInteger<Strategy: BinaryIntegerEncodingStrategy>(_ strategy: Strategy) -> Int? {
+        strategy.readInteger(as: Int.self, from: &self)
+    }
+
+    /// Write a binary encoded integer.
+    ///
+    /// - Returns: The number of bytes written
+    @discardableResult
+    @inlinable
+    public mutating func writeEncodedInteger<
+        Integer: FixedWidthInteger,
+        Strategy: BinaryIntegerEncodingStrategy
+    >(
+        _ value: Integer,
+        strategy: Strategy
+    ) -> Int {
+        strategy.writeInteger(value, to: &self)
+    }
+
+    /// Prefixes a message written by `writeMessage` with the number of bytes written using `strategy`
+    ///
+    /// Implementation note: This function works by reserving the number of bytes suggested by ``BinaryIntegerEncodingStrategy.reservedSpace`` before the data
+    /// It then writes the data, and then goes back to write the length
+    /// If the reserved capacity turns out to be too little or too much, then the data will be shifted
+    /// Therefore, this function is most performant if the strategy is able to use the same number of bytes that it reserved
+    ///
+    /// - Parameters:
+    ///     - strategy: The strategy to use for encoding the length
+    ///     - writeMessage: A closure that takes a buffer, writes a message to it and returns the number of bytes written
+    /// - Returns: Number of total bytes written. This is the length of the written message + the number of bytes used to write the length before it
+    @discardableResult
+    @inlinable
+    public mutating func writeLengthPrefixed<Strategy: BinaryIntegerEncodingStrategy>(
+        strategy: Strategy,
+        writeMessage: (inout ByteBuffer) throws -> Int
+    ) rethrows -> Int {
+        /// The index at which we write the length
+        let lengthPrefixIndex = self.writerIndex
+        /// The space which we reserve for writing the length
+        let reservedSpace = strategy.reservedSpaceForInteger
+        self.writeRepeatingByte(0, count: reservedSpace)
+
+        /// The index at which we start writing the message originally. We may later move the message if the reserved space for the length wasn't right
+        let originalMessageStartIndex = self.writerIndex
+        /// The length of the message written
+        let messageLength: Int
+        do {
+            messageLength = try writeMessage(&self)
+        } catch {
+            // Clean up our write so that it as if we never did it.
+            self.moveWriterIndex(to: lengthPrefixIndex)
+            throw error
+        }
+        /// The index at the end of the written message originally. We may later move the message if the reserved space for the length wasn't right
+        let originalMessageEndIndex = self.writerIndex
+
+        // Quick check to make sure the user didn't do something silly
+        precondition(
+            originalMessageEndIndex - originalMessageStartIndex == messageLength,
+            "writeMessage returned \(messageLength) bytes, but actually \(originalMessageEndIndex - originalMessageStartIndex) bytes were written. They should be the same"
+        )
+
+        // We write the length after the message to begin with. We will move it later
+
+        /// The actual number of bytes used to write the length written. The user may write more or fewer bytes than what we reserved
+        let actualIntegerLength = strategy.writeIntegerWithReservedSpace(
+            messageLength,
+            reservedSpace: reservedSpace,
+            to: &self
+        )
+
+        switch actualIntegerLength {
+        case reservedSpace:
+            // Good, exact match, swap the values and then "delete" the trailing bytes by moving the index back
+            self._moveBytes(from: originalMessageEndIndex, to: lengthPrefixIndex, size: actualIntegerLength)
+            self.moveWriterIndex(to: originalMessageEndIndex)
+        case ..<reservedSpace:
+            // We wrote fewer bytes. We now have to move the length bytes from the end, and
+            // _then_ shrink the rest of the buffer onto it.
+            self._moveBytes(from: originalMessageEndIndex, to: lengthPrefixIndex, size: actualIntegerLength)
+            let newMessageStartIndex = lengthPrefixIndex + actualIntegerLength
+            self._moveBytes(
+                from: originalMessageStartIndex,
+                to: newMessageStartIndex,
+                size: messageLength
+            )
+            self.moveWriterIndex(to: newMessageStartIndex + messageLength)
+        case reservedSpace...:
+            // We wrote more bytes. We now have to create enough space. Once we do, we have the same
+            // implementation as the matching case.
+            let extraSpaceNeeded = actualIntegerLength - reservedSpace
+            self._createSpace(before: lengthPrefixIndex, requiredSpace: extraSpaceNeeded)
+
+            // Clean up the indices.
+            let newMessageEndIndex = originalMessageEndIndex + extraSpaceNeeded
+            // We wrote the length after the message, so we have to move those bytes to the space at the front
+            self._moveBytes(from: newMessageEndIndex, to: lengthPrefixIndex, size: actualIntegerLength)
+            self.moveWriterIndex(to: newMessageEndIndex)
+        default:
+            fatalError("Unreachable")
+        }
+
+        let totalBytesWritten = self.writerIndex - lengthPrefixIndex
+        return totalBytesWritten
+    }
+
+    /// Reads a slice which is prefixed with a length. The length will be read using `strategy`, and then that many bytes will be read to create a slice
+    /// - Returns: The slice, if there are enough bytes to read it fully. In this case, the readerIndex will move to after the slice
+    /// If there are not enough bytes to read the full slice, the readerIndex will stay unchanged
+    public mutating func readLengthPrefixedSlice<Strategy: BinaryIntegerEncodingStrategy>(
+        _ strategy: Strategy
+    ) -> ByteBuffer? {
+        let originalReaderIndex = self.readerIndex
+        guard let length = strategy.readInteger(as: Int.self, from: &self), let slice = self.readSlice(length: length)
+        else {
+            self.moveReaderIndex(to: originalReaderIndex)
+            return nil
+        }
+        return slice
+    }
+}
+
+// MARK: - Helpers for writing length-prefixed things
+
+extension ByteBuffer {
+    /// Write the length of `buffer` using `strategy`. Then write the buffer
+    /// - Parameters:
+    ///   - buffer: The buffer to be written
+    ///   - strategy: The encoding strategy to use
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the buffer itself
+    @discardableResult
+    public mutating func writeLengthPrefixedBuffer<
+        Strategy: BinaryIntegerEncodingStrategy
+    >(
+        _ buffer: ByteBuffer,
+        strategy: Strategy
+    ) -> Int {
+        var written = 0
+        written += self.writeEncodedInteger(buffer.readableBytes, strategy: strategy)
+        written += self.writeImmutableBuffer(buffer)
+        return written
+    }
+
+    /// Write the length of `string` using `strategy`. Then write the string
+    /// - Parameters:
+    ///  - string: The string to be written
+    ///  - strategy: The encoding strategy to use
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the string itself
+    @discardableResult
+    public mutating func writeLengthPrefixedString<
+        Strategy: BinaryIntegerEncodingStrategy
+    >(
+        _ string: String,
+        strategy: Strategy
+    ) -> Int {
+        var written = 0
+        // writeString always writes the String as UTF8 bytes, without a null-terminator
+        // So the length will be the utf8 count
+        written += self.writeEncodedInteger(string.utf8.count, strategy: strategy)
+        written += self.writeString(string)
+        return written
+    }
+
+    /// Write the length of `bytes` using `strategy`. Then write the bytes
+    /// - Parameters:
+    ///  - bytes: The bytes to be written
+    ///  - strategy: The encoding strategy to use
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the bytes themselves
+    @inlinable
+    public mutating func writeLengthPrefixedBytes<
+        Bytes: Sequence,
+        Strategy: BinaryIntegerEncodingStrategy
+    >(
+        _ bytes: Bytes,
+        strategy: Strategy
+    ) -> Int
+    where Bytes.Element == UInt8 {
+        let numberOfBytes = bytes.withContiguousStorageIfAvailable { b in
+            UnsafeRawBufferPointer(b).count
+        }
+        if let numberOfBytes {
+            var written = 0
+            written += self.writeEncodedInteger(numberOfBytes, strategy: strategy)
+            written += self.writeBytes(bytes)
+            return written
+        } else {
+            return self.writeLengthPrefixed(strategy: strategy) { buffer in
+                buffer.writeBytes(bytes)
+            }
+        }
+    }
+}
+
+extension ByteBuffer {
+    /// Creates `requiredSpace` bytes of free space immediately before `index`
+    /// e.g. given [a, b, c, d, e, f, g, h, i, j] and calling this function with (before: 4, requiredSpace: 2) would result in
+    /// [a, b, c, d, 0, 0, e, f, g, h, i, j]
+    /// 2 extra bytes of space were created before index 4 (the letter e)
+    /// The total bytes written will be equal to `requiredSpace`, and the writer index will be moved accordingly
+    @usableFromInline
+    mutating func _createSpace(before index: Int, requiredSpace: Int) {
+        let bytesToMove = self.writerIndex - index
+
+        // Add the required number of bytes to the end first
+        self.writeRepeatingByte(0, count: requiredSpace)
+        // Move the message forward by that many bytes, to make space at the front
+        self.withVeryUnsafeMutableBytes { pointer in
+            _ = memmove(
+                // The new position is forward from where the user written message currently begins
+                pointer.baseAddress!.advanced(by: index + requiredSpace),
+                // This is the position where the user written message currently begins
+                pointer.baseAddress!.advanced(by: index),
+                bytesToMove
+            )
+        }
+    }
+
+    /// Move the `size` bytes starting from `source` to `destination`
+    /// `source` and `destination` must both be within the writable range
+    @usableFromInline
+    mutating func _moveBytes(from source: Int, to destination: Int, size: Int) {
+        precondition(source >= self.readerIndex && destination < self.writerIndex && source >= destination)
+        precondition(source + size <= self.writerIndex)
+
+        // The precondition above makes this safe: our indices are in the valid range, so we can safely use them here
+        self.withVeryUnsafeMutableBytes { pointer in
+            _ = memmove(
+                pointer.baseAddress!.advanced(by: destination),
+                pointer.baseAddress!.advanced(by: source),
+                size
+            )
+        }
+    }
+}
--- a/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
@@ -12,20 +12,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if os(Windows)
-import ucrt
-#elseif canImport(Darwin)
-import Darwin
-#elseif canImport(Glibc)
-import Glibc
-#elseif canImport(Musl)
-import Musl
-#elseif canImport(Bionic)
-import Bionic
-#else
-#error("The Byte Buffer module was unable to identify your C library.")
-#endif
-
 /// Describes a way to encode and decode an integer as bytes
 ///
 public protocol NIOBinaryIntegerEncodingStrategy {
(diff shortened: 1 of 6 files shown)


Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,313 @@
...
+    /// If the actual bytes used by the write function is more or less than this, it will be necessary to shuffle bytes
+    /// Therefore, an accurate prediction here will improve performance
+    var reservedSpaceForInteger: Int { get }
+
+    /// Write an integer to a buffer. The writer index should be moved accordingly
+    /// This function takes  a `reservedSpace` parameter which is the number of bytes we already reserved for writing this integer
+    /// If you use write more or less than this many bytes, we wll need to move bytes around to make that possible
+    /// Therefore, you may decide to encode differently to use more bytes than you actually would need, if it is possible for you to do so
+    /// That way, you waste the reserved bytes, but improve writing performance
+    func writeIntegerWithReservedSpace(
```
glbrntt, line 55: This API 'stutters' a little bit: could we use `reservedCapacity` to overload `writeInteger`? i.e.

```swift
func writeInteger(_ value: Int, to buffer: inout ByteBuffer, reservedCapacity: Int) -> Int
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,313 @@
...
+        from buffer: inout ByteBuffer
+    ) -> IntegerType?
+
+    /// Write an integer to a buffer. The writer index should be moved accordingly
+    func writeInteger<IntegerType: FixedWidthInteger>(
+        _ integer: IntegerType,
+        to buffer: inout ByteBuffer
+    ) -> Int
+
+    /// When writing an integer using this strategy, how many bytes we should reserve
```
glbrntt, line 45: I found this and `writeIntegerWithReservedSpace` quite unintuitive without looking at how they were used. It became clearer when I started reading "Integer" as "Length". Is that the right way to think about it?

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
```diff
@@ -0,0 +1,135 @@
...
+        @inlinable
+        public func writeInteger<IntegerType: FixedWidthInteger>(
+            _ integer: IntegerType,
+            to buffer: inout ByteBuffer
+        ) -> Int {
+            self.writeIntegerWithReservedSpace(integer, reservedSpace: 0, to: &buffer)
+        }
+
+        @inlinable
+        public var reservedSpaceForInteger: Int { 4 }
```
glbrntt, line 90: Should this be `self.reservedSpace`?

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,313 @@
...
+    /// If the actual bytes used by the write function is more or less than this, it will be necessary to shuffle bytes
+    /// Therefore, an accurate prediction here will improve performance
```
glbrntt, line 47: General comment which applies to pretty much all of your docs: you need to use full stops. Because it's effectively markdown newlines will be concatenated when rendered:

Screenshot 2024-09-05 at 09 47 52

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,313 @@
...
+    >(
+        _ value: Integer,
+        strategy: Strategy
+    ) -> Int {
+        strategy.writeInteger(value, to: &self)
+    }
+
+    /// Prefixes a message written by `writeMessage` with the number of bytes written using `strategy`
+    ///
+    /// Implementation note: This function works by reserving the number of bytes suggested by ``BinaryIntegerEncodingStrategy.reservedSpace`` before the data
```
glbrntt, line 101: nit: You can use `- Note:`

glbrntt (changes requested):
This is neat! I've not reviewed it all but left some initial feedback and questions to make sure I understood how it all fits together.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
hamzahrmalik: Yeah i agree this is confusing

The idea here is that `reservedSpaceForInteger` is the amount of space that someone should reserve if theyre intending to write an integer using your strategy
And the `writeIntegerWithReservedSpace` is for when someone has already reserved some space, and now wants you to write an integer into that space. You can then make the decision on whether you want to try and use exactly that much space (this is what the quic impl does) even if you could've used less. Ie you choose whether you optimize for the smallest possible encoding of the int, or for trying to avoid moving bytes around

This is in practice currently only used by the writePrefixed* functions so yes, there the integer is a length. But i don't think it _has_ to be a length

glbrntt: It _can_ be used to write an integer, but in the context of this protocol _should_ it be used for anything other than the length? 

IMO we should make the intent of the requirement clear, if you wanted to write an integer then you should just call the `writeInteger(_:to:)` method.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+    /// If the actual bytes used by the write function is more or less than this, it will be necessary to shuffle bytes.
+    /// Therefore, an accurate prediction here will improve performance.
+    var reservedCapacityForInteger: Int { get }
+
+    /// Write an integer to a buffer. The writer index should be moved accordingly.
+    /// This function takes  a `reservedCapacity` parameter which is the number of bytes we already reserved for writing this integer.
+    /// If you use write more or less than this many bytes, we wll need to move bytes around to make that possible.
+    /// Therefore, you may decide to encode differently to use more bytes than you actually would need, if it is possible for you to do so.
+    /// That way, you waste the reserved bytes, but improve writing performance.
+    func writeIntegerWithReservedCapacity(
```
glbrntt, line 63: nit: repeating "reserved" in the name and in the parameter name. Is that necessary?

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+    ///
+    /// - Parameters:
+    ///     - strategy: The strategy to use for encoding the length.
+    ///     - writeMessage: A closure that takes a buffer, writes a message to it and returns the number of bytes written.
+    /// - Returns: Number of total bytes written. This is the length of the written message + the number of bytes used to write the length before it.
+    @discardableResult
+    @inlinable
+    public mutating func writeLengthPrefixed<Strategy: NIOBinaryIntegerEncodingStrategy>(
+        strategy: Strategy,
+        writeMessage: (inout ByteBuffer) throws -> Int
```
glbrntt, line 122: nit: name closure parameters (per the [API design guidelines](https://www.swift.org/documentation/api-design-guidelines/#special-instructions)) as they act as documentation

```suggestion
        writeMessage: (_ buffer: inout ByteBuffer) throws -> Int
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+        Integer: FixedWidthInteger,
+        Strategy: NIOBinaryIntegerEncodingStrategy
+    >(
+        _ value: Integer,
+        strategy: Strategy
+    ) -> Int {
+        strategy.writeInteger(value, to: &self)
+    }
+
+    /// Prefixes a message written by `writeMessage` with the number of bytes written using `strategy`.
```
glbrntt, line 107: naming nit: this isn't necessarily a 'message' being written, 'bytes' is a little more generic

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+            self.moveWriterIndex(to: lengthPrefixIndex)
+            throw error
+        }
+        /// The index at the end of the written message originally. We may later move the message if the reserved space for the length wasn't right
+        let originalMessageEndIndex = self.writerIndex
+
+        // Quick check to make sure the user didn't do something silly
+        precondition(
+            originalMessageEndIndex - originalMessageStartIndex == messageLength,
+            "writeMessage returned \(messageLength) bytes, but actually \(originalMessageEndIndex - originalMessageStartIndex) bytes were written. They should be the same"
```
glbrntt, line 147:
```suggestion
            "writeMessage returned \(messageLength) bytes, but actually \(originalMessageEndIndex - originalMessageStartIndex) bytes were written. They must be the same."
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        let totalBytesWritten = self.writerIndex - lengthPrefixIndex
+        return totalBytesWritten
+    }
+
+    /// Reads a slice which is prefixed with a length. The length will be read using `strategy`, and then that many bytes will be read to create a slice.
+    /// - Returns: The slice, if there are enough bytes to read it fully. In this case, the readerIndex will move to after the slice.
+    /// If there are not enough bytes to read the full slice, the readerIndex will stay unchanged.
+    public mutating func readLengthPrefixedSlice<Strategy: NIOBinaryIntegerEncodingStrategy>(
```
glbrntt, line 197: This should be `@inlinable`

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+// MARK: - Helpers for writing length-prefixed things
+
+extension ByteBuffer {
+    /// Write the length of `buffer` using `strategy`. Then write the buffer.
+    /// - Parameters:
+    ///   - buffer: The buffer to be written.
+    ///   - strategy: The encoding strategy to use.
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the buffer itself.
+    @discardableResult
+    public mutating func writeLengthPrefixedBuffer<
```
glbrntt, line 219: this one also needs an `@inlinable`

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+        Strategy: NIOBinaryIntegerEncodingStrategy
+    >(
+        _ buffer: ByteBuffer,
+        strategy: Strategy
+    ) -> Int {
+        var written = 0
```
glbrntt, line 225: It might be worth reserving enough capacity ahead of time here (and for the `String` version too)

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,321 @@
...
+        written += self.writeImmutableBuffer(buffer)
+        return written
+    }
+
+    /// Write the length of `string` using `strategy`. Then write the string.
+    /// - Parameters:
+    ///  - string: The string to be written.
+    ///  - strategy: The encoding strategy to use.
+    /// - Returns: The total bytes written. This is the bytes needed to write the length, plus the length of the string itself.
+    @discardableResult
```
glbrntt, line 236: `@inlinable`

glbrntt (changes requested):
Left a few more comments. The bulk of this looks great but still unsure around some of the naming.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
hamzahrmalik: we don't know how many bytes writeEncodedInteger will need

We could make use of strategy.reservedCapacity here, but it might be wrong

glbrntt: Being wrong by a couple of bytes should be fine because we round up to the next power of two anyway. Reserving capacity would help us to avoid situations where we allocate a couple of bytes for the length then immediately reallocate because we need more space to write the buffer after it.

hamzahrmalik: ok, done

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,337 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if os(Windows)
+import ucrt
```
weissi, line 16: @hamzahrmalik why do we need the libcs here?

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,337 @@
...
+#elseif canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
+#else
+#error("The Byte Buffer module was unable to identify your C library.")
```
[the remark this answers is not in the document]
MaxDesiatov: Yes, thanks for catching this!

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,337 @@
...
+#if os(Windows)
+import ucrt
+#elseif canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
```
MaxDesiatov, line 24:
```suggestion
import Bionic
#elseif canImport(WASILibc)
import WASILibc
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,337 @@
...
+    /// 2 extra bytes of space were created before index 4 (the letter e).
+    /// The total bytes written will be equal to `requiredSpace`, and the writer index will be moved accordingly.
+    @usableFromInline
+    mutating func _createSpace(before index: Int, requiredSpace: Int) {
+        let bytesToMove = self.writerIndex - index
+
+        // Add the required number of bytes to the end first
+        self.writeRepeatingByte(0, count: requiredSpace)
+        // Move the data forward by that many bytes, to make space at the front
+        self.withVeryUnsafeMutableBytes { pointer in
```
weissi, line 309: could we avoid `*Unsafe*` here?

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
hamzahrmalik: for the memmove function

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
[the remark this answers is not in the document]
hamzahrmalik: If we need this, then it's missing across all of NIO

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
Lukasa: Not really: this is the safe wrapper around the access.

Ultimately this is implementing `memmove` on `ByteBuffer`. Implementing it safely is somewhere between very hard and impossible using the APIs we currently have. We can run through the options:

1. Bytewise copy using `writeInteger`. This is going to be very costly for large moves, which is the concern we have in this method.
2. Manually-autovectorised copy using `writeMultipleInteger` and wider integers. This can work, but it's still going to be _very_ slow compared to `memmove` and it will be much more complex code as we have to handle a wide range of edge-cases.
3. Move using `writeImmutableBuffer`. This will, unfortunately, trigger a CoW.
4. Move using `ByteBufferView`. This will also trigger a CoW.

I fundamentally can't see an implementation here that performs well enough to be used on larger buffers, which is also the circumstance in which it is most likely that we'll need a move. But I'm open to other options.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
weissi: Okay, in this case we should remove the imports here.

- we shouldn't use `*Unsafe*` here, the most minimal subset of `ByteBuffer-core.swift` needs to provide enough
- `memmove` isn't necessary to be called directly. `UnsafeRawBufferPointer` can move memory with a higher-level API (I think it's `bufferPointer[0..<100] = bufferPointer[2..<102]` or something like that; but again, no Unsafe in here please)

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
weissi: Okay, if this is necessary functionality, can we add it to `ByteBuffer-core.swift` and use in other places too? If my memory serves me right, then @glbrntt added some memory-moving API already a few years back, might be wrong about that though

glbrntt: Close enough, I added `copyBytes(at:to:length:)` 5 years ago as part of making `ByteBufferView` mutable. Good memory!

 https://github.com/apple/swift-nio/blame/282f5935cf3352b3d026c35eb57cb3619dd9536f/Sources/NIOCore/ByteBuffer-core.swift#L1077

weissi: @hamzahrmalik isn't that exactly what we need? `memmove` == copying bytes too

Lukasa: Yeah, this looks like we could use it here.

hamzahrmalik: Yep, this works perfectly

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// Describes a way to encode and decode an integer as bytes
+///
+public protocol NIOBinaryIntegerEncodingStrategy {
```
Lukasa, line 17: We can x-ref the narrative docs you wrote from here.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
Lukasa: George's note here is good, we can probably remove the note in the name.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// - Parameters:
+    ///    - integer: The type of the integer to write.
+    ///    - buffer: The buffer to write to.
+    /// - Returns: The number of bytes used to write the integer.
+    func writeInteger<IntegerType: FixedWidthInteger>(
+        _ integer: IntegerType,
+        to buffer: inout ByteBuffer
+    ) -> Int
+
+    /// An estimate of the bytes required to write integers using ths strategy.
```
Lukasa, line 40:
```suggestion
    /// An estimate of the bytes required to write integers using this strategy.
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// Callers may use this to reserve bytes before writing the integer.
+    /// If the actual bytes used by the write function is more or less than this, it may be necessary to shuffle bytes.
+    /// Therefore, an accurate prediction here will improve performance.
+    var requiredBytesHint: Int { get }
+
+    /// Write an integer to a buffer. Move the writer index to after the written integer.
+    /// Call this function if you have already reserved some capacity for an integer to be written.
```
Lukasa, line 47: The tone of this sentence seems wrong: it's an instruction to the caller, but the caller is us.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// Implementors should consider using a less efficient encoding, if possible,to fit exactly within the reserved capacity.
```
Lukasa, line 48:
```suggestion
    /// Implementers should consider using a less efficient encoding, if possible,to fit exactly within the reserved capacity.
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// Otherwise, callers may need to shift bytes to reconcile the difference.
+    /// It is up to the implementor to find the balance between performance and size.
```
Lukasa, line 50:
```suggestion
    /// It is up to the implementer to find the balance between performance and size.
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    /// - Returns: The number of bytes used to write the integer.
+    func writeInteger<IntegerType: FixedWidthInteger>(
+        _ integer: IntegerType,
+        to buffer: inout ByteBuffer
+    ) -> Int
+
+    /// An estimate of the bytes required to write integers using ths strategy.
+    /// Callers may use this to reserve bytes before writing the integer.
+    /// If the actual bytes used by the write function is more or less than this, it may be necessary to shuffle bytes.
+    /// Therefore, an accurate prediction here will improve performance.
```
Lukasa, line 43: We should clarify that this is only used in the function below.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,310 @@
...
+    ) -> Int {
+        self.writeInteger(integer, to: &buffer)
+    }
+}
+
+extension ByteBuffer {
+    /// Read a binary encoded integer, moving the `readerIndex` appropriately.
+    /// If there are not enough bytes, nil is returned.
+    @inlinable
+    public mutating func readEncodedInteger<Strategy: NIOBinaryIntegerEncodingStrategy>(_ strategy: Strategy) -> Int? {
```
Lukasa, line 81: These should probably be generic over `FixedWidthInteger`, the way the existing `readInteger` and `writeInteger` functions are.

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+extension ByteBuffer {
+    public struct QUICBinaryEncodingStrategy: NIOBinaryIntegerEncodingStrategy {
```
Lukasa, line 16: It'd be nice to have a doc comment here.

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
```diff
@@ -0,0 +1,138 @@
...
+        public var requiredBytesHint: Int
+
+        /// - Parameter requiredBytesHint: An estimate of the bytes required to write integers using this strategy
+        @inlinable
+        public init(requiredBytesHint: Int) {
```
Lukasa, line 21: Do we think it's useful to make this configurable?

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+## ``NIOBinaryIntegerEncodingStrategy`` protocol
+
+The first building block is a protocol which describes how to encode and decode an integer.
+
+An implementation of this protocol is needed for any encoding strategy. Currently, we provide one
+implementation: ``ByteBuffer/QUICBinaryEncodingStrategy``
+
+This protocol only has 2 requirements which don't have default implementations:
+
+- `readInteger`: Reads an integer from the ByteBuffer using this encoding. Implementations will read as many bytes as
```
glbrntt, line 24:
```suggestion
- `readInteger`: Reads an integer from the `ByteBuffer` using this encoding. Implementations will read as many bytes as
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+  they need to, according to their wire format, and move the reader index accordingly
+- `writeInteger`: Write an integer to the ByteBuffer using this encoding. Implementations will write as many bytes as
```
glbrntt, line 26:
```suggestion
- `writeInteger`: Write an integer to the `ByteBuffer` using this encoding. Implementations will write as many bytes as
```

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
hamzahrmalik: In some cases, a user might have a rough idea of how many bytes they're expecting to write. In such cases, allowing them to use a smaller number would result in saved bytes

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+
+- Encode the length of the integer into the integer itself when writing, so it knows how many bytes to read when
+  reading. This is what QUIC does.
+- Always use the same length, e.g. a simple strategy which always writes the integer as a UInt64.
+
+## Extensions on ``ByteBuffer``
+
+To provide a more user-friendly API, we have added extensions on `ByteBuffer` for writing integers with a
+chosen ``NIOBinaryIntegerEncodingStrategy``. These are ``ByteBuffer/writeEncodedInteger(_:strategy:)``
+and ``ByteBuffer/readEncodedInteger(_:as:)``.
```
glbrntt, line 39: Feels a bit odd to have:

```swift
buffer.writeEncodedInteger(42, strategy: .quic)
```

vs.

```swift
buffer.readEncodedInteger(.quic: as Int32.self)
```

IMO the latter would make more sense as:

```swift
// Explicit as
buffer.readEncodedInteger(as: Int32.self, strategy: .quic)
// Or, inferring the type:
buffer.readEncodedInteger(strategy: .quic)
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+    writeData: (_ buffer: inout ByteBuffer) throws -> Int
+) rethrows -> Int
+```
+
+Users could use the function as follows:
+
+```swift
+myBuffer.writeLengthPrefixed(strategy: .quic) { writer in
+    writer.writeString("something")
+    writer.writeSomethingComplex(something)
```
glbrntt, line 72: nit: docs above use `buffer` as the name of the param passed to the closure. Seems odd to not use that here as well.

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+Writing the implementation of `writeLengthPrefixed` presented a challenge. We need to write the length _before_ the
+data. But we do not know the length until the data is written.
```
glbrntt, line 77:
```suggestion
Writing the implementation of `writeLengthPrefixed` presents a challenge. We need to write the length _before_ the
data. But we do not know the length until the data is written.
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+
+Ideally, we would reserve some number of bytes, then call the writeData closure, and then go back and write the length
```
glbrntt, line 79:
```suggestion
Ideally, we would reserve some number of bytes, then call the `writeData` closure, and then go back and write the length
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+in the reserved space. However, we would not even know how many bytes of space to reserve, because the number of bytes
+needed to write an integer will depend on the integer!
+
+The solution we landed on was the following:
```
glbrntt, line 83: This _is_ the current solution

```suggestion
The solution we landed on is the following:
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+- Added ``NIOBinaryIntegerEncodingStrategy/requiredBytesHint``. This allows strategies to provide an estimate of how
+  many bytes they need for encoding a length
+- Using this property, reserve the estimated number of bytes
+- Call the `writeData` closure to write the data
+- Go back to the reserved space to write the length
+    - If the length ends up needing fewer bytes than we had reserved, shuffle the data back to close the gap
+    - If the length ends up needing more bytes than we had reserved, shuffle the data forward to make space
+
+This code will be most performant when the `requiredBytesHint` is exactly correct, because it will avoid needing to
+shuffle any bytes. With that in mind, we can actually make one more optimisation: when we call the writeInteger function
```
glbrntt, line 94:
```suggestion
shuffle any bytes. With that in mind, we can actually make one more optimisation: when we call the `writeInteger` function
```

Sources/NIOCore/Docs.docc/ByteBuffer-lengthPrefix.md
```diff
@@ -0,0 +1,119 @@
...
+///   - buffer: The buffer to write into.
+/// - Returns: The number of bytes used to write the integer.
+func writeInteger(
+    _ integer: Int,
+    reservedCapacity: Int,
+    to buffer: inout ByteBuffer
+) -> Int
+```
+
+Many strategies will not be able to do anything useful with the additional `reservedCapacity` parameter. For e.g., in
```
glbrntt, line 114: nit: "e.g." is the abbreviation of "exempli gratia" which means "for the sake of example", so no need to write "For e.g.", Just use "e.g." or "for example"

```suggestion
Many strategies will not be able to do anything useful with the additional `reservedCapacity` parameter. For example, in
```

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
hamzahrmalik: An alternative would be, make this not configurable, but then add a parameter to ByteBuffer/writeLengthPrefixed(strategy:writeData:)

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,315 @@
...
+}
+
+extension ByteBuffer {
+    /// Read a binary encoded integer, moving the `readerIndex` appropriately.
+    /// If there are not enough bytes, nil is returned.
+    @inlinable
+    public mutating func readEncodedInteger<Strategy: NIOBinaryIntegerEncodingStrategy, Integer: FixedWidthInteger>(
+        _ strategy: Strategy,
+        as: Integer.Type = Integer.self
+    ) -> Integer? {
```
glbrntt, line 86: `readEncodedInteger(.quic)` is a bit odd when using type inference (and is different to `writeEncodedInteger(42, strategy: .quic)`).

I think this should be spelled as `readEncodedInteger(as:strategy:)` which works well with or without type inference and mirrors the write API nicely.

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,315 @@
...
+        Strategy: NIOBinaryIntegerEncodingStrategy
+    >(
+        _ value: Integer,
+        strategy: Strategy
+    ) -> Int {
+        strategy.writeInteger(value, to: &self)
+    }
+
+    /// Prefixes bytes written by `writeData` with the number of bytes written.
+    /// The number of bytes written is encoded using `strategy`
```
glbrntt, line 106:
```suggestion
    /// The number of bytes written is encoded using `strategy`.
```

Sources/NIOCore/ByteBuffer-binaryEncodedLengthPrefix.swift
```diff
@@ -0,0 +1,315 @@
...
+        let totalBytesWritten = self.writerIndex - lengthPrefixIndex
+        return totalBytesWritten
+    }
+
+    /// Reads a slice which is prefixed with a length. The length will be read using `strategy`, and then that many bytes will be read to create a slice.
+    /// - Returns: The slice, if there are enough bytes to read it fully. In this case, the readerIndex will move to after the slice.
+    /// If there are not enough bytes to read the full slice, the readerIndex will stay unchanged.
+    @inlinable
+    public mutating func readLengthPrefixedSlice<Strategy: NIOBinaryIntegerEncodingStrategy>(
+        _ strategy: Strategy
```
glbrntt, line 198: Not sure about dropping the label on the `strategy` here, I think we should keep the strategy label.

glbrntt (changes requested):
This mostly looks really good. Most of my comments are doc nits but a couple are minor API tweaks that I think we should fix.

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
```diff
@@ -129,10 +131,14 @@ extension ByteBuffer {
 
 extension NIOBinaryIntegerEncodingStrategy where Self == ByteBuffer.QUICBinaryEncodingStrategy {
     @inlinable
+    /// Encodes bytes as defined in RFC 9000 § 16
+    /// - Parameter requiredBytesHint: An estimate of the bytes required to write integers using this strategy
+    /// - Returns: <#description#>
```
glbrntt, line 136: This one needs filling out, `<#description#>`

glbrntt (approved):
Looks good modulo one doc nit!

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
```diff
@@ -0,0 +1,144 @@
...
+                fatalError("Unreachable")
+            }
+        }
+
+        private enum NumBytes {
+            case one, two, four, eight
+        }
+
+        @usableFromInline
+        func bytesNeededForInteger<IntegerType: FixedWidthInteger>(_ integer: IntegerType) -> Int {
```
ehaydenr, line 69: I think exposing this logic publicly would be useful. Consider a case where I want to write multiple variable length integers to a buffer - how can I appropriately size the buffer's capacity ahead of time? I can assume they'll all need 8 bytes, but then I'm going to over allocate. 
```
var buffer = allocator.buffer(capacity: enoughRoomForTwoVarints)
buffer.writeInteger(myInt, strategy: .quic)
buffer.writeInteger(myOtherInt, strategy: .quic)
```

Sources/NIOCore/ByteBuffer-quicBinaryEncodingStrategy.swift
hamzahrmalik: leaning towards the alternative
Having the requiredBytesHint parameter here is quite confusing

Lukasa (approved):
Nice one, let's get this landed.

Tests/NIOCoreTests/ByteBufferQUICBinaryEncodingStrategyTests.swift
```diff
@@ -0,0 +1,136 @@
...
+        // 2 bit mask is 10 for 4 bytes so this becomes 0b10011101_01111111_00111110_01111101
+        XCTAssertEqual(buffer.readInteger(as: UInt32.self), 0b10011101_01111111_00111110_01111101)
+        XCTAssertEqual(buffer.readableBytes, 0)
+    }
+
+    func testWriteEightByteQUICVariableLengthInteger() {
+        var buffer = ByteBuffer()
+        let strategy = ByteBuffer.QUICBinaryEncodingStrategy.quic
+        let bytesWritten = strategy.writeInteger(
+            0b00000010_00011001_01111100_01011110_11111111_00010100_11101000_10001100,
```
finagolfin, line 75: @hamzahrmalik, this line and a couple more writes below [broke compilation of these tests for 32-bit platforms like Android armv7 on my Android CI](https://github.com/finagolfin/swift-android-sdk/actions/runs/11159109574/job/31016848537#step:15:6552).

weissi: @finagolfin @Lukasa @FranzBusch can we integrate the Android & 32 bit CIs into NIO's own github actions?

finagolfin: Franz and I discussed getting Android tested on this CI, I'm aiming to have something for that in the coming months. However, that just builds the source for armv7 but wouldn't run the tests.

If you have a 32-bit CI like watchOS or embedded that you could add quickly, that may be a better first step to get 32-bit platforms tested.

hamzahrmalik: I've raised a PR to fix this: #2904

Issue #2752, opened 2024-06-25
FileChunks isn't Sendable


weissi:
`FileChunks` is an `AsyncSequence` but not `Sendable`. That makes it annoying to use in places.

PR #2871, opened 2024-09-07
Issue-2752 - Make FileChunk sendable

ali-ahsan-ali:
Motivation is to solve https://github.com/apple/swift-nio/issues/2752


glbrntt (approved):
Great, thank you @ali-ahsan-ali!

PR #2879, opened 2024-09-10
Adopt `NIOThrowingAsyncSequenceProducer`

rnro:
### Motivation:

Adopt `NIOThrowingAsyncSequenceProducer` in NIOFileSystem to reduce code duplication.

### Modifications:

Adopt `NIOThrowingAsyncSequenceProducer` in NIOFileSystem `DirectoryEntryProducer` and `FileChunkProducer`

### Result:

No functional changes. Internal changes reduce code duplication.

### DIFF
--- a/Sources/NIOFileSystem/DirectoryEntries.swift
+++ b/Sources/NIOFileSystem/DirectoryEntries.swift
@@ -119,7 +119,9 @@ extension DirectoryEntries {
         public struct BatchedIterator: AsyncIteratorProtocol {
             private var iterator: BufferedOrAnyStream<[DirectoryEntry], DirectoryEntryProducer>.AsyncIterator
 
-            fileprivate init(wrapping iterator: BufferedOrAnyStream<[DirectoryEntry], DirectoryEntryProducer>.AsyncIterator) {
+            fileprivate init(
+                wrapping iterator: BufferedOrAnyStream<[DirectoryEntry], DirectoryEntryProducer>.AsyncIterator
+            ) {
                 self.iterator = iterator
             }
 
@@ -136,15 +138,23 @@ extension DirectoryEntries.Batched.AsyncIterator: Sendable {}
 // MARK: - Internal
 
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-extension NIOThrowingAsyncSequenceProducer where Element == [DirectoryEntry], Failure == Error,
-        Strategy == NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark, Delegate == DirectoryEntryProducer {
+extension NIOThrowingAsyncSequenceProducer
+where
+    Element == [DirectoryEntry],
+    Failure == (any Error),
+    Strategy == NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark,
+    Delegate == DirectoryEntryProducer
+{
     fileprivate static func makeBatchedDirectoryEntryStream(
         handle: SystemFileHandle,
         recursive: Bool,
         entriesPerBatch: Int,
         lowWatermark: Int,
         highWatermark: Int
-    ) -> NIOThrowingAsyncSequenceProducer<[DirectoryEntry], any Error, NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark, DirectoryEntryProducer> {
+    ) -> NIOThrowingAsyncSequenceProducer<
+        [DirectoryEntry], any Error, NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark,
+        DirectoryEntryProducer
+    > {
         let producer = DirectoryEntryProducer(
             handle: handle,
             recursive: recursive,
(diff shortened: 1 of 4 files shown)


FranzBusch (commented):
In general this looks good to me. Have you run some benchmarks between the two to make sure we still have the same performance?

Sources/NIOFileSystem/DirectoryEntries.swift
```diff
@@ -135,52 +136,85 @@ extension DirectoryEntries.Batched.AsyncIterator: Sendable {}
 // MARK: - Internal
 
 @available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
-extension BufferedStream where Element == [DirectoryEntry] {
+extension NIOThrowingAsyncSequenceProducer where Element == [DirectoryEntry], Failure == Error,
```
glbrntt, line 139: nit
```suggestion
extension NIOThrowingAsyncSequenceProducer where Element == [DirectoryEntry], Failure == (any Error),
```

Sources/NIOFileSystem/DirectoryEntries.swift
```diff
@@ -353,22 +405,40 @@ private struct DirectoryEnumerator: Sendable {
...
 
-    internal func produceMore() -> NIOThreadPool? {
+    internal mutating func produceMore() -> NIOThreadPool? {
         switch self.state {
         case let .idle(handle, _):
             return handle.threadPool
         case let .open(threadPool, _, _):
             return threadPool
+        case .openPausedProducing(let threadPool, let source, let array):
+            self.state = .modifying
```
glbrntt, line 415: `.modifying` isn't required here because we aren't modifying any of the state held in the `openPausedProducing` case

Sources/NIOFileSystem/DirectoryEntries.swift
```diff
@@ -353,22 +405,40 @@ private struct DirectoryEnumerator: Sendable {
...
             return nil
         case .modifying:
             fatalError()
         }
     }
 
+    internal mutating func pauseProducing() {
+        switch self.state {
+        case .open(let nIOThreadPool, let source, let array):
+            self.state = .modifying
```
glbrntt, line 428: same here, no need to do `.modifying`

Sources/NIOFileSystem/DirectoryEntries.swift
```diff
@@ -353,22 +405,40 @@ private struct DirectoryEnumerator: Sendable {
...
         case .done:
             return nil
         case .modifying:
             fatalError()
         }
     }
 
+    internal mutating func pauseProducing() {
+        switch self.state {
+        case .open(let nIOThreadPool, let source, let array):
```
glbrntt, line 427: nit: `threadPool`

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -85,61 +85,89 @@ extension FileChunks.FileChunkIterator: Sendable {}
...
-private struct FileChunkProducer: Sendable {
+private final class FileChunkProducer: NIOAsyncSequenceProducerDelegate, Sendable {
     let state: NIOLockedValueBox<ProducerState>
-    let source: BufferedStream<ByteBuffer>.Source
+
     let path: FilePath
     let length: Int64
 
+    init(range: FileChunks.ChunkRange, handle: SystemFileHandle, length: Int64) {
+
```
glbrntt, line 132: nit: no blank lines at the start of funcs
```suggestion
    init(range: FileChunks.ChunkRange, handle: SystemFileHandle, length: Int64) {
```

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -85,61 +85,89 @@ extension FileChunks.FileChunkIterator: Sendable {}
...
+        let state: ProducerState = switch range {
+        case .entireFile:
+                .init(handle: handle, range: nil)
+        case .partial(let partialRange):
+                .init(handle: handle, range: partialRange)
+        }
```
glbrntt, line 138: Hmm, I don't think our minimum Swift version is new enough to support this? Also the indentation is way off for the body of each case.

glbrntt: (Also I really don't like this syntax but maybe I'm just a curmudgeon...)

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -302,45 +347,75 @@ private struct ProducerState: Sendable {
...
     mutating func done() {
-        self.state = .done
+        self.state = .done(emptyRange: false)
     }
 }
 
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension ProducerState.Producing {
+    mutating func updateRange(count: Int) -> Bool {
```
glbrntt, line 402: It's not obvious _how_ the range is updated without looking at the implementation, also the return value is quite misleading: `if updateRange(count: blah) { ... }` looks a whole lot like "if the range was updated then ...".

Would probably be clearer if you call this `didReadBytes(_ count: Int)` and documented the return value (you could use an enum for the return type but that's probably overkill)

glbrntt (changes requested):
This looks good! I left a couple of small comments which need addressing.

glbrntt:
> In general this looks good to me. Have you run some benchmarks between the two to make sure we still have the same performance?

I would be surprised if there was a substantial difference. You'd think the syscalls would dominate here but it's worth doing some validation nonetheless.

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -193,65 +221,74 @@ private struct FileChunkProducer: Sendable {
...
+            source.finish()
+            self.clearSource()
+            return
+        }
+
+        switch writeResult {
+        case .produceMore:
+            self.produceMore()
+        case .stopProducing:
+            self.state.withLockedValue { state in state.pauseProducing()}
```
glbrntt, line 275:
```suggestion
            self.state.withLockedValue { state in state.pauseProducing() }
```

glbrntt (approved):
Looks good modulo some failing CI!

rnro:
Performance-wise there doesn’t seem to be much of a difference.

I tested listing the files on disk, exercising the `DirectoryEntries` path. I listed a directory I happened to have with 772492 files in it making up 23GiB.

For the existing code:
3 runs of: 27.51s, 26.53s, 26.34s
With an average of:  26.8s

For the new code:
3 runs of: 26.08s, 27.71s, 27.35s
With an average of:  27.05s

FranzBusch (approved):

finagolfin:
This pull consistently causes hanging `FileHandle` tests in my Android CI, that runs them in an Android x86_64 emulator on both [linux](https://github.com/finagolfin/swift-android-sdk/actions/runs/10899962069/job/30246566565#step:38:1288) and [macOS](https://github.com/finagolfin/swift-android-sdk/actions/runs/10861194727/job/30142820028#step:38:2747) built with 5.10, 6.0, and trunk 6.1. I can reproduce locally on Android 13 AArch64 with a recently-built trunk 6.1 toolchain from the Sep. 4 source snapshot and [the Termux 5.10.1 toolchain](https://github.com/termux/termux-packages/blob/master/packages/swift/build.sh).

I'll see if I can track it down and file an issue.

PR #2883, opened 2024-09-13
ci: Use ${GITHUB_BASE_REF} as treeish for checking API break

simonjbeaumont:
### Motivation:

This repo is providing a reusable workflow for checking API breaks but it has two shortcomings:

1. It presumes the target branch name for the PR to be `main`.
2. It presumes the remote is named `origin`.

(2) is always the case in CI, but isn't always locally. My remote is named `upstream` so cannot run this using `act`. (1) is worse, because it's not actually testing against the target branch of the PR—consider if someone was enabling this reusable workflow on a feature branch or other-named branch.

Fortunately, Github Actions makes the target ref of the pull request available in the `GITHUB_BASE_REF` environment variable[^0].

### Modifications:

Use `GITHUB_BASE_REF` as the argument to `swift package diagnose-api-breaking-changes`.

### Result:

This can now be run locally.

[^0]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -65,14 +65,14 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
       with:
-        # We need to fetch everything otherwise only the head commit will be fetched.
-        fetch-depth: 0
         persist-credentials: false
     - name: Mark the workspace as safe
       # https://github.com/actions/checkout/issues/766
       run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
     - name: Run API breakage check
-      run: swift package diagnose-api-breaking-changes origin/main
+      run:  |
+        git fetch ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY} ${GITHUB_BASE_REF}:pull-base-ref
+        swift package diagnose-api-breaking-changes pull-base-ref
 
   docs-check:
     name: Documentation check


rnro (approved):
Looks great!

rnro:
> If I run this locally now without passing the env it will fail right? Can we make the default ref for local act runs `main`?

Ah, I had assumed that `act` took care of setting up the environment for you. I agree that if it doesn't we should have a default.

simonjbeaumont:
> If I run this locally now without passing the env it will fail right? Can we make the default ref for local act runs `main`?

I don't think so. Did it actually fail for you? When I run it without setting it, it works fine.

> Ah, I had assumed that `act` took care of setting up the environment for you. I agree that if it doesn't we should have a default.

No default is required.

Defaulting the ref is a bad thing IMO. By defaulting it to `main` you are presuming the target branch of the PR, when it isn't explicitly set. When it isn't set, I would expect the best choice for default would be the default branch of the repo. Which you can get by just using nothing:

```
❯ git fetch https://github.com/apple/swift-nio :one

❯ git fetch https://github.com/apple/swift-nio main:two

❯ git rev-parse one
1851ae6e7d49860ea42e8ba9c4cf0aee9633d524

❯ git rev-parse two
1851ae6e7d49860ea42e8ba9c4cf0aee9633d524
```

FranzBusch (approved):

PR #2888, opened 2024-09-19
FileChunkProducer.shouldProduceMore updates state

rnro:
### Motivation:

`FileChunkProducer` would not reflect its internal state when requested to produce more elements for the sequence after being paused.

### Modifications:

`FileChunkProducer.shouldProduceMore` updates the internal state and returns the threadpool leading to more elements being produced.

### Result:

`FileChunkProducer` should no longer hang after being paused.

### DIFF
--- a/Sources/NIOFileSystem/FileChunks.swift
+++ b/Sources/NIOFileSystem/FileChunks.swift
@@ -329,12 +329,14 @@ private struct ProducerState: Sendable {
         }
     }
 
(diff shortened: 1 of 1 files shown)


FranzBusch (approved):

rnro:
Hi @finagolfin thanks for following up on this, I was seeing this too in my testing. I've proposed that we revert the change for now and circle-back https://github.com/apple/swift-nio/pull/2892.

PR #2890, opened 2024-09-19
Add concrete description for `EmbeddedEventLoop`

aryan-25:
### Motivation:

We want to avoid going through `EventLoopGroup`'s [`String(describing: self)`](https://github.com/apple/swift-nio/blob/8307ad610a358513b2190955379829d70f04b8c2/Sources/NIOCore/EventLoop.swift#L414) when obtaining the description of an `EventLoopGroup` or `EventLoop`.

The  concrete `MultiThreadedEventLoopGroup` and `SelectableEventLoop` already implement a `description` property. `EmbeddedEventLoop` does not.

### Modifications:

Added a concrete description property in `EmbeddedEventLoop`.

### Result:

The description of `EmbeddedEventLoop` can be obtained without going through `String(describing: self)`.

### DIFF
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -97,6 +97,8 @@ public final class EmbeddedEventLoop: EventLoop {
     // execution.
     private var taskNumber: UInt64 = 0
 
+    public var description = "EmbeddedEventLoop"
+
     private func nextTaskNumber() -> UInt64 {
         defer {
             self.taskNumber += 1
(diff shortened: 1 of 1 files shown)


FranzBusch (approved):

Lukasa (approved):

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -97,6 +97,8 @@ public final class EmbeddedEventLoop: EventLoop {
     // execution.
     private var taskNumber: UInt64 = 0
 
+    public var description = "EmbeddedEventLoop"
```
glbrntt, line 100: This should be a `let`!

glbrntt (changes requested):

glbrntt (approved):

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -97,6 +97,8 @@ public final class EmbeddedEventLoop: EventLoop {
     // execution.
     private var taskNumber: UInt64 = 0
 
+    public let description = "EmbeddedEventLoop"
```
weissi, line 100: This should maybe include the pointer value of `self`? Just so we can distinguish them

PR #2891, opened 2024-09-19
Expose UDP_MAX_SEGMENTS via System

rnro:
### Motivation:

Linux kernels recently increased the value of `UDP_MAX_SEGMENTS` from 64 to 128 causing hardcoded assumptions in our tests to become invalid.

### Modifications:

Surface the value of `UDP_MAX_SEGMENTS` if it is defined on the platform in use via new public API `System.udpMaxSegments()` (otherwise return `nil`) and make use of it in `DatagramChannelTests`.

`System.udpMaxSegments()` is added as new public API in analogue to e.g. `System.supportsUDPReceiveOffload`.

### Result:

`DatagramChannelTests` pass and new API.

### DIFF
--- a/Sources/CNIOLinux/shim.c
+++ b/Sources/CNIOLinux/shim.c
@@ -216,16 +216,8 @@ const unsigned long CNIOLinux_UTIME_OMIT = UTIME_OMIT;
 const unsigned long CNIOLinux_UTIME_NOW = UTIME_NOW;
 
 
-bool CNIOLinux_udp_max_segments_defined() {
-    #ifdef UDP_MAX_SEGMENTS
-    return true;
-    #else
-    return false;
-    #endif
-}
-
 #ifdef UDP_MAX_SEGMENTS
-const unsigned long CNIOLinux_UDP_MAX_SEGMENTS = UDP_MAX_SEGMENTS;
+const long CNIOLinux_UDP_MAX_SEGMENTS = UDP_MAX_SEGMENTS;
 #endif
-const unsigned long CNIOLinux_UDP_MAX_SEGMENTS = 0;
+const long CNIOLinux_UDP_MAX_SEGMENTS = -1;
 #endif
--- a/Sources/CNIOLinux/shim.c
+++ b/Sources/CNIOLinux/shim.c
@@ -215,4 +215,17 @@ const int CNIOLinux_AT_EMPTY_PATH = AT_EMPTY_PATH;
 const unsigned long CNIOLinux_UTIME_OMIT = UTIME_OMIT;
 const unsigned long CNIOLinux_UTIME_NOW = UTIME_NOW;
 
+
+bool CNIOLinux_udp_max_segments_defined() {
+    #ifdef UDP_MAX_SEGMENTS
+    return true;
+    #else
+    return false;
+    #endif
+}
+
+#ifdef UDP_MAX_SEGMENTS
+const unsigned long CNIOLinux_UDP_MAX_SEGMENTS = UDP_MAX_SEGMENTS;
+#endif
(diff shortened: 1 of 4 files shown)


Sources/NIOCore/Utilities.swift
```diff
@@ -258,3 +258,18 @@ extension System {
     public static let supportsUDPReceiveOffload: Bool = false
     #endif
 }
+
+extension System {
+    /// Returns the UDP maximum segment count if the platform supports and defines it.
+    public static func udpMaxSegments() -> Int? {
```
FranzBusch, line 264: This should really be a `static let` that's computed on first access

Sources/NIOCore/Utilities.swift
```diff
@@ -258,3 +258,18 @@ extension System {
     public static let supportsUDPReceiveOffload: Bool = false
     #endif
 }
+
+extension System {
```
FranzBusch, line 262: Wy is this in a new extension?

Sources/NIOCore/Utilities.swift
Lukasa: I'm not sure that's true: the cost of this function is pretty close to the cost of the `dispatch_once`, so we might prefer a static `var`.

Sources/CNIOLinux/shim.c
```diff
@@ -215,4 +215,17 @@ const int CNIOLinux_AT_EMPTY_PATH = AT_EMPTY_PATH;
...
+    return true;
+    #else
+    return false;
+    #endif
+}
+
+#ifdef UDP_MAX_SEGMENTS
+const unsigned long CNIOLinux_UDP_MAX_SEGMENTS = UDP_MAX_SEGMENTS;
+#endif
+const unsigned long CNIOLinux_UDP_MAX_SEGMENTS = 0;
```
Lukasa, line 230: Rather than have `max_segments_defined`, we should probably just return a sentinel value. `-1` is good.

Sources/NIOCore/Utilities.swift
rnro: Just for separation of groups of related code - but I'm happy to lump it together if you'd rather.

Sources/NIOCore/Utilities.swift
rnro: I've changed this to a `static var`.

Sources/NIOCore/Utilities.swift
rnro: I've consolidated them now that I'm only adding one thing.

Sources/CNIOLinux/shim.c
rnro: I normally try to avoid sentinel values as a rule but given we know that this value should comfortably fit within an `Int64` I can live with it here, particularly if it might help the optimizer see through this logic.

Sources/NIOCore/Utilities.swift
```diff
@@ -257,4 +257,14 @@ extension System {
     /// The option can be enabled by setting the ``ChannelOptions/Types/DatagramReceiveOffload`` channel option.
     public static let supportsUDPReceiveOffload: Bool = false
     #endif
+
+    /// Returns the UDP maximum segment count if the platform supports and defines it.
+    public static var udpMaxSegments: Int? {
+        #if os(Linux)
+        if CNIOLinux_UDP_MAX_SEGMENTS != -1 {
```
Lukasa, line 264: Mild nit: I suspect the optimizer will end up calling this function twice, as it doesn't know what it does. I recommend that you save the value to a temporary and branch from there.

rnro: I _think_ I've addressed this but your comment isn't 100% clear to me.

Lukasa (approved):
Nice one, LGTM.

Issue #2893, opened 2024-09-20
Build errors when building against a Static Linux SDK


briancoyner:
### Expected behavior
An application with a dependency on SwiftNIO successfully builds with a Swift 6 OSS toolchain + Static Linux SDK.

**Note**: The `swift-6.0-DEVELOPMENT-SNAPSHOT-2024-09-17-a` toolchain is used vs the `swift-6.0-RELEASE` because of another issue with `swift-collections`. That issue is fixed in the "09-17" toolchain release.
- https://github.com/apple/swift-collections/issues/420

**Note**: Xcode 15.4.0 vs Xcode 16.0.0 because of an issue described here:
- https://forums.swift.org/t/swift-6-toolchain-static-linux-sdk-build-issue/74694

#### `main` branch error

```
$ xcrun --toolchain org.swift.600202409171a swift build --swift-sdk aarch64-swift-linux-musl
Building for debugging...
In file included from /Users/someuser/Developer/some-app/.build/checkouts/swift-nio/Sources/CNIOLinux/shim.c:25:
/Users/someuser/Developer/some-app/.build/checkouts/swift-nio/Sources/CNIOLinux/include/CNIOLinux.h:33:10: fatal error: 'linux/udp.h' file not found
   33 | #include <linux/udp.h>
      |          ^~~~~~~~~~~~~
1 error generated.
```

#### `2.72.0` tag error

```
$ xcrun --toolchain org.swift.600202409171a swift build --swift-sdk aarch64-swift-linux-musl
/Users/someuser/Developer/some-app/.build/checkouts/swift-nio/Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift:76:12: error: conflicting arguments to generic parameter 'T' ('Optional<UnsafeRawPointer>.AtomicRepresentation' vs. 'Optional<UnsafeRawPointer>.AtomicRepresentation')

...

/Users/someuser/Developer/some-app/.build/checkouts/swift-nio/Sources/NIOPosix/PosixSingletons+ConcurrencyTakeOver.swift:112:35: error: 'nil' requires a contextual type

...

additional errors 
```

### Steps to reproduce

1. Install the Swift 6 OSS toolchain + SDK (`swift-6.0-DEVELOPMENT-SNAPSHOT-2024-09-17-a`).
2. A simple application using SPM with a dependency on SwiftNIO.
3. Execute the following to build the application.

```
[...]

PR #2894, opened 2024-09-20
Conditionally include linux/udp.h

rnro:
### Motivation:

To enable access to `UDP_MAX_SEGMENTS` we switched from including `netinet/udp.h` to including `linux/udp.h` however the latter isn't available in MUSL.

### Modifications:

if `linux/udp.h` is available then import it in favor of `netinet/udp.h`. If we fall back to the latter then `UDP_MAX_SEGMENTS` won't be defined but that will be caught by the existing paths which account for this case.

### Result:

NIO should be able to build against a static MUSL SDK again.

### DIFF
--- a/Sources/CNIOLinux/include/CNIOLinux.h
+++ b/Sources/CNIOLinux/include/CNIOLinux.h
@@ -30,7 +30,11 @@
 #include <errno.h>
 #include <pthread.h>
 #include <netinet/ip.h>
+#if __has_include(<linux/udp.h>)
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

rnro:
This should help with part of https://github.com/apple/swift-nio/issues/2893

briancoyner:
Confirmed... `main` no longer generates the `#include <linux/udp.h>` error as described in #2893.

PR #2895, opened 2024-09-23
Update Workflows for Swift 6 release

0xTim:
Update for Swift 6 release

Update CI scripts for Swift 6 release

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -22,9 +22,9 @@ on:
             type: boolean
             description: "Boolean to enable the Linux 5.10 Swift version matrix job. Defaults to true."
             default: true
-        linux_nightly_6_0_enabled:
+        linux_6_0_enabled:
             type: boolean
-            description: "Boolean to enable the Linux nightly 6.0 Swift version matrix job. Defaults to true."
+            description: "Boolean to enable the Linux 6.0 Swift version matrix job. Defaults to true."
             default: true
         linux_nightly_main_enabled:
             type: boolean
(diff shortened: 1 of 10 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -22,9 +22,9 @@ on:
             type: boolean
             description: "Boolean to enable the Linux 5.10 Swift version matrix job. Defaults to true."
             default: true
-        linux_nightly_6_0_enabled:
+        linux_6_0_enabled:
```
MahdiBM, line 25: Bad news: If there are other repos who call these reusable workflows and do use these parameters, their CIs will start to fail.

.github/workflows/soundness.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     if: ${{ inputs.format_check_enabled }}
     runs-on: ubuntu-latest
     container:
-      image: swiftlang/swift:nightly-6.0-jammy
+      image: swift:6.0-jammy
```
MahdiBM, line 146: There are noble images as well.
I know I filed https://github.com/swiftlang/swift/issues/76555 which mentions 40-50% build regression in _our_ repo, but it should be fine for light workflows.

.github/workflows/benchmarks.yml
FranzBusch: That is intentional. This is where we landed when we wanted to achieve both an easy way to roll out new Swift versions but also give the repos the opportunity to define their own custom arguments that are passed.

.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ jobs:
             linux_5_8_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
             linux_5_9_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
             linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
-            linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_6_0_arguments_override: "--explicit-target-dependency-import-check error"
```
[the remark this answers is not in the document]
FranzBusch: We are not yet warning free on Swift 6

FranzBusch:
@0xTim Closing this since we updated the workflow in #2920

.github/workflows/benchmarks.yml
sebsto: I confirm Swift Runtime for AWS Lambda CI failed.  Fixed now :-)

PR #2901, opened 2024-10-02
ConditionLock deallocs its pthread_cond_t in more cases

rnro:
### Motivation:

Recently we changed the init and deinit behavior of `ConditionLock`, at that time the `UnsafeMutablePointer` allocation of the held `pthread_cond_t` was put behind a conditional compiler and runtime check. The deallocation of the same object was also put behind a conditional, however the conditions were mismatched leading to a leak on some platforms. Notably this surfaced when `wait`ing on an event loop future.

### Modifications:

`ConditionLock.deinit` now deallocs the held `pthread_cond_t` under the same conditions in which it allocs it.

### Result:

Creating a `ConditionLock` no longer leaks memory.


MaxDesiatov (approved):

Lukasa (approved):

Lukasa:
Worth considering whether we should add a regression benchmark test for this.

PR #2904, opened 2024-10-04
Make ByteBufferQUICBinaryEncodingStrategyTests compatible with 32-bit systems

hamzahrmalik:
This is a followup to #2867

Some of the literals used in the test cases were too big to fit in an Int32
This is fine on 64-bit systems, because the literals are considered as `Int`, which is Int64 on those systems
However, on 32-bit systems, those literals are considered as Int64

Change: Add `as Int64` where needed, to tell the compiler we want these literals to be treated as Int64, which should allow these tests to run on 32 bit systems too


Lukasa (approved):

finagolfin:
Thanks, I can confirm this pull fixed the 32-bit compilation issue for Android armv7.

PR #2909, opened 2024-10-09
Add convenience conformances to `ByteCount`

clintonpi:
Motivation:

Several arithmetic and comparison operations are done with `ByteCount` using the `bytes` property. Operating on the `ByteCount` type directly will be more convenient.

Modifications:

- Add `AdditiveArithmetic` and `Comparable` conformances to `ByteCount`.

Result:

The `ByteCount` type can undergo arithmetic and comparison operations.

### DIFF
--- a/Sources/NIOFileSystem/ByteCount.swift
+++ b/Sources/NIOFileSystem/ByteCount.swift
@@ -81,7 +81,7 @@ public struct ByteCount: Hashable, Sendable {
 }
 
 extension ByteCount: AdditiveArithmetic {
-    public static var zero: ByteCount {  ByteCount(bytes: 0) }
+    public static var zero: ByteCount { ByteCount(bytes: 0) }
 
     public static func + (lhs: ByteCount, rhs: ByteCount) -> ByteCount {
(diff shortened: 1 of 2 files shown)


Sources/NIOFileSystem/ByteCount.swift
```diff
@@ -80,4 +80,22 @@ public struct ByteCount: Hashable, Sendable {
     }
 }
 
+extension ByteCount: AdditiveArithmetic {
+    public static var zero = ByteCount(bytes: 0)
```
glbrntt, line 84: This should either be a `static let` or a `static` computed property:

```suggestion
    public static var zero: ByteCount {  ByteCount(bytes: 0) }
```

It's not a good idea to let any user change the meaning of `zero`!

glbrntt (changes requested):
Looks good, just need to fix up one thing first

glbrntt (commented):
Looks good but the format checking CI failed

glbrntt (approved):
Fab, thank you!

Issue #2877, opened 2024-09-10
NIOFileSytem: crash when passing `maximumSizeAllowed: .gibibytes(.max)` to disable limits


weissi:
```
request.body = try await Data(buffer: ByteBuffer(
    contentsOf: filePath,
    maximumSizeAllowed: .gibibytes(.max)
))
```

crashes with

```
Thread 5 Crashed:: NIO-SGLTN-0-#4
0   0x104f0e510 Swift runtime failure: arithmetic overflow + 0 [inlined]
1   0x104f0e510 static ByteCount.gibibytes(_:) + 76 (ByteCount.swift:79)
```

but it shouldn't.

glbrntt:
Agree. This intersects a bit weirdly with https://github.com/apple/swift-nio/issues/2878 – if `.gigabytes(.max)` is okay, is `.megabytes(.max)` okay? Maybe we should reshape the API and allow `maximumSizeAllowed` to be `nil` _or_ add a sentinel value to the `ByteCount` which makes the intent clearer (`.max`, maybe?).

weissi:
> Agree. This intersects a bit weirdly with #2878 – if `.gigabytes(.max)` is okay, is `.megabytes(.max)` okay? Maybe we should reshape the API and allow `maximumSizeAllowed` to be `nil` _or_ add a sentinel value to the `ByteCount` which makes the intent clearer (`.max`, maybe?).

Yes, I'd suggest just `.unlimitedDangerousForSecurityReasons` or something straight on `ByteCount`.

glbrntt:
Feels weird having it on `ByteCount` but it might just be the pragmatic choice.

weissi:
> Feels weird having it on `ByteCount` but it might just be the pragmatic choice.

Why? `ByteCount.unlimited` sounds alright, no?

PR #2914, opened 2024-10-10
[NIOFileSystem] Provide an API to specify allowing unlimited sized reads

clintonpi:
Motivation:

As described in issue [#2877](https://github.com/apple/swift-nio/issues/2877), there is no API available to specify allowing a read of unlimited size.

Modifications:

- Add a new `public static` property, `unlimited`, to `ByteCount` representing an unlimited amount of bytes.
- Adapt `ReadableFileHandleProtocol.readToEnd(fromAbsoluteOffset:maximumSizeAllowed:)` to work with `maximumSizeAllowed` being `ByteCount.unlimited`.

Result:

An API to specify allowing a read of an unlimited size will be available.

### DIFF
--- a/Sources/NIOFileSystem/FileHandleProtocol.swift
+++ b/Sources/NIOFileSystem/FileHandleProtocol.swift
@@ -335,11 +335,7 @@ extension ReadableFileHandleProtocol {
         fromAbsoluteOffset offset: Int64 = 0,
         maximumSizeAllowed: ByteCount
     ) async throws -> ByteBuffer {
-        var maximumSizeAllowed = maximumSizeAllowed
-        if maximumSizeAllowed == .unlimited {
-            maximumSizeAllowed = .byteBufferCapacity
-        }
-
+        let maximumSizeAllowed = maximumSizeAllowed == .unlimited ? .byteBufferCapacity : maximumSizeAllowed
         let info = try await self.info()
         let fileSize = Int64(info.size)
         let readSize = max(Int(fileSize - offset), 0)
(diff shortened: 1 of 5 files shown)


Sources/NIOFileSystem/FileHandleProtocol.swift
```diff
@@ -335,6 +335,11 @@ extension ReadableFileHandleProtocol {
         fromAbsoluteOffset offset: Int64 = 0,
         maximumSizeAllowed: ByteCount
     ) async throws -> ByteBuffer {
+        var maximumSizeAllowed = maximumSizeAllowed
+        if maximumSizeAllowed == .unlimited {
+            maximumSizeAllowed = .byteBufferCapacity
+        }
```
glbrntt, line 341: We can do this in a one line to keep `maximumSizeAllowed` constant:

```swift
let maximumSizeAllowed = maximumSizeAllowed == .unlimited ? .byteBufferCapacity : maximumSizeAllowed
```

glbrntt (changes requested):
One nit but looks great otherwise

glbrntt (approved):

Tests/NIOFileSystemIntegrationTests/XCTestExtensions.swift
```diff
@@ -69,4 +69,16 @@ func XCTAssertThrowsFileSystemErrorAsync<R>(
         }
     }
 }
+
+func XCTAssertNoThrowAsync<T>(
```
weissi, line 73: we shouldn't squat on XCTest's `XCT*` name space, you can make it `NIOAssertNoThrowAsync` or smth

PR #2917, opened 2024-10-11
Adopt NIOThrowingAsyncSequenceProducer 2nd try

rnro:
### Motivation:

Adopt `NIOThrowingAsyncSequenceProducer` in NIOFileSystem to reduce code
duplication.

### Modifications:

Adopt `NIOThrowingAsyncSequenceProducer` in NIOFileSystem `DirectoryEntryProducer` and `FileChunkProducer`.

This change was previously merged and then backed out due to issues (#2879). The original change is in the first commit, the second commit contains additional changes.

In the original adoption of `NIOThrowingAsyncSequenceProducer` the code did not deal with backpressure being applied very well, in some cases causes hangs.

A key bug was that it was not protected against re-entrant calls to `produceMore`. Such calls may happen e.g. when the producer has been asked to pause producing and then resume again before the initial read had completed. This resulted in overlapping reads and re-ordered data.

This change introduces a new activity state which is protected by a lock and keeps track of if we are in the critical section to serialize access.

`DirectoryEntryProducer` performs most of its logic within a lock and so
doesn't seem to be impacted in the same way.

### Result:

No functional changes. Internal changes reduce code duplication.

### DIFF
--- a/Sources/NIOFileSystem/FileChunks.swift
+++ b/Sources/NIOFileSystem/FileChunks.swift
@@ -29,7 +29,7 @@ public struct FileChunks: AsyncSequence, Sendable {
     public typealias Element = ByteBuffer
 
     /// The underlying buffered stream.
-    private let stream: BufferedOrAnyStream<ByteBuffer>
+    private let stream: BufferedOrAnyStream<ByteBuffer, FileChunkProducer>
 
     /// Create a ``FileChunks`` sequence backed by wrapping an `AsyncSequence`.
     public init<S: AsyncSequence & Sendable>(wrapping sequence: S) where S.Element == ByteBuffer {
@@ -50,7 +50,7 @@ public struct FileChunks: AsyncSequence, Sendable {
 
         // TODO: choose reasonable watermarks; this should likely be at least somewhat dependent
         // on the chunk size.
-        let stream = BufferedStream.makeFileChunksStream(
+        let stream = NIOThrowingAsyncSequenceProducer.makeFileChunksStream(
             of: ByteBuffer.self,
             handle: handle,
             chunkLength: chunkLength.bytes,
@@ -67,9 +67,9 @@ public struct FileChunks: AsyncSequence, Sendable {
     }
 
     public struct FileChunkIterator: AsyncIteratorProtocol {
-        private var iterator: BufferedOrAnyStream<ByteBuffer>.AsyncIterator
+        private var iterator: BufferedOrAnyStream<ByteBuffer, FileChunkProducer>.AsyncIterator
 
-        fileprivate init(wrapping iterator: BufferedOrAnyStream<ByteBuffer>.AsyncIterator) {
+        fileprivate init(wrapping iterator: BufferedOrAnyStream<ByteBuffer, FileChunkProducer>.AsyncIterator) {
             self.iterator = iterator
         }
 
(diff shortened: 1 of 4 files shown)


Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -85,97 +85,168 @@ extension FileChunks.FileChunkIterator: Sendable {}
...
-        let producer = FileChunkProducer(
-            state: protectedState,
-            source: source,
-            path: handle.path,
-            length: chunkLength
+        let nioThrowingAsyncSequence = NIOThrowingAsyncSequenceProducer.makeSequence(
+            elementType: ByteBuffer.self,
+            backPressureStrategy: NIOAsyncSequenceProducerBackPressureStrategies.HighLowWatermark(
+                lowWatermark: 4,
+                highWatermark: 8
```
glbrntt, line 119: We pass in the watermarks to `makeFileChunksStream` so presumably we should use them here?

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -260,87 +334,216 @@ private struct ProducerState: Sendable {
...
+
+        case .done:
+            return .stop
+        }
+    }
+
+    internal enum RangeUpdateOutcome {
+        case moreToRead
+        case cannotReadMore
+    }
```
glbrntt, line 457: nit: can you move this to the extension where its used?

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -260,87 +334,216 @@ private struct ProducerState: Sendable {
...
 
-    mutating func produceMore() -> Result<(FileDescriptor, Range<Int64>?)?, FileSystemError> {
+    mutating func fileReadingState() -> Result<(FileDescriptor, Range<Int64>?)?, FileSystemError> {
         switch self.state {
-        case let .producing(state):
-            if let descriptor = state.handle.descriptorIfAvailable() {
-                return .success((descriptor, state.range))
-            } else {
+        case .producing(let producingState), .pausedProducing(let producingState):
+            guard let descriptor = producingState.handle.descriptorIfAvailable() else {
```
glbrntt, line 471: nit: I don't think a `guard` is any clearer here as the exit on the failure path isn't earlier than on the success path. IMO if both paths immediately return then `if`/`else` is clearer.

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -260,87 +334,216 @@ private struct ProducerState: Sendable {
...
+        self.state = .done(emptyRange: false)
     }
 }
 
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension ProducerState.Producing {
+    /// Updates the range (the offsets to read from and up to) to reflect the number of bytes which have been read.
+    /// - Parameter count: The number of bytes which have been read.
+    /// - Returns: Returns `True` if there are no remaining bytes to read, `False` otherwise.
+    mutating func updateRangeWithReadBytes(_ count: Int) -> ProducerState.RangeUpdateOutcome {
```
glbrntt, line 527: This is definitely clearer than the existing in-line range updates 🙂

Sources/NIOFileSystem/FileChunks.swift
```diff
@@ -260,87 +334,216 @@ private struct ProducerState: Sendable {
...
+        // we have run out of bytes to read, we are done
+        if newLowerBound >= currentRange.upperBound {
+            self.range = currentRange.upperBound..<currentRange.upperBound
+            return .cannotReadMore
+        }
+
+        // update range, we are not done
+        self.range = newLowerBound..<currentRange.upperBound
+        if currentRange.upperBound - newLowerBound < 1 {
+            fatalError("here \(newLowerBound) \(currentRange.upperBound)")
```
glbrntt, line 544: Leftover debugging?

glbrntt (changes requested):
This looks great aside from a couple of small things to fix.

Sources/NIOFileSystem/FileChunks.swift
rnro: Good catch!

glbrntt (approved):
Awesome, thanks @rnro!

PR #2918, opened 2024-10-13
Add new protocol for ChannelHandler to get buffered bytes in the channel handler

johnnzhou:
Add new protocol to get buffered bytes from `ChannelHandler`s and `ChannelPipeline` API to query the buffered bytes from `ChannelHandler`s

### Motivation:

In #2849, a new `ChannelOption` is introduced to retrieve the number of buffered outbound bytes in a `Channel`. However, this solution does not account for bytes that may be buffered within individual `ChannelHandler`s. This PR builds on #2849 by adding functionality to audit buffered bytes residing in `ChannelHandler`s and exposing this information through new `ChannelPipeline` APIs.

### Modifications:

- Two new protocols for `ChannelHandler` to audit buffered bytes for inbound and outbound.
    - `OutboundBufferedBytesAuditableChannelHandler`
    - `InboundBufferedBytesAuditableChannelHandler`
- New `ChannelPipeline` APIs
    - auditOutboundBufferedBytes()
    - auditOutboundBufferedBytes(name: String)
    - auditOutboundBufferedBytes(handler: ChannelHandler)
    - auditInboundBufferedBytes()
    - auditInboundBufferedBytes(name: String)
    - auditInboundBufferedBytes(handler: ChannelHandler)

### Result:

Users can now easily query the amount of bytes buffered in `ChannelHandler`s using the new `ChannelPipeline` APIs, enhancing the visibility of `ChannelHandler` performance.

---

I'd like to make PR to gather some feedback from the community regarding the design before proceeding with further implementation. I'm planning to add more test cases and make existing `ChannelHandler`s in SwiftNIO project conform to new protocols.


Lukasa:
Thanks so much for opening this @johnnzhou! cc @weissi @joannis @FranzBusch who are all going to care about this API.

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -343,3 +343,19 @@ extension RemovableChannelHandler {
         context.leavePipeline(removalToken: removalToken)
     }
 }
+
+/// A `OutboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that 
+/// audits and reports the number of bytes buffered for outbound direction.
+public protocol OutboundBufferedBytesAuditableChannelHandler {
+    /// Returns the number of bytes buffered in the channel handler, which are queued to be sent to
+    /// the next outbound channel handler.
+    func auditOutboundBufferedBytes() -> Int
```
Lukasa, line 352: Minor nit: these should probably be computed properties, not functions.

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -343,3 +343,19 @@ extension RemovableChannelHandler {
         context.leavePipeline(removalToken: removalToken)
     }
 }
+
+/// A `OutboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that 
+/// audits and reports the number of bytes buffered for outbound direction.
+public protocol OutboundBufferedBytesAuditableChannelHandler {
```
Lukasa, line 349: Per our API evolution rules, these will need to be prefixed with `NIO`: 

```suggestion
public protocol NIOOutboundBufferedBytesAuditableChannelHandler {
```

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -343,3 +343,19 @@ extension RemovableChannelHandler {
...
+    /// Returns the number of bytes buffered in the channel handler, which are queued to be sent to
+    /// the next outbound channel handler.
+    func auditOutboundBufferedBytes() -> Int
+}
+
+/// A `InboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that 
+/// audits and reports the number of bytes buffered for inbound direction.
+public protocol InboundBufferedBytesAuditableChannelHandler {
```
Lukasa, line 357:
```suggestion
public protocol NIOInboundBufferedBytesAuditableChannelHandler {
```

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1492,6 +1492,8 @@ public enum ChannelPipelineError: Error {
     case alreadyRemoved
     /// `ChannelHandler` was not found.
     case notFound
+    /// `ChannelHandler` is not auditable.
+    case notAuditable
```
Lukasa, line 1496: We're not going to be able to add a new error case here: we'll either need a new error, or we'll need to find a way to use one we already have.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2091,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
         return handlers
     }
 }
+
+extension ChannelPipeline {
```
Lukasa, line 2095: These APIs will definitely want `syncOperations` versions.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2091,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        }
+
+        return future
+    }
+    
+    /// Audit the total numbers of outbound bytes buffered in the channel handler with the given `name`.
+    ///
+    /// - parameters:
+    ///     - name: the name of the channel handler whose outbound buffered bytes will be audited.
+    public func auditOutboundBufferedBytes(name: String) -> EventLoopFuture<Int> {
```
Lukasa, line 2135: I'm not sure this API is pulling its weight: users can always obtain channel handlers by name and call the relevant functions directly.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2091,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        }
+        
+        return future
+    }
+    
+    /// Audit the total numbers of outbound bytes buffered in the channel handler with the given `handler`.
+    ///
+    /// - parameters:
+    ///     - handler: the channel handler object whose outbound buffered bytes will be audited.
+    public func auditOutboundBufferedBytes(handler: ChannelHandler) -> EventLoopFuture<Int> {
```
Lukasa, line 2153: Same here: this API is just a function call on the handler directly, I don't think any extra API surface is needed for this.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2091,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        let result = self.contextSync(handler: handler)
+        switch result {
+        case .success(let context):
+            return audit0(context: context, direction: direction)
+        case .failure(let error):
+            return .failure(error)
+        }
+    }
+    
+    private func audit0(context: ChannelHandlerContext, direction: AuditDirection) -> Result<Int, Error> {
```
Lukasa, line 2223: I'm not sure this needs an error result for not being able to do the type conversion, it might just want to return `Int?`

Sources/NIOCore/ChannelHandler.swift
dnadoba: Only if we want to add an `O(1)` requirement. I don't we think we want though because some ChannelHandlers have `n` child ChannelHandlers which means they wouldn't be able to conform, strictly speaking. 

It would also be unnecessarily different from the other overloads that take an argument.

Sources/NIOCore/ChannelPipeline.swift
dnadoba: It could just throw if it really wants to propagate an error.

johnnzhou: Given that it is not a good design choice to add an error in `ChannelPipelineError` and error conveys the same outcome as `Int?`, I agree it is better to return `Int?` directly.

Sources/NIOCore/ChannelPipeline.swift
johnnzhou: added in https://github.com/apple/swift-nio/pull/2918/commits/f5f85b531ca23799de9c80b87d174be296db31e3

FranzBusch (commented):
In general I am happy to see an API added for this. I am a bit unsure if the word `audit` is the right one here. Could we just drop that from all the APIs here? What do others think?

johnnzhou:
>  I am a bit unsure if the word `audit` is the right one here. Could we just drop that from all the APIs here? What do others think?

Hi @FranzBusch, thanks for this suggestion. I originally designed this functionality as a way for users to inspect the number of bytes buffered in the channel handlers. Perhaps `audit` is more suited for financial contexts.

How about we use the word `count` instead? For instance, we could rename the protocol to `NIOOutboundBufferedBytesCountableChannelHandler` and the public method to `countOutboundBufferedBytes`. This change would preserve the idea of inspection while emphasizing the action of counting, making it more intuitive.

What do you think?

Joannis:
Would this PR also (intend to) support child channels, such as in HTTP/2.

FranzBusch:
> How about we use the word `count` instead? For instance, we could rename the protocol to `NIOOutboundBufferedBytesCountableChannelHandler` and the public method to `countOutboundBufferedBytes`. This change would preserve the idea of inspection while emphasizing the action of counting, making it more intuitive.

Do we need count even here? What about `NIOOutboundByteBufferingChannelHandler`? The current name of the property sounds good to me.

> Would this PR also (intend to) support child channels, such as in HTTP/2.

Depends what you expect. Every `ChannelHandler` that does multiplexing such as the `H2Handler` needs to decide how to treat. When a handler in a stream channel asks about the outbound buffered bytes it could return both the buffer from just the stream or the whole connection. We need to decide and implement that on a case-by-case basis.

johnnzhou:
> Do we need count even here? What about `NIOOutboundByteBufferingChannelHandler`? The current name of the property sounds good to me.

I'm generally ok with `NIOOutboundByteBufferingChannelHandler`. However, my main concern is that this name might be a bit misleading. Two new protocols are not intended to help provide buffering mechanism or manage bytes storage in the channel handlers. Instead, their main function is to collect the number of bytes buffered in the channel handler and allow users to make informed decision on their network load. 

> 
> > Would this PR also (intend to) support child channels, such as in HTTP/2.
> 
> Depends what you expect. Every `ChannelHandler` that does multiplexing such as the `H2Handler` needs to decide how to treat. When a handler in a stream channel asks about the outbound buffered bytes it could return both the buffer from just the stream or the whole connection. We need to decide and implement that on a case-by-case basis.

My original plan is to implement this protocol for channel handlers provided by NIO. But maybe we need to make separate PRs for each handler?

FranzBusch:
> I'm generally ok with `NIOOutboundByteBufferingChannelHandler`. However, my main concern is that this name might be a bit misleading. Two new protocols are not intended to help provide buffering mechanism or manage bytes storage in the channel handlers. Instead, their main function is to collect the number of bytes buffered in the channel handler and allow users to make informed decision on their network load.

I understand where you are coming from. I personally still prefer that name but I would like to hear what @Lukasa @glbrntt or @weissi think. 

> My original plan is to implement this protocol for channel handlers provided by NIO. But maybe we need to make separate PRs for each handler?

We should definitely separate the API additions in this PR from the adoption.

Lukasa:
I'm happy for us to go with `NIOOutboundByteBufferingChannelHandler`

johnnzhou:
Sounds good. I will make the change accordingly.

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -346,15 +346,15 @@ extension RemovableChannelHandler {
 
 /// A `OutboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that
 /// audits and reports the number of bytes buffered for outbound direction.
```
FranzBusch, line 348: Can we update this document please. Both to adjust the name of the protocol and remove the word audit

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -346,15 +346,15 @@ extension RemovableChannelHandler {
...
-public protocol NIOOutboundBufferedBytesAuditableChannelHandler {
+public protocol NIOOutboundByteBufferingChannelHandler {
     /// The number of bytes buffered in the channel handler, which are queued to be sent to
     /// the next outbound channel handler.
     var outboundBufferedBytes: Int { get }
 }
 
 /// A `InboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that
 /// audits and reports the number of bytes buffered for inbound direction.
```
FranzBusch, line 356: Same here

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2091,20 +2091,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
 }
 
 extension ChannelPipeline {
-    private enum AuditDirection: Equatable {
+    private enum BufferingDirection: Equatable {
         case inbound
         case outbound
     }
 
     /// Audit the total number of bytes buffered for outbound.
```
FranzBusch, line 2099: Same here

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2091,20 +2091,20 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
-    public func auditOutboundBufferedBytes() -> EventLoopFuture<Int> {
+    public func outboundBufferedBytes() -> EventLoopFuture<Int> {
```
FranzBusch, line 2100: Should this be a computed var instead of a method?

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2120,29 +2120,29 @@ extension ChannelPipeline {
     ///            referenced by the `ChannelHandlerContext` parameter `in` is collected.
     ///            If the `ChannelHandler` in the `ChannelHandlerContext` does not conform to
     ///            `NIOOutboundBufferedBytesAuditableChannelHandler`, the future will contain`nil`.
-    public func auditOutboundBufferedBytes(in context: ChannelHandlerContext) -> EventLoopFuture<Int?> {
+    public func outboundBufferedBytes(in context: ChannelHandlerContext) -> EventLoopFuture<Int?> {
```
FranzBusch, line 2123: Similar here. Should this be a computed var?

Sources/NIOCore/ChannelHandler.swift
```diff
@@ -343,3 +343,19 @@ extension RemovableChannelHandler {
...
+/// A `OutboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that
+/// audits and reports the number of bytes buffered for outbound direction.
+public protocol NIOOutboundByteBufferingChannelHandler {
+    /// The number of bytes buffered in the channel handler, which are queued to be sent to
+    /// the next outbound channel handler.
+    var outboundBufferedBytes: Int { get }
+}
+
+/// A `InboundBufferedBytesAuditableChannelHandler` is a `ChannelHandler` that
+/// audits and reports the number of bytes buffered for inbound direction.
```
FranzBusch, line 356: Same here

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        } else {
+            future = self.eventLoop.submit {
+                self.countAll(direction: .outbound)
+            }
+        }
+
+        return future
+    }
+
+    /// Audit and retrieve the number of outbound bytes buffered in the `ChannelHandler` associated with the given `ChannelHandlerContext`.
```
FranzBusch, line 2114: Can we remove the word audit here

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        } else {
+            future = self.eventLoop.submit {
+                self.count0(context: context, direction: .outbound)
+            }
+        }
+
+        return future
+    }
+
+    /// Audit the total number of bytes buffered for inbound.
```
FranzBusch, line 2137: In general just everywhere where we used audit :)

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        } else {
+            future = self.eventLoop.submit {
+                self.count0(context: context, direction: .inbound)
+            }
+        }
+
+        return future
+    }
+
+    private func count0(context: ChannelHandlerContext, direction: BufferingDirection) -> Int? {
```
FranzBusch, line 2175:
```suggestion
    private func countBufferedBytes(context: ChannelHandlerContext, direction: BufferingDirection) -> Int? {
```

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        case .outbound:
+            guard let handler = context.handler as? NIOOutboundByteBufferingChannelHandler else {
+                return nil
+            }
+            return handler.outboundBufferedBytes
+        }
+
+    }
+
+    private func countAll(direction: BufferingDirection) -> Int {
```
FranzBusch, line 2191:
```suggestion
    private func countAllBufferedBytes(direction: BufferingDirection) -> Int {
```

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+
+        return total
+    }
+}
+
+extension ChannelPipeline.SynchronousOperations {
+    /// Audit the total number of bytes buffered for outbound.
+    ///
+    /// - Important: This *must* be called on the event loop.
+    public func auditOutboundBufferedBytes() -> Int {
```
FranzBusch, line 2219: Can we remove audit from all those methods please

FranzBusch (approved):
LGTM! @Lukasa you want to give this one final review as well?

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        case .outbound:
+            guard let handler = context.handler as? NIOOutboundByteBufferingChannelHandler else {
+                return nil
+            }
+            return handler.outboundBufferedBytes
+        }
+
+    }
+
+    private func countAllBufferedBytes(direction: BufferingDirection) -> Int {
```
Lukasa, line 2191: Can we add an `assertInEventLoop` here?

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+    /// Retrieve the number of outbound bytes buffered in the `ChannelHandler` associated with the given `ChannelHandlerContext`.
+    ///
+    /// - Parameters:
+    ///     - in: the `ChannelHandlerContext` from which the outbound buffered bytes of the `ChannelHandler` will be retrieved.
+    ///
+    /// - Returns: The `EventLoopFuture` which will be notified once the number of outbound bytes buffered in the `ChannelHandler`
+    ///            referenced by the `ChannelHandlerContext` parameter `in` is collected.
+    ///            If the `ChannelHandler` in the `ChannelHandlerContext` does not conform to
+    ///            `NIOOutboundByteBufferingChannelHandler`, the future will contain`nil`.
+    public func outboundBufferedBytes(in context: ChannelHandlerContext) -> EventLoopFuture<Int?> {
```
Lukasa, line 2123: Looking at this API I'm not really sure it's right. Users can't safely hold a `ChannelHandlerContext` from _off_ the `EventLoop` (they aren't `Sendable`), so I don't think this API makes sense here. It only really makes sense on the `syncOperations` view.

We could write alternative APIs here that _do_ work from off the EL, but I'm not all that sure those are useful anyway.

johnnzhou: Thanks for catching this! Indeed, we cannot ensure that `ChannelHandlerContext` is safe unless user uses

```swift
try channel.pipeline.context(name: inboundChannelHandlerName).flatMap { context in
      channel.pipeline.inboundBufferedBytes(in: context)
}
```

Sources/NIOCore/ChannelPipeline.swift
johnnzhou: fixed in https://github.com/apple/swift-nio/pull/2918/commits/84b9fd438c4f93fb16fcdc2b85a4af6eb5aae48e

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -2089,3 +2089,171 @@ extension ChannelPipeline: CustomDebugStringConvertible {
...
+        } else {
+            future = self.eventLoop.submit {
+                self.countBufferedBytes(context: context, direction: .inbound)
+            }
+        }
+
+        return future
+    }
+
+    private func countBufferedBytes(context: ChannelHandlerContext, direction: BufferingDirection) -> Int? {
```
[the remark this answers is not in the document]
johnnzhou: fixed in https://github.com/apple/swift-nio/pull/2918/commits/84b9fd438c4f93fb16fcdc2b85a4af6eb5aae48e

Lukasa (approved):
Very nice patch, thanks @johnnzhou! ✨

Sources/NIOCore/ChannelHandler.swift
dnadoba: Why is this now a computed property? Do we really want it to be O(1)? We should at least document it to be potentially O(n) where n is the number of child channels to be in line with the API Design Guidelines:
> Document the complexity of any computed property that is not O(1). People often assume that property access involves no significant computation, because they have stored properties as a mental model. Be sure to alert them when that assumption may be violated.
> https://www.swift.org/documentation/api-design-guidelines/#conventions

cc: @Lukasa

Issue #2876, opened 2024-09-10
NIOFileSystemFoundationCompat should provide `Data(contentsOf:maximumSizeAllowed:)`


weissi:
NIOFileSystem allows loading files into a `ByteBuffer` which is great for most purposes but there are still APIs (like `Codable`) that natively only want `Data`s. So `NIOFileSytemFoundationCompat` should provide `Data(contentsOf:maximumSizeAllowed:)`.

natikgadzhi:
@weissi ditto, I think this one is implemented in #2923. Time to close?

Issue #2875, opened 2024-09-10
NIOFileSystem should provide `Array(contentsOf:maximumSizeAllowed:)`


weissi:
NIOFileSystem allows `ByteBuffer(contentsOf:maximumSizeAllowed:)` which is great but occasionally a `Array<UInt8>` would be great too.

natikgadzhi:
@weissi this one is done too! Close it!

PR #2923, opened 2024-10-14
Provide APIs to read file into more data types

clintonpi:
Motivation:

As requested in issues [#2875](https://github.com/apple/swift-nio/issues/2875) and [#2876](https://github.com/apple/swift-nio/issues/2876), it would be convenient to be able to read the contents of a file into more data types such as `Array`, `ArraySlice` & Foundation's `Data`.

Modifications:

- Extend `Array`, `ArraySlice` & `Data` to be initialisable with the contents of a file.

Result:

The contents of a file can be read into more data types.


Sources/NIOFileSystemFoundationCompat/Data+FileSystem.swift
```diff
@@ -0,0 +1,59 @@
...
+        fileSystem: some FileSystemProtocol
+    ) async throws {
+        let byteBuffer = try await fileSystem.withFileHandle(forReadingAt: path) { handle in
+            try await handle.readToEnd(
+                fromAbsoluteOffset: 0,
+                maximumSizeAllowed: maximumSizeAllowed
+            )
+        }
+
+        self = byteBuffer.withUnsafeReadableBytes { Self($0) }
```
glbrntt, line 40: The [NIOFoundationCompat](https://github.com/apple/swift-nio/tree/main/Sources/NIOFoundationCompat) module provides an extension on `Data` which you should use instead: https://github.com/apple/swift-nio/blob/f7dc3f527576c398709b017584392fb58592e7f5/Sources/NIOFoundationCompat/ByteBuffer-foundation.swift#L403-L406

Sources/NIOFileSystem/Array+FileSystem.swift
```diff
@@ -0,0 +1,57 @@
...
+    public init(
+        contentsOf path: FilePath,
+        maximumSizeAllowed: ByteCount,
+        fileSystem: some FileSystemProtocol
+    ) async throws {
+        let byteBuffer = try await fileSystem.withFileHandle(forReadingAt: path) { handle in
+            try await handle.readToEnd(
+                fromAbsoluteOffset: 0,
+                maximumSizeAllowed: maximumSizeAllowed
+            )
```
glbrntt, line 35: 0 is the default, so no need to specify it here
```suggestion
            try await handle.readToEnd(maximumSizeAllowed: maximumSizeAllowed)
```

Sources/NIOFileSystemFoundationCompat/Data+FileSystem.swift
```diff
@@ -0,0 +1,59 @@
...
+    public init(
+        contentsOf path: FilePath,
+        maximumSizeAllowed: ByteCount,
+        fileSystem: some FileSystemProtocol
+    ) async throws {
+        let byteBuffer = try await fileSystem.withFileHandle(forReadingAt: path) { handle in
+            try await handle.readToEnd(
+                fromAbsoluteOffset: 0,
+                maximumSizeAllowed: maximumSizeAllowed
+            )
```
glbrntt, line 37: Same here re: 0
```suggestion
            try await handle.readToEnd(maximumSizeAllowed: maximumSizeAllowed)
```

Tests/NIOFileSystemFoundationCompatTests/FileSystemFoundationCompatTests.swift
```diff
@@ -33,5 +33,18 @@ final class FileSystemBytesConformanceTests: XCTestCase {
             Date(timeIntervalSince1970: 1.000000001)
         )
     }
+
+    func testReadFileIntoData() async throws {
+        let fs = FileSystem.shared
+        let path = FilePath(#filePath)
+
+        try await fs.withFileHandle(forReadingAndWritingAt: path) { fileHandle in
```
glbrntt, line 41: Writing into this source file is a _bad_ idea (think about the local development experience of running this test). Reading is fine, writing less so.

Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
```diff
@@ -1828,6 +1828,30 @@ extension FileSystemTests {
...
+        let path = FilePath(#filePath)
+
+        try await self.fs.withFileHandle(forReadingAndWritingAt: path) { fileHandle in
+            _ = try await fileHandle.write(contentsOf: [0, 1, 2], toAbsoluteOffset: 0)
+        }
+
+        let contents = try await ArraySlice(contentsOf: path, maximumSizeAllowed: .bytes(1024))
+
+        XCTAssertEqual(contents, [0, 1, 2])
+    }
```
glbrntt, line 1854: Same comments apply here as well: please don't write into this source file.

glbrntt (changes requested):
Looks good but a few things need fixing up before we can merge this.

Tests/NIOFileSystemFoundationCompatTests/FileSystemFoundationCompatTests.swift
clintonpi: True, I missed that. I worked with `temporaryFilePath()` locally but swapped it with `FilePath(#filePath)` just before committing. `FilePath(#filePath)` fails for me locally due to permission issues so I was depending on the CI tests and didn't re-test locally after swapping.

Sources/NIOFileSystemFoundationCompat/Data+FileSystem.swift
```diff
@@ -0,0 +1,58 @@
...
+    ///
+    /// - Parameters:
+    ///   - path: The path of the file to read.
+    ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
+    ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    public init(
+        contentsOf path: FilePath,
+        maximumSizeAllowed: ByteCount,
+        fileSystem: some FileSystemProtocol,
+        byteTransferStrategy: ByteBuffer.ByteTransferStrategy = .automatic
```
glbrntt, line 33: We don't think we need to expose this, we can just use the default behaviour

glbrntt (changes requested):
One nit, but looks good otherwise.

PR #2924, opened 2024-10-14
Drop support for Swift 5.8

FranzBusch:
# Motivation

We only support the last three Swift released versions which are at this time 5.9, 5.10 and 6.

# Modification

This PR drops anything related to Swift 5.8.

# Result

Version support aligned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ This is the current version of SwiftNIO and will be supported for the foreseeabl
 ### Swift Versions
 
 We commit to support the most recently released swift version (currently 5.10) and the last two minor releases before that unless this is impossible to do in one codebase.
-In addition checks are run against the latest beta release (if any) as well as the nightly swift builds and the intent is that these should pass. 
+In addition checks are run against the latest beta release (if any) as well as the nightly swift builds and the intent is that these should pass.
 
 The most recent versions of SwiftNIO support Swift 5.9 and newer. The minimum Swift version supported by SwiftNIO releases are detailed below:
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -86,7 +86,8 @@ SwiftNIO            | Minimum Swift Version
 `2.43.0 ..< 2.51.0` | 5.5.2
 `2.51.0 ..< 2.60.0` | 5.6
 `2.60.0 ..< 2.65.0` | 5.7
-`2.65.0 ...`        | 5.8
+`2.65.0 ..< 2.75.0  | 5.8
+`2.75.0 ...`        | 5.9
```
glbrntt, line 90: You've got an off-by-one error here: 2.75.0 has shipped and supports 5.8. Dropping 5.8 is a minor so this should be 2.76.0.

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -89,8 +89,7 @@ public struct NIOScheduledCallback: Sendable {
 }
 
 extension EventLoop {
-    // This could be package once we drop Swift 5.8.
-    public func _scheduleCallback(
+    package func _scheduleCallback(
```
glbrntt, line 92: Should we drop the `_` here?

README.md
```diff
@@ -86,7 +86,8 @@ SwiftNIO            | Minimum Swift Version
 `2.43.0 ..< 2.51.0` | 5.5.2
 `2.51.0 ..< 2.60.0` | 5.6
 `2.60.0 ..< 2.65.0` | 5.7
-`2.65.0 ...`        | 5.8
+`2.65.0 ..< 2.75.0  | 5.8
```
Lukasa, line 89: This is still wrong, it's 
```suggestion
`2.65.0 ..< 2.76.0  | 5.8
```

FranzBusch:
@glbrntt I kept the `_scheduleCallback` underscored since there is another `scheduleCallback` method on EL already that would conflict otherwise. That methods actually calls the underscored method.

rnro (approved):

Lukasa (approved):

PR #2925, opened 2024-10-14
[CI] Fix python lint

FranzBusch:
# Motivation

Fix the python lint check.

# Modification

This PR fixes all the warnings related to the new python lint check.

# Result

Only green CI.


PeterAdams-A:
Appears to still be failing python lint.

Lukasa (approved):

FranzBusch:
@Lukasa FYI I disabled `warnings-as-errors` on the Swift 6 pipelines here. Enabling it in the first place was a mistake since we were never clean on Swift 6.

PR #2926, opened 2024-10-15
Disable semver-label-check locally

rnro:
Disable semver-label-check locally when run via `act`.

### Motivation:

This check makes no sense when run locally.

### Modifications:

Disable semver-label-check locally when run via `act`.

### Result:

The label check should no longer run and fail locally.

### DIFF
--- a/.github/workflows/pull_request_label.yml
+++ b/.github/workflows/pull_request_label.yml
@@ -11,8 +11,10 @@ jobs:
         timeout-minutes: 1
         steps:
             - name: Checkout repository
+              if: ${{ !env.ACT }}
               uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request_label.yml
```diff
@@ -11,8 +11,10 @@ jobs:
         timeout-minutes: 1
         steps:
             - name: Checkout repository
+              if: ${{ !env.ACT }}
```
FranzBusch, line 14: This isn't the right place for this. I was expecting to put it into the `action.yml` of the semver label action

rnro:
https://github.com/apple/swift-nio/pull/2927

PR #2931, opened 2024-10-17
Add future wait benchmark to catch memory leaks

rnro:
### Motivation:

In the past we introduced a memory leak around the creation of and waiting on futures - we should protect against leaks in this fundamental operation.

### Modifications:

Add a new benchmark which would have failed with the previous bug

### Result:

Regression protection for this type of memory leak.

### DIFF
--- a/Benchmarks/Benchmarks/NIOCoreBenchmarks/Benchmarks.swift
+++ b/Benchmarks/Benchmarks/NIOCoreBenchmarks/Benchmarks.swift
@@ -23,7 +23,7 @@ let benchmarks = {
 
     let leakMetrics: [BenchmarkMetric] = [
         .mallocCountTotal,
-        .memoryLeaked
+        .memoryLeaked,
     ]
 
     Benchmark(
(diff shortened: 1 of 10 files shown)


Lukasa (approved):

Benchmarks/Benchmarks/NIOCoreBenchmarks/Benchmarks.swift
```diff
@@ -21,6 +21,11 @@ let benchmarks = {
         .mallocCountTotal
     ]
 
+    let leakMetrics: [BenchmarkMetric] = [
+        .mallocCountTotal,
+        .memoryLeaked,
+    ]
```
FranzBusch, line 27: I wanted to add one comment here. Should we add leaks to all of our benchmarks. Seems like we had that before in the old school alloc tests.

rnro: _Maybe_. Unfortunately it seems like there isn't a simple "detect leaks" lever to pull here. Even in a case where I manufactured a leak I had to tweak the configuration carefully to have it detected. We could copy that configuration and hope it would detect _any_ leak but doing so would slow down the benchmarks significantly. I'm planning on investigating jemalloc in more detail to see if we can do anything to make leak detection more robust.

Issue #2934, opened 2024-10-18
`ConditionLock` doesn't build on Windows.


jeffdav:
### Expected behavior
When running `swift build` on Windows, it should build.

### Actual behavior
It doesn't build:
```pwsh
PS C:\Users\jeffd\src\swift-nio> swift build -Xcc -DBYTE_ORDER=LITTLE_ENDIAN
Building for debugging...
C:\Users\jeffd\src\swift-nio\Sources\NIOConcurrencyHelpers\lock.swift:234:37: error: value of type 'LockStorage<Void>' has no member 'mutex'
232 |             if !SleepConditionVariableSRW(
233 |                 self.cond,
234 |                 self.mutex._storage.mutex,
    |                                     `- error: value of type 'LockStorage<Void>' has no member 'mutex'
235 |                 dwMilliseconds,
236 |                 0
[31/99] Emitting module NIOConcurrencyHelpers
```

And indeed it doesn't have that member. I believe the fix is to use `withLockPrimitive` instead.

### Steps to reproduce

1. Clone the repository on Windows.
2. `swift build -Xcc -DBYTE_ORDER=LITTLE_ENDIAN`

### If possible, minimal yet complete reproducer code (or URL to code)

N/a.

### SwiftNIO version/commit hash
```commit ff98c93fe95c65cce6d160605b8e2efcfbe65f8e (HEAD -> main, origin/main, origin/HEAD)```

### System & version information

Please provide at the very least your operating system and Swift version
information.

```pwsh
PS C:\Users\jeffd\src\swift-nio> swift --version
compnerd.org Swift version 6.0-dev (LLVM a527f49f462b0d7, Swift eef85a7be270421)
Target: x86_64-unknown-windows-msvc

PS C:\Users\jeffd\src\swift-nio> [System.Environment]::OSVersion

Platform ServicePack Version      VersionString
-------- ----------- -------      -------------
 Win32NT             10.0.22631.0 Microsoft Windows NT 10.0.22631.0
```

PR #2935, opened 2024-10-18
Fix Windows build break.

jeffdav:
Use `withLockPrimitive` to access the underlying lock.

### Motivation:

The current method of accessing the lock doesn't build.
This fixes bug #2934.

### Modifications:

Access the lock via `withLockPrimitive`.

### Result:

Now it builds! :)


Lukasa (approved):
Nice fix, thanks!

PR #2940, opened 2024-10-22
[CI] Enable benchmarks & cxx Windows CI

FranzBusch:
This PR makes it possible to enable benchmarks and cxx interop checks on Windows. By default they are turned off. This PR also fixes the job of the Windows 6 pipeline which was broken since single configuration matrix jobs seem to behave differently.

### DIFF
--- a/.github/workflows/swift_matrix.yml
+++ b/.github/workflows/swift_matrix.yml
@@ -149,7 +149,7 @@ jobs:
           curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/check-matrix-job.sh | bash
 
   windows:
-    name: Windows (${{ matrix.swift.swift_version }})
+    name: Windows (6.0)
     runs-on: windows-2022
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/swift_matrix.yml
```diff
@@ -149,7 +149,7 @@ jobs:
           curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/check-matrix-job.sh | bash
 
   windows:
-    name: Windows (${{ matrix.swift.swift_version }})
+    name: Windows (6.0)
```
PeterAdams-A, line 152: Is this just waiting to go stale?  Perhaps better not to even mention version?

PeterAdams-A (approved):

.github/workflows/swift_matrix.yml
FranzBusch: This was intentional. The problem appears that a matrix with a single element doesn't expand properly. That's why I hard coded it for now. As soon as we get 6.1 we need to expand the matrix and make this dynamic again. The idea is that the check name won't change.

PR #2941, opened 2024-10-23
[CI] Fix the Windows 6.0 check name

FranzBusch:
# Motivation

Currently, the Windows 6.0 check name includes the matrix properties.

# Modification

This PR makes it dynamic with the swift version again.

# Result

Aligned CI check names across platforms.

### DIFF
--- a/.github/workflows/swift_matrix.yml
+++ b/.github/workflows/swift_matrix.yml
@@ -149,7 +149,7 @@ jobs:
           curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/check-matrix-job.sh | bash
 
   windows:
-    name: Windows (6.0)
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_matrix.yml
```diff
@@ -159,7 +159,6 @@ jobs:
           - image: swift:6.0-windowsservercore-ltsc2022
             swift_version: "6.0"
             enabled: ${{ inputs.matrix_windows_6_0_enabled }}
-    if: ${{ inputs.matrix_windows_6_0_enabled }}
```
FranzBusch, line 162: These if checks caused the naming problems since those are evaluated before the matrix so if we disabled the check then the names would get bad

PeterAdams-A (approved):

PR #2946, opened 2024-10-24
Checkout any submodules when running Actions

PeterAdams-A:
Motivation:

Some repos using this as their source of GitHub Actions logic have submodules.

Modifications:

Change workflows to checkout submodules.

Result:

All CI will work cleanly

### DIFF
--- a/.github/workflows/swift_matrix.yml
+++ b/.github/workflows/swift_matrix.yml
@@ -130,6 +130,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           persist-credentials: false
+          submodules: true
       - name: Mark the workspace as safe
         if: ${{ matrix.swift.enabled }}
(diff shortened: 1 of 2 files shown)


.github/workflows/swift_matrix.yml
```diff
@@ -168,6 +169,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           persist-credentials: false
+          format_check_enabled: true
```
gjcairo, line 172: Nit but do we want to do this in a separate PR, since it's not related to checking out submodules?

PeterAdams-A: Yikes, didn't mean to do that.  Will take ti out.

ali-ahsan-ali (approved):

Issue #2950, opened 2024-10-24
EmbeddedScheduledCallbackTests: violates `EmbeddedEventLoop` requirements


weissi:
`EmbeddedScheduledCallbackTests` uses `EmbeddedEventLoop` as if it were thread safe, that's not right.

Lukasa:
Resolved in #2951.

PR #2951, opened 2024-10-24
EmbeddedEventLoop/Channel: check correct thread

### LINKED ISSUE #2949: EmbeddedEventLoop/Channel violates Sendable: Store the current thread in `init()` and then assert that we're on that thread
`EmbeddedChannel` & `EmbeddedEventLoop` currently violate `Sendable`. They should store the current thread in `init` and then implement `inEventLoop` with `precondition(self.myThread == Thread.current); return true`.

Otherwise, EmbeddedChannel/EventLoop are just trivially violating the rules.

weissi:
### Motivation:

`EmbeddedChannel` & `EmbeddedEventLoop` currently violate `Sendable`. They should store the current thread in `init` and then implement `inEventLoop` and other functions with a check that the current thread is correct.
Since NIO 1.0 `Embedded*` were always documented to not be thread-safe but we should finally police this.

### Modifications:

- Implement the thread check
- For now, just warn (soon hopefully crash)
- Delete EmbeddedScheduledCallbackTests (#2950)

### Result:

- Less Embedded* abuse
- fixes #2949

### DIFF
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -22,6 +22,20 @@ import _NIODataStructures
 import Dispatch
 #endif
 
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#elseif canImport(Musl)
+import Musl
+#elseif canImport(Bionic)
+import Bionic
+#elseif canImport(WASILibc)
+import WASILibc
+#else
+#error("Unknown C library.")
+#endif
+
 internal struct EmbeddedScheduledTask {
     let id: UInt64
     let task: () -> Void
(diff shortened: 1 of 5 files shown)


Sources/NIOEmbedded/Embedded.swift
```diff
@@ -99,6 +113,18 @@ public final class EmbeddedEventLoop: EventLoop, CustomStringConvertible {
...
 
+    #if canImport(Darwin) || canImport(Glibc) || canImport(Musl) || canImport(Bionic)
+    private let myThread: pthread_t = pthread_self()
+
+    func isCorrectThread() -> Bool {
+        pthread_equal(self.myThread, pthread_self()) != 0
+    }
+    #else
+    func isCorrectThread() -> Bool {
+        true  // let's be conservative
```
Lukasa, line 124: By "conservative" this means "keep working"?

Lukasa (approved):

PR #2952, opened 2024-10-28
Fix `NIOAsyncSequenceProducer` watermark strategy.

FranzBusch:
# Motivation

It was currently possible that the producer's delegate is getting called twice with `produceMore` even if no `yield` returned a `stopProducing`. This could happen when we expected the producer to yield elements but the consumer went below the low watermark again. Resulting in two subsequent calls.

# Modification

This PR stores the current demand state in the strategy which let's us avoid flipping the `hasOustandingDemand` state of the sequence.

# Result

Correctly, ensured the call order of `produceMore` and `stopProducing`.

### DIFF
--- a/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducerStrategies.swift
+++ b/Sources/NIOCore/AsyncSequences/NIOAsyncSequenceProducerStrategies.swift
@@ -22,6 +22,7 @@ public enum NIOAsyncSequenceProducerBackPressureStrategies {
     public struct HighLowWatermark: NIOAsyncSequenceProducerBackPressureStrategy {
         private let lowWatermark: Int
         private let highWatermark: Int
+        private var hasOustandingDemand: Bool = true
 
         /// Initializes a new ``NIOAsyncSequenceProducerBackPressureStrategies/HighLowWatermark``.
         ///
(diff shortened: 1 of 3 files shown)


IgorMuzyka:
@rnro hi, not sure who to mention here but you approved changes here so maybe it is you. My question is how soon a new release with this changes merged can be expected?

Lukasa:
We are unlikely to release before next week at the absolute earliest. However, if we released today our `main` branch would break Hummingbird (and a few other packages) because of apple/swift#77459. So we're waiting to see what the path of that patch is going to be.

IgorMuzyka:
@Lukasa thanks, thats an unexpected (for me) turn, but next week as an optimistic release date is great nevertheless. Won't pretend to understand fully the issue you mentioned, but since it concerns those who use swift 6, can't the changes in this issue be wrapped with `#if swift(<=6.0)` ?

Lukasa:
This issue isn't the one that's affected, the issue is a number of other fixes that have landed in `main` as well.

If we aren't going to get a fix to that issue in a timely fashion then we'll need to back those fixes out.

Volodymyr-13:
@Lukasa Hi Cory, can you give me an update on the release date for this? My app crashes a lot because of this, using Vapor, and  they are waiting for the release of NIO. It's been a long wait! Waiting -> Waiting)

Lukasa:
Hello! I'll aim to get you a release this week or next: we'll have to back some stuff out to work around the linked bug.

PR #2953, opened 2024-10-28
Handle Sendability of RemovableChannelHandler

Lukasa:
Motivation:

RemovableChannelHandlers have a large API surface in NIOCore. That API surface is a bit awkward with regard to strict concurrency, and needs some cleanup.

Modifications:

This patch adds some new API that is necessary to safely work with RemovableChannelHandlers, deprecates some API that cannot plausibly be used, and cleans up some other parts of the API.

Result:

Easier to work with RemovableChannelHandlers


FranzBusch (approved):

Lukasa:
Merging over the API breakage, which is expected and safe.

PR #2963, opened 2024-11-02
Fix leaking technique

asdf-bro:
Using `Unmanaged.passRetained` is correct and therefore safe under all compiler optimizations


Lukasa (approved):
Thanks!

PR #2964, opened 2024-11-06
Use condition variable in NIOThreadPool

### LINKED ISSUE #2960: NIOFileSystem operations thread warnings
### Expected behavior

Not to see thread warnings when running `NIOFileSystem` methods in a Task in a SwiftUI app.

### Actual behavior

I believe this is very similar to: https://github.com/apple/swift-nio/issues/2686

That specific issue was fixed by: https://github.com/apple/swift-nio/pull/2687

In `NIOFileSystem`, the usage of `DispatchSemaphore` seems to cause the same issues. I created a reproducible repo that you can check out here: https://github.com/fortmarek/NIOFileSystemThreadWarningReproduction

Note you will need to change the path [here](https://github.com/fortmarek/NIOFileSystemThreadWarningReproduction/blob/main/App/Sources/ContentView.swift#L12) to point to an existing file on your local disk. Once you run the app, you will see the following thread warning:
image

### Steps to reproduce

1. Clone the [repro repo](https://github.com/fortmarek/NIOFileSystemThreadWarningReproduction)
2. Update the path to point to an existing file
3. Run the macOS app
4. See the thread warning

```swift
import SwiftUI
import NIOFileSystem

public struct ContentView: View {
    public init() {}

    public var body: some View {
        Text("Hello, World!")
            .padding()
            .onAppear() {
                Task {
                    let handle = try await NIOFileSystem.FileSystem.shared.openFile(forReadingAt: .init("/Users/marekfort/Downloads/devices.json"), options: .init())
                    try await handle.close()
                }
            }
    }
}
```

`2.76.1`

[nio-diagnose.md](https://github.com/user-attachments/files/17600640/nio-diagnose.md)

glbrntt:
Motivation:

The thread performance checker warns that a threads with QoS are waiting on threads in the NIOThreadPool. We can avoid these warnings by using a condition variable instead.

Modifications:

- Replace the usage of a semaphore and lock in NIOThreadPool with a condition lock.
- The condition value indicates whether the threads have work to do (where work is processing work items or exiting their run loop).

Result:

- Fewer warnings
- Resolves #2960


Sources/NIOPosix/NIOThreadPool.swift
```diff
@@ -133,17 +139,17 @@ public final class NIOThreadPool {
...
-                }
-                return threads
+                self.threads = nil
+
+                // Each thread has work to do: shutdown.
+                return (unlockWith: .hasWork, result: threads)
+
             case .shuttingDown, .stopped:
-                return []
+                return (unlockWith: .hasWork, result: [])
```
Lukasa, line 151: Is this state right?

glbrntt: Good question. I don't think we need to unlock with a value here, if we're already shutting down then we've already signalled there's work for the threads to do, we don't need to do that again.

Lukasa (approved):

PR #2966, opened 2024-11-06
Use Swift 6.0 docs pipeline

rnro:
### Motivation:

Documentation checking catches more issues in Swift 6.0.

### Modifications:

Adopt the Swift 6.0 image and fix the errors.

### Result:

More accurate docs.


Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1787,17 +1787,17 @@ extension EventLoopFuture {
...
-    /// - parameters:
-    ///     - orReplace: the value of the returned `EventLoopFuture` when then resolved future's value is `Optional.some()`.
-    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and the value passed in the `orReplace` parameter.
+    /// - Parameters:
+    ///   - orReplace: the value of the returned `EventLoopFuture` when then resolved future's value is `Optional.some()`.
+    /// - Returns: an new `EventLoopFuture` with new type parameter `NewValue` and the value passed in the `orReplace` parameter.
     @preconcurrency
     @inlinable
     public func unwrap<NewValue: Sendable>(
-        orReplace replacement: NewValue
```
glbrntt, line 1796: I think the docs and code are clearer in terms of `replacement` here

Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -1812,17 +1812,17 @@ extension EventLoopFuture {
...
+    /// - Parameters:
+    ///   - orElse: a closure that returns the value of the returned `EventLoopFuture` when then resolved future's value
     ///         is `Optional.some()`.
-    /// - returns: an new `EventLoopFuture` with new type parameter `NewValue` and with the value returned by the closure
+    /// - Returns: an new `EventLoopFuture` with new type parameter `NewValue` and with the value returned by the closure
     ///     passed in the `orElse` parameter.
     @inlinable
     @preconcurrency
     public func unwrap<NewValue>(
-        orElse callback: @escaping @Sendable () -> NewValue
```
glbrntt, line 1823: Same here, I think `callback` is clearer than `orElse` in the docs etc.

glbrntt (changes requested):
Awesome! Thanks for doing this Rick, I left a couple of comments on small things to fix.

PR #2969, opened 2024-11-07
Add isolated views to EventLoop, Promise, and Future

Lukasa:
Motivation:

Users writing NIO code in a strict concurrency world often need to interact with futures, promises, and event loops. The main interface to these has strict sendability requirements, as it is possible the user is doing so from outside the EventLoop that provides the isolation domain for these types.

However, in many cases the user knows that they are on the isolation domain in question. In that case, they need more capabilities. While they can achieve their goals with NIOLoopBound today, it'd be nice if they had a better option.

Modifications:

- Make EventLoop.Isolated public.
- Make EventLoopFuture.Isolated public.
- Make EventLoopPromise.Isolated public.
- Make all their relevant methods public.
- Move the runtime isolation check from the point of use to the point of construction.
- Make the types non-Sendable to ensure that isolation check is sufficient.
- Add unsafeUnchecked options to create these types when performance matters and correctness is clear.
- Add tests for their behaviour.
- Update the documentation.

Result:

Writing safe code with promises, futures, and event loops is easier.


Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -14,8 +14,14 @@
...
-@usableFromInline
-struct IsolatedEventLoop {
+///
+/// This type is explicitly not `Sendable`. It may only be constructed on an event loop,
+/// using ``EventLoop/assumeIsolated``, and may not subsequently be passed to other isolation
+/// domains.
+///
+/// Using this type relaxes the need to have the closures for ``EventLoop/execute(_:)``,
+/// ``EventLoop/submit(_:)``, and ``EventLoop/scheduleTask(in:_:)`` to be `@Sendable`.
+public struct IsolatedEventLoop {
```
glbrntt, line 24: We need a `NIO` prefix here

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -458,34 +478,70 @@ extension EventLoopFuture {
...
+        }
+
         /// Deliver a successful result to the associated `EventLoopFuture<Value>` object.
         ///
         /// - Parameters:
         ///   - value: The successful result of the operation.
         @inlinable
-        func succeed(_ value: Value) {
-            self._wrapped.futureResult.eventLoop.assertInEventLoop()
+        public func succeed(_ value: Value) {
```
glbrntt, line 544: We don't have a `fail(_:)` method here. `Error`s are always `Sendable` so I suppose it's technically not necessary but it still feels like an odd omission given we have `completeWith(_:)`.

Tests/NIOPosixTests/EventLoopFutureIsolatedTests.swift
```diff
@@ -0,0 +1,354 @@
...
+
+@available(*, unavailable)
+extension SuperNotSendable: Sendable {}
+
+final class EventLoopFutureIsolatedTest: XCTestCase {
+    func testCompletingPromiseWithNonSendableValue() throws {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            try! group.syncShutdownGracefully()
+        }
```
glbrntt, line 31: Not that it matters much, but should we prefer singletons in tests as they save a little boilerplate.

Tests/NIOPosixTests/EventLoopFutureIsolatedTests.swift
```diff
@@ -0,0 +1,354 @@
...
+            let future = promise.futureResult.assumeIsolated()
+            let originalValue = SuperNotSendable()
+
+            // Note that for this test it is _very important_ that all of these
+            // close over `originalValue`. This proves the non-Sendability of
+            // the closure.
+
+            // This block is the main happy path.
+            let newFuture = future.flatMap { result in
+                XCTAssertTrue(originalValue === result)
```
glbrntt, line 157: FWIW: `XCTAssertIdentical` exists for this

glbrntt (changes requested):
This looks great, we're missing a NIO prefix though.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: Yeah, I almost added it and then concluded that it was nicer if we only had methods on this type that dropped the `Sendable` constraint. `completeWith` meets that bar, as it drops the `Sendable` requirement on the type, which _is_ otherwise present.

glbrntt (approved):
LGTM, thanks Cory!

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -25,9 +31,9 @@ struct IsolatedEventLoop {
     }
 
     /// Submit a given task to be executed by the `EventLoop`
+    @available(*, noasync)
     @inlinable
-    func execute(_ task: @escaping () -> Void) {
-        self._wrapped.assertInEventLoop()
```
FranzBusch, line 30: Meta question: I understand why you removed those but should we keep them in in-case somebody really mishandles this type? In release builds these all should go away anyways.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -14,8 +14,14 @@
 
 /// A struct wrapping an ``EventLoop`` that ensures all calls to any method on this struct
 /// are coming from the event loop.
-@usableFromInline
-struct IsolatedEventLoop {
+///
+/// This type is explicitly not `Sendable`. It may only be constructed on an event loop,
+/// using ``EventLoop/assumeIsolated()``, and may not subsequently be passed to other isolation
+/// domains.
```
FranzBusch, line 20: I think other isolation domains is actually too weak of a statement. This type must not be held inside any asynchronous context.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -119,26 +130,54 @@ struct IsolatedEventLoop {
 
     /// Returns the wrapped event loop.
     @inlinable
-    func nonisolated() -> any EventLoop {
+    public func nonisolated() -> any EventLoop {
```
FranzBusch, line 133: Can we make this method also `noasync`

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -119,26 +130,54 @@ struct IsolatedEventLoop {
...
+        return NIOIsolatedEventLoop(self)
+    }
+
+    /// Assumes the calling context is isolated to the event loop.
+    ///
+    /// This version of ``EventLoop/assumeIsolated()`` omits the runtime
+    /// isolation check in release builds. It retains it in debug mode to
+    /// ensure correctness.
+    @inlinable
+    public func assumeIsolatedUnsafeUnchecked() -> NIOIsolatedEventLoop {
```
FranzBusch, line 153: Same here. Can we make this `noasync`

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: I agree. You can always get the underlying promise back by calling `.nonisolated.fail()`

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -458,34 +502,73 @@ extension EventLoopFuture {
...
+    ///
+    /// Using this type relaxes the need to have the promise completion functions accept `Sendable`
+    /// values, as this type can only be handled on the ``EventLoop``.
+    ///
+    /// This type does not offer the full suite of completion functions that ``EventLoopPromise``
+    /// does, as many of those functions do not require `Sendable` values already. It only offers
+    /// versions for the functions that do require `Sendable` types. If you have an
+    /// ``EventLoopPromise/Isolated`` but need a regular ``EventLoopPromise``, use
+    /// ``EventLoopPromise/Isolated/nonisolated()`` to unwrap the value.
+    public struct Isolated {
```
FranzBusch, line 556: I think all methods in this type or in general in all those new types should be marked `noasync`. None of them should ever be held in an async context.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -504,22 +587,43 @@ extension EventLoopPromise {
...
+    /// Returns a variant of this ``EventLoopPromise`` with less strict
+    /// `Sendable` requirements. Can only be called from on the
+    /// ``EventLoop`` to which this ``EventLoopPromise`` is bound, will crash
+    /// if that invariant fails to be met.
+    ///
+    /// This is an unsafe version of ``EventLoopPromise/assumeIsolated()`` which
+    /// omits the runtime check in release builds. This improves performance, but
+    /// should only be used sparingly.
+    @inlinable
+    public func assumeIsolatedUnsafeUnchecked() -> Isolated {
```
FranzBusch, line 622:
```suggestion
    @available(*, noasync)
    public func assumeIsolatedUnsafeUnchecked() -> Isolated {
```

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: I think that's not appropriate either: non-asynchronous contexts are themselves dangerous. As task executors aren't supported today, I think the statement as written is correct. I would now reject any plan to _add_ task executors until these issues are resolved, so I'm inclined to keep the text.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: I don't think we need to: this is the only method that is _always_ safe to call on this type.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: I deliberately didn't do that. This is already marked unsafe-unchecked. I have therefore left it for the possible use-case that someone is actually trying to do the thing we're trying to prevent them doing, and is willing to take on the risk of doing it correctly themselves.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: Same note here, I want to leave the unsafe unchecked methods without the annotation.

glbrntt (approved):

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: What non-asynchronous context are you referring to? I consider inside an actor an asynchronous context since you need to be inside a `Task`.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: I understand your argument my reason is that if a user ever ends up holding this type in an `async` method I really want them to see an error. Even if that place is not where the error happened but they should question how they got the type there in the first place.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
FranzBusch: Okay fair enough. Can we add that to the documentation then.

FranzBusch (approved):

PR #2970, opened 2024-11-12
Fix SemVer major label check

FranzBusch:


glbrntt (approved):
I assume we've rolled this label out to all repos?

FranzBusch:
> I assume we've rolled this label out to all repos?

Yes. The label has been rolled out to all of our repos.

PR #2971, opened 2024-11-12
Add WebSocketFrame.reservedBits OptionSet

adam-fowler:
Add support for treating WebSocketFrame reserved bits as an option set

### Motivation:

I would like to do mask operations on the reserved bits of the WebSocketFrame eg check none are set, check only bits in a mask are set. With the current public interface it is only possible to check the state of one bit at a time.

### Modifications:

Add `WebSocketFrame.ReservedBits` OptionSet
Add computed member `WebSocketFrame.reservedBits`  

### Result:

I can now check the status of multiple bits in one operation eg
`frame.reservedBits.isEmpty`
`frame.reservedBits.itersection([.rsv1, .rsv2]).isEmpty`

### DIFF
--- a/Sources/NIOWebSocket/WebSocketFrame.swift
+++ b/Sources/NIOWebSocket/WebSocketFrame.swift
@@ -420,8 +420,8 @@ extension WebSocketFrame {
     /// WebSocketFrame reserved bits option set
     public struct ReservedBits: OptionSet, Sendable {
         public var rawValue: UInt8
-        
-        @inlinable 
+
+        @inlinable
         public init(rawValue: UInt8) {
             self.rawValue = rawValue
         }
--- a/Sources/NIOWebSocket/WebSocketFrame.swift
+++ b/Sources/NIOWebSocket/WebSocketFrame.swift
@@ -409,7 +409,7 @@ extension WebSocketFrame: CustomDebugStringConvertible {
 
 extension WebSocketFrame {
     /// WebSocketFrame reserved bits option set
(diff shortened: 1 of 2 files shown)


Sources/NIOWebSocket/WebSocketFrame.swift
```diff
@@ -406,3 +406,29 @@ extension WebSocketFrame: CustomDebugStringConvertible {
         "(\(self.description))"
     }
 }
+
+extension WebSocketFrame {
+    /// WebSocketFrame reserved bits option set
+    public struct ReservedBits: OptionSet, Sendable {
+        public var rawValue: UInt8
+
+        public init(rawValue: UInt8) {
```
Lukasa, line 415: Let's make this `@inlinable` and all the `static var`s as well. That unlocks the meaningful performance improvement.

Lukasa (commented):
This mostly looks good, thanks Adam! A quick tweak on the diff.

Sources/NIOWebSocket/WebSocketFrame.swift
adam-fowler: I just noticed the get/set accessors on `firstByte` are not `@inlinable`. Is it worthwhile tagging all of those?

adam-fowler: I tagged them anyway

Lukasa (approved):
Nice one, thanks Adam!

adam-fowler:
aargh swift-format

PR #2979, opened 2024-11-19
Better handle ECONNRESET on connected datagram sockets.

Lukasa:
Motivation:

When a connected datagram socket sends a datagram to a port or host that is not listening, an ICMP Destination Unreachable message may be returned. That message triggers an ECONNRESET to be produced at the socket layer.

On Darwin we handled this well, but on Linux it turned out that this would push us into connection teardown and, eventually, into a crash. Not so good.

To better handle this, we need to distinguish EPOLLERR from EPOLLHUP on datagram sockets. In these cases, we should check whether the socket error was fatal and, if it was not, we should continue our execution having fired the error down the pipeline.

Modifications:

Modify the selector code to distinguish reset and error. Add support for our channels to handle errors.
Have most channels handle errors as resets.
Override the logic for datagram channels to duplicate the logic in readable.
Add a unit test.

Result:

Better datagrams for all.

### DIFF
--- a/Tests/NIOPosixTests/DatagramChannelTests.swift
+++ b/Tests/NIOPosixTests/DatagramChannelTests.swift
@@ -34,6 +34,17 @@ extension Channel {
         }.wait()
     }
 
+    func waitForErrors(count: Int) throws -> [any Error] {
+        try self.pipeline.context(name: "ByteReadRecorder").flatMap { context in
+            if let future = (context.handler as? DatagramReadRecorder<ByteBuffer>)?.notifyForErrors(count) {
+                return future
+            }
+
+            XCTFail("Could not wait for errors")
+            return self.eventLoop.makeSucceededFuture([])
+        }.wait()
+    }
+
     func readCompleteCount() throws -> Int {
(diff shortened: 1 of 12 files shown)


Tests/NIOPosixTests/DatagramChannelTests.swift
```diff
@@ -1712,6 +1744,29 @@ class DatagramChannelTests: XCTestCase {
...
+        // Ok, close on the second channel.
+        XCTAssertNoThrow(try self.secondChannel.close(mode: .all).wait())
+        print("closed")
+
+        // Write again.
+        XCTAssertNoThrow(try self.firstChannel.writeAndFlush(writeData).wait())
+
+        // This should trigger an error.
+        let errors = try self.firstChannel.waitForErrors(count: 1)
+        XCTAssertEqual((errors[0] as? IOError)?.errnoCode, ECONNREFUSED)
```
FranzBusch, line 1767: This is really interesting. So the write succeeds since we send out the data on the socket however when the subsequently get the `ECONNREFUSED` since there was no reachable destination for the packet.

FranzBusch (approved):

PR #2984, opened 2024-11-21
give common blocking functions a clear name

weissi:
### Motivation:

In many situations, for example continuous profiling with sampling profilers, it's important to distinguish between on- and off-CPU work. That's most easily done if there are clear function names/prefixes to grep for that are common places to just wait (off-CPU) and do no real work. In SwiftNIO those are chiefly three places:
1. The `EventLoops` waiting for work
2. `The NIOThreadPool` threads waiting for work
3. `EventLoopFuture.wait()`

This patch makes sure that each of those will have a function with prefix `_blockingWaitFor` in the function name which is easily greppable.

### Modifications:

- Create some non-inlinable functions with `_blockingWaitFor...` that just wait for `...`.

### Result:

SwiftNIO makes operational excellence easier and your SRE team more happy.

### DIFF
--- a/Sources/NIOPosix/System.swift
+++ b/Sources/NIOPosix/System.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2021 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2024 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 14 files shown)


Sources/NIOPosix/System.swift
```diff
@@ -534,14 +567,15 @@ internal enum Posix {
     }
 
     @inline(never)
-    internal static func bind(descriptor: CInt, ptr: UnsafePointer<sockaddr>, bytes: Int) throws {
+    public static func bind(descriptor: CInt, ptr: UnsafePointer<sockaddr>, bytes: Int) throws {
```
Lukasa, line 570: Can we drop the changes to `public` here? They're very misleading.

weissi: @Lukasa there were other, newer (for NIOFileSystem etc) functions that used `public`, so I thought I'd make it consistent with the other new stuff

Lukasa (approved):

PR #2985, opened 2024-11-21
EventLoopFuture.waitSpinningRunLoop()

weissi:
### Motivation:

In some (probably niche) scenarios, especially in pre-Concurrency UI applications on Darwin, it can be useful to wait for an a value whilst still running the current `RunLoop`. That allows the UI and other things to work whilst we're waiting for a future to complete.

### Modifications:

- Add `NIOFoundationCompat.EventLoopFuture.waitSpinningRunLoop()`.

### Result:

Better compatibility with Cocoa.

### DIFF
--- a/Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
+++ b/Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
@@ -1,0 +1,72 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Atomics
+import Foundation
+import NIOConcurrencyHelpers
+import NIOCore
+
+extension EventLoopFuture {
+    /// Wait for the resolution of this `EventLoopFuture` by spinning `RunLoop.current` in `mode` until the future
+    /// resolves. The calling thread will be blocked albeit running `RunLoop.current`.
+    ///
+    /// If the `EventLoopFuture` resolves with a value, that value is returned from `waitSpinningRunLoop()`. If
+    /// the `EventLoopFuture` resolves with an error, that error will be thrown instead.
+    /// `waitSpinningRunLoop()` will block whatever thread it is called on, so it must not be called on event loop
+    /// threads: it is primarily useful for testing, or for building interfaces between blocking
+    /// and non-blocking code.
+    ///
+    /// This is also forbidden in async contexts: prefer `EventLoopFuture/get()`.
+    ///
+    /// - Note: The `Value` must be `Sendable` since it is shared outside of the isolation domain of the event loop.
+    ///
+    /// - Returns: The value of the `EventLoopFuture` when it completes.
+    /// - Throws: The error value of the `EventLoopFuture` if it errors.
+    @available(*, noasync, message: "waitSpinningRunLoop() can block indefinitely, prefer get()", renamed: "get()")
+    @inlinable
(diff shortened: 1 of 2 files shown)


Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// and non-blocking code.
+    ///
+    /// This is also forbidden in async contexts: prefer `EventLoopFuture/get()`.
+    ///
+    /// - Note: The `Value` must be `Sendable` since it is shared outside of the isolation domain of the event loop.
+    ///
+    /// - Returns: The value of the `EventLoopFuture` when it completes.
+    /// - Throws: The error value of the `EventLoopFuture` if it errors.
+    @available(*, noasync, message: "waitSpinningRunLoop() can block indefinitely, prefer get()", renamed: "get()")
+    @preconcurrency
```
Lukasa, line 37: This isn't `@preconcurrency`.

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
```diff
@@ -0,0 +1,77 @@
...
+    public func waitSpinningRunLoop(
+        inMode mode: RunLoop.Mode = .default,
+        file: StaticString = #file,
+        line: UInt = #line
+    ) throws -> Value where Value: Sendable {
+        try self._blockingWaitForFutureCompletion(mode: mode, file: file, line: line)
+    }
+
+    @inlinable
+    func _blockingWaitForFutureCompletion(
```
Lukasa, line 48: Per your blocking PR elsewhere, should this be `@inlinable @inline(never)`?

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
```diff
@@ -0,0 +1,77 @@
...
+
+        let value: NIOLockedValueBox<Result<Value, any Error>?> = NIOLockedValueBox(nil)
+        self.whenComplete { result in
+            value.withLockedValue { value in
+                value = result
+            }
+        }
+
+        while value.withLockedValue({ $0 }) == nil {
+            _ = runLoop.run(mode: mode, before: Date().addingTimeInterval(0.01))
```
Lukasa, line 65: Should this time interval be configurable?

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        return try value.withLockedValue { value in
+            switch value! {
+            case .success(let result):
+                return result
+            case .failure(let error):
+                throw error
+            }
```
Lukasa, line 74: This is `value!.get()`.

Lukasa (changes requested):
Thanks, this generally looks good! A few notes in the diff.

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
weissi: thanks, ys

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
weissi: Don't think so because what I _really_ want to do is

```
while value.withLockedValue({ $0 }) == nil {
 _ = runLoop.run(mode: mode, before: .distantFuture)
}
```

and then use `runLoop.schedule { /* just poking the run loop */ }` from the future callback when the future is succeeded.

Unfortunately, `RunLoop` says that it's not thread-safe at all. I don't think that's actually true for functions that you clearly _need_ from other threads (like `schedule`) but I'm waiting for confirmation from @parkera

Sources/NIOFoundationCompat/WaitSpinningRunLoop.swift
weissi: ta!

Lukasa (approved):
Fab, thanks @weissi!

Lukasa:
Note that the unit tests aren't passing.

PR #2994, opened 2024-11-23
fix almost all Sendable warnings

weissi:
### Motivation:

Opening the `swift-nio` repository made me warning blind because there were always so many trivially fixable warnings about things that were correct but cannot be understood by the compiler.

### Modifications:

Fix all the sendable warnings that popped up, except for one test where `NIOLockedValueBox<Thread?>` isn't sendable because `Foundation.Thread` seemingly isn't `Sendable` which is odd. Guessing that'll be fixed on their end.

### Result:

- Fewer warnings
- Less warning-blindness
- More checks

### DIFF
--- a/Sources/NIOCore/ChannelPipeline.swift
+++ b/Sources/NIOCore/ChannelPipeline.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2017-2024 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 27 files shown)


dnadoba (commented):
And that without any unsafe/unchecked swift! Nicely done!

Lukasa:
The intention here has been to wait until we could use the isolated views on ELP and ELF to do this work, to minimise the amount of runtime isolation checking. As the principal advocate for avoiding `preconditionOnEventLoop`, would you prefer to silence these warnings at that cost?

Lukasa:
That comment applies to the changes in the production code, naturally: the changes in the tests are going to be just fine either way.

weissi:
> The intention here has been to wait until we could use the isolated views on ELP and ELF to do this work, to minimise the amount of runtime isolation checking. As the principal advocate for avoiding `preconditionOnEventLoop`, would you prefer to silence these warnings at that cost?

@Lukasa I really would want to see these warnings go away because there are just too many of them, I lose track of others (point in case this one: https://github.com/apple/swift-nio/pull/2995#issuecomment-2497223697).

I hear you w.r.t. to the `preconditionInEventLoop` but here's why I think we should still do it: At first, I only wanted to do tests & the upgraders (because they're not super perf critical). But then I noticed that's like 99% of all of them. None of the code I changed is in actually perf-critical code or changes any of the properties of the code.

And once we have the isolated views (which are fab btw), we can still replace all the code, it's super easy to spot by grepping for `NIOLoopBound`, right?

tl;dr: This improves things and I mandate that it doesn't regress anything. So I don't think there's any point in waiting.

Lukasa (approved):

PR #2995, opened 2024-11-24
fix warnings: syncShutdownGracefully not available in async contexts

weissi:
### Motivation:

Warnings are annoying. Companion of #2994 

### Modifications:

Fix all the `syncShutdownGracefully` not being available in `async` contexts warnings.

### Result:

Everybody happier.

### DIFF
--- a/Tests/NIOPosixTests/SerialExecutorTests.swift
+++ b/Tests/NIOPosixTests/SerialExecutorTests.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2023-2024 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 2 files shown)


Tests/NIOPosixTests/SerialExecutorTests.swift
```diff
@@ -47,39 +47,43 @@ actor EventLoopBoundActor {
...
+            throw error
+        }
     }
 
     func testAssumeIsolation() async throws {
         #if compiler(<6.0)
         throw XCTSkip("Custom executors are only supported in 5.9")
         #else
 
-        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
```
weissi, line 81: there was actually a bug here which would always leak the group...

Lukasa:
```
/__w/swift-nio/swift-nio/Tests/NIOPosixTests/SerialExecutorTests.swift:59:13: error: initialization of immutable value 'eventLoopGroup' was never used; consider replacing with assignment to '_' or removing it
        let eventLoopGroup = self.group
        ~~~~^~~~~~~~~~~~~~
```

PR #2999, opened 2024-11-25
Add Cxx interop swift settings to CI

rnro:
### Motivation:

At the moment the Cxx interoperability CI workflow doesn't actually test Cxx interoperability.

### Modifications:

The script now looks for the precise way that the `Package.swift` currently formats the target entry and appends a `swiftSettings` entry to it.

This approach is pretty brittle but has the advantage that it's the same on Linux and Darwin. If this formatting changes too often then we could give up on this and assume the availability of gnu-sed instead (and make sure it's present on our CI containers).

### Result:

All Cxx interoperability checks will check what they are named for.

### DIFF
--- a/scripts/check-cxx-interop-compatibility.sh
+++ b/scripts/check-cxx-interop-compatibility.sh
@@ -30,25 +30,18 @@ package_name=$(swift package dump-package | jq -r '.name')
 
 cd "$working_dir"
 swift package init
+
+{
+  echo "let swiftSettings: [SwiftSetting] = [.interoperabilityMode(.Cxx)]"
+  echo "for target in package.targets { target.swiftSettings = swiftSettings }"
+} >> Package.swift
+
 echo "package.dependencies.append(.package(path: \"$source_dir\"))" >> Package.swift
 
 for product in $library_products; do
   echo "package.targets.first!.dependencies.append(.product(name: \"$product\", package: \"$package_name\"))" >> Package.swift
   echo "import $product" >> "$source_file"
 done
-
-# A platform independent (but brittle) way to add the swift settings block
-sed '/name: "'"${project_name}"'")/a\
-        swiftSettings: [\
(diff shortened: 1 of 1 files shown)


scripts/check-cxx-interop-compatibility.sh
```diff
@@ -37,6 +37,19 @@ for product in $library_products; do
   echo "import $product" >> "$source_file"
 done
 
+# A platform independent (but brittle) way to add the swift settings block
```
glbrntt, line 40: Can't we use the fact that `Package` has reference semantics and just add settings after the fact?

i.e something like:

```swift
let package = Package(...)

// Generate this bit:
for target in package.targets {
  target.swiftSettings = [....]
}

File: scripts/check-cxx-interop-compatibility.sh [hunk 2]
```diff
@@ -30,6 +30,12 @@ package_name=$(swift package dump-package | jq -r '.name')
 
 cd "$working_dir"
 swift package init
+
+{
+  echo "let swiftSettings: [SwiftSetting] = [.interoperabilityMode(.Cxx)]"
+  echo "for target in package.targets { target.swiftSettings = swiftSettings }"
```
czechboy0, line 36: This overwrites any existing build settings, could we do `target.swiftSettings.append(contentsOf: swiftSettings)` instead?

rnro: I wondered about that approach myself and went back and forth. I landed on this approach because I thought it kept the test simple and targeted on the particular facet of the build we're interested in. Do you have specific reasons to use the combination?

czechboy0: Yes, downgrading the Swift language mode to 5 is done this way, so the code above breaks that and builds it in Swift 6 mode (for packages with tools-version 6.0)

PR #3001, opened 2024-11-26
Remove noasync from NIOFileHandle

glbrntt:
Motivation:

In #2598, `NIOFileHandle` was annotated with `noasync` in a few places. This is, unfortunately, a breaking change.

Modifications:

- Remove the `noasync` annotations

Result:

Adopters aren't broken

### DIFF
--- a/Sources/NIOCore/FileHandle.swift
+++ b/Sources/NIOCore/FileHandle.swift
@@ -387,6 +387,7 @@ extension NIOFileHandle {
     ///   - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
     ///   - mode: Access mode. Default mode is `.read`.
     ///   - flags: Additional POSIX flags.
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

Sources/NIOCore/FileHandle.swift
```diff
@@ -388,7 +387,6 @@ extension NIOFileHandle {
     ///   - path: The path of the file to open. The ownership of the file descriptor is transferred to this `NIOFileHandle` and so it will be closed once `close` is called.
     ///   - mode: Access mode. Default mode is `.read`.
     ///   - flags: Additional POSIX flags.
-    @available(*, noasync, message: "This method may block the calling thread")
```
weissi, line 391: @glbrntt We can keep it on the new methods though, no?

glbrntt: Yeah, you're absolutely right.

PR #3003, opened 2024-11-26
Prevent crash in Happy Eyeballs Resolver

Lukasa:
Motivation:

In vanishingly rare situations it is possible for the AAAA results to come in on the same tick as the resolution delay timer completes. In those cases, depending on the ordering of the tasks, we can get situations where the resolution delay timer completion causes a crash.

Modifications:

Tolerate receiving the resolution delay timer after resolution completes.

Result:

Fewer crashes


weissi (approved):
Awesome, thank you!

PR #3006, opened 2024-11-27
Avoid converting array holding existentials

Lukasa:
Motivation:

When converting an Array that holds existentials, it is necessary for Swift to allocate a new Array and copy the elements into it, so that it can fix up their existential boxes. We forced this to happen with taking a `[ChannelHandler]` to `[ChannelHandler & Sendable]`, without rewriting the other functions.

The result of that was that the other functions needed to convert the channel handler types, causing extra allocations in this path.

Modifications:

Propagate the constraint down to the point where we iterate the Array.

Result:

Allocations reduced.


Lukasa:
Of course, this change just moved the cost into the SyncOperations instead. So they now need to have totally separate codepaths to avoid reprojecting the array.

dnadoba:
> it is necessary for Swift to allocate a new Array [...] We forced this to happen with taking a [ChannelHandler] to [ChannelHandler & Sendable]

That should be same type at runtime as Sendable is just a marker protocol. @Lukasa did you opened a swiftlang issue?

Lukasa:
I didn't, but I don't think the fact that it's the same type at runtime is quite true.

Issue #2933, opened 2024-10-17
`removeItem` performance


pepicrft:
👋🏼 Hi,

First of all, thanks for this amazing work.

We are using the file-system IO components at [Tuist](https://github.com/tuist/tuist), and I'd like to bring up something that we notice regarding the performance of the `NIOFileSystem.FileSystem.removeItem` function.

When we run it against a large directory with many nested directories and and files, we noticed the performance is much worse than using `FileManager` or just `rm -rf` in a Linux environment. We benchmarked the performance with [hyperfine](https://github.com/sharkdp/hyperfine)  and the results are below.

Looking at the implementation, it seems that we are not parallelising at all, so I was wondering if this is something intentional, or if it's ok to refactor that code to run as much as possible in parallel.

#### With NIOFileSystem

Benchmark
- Time (mean ± σ):     67.939 s ±  4.450 s    [User: 61.906 s, System: 3.205 s]
- Range (min … max):   63.684 s … 77.757 s    10 runs

#### With Foundation's FileManager

Benchmark:
- Time (mean ± σ):      5.857 s ±  0.251 s    [User: 1.430 s, System: 2.067 s]
- Range (min … max):    5.419 s …  6.190 s    10 runs

#### With `rm -rf`

Benchmark:
  Time (mean ± σ):      3.150 s ±  0.386 s    [User: 0.022 s, System: 1.666 s]
  Range (min … max):    2.225 s …  3.585 s    10 runs

FranzBusch:
Thanks for filling this! This is definitely something that we would take a patch for and parallelise the removal similar to how we did for copying in https://github.com/apple/swift-nio/pull/2806.

mimischi:
@pepicrft I've raised a PR (#3008) to address the issue of handling removal of items sequentially. I seem to be unable to reproduce your results though. The existing logic in `main` appears to be performing _ok_, when compared to `rm -rf` (albeit slower).

If you still have your test setup, would you mind sharing that, so I can try and reproduce that?

pepicrft:
> @pepicrft I've raised a PR (#3008) to address the issue of handling removal of items sequentially. I seem to be unable to reproduce your results though. The existing logic in `main` appears to be performing _ok_, when compared to `rm -rf` (albeit slower).
> 
> If you still have your test setup, would you mind sharing that, so I can try and reproduce that?

[...]

PR #3008, opened 2024-11-29
Add parallel removal of items

mimischi:
Introduce parallel removal of items, instead of only doing a sequential removal (and discovery) one-by-one

### Motivation:

#2933 mentioned this function call is slow. If we attempt to delete a directory with lots of subdirectories, we end up discovering each subdirectory one by one on a single thread.

### Modifications:

This PR refactors the logic to remove items. Instead of sequentially discovering and deleting items one by one, we concurrently traverse all directories that we find and then sequentially issue file deletions.

This new deletion logic is currently unbounded, and will spawn a bunch of tasks no matter the underlying system.

### Result:

Benchmark using `--configration=release` for both the `main` and `parallel-removal` branches. This new solution does not appear to be _that much_ faster than the current solution on `main`. Curious to hear where I'm doing things in a suboptimal fashion:
```
$ rm -rf .build && git switch main && swift build --configuration=release && cp .build/release/NIOTestCLI NIOTestCLI_main
[...]
Building for production...
[134/134] Linking NIOPerformanceTester
Build complete! (32.65s)
$ rm -rf .build && git switch parallel-removal && swift build --configuration=release
[...]
Building for production...
[134/134] Linking NIOPerformanceTester
Build complete! (32.19s)
$ hyperfine --max-runs 10 --prepare "rm -rf /tmp/swift-nio && cp -r $(pwd) /tmp/swift-nio" "./NIOTestCLI_main /tmp/swift-nio" ".build/release/NIOTestCLI /tmp/swift-nio" "rm -rf /tmp/swift-nio"
Benchmark 1: ./NIOTestCLI_main /tmp/swift-nio
  Time (mean ± σ):      3.019 s ±  0.032 s    [User: 0.601 s, System: 11.123 s]
  Range (min … max):    2.985 s …  3.072 s    10 runs

Benchmark 2: .build/release/NIOTestCLI /tmp/swift-nio
  Time (mean ± σ):      2.295 s ±  0.065 s    [User: 0.673 s, System: 7.677 s]
  Range (min … max):    2.212 s …  2.370 s    10 runs

Benchmark 3: rm -rf /tmp/swift-nio
  Time (mean ± σ):      1.517 s ±  0.037 s    [User: 0.015 s, System: 1.442 s]
  Range (min … max):    1.480 s …  1.590 s    10 runs

Summary
  rm -rf /tmp/swift-nio ran
    1.51 ± 0.06 times faster than .build/release/NIOTestCLI /tmp/swift-nio
    1.99 ± 0.05 times faster than ./NIOTestCLI_main /tmp/swift-nio
```

Resolves https://github.com/apple/swift-nio/issues/2933


mimischi:
I'll need a hand with the API breakage check. I can't reproduce the result locally, and I'm not entirely clear how to prevent the breakage reported in CI—would we not add a new `removeItem` method with overloaded arguments to `FileSystemProtocol.swift` to prevent this?

> API breakage: func FileSystemProtocol.removeItem(at:strategy:recursively:) has been added as a protocol requirement

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -599,7 +605,29 @@ extension FileSystemProtocol {
...
+    ///   - path: The path to delete.
+    ///   - removalStrategy: Whether to delete files sequentially (one-by-one), or perform a
+    ///       concurrent scan of the tree at `path` and delete files when they are found.
+    /// - Returns: The number of deleted items which may be zero if `path` did not exist.
+    @discardableResult
+    public func removeItem(
+        at path: FilePath,
+        strategy removalStrategy: RemovalStrategy
+    ) async throws -> Int {
+        try await self.removeItem(at: path, strategy: removalStrategy, recursively: true)
```
glbrntt, line 630: We want the strategy and the recursion behaviour to both be optional. At the moment it's not possible to call `removeItem(at: "foo", recursive: true)` as the strategy isn't defaulted.

Annoyingly I think we need have the following:

- `removeItem(at:strategy:recursively:)` as the requirement, and
- `removeItem(at:)` and
- `removeItem(at:strategy:)`, (defaults recursively) and
- `removeItem(at:recursively:)` (defaults strategy) as extensions.

glbrntt (changes requested):
This looks great so far, I left a couple of notes inline.

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
```diff
@@ -0,0 +1,81 @@
...
+                                // dependants.
+                                let numberOfDeletedFiles = try await self.removeOneItem(at: entry.path)
+                                atomicCounter.wrappingIncrement(by: numberOfDeletedFiles, ordering: .relaxed)
+                            }
+                        default:
+                            // Delete anything that is not a directory.
+                            group.addTask {
+                                let numberOfDeletedFiles = try await self.removeOneItem(at: entry.path)
+                                atomicCounter.wrappingIncrement(by: numberOfDeletedFiles, ordering: .relaxed)
+                            }
```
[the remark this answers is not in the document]
glbrntt: You're not missing anything, I was just curious as adding tasks isn't free and I wasn't sure _how much_ we were gaining.

Sources/NIOFileSystem/FileSystemProtocol.swift
mimischi: Having a hard time replicating the API breakage failures locally, and my CI-based-debugging is not the most efficient here. Reached out to Rick for some guidance.

I think I've done as you've suggested (`removeItem(at:strategy:recursively:)` as the only requirement, and the other three as extensions to `FileSystemProtocol`), but I'm still seeing issues. Don't I also need an extension for `removeItem(at:strategy:recursively:)` to define a concrete implementation of that variant?

Sources/NIOFileSystem/Internal/Concurrency Primitives/TokenBucket.swift
```diff
@@ -0,0 +1,88 @@
...
+            return
+        }
+        self.lock.unlock()
+
+        await withCheckedContinuation {
+            self.lock.lock()
+            self.waiters.append($0)
+            self.lock.unlock()
+        }
+    }
```
mimischi, line 76: Trying to use `{ defer self.lock.unlock() }` in this call leads to a deadlock. I'm probably not seeing something here, but thinking that `await withCheckedContinuation` isn't compatible with this approach? Seems a tad repetitive to have `self.lock.unlock()` on lines 66 and 69 (albeit in different branches).

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension FileSystem {
+    /// Iterative implementation of a recursive parallel removal of objects
+    /// found within `path`.
+    func removeConcurrently(
```
mimischi, line 21: This could probably live inside of `NIOFileSystem/FileSystem.swift`. I've got no  hard opinions if this should be here or there.

Sources/NIOFileSystem/FileSystemProtocol.swift
mimischi: Fixed this at last. Thanks @rnro for the pointers!

mimischi:
Updated the PRs description with the latest benchmarks

Sources/NIOFileSystem/IOStrategy.swift
```diff
@@ -0,0 +1,211 @@
...
+    /// Traversal of directories and removal of files will be done sequentially without any
+    /// parallelization.
+    public static let sequential: Self = Self(.sequential)
+
+    /// Allow for one or more directory scans to run at the same time. Removal of files will happen
+    /// on asynchronous tasks in parallel.
+    ///
+    /// Setting `maxDescriptors` to 1, will limit the speed of directory discovery. Deletion of
+    /// files within that directory will run in parallel, but discovery of subdirectories will be
+    /// limited to one at a time.
```
mimischi, line 187: Writing these out, and having talked to George earlier, I wonder if we should also update `FileSystem.removeItemSequentially` to perform file deletion in parallel. All we are controlling with `RemovalStrategy` is the number of directory handles that are held by one process at a time. Deletion of items does not require depletion of file descriptors, and we could spin out additional tasks to remove items when users prefer to use `.sequential`.

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
```diff
@@ -0,0 +1,85 @@
...
+    private func discoverItemsInTree(
+        at path: FilePath,
+        _ bucket: TokenBucket
+    ) async throws -> Int {
+        // Discover current directory and find all files/directories. Free up
+        // the handle as fast as possible.
+        let (directoriesToRecurseInto, itemsToDelete) = try await bucket.withToken {
+            try await self.withDirectoryHandle(atPath: path) { directory in
+                var subdirectories: [DirectoryEntry] = []
+                var itemsInDirectory: [DirectoryEntry] = []
```
mimischi, line 41: These could be `[FilePath]` instead, because we do not need to hold onto the `DirectoryEntry` reference afterwards. Would also align it with how we do things in `removeItemSequentially`

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// on each directory that we find, as soon as the file descriptor for
+    /// `path` has been closed; also delete all files that we come across.
+    private func discoverItemsInTree(
+        at path: FilePath,
+        _ bucket: TokenBucket
+    ) async throws -> Int {
+        // Discover current directory and find all files/directories. Free up
+        // the handle as fast as possible.
+        let (directoriesToRecurseInto, itemsToDelete) = try await bucket.withToken {
+            try await self.withDirectoryHandle(atPath: path) { directory in
```
mimischi, line 39: We could actually re-use part of `removeItemSequentially` here. The entire `try away self.withDirectoryHandle {}` piece can be made the same.

Sources/NIOFileSystem/IOStrategy.swift
mimischi: I've got changes for this locally that makes `removeItemSequentially` about 1.2x faster than the current implementation. Not a huge difference, but a simple change.

```
$ hyperfine --runs 10 --warmup 2 --prepare "rm -rf /tmp/swNIOTestCLI --sequential /tmp/swift-nio" ".build/release/NIOTestCLI --parallel /ttmp/swift-nio" "rm -rf /tmp/swift-nio"
Benchmark 1: .build/release/NIOTestCLI --sequential /tmp/swift-nio
  Time (mean ± σ):      2.298 s ±  0.065 s    [User: 0.504 s, System: 9.054 s]
  Range (min … max):    2.184 s …  2.412 s    10 runs

Benchmark 2: .build/release/NIOTestCLI --parallel /tmp/swift-nio
  Time (mean ± σ):     961.5 ms ±  57.0 ms    [User: 327.1 ms, System: 5753.0 ms
  Range (min … max):   909.1 ms … 1090.0 ms    10 runs

Benchmark 3: .build/release/NIOTestCLI --filemanager /tmp/swift-nio
  Time (mean ± σ):      1.340 s ±  0.095 s    [User: 0.021 s, System: 1.226 s]
  Range (min … max):    1.281 s …  1.599 s    10 runs

Benchmark 4: rm -rf /tmp/swift-nio
  Time (mean ± σ):      1.462 s ±  0.032 s    [User: 0.016 s, System: 1.382 s]
  Range (min … max):    1.413 s …  1.501 s    10 runs

Summary
  .build/release/NIOTestCLI --parallel /tmp/swift-nio ran
    1.39 ± 0.13 times faster than .build/release/NIOTestCLI --filemanager /tmp/swift-nio
    1.52 ± 0.10 times faster than rm -rf /tmp/swift-nio
    2.39 ± 0.16 times faster than .build/release/NIOTestCLI --sequential /tmp/swift-nio
$ tree -a . | grep directories
3768 directories, 12118 files
```

Sources/NIOFileSystem/Internal/Concurrency Primitives/TokenBucket.swift
```diff
@@ -0,0 +1,88 @@
...
+///
+/// This is an adaptation of the TokenBucket found in Swift Package Manager.
+/// Instead of using an ``actor``, we define a class and limit access through
+/// ``NIOLock``.
+
+import DequeModule
+import NIOConcurrencyHelpers
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+class TokenBucket {
```
mimischi, line 37: Reviewing this myself now, it looks like I added the imports between the `class` definition and its comment. Surprised that no linter is barfing at me here.

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
```diff
@@ -0,0 +1,85 @@
...
+                group.addTask {
+                    try await self.discoverItemsInTree(at: directory.path, bucket)
+                }
+            }
+
+            // Await task groups to finish and sum all items deleted so far.
+            var numberOfDeletedItems: Int = 0
+            for try await result in group {
+                numberOfDeletedItems += result
+            }
```
glbrntt, line 77: FWIW there's a shorthand for this:

```suggestion
            var numberOfDeletedItems = try await group.reduce(0, +)
```

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
```diff
@@ -305,7 +305,11 @@ final class FileHandleTests: XCTestCase {
     func testWriteAndReadUnseekableFile() async throws {
         let privateTempDirPath = try await FileSystem.shared.createTemporaryDirectory(template: "test-XXX")
         self.addTeardownBlock {
-            try await FileSystem.shared.removeItem(at: privateTempDirPath, recursively: true)
+            try await FileSystem.shared.removeItem(
```
glbrntt, line 308: The changes to add `.platformDefault` shouldn't be required as it's the default? (It's useful to _not_ make these changes to make sure we have an appropriate default.)

glbrntt (changes requested):
This is looking really good so far!

Sources/NIOFileSystem/IOStrategy.swift
mimischi: Fair enough. Good point on users likely being surprised about this behavior when using `.sequential`.

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
mimischi: TIL. Thanks!

Sources/NIOFileSystem/Internal/Concurrency Primitives/TokenBucket.swift
mimischi: Ah, makes sense. Also seeing now how your suggestion is a more condense variation of mine. Thanks!

Sources/NIOFileSystem/Internal/Concurrency Primitives/TokenBucket.swift
```diff
@@ -0,0 +1,88 @@
...
+        await withCheckedContinuation {
+            self.lock.lock()
+            self.waiters.append($0)
+            self.lock.unlock()
+        }
+    }
+
+    func returnToken() {
+        self.lock.lock()
+        defer { self.lock.unlock() }
```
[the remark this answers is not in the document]
mimischi: Got that to work. Not convinced it's more readable than this simple two-liner though, but happy to adapt if that's the way.

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -393,6 +396,42 @@ public struct FileSystem: Sendable, FileSystemProtocol {
     public func removeItem(
         at path: FilePath,
         recursively removeItemRecursively: Bool
+    ) async throws -> Int {
+        try await self.removeItem(at: path, strategy: .platformDefault, recursively: removeItemRecursively)
+    }
```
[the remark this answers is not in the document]
mimischi: Breadcrumb: I added this believing we need it to not break the API. We now think this might just be the checker that's overly sensitive and throws a false-positive at us.

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -599,7 +606,80 @@ extension FileSystemProtocol {
...
+    ///       concurrent scan of the tree at `path` and delete files when they are found.
+    ///   - removeItemRecursively: Whether or not to remove items recursively.
+    /// - Returns: The number of deleted items which may be zero if `path` did not exist.
+    @discardableResult
+    public func removeItem(
+        at path: FilePath,
+        strategy removalStrategy: RemovalStrategy,
+        recursively removeItemRecursively: Bool
+    ) async throws -> Int {
+        try await self.removeItem(at: path, strategy: removalStrategy, recursively: removeItemRecursively)
```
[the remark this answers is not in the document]
mimischi: Breadcrumb: I added this believing we need it to not break the API. We now think this might just be the checker that's overly sensitive and throws a false-positive at us.

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
mimischi: I might have changes these tests after adding only some of the protocol extensions, as I've had those tests fail. Now with unnecessary implementations removed, tests are still passing. Going to revert these changes (including the format change—where I'm again surprised the Swift formatter isn't trying to remove the line breaks that are now not required).

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
glbrntt: I don't mind either way to be honest.

Sources/NIOFileSystem/Internal/Concurrency Primitives/TokenBucket.swift
```diff
@@ -0,0 +1,88 @@
...
+import NIOConcurrencyHelpers
+
+/// Type modeled after a "token bucket" pattern, which is similar to a semaphore, but is built with
+/// Swift Concurrency primitives.
+///
+/// This is an adaptation of the TokenBucket found in Swift Package Manager.
+/// Instead of using an ``actor``, we define a class and limit access through
+/// ``NIOLock``.
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+final class TokenBucket {
```
glbrntt, line 36: I think this should be `@unchecked Sendable`

glbrntt (commented):
One nit but otherwise this looks great!

Sources/NIOFileSystem/Internal/ParallelRemoval.swift
mimischi: I went ahead and refactored it in [338ff35](https://github.com/apple/swift-nio/pull/3008/commits/338ff350730d9193eb03dc38d47cffc55f3964cd)

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -438,29 +438,37 @@ public struct FileSystem: Sendable, FileSystemProtocol {
         }
     }
 
-    @discardableResult
-    private func removeItemSequentially(
-        at path: FilePath
-    ) async throws -> Int {
-        var (subdirectories, filesRemoved) = try await self.withDirectoryHandle(
+    func _collectItemsInDirectory(at path: FilePath) async throws -> ([FilePath], [FilePath]) {
```
glbrntt, line 441: Can you add labels to the tuples? It's too easy to mix them up otherwise.

glbrntt: Actually, do we even need to split them? We could just return the directory entries and the caller can then iterate over them and do as they need. This would avoid allocating an array unnecessarily as well (not that I think it'll have any meaningful impact on perf).

mimischi: I can get behind adding labels. If we only return a single array, then the callers would need to perform the sorting themselves, and thus replicate the bulk of what this utility does. I'm not wed to this existing, but I think if we have it then might as well use it to remove some level of duplication?

mimischi: Had a quick discussion offline. Reverted the addition of `_collectItemsInDirectory` in [e9df08e](https://github.com/apple/swift-nio/pull/3008/commits/e9df08e1c69778b5afeb743d5dc050b507a62d3b)

glbrntt (approved):
This LGTM, thanks @mimischi!

glbrntt:
API break is expected; we are changing the requirement to add the strategy

```
3 breaking changes detected in _NIOFileSystem:
  💔 API breakage: func FileSystemProtocol.removeItem(at:strategy:recursively:) has been added as a protocol requirement
```

These fall out of that:

```
  💔 API breakage: func FileSystem.removeItem(at:recursively:) has parameter 1 type change from Swift.Bool to _NIOFileSystem.RemovalStrategy
  💔 API breakage: func FileSystem.removeItem(at:recursively:) has been renamed to func removeItem(at:strategy:recursively:)
```

PR #3013, opened 2024-12-03
Introduce JSON matrix workflow

rnro:
Introduce and adopt a new method for defining test matrices, `swift_test_matrix.yml`.

⚠️ Any external adopters of the unit tests, Cxx interop and benchmarks workflows are automatically opted in to use the new infrastructure.

### Motivation:

* The current matrix workflow has the limitation that it only supports pre-defined sets of variables which are explored in the test matrix. At the moment this is a pre-defined set of Swift versions on Linux and Windows.
* Adding more means hard-coding them at multiple levels of the workflow hierarchy.
* Currently skipped Windows matrix jobs show up as successes in the GitHub UI leading to a misleading impression of good coverage.

### Modifications:

Introduce and adopt a new method for defining test matrices, `swift_test_matrix.yml`. The new method is based around the approach of defining the test matrix via a JSON object which may be supplied via an input string from another workflow or from a file on-disk in a repository.

Taking this approach means that we have the ability to add new targets to the matrix simply by adding new elements to the JSON array, increasing flexibility and the scope for future growth.

The unit tests, Cxx interop and benchmarks workflows are all modified to use the new approach, this opts-in all downstream adopters. This should be transparent to all downstream adopters.

In order to unify the Linux and Windows jobs I removed the use of the `container:` GitHub Actions affordance in the Linux jobs which transparently means all steps are executed within the nested container. Instead we must manually call in to the docker container which complicates scripting a little. I tested to see if doing this slowed down the jobs (perhaps GitHub was caching the docker images more intelligently) but it does not.

This approach follows the pattern of @FranzBusch 's open PR https://github.com/apple/swift-nio/pull/2942

### Result:

* More flexible test matrix definitions
* No more false-passes for disabled Windows targets

### DIFF
--- a/scripts/generate_matrix.sh
+++ b/scripts/generate_matrix.sh
@@ -1,0 +1,153 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftNIO open source project
+##
+## Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+# Parameters
+linux_command="$MATRIX_LINUX_COMMAND"  # required if any Linux pipeline is enabled
+linux_setup_command="$MATRIX_LINUX_SETUP_COMMAND"
+linux_5_9_enabled="${MATRIX_LINUX_5_9_ENABLED:=true}"
+linux_5_9_command_arguments="$MATRIX_LINUX_5_9_COMMAND_ARGUMENTS"
+linux_5_10_enabled="${MATRIX_LINUX_5_10_ENABLED:=true}"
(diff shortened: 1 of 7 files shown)


scripts/generate_matrix.sh
```diff
@@ -0,0 +1,153 @@
...
+    --arg runner "$linux_runner" \
+    '.swift[.swift| length] |= . + { "name": "nightly-main", "image": $container_image, "swift_version": "nightly-main", "platform": "Linux", "command": $command, "command_arguments": $command_arguments, "setup_command": $setup_command, "runner": $runner}')
+fi
+
+## Windows
+if [[ "$windows_6_0_enabled" == "true" || "$windows_nightly_6_0_enabled" == "true" || "$windows_nightly_main_enabled" == "true" ]]; then
+      if [[ -z "$windows_command" ]]; then
+        echo "No windows command defined"; exit 1
+      fi
+fi
```
glbrntt, line 121: nit: inconsistent indentation

glbrntt (approved):
From my limited knowledge of how this all works, this looks good. I think it'll be a really beneficial change!

scripts/generate_matrix.sh
rnro: Good catch, thanks George

glbrntt (approved):

PR #3014, opened 2024-12-03
Revert `fastRebase` implementation

clintonpi:
Motivation:

`UnsafeRawBufferPointer.init(fastRebase:)` and `UnsafeMutableRawBufferPointer.init(fastRebase:)` were shimmed into NIO in https://github.com/apple/swift-nio/pull/1696. The shim is no longer necessary.

Modifications:

- Revert the use of the `fastRebase` inits to the native `.init(rebasing:)`.

Result:

Use of native APIs instead.


glbrntt (approved):
Thanks @clintonpi

PR #3015, opened 2024-12-03
Add EventLoop.now API for getting the current time

simonjbeaumont:
### Motivation

Code that tries to work with `NIODeadline` can be challenging to test using `EmbeddedEventLoop` and `NIOAsyncTestingEventLoop` because they have their own, fake clock.

These testing event loops do implement `scheduleTask(in:_)` to submit work in the future, relative to the event loop clock, but there is no way to get the event loop's current notion of "now", as a `NIODeadline`, and users sometimes find that `NIODeadline.now`, which returns the time of the real clock, to be surprising.

### Modifications

Add `EventLoop.now`  to get the current time of the event loop clock.

### Result

New APIs to support writing code that's easier to test.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -269,6 +269,9 @@ public protocol EventLoop: EventLoopGroup {
     @preconcurrency
     func submit<T>(_ task: @escaping @Sendable () throws -> T) -> EventLoopFuture<T>
 
+    /// The current time of the event loop.
+    var now: NIODeadline { get }
+
     /// Schedule a `task` that is executed by this `EventLoop` at the given time.
     ///
     /// - Parameters:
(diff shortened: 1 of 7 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -394,6 +397,11 @@ public protocol EventLoop: EventLoopGroup {
     func cancelScheduledCallback(_ scheduledCallback: NIOScheduledCallback)
 }
 
+extension EventLoop {
+    /// Default implemnetation of `now`: Returns `NIODeadline.now()`.
```
glbrntt, line 401:
```suggestion
    /// Default implementation of `now`: Returns `NIODeadline.now()`.
```

glbrntt (approved):
One typo, but looks good otherwise!

PR #3019, opened 2024-12-06
A re-usable static SDK compilation workflow

rnro:
A re-usable static SDK compilation workflow

### Motivation:

We would like to offer the ability to opt-in to CI coverage for compilation against the Linux static SDK.

### Modifications:

Introduce a workflow which will install a Swift snapshot and Linux SDK
and test the repository builds against it.

The script which sets up the VM will install snapshots and SDKs for
either branch or released-version snapshots. They can be specified as:
- a specific version string
- the 'latest' version, the script attempts to find the most recent
  version enumerated on swift.org with both SDK and toolchain available
- the most recent snapshot for the specified branch name enumerated on
  swift.org with both SDK and toolchain available

At the moment the script should work on ubuntu and amazonlinux OSes,
this can be expanded in the future.

The current matrix covers testing building on Ubuntu Jammy for the main
branch and latest released Swift version (6.0.3) and enables it on our periodic CI runs.

See https://github.com/apple/swift-nio/actions/runs/13722773053/job/38381907947?pr=3019 for a recent example of this workflow in action.

### Result:

Increased CI coverage


.github/workflows/static_sdk.yml
```diff
@@ -0,0 +1,37 @@
...
+    # Workaround https://github.com/nektos/act/issues/1875
+    uses: apple/swift-nio/.github/workflows/swift_test_matrix.yml@static_sdk_workflow  # TODO: change to @main
+    with:
+      name: "Static SDK"
+      matrix_string: >-
+        {
+          "swift":[
+            {
+              "name":"latest-release Jammy",
+              "swift_version":"6.0.2",
```
FranzBusch, line 18: Is this outdated as of today since 6.0.3 got released?

0xTim:
Is there any reason not to add this to the Swiftlang workflow repo straight away?

FranzBusch:
> Is there any reason not to add this to the Swiftlang workflow repo straight away?

We are not using the swiftlang repo for our matrix workflows right now. We plan on opening a PR to propose merging our matrix workflows into swiftlang but for now we are keeping them separate to iterate faster to a state where we feel good in upstreaming them.

.github/workflows/static_sdk.yml
rnro: It was informational only and will just get out of date so I changed it to `latest-release`

FranzBusch (approved):

PR #3009, opened 2024-12-01
Use the new Android overlay in the tests and update some Bionic declarations

finagolfin:
### Motivation:

Get this repo building again for Android with NDK 27

### Modifications:

- Update some networking declarations for newly added nullability annotations
- Import the new Android overlay instead in some tests
- Add two force-unwraps on all platforms, that are needed for Android

### Result:

This repo and its tests build for Android again

I've been [using these patches on my Android CI](https://github.com/finagolfin/swift-android-sdk/blob/main/swift-nio-ndk27.patch) and natively on Android for a couple months now. I didn't bother keeping this patch building for Android with Swift 5 anymore, as my Android CI no longer tests Swift 5.

I built this pull and ran the tests on linux x86_64 to make sure there was no regression.

PR #3025, opened 2024-12-12
Remove now unneeded fts_open bitcast for Android

finagolfin:
I added this changed function call for Android in #2660 earlier this year, because of [an incorrect Bionic function signature as explained recently](https://github.com/apple/swift-nio/pull/3009#discussion_r1864807270), but now that it has been worked around through a new API note as mentioned there, this change is no longer needed.

I simply did not notice this earlier because it still compiles and merely gives a warning, which removing this call now silences.


Lukasa (approved):

finagolfin:
Few CI failures are unrelated, as this only affects Android.

PR #3032, opened 2024-12-17
Wait for `closeFuture` instead of close promise in `NIOAsyncChannel`'s `executeThenClose`

gjcairo:
After the changes introduced in https://github.com/apple/swift-nio-http2/pull/487, we need to make a small change in the implementation of `NIOAsyncChannel` to wait on the `closeFuture` instead of on `close`'s promise in the `executeThenClose` implementation.

### Motivation:

`executeThenClose` shouldn't fail from errors arising from closing the channel - at this point, the user of the channel cannot really do anything, and since the channel has been closed, we should not fail since resources have been cleaned up anyways.

### Modifications:

This PR changes the implementation of `NIOAsyncChannel` to wait on the `closeFuture` instead of on `close`'s promise in the `executeThenClose` implementation.
It also updates the docs for `closeFuture` to better explain when it will be succeeded and why it won't ever be failed.

### Result:

`executeThenClose` won't throw errors upon closing.


Sources/NIOCore/Channel.swift
```diff
@@ -107,6 +107,11 @@ public protocol Channel: AnyObject, ChannelOutboundInvoker, _NIOPreconcurrencySe
     var allocator: ByteBufferAllocator { get }
 
     /// The `closeFuture` will fire when the `Channel` has been closed.
+    ///
+    /// - Important: This future will never be failed, as it signals when the channel has been closed, and this action cannot fail.
```
glbrntt, line 111: It can absolutely fail if a `Channel` doesn't implement this as expected, which is quite likely given that this is the first time we're documenting it 😄

I think we should say this it _should_ never be failed.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -301,15 +301,13 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
-                return result
-            }
-            throw error
-        }
+        self._outbound.finish()
+        // We ignore errors from close, since all we care about is that the channel has been closed
+        // at this point.
+        self.channel.close(promise: nil)
+        // `closeFuture` is never failed, so we can ignore the error
+        try? await self.channel.closeFuture.get()
```
glbrntt, line 309: Not sure whether we decided to ignore `closeFuture` errors entirely or not – it seems completely reasonable to ignore them though. 

Suspect @FranzBusch has opinions.

glbrntt (approved):
LGTM!

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
gjcairo: This future should never be failed; if we're considering a failing `closeFuture` the result of a bad implementation of a `Channel` but not something that should happen in any other scenario, it seemed reasonable to me to ignore all errors.

FranzBusch: The original goals were:
1. Make sure we only excited from `executeThenClose` when the channel is actually closed
2. Surface any errors if closing fails

Now 1. is still being upheld here since we wait for the `closeFuture`. It also turns out that closing can't fail and a call to `close()` must always result in the closeFuture completing. Strictly speaking succeeding actually.

Now we have two options here. We can just assume that the "strict" part is true so there will never be any errors. However, there might be some code out there that actually fails the close future. What if we catch the error and add an assert that triggers if this close future actually fails to help us spot incorrect implementations?

gjcairo: Fair enough, adding an assertion sounds useful.

FranzBusch (approved):

PR #3034, opened 2024-12-18
Add Sendability annotations on SocketOptionProvider

Lukasa:
### Motivation:

SocketOptionProvider is used to apply a number of socket options. As part of using it in NIOPosix, it became apparent that several necessary sendability annotations are missing. Let's add those.

### Modifications:

- Add Sendable constraints with @preconcurrency to SocketOptionProvider.
- Update the BaseSocketChannel conformance to match

### Result:

Tighter Sendable constraints on SocketOptionProvider.


glbrntt (approved):

Lukasa:
Merging over API breakage failure, which is wrong.

PR #3039, opened 2024-12-30
Add stale action to mark issues and PRs as stale and close them

gjcairo:
### Motivation:

To keep good repo hygiene, it would be good to have old issues/PRs automatically commented on after a period of inactivity, and if still inactive after this, subsequently closed.
For this we can use the `stale` GH action. We can try out a first integration here before potentially moving it to other repos too.

### Modifications:

- Issues will be marked as stale after 30 days of inactivity, and subsequently closed after 7 more days of inactivity.
- PRs will be marked as stale after 45 days of inactivity, and subsequently closed after 10 more days of inactivity.
- A label `stale` will be added to issues and PRs marked as stale, and a comment will be left.

All of these things can be changed if we think the durations, the message, or the label name (if any) should be different.

### Result:

Hopefully better responsiveness in issues/PRs and cleaner repos.

### DIFF
--- a/.github/workflows/stale_issues_prs.yml
+++ b/.github/workflows/stale_issues_prs.yml
@@ -1,0 +1,27 @@
+name: 'Close stale issues and PRs'
+
+on:
+  schedule:
+    - cron: '0 7 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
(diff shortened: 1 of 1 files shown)


Lukasa:
Just as an FYI before I'm back from break, but I'm pretty strongly opposed to doing this. We can discuss later.

PR #3040, opened 2025-01-05
Fix WebSocketProtocolErrorHandler sending the close frame with appropriate masking key

johnnzhou:
Enhance `WebSocketProtocolErrorHandler` to correctly add masking key for client/server.

### Motivation:
In `NIOWebSocket`, the automatic error handling provided by `WebSocketProtocolErrorHandler` offers a convenient way for both clients and servers to handle protocol-level errors. However, the `WebSocketFrame` used in `WebSocketProtocolErrorHandler` does not fully adhere to the RFC 6455 standard. Specifically, as per [RFC 6455 Section 5.1](https://datatracker.ietf.org/doc/html/rfc6455#section-5.1), a client *must* mask all frames it sends to the server, while a server *must not* mask any frames it sends to the client. This PR addressed this discrepancy to ensure compliance with the WebSocket protocol.

### Modifications:
In the `WebSocketProtocolErrorHandler` initializer, the user can specify whether the handler is used for a WebSocket client or server. Within the `errorCaught` method, the `WebSocketFrame` will include a maskingKey if the handler is used by a client, or nil if it is used by a server.

Additionally, the `@unchecked` annotation is removed from `NIOWebSocketServerUpgrader`  since the project is now using swift 5.9 toolchain.

### Result:
The `WebSocketProtocolErrorHandler` is more robust and can correctly close the connection is error occurs.


adam-fowler:
Given previously a masking key wasn't added shouldn't the default continue to be that. Also adding a parameter to a function is considered a breaking change, even if it has a default value. You need to add a new function which sets the `isServer` parameter.

johnnzhou:
Hi @adam-fowler, thanks for the review. I addressed your comments in the most recent version. Please let me know if everything looks good now or if there’s anything else you’d like me to adjust.

Thanks again!

adam-fowler:
Instead of having a mutable isServer member variable. You could have two initializers. The original and a new one which takes a isServer parameter. 

NIO folks will have last say on this though.

Lukasa:
I like Adam's suggestion of using a `let` instead of a `var`, and setting it in the initializer. The value should never change at runtime anyway.

Sources/NIOWebSocket/WebSocketProtocolErrorHandler.swift
```diff
@@ -22,7 +22,12 @@ public final class WebSocketProtocolErrorHandler: ChannelInboundHandler {
     public typealias InboundIn = Never
     public typealias OutboundOut = WebSocketFrame
 
-    public init() {}
+    /// Indicate that this `ChannelHandeler` is used by a WebSocket server or client. Default is true.
+    public private(set) var isServer: Bool
```
Lukasa, line 26: Let's avoid making anything `public` we don't have to. I can't think of many uses for having this be public so we should probably avoid it.

Lukasa (commented):
Generally looks good, just a quick note!

johnnzhou:
Thanks @Lukasa for the review! I updated the PR to address your comments. Could you please review whenever you are available? Appreciate it!

Lukasa (approved):
Looks great, thanks @johnnzhou!

PR #3042, opened 2025-01-08
GetaddrinfoResolver succeeds futures on eventLoop

rnro:
### Motivation:

`testClientBindWorksOnSocketsBoundToEitherIPv4OrIPv6Only` would fail sometimes leaking the IPv4 promise in `GetaddrinfoResolver`

`HappyEyeballsConnector` returns the connection when it resolves either IPv4 of IPv6. It uses the `GetaddrinfoResolver` which holds a promise for each of the IPv4 and IPv6 resolution; when one is completed the connection will be returned and it is possible to start tearing down the test and shutting down the event loop before the other is completed and we leak the promise.

### Modifications:

Complete both futures on the event loop rather than the dispatch queue.

### Result:

The futures are completed in the same event loop tick meaning that we cannot continue execution and leak one.


glbrntt (approved):
Nice one!

Lukasa (approved):
Nice one.

glbrntt (approved):

PR #3044, opened 2025-01-09
NIOAsyncWriter: Fix suspending yield when writer is finished

orobio:
This fixes hitting the following preconditionFailure in NIOAsyncWriter: "This should have already been handled by `yield()`".

It doesn't expect a yield to be suspended when the state is `.writerFinished`, but this can definitely happen. This seemed to be the correct solution to me, but please check it carefully, because I'm unfamiliar with this code.

I'm not happy about the test for this. It's blocking the `didYield` call for a while, to make sure the correct state is reached. See also the 'FIXME' line. What would be a better way to do this?


Lukasa (commented):
This is a great catch, thanks so much!

Regarding your test, it might be useful for the test to use a new delegate that makes it easier to arrange the ordering to work out. NIO's ConditionLock can be used to set up a nice ordering chain, so that instead of using the usleeps we can just block on that lock being unlocked with the right value.

orobio:
@Lukasa : I updated it with ConditionLock. Is this what you had in mind?

This is still blocking a thread from the concurrency pool and then depending on another thread to unblock it. Will that not be a problem? The only way I see around that is to add an async hook to NIOAsyncWriter.

Lukasa:
For testing purposes, that blocking will be safe. We will eventually resolve it, and we don't rely on Dispatch to make that happen (it requires the NIO thread to do it), so it'll be ok.

Lukasa (approved):
This is a really lovely patch, thanks so much @orobio!

orobio:
> For testing purposes, that blocking will be safe. We will eventually resolve it, and we don't rely on Dispatch to make that happen (it requires the NIO thread to do it), so it'll be ok.

I'm not sure I fully understand what you're saying here, but I assume it's fine then.

I do still have that 'FIXME' line in there, which would probably be good to remove. Well..., this got merged while I was typing, so I'll open a new pull request for that.

finagolfin:
It appears this pull gets [the NIO tests to hang](https://github.com/finagolfin/swift-android-sdk/actions/runs/12784959691/job/35638994901) in [the Android x86_64 emulator](https://github.com/finagolfin/swift-android-sdk/actions/runs/12784959691/job/35638996500), but weirdly only when compiled and run on macOS, ie the linux CI runs shown there work fine. I don't know if that's a macOS cross-compilation issue or some incompatibility with the macOS emulator triggered by this pull, possibly both. Reverting this pull gets [my Android CI runs on macOS to work again](https://github.com/finagolfin/swift-android-sdk/actions/runs/12794532278/job/35669869104).

Let me know what you guys think.

Lukasa:
That’s very interesting. Do you have a smaller repro? The OS-specificity makes me think this is probably a Swift bug, but if it’s easy enough to reproduce we may be able to figure out if the test that hangs is just flaky.

finagolfin:
A bit tough for me to reproduce since I don't use macOS. @marcprux, any interest in looking into this on your Mac?

marcprux:
When I test against a local Android emulator (with `skip android test`) on my ARM macOS 15 machine, the `NIOAsyncWriterTests `tests (which appears to be where your CI is hanging) do all pass:

```
Test Suite 'NIOAsyncWriterTests' passed at 2025-01-16 09:23:46.416
	 Executed 33 tests, with 0 failures (0 unexpected) in 0.022 (0.022) seconds
```

However, the test run does ultimately fail due to a leaked promise.

I also set up an Android CI in a fork of swift-nio and both macos-13 and ubuntu-24.04. In the [run](https://github.com/marcprux/swift-nio/actions/runs/12811481892), macOS hangs like you are seeing and Ubuntu fails with the same leaked promise fatalError. I tried both with and without applying your patches in the [workflow](https://github.com/marcprux/swift-nio/blob/main/.github/workflows/android.yml).

I'm not sure why the CI's macOS testing hangs but my local tests complete, but one difference is that the CI's Android emulator is API 29 and mine is API 34. I'll try with some different Android API levels and see if that changes anything.

Another possibility is architecture: my local emulator is running on aarch64, but the CI runs x86_64 (which is why we are running on macos-13: it is the last public macOS image to support x86). I won't have a macOS x86 machine handy, but I can get access to one later on, so I can try that out if you suspect that it could be architecture related.


Here's the leaked promise, although it is likely a red herring:

```
Test Case 'ChannelTests.testAskForLocalAndRemoteAddressesAfterChannelIsClosed' started at 2025-01-16 09:23:53.419
NIOPosixTests/TestUtils.swift:697: Fatal error: leaking promise created at (file: "NIOPosixTests/TestUtils.swift", line: 697)
Current stack trace:
0    libswiftCore.so                    0x00000075619ad878 _swift_stdlib_reportFatalErrorInFile + 128
1    libswiftCore.so                    0x00000075617cfb64 _assertionFailure(_:_:file:line:flags:) + 244
2    swift-nioPackageTests.xctest       0x00000061dced52f0 <unavailable> + 19673840
3    swift-nioPackageTests.xctest       0x00000061dcf77a40 <unavailable> + 20339264
4    swift-nioPackageTests.xctest       0x00000061dcf77e34 <unavailable> + 20340276
5    swift-nioPackageTests.xctest       0x00000061dcee49ec <unavailable> + 19737068
6    swift-nioPackageTests.xctest       0x00000061dcee4aac <unavailable> + 19737260
7    libswiftCore.so                    0x000000756191a79c <unavailable> + 4884380
8    libswiftCore.so                    0x000000756191b274 <unavailable> + 4887156
9    swift-nioPackageTests.xctest       0x00000061dda7ca48 <unavailable> + 31894088
10   swift-nioPackageTests.xctest       0x00000061dda705b0 <unavailable> + 31843760
11   swift-nioPackageTests.xctest       0x00000061dda7334c <unavailable> + 31855436
12   swift-nioPackageTests.xctest       0x00000061dda754fc <unavailable> + 31864060
13   swift-nioPackageTests.xctest       0x00000061dda7cf88 <unavailable> + 31895432
14   swift-nioPackageTests.xctest       0x00000061dda7bab4 <unavailable> + 31890100
15   swift-nioPackageTests.xctest       0x00000061dda7b4c8 <unavailable> + 31888584
16   swift-nioPackageTests.xctest       0x00000061dd7611d4 <unavailable> + 28635604
17   swift-nioPackageTests.xctest       0x00000061dd76e1b8 <unavailable> + 28688824
18   libXCTest.so                       0x000000755b6ccbcc <unavailable> + 236492
19   libXCTest.so                       0x000000755b6ce564 XCTAssertNoThrow<A>(_:_:file:line:) + 68
20   swift-nioPackageTests.xctest       0x00000061dd7907cc <unavailable> + 28829644
21   swift-nioPackageTests.xctest       0x00000061ddd54300 <unavailable> + 34874112
22   swift-nioPackageTests.xctest       0x00000061dce81c6c <unavailable> + 19332204
23   swift-nioPackageTests.xctest       0x00000061ddd829ac <unavailable> + 35064236
24   libXCTest.so                       0x000000755b6c5c44 <unavailable> + 207940
25   libXCTest.so                       0x000000755b6c50bc <unavailable> + 204988
26   libXCTest.so                       0x000000755b6c5bac <unavailable> + 207788
27   libXCTest.so                       0x000000755b6c5b8c <unavailable> + 207756
28   libXCTest.so                       0x000000755b6c5f00 <unavailable> + 208640
29   libXCTest.so                       0x000000755b6c405c <unavailable> + 200796
30   libXCTest.so                       0x000000755b6c6164 XCTestCase.invokeTest() + 356
31   libXCTest.so                       0x000000755b6c5f48 XCTestCase.perform(_:) + 128
32   libXCTest.so                       0x000000755b6c8a54 XCTest.run() + 164
33   libXCTest.so                       0x000000755b6c706c XCTestSuite.perform(_:) + 164
34   libXCTest.so                       0x000000755b6c8a54 XCTest.run() + 164
35   libXCTest.so                       0x000000755b6c706c XCTestSuite.perform(_:) + 164
36   libXCTest.so                       0x000000755b6c8a54 XCTest.run() + 164
37   libXCTest.so                       0x000000755b6c706c XCTestSuite.perform(_:) + 164
38   libXCTest.so                       0x000000755b6c8a54 XCTest.run() + 164
39   libXCTest.so                       0x000000755b6c40ec XCTMain(_:arguments:observers:) + 312
40   libXCTest.so                       0x000000755b6c42c4 XCTMain(_:) + 40
41   swift-nioPackageTests.xctest       0x00000061ddddf170 <unavailable> + 35443056
42   swift-nioPackageTests.xctest       0x00000061ddddf190 <unavailable> + 35443088
43   swift-nioPackageTests.xctest       0x00000061ddddf1a8 <unavailable> + 35443112
44   libc.so                            0x000000755d8f25a8 __libc_init + 108
Trap 
```

marcprux:
Actually, scratch all that. It seems to be hanging randomly, on both macOS and Ubuntu runners, irrespective of API level. See the runs at https://github.com/marcprux/swift-nio/actions/runs/12812276673

I'm still not able to get it to hang on my local machine, but I'll keep trying.

Lukasa:
Does it always hang on the same test?

marcprux:
No, it seems to vary. Sometimes in `NIOAsyncSequenceProducerTests.testYield_whenStreaming_andNotSuspended_andDemandMore `, sometimes in `NIOAsyncWriterTests.testSetWritability_whenInitial`. It's hard to tell exactly because I think there's some output buffering going on (because sometimes the log will just stop at an incomplete line like:

```
Test Case 'NIOAsyncSequenceProducerTests.testYield_whenS
Error: The operation was canceled.
```

Lukasa:
Yeah, that seems fairly likely. Can you run the underlying test binary directly? Potentially running it under `stdbuf -o0` to try to disable output buffering entirely.

finagolfin:
On my Android CI, it is entirely consistent: linux CI runs pass every time while macOS runs always hang. See [the run yesterday: across two attempts with this pull, I had four macOS hangs and six linux runs pass](https://github.com/finagolfin/swift-android-sdk/actions/runs/12805530874/attempts/1) (ignore the devel macOS failures, that's a separate issue). So much so that today I only reverted this pull for macOS, finagolfin/swift-android-sdk@7476f2be5, and [all the runs passed](https://github.com/finagolfin/swift-android-sdk/actions/runs/12825967607).

Lukasa:
@finagolfin What happens if you only revert the test?

finagolfin:
I don't think this test is the issue, as it appears to be failing elsewhere in the log, but I will check that and let you know.

finagolfin:
Nope, you were right to suspect the test, as simply [reverting this test on macOS got the NIO tests to pass again](https://github.com/finagolfin/swift-android-sdk/actions/runs/12845354927/job/35819332059). 😄

Marc, can you confirm?

marcprux:
> Marc, can you confirm?

Confirmed: if I remove the test, then the hangs never occurs in the emulator (running on macOS-13 or Ubuntu).

finagolfin:
@marcprux, are you disabling all NIOAsyncWriterTests though, as shown in the linked commit above, marcprux/swift-nio@87d02310d? If so, it would be better to just disable this new test, as I did above, to see if this code change affects the older tests too, which you are no longer running.

I'm going to try running the linux-compiled tests on the macOS emulator and vice versa next, to see if that helps narrow down where this strange hang is coming from, as it could be that we're hitting a bug in the Android emulator for macOS.

Lukasa:
This could be an emulator bug, but it could also be that we're seeing an ordering issue flushed out by the performance overhead of the emulator. It'd be useful to try to get a "sequence of events" log from your hanging test as well, as concurrent tests are horrible to debug.

finagolfin:
Alright, I ran the NIO tests alone in the Android emulator, once [swapping the mac-compiled tests into the linux-hosted emulator and vice versa](https://github.com/finagolfin/swift-android-sdk/actions/runs/12935522527/attempts/1), once [keeping the mac-built tests in the mac-hosted emulator and so on, where only the mac version hung](https://github.com/finagolfin/swift-android-sdk/actions/runs/12935540737).

Looking at the results from the swapped run, the macos-hosted emulator again hung each time, with the linux-compiled tests not finishing for 20 minutes before I killed it (I have a 5 macOS job limit on the OSS free tier of github Actions, so I killed it once it seemed to hang, so other jobs could proceed). I just kicked off [a swapped run again that I will let run for longer](https://github.com/finagolfin/swift-android-sdk/actions/runs/12935522527) to make sure it does hang after longer.

This means the problem is in the interaction between this test and the macOS-hosted Android emulator, since the linux-compiled test runs fine in the linux-hosted emulator but hangs in the mac-hosted emulator. The mac cross-compile and mac-hosted Android emulator are consistently slower on github Actions, because of greater contention for fewer macOS CI machines or whatever, so either that lack of speed screws up this test or there is a bug in the macOS-hosted Android emulator.

Regarding "a 'sequence of events' log," you want me to spray this test with print statements to figure out what's running when?

Lukasa:
Yeah, that was what I was asking for. Presumably this is the result of a specific ordering sequence caused by the slowdown in the emulator.

orobio:
Sorry, I was away for a while and completely missed this. I'd like to mention that I'm still a bit uncertain about one Task being blocked, waiting for another Task. Could that have anything to do with this? I can imagine that could cause a problem when other tests are doing a similar thing and we're running out of Swift concurrency threads that are making forward progress.

finagolfin:
No idea, I will stick a bunch of `print()`s in later this week and try to get you a log. You may be able to reproduce on linux yourself with slow or constrained hardware.

orobio:
It'd be great to have a more detailed log, thanks! Perhaps next week or so I could have a look at trying to reproduce this on constrained hardware.

finagolfin:
Alright, finally got around to this, but not sure how useful it is considering the test output may not always flush.

I [disabled all other tests in this test file and stuck `print()`s all over this new test](https://github.com/finagolfin/swift-android-sdk/pull/205/files#diff-f0157e7f7394e4377ce7516dcad925242f1faf112a034c9509e0bec81afe5f6b). Most NIO test runs on the Android emulator in macOS [then](https://github.com/finagolfin/swift-android-sdk/actions/runs/13239608343/job/36952500142#step:4:4499) [hang](https://github.com/finagolfin/swift-android-sdk/actions/runs/13239608343/job/36952501117#step:4:1299) the output in [random](https://github.com/finagolfin/swift-android-sdk/actions/runs/13240887013/job/36955811704?pr=205#step:19:1818) [places](https://github.com/finagolfin/swift-android-sdk/actions/runs/13240887013/job/36955812351?pr=205#step:19:5205), but [two of them](https://github.com/finagolfin/swift-android-sdk/actions/runs/13239608343/job/36952500624#step:4:3885) actually [show something inside this test](https://github.com/finagolfin/swift-android-sdk/actions/runs/13240887013/job/36955813038?pr=205#step:19:3885). You can compare those last two failing macOS runs to [a passing run on linux](https://github.com/finagolfin/swift-android-sdk/actions/runs/13240887013/job/36955811380#step:19:3813).

Given that the test output might cut off anywhere before the actual hang, not sure how useful that is.

orobio:
Thanks for trying this and the detailed report! Indeed, it's hard to say what's going on without a proper flush. I tried to reproduce this on a VM with only one CPU, but that worked fine. However, I was able to trigger a hanging state by adding a delay in one of the tasks. This causes a timeout when waiting for both yields being suspended, which results in the ConditionLock value never being set to true. The timeout is only 1 second, so maybe that one triggers in the Android emulator, causing the issue?

I created the following patch: [Fix hanging NIOAsyncWriter test](https://github.com/orobio/swift-nio/commit/1358a77f9e5ec0811ed93b6d73abd2d6a384dda3)

It does 2 things:
1. Increase the timeout for waiting for both yields being suspended from 1 to 5 seconds.
2. Add a timeout to waiting for the ConditionLock, so that the test will always complete.

I'm hoping that the former fixes the issue, but if not, the latter should make sure that the test doesn't hang and we can hopefully get more information from the log. Could you give this a test run?

finagolfin:
I just ran it with your test modifications: [doesn't hang anymore but now fails, see full log output on my CI](https://github.com/finagolfin/swift-android-sdk/actions/runs/13358056336/job/37303539115#step:19:2176).

orobio:
It looks like what's happening is what I mentioned earlier. It's waiting on the condition lock and there is no thread available on which the other Task can be scheduled to set the condition. I am not able to reproduce this on my Linux VM with one CPU core. I did some tests and I found that when I keep creating Tasks that block on a ConditionLock, Swift just keeps spawning new threads to schedule the work on, until a maximum of 128 threads is reached. If I understand correctly, the underlying libDispatch is doing this and the Android implementation does not use libDispatch, which explains the difference. I guess the Android implementation really has a fixed-width thread pool.

@finagolfin: After some more digging, I can see that all your configurations are using 2 CPU cores for the Android emulator, but I found the following warning in your Android emulator launch logs for the MacOS runs:
`WARNING      | Running on a system with less than 6 logical cores. Setting number of virtual cores to 1`

So, it looks like it's a combination of having only 1 CPU core and a fixed-width concurrency thread pool that causes the issue.

@Lukasa: A solution could be to add an async test hook into NIOAsyncWriter, so that we can just suspend instead of blocking the thread. For example, something like `_didYieldAsync`, or perhaps `_willPerformAction`. Would this be acceptable? Or do you know of any other way we can solve this without the need for another test hook?

Lukasa:
Having an async method on the writer delegate is a weird design, because it arguably misses the intended use-case, which is to bridge async producers into synchronous consumers. The sink delegate is supposed to be synchronous because it is the consumer.

A better option is to find a way to skip the test in cases where it is likely to run into trouble. If we needed a real sledgehammer we could just skip the entire test on Android, but I wonder if we can construct a better question to ask the runtime.

orobio:
> Having an async method on the writer delegate is a weird design, because it arguably misses the intended use-case, which is to bridge async producers into synchronous consumers.

I understand, but I meant for it to only be used for testing. We need some way to pause execution of the first yield at the right moment, to make sure the state machine is in the correct state for the second yield. With an async hook we can add this pause without blocking a thread from the concurrency pool. The didYield call is actually performed from an async context, so it should be easy to add an async callback for testing next to it.

> Swift Concurrency on Android uses libdispatch, and I haven't seen any big differences in this repo either, so I don't think that's it.

Ah, I didn't know it used libdipatch as well. Could it be that libdispatch behaves differently on Android for some reason? On my Linux machine it starts a maximum of 128 threads. When I start 127 Tasks at the start of the test and block them all, leaving exactly 1 concurrency thread free, I can see that the test fails in the same way as on the MacOS Android emulator. When I start 126 blocked Tasks, leaving 2 concurrency threads free, the test is successful. That, combined with the fact that the MacOS Android emulator runs on 1 core and the Linux one on 2 cores makes me think that it's very likely that's the issue. If you want, I can create something with which we can test the behavior of the concurrency thread pool in the Android emulator.

Lukasa:
> I understand, but I meant for it to only be used for testing, just like the already existing didSuspend callback, which is directly set on the storage with [_setDidSuspend(...)](https://github.com/orobio/swift-nio/blob/a5fbe75546dfeee247d4e303907adf3dec289c76/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift#L474-L480).

My nervousness there though is that `didSuspend` is still synchronous. While the hook may only be _intended_ for use in testing, it's actually available for use everywhere, and any future reading of the code has to reason about it in the face of potential concurrent calls.

With that said, I'm not 100% definitely opposed. Just noting that it conceptually complicates matters somewhat.

orobio:
There is another bug in the implementation that was revealed by the Android run. In the log it can be seen that an error was thrown from one of the tasks. This happens due to that the continuation of the second task (which should suspend again) will not be resumed until the yield of the first task is fully completed. The bug is that when the first yield finishes up, the state machine determines that there are no suspended yields and it moves the state to .finished, resulting in an alreadyFinished error from the second task.

I can have a look at fixing that issue later, but if we're going with the async test hook approach, I anticipate that we probably need another hook to be able to test that issue, which means that we're already looking at adding 2 more hooks. After thinking about it some more, I'm now leaning towards using a custom executor and setting an executor preference on the tasks. That way we can guarantee that there are at least 2 threads available. Later, we can use the same mechanism to make sure there is only 1 thread available for both tasks, which will reproduce the issue described above.

I'm not really familiar with custom executors, but I think this can work and will investigate it more.

finagolfin:
I think you've managed to track this down, the number of cores appears to be the cause. First, I applied your patch with the new test and ran it on the linux and mac Android emulators as usual. [The linux-hosted emulator passed as always](https://github.com/finagolfin/swift-android-sdk/actions/runs/13629559329/job/38094141353#step:19:1382):
```
Test Case 'NIOAsyncWriterTests.testConcurrencyThreadPoolWidth' started at 2025-03-03 11:21:29.021
2025-03-03 11:21:29 +0000: Starting Task 0
2025-03-03 11:21:29 +0000: Task 0 started
2025-03-03 11:21:29 +0000: Starting Task 1
2025-03-03 11:21:29 +0000: Task 1 started
2025-03-03 11:21:29 +0000: Starting Task 2
Detected initial number of threads: 2
Detected maximum number of threads: 2
2025-03-03 11:21:39 +0000: Task 2 started
Test Case 'NIOAsyncWriterTests.testConcurrencyThreadPoolWidth' passed (10.01 seconds)
```
while [the mac-hosted emulator hung](https://github.com/finagolfin/swift-android-sdk/actions/runs/13629559329/job/38094141732#step:19:5324):
```
Test Case 'NIOAsyncWriterTests.testConcurrencyThreadPoolWidth' started at 2025-03-03 11:33:12.245
2025-03-03 11:33:12 +0000: Starting Task 0
2025-03-03 11:33:12 +0000: Task 0 started
2025-03-03 11:33:12 +0000: Starting Task 1
Detected initial number of threads: 1
Detected maximum number of threads: 1
2025-03-03 11:33:22 +0000: Task 1 started
```
I then [limited the linux-hosted emulator to a single core just like the mac and they all hung](https://github.com/finagolfin/swift-android-sdk/actions/runs/13630956604/job/38098433363#step:19:1253):
```
Test Case 'NIOAsyncWriterTests.testConcurrencyThreadPoolWidth' started at 2025-03-03 12:38:48.854
2025-03-03 12:38:48 +0000: Starting Task 0
2025-03-03 12:38:48 +0000: Task 0 started
2025-03-03 12:38:48 +0000: Starting Task 1
Detected initial number of threads: 1
Detected maximum number of threads: 1
2025-03-03 12:38:58 +0000: Task 1 started
```

orobio:
Great, that definitely shows the problem. I will investigate the use of a custom task executor further and can hopefully provide a fix based on that.

This indeed shows different behavior between (GNU) Linux and Android. Would it make sense to align them? I've always understood the Swift global concurrency thread pool to be a fixed width pool, so from that perspective the behavior on Android is correct. However, I can imagine that changing the behavior on Linux would have some undesired impact on existing projects.

One thing that bothers me is that the test hangs when a single core is used. I've been staring at the code for a while, but I don't see why that would happen. If you have any ideas, I'd be happy to hear them!

PR #3046, opened 2025-01-12
feat: NIO.TimeAmount(string:) and TimeAmount.description

### LINKED ISSUE #2504: unit parsing/pretty printing support for `TimeAmount`
I frequently need to parse `500 ms` into `TimeAmount.milliseconds(500)` and also render `.nanoseconds(17)` as `17 ns`. So I usually have to write code akin to

```swift
extension TimeAmount {
     init(_ userProvidedString: String, defaultUnit: String) throws {
         let string = String(userProvidedString.filter { !$0.isWhitespace }).lowercased()
         let parsedNumbers = string.prefix(while: { $0.isWholeNumber || $0.isPunctuation })
         let parsedUnit = string.dropFirst(parsedNumbers.count)

         guard let numbers = Int64(parsedNumbers) else {
             throw ValidationError("'\(userProvidedString)' cannot be parsed as number and unit")
         }
         let unit = parsedUnit.isEmpty ? defaultUnit : String(parsedUnit)

         switch unit {
         case "h", "hr":
             self = .hours(numbers)
         case "min":
             self = .minutes(numbers)
         case "s":
             self = .seconds(numbers)
         case "ms":
             self = .milliseconds(numbers)
         case "us":
             self = .microseconds(numbers)
         case "ns":
             self = .nanoseconds(numbers)
         default:
             throw ValidationError("Unknown unit '\(unit)' in '\(userProvidedString)")
         }
     }

     var prettyPrint: String {
         let fullNS = self.nanoseconds
         let (fullUS, remUS) = fullNS.quotientAndRemainder(dividingBy: 1_000)
         let (fullMS, remMS) = fullNS.quotientAndRemainder(dividingBy: 1_000_000)
         let (fullS, remS) = fullNS.quotientAndRemainder(dividingBy: 1_000_000_000)

         if remS == 0 {
             return "\(fullS) s"
         } else if remMS == 0 {
             return "\(fullMS) ms"
         } else if remUS == 0 {
             return "\(fullUS) us"
         } else {
             return "\(fullNS) ns"
         }
     }
 }
```

and I don't think I should have to write this. NIO should have it :).

natikgadzhi:
Adds `TimeAmount.init(string:)` and `TimeAmount.description` for parsing time amounts from strings and pretty printing them. 

Closes #2504.

### Motivation:

Had a minute, wanted to work on Swift-NIO a bit more, and saw @weissi still wanted this.

### Modifications:

It's largely based on the snippet @weissi made in #2504 with a few changes:
- Added unit tests.
- Added support for more unit aliases for convenience based on @glbrntt's suggestion.
- Changed `.gitignore` to ignore `.build` everywhere, I've got a few of them when testing locally.

### Open questions:

- I originally thought perhaps I should add support for multiple number and unit pairs, i.e. `1h 31m`, but decided against it. Feels like an edge case. Happy to add this if you think it's needed.


Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
     }
 }
 
+/// Contains the logic for parsing time amounts from strings,
+/// and printing pretty strings to represent time amounts.
+extension TimeAmount: CustomStringConvertible {
+
+    /// Errors thrown when parsint a TimeAmount from a string
+    public enum ValidationError: Error, Equatable {
```
natikgadzhi, line 559: I'm not sure it's a great idea to introduce a new public error type. Can I get away with re-using something?

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+
+    /// Creates a TimeAmount from a string representation with an optional default unit.
+    ///
+    /// Supports formats like:
+    /// - "5s" (5 seconds)
+    /// - "100ms" (100 milliseconds)
+    /// - "42" (42 of default unit)
+    /// - "1 hr" (1 hour)
+    ///
+    /// This function only supports one pair of the number and units, i.e. "5s" or "100ms" but not "5s 100ms".
```
natikgadzhi, line 575: Doc string to be explicit on not supporting complex strings.

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+    /// - ns, nanos (nanoseconds)
+    ///
+    /// - Parameters:
+    ///   - userProvidedString: The string to parse
+    ///   - defaultUnit: Unit to use if no unit is specified in the string
+    ///
+    /// - Throws: ValidationError if the string cannot be parsed
+    public init(_ userProvidedString: String, defaultUnit: String = "s") throws {
+        // First parse the string into a number and supported units.
+        // Clean out white space from the string.
```
natikgadzhi, line 592: These are for my sanity when I was originally reading @weissi snippet. Happy to remove the comments.

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+        case "ns", "nanos":
+            self = .nanoseconds(numbers)
+        default:
+            throw ValidationError.unsupportedUnit("Unknown unit '\(unit)' in '\(userProvidedString)'")
+        }
+    }
+
+    /// Returns a human-readable string representation of the time amount
+    /// using the most appropriate unit
+    public var description: String {
```
natikgadzhi, line 624: Should this be `@inlinable`?

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+    /// - ms, millis (milliseconds)
+    /// - us, µs, micros (microseconds)
+    /// - ns, nanos (nanoseconds)
+    ///
+    /// - Parameters:
+    ///   - userProvidedString: The string to parse
+    ///   - defaultUnit: Unit to use if no unit is specified in the string
+    ///
+    /// - Throws: ValidationError if the string cannot be parsed
+    public init(_ userProvidedString: String, defaultUnit: String = "s") throws {
```
natikgadzhi, line 590: Should this be `@inlinable` or at least `@usableFromInline`?

natikgadzhi (commented):
Code walkthrough comments to kick things off.

Sources/NIOCore/EventLoop.swift
weissi: I doubt that anybody would really _catch_ that error, so we could also just make this an `internal` error type. That way you can still throw & print it but NIO wouldn't be forced into the API.

Sources/NIOCore/EventLoop.swift
Lukasa: `@usableFromInline` is not usable on `public` methods. I have no objection to making it `@inlinable`, but I don't think it's necessary either.

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+        // First parse the string into a number and supported units.
+        // Clean out white space from the string.
+        let string = String(userProvidedString.filter { !$0.isWhitespace }).lowercased()
```
Lukasa, line 593: Up to us whether this matters, but doing this computation on the `String` type pays a lot of performance costs for unicode-correctness. We can always merge it this way for now, of course.

natikgadzhi: Agreed — I don't think this method is will be _as high usage_ so that it unicode correctness penalty would matter?

natikgadzhi:
Alright, cleaned this up. Thank you for taking a look at this quickly, and agreeing to have this as part of NIO, @Lukasa!

Sources/NIOCore/EventLoop.swift
Lukasa: @weissi Can you weigh in here? You are the one who originally requested the feature, and most likely to be grumpy if we pay a cost we didn't need to.

weissi: I typically use this for two purposes:
- command line parsing (e.g. `--timeout 10s`)
- pretty printing

I'm completely fine with Unicode but I don't need it. The implementation that I keep copy & pasting around just supports a basic ASCII form, that was always fine for me. Performance of this hasn't been a concern for me.

Lukasa: Awesome, then let's accept this as-is.

Lukasa (approved):
Nice one, thanks!

Sources/NIOCore/EventLoop.swift
```diff
@@ -551,6 +551,94 @@ public struct TimeAmount: Hashable, Sendable {
...
+    /// - ms, millis (milliseconds)
+    /// - us, µs, micros (microseconds)
+    /// - ns, nanos (nanoseconds)
+    ///
+    /// - Parameters:
+    ///   - userProvidedString: The string to parse
+    ///   - defaultUnit: Unit to use if no unit is specified in the string
+    ///
+    /// - Throws: ValidationError if the string cannot be parsed
+    public init(_ userProvidedString: String, defaultUnit: String = "s") throws {
```
weissi, line 590: FWIW, I wouldn't default `defaultUnit` to anything. Why is it seconds?

natikgadzhi:
@Lukasa @weissi I cleaned this up, looking into what I'm doing wrong with formatting, and picking up how to run soundness checks locally so I don't waste CI time more.

natikgadzhi:
Thank you for helping me though this folks 🖤

PR #3047, opened 2025-01-12
Adds `Int(buffer:)` initializer to `FixedWidthInteger` types

natikgadzhi:
### Motivation:

Makes it possible to make integers out of ByteBuffers directly with an initializer. I.e. `UInt32(buffer: ByteBuffer(bytes: [0, 1, 2, 3]))`.

### Modifications:

- Adds the `Int(buffer:)` initializer in `ByteBuffer-int.swift`
- Adds tests in `ByteBufferTest.swift`. _Holy hell this file is huge._
- Note that `Int(buffer:)` will crash if the buffer does not have enough bytes in it to represent the desired integer type.
- It also does _not_ expose endianness and uses `Endianness.host`, to keep the public API clean and simple. It's a convenience initializer, so I thought keeping it minimal is a good idea.

### Result:

Nicer direct initializer for integers from ByteBuffers.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -134,6 +134,22 @@ extension FixedWidthInteger {
 
         return 1 << ((Self.bitWidth - 1) - self.leadingZeroBitCount)
     }
+
+    /// Initialize an integer from a ByteBuffer. The buffer must contain enough bytes to represent the integer.
+    /// The bytes will be read using the host system's endianness.
+    ///
+    /// - Parameters:
+    ///   - buffer: The ByteBuffer to read from
+    ///
+    /// - Returns: The integer value read from the buffer, or nil if the value could not be read.
+    @inlinable
+    public init?(buffer: ByteBuffer) {
+        var buffer = buffer
+        guard let value = buffer.readInteger(endianness: .host, as: Self.self) else {
+            return nil
+        }
+        self = value
+    }
 }
 
 extension UInt32 {
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -137,18 +137,16 @@ extension FixedWidthInteger {
 
     /// Initialize an integer from a ByteBuffer. The buffer must contain enough bytes to represent the integer.
     /// The bytes will be read using the host system's endianness.
+    /// If the buffer doesn't contain enough bytes to represent the integer, the program will crash.
     ///
     /// - Parameters:
     ///   - buffer: The ByteBuffer to read from
     ///
-    /// - Returns: The integer value read from the buffer, or nil if the value could not be read.
+    /// - Returns: The integer value read from the buffer
     @inlinable
-    public init?(buffer: ByteBuffer) {
+    public init(buffer: ByteBuffer) {
         var buffer = buffer
-        guard let value = buffer.readInteger(endianness: .host, as: Self.self) else {
-            return nil
-        }
-        self = value
+        self = buffer.readInteger(endianness: .host, as: Self.self)!
     }
 }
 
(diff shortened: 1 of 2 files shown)


natikgadzhi:
Hmmmm. Unlike `Array(buffer:)`, this will fail if there's not enough stuff in the Buffer. The other two just read whatever is readable, so they should not fail arbitrarily if the user didn't check the length. 

Perhaps it's worth adding a check and making this throwing, or keeping this optional? Or documenting that the user should ensure the length of the buffer is enough?

natikgadzhi:
I checked the docs for `ByteBuffer`, and they already explain that ByteBuffer can cast to/from "various unsigned int types", so I think that's enough.

Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -134,6 +134,20 @@ extension FixedWidthInteger {
...
+    /// Initialize an integer from a ByteBuffer. The buffer must contain enough bytes to represent the integer.
+    /// The bytes will be read using the host system's endianness.
+    /// If the buffer doesn't contain enough bytes to represent the integer, the program will crash.
+    ///
+    /// - Parameters:
+    ///   - buffer: The ByteBuffer to read from
+    ///
+    /// - Returns: The integer value read from the buffer
+    @inlinable
+    public init(buffer: ByteBuffer) {
```
Lukasa, line 147: This should definitely be failable, so that we can avoid the force-unwrap. We also want to add tests.

There's a separate question about whether this should be _exactly_ the right size.

natikgadzhi: @weissi strong opinions on checking that the buffer size matches the int size exactly?

If we should verify the size match, the most straightforward way would be to check `buffer.readableBytes == MemoryLayout<Self.self>.size`, right?

weissi: I'm definitely in favour of checking that this is exactly the right size. It's weird to left garbage around, if that's expected `readInteger` is the right tool. Endianness should also be configurable (but defaulted to `.host`) Something like

```
var buffer = buffer
guard let value = buffer.readInteger(endianness: endianness, as: Self.self), buffer.readableBytes == 0 else {
    return nil
}
return value
```

natikgadzhi: Done. You did all the work for me. 

I thought making it optional is a better choice than making it throwing — I don't think anyone need to capture the specific error, so no need to make an extra error type.

weissi (approved):
Thank you, I'm happy with this but @Lukasa or someone from the NIO should approve too.

Lukasa (approved):
This is a lovely patch @natikgadzhi, thank you so much! ✨

Lukasa:
Ah dang, one quick note from the docs:

```
error: Return value documented for initializer returning void
   --> ../ByteBuffer-int.swift:146:9-146:117
144 |     ///   - endianness: The endianness to use when reading the integer, defaults to the host system's endianness.
145 |     ///
146 +     /// - Returns: The integer value read from the buffer, or nil if the buffer size did not match the integer type.
    |         ╰─suggestion: Remove return value documentation
147 |     @inlinable
148 |     public init?(buffer: ByteBuffer, endianness: Endianness = .host) {
```

Looks like you should change this to just say "this initializer will fail if the buffer size does not match the size of the integer type".

Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -134,6 +135,23 @@ extension FixedWidthInteger {
...
     }
+
+    /// Initialize an integer from a byte buffer of exactly the right size.
+    /// The bytes will be read using the host system's endianness.
+    ///
+    /// - Parameters:
+    ///   - buffer: The ByteBuffer to read from
+    ///   - endianness: The endianness to use when reading the integer, defaults to the host system's endianness.
+    ///
+    /// - Returns: The integer value read from the buffer, or nil if the buffer size did not match the integer type.
```
natikgadzhi, line 146:
```suggestion
```

natikgadzhi:
Ooops. That should do it.

PR #3048, opened 2025-01-13
feat: Add `String(contentsOf: FilePath ...)`

### LINKED ISSUE #3010: NIOFileSystem: missing String(contentsOf:maximumSizeAllowed:)
### Expected behavior
```
        let string = try await String(
            contentsOf: "/etc/hosts",
            maximumSizeAllowed: .mebibytes(12)
        )
```

should work.

### Actual behavior

API not available

natikgadzhi:
### Motivation:

Makes it possible to read contents of an NIOFileSystem.FilePath into a String directly. Closes #3010.

### Modifications:

- Added a new `String+FileSystem.swift` into `NIOFileSystem` target.
- Added basic tests to the existing `ConvenienceTests.swift`.
- Added a test to cover `Array+FileSystem` initializer as well.

### Result:

- String initializer for `FilePath` is now available
- Array initializer for `FilePath` now has a test

/cc @weissi


Sources/NIOFileSystem/String+FileSystem.swift
```diff
@@ -0,0 +1,56 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
```
weissi, line 5: 2025? Or 2024-2025?

Lukasa (commented):
So I'm 100% fine with this change, but given that we're considering using a type other than `FilePath` in this module, I wonder if we should sit on this a little bit. @glbrntt?

weissi:
> So I'm 100% fine with this change, but given that we're considering using a type other than `FilePath` in this module, I wonder if we should sit on this a little bit. @glbrntt?

@Lukasa This is in the `_NIOFileSystem` module which
- currently uses `FilePath`
- is not API-stable (`_...`)
- will probably get a big change (`s/FilePath/NIOFilePath/`) at some point soon

So I would argue that it's _better_ to merge this as it lessens the likelihood that this gets forgotten. And because it's in the underscored module, there's no risk to paint ourselves into a corner.

Lukasa:
It's the big change that I'm concerned about: I wonder whether it's particularly egregious to add a method whose immediate future is to be deprecated or removed in favour of one with a different type.

glbrntt:
> It's the big change that I'm concerned about: I wonder whether it's particularly egregious to add a method whose immediate future is to be deprecated or removed in favour of one with a different type.

I'd be more concerned if this was isolated to a single API, but given we'll have to deprecate and replace many APIs I don't really see this as an issue.

weissi:
> > It's the big change that I'm concerned about: I wonder whether it's particularly egregious to add a method whose immediate future is to be deprecated or removed in favour of one with a different type.
> 
> I'd be more concerned if this was isolated to a single API, but given we'll have to deprecate and replace many APIs I don't really see this as an issue.

+1, same.

glbrntt (changes requested):
Cool, thanks for opening this! DocC is unhappy about a reference and Johannes' comment re: the copyright header needs addressing but otherwise this looks great.

Sources/NIOFileSystem/String+FileSystem.swift
```diff
@@ -0,0 +1,56 @@
...
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension String {
+    /// Reads the contents of the file at the path into a String.
+    ///
+    /// - Parameters:
+    ///   - path: The path of the file to read.
+    ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``. If the file is larger than this, an error is thrown.
+    ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    ///
+    /// - Throws: If the file is larger than `maximumSizeAllowed`, an ``FileSystemError.resourceExhausted`` error will be thrown.
```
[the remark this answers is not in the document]
natikgadzhi: TIL. Seems to work, pushing up now.

Sources/NIOFileSystem/String+FileSystem.swift
glbrntt: This one still needs addressing. It should be 2025.

natikgadzhi: 🤦🏼  done

glbrntt (approved):

PR #3049, opened 2025-01-13
Ensuring NIOAsyncChannel flushes all writes before closing

adam-fowler:
Currently NIOAsyncChannel can close the channel without having flushed all the writes it has made. 

### Motivation:

I believe there are two situations where writes are not getting written. At the point the NIOAsyncChannel starts the channel close procedure it calls `finish` on the writer which will resume all the pending continuations but then closes the channel immediately so those writes are never written. The second situation is related to the fact that there is no way to wait on a write actually making it all the way through the channel pipeline.

These proposed changes are more to start a discussion and this PR is still very much a draft. 

### Modifications:

Instead of passing Values through the NIOAsyncWriter we pass actions. There are currently three actions 
- `write`: write value to channel
- `writeAndFlush`: write value to channel and flush the pipeline (returning when write promise either fails or succeeds).
- `flush`: Ensure all previous writes have been written. 

### Result:

As long as the last write a user makes is with `writeAndFlush` everything will get written out before the channel is closed.

### DIFF
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
@@ -481,9 +481,10 @@ extension NIOAsyncChannelHandler {
 
     @inlinable
     func _doOutboundWrites(context: ChannelHandlerContext?, writes: Deque<OutboundAction<OutboundOut>>) {
-        for write in writes {
+        // write everything but the last item
+        for write in writes.dropLast() {
             switch write {
-            case .write(let value):
+            case .write(let value), .writeAndFlush(let value, _):
                 guard let context = self.context else {
                     // Already removed from the channel by now, we can stop.
                     return
@@ -492,14 +493,28 @@ extension NIOAsyncChannelHandler {
                 context.flush()
             case .flush(let promise):
                 promise.succeed()
-            case .writeAndFlush(let value, let promise):
-                guard let context = self.context else {
-                    // Already removed from the channel by now, we can stop.
-                    promise.succeed()
-                    return
-                }
-                context.writeAndFlush(Self.wrapOutboundOut(value), promise: promise)
             }
+        }
+        // write last item
+        switch writes.last {
+        case .write(let value):
+            guard let context = self.context else {
+                // Already removed from the channel by now, we can stop.
+                return
+            }
+            context.write(Self.wrapOutboundOut(value), promise: nil)
+            context.flush()
+        case .flush(let promise):
+            promise.succeed()
+        case .writeAndFlush(let value, let promise):
+            guard let context = self.context else {
+                // Already removed from the channel by now, we can stop.
+                promise.succeed()
+                return
+            }
+            context.writeAndFlush(Self.wrapOutboundOut(value), promise: promise)
+        case .none:
+            break
         }
     }
 
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
@@ -14,6 +14,16 @@
 
 import DequeModule
 
+@usableFromInline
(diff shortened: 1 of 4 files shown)


Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
```diff
@@ -430,35 +443,27 @@ struct NIOAsyncChannelHandlerWriterDelegate<Element: Sendable>: NIOAsyncWriterSi
...
-    func _didYield(sequence: Deque<OutboundOut>) {
+    func _didYield(sequence: Deque<OutboundAction<OutboundOut>>) {
         // This is always called from an async context, so we must loop-hop.
         // Because we always loop-hop, we're always at the top of a stack frame. As this
         // is the only source of writes for us, and as this channel handler doesn't implement
         // func write(), we cannot possibly re-entrantly write. That means we can skip many of the
         // awkward re-entrancy protections NIO usually requires, and can safely just do an iterative
         // write.
         self.eventLoop.preconditionInEventLoop()
-        guard let context = self.context else {
```
adam-fowler, line 441: This test has been moved into _doOutboundWrites as we need to complete promises even if the channel handler is no longer there.

Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
```diff
@@ -430,35 +443,27 @@ struct NIOAsyncChannelHandlerWriterDelegate<Element: Sendable>: NIOAsyncWriterSi
...
+    func _didYield(element: OutboundAction<OutboundOut>) {
         // This is always called from an async context, so we must loop-hop.
         // Because we always loop-hop, we're always at the top of a stack frame. As this
         // is the only source of writes for us, and as this channel handler doesn't implement
         // func write(), we cannot possibly re-entrantly write. That means we can skip many of the
         // awkward re-entrancy protections NIO usually requires, and can safely just do an iterative
         // write.
         self.eventLoop.preconditionInEventLoop()
-        guard let context = self.context else {
-            // Already removed from the channel by now, we can stop.
```
adam-fowler, line 459: This test has been moved into _doOutboundWrites as we need to complete promises even if the channel handler is no longer there.

Sources/NIOCore/AsyncChannel/AsyncChannelOutboundWriter.swift
```diff
@@ -72,6 +72,9 @@ public struct NIOAsyncChannelOutboundWriter<OutboundOut: Sendable>: Sendable {
     @usableFromInline
     internal let _backing: Backing
 
+    @usableFromInline
+    internal let eventLoop: EventLoop?
+
```
adam-fowler, line 77: Need an eventLoop to make promises

adam-fowler:
> 1. Introduce an explicit flush method on the `NIOAsyncChannelOutboundWriter` (this is basically your PR here)
> 
> * The problem with this approach is that it extends the public API and users have to handle manual flushing again. Our goal was and is that this is happening implicitly and users can just write and we will make sure they are getting out appropriately
> 

The problem with not extending the public PR is you are trying to match a single `write` API to an api that includes a number of combinations (write, write and wait on promise, writeAndFlush, writeAndFlush and wait on Promise). You have to choose to map to one of these. Currently you are mapping to writeAndFlush. 

The channel pipeline write APIs provide all this functional flexibility. Is there a reason NIOAsyncChannel is wanting to hide it all?

> 2. Create a promise for every write and wait until the write has been written to the socket
> 

Please do not do this. The performance implications are dreadful. 

I did think about doing something a little more internal where any write that could be considered a final write would have a promise created for it internally, but the write function wouldn't wait on this promise completing but it would be stored "somewhere" and when we get to the executeAndClose {} closing phase it would wait on the last promise allocated for a write. 

As I understand it the only way to verify a write has made it all the way through the pipeline is to create a promise for it. I think you'll need a higher level API to indicate the possibility of a write being the final write to avoid creating promises for ever write.

PR #3051, opened 2025-01-13
Clean up sendability in the bootstraps

Lukasa:
### Motivation:

The bootstraps are the next most obvious target for cleaning up Sendability issues. These issues are mostly just missing annotations, but there are a few places where we had actual latent threading bugs that were missed. A lot more of the control flow is made more explicit in this patch, and in general it should get a lot easier to be confident that the code is correct.

### Modifications:

- Add necessary `@Sendable` annotations
- Clean up some incorrect `self` captures of bootstraps, which are not `Sendable`, and which could lead to real threading bugs
- Make a few methods `static` to avoid needing to capture `self` at all.
- Make a few threading assumptions clear by using the isolated views or the sync operations
- Add some missing `Sendable` constraints on interior generic functions.

### Result:

Better safety, better correctness in the bootstraps.


Sources/NIOPosix/Bootstrap.swift
```diff
@@ -474,17 +479,17 @@ public final class ServerBootstrap {
             func fireThroughPipeline(_ future: EventLoopFuture<Void>, context: ChannelHandlerContext) {
                 ctxEventLoop.assertInEventLoop()
                 assert(ctxEventLoop === context.eventLoop)
-                let loopBoundContext = context.loopBound
+                let loopBoundValues = NIOLoopBound((context: context, self: self), eventLoop: context.eventLoop)
```
FranzBusch, line 482: Can't we use the assume isolated APIs here?

Lukasa: Yes, we can. Updated.

FranzBusch: Love it!

FranzBusch (approved):

Lukasa:
Unfortunately this has regressed allocations. For example, in `test_1000_tcpconnections` we've seen 6 extra allocations per loop iteration. These appear to all be coming from the isolated views, so we should take a moment to clean them up.

Lukasa:
Ok, hopefully #3055 should have resolved those regressions, so let's see.

Lukasa:
Oh, it not only resolved those regressions, but led to a benchmark _improvement_. That's nice!

PR #3057, opened 2025-01-14
Fix compiling on 6.1

Lukasa:
Motivation:

Swift 6.1 nightlies have added many new warnings and errors to NIO. This patch resolves all of them.

Modifications:

- Clean up missing imports
- Move away from @_implementationOnly where necessary
- Fix a few things where Sendable mismatches were present.

Result:

Clean compile again

### DIFF
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import DequeModule
+@preconcurrency import DequeModule
 
 /// A ``ChannelHandler`` that is used to transform the inbound portion of a NIO
 /// ``Channel`` into an asynchronous sequence that supports back-pressure. It's also used
--- a/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
+++ b/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-@preconcurrency import DequeModule
+import DequeModule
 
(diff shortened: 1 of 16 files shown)


Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import DequeModule
+@preconcurrency import DequeModule
```
glbrntt, line 15: Do we actually need a `@preconcurrency` import? `Deque` is `Sendable` if its `Element` is `Sendable`.

Sources/NIOHTTP1/HTTPDecoder.swift
```diff
@@ -12,7 +12,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if compiler(>=6.1)
```
glbrntt, line 15: Why is this `>=6.1`? We can do imports with access level from 6.0 IIRC.

glbrntt (changes requested):

Sources/NIOHTTP1/HTTPDecoder.swift
Lukasa: For no better reason than 6.1 is where the errors started to be observed.

Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift
Lukasa: Ah, because we support versions of swift-collections so old that they don't have that conformance. Let's lift the constraint.

Lukasa: Now fixed.

Lukasa:
I'll be merging over the API breakage checker, which is wrong.

PR #3062, opened 2025-01-15
Fix the HappyEyeballsResolver and core Bootstraps under strict concurrency

Lukasa:
### Motivation:

The HappyEyeballsResolver, being an old part of our stack, has a lot of code in it that fails to pass strict concurrency checking. That's deeply suboptimal.

### Modifications:

- Clean up the happy eyeballs resolver under strict concurrency
- Further cleanups to the bootstraps

### Result:

Another step taken on the road to strict concurrency.


glbrntt (approved):
LGTM, thanks Cory

Lukasa:
Ok, I'm going to take the L on this extra allocation. I'm pretty sure this is the result of needing to no longer close over `self` in the bootstraps, and I think the trade on that for extra allocations is appropriate.

Sources/NIOPosix/Bootstrap.swift
```diff
@@ -992,14 +1005,26 @@ public final class ClientBootstrap: NIOClientTCPBootstrapProtocol {
...
         ) { eventLoop, protocolFamily in
-            self.initializeAndRegisterNewChannel(eventLoop: eventLoop, protocolFamily: protocolFamily) {
+            Self.initializeAndRegisterNewChannel(
+                eventLoop: eventLoop,
+                protocolFamily: protocolFamily,
+                enableMPTCP: enableMPTCP,
+                channelInitializer: channelInitializer,
+                channelOptions: channelOptions,
+                bindTarget: bindTarget
+            ) {
```
Lukasa, line 1027: Specifically I think this is our new allocation.

PR #3065, opened 2025-01-16
Add ChannelPipeline.SynchronousOperations.Position

0xTim:
Adds a new `ChannelPipeline.SynchronousOperations.Position` type to fill a hole in the API that prevented non-sendable Channel handlers being added at a position to the pipeline via the synchronous operations.


0xTim:
@Lukasa I've updated the existing public APIs to use the new type to avoid confusing the compiler since the APIs are not usable in their current form. It you'd prefer all the existing APIs to be deprecated and add `@_disfavoredOverload` instead I can do that

Lukasa:
Thanks Tim. Yeah, can we use overloads and `@_disfavoredOverload`? I think it would also be enough to deprecate the old APIs: you can easily test that by omitting the disfavored overload keyword and confirming everything still compiles without warnings.

0xTim:
Updated - I was getting compiler errors due to "ambiguous references..." without the disfavoured overloads so added them in

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -175,11 +175,12 @@ public final class ChannelPipeline: ChannelInvoker {
     ) -> EventLoopFuture<Void> {
         let future: EventLoopFuture<Void>
 
+        let syncPosition = ChannelPipeline.SynchronousOperations.Position(position)
```
Lukasa, line 178: This conversion isn't right: we can't create the synchronous position object until we're on the loop.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1574,6 +1626,38 @@ extension ChannelPipeline {
     }
 }
 
+extension ChannelPipeline.SynchronousOperations {
+    /// A `Position` within the `ChannelPipeline`'s `SynchronousOperations` used to insert non-sendable handlers
+    /// into the `ChannelPipeline` at a certain position.
+    @preconcurrency
```
Lukasa, line 1632: No need for a `preconcurrency` annotation here, we should get this right the first time.

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1574,6 +1626,38 @@ extension ChannelPipeline {
...
+    public enum Position {
```
Lukasa, line 1633: Let's mark this type explicitly non-`Sendable`

Sources/NIOCore/ChannelPipeline.swift
```diff
@@ -1574,6 +1626,38 @@ extension ChannelPipeline {
...
+        /// The last `ChannelHandler` -- the back of the `ChannelPipeline`.
+        case last
+
+        /// Before the given `ChannelHandler`.
+        case before(ChannelHandler)
+
+        /// After the given `ChannelHandler`.
+        case after(ChannelHandler)
+
+        package init(_ position: ChannelPipeline.Position) {
```
Lukasa, line 1646: I can't really see any reason to not make this `public`: we may need this conversion logic in other places.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -83,10 +83,11 @@ extension ChannelPipeline {
     ) -> EventLoopFuture<Void> {
         let future: EventLoopFuture<Void>
 
+        let syncPosition = ChannelPipeline.SynchronousOperations.Position(position)
```
Lukasa, line 86: This should only be created on the EL.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -126,10 +127,11 @@ extension ChannelPipeline {
...
```
Lukasa, line 130: This should only be created on the EL.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -173,10 +175,11 @@ extension ChannelPipeline {
...
```
Lukasa, line 178: This should only be created on the EL.

Sources/NIOHTTP1/HTTPPipelineSetup.swift
```diff
@@ -342,10 +345,11 @@ extension ChannelPipeline {
...
```
Lukasa, line 348: This should only be created on the EL.

Lukasa (commented):
Thanks for this! I've left a few notes in the diff.

Sources/NIOCore/ChannelPipeline.swift
fabianfett: I assume we use the `sending` property here.

Lukasa: We do not, and should not. We should be very reluctant to adopt `sending` until we have entirely Swift 6 usage, or it gets very easy for users to accidentally write code that works in one Swift version but not older ones.

0xTim:
@Lukasa all these should be fixed now

Lukasa (commented):
Ok great, this looks really good. One quick note though, can we add a few very simple tests of these new APIs? Just inserting handlers and confirming they're there is sufficient.

0xTim:
> Ok great, this looks really good. One quick note though, can we add a few very simple tests of these new APIs? Just inserting handlers and confirming they're there is sufficient.

Yes that would be helpful - added!

Tests/NIOPosixTests/ChannelPipelineTest.swift
```diff
@@ -2693,6 +2693,132 @@ class ChannelPipelineTest: XCTestCase {
...
+                IndexWritingHandler(3),
+                position: .after(firstHandler)
+            ).wait()
+        )
+
+        channel.assertReadIndexOrder([1, 3, 2])
+        channel.assertWriteIndexOrder([2, 3, 1])
+    }
+
+    func testAddBeforeForSynchronousPosition() {
```
Lukasa, line 2717: This test isn't actually using the synchronous API.

Tests/NIOPosixTests/ChannelPipelineTest.swift
```diff
@@ -2693,6 +2693,132 @@ class ChannelPipelineTest: XCTestCase {
 
         XCTAssertTrue(try channel.finish().isClean)
     }
+
+    func testAddAfterForSynchronousPosition() {
```
Lukasa, line 2697: This test isn't using the synchronous API.

Tests/NIOPosixTests/ChannelPipelineTest.swift
0xTim: 🤦‍♂️

Lukasa (approved):
Ok, this LGTM. Let's see what CI says.

Lukasa:
Looks like the format check wants some cleanups, mind tackling that for me?

0xTim:
Done

Issue #2770, opened 2024-07-07
ByteBuffer allocator should use malloc_good_size() on Darwin


weissi:
Darwin has a useful function `malloc_good_size()` which returns a good size to malloc. ByteBuffer's allocator should use that instead of just rounding up to the next power of 2.

>     The malloc_good_size() function rounds size up to a value that the allocator implementation can allocate without adding any padding; it then
>     returns that rounded-up value.

On Linux, we should also have a better strategy.

natikgadzhi:
Halp! We could:
1. Just patch the current `_Storage` behavior to respect `malloc_good_size` instead of next power of 2 on Darwin — prop fine, but adds a discrepancy of allocation strategy between platforms, and ditto for Linux.
2. Make `_Storage` swappable and implement `_DarwinStorage` or something that kicks in on, well, Darwin. I don't understand how that is a good idea — Storage is internal anyway, so it's not like our users can swap it out for their own alloc strategy storage.
3. Make `ByteBufferAllocator` responsible for `realloc`, and therefore be able to decide how much to allocate. Document how to make alternative ByteBufferAllocators, and see which APIs we would have to patch to allow people to plug in their allocators if they wish to (i.e. make buffers from files, etc etc). Or perhaps I'm overthinking it. 

Feels like these are solutions in search of two different problems, and you just want pt 1 of this, right? Defer the whole Allocator conversation to https://github.com/apple/swift-nio/issues/2729#issuecomment-2544918078 and just make things slightly more memory efficient here?

Lukasa:
Yeah, I think this just wants the part 1 bit.

natikgadzhi:
Ok. The smallest first step that I can get in a PR is using `malloc_good_size` on Dawrin in the existing `_Storage`, and get this merged.

Step 2 could be to extend this and use some combination of `posix_memalign` and `malloc` on Linux. Problem here is that if we wanted to just plug `posix_memalign` or `aligned_alloc`, we would have to use that _instead_ of `malloc`, and therefore we would have to figure out how to swap out `ByteBufferAllocator`. This sounds like a broader conversation.

PR #3066, opened 2025-01-17
byte-buffer: use `malloc_good_size` on Darwin to allocate memory

natikgadzhi:
### Motivation:

With this change, on Darwin, ByteBuffer will use less memory, and memory allocations will be aligned with pages better, hopefully improving memory efficiency.

This is part 1 of #2770. The next step is to figure out better allocation strategy for Libc, Musl, etc.

### Modifications:

- Introduces a new internal helper method to `_Storage` that returns the optimal storage size
- Uses that to malloc capacity.
- A few tests to verify allocation sizes.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -299,6 +299,19 @@ public struct ByteBuffer {
             _ByteBufferSlice(0..<self.capacity)
         }
 
+        /// Returns the size that this storage would allocate for a given buffer capacity.
+        ///
+        /// On Darwin, use `malloc_good_size` to get optimal size (capacity rounded up, but hopefully not allocating extra memory pages).
+        /// On other platforms, use the next power of 2 clamped to `UInt32.max`.
+        @inlinable
+        static func mallocSize(capacity: _Capacity) -> _Capacity {
+            #if canImport(Darwin)
+            return _Capacity(malloc_good_size(Int(capacity)))
+            #else
+            return capacity.nextPowerOf2ClampedToMax()
+            #endif
+        }
+
         @inlinable
         static func _allocateAndPrepareRawMemory(bytes: _Capacity, allocator: Allocator) -> UnsafeMutableRawPointer {
             let ptr = allocator.malloc(size_t(bytes))!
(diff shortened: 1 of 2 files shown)


Tests/NIOCoreTests/ByteBufferStorageMallocTest.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+
+@testable import NIOCore
+
+#if canImport(Darwin)
```
Lukasa, line 19: Let's make these tests cross-platform by redefining the logic here. We can just have an abstraction function that is essentially equivalent to `mallocSize`, but allows us to validate that that's what we're actually using.

Lukasa (commented):
Mostly great, one quick note!

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -332,7 +345,7 @@ public struct ByteBuffer {
 
         @inlinable
         func reallocStorage(capacity minimumNeededCapacity: _Capacity) {
-            let newCapacity = minimumNeededCapacity.nextPowerOf2ClampedToMax()
+            let newCapacity = minimumNeededCapacity == 0 ? 0 : _Storage.mallocSize(capacity: minimumNeededCapacity)
```
natikgadzhi, line 348: Wait, this is a bug — this should not have the comparison with 0, bad copy & paste.

natikgadzhi:
```suggestion
            let newCapacity = _Storage.mallocSize(capacity: minimumNeededCapacity)
```

Tests/NIOCoreTests/ByteBufferStorageMallocTest.swift
natikgadzhi: @Lukasa apologies for the delay — I think this gets it to a good spot. Once that gets in, I'll get a bit more experience around ByteBuffer and then see if it's useful to figure out better Linux allocation strategies.

Tests/NIOCoreTests/ByteBufferStorageMallocTest.swift
```diff
@@ -0,0 +1,64 @@
...
+        // This will trigger reallocation
+        buffer.writeBytes(Array(repeating: UInt8(0), count: Int(newSize)))
+
+        XCTAssertEqual(Int(buffer._storage.capacity), expectedCapacity)
+    }
+
+    func testZeroCapacity() {
+        let allocator = ByteBufferAllocator()
+        let buffer = allocator.buffer(capacity: 0)
+        XCTAssertEqual(buffer.capacity, 0)
```
Lukasa, line 61: Why are we asking for capacity differently here than in the rest of the tests?

natikgadzhi: Same approach as the rest of the tests would be fine, BUT zero being explicit feels nice.

Lukasa: I think it might be nice to keep the tests discussing capacity the same way.

PR #3067, opened 2025-01-17
`NIOFileSystem`: Try ${TMPDIR} first for temporary directory

### LINKED ISSUE #2861: NIOFileSystem.temporaryDirectory: Consider falling back to TMPDIR
https://github.com/apple/swift-nio/blob/30df8551f4e636b8f68627dbc205221bcfc57782/Sources/NIOFileSystem/FileSystem.swift#L638

Currently, NIOFS's `temporaryDirectory` uses `_CS_DARWIN_USER_TEMP_DIR` on Darwin and `/tmp`/`/data/local/tmp` anywhere else. I think we should consider the env var `TMPDIR` if set.

So maybe:
- Darwin: `_CS_DARWIN_USER_TEMP_DIR` -> `TMPDIR` -> `/tmp`
- Linux: `TMPDIR` -> `/tmp`
- Android: `TMPDIR` -> `/data/local/tmp`
?

natikgadzhi:
### Motivation:

This PR aligns what temp directory NIOFileSystem will return first:
- First try `${TMPDIR}` as it's an expressed user preference
- On Darwin, try `_CS_DARWIN_USER_TMP_DIR` next. If it's set, return that.
- Finally, fall back on `/tmp` on Darwin and Linux, and `/data/local/tmp` on Android.

Closes #2861.

### Modifications:

- Reworks `NIOFileSystem.temporaryDirectory`.

### Result:

- NIOFileSystem will now try `${TMPDIR}` first for the temporary directory.

### Caveats:

- We might want to align how Swift-NIO and FoundationEssentials resolve temp directory, and [FoundationEssentials have a different approach](https://github.com/swiftlang/swift-foundation/blob/9d57f36de757b3d5e3a2f7ffcf27aaec3033509f/Sources/FoundationEssentials/String/String%2BPath.swift#L484-L533). But, I agree with [this comment](https://github.com/apple/swift-nio/issues/2861#issuecomment-2327010713), it feels like TMPDIR is an expected default.


Lukasa:
Thanks for this patch! I've asked @glbrntt to review, as he's most familiar with this section of the code.

finagolfin:
Thanks, I prefer this for Android and currently patch this to use [Foundation's `temporaryDirectory` logic instead, both on my Android CI and when testing this repo natively in the Termux app](https://github.com/finagolfin/swift-android-sdk/blob/0ea8da997bc3195b80a83c255b100eaad3884e74/swift-nio-filesystem.patch#L19), which I can finally remove once this goes in. 👍

Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
```diff
@@ -1681,6 +1681,22 @@ extension FileSystemTests {
...
+        defer {
+            if let originalTMPDIR {
+                setenv("TMPDIR", String(cString: originalTMPDIR), 1)
+            } else {
+                unsetenv("TMPDIR")
+            }
+        }
+
+        let customTmpDir = "/custom/tmp/dir"
+        setenv("TMPDIR", customTmpDir, 1)
```
glbrntt, line 1695: I'm a bit hesitant to mess around with `TMPDIR` like this in case other tests running in parallel are relying on it. 

II think my preference here would be to check if `TMPDIR` is set and then check `fs.temporaryDirectory` if it was set.

glbrntt (changes requested):
Thanks, the "actual" change looks great. I left a comment on the test though.

Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
natikgadzhi: My bad, I haven't considered tests running in parallel, I see how this is potentially problematic.

glbrntt (approved):
Thank you!

finagolfin:
Thanks for getting this in, Natik. I was able to remove my similar patch from my Android CI that I linked above, and I can now build this repo natively on Android without _any_ local patches in [the Termux app, which sets its own `TMPDIR`](https://termux.dev/en/). 🥳

PR #3070, opened 2025-01-17
Move Isolated EL operations to EL protocol

Lukasa:
Motivation

Right now the isolated view on the EventLoop has to implement its interface in terms of UnsafeTransfer, because there are no non-Sendable operations it can use.

That hurts us because we have to allocate an extra closure for all these operations, making them slower than they need to be. This is unavoidable in some contexts, but most EventLoops would be able to offer a fast-path that avoids this extra allocation.

To achieve that, we need to move this logic into a customisation point on the EventLoop protocol. Our fallback default implementation can be the same as we do today.

Modifications

- Add customisation points for EventLoop for each of its operations that has a protocol witness already.
- Add default implementations of these, using the implementation from EventLoop.Isolated.

Results

No behavioural changes, but the code has moved.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -395,6 +395,69 @@ public protocol EventLoop: EventLoopGroup {
     ///
     /// - NOTE: Event loops only need to implemented this if they provide a custom scheduled callback implementation.
     func cancelScheduledCallback(_ scheduledCallback: NIOScheduledCallback)
+
+    /// Submit a given task to be executed by the ``EventLoop``, from a context where the caller
+    /// statically knows that the context is isolated.
+    ///
+    /// This is an optional performance hook. ``EventLoop`` implementers are not required to implement
+    /// this witness, but may choose to do so to enable better performance of the isolated EL views. If
+    /// they do so, ``EventLoop/Isolated/execute`` will perform better.
+    func _executeIsolatedUnsafeUnchecked(_ task: @escaping () -> Void)
+
+    /// Submit a given task to be executed by the ``EventLoop```, from a context where the caller
+    /// statically knows that the context is isolated.
+    ///
+    /// Once the execution is complete the returned ``EventLoopFuture`` is notified.
+    ///
+    /// This is an optional performance hook. ``EventLoop`` implementers are not required to implement
+    /// this witness, but may choose to do so to enable better performance of the isolated EL views. If
+    /// they do so, ``EventLoop/Isolated/submit`` will perform better.
+    ///
+    /// - Parameters:
+    ///   - task: The closure that will be submitted to the ``EventLoop`` for execution.
+    /// - Returns: ``EventLoopFuture`` that is notified once the task was executed.
+    func _submitIsolatedUnsafeUnchecked<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T>
+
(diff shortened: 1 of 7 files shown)


Sources/NIOCore/EventLoop.swift
```diff
@@ -422,6 +485,50 @@ extension EventLoop {
...
+    public func _executeIsolatedUnsafeUnchecked(_ task: @escaping () -> Void) {
+        self.assertInEventLoop()
+        let unsafeTransfer = UnsafeTransfer(task)
+        self.execute {
+            unsafeTransfer.wrappedValue()
+        }
+    }
+
+    /// Default implementation: wraps the task in an UnsafeTransfer.
+    public func _submitIsolatedUnsafeUnchecked<T>(_ task: @escaping () throws -> T) -> EventLoopFuture<T> {
```
glbrntt, line 499: Should these be `@inlinable`?

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -122,10 +110,11 @@ public struct NIOIsolatedEventLoop {
...
         _ task: @escaping () throws -> EventLoopFuture<T>
     ) -> Scheduled<T> {
-        let unsafeTransfer = UnsafeTransfer(task)
-        return self._wrapped.flatScheduleTask(deadline: deadline, file: file, line: line) {
-            try unsafeTransfer.wrappedValue()
-        }
+        let promise: EventLoopPromise<T> = self._wrapped.makePromise(file: file, line: line)
+        let scheduled = self.scheduleTask(deadline: deadline, task)
+
+        scheduled.futureResult.flatMap { $0 }.cascade(to: promise)
```
glbrntt, line 116: Can we save an alloc here if we use `scheduled.futureResult.whenComplete { ... }`?

glbrntt (commented):
Couple of questions but looks good otherwise

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
Lukasa: I don't think so: the `flatMap` unwraps an extra level of futureness.

Sources/NIOCore/EventLoop.swift
Lukasa: Sure, added those.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
glbrntt: I mean like this (I appreciate it's more code than the `flatMap` + `cascade`):

```swift
scheduled.futureResult.whenComplete { result in
    switch result {
    case .success(let futureResult):
        promise.completeWith(futureResult)
    case .failure(let error):
        promise.fail(error)
    }
}
```

Lukasa: Yes, you're probably right. However, all the `flatScheduleTask` implementations in swift-nio follow this spelling. I've filed #3071 to track making the improvement generically, and so for now would rather keep this in line with the others.

glbrntt: Thanks for filing that, makes sense to track them together.

glbrntt (approved):

PR #3073, opened 2025-01-20
Fast-path isolated view for EmbeddedEventLoop

Lukasa:
Motivation:

This patch takes advantage of the fast-path for isolated views on EmbeddedEventLoop, reducing overhead when using EmbeddedEL.

Modifications:

- Implement the new witnesses in EmbeddedEL

Result:

EmbeddedEL is faster when using isolated views.

### DIFF
--- a/Sources/NIOEmbedded/Embedded.swift
+++ b/Sources/NIOEmbedded/Embedded.swift
@@ -381,6 +381,7 @@ public final class EmbeddedEventLoop: EventLoop, CustomStringConvertible {
         in delay: TimeAmount,
         _ task: @escaping () throws -> T
     ) -> Scheduled<T> {
+        // Because of the way EmbeddedEL works, we can delegate this directly to schedule.
         self.scheduleTask(in: delay, task)
     }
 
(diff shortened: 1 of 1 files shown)


Sources/NIOEmbedded/Embedded.swift
```diff
@@ -355,6 +355,35 @@ public final class EmbeddedEventLoop: EventLoop, CustomStringConvertible {
...
+        // Because of the way EmbeddedEL works, we can delegate this directly to schedule.
+        self.scheduleTask(deadline: deadline, task)
+    }
+
+    @discardableResult
+    public func _scheduleTaskIsolatedUnsafeUnchecked<T>(
+        in delay: TimeAmount,
+        _ task: @escaping () throws -> T
+    ) -> Scheduled<T> {
+        self.scheduleTask(in: delay, task)
```
gjcairo, line 384: Nit: do we want to add the same "because of the way EmbeddedEL works..." comment here?

gjcairo (approved):

Sources/NIOEmbedded/Embedded.swift
Lukasa: Sure, I'll add that

PR #3075, opened 2025-01-20
Adopt optimized isolated path for SelectableEL

Lukasa:
Motivation:

Users of SelectableEventLoop deserve to have their code that uses the isolated event loop view perform at least as well as the code that doesn't.

Modifications:

- Adopt the extra protocol witnesses.
- Refactor out common code.

Result:

Better performance


Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -308,6 +308,29 @@ internal final class SelectableEventLoop: EventLoop, @unchecked Sendable {
...
+            promise.fail(error)
+        }
+
+        return scheduled
+    }
+
+    /// - see: `EventLoop.scheduleTask(in:_:)`
+    @inlinable
+    internal func scheduleTask<T>(in: TimeAmount, _ task: @escaping () throws -> T) -> Scheduled<T> {
+        scheduleTask(deadline: .now() + `in`, task)
```
glbrntt, line 325:
```suggestion
        self.scheduleTask(deadline: .now() + `in`, task)
```

glbrntt (approved):

PR #3076, opened 2025-01-21
Update nightly CI pipelines to 6.1 instead of 6.0

rnro:
### Motivation:

Active development has moved to 6.1 so the Swift 6.0 nightlies no longer deliver much value.

### Modifications:

* Deprecate 6.0 nightly CI configuration parameters
* Introduce new 6.1 nightly CI configuration parameters
* Update the matrix generation script to handle 6.1
 
### Result:

* No more nightly 6.0 CI, 6.1 nightlies are introduced.
* Configuration for 6.0 nightlies no longer has any effect.
* All repos are automatically opted-in to 6.1 nightly linux pipelines **even if they opted out of 6.0**.


.github/workflows/unit_tests.yml
```diff
@@ -29,11 +29,19 @@ on:
         default: ""
       linux_nightly_6_0_enabled:
         type: boolean
-        description: "Boolean to enable the Linux nightly 6.0 Swift version matrix job. Defaults to true."
+        description: "⚠️  Deprecated, we no longer run against 6.0 nightly."
```
Lukasa, line 32: Should these defaults be changed?

rnro: I don’t think it matters, we just drop it on the floor now

Lukasa (approved):

PR #3078, opened 2025-01-22
more 6.0 nightly to 6.1 nightly CI changes

rnro:
### Motivation:

Some changes were missed in #3076, passing through parameters for the new 6.1 version.

### Modifications:

Pass through more parameters, clean up misleading comments.

### Result:

6.1 nightly runs will be more like 6.0 nightly runs were


Lukasa:
Can you add the 6.1 thresholds so the integration tests pass?

Issue #3071, opened 2025-01-20
Improve flatScheduleTask


Lukasa:
swift-nio has multiple implementations of `flatScheduleTask` which are all basically identical:

```
        let promise: EventLoopPromise<T> = self.makePromise(file: file, line: line)
        let scheduled = self.scheduleTask(deadline: deadline, task)

        scheduled.futureResult.flatMap { $0 }.cascade(to: promise)
        return .init(promise: promise, cancellationTask: { scheduled.cancel() })
```

Each of these is meaningfully different, e.g. in whether they take a deadline or a delay, or whether they require sendable closures and results. 

@glbrntt has pointed out that the line `scheduled.futureResult.flatMap { $0 }.cascade(to: promise)`, which is common to all of them, is potentially suboptimal. A better implementation would be:

```
scheduled.futureResult.whenComplete { result in
    switch result {
    case .success(let futureResult):
        promise.completeWith(futureResult)
    case .failure(let error):
        promise.fail(error)
    }
}
```

We should change the implementations, and then confirm that we do actually see an allocation reduction from the change. Assuming we do, we can safely merge it.

PR #3079, opened 2025-01-22
Improve `flatScheduleTask`

clintonpi:
Motivation:

As pointed out in #3071, the `flatScheduleTask` implementations can be improved.

Modifications:

- Refactor the `flatScheduleTask` implementations to skip `flatMap` calls, which avoids creating an extra promise.
- As there is now a lower number of allocations, reduce the necessary thresholds for the allocation tests.

Result:

Reduction in the number of allocations in the package.


glbrntt (approved):
Looks great, thanks @clintonpi!

PR #3081, opened 2025-01-23
Clean up Sendability in Thread and ThreadPool

Lukasa:
Motivation

Next on my journey of discovery is NIOThread and NIOThreadPool. These are basically fine, but they need some Sendability annotations and a bit of a cleanup.

Modifications

Add some Sendability
Clean up some shutdown logic to avoid semaphores.

Result

Another step forward to a strict-concurrency safe world.

### DIFF
--- a/Sources/NIOPosix/ThreadPosix.swift
+++ b/Sources/NIOPosix/ThreadPosix.swift
@@ -74,8 +74,27 @@ private func sysPthread_create(
 
 typealias ThreadOpsSystem = ThreadOpsPosix
 
+struct PthreadWrapper: @unchecked Sendable {
+    var handle: pthread_t
+}
+
+extension Optional where Wrapped == PthreadWrapper {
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/ThreadPosix.swift
```diff
@@ -74,8 +74,27 @@ private func sysPthread_create(
 
 typealias ThreadOpsSystem = ThreadOpsPosix
 
+struct PthreadWrapper: @unchecked Sendable {
```
FranzBusch, line 77: Nit: Can't we use our existing `UnsafeTransfer` for this?

Lukasa: We could, but in this case I wanted a special-case wrapper that we can see, and aim to remove when it's no longer needed.

Lukasa:
The API breakages are expected.

PR #3089, opened 2025-01-24
Avoid using stderr explicitly

Lukasa:
Motivation

It's tempting for us to use the FILE * for stderr whenever we want to print to it. However, the Linux libc modules aren't appropriately annotated to call that field constant, and indeed it isn't. This causes Swift to throw a fit if we even touch it.

Modifications

Introduce an awkward function that issues direct writes to stderr instead of using the FILE *.

Result

Swift is happier.

### DIFF
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -19,6 +19,22 @@ import NIOConcurrencyHelpers
 import NIOCore
 import _NIODataStructures
 
+private func printError(_ string: StaticString) {
+     string.withUTF8Buffer { buf in
+         var buf = buf
+         while buf.count > 0 {
+             // 2 is stderr
+             let rc = write(2, buf.baseAddress, buf.count)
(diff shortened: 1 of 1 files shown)


Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -19,6 +19,22 @@ import NIOConcurrencyHelpers
...
+        var buf = buf
+        while buf.count > 0 {
+            // 2 is stderr
+            let rc = write(2, buf.baseAddress, buf.count)
+            if rc < 0 {
+                let err = errno
+                if err == EINTR { continue }
+                fatalError("Unexpected error writing: \(err)")
+            }
+            buf = .init(rebasing: buf.dropFirst(Int(rc)))
```
rnro, line 33: Is using `Int` here going to trip us up on some less common platform? Is it being used as a proxy to express some known number of bytes which we might be better using?

gjcairo (approved):

PR #3090, opened 2025-01-24
Make NIOPosix strict-concurrency clean

Lukasa:
Motivation

NIOPosix should be clean under strict concurrency

Modifications

- Fix newly-added issue in GetaddrinfoResolver
- Fix newly-added issue in SEL
- Lock in the win in Package.swift

Result

We did it!


glbrntt (approved):
🎉

PR #3091, opened 2025-01-26
Make `NIOLoopBoundBox` off-EL with sending

### LINKED ISSUE #2754: Give NIOLoopBoundBox a `sending` off-EL initialiser once compiler is ready
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0430-transferring-parameters-and-results.md will be useful to construct `NIOLoopBoundBox` with a non-`Sendable` value off EL.

But the compiler isn't ready quite yet.

This is a followup for #2753 (which allows `Sendable` values)

```
#if compiler(>=6.0) // `sending` is >= 6.0
extension NIOLoopBoundBox {
    /// Initialise a ``NIOLoopBoundBox`` by `sending` (i.e. transferring) a value, validly callable off `eventLoop`.
    ///
    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because we are `sending` the value across the isolation domain.
    /// Because we're `sending` `value`, we just need to protect it against mutations (because nothing else can have access to it anymore).
    public static func makeBoxSendingValue(
        _ value: sending Value,
        as: Value.Type = Value.self,
        eventLoop: EventLoop
    ) -> NIOLoopBoundBox<Value> {
        ...
    }
}
#endif

### PR DESCRIPTION
**Pull request by natikgadzhi**
## Motivation:

This closes #2754. It's a follow-up for #2753 that adds a way to `makeLoopBoundBox` by sending a value, leaning into Swift 6 concurrency!

/cc @weissi and @FranzBusch.

## Modifications:

- Adds `makeBoxSendingValue`. Can't take the credit, @weissi wrote it in the issue.
- Adds a test for it.

## Result:

Nice NIOLoopBoundBox API for Swift 6.

### DIFF
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -17,7 +17,7 @@
 ///
 /// ``NIOLoopBound`` is useful to transport a value of a non-`Sendable` type that needs to go from one place in
 /// your code to another where you (but not the compiler) know is on one and the same ``EventLoop``. Usually this
-/// involves `@Sendable` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-2fxvb``)
+/// involves `@Sendable` or `sending` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-2fxvb``)
 /// that this is actually true.
 ///
 /// A ``NIOLoopBound`` can only be constructed, read from or written to when you are provably
@@ -59,7 +59,7 @@ public struct NIOLoopBound<Value>: @unchecked Sendable {
 ///
 /// ``NIOLoopBoundBox`` is useful to transport a value of a non-`Sendable` type that needs to go from one place in
 /// your code to another where you (but not the compiler) know is on one and the same ``EventLoop``. Usually this
-/// involves `@Sendable` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-7ukrq``)
+/// involves `@Sendable` or `sending` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-7ukrq``)
 /// that this is actually true.
 ///
 /// A ``NIOLoopBoundBox`` can only be read from or written to when you are provably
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Sources/NIOCore/NIOLoopBound.swift [hunk 1]
```diff
@@ -17,7 +17,7 @@
 ///
 /// ``NIOLoopBound`` is useful to transport a value of a non-`Sendable` type that needs to go from one place in
 /// your code to another where you (but not the compiler) know is on one and the same ``EventLoop``. Usually this
-/// involves `@Sendable` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-2fxvb``)
+/// involves `@Sendable` or `sending` closures. This type is safe because it verifies (using ``EventLoop/preconditionInEventLoop(file:line:)-2fxvb``)
```
natikgadzhi, line 20: Unrelated, but thought this would be more precise.

Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -142,3 +142,26 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
...
+    /// Initialise a ``NIOLoopBoundBox`` by `sending` (i.e. transferring) a value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because we are `sending` the value across the isolation domain.
+    /// Because we're `sending` `value`, we just need to protect it against mutations (because nothing else can have access to it anymore).
+    public static func makeBoxSendingValue(
+        _ value: sending Value,
+        as: Value.Type = Value.self,
+        eventLoop: EventLoop
+    ) -> NIOLoopBoundBox<Value> {
+        // Here, we -- possibly surprisingly -- do not precondition being on the EventLoop. This is okay for a few
```
natikgadzhi, line 157: The comment is a copy-pasta from the other off-EL initializers, but I think it's fair to keep this for consistency.

Tests/NIOPosixTests/NIOLoopBoundTests.swift
```diff
@@ -75,6 +75,43 @@ final class NIOLoopBoundTests: XCTestCase {
...
+    }
+
+    func testLoopBoundBoxCanBeInitialisedWithSendingValueOffLoopAndLaterSetToValue() {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: 1)
+        defer {
+            XCTAssertNoThrow(try group.syncShutdownGracefully())
+        }
+
+        let loop = group.any()
+
```
natikgadzhi, line 93:
```suggestion
```

Tests/NIOPosixTests/NIOLoopBoundTests.swift
```diff
@@ -75,6 +75,43 @@ final class NIOLoopBoundTests: XCTestCase {
         )
     }
 
+    #if compiler(>=6.0)
+    class NonSendableThingy {
```
natikgadzhi, line 79: Could be an overkill. Also, should I explicitly mark this fileprivate?

natikgadzhi, line 79:
```suggestion
    fileprivate class NonSendableThingy {
```

Sources/NIOCore/NIOLoopBound.swift
natikgadzhi: Related: should I add more explicit documentation outlining that there are a bunch of functions to make a `NIOLoopBoundBox` off the event loop called `make*`, and outline the use cases for them at a high level? Separate PR?

Lukasa (changes requested):
I'm actually a bit nervous about merging this.

We have a number of places where our `Sendable` constraints can be relaxed to `sending`, but I don't think we should actually do that yet. Many projects that depend on NIO support multiple Swift versions, just as we do, and it makes life quite a bit harder when doing implementation to have the APIs be different between those Swift versions.

In some cases this is acceptable, but in this instance I'm inclined to suggest we should wait, and do this work when we drop all compilers below 6.0. That would be at the release of Swift 6.2 (presuming no new Swift language versions).

weissi:
> I'm actually a bit nervous about merging this.
> 
> We have a number of places where our `Sendable` constraints can be relaxed to `sending`, but I don't think we should actually do that yet. Many projects that depend on NIO support multiple Swift versions, just as we do, and it makes life quite a bit harder when doing implementation to have the APIs be different between those Swift versions.
> 
> In some cases this is acceptable, but in this instance I'm inclined to suggest we should wait, and do this work when we drop all compilers below 6.0. That would be at the release of Swift 6.2 (presuming no new Swift language versions).

What about adding it _now_ but underscored and when 6.0 is the floor making it proper public?

That way folks who have a 6.0 compiler can use it straight away?

Lukasa:
We can do that, but the 6.2 move would be to just change the other initializer to drop the Sendable constraint. So we’d end up with two essentially identical initializers.

PR #3110, opened 2025-02-05
Moving `PooledRecvBufferAllocator` from NIOPosix to NIOCore

rafaelcepeda:
Moving `PooledRecvBufferAllocator` from NIOPosix to NIOCore, making it part of its public interface.

### Motivation:

We want to introduce the capability of reusing receiving buffers from a pool into swift-nio-transport-services project. To prevent duplicating this functionality there, this change is making the existing functionality from NIO part of its public API. 

### Modifications:

Moved the type `PooledRecvBufferAllocator` from NIOPosix module to NIOCore and turned it into a public type.

### Result:

`PooledRecvBufferAllocator` is part of NIOCore public API.


Sources/NIOCore/PooledRecvBufferAllocator.swift
```diff
@@ -58,7 +61,7 @@ internal struct PooledRecvBufferAllocator {
     }
 
     /// Update the capacity of the underlying buffer pool.
-    mutating func updateCapacity(to newCapacity: Int) {
+    public mutating func updateCapacity(to newCapacity: Int) {
```
gjcairo, line 64: Can we update the docs for this and the other `public` methods to include a list of arguments (i.e. the `Parameters` keyword as in the `public init` above)?

Sources/NIOCore/PooledRecvBufferAllocator.swift
```diff
@@ -12,10 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
-
 /// A receive buffer allocator which cycles through a pool of buffers.
```
Lukasa, line 15: We should extend this to express when users might want to use this type.

Sources/NIOCore/PooledRecvBufferAllocator.swift
rafaelcepeda: I've added the missing documentation. Let me know if you think I should change anything else.

Sources/NIOCore/PooledRecvBufferAllocator.swift
gjcairo: Allocating is always expensive, so I'm not sure this is particularly insightful. Maybe something like:

> Channels can read multiple times per cycle (based on `ChannelOptions.maxMessagesPerRead`), and they reuse the inbound buffer for each read. If a `ChannelHandler` holds onto this buffer, then CoWing will be needed. A `PooledRecvBufferAllocator` cycles through preallocated buffers to avoid CoWs during the same read cycle.

rafaelcepeda: I added your suggestion in the documentation.

Sources/NIOCore/PooledRecvBufferAllocator.swift
```diff
@@ -81,14 +89,21 @@ internal struct PooledRecvBufferAllocator {
 
     /// Record the number of bytes which were read.
     ///
-    /// Returns whether the next buffer will be larger than the last.
-    mutating func record(actualReadBytes: Int) {
+    /// - Parameters:
+    ///   - actualReadBytes: Number of bytes being recorded
+    /// - Returns: whether the next buffer will be larger than the last.
```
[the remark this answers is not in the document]
rafaelcepeda: Good catch. Removed from the code.

Sources/NIOCore/PooledRecvBufferAllocator.swift
```diff
@@ -12,10 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
-import NIOCore
-
 /// A receive buffer allocator which cycles through a pool of buffers.
-internal struct PooledRecvBufferAllocator {
+///
+/// This type is useful when allocating new buffers for every IO read is expensive or undesirable.
+public struct PooledRecvBufferAllocator {
```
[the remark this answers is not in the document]
rafaelcepeda: I've added the prefix. Great catch.

Sources/NIOCore/PooledRecvBufferAllocator.swift
```diff
@@ -81,14 +89,21 @@ internal struct PooledRecvBufferAllocator {
...
 
     /// Provides a buffer with enough writable capacity as determined by the underlying
     /// receive allocator to the given closure.
-    mutating func buffer<Result>(
+    ///
+    /// - Parameters:
+    ///    - allocator: `ByteBufferAllocator` used to construct a new buffer if needed
+    ///    - body: Closure where the caller can use the new or existing buffer
+    /// - Returns: A tuple containing the `ByteBuffer` used and the `Result` yielded by the closure provided.
+    public mutating func buffer<Result>(
```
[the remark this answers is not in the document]
rafaelcepeda: I've changed this and related functions/properties to be `@inlinable/@usableFromInline` as you suggested.

Sources/NIOCore/NIOPooledRecvBufferAllocator.swift
```diff
@@ -12,147 +12,170 @@
...
 
-import NIOCore
-
 /// A receive buffer allocator which cycles through a pool of buffers.
-internal struct PooledRecvBufferAllocator {
+///
+/// Channels can read multiple times per cycle (based on `ChannelOptions.maxMessagesPerRead`), and they reuse
+/// the inbound buffer for each read. If a `ChannelHandler` holds onto this buffer, then CoWing will be needed.
+/// A `NIOPooledRecvBufferAllocator` cycles through preallocated buffers to avoid CoWs during the same read cycle.
+public struct NIOPooledRecvBufferAllocator {
```
glbrntt, line 20: As a public type we should annotate whether this is `Sendable` or not. It _looks_ like it is.

glbrntt (changes requested):
One nit but LGTM otherwise

Sources/NIOCore/NIOPooledRecvBufferAllocator.swift
rafaelcepeda: I've introduced the change requested.

glbrntt (approved):

PR #3112, opened 2025-02-07
Enable strict concurrency checking for NIOTLS

rnro:
### Motivation:

To ensure NIOTLS concurrency safety.

### Modifications:

* Require `NegotiationResult` to be `Sendable`
  * In `NIOTypedApplicationProtocolNegotiationHandler<NegotiationResult>` the
result is used to fulfil promises and perform hops so we need it to be
sendable when accessed from these other concurrency domains.

### Result:

Builds will warn and CI will fail if regressions are introduced.


Lukasa (approved):
Nice, this LGTM.

PR #3113, opened 2025-02-10
Add bpftrace allocations script

Lukasa:
Motivation:

When debugging allocations it is frequently helpful to be able to use bpftrace instead of the more limited heaptrack.

Modifications:

Offer an equivalent of malloc-aggregation.d for bpftrace.

Result:

Allocation recording on Linux is available.

### DIFF
--- a/.licenseignore
+++ b/.licenseignore
@@ -45,5 +45,6 @@ dev/git.commit.template
 dev/lldb-smoker
 dev/make-single-file-spm
 dev/malloc-aggregation.d
+dev/malloc-aggregation.bt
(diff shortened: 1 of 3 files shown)


glbrntt (approved):
Nice!

glbrntt (changes requested):
Looks like you need to add `dev/malloc-aggregation.bt` to: https://github.com/apple/swift-nio/blob/main/.licenseignore

Lukasa:
Done.

glbrntt (approved):

PR #3117, opened 2025-02-13
Add more methods to work with Isolated futures

rnro:
### Motivation:

Whilst working on the NIOHTTP1 strict concurrency work I encountered a few gaps in the API where I needed isolated analogues of methods on futures.

### Modifications:

* Add `makeSucceededIsolatedFuture`
* Add a new internal initializer for `EventLoopFuture` which does not require the value to be Sendable for use in isolated contexts.
* Add a variant of `flatMapError` which can handle closures which return isolated futures
* Add a `futureResult` computed property which allows us to avoid the otherwise clumsy `future.nonIsolated().futureResult.assumeIsolated()`

### Result:

More ability to work with isolated futures.

### DIFF
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -1107,7 +1107,7 @@ extension EventLoop {
         }
     }
 
-    /// Creates and returns a new isolated `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop` as execution `NIOThread`.
+    /// Creates and returns a new isolated `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop.
     ///
     /// - Parameters:
     ///   - value: the value that is used by the `EventLoopFuture.Isolated`.
--- a/Sources/NIOCore/EventLoop.swift
+++ b/Sources/NIOCore/EventLoop.swift
@@ -1107,6 +1107,22 @@ extension EventLoop {
         }
     }
 
+    /// Creates and returns a new isolated `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop` as execution `NIOThread`.
+    ///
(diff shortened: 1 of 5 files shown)


Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -643,6 +680,11 @@ extension EventLoopPromise {
             self._wrapped = _wrapped
         }
 
+        @usableFromInline
+        var futureResult: EventLoopFuture<Value>.Isolated {
```
Lukasa, line 684: As you've added this, we may as well make it public and inlinable.

Lukasa (commented):
Basically looking good, a few notes.

Sources/NIOCore/EventLoop.swift
```diff
@@ -1107,6 +1107,22 @@ extension EventLoop {
         }
     }
 
+    /// Creates and returns a new isolated `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop` as execution `NIOThread`.
```
weissi, line 1110:
```suggestion
    /// Creates and returns a new isolated `EventLoopFuture` that is already marked as success. Notifications will be done using this `EventLoop.
```

weissi: NIOThread isn't public and also not all ELs are implemented using NIOThread

weissi: oh, this seems to be in other places too. Probably some very old oversight ;)

Lukasa: Yeah, it's a worthy doc comment change in the general case, but we can avoid making it again here.

glbrntt (approved):

PR #3119, opened 2025-02-23
Equatable SocketAddressError

0xTim:
Make `SocketAddressError` `Equatable`. This is useful when running tests that you expect to throw this error as it allows you to write for example
```swift
await #expect(throws: SocketAddressError.unknown(host: "127.0.0.1", port: Int.max)) {
    try await app.startup()
}
```


Lukasa (commented):
Let’s make it Hashable while we’re here.

Lukasa (approved):
Nice one, thanks Tim!

PR #3121, opened 2025-02-25
Changed NIOAsyncChannel.executeThenClose to accept actor-isolated par…

rafaelcepeda:
Improve usability of NIOAsyncChannel.executeThenClose function from an Actor

### Motivation:

When calling executeThenClose from inside an actor, the closure which we pass is not isolated to the same actor
This makes it hard to use

e.g. the following code won’t compile on Swift 6 because the closure is `self-isolated` so can’t be passed

```
await withDiscardingTaskGroup { group in
            do {
                try await serverChannel.executeThenClose { inboundStream, _ in
                    try await self.handleInboundStream(inboundStream: inboundStream, group: &group)
                }
             } catch {}
}
```

Making the closure as @Sendable allows it to be passed, but then we can’t access `group` anymore

### Modifications:

Added actor isolation parameter to the `executeThenClose` function. 

### Result:

`executeThenClose` will be able to be asynchronously called from inside an Actor, as it will now be part of the Actor's isolation domain.


Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -283,8 +283,11 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
     ///
     /// - Important: After this method returned the underlying ``Channel`` will be closed.
     ///
-    /// - Parameter body: A closure that gets scoped access to the inbound and outbound.
+    /// - Parameter:
+    ///     - isolatedTo: actor where this function should be isolated to
```
gjcairo, line 287: Nit: the parameter name here is `actor` - `isolatedTo` is the label (you can think of it as a part of the method's name).

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -283,8 +283,11 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+    ///     - body: A closure that gets scoped access to the inbound and outbound.
     public func executeThenClose<Result>(
+        isolatedTo actor: isolated(any Actor)? = #isolation,
```
gjcairo, line 290: This change is actually API-breaking: even if the parameter you add is defaulted, the function signature changes.

For instance, imagine you have:
```swift
func someFunc(foo: Int) { ... }

func doSomething() {
  let f: (Int) -> Void  = someFunc(foo:)
  // ...
}
```

Say you add a new defaulted parameter...
```swift
func someFunc(foo: Int, bar: String = "baz") { ... }

func doSomething() {
  let f1: (Int) -> Void  = someFunc(foo:)  //  this will fail to build: "no member named someFunc(foo:)"
  // Should be: 
  let f2: (Int, String?) -> Void  = someFunc(foo:bar:)
}
```


You will have to add a new `executeThenClose` overload that takes a (non-optional) isolated param.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
rafaelcepeda: Fixed the issue.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
rafaelcepeda: Great point. I've added the old API back and have implemented it using the new one.

hamzahrmalik (commented):
Thanks for this change, it will make it much easier to use asyncchannels!

Couple of points:
1) this will not work on older compilers, so probably needs some #if compiler >= 6.0 sprinkled around
2) following on from gus's comment, are we sure the overload resolution will pick the right thing here? Can we add a test which uses this from an actor to make sure that compiles fine?

The testcase would be something like having an actor-isolated function call executeThenClose and then attempt to reference self inside the closure

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -283,10 +283,25 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+
+    /// Provides scoped access to the inbound and outbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameter:
+    ///     - actor: actor where this function should be isolated to
+    ///     - body: A closure that gets scoped access to the inbound and outbound.
+    public func executeThenClose<Result>(
+        isolatedTo actor: isolated (any Actor)?,
```
rnro, line 302: We seem to have dropped the defaulted parameter here which I think we want.

rnro:
Looks good in general, unfortunately there are some fiddly considerations here. Here's a link to a similar block of code which includes the compiler check which Hamzah mentions https://github.com/apple/swift-service-context/blob/8946c930cae601452149e45d31d8ddfac973c3c7/Sources/ServiceContextModule/ServiceContext.swift#L232-L266

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
gjcairo: The problem is the compiler picking the right func if we default this parameter (i.e. if you call `executeThenClose { ... }` are we calling the old version, or the one defaulting `isolation`?). 
We could use `@_disfavoredOverload`, but have we agreed that is the route we generally want to take?

Specifically in this case, if what we want is to enable adopters to use `executeThenClose` from actors, I don't think it's particularly bad to have to explicitly pass `self` as `isolation`, as it makes it clear you want to run this _on the actor_, even though you're `await`ing and would otherwise be hopping off the isolation domain.

rafaelcepeda: As spoken offline, we decided to favour the use of `@_disfavoredOverload` with this API. I'll change the code.

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -301,6 +302,53 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
...
+    #if compiler(>=6.0)
+    /// Provides scoped access to the inbound and outbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameters:
+    ///     - actor: actor where this function should be isolated to
+    ///     - body: A closure that gets scoped access to the inbound and outbound.
+    public func executeThenClose<Result>(
+        isolatedTo actor: isolated (any Actor)? = #isolation,
```
FranzBusch, line 335: I think this follows more what we have been doing elsewhere https://github.com/swift-server/swift-service-lifecycle/blob/e2b442cc73b8100f16df23eb2431aa0d4660fc19/Sources/ServiceLifecycle/GracefulShutdown.swift#L41
```suggestion
        isolation actor: isolated (any Actor)? = #isolation,
```

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -332,13 +381,61 @@ extension NIOAsyncChannel {
...
+    #if compiler(>=6.0)
+    /// Provides scoped access to the inbound side of the underlying ``Channel``.
+    ///
+    /// - Important: After this method returned the underlying ``Channel`` will be closed.
+    ///
+    /// - Parameters:
+    ///     - actor: actor where this function should be isolated to
+    ///     - body: A closure that gets scoped access to the inbound.
+    public func executeThenClose<Result>(
+        isolatedTo actor: isolated (any Actor)? = #isolation,
```
FranzBusch, line 431:
```suggestion
        isolation actor: isolated (any Actor)? = #isolation,
```

Sources/NIOCore/AsyncChannel/AsyncChannel.swift
```diff
@@ -284,6 +284,7 @@ public struct NIOAsyncChannel<Inbound: Sendable, Outbound: Sendable>: Sendable {
     /// - Important: After this method returned the underlying ``Channel`` will be closed.
     ///
     /// - Parameter body: A closure that gets scoped access to the inbound and outbound.
+    @_disfavoredOverload
```
FranzBusch, line 287: I think we should also deprecate those methods since in theory those are not concurrency safe and we want users to adopt a Swift 6 compiler to make those methods safe again.

Lukasa: I don't think it's appropriate to use deprecation to encourage users to update _compiler_ version. Deprecation warnings really need to be something that can be resolved without a toolchain change.

Lukasa (commented):
It seems like it might be a good idea for us to write tests for these functions. We should grab some test of executeAndClose and tweak it both a) be called from within an actor, and b) return a non-Sendable value across the boundary.

gjcairo (approved):

PR #3123, opened 2025-02-26
Expose GITHUB_ACTIONS env var to docker containers

rnro:
### Motivation:

It can be useful for scripts to be able to check if they are being run in CI. Because we invoke docker on the CI runners, the env var which people may use to do this isn't present.

### Modifications:

Expose `GITHUB_ACTIONS` env var to docker containers in the matrix workflow

### Result:

Adopters of the matrix workflow or downstream workflows will be able to check if they are running in CI

### DIFF
--- a/.github/workflows/swift_test_matrix.yml
+++ b/.github/workflows/swift_test_matrix.yml
@@ -43,6 +43,7 @@ jobs:
           workspace="/$(basename ${{ github.workspace }})"
           docker run -v ${{ github.workspace }}:"$workspace" \
             -w "$workspace" \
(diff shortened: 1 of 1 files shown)


Lukasa (approved):

simonjbeaumont:
Might be more generalisable for the scripts to check for `CI`, which is also populated by Github actions (and other CI systems usually).

This would allow our scripts to be more portable and not Github-actions–centric.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

PR #3127, opened 2025-03-05
Enable strict concurrency checking for NIOWebSocket

rnro:
Enable strict concurrency checking for NIOWebSocket

### Motivation:

To ensure NIOWebSocket concurrency safety.

### Modifications:

* `NIOWebSocketClientUpgrader._upgrade` private function generic return
type is marked as `Sendable` in line with the only method which calls
it.
* `NIOWebSocketServerUpgrader._upgrade` private function generic return
type is marked as `Sendable` in line with the only method which calls
it.
* `WebSocketOpcode.allCases` public computed variable is replaced with a
`let` which manually enumerates the cases instead

### Result:

`NIOWebSocket` does not warn of concurrency issues, builds will warn and CI will fail if regressions are introduced.

### DIFF
--- a/Sources/NIOWebSocket/WebSocketOpcode.swift
+++ b/Sources/NIOWebSocket/WebSocketOpcode.swift
@@ -57,7 +57,24 @@ extension WebSocketOpcode: Equatable {}
 extension WebSocketOpcode: Hashable {}
 
 extension WebSocketOpcode: CaseIterable {
-    public static var allCases = (0..<0x10).map { WebSocketOpcode(rawValue: $0) }
+    public static let allCases: [WebSocketOpcode] = [
+        .init(rawValue: 0x0),
+        .init(rawValue: 0x1),
+        .init(rawValue: 0x2),
+        .init(rawValue: 0x3),
+        .init(rawValue: 0x4),
+        .init(rawValue: 0x5),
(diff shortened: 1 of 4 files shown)


Sources/NIOWebSocket/WebSocketOpcode.swift
```diff
@@ -57,7 +57,24 @@ extension WebSocketOpcode: Equatable {}
 extension WebSocketOpcode: Hashable {}
 
 extension WebSocketOpcode: CaseIterable {
-    public static var allCases = (0..<0x10).map { WebSocketOpcode(rawValue: $0) }
+    public static let allCases: [WebSocketOpcode] = [
```
Lukasa, line 60: So this has the unfortunate property of being a technical API break.

Can I suggest making this a computed property that returns the array on get, and on set just does nothing?

Lukasa (changes requested):

Sources/NIOWebSocket/WebSocketOpcode.swift
rnro: 👍 sure. I'd misread this as already being a get-only computed property so I missed the break. I've also deprecated the setter.

Lukasa (approved):

PR #3128, opened 2025-03-05
NIOSingletons docs: Clarify the set once-ness

weissi:
### Motivation:

The docs don't make it clear that the thread count suggestions for the singletons are:

- set once
- privileged
- need to happen very early on to be effective

### Modifications:

Clarify that.

### Result:

Fewer surprises.


Lukasa (approved):
Fab, thanks JW

PR #3129, opened 2025-03-05
Remove side effect from assertion in IPHeaderRemoverHandler

glbrntt:
Motivation:

The 'testRawSocketBootstrap_withProtocolNegotiation' test times out when run in release mode. This is because 'IPHeaderRemoverHandler' removes the IPv4 headers as a side effect of an assertion. This means that the headers are only removed when compiled in debug mode.

Modifications:

Always remove the headers.

Result:

The test passes in release mode.


Lukasa (approved):
Good catch!

glbrntt:
Suspect CI will fail for the `unsafe` issue in nightly. I verified this locally with by temporarily renaming the `unsafe` variables.

PR #3130, opened 2025-03-06
Enable strict concurrency checking for NIOTestUtils

rnro:
### Motivation:

To ensure `NIOTestUtils` concurrency safety.

### Modifications:

* `NIOHTTP1TestServer` does pipeline configuration as synchronous operations.

### Result:

`NIOTestUtils` does not warn of concurrency issues, builds will warn and CI will fail if regressions are introduced.

### DIFF
--- a/Sources/NIOTestUtils/NIOHTTP1TestServer.swift
+++ b/Sources/NIOTestUtils/NIOHTTP1TestServer.swift
@@ -263,18 +263,16 @@ public final class NIOHTTP1TestServer {
             self.handleChannels()
             return
         }
-        channel.pipeline.configureHTTPServerPipeline().flatMap {
+        do {
+            try channel.pipeline.syncOperations.configureHTTPServerPipeline()
             if self.aggregateBody {
-                return channel.pipeline.addHandler(AggregateBodyHandler())
-            } else {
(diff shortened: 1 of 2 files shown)


Sources/NIOTestUtils/NIOHTTP1TestServer.swift
```diff
@@ -263,18 +263,16 @@ public final class NIOHTTP1TestServer {
...
-            channel.pipeline.addHandler(WebServerHandler(webServer: self))
-        }.flatMap {
-            channel.pipeline.addHandler(TransformerHandler())
-        }.whenSuccess {
-            _ = channel.setOption(.autoRead, value: true)
+            try channel.pipeline.syncOperations.addHandler(WebServerHandler(webServer: self))
+            try channel.pipeline.syncOperations.addHandler(TransformerHandler())
+            _ = try channel.syncOptions!.setOption(.autoRead, value: true)
+        } catch {
+            assertionFailure("Channel initialization failed with: \(error)")
```
Lukasa, line 275: Let's just fully `fatalError` here, no reason to be shy.

Lukasa (approved):

PR #3135, opened 2025-03-07
Fix NioAsyncWriter test on concurrency thread pool with single thread

orobio:
### Motivation:

The testSuspendingBufferedYield_whenWriterFinished test fails on the Android emulator. See also the discussion in #3044.

### Modifications:

The test requires at least two threads in the concurrency thread pool because it blocks one task, which waits for another task to set a condition. This PR adds support for running a task executor based on a NIOThreadPool and uses it for the test. Using a custom task executor guarantees that at least two threads are available for the test.

Additionally, the test has been renamed to testWriterFinish_AndSuspendBufferedYield, which is more in line with the other test names.

### Result:

The test will pass regardless of the width of the global concurrency thread pool.


orobio:
@finagolfin : Could you verify this fix on the Android emulator?

@Lukasa : I wanted to get some feedback first before writing tests for `withNIOThreadPoolTaskExecutor`.  I will start creating some tests for that function if the fix works, and if this solution is deemed acceptable.

orobio:
@finagolfin : That is good news. Thank you for verifying! I've added the dependency in Package.swift.

@Lukasa : Would this solution be acceptable? Using `NIOThreadPool` seemed like the obvious choice, and I designed it to ensure everything is cleaned up by the end of the test. However, incorrect usage could cause `withNIOThreadPoolTaskExecutor` to wait indefinitely for tasks to finish. I think guaranteed clean-up is a good thing to have for this, but let me know if you'd prefer a different approach.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,92 @@
...
+            guard case shouldRun = NIOThreadPool.WorkItemState.active else {
+                // Should never happen. We keep the thread pool running until self is deinitialized.
+                fatalError("Thread pool stopped unexpectedly")
+            }
+            unownedJob.runSynchronously(on: self.asUnownedTaskExecutor())
+        }
+    }
+
+    deinit {
+        self.deinitContinuation.finish()
```
Lukasa, line 88: We shouldn't use `deinit` for this: this kind of RAII-based cleanup has some nasty side effects when it goes wrong. A better design might be to force structured concurrency here, and say that if Tasks are running on this executor that is an immediate crash.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,92 @@
+#if compiler(>=6)
+
+import NIOPosix
+
+/// Run a `NIOThreadPool` based `TaskExecutor` while executing the given `body`.
```
Lukasa, line 5: Let's clarify in this documentation that this is a `TaskExecutor`, but not a _serial_ executor.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,92 @@
...
+///   - numberOfThreads: The number of threads in the pool.
+///   - body: The closure that will accept the task executor.
+///
+/// - Throws: When `body` throws.
+///
+/// - Returns: The value returned by `body`.
+///
+public func withNIOThreadPoolTaskExecutor<T, Failure>(
+    numberOfThreads: Int,
+    body: (any TaskExecutor) async throws(Failure) -> T
```
Lukasa, line 32: No need for an `any` here, this can be a `some`.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,92 @@
...
+///
+/// - Parameters:
+///   - numberOfThreads: The number of threads in the pool.
+///   - body: The closure that will accept the task executor.
+///
+/// - Throws: When `body` throws.
+///
+/// - Returns: The value returned by `body`.
+///
+public func withNIOThreadPoolTaskExecutor<T, Failure>(
```
Lukasa, line 30: Let's make this `@inlinable`.

Lukasa (commented):
I think in general I'd accept this for this use-case, yeah. Let's just tweak a few things.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
orobio: Just to be clear. Do you mean that just trying to run an unstructured Task on the executor should result in a crash, or that we should crash if any Task is still running at the point we want to shutdown the thread pool?

Lukasa: The latter.

orobio: Okay, clear. The reason I didn't do that in the first place was that it's non-deterministic; a task may or may not still be running when we want to shutdown, which may result in a crash only in some situations. I guess we'll need to make that very clear in the documentation, then hopefully it will be fine. I'll have a look at it.

Lukasa: Agreed, and as this is testing-only code I think it's acceptable.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
orobio: `some` isn't possible here, but I changed it to the concrete type.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
orobio: Done

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
orobio: Done

glbrntt (changes requested):

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,116 @@
...
+    /// Make sure that all tasks running on the executor are finished before shutting down.
+    ///
+    /// - warning: If any task is still running on the executor, this results in a fatalError.
+    public func shutdownGracefully() async {
+        do {
+            try await nioThreadPool.shutdownGracefully()
+        } catch {
+            fatalError("Failed to shutdown NIOThreadPool")
+        }
+    }
```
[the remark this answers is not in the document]
orobio: I kept it internal at first for that reason, but with the changes to `withNIOThreadPoolTaskExecutor`, to make it inlinable and provide the concrete type to the closure, I had the choice of making it fully public, or annotate everything with `usableFromInline`. Eventually I settled on just making it public, thinking that perhaps it could also be useful in some cases without the `withNIOThreadPoolTaskExecutor` function. I have no problem changing it back to an internal type.

Sources/NIOTestUtils/NIOThreadPoolTaskExecutor.swift
```diff
@@ -0,0 +1,116 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftNIO project authors
```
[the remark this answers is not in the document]
orobio: Ah, I forgot that one. Good catch!

glbrntt (approved):
Thank you!

Tests/NIOCoreTests/AsyncSequences/NIOAsyncWriterTests.swift
```diff
@@ -606,48 +607,57 @@ final class NIOAsyncWriterTests: XCTestCase {
...
-                suspendedAgain.unlock(withValue: true)
-            }
-        }
-
-        self.delegate.didYieldHandler = { _ in
-            if self.delegate.didYieldCallCount == 1 {
-                // Delay this yield until the other yield is suspended again.
-                suspendedAgain.lock(whenValue: true)
-                suspendedAgain.unlock()
+    func testWriterFinish_AndSuspendBufferedYield() async throws {
```
FranzBusch, line 610: Sorry for being late to the party but I just realized that this test could be completely re-written to make it work without any thread backed executors and fully reliable. What I wanted to achieve with this test originally was to control the order of the yields and their suspension. I think using a manual task executor would make this test 100 times more maintainable and remove any of the locks.

This is how a manual executor looks like:

```
import DequeModule
import Synchronization

@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
final class ManualTaskExecutor: TaskExecutor {
    private let jobs = Mutex<Deque<UnownedJob>>(.init())
    
    func enqueue(_ job: UnownedJob) {
        self.jobs.withLock { $0.append(job) }
    }
    
    func run() {
        while let job = self.jobs.withLock({ $0.popFirst() }) {
            job.runSynchronously(on: self.asUnownedTaskExecutor())
        }
    }
}
```

With such an executor we should be able to exactly control what child task runs at what point and control when they suspend.

FranzBusch: Oh sorry. I thought this was one of my original tests. Regardless I think we can achieve the ordering that we want here by manually controlling the order of execution.

orobio: @FranzBusch : I pushed an update to switch to a ManualTaskExecutor, which works nicely. Is this what you had in mind?

One note: I have no knowledge of the versions in Apple's ecosystem, so I just copied the `@available` line from your example, assuming that it is correct.

orobio: @FranzBusch : Could you have a look at this when you have a chance?

Sources/NIOTestUtils/ManualTaskExecutor.swift
FranzBusch, line 1: I don't think NIO should provide this from the test utils really. Can we at least make everything `package` for now so it isn't public?

Sources/NIOTestUtils/ManualTaskExecutor.swift
```diff
@@ -0,0 +1,164 @@
...
+    /// and makes sure that no new jobs can be enqueued.
+    @usableFromInline
+    func shutdown() {
+        self.storage.withLock { storage in
+            if !storage.jobs.isEmpty {
+                fatalError("Shutdown of manual executor with jobs in queue")
+            }
+            storage.isShutdown = true
+        }
+    }
```
FranzBusch, line 161: I don't know if we really need this complicated shutdown handling. Since task executors can only be set on a scope we shouldn't have to do any manual clean up or reject jobs.

FranzBusch (commented):
Looks great! Just two minor comments

Sources/NIOTestUtils/ManualTaskExecutor.swift
orobio: Changed to `package`

Sources/NIOTestUtils/ManualTaskExecutor.swift
orobio: It can catch some incorrect usage, like for example:
```swift
await withManualTaskExecutor { taskExecutor in
    Task(executorPreference: taskExecutor) {
        try await Task.sleep(for: .seconds(3))
    }
    taskExecutor.runUntilQueueIsEmpty()  // Runs until the sleep starts, but the task is not finished yet
}
```
It will definitely not catch all problems, but I was thinking there could be value in catching some.

FranzBusch: Since it is not a public thing anymore I'm fine with leaving it. @Lukasa @glbrntt you want to take a look before we merge it?

FranzBusch (approved):

Lukasa (approved):
Nice, this LGTM. Thanks @orobio!

Lukasa:
Looks like a few CI failures need cleaning up.

orobio:
> Looks like a few CI failures need cleaning up.

Should be fixed now, except for this one, which doesn't seem related:
https://github.com/apple/swift-nio/actions/runs/14705288968/job/41272725117?pr=3135

orobio:
@Lukasa : Would we need to do anything for the [third issue](https://github.com/apple/swift-nio/actions/runs/14705288968/job/41272725117?pr=3135)? Or can we rerun the CI?

glbrntt:
@orobio -- apologies for missing this one. I've just kicked the CI and enabled auto-merge. The last issue you linked above is a known issue which shouldn't block the merge.

orobio:
It looks like another issue popped up, but I'm not sure what to do with that one either.

glbrntt:
Looks like an infra problem, I've re-run that job.

glbrntt:
Something is up with that runner unrelated to your PR so I'm going to merge over the failure.

PR #3138, opened 2025-03-11
Allow NIOLoopBoundBox to be initialized by sending in a value

hamzahrmalik:
Allow NIOLoopBoundBox to be initialized by sending in a value
### Motivation:

It should be possible to initialize a loopbound box off-loop if the initial value of the box is being moved onto the loop.

### Modifications:

Add another static function, to create a loopboundbox by `sending` a value. This is safe because the compiler will not let the caller use the value after putting it into the box. Only the box (or rather, the el which the box is bound to) will be able to access the value thereafter.

### DIFF
--- a/Sources/NIOCore/NIOLoopBound.swift
+++ b/Sources/NIOCore/NIOLoopBound.swift
@@ -127,6 +127,28 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         .init(_value: value, uncheckedEventLoop: eventLoop)
     }
 
+    #if compiler(>=6.0)
+    /// Initialise a ``NIOLoopBoundBox`` by sending a  value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because `value` is moved into the box and can no longer be accessed outside the box.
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/NIOLoopBound.swift
```diff
@@ -127,6 +127,28 @@ public final class NIOLoopBoundBox<Value>: @unchecked Sendable {
         .init(_value: value, uncheckedEventLoop: eventLoop)
     }
 
+    #if compiler(>=6.0)
+    /// Initialise a ``NIOLoopBoundBox`` by sending a  value, validly callable off `eventLoop`.
+    ///
+    /// Contrary to ``init(_:eventLoop:)``, this method can be called off `eventLoop` because `value` is moved into the box and can no longer be accessed outside the box.
+    /// So we don't need to protect `value` itself, we just need to protect the ``NIOLoopBoundBox`` against mutations which we do because the ``value``
+    /// accessors are checking that we're on `eventLoop`.
+    public static func makeBoxTakingValue(
```
Lukasa, line 136: I think if we're going to add this API, it needs to have the same name as the one above. It's ultimately the same function, with some slightly less strong generic constraints.

Lukasa (approved):

PR #3141, opened 2025-03-12
Experimental macOS CI

rnro:
Enable macOS CI

### Motivation:

To expand test coverage to the Darwin platforms.

### Modifications:

Create a new re-usable workflow `macos_tests.yml` which offers testing and compilation checks for multiple Xcodes and simulator destinations.

The new workflow first performs `swift test` with configurable parameters and then goes on to build the code targeting generic destinations for:
* iOS
* macOS
* macOS Catalyst
* tvOS
* visionOS (temporarily disabled due to a possible runner issue)
* watchOS

At the moment this is workflow is enabled to run on the NIO repository:
* on each commit to a PR
* on each merge to `main`
* overnight on a timer

This will likely be revised due to capacity.

This functionality is implemented as a standalone workflow to preserve flexibility as requirements and capacity change over time.

### Result:

Increased test coverage.

### DIFF
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -36,6 +36,9 @@ on:
         description: "The arguments passed to swift test in the Xcode version 16.2 job."
         default: ""
 
+      build_scheme:
+        type: string
+        description: "The build scheme used in the Xcode builds."
       macos_xcode_build_enabled:
         type: boolean
         description: "Boolean to enable the Xcode build targetting macOS. Defaults to true."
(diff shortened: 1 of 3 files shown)


rnro:
I will review the test shim commit (18256679402a78c800f2d0e0fd1775dd2a35ac66) prior to merging - it's there to show the workflow in operation.

.github/workflows/macos_tests.yml
```diff
@@ -0,0 +1,157 @@
+name: macOS tests
```
FranzBusch, line 1: Why do we provide a separate workflow for this instead of inside the existing testing workflow? We can disable it by default there but in the end all of our projects want to test on macOS and it would be easier to roll it out wide without requiring a new workflow. Maybe I'm missing something though.

rnro: I tried to give a nod to this in the description but perhaps more detail is needed:
> This functionality is implemented as a standalone workflow to preserve flexibility as requirements and capacity change over time.

The pattern of use (on PRs?, run nightly?), requirements, and node capacity are likely to be in flux over the short term as more repositories are onboarded and we understand load - I wanted to reserve as much flexibility as possible. I agree in an ideal world they'd be one matrix.

rnro: The actual task(s) we're performing here is also very different, we want to run a bunch of checks on one node to avoid the recycle time, not just `swift test`.

FranzBusch: Makes sense. Thanks for expanding

FranzBusch (approved):

PR #3144, opened 2025-03-13
A re-usable CMake check workflow

rnro:
### Motivation

A few repositories have custom workflows for checking CMake, we would like to unify this for maintenance and adoption simplicity.

Further to this at the moment they only check that the scripts are up-to-date, we'd also like to have them build using CMake.

### Modifications

Add the re-usable workflow and supporting scripts

### Result:

Repositories using CMake can switch to the centralised workflow.

An example of it in action (updated) https://github.com/apple/swift-crypto/actions/runs/13837092514/job/38782284575

### DIFF
--- a/.github/workflows/cmake_tests.yml
+++ b/.github/workflows/cmake_tests.yml
@@ -41,7 +41,7 @@ jobs:
           curl -s https://raw.githubusercontent.com/apple/swift-nio/main/scripts/update-cmake-lists.sh | CONFIG_JSON="${{ inputs.update_cmake_lists_config }}" FAIL_ON_CHANGES=true bash
       - name: CMake build
         run: |
-          which curl cmake ninja || apt -q update 
+          which curl cmake ninja || apt -q update
           which curl || apt -yq install curl
           which cmake || apt -yq install cmake
           which ninja || apt -yq install ninja-build
--- a/scripts/cmake-build.sh
+++ b/scripts/cmake-build.sh
@@ -25,11 +25,11 @@ if [ -z "$target_dir" ]; then
   fatal "Target directory must be specified."
 fi
 
-CURL_BIN=${CURL_BIN:-$(which curl)} || fatal "CURL_BIN unset and no curl on PATH"
-TAR_BIN=${TAR_BIN:-$(which tar)} || fatal "TAR_BIN unset and no tar on PATH"
-CMAKE_BIN=${CMAKE_BIN:-$(which cmake)} || fatal "CMAKE_BIN unset and no cmake on PATH"
-NINJA_BIN=${NINJA_BIN:-$(which ninja)} || fatal "NINJA_BIN unset and no ninja on PATH"
-ASSEMBLY_COMPILER_BIN=${ASSEMBLY_COMPILER_BIN:-$(which clang)} || fatal "ASSEMBLY_COMPILER_BIN unset and no clang on PATH"
+CURL_BIN="${CURL_BIN:-$(which curl)}" || fatal "CURL_BIN unset and no curl on PATH"
+TAR_BIN="${TAR_BIN:-$(which tar)}" || fatal "TAR_BIN unset and no tar on PATH"
+CMAKE_BIN="${CMAKE_BIN:-$(which cmake)}" || fatal "CMAKE_BIN unset and no cmake on PATH"
+NINJA_BIN="${NINJA_BIN:-$(which ninja)}" || fatal "NINJA_BIN unset and no ninja on PATH"
+ASSEMBLY_COMPILER_BIN="${ASSEMBLY_COMPILER_BIN:-$(which clang)}" || fatal "ASSEMBLY_COMPILER_BIN unset and no clang on PATH"
 
 log "Building Ninja build files for target"
 build_dir="${target_dir}/build"


scripts/cmake-build.sh
```diff
@@ -0,0 +1,53 @@
...
+log "Installing tools for this script"
+apt-get update -y -q && apt-get install -y -q curl rsync
+
+log "Obtaining CMake (${CMAKE_VERSION})"
+curl -sL "https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-linux-x86_64.tar.gz" -o /tmp/cmake.tar.gz
+
+log "Installing CMake"
+mkdir -p /tmp/cmake
+tar xfvz /tmp/cmake.tar.gz --strip-components 1 -C /tmp/cmake
+rsync -q -a /tmp/cmake/* /usr/
```
simonjbeaumont, line 41: This seems a bit of an unnecessary indirection. If we're extracting to a temporary and immediately shovelling everything into `/usr/`, can't we just extract there?

scripts/cmake-build.sh
```diff
@@ -0,0 +1,53 @@
...
+fi
+if [ -z "$target_dir" ]; then
+  fatal "Target directory must be specified."
+fi
+
+log "Installing tools for this script"
+apt-get update -y -q && apt-get install -y -q curl rsync
+
+log "Obtaining CMake (${CMAKE_VERSION})"
+curl -sL "https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSION}/cmake-${CMAKE_VERSION}-linux-x86_64.tar.gz" -o /tmp/cmake.tar.gz
```
simonjbeaumont, line 36: Is there a specific need to pin to the CMake version? If not, we could just get this from `apt` like the other packages?

scripts/cmake-build.sh
```diff
@@ -0,0 +1,53 @@
...
+
+log "Installing CMake"
+mkdir -p /tmp/cmake
+tar xfvz /tmp/cmake.tar.gz --strip-components 1 -C /tmp/cmake
+rsync -q -a /tmp/cmake/* /usr/
+
+log "Installing Ninja"
+apt-get install -y -q ninja-build
```
simonjbeaumont, line 44: May as well move this up to the other install?

scripts/cmake-build.sh
```diff
@@ -0,0 +1,53 @@
...
+
+if [ -z "$version" ]; then
+  fatal "CMake version must be specified."
+fi
+if [ -z "$target_dir" ]; then
+  fatal "Target directory must be specified."
+fi
+
+log "Installing tools for this script"
+apt-get update -y -q && apt-get install -y -q curl rsync
```
simonjbeaumont, line 33: All your `apt-get` commands can just be `apt`.

scripts/cmake-build.sh
```diff
@@ -0,0 +1,53 @@
...
+
+log "Installing CMake"
+mkdir -p /tmp/cmake
+tar xfvz /tmp/cmake.tar.gz --strip-components 1 -C /tmp/cmake
+rsync -q -a /tmp/cmake/* /usr/
+
+log "Installing Ninja"
+apt-get install -y -q ninja-build
+
+log "Building makefiles for target"
```
simonjbeaumont, line 46: This comment seems wrong if we're using the Ninja CMake generator (not the Makefile one).

scripts/update-cmake-lists.sh
simonjbeaumont, line 1: Are these files covered by the shellcheck CI?

.github/workflows/cmake_tests.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+          submodules: true
+      - name: Mark the workspace as safe
+        # https://github.com/actions/checkout/issues/766
+        run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
+      - name: Check CMake lists
```
simonjbeaumont, line 32: This irks me. These are referred to as "CMakeLists files" in the docs: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Writing%20CMakeLists%20Files.html

scripts/cmake-build.sh
simonjbeaumont: Separately, why do we have installation of tools here _and_ in the GitHub Actions YAML?

Can it be in one place. Is it more portable for it to be in the YAML, where the distro is also declared, since then this script would work on other distros, which use different package managers.

scripts/update-cmake-lists.sh
simonjbeaumont: So, this file can likely land as-is, since it's been working fine in Swift Crypto. But if you wanted to rework it to be more like the others we have in the workflows you could. E.g. the delicate use of `find` can probably be replaced with a `git-ls-files` construction similar to what we have in the other checks. The `exceptions` could be handled with an ignore file, etc.

WDYT? Not bother ever? Defer? Do now?

.github/workflows/cmake_tests.yml
rnro: So you'd rather this said the following?
```suggestion
      - name: Check CMakeLists files
```

scripts/update-cmake-lists.sh
rnro: > Are these files covered by the shellcheck CI?
Yes, it's complained at me about parts of it 😂

rnro: I think the ignore files idea is a good one for us to put down as future work, then if any adopter repo has a specific need they can pick it up (or if anyone has free time).

scripts/cmake-build.sh
rnro: I've tried to have installation of tools just before we use them so that the components are modular. I've changed this line to also check if it's already installed with a `which foo` check to try and make it more lightweight.

scripts/cmake-build.sh
rnro: I'd prefer keeping them separate so that we can have separate checks using `which` to see if they are already installed (especially since we know in CI `curl` will be)

scripts/cmake-build.sh
rnro: I don't think the pinning is necessary from our side, but it doesn't seem to be available in `apt` by default nor is there an 'official' repository I could add to pull it from that I could see.

rnro: Huh - strange, I'm not sure why my attempt to find it failed. I'll use this one

scripts/cmake-build.sh
rnro: I've switched the new build script to this approach, it agree it's good to have it portable

Issue #3139, opened 2025-03-11
`isOutboundHalfClosureEnabled` closes channel before all writes are flushed


adam-fowler:
### Expected behavior
If I create a `NIOAsyncChannel` with `isOutboundHalfClosureEnabled` set to true. When I call `outbound.finish()` I expect that all writes will have been flushed before the channel is closed.

This was meant to be a workaround for the fact that `NIOAsyncChannel.executeThenClose` didn't flush writes, but it has the same issue.

### Steps to reproduce

1. Write small HTTP server that outputs a large response
2. Use curl to get response.
3. curl will report with something like `curl: (18) transfer closed with 672832 bytes remaining to read`

### If possible, minimal yet complete reproducer code (or URL to code)

Here is a small HTTP server that returns 1000000 bytes in its response

```swift
import NIOCore
import NIOHTTP1
import NIOPosix

/// Sendable server response that doesn't use ``IOData``
public typealias SendableHTTPServerResponsePart = HTTPPart<HTTPResponseHead, ByteBuffer>

/// Channel to convert HTTPServerResponsePart to the Sendable type HBHTTPServerResponsePart
final class HTTPSendableResponseChannelHandler: ChannelOutboundHandler, RemovableChannelHandler {
    typealias OutboundIn = SendableHTTPServerResponsePart
    typealias OutboundOut = HTTPServerResponsePart

    func write(context: ChannelHandlerContext, data: NIOAny, promise: EventLoopPromise<Void>?) {
        let part = unwrapOutboundIn(data)
        switch part {
        case .head(let head):
            context.write(self.wrapOutboundOut(.head(head)), promise: promise)
        case .body(let buffer):
            context.write(self.wrapOutboundOut(.body(.byteBuffer(buffer))), promise: promise)
        case .end:
            context.writeAndFlush(self.wrapOutboundOut(.end(nil)), promise: promise)
        }
    }
}

@available(macOS 14, *)
func server() async throws {
    let asyncChannel = try await ServerBootstrap(group: MultiThreadedEventLoopGroup.singleton)
        // Specify backlog and enable SO_REUSEADDR for the server itself
        .serverChannelOption(ChannelOptions.socketOption(.so_reuseaddr), value: 1)
        .childChannelOption(ChannelOptions.allowRemoteHalfClosure, value: true)
        .bind(host: "127.0.0.1", port: 8888, serverBackPressureStrategy: nil) { channel in
            return channel.eventLoop.makeCompletedFuture {
[...]

PR #3148, opened 2025-03-18
BaseStreamSocketChannel half-close allows outstanding writes to complete

rnro:
### Motivation:

At the moment half-closes are actioned immediately and fails all outstanding writes. We should refuse new writes but allow these writes to complete before completing the close.

### Modifications:

Modify the PendingWritesManager internal buffer to hold an enum of either writes or close events. We use this to store the close and only action it when the preceding writes have been handled.

### Result:

Outbound close should no longer fail outstanding writes

Should resolve https://github.com/apple/swift-nio/issues/3139

### DIFF
--- a/Sources/NIOPosix/BaseStreamSocketChannel.swift
+++ b/Sources/NIOPosix/BaseStreamSocketChannel.swift
@@ -194,27 +194,33 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                     self.close0(error: error, mode: .all, promise: promise)
                     return
                 }
-                self.unregisterForWritable()
 
-                let writesCloseResult = self.pendingWrites.closeOutbound(promise)
-                switch writesCloseResult {
+                let result = self.pendingWrites.closeOutbound(promise)
+                switch result {
                 case .pending:
-                    ()  // promise is stored in `pendingWrites` state for completing on later call to `closeComplete`
+                    ()  // promise is stored in `pendingWrites` state for completing later
+
                 case .readyForClose(let closePromise):
                     // Shutdown the socket only when the pending writes are dealt with
                     do {
                         try self.shutdownSocket(mode: mode)
-                        self.pendingWrites.closeComplete()
+                        closePromise?.succeed(())
                     } catch let err {
-                        self.pendingWrites.closeComplete(err)
+                        closePromise?.fail(err)
                     }
-                    closePromise?.succeed(())
+                    self.unregisterForWritable()
                     self.pipeline.fireUserInboundEventTriggered(ChannelEvent.outputClosed)
+
                 case .closed(let closePromise):
                     closePromise?.succeed(())
-                case .open:
-                    promise?.fail(ChannelError.inappropriateOperationForState)
-                    assertionFailure("Close resulted in an open state, this should never happen")
+
+                case .errored(let err, let closePromise):
+                    assertionFailure("Close errored: \(err)")
+                    closePromise?.fail(err)
+
+                    // Escalate to full closure
+                    // promise is nil here because we have used the supplied promise to convey failure of the half-close
+                    self.close0(error: err, mode: .all, promise: nil)
                 }
 
             case .input:
@@ -264,16 +270,8 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
     }
 
     final override func cancelWritesOnClose(error: Error) {
-        let closeResult = self.pendingWrites.failAll(error: error, close: true)
-        switch closeResult {
-        case .closed(let eventLoopPromise):
-            if let eventLoopPromise {
-                eventLoopPromise.fail(error)
-            }
-        case .open, .pending, .readyForClose:
-            preconditionFailure("failAll with close should never return \(closeResult!)")
-        case .none:
-            preconditionFailure("failAll with close should return a close result")
+        if let eventLoopPromise = self.pendingWrites.failAll(error: error) {
+            eventLoopPromise.fail(error)
         }
     }
 
--- a/Sources/NIOPosix/BaseStreamSocketChannel.swift
+++ b/Sources/NIOPosix/BaseStreamSocketChannel.swift
@@ -196,11 +196,11 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                 }
                 self.unregisterForWritable()
 
-                let writesCloseResult = self.pendingWrites.close(promise)
+                let writesCloseResult = self.pendingWrites.closeOutbound(promise)
                 switch writesCloseResult {
                 case .pending:
                     ()  // promise is stored in `pendingWrites` state for completing on later call to `closeComplete`
-                case .readyForClose:
+                case .readyForClose(let closePromise):
                     // Shutdown the socket only when the pending writes are dealt with
                     do {
                         try self.shutdownSocket(mode: mode)
@@ -208,9 +208,10 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                     } catch let err {
                         self.pendingWrites.closeComplete(err)
                     }
+                    closePromise?.succeed(())
                     self.pipeline.fireUserInboundEventTriggered(ChannelEvent.outputClosed)
-                case .closed:
-                    promise?.succeed(())
+                case .closed(let closePromise):
+                    closePromise?.succeed(())
                 case .open:
                     promise?.fail(ChannelError.inappropriateOperationForState)
                     assertionFailure("Close resulted in an open state, this should never happen")
@@ -263,7 +264,17 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
     }
 
     final override func cancelWritesOnClose(error: Error) {
-        self.pendingWrites.failAll(error: error, close: true)
+        let closeResult = self.pendingWrites.failAll(error: error, close: true)
+        switch closeResult {
+        case .closed(let eventLoopPromise):
+            if let eventLoopPromise {
+                eventLoopPromise.fail(error)
+            }
+        case .open, .pending, .readyForClose:
+            preconditionFailure("failAll with close should never return \(closeResult!)")
+        case .none:
+            preconditionFailure("failAll with close should return a close result")
+        }
     }
 
     @discardableResult
--- a/Sources/NIOPosix/BaseStreamSocketChannel.swift
+++ b/Sources/NIOPosix/BaseStreamSocketChannel.swift
@@ -201,7 +201,7 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                 case .open:
                     preconditionFailure("Close resulted in an open state, this should never happen")
                 case .pending:
-                    () // nothing to do
+                    ()  // nothing to do
                 case .readyForClose(let closePromise):
                     assert(promise == closePromise)
                     // Shutdown the socket only when the pending writes are dealt with
@@ -213,7 +213,7 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                     }
                     self.pendingWrites.outboundCloseState = .closed
                 case .closed:
-                    () // nothing to do
+                    ()  // nothing to do
                 }
 
                 self.pipeline.fireUserInboundEventTriggered(ChannelEvent.outputClosed)
--- a/Sources/NIOPosix/BaseStreamSocketChannel.swift
+++ b/Sources/NIOPosix/BaseStreamSocketChannel.swift
@@ -194,11 +194,19 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
                     self.close0(error: error, mode: .all, promise: promise)
                     return
                 }
-                try self.shutdownSocket(mode: mode)
-                // Fail all pending writes and so ensure all pending promises are notified
-                self.pendingWrites.failAll(error: error, close: false)
                 self.unregisterForWritable()
-                promise?.succeed(())
+
+                // Shutdown the socket only when the pending writes are dealt with
+                let closePromise = self.eventLoop.makePromise(of: Void.self)
+                closePromise.futureResult.whenComplete { _ in
+                    do {
+                        try self.shutdownSocket(mode: mode)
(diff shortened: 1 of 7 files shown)


Sources/NIOPosix/BaseStreamSocketChannel.swift
```diff
@@ -194,11 +194,19 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
...
+                // Shutdown the socket only when the pending writes are dealt with
+                let closePromise = self.eventLoop.makePromise(of: Void.self)
+                closePromise.futureResult.whenComplete { _ in
+                    do {
+                        try self.shutdownSocket(mode: mode)
+                        promise?.succeed(())
+                    } catch let err {
+                        promise?.fail(err)
+                    }
+                }
```
Lukasa, line 208: I don't think we should need a promise here: we can manage a state machine directly, without resorting to it.

Sources/NIOPosix/BaseStreamSocketChannel.swift
```diff
@@ -194,11 +194,27 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
...
+                    ()  // nothing to do
+                case .readyForClose(let closePromise):
+                    assert(promise == closePromise)
+                    // Shutdown the socket only when the pending writes are dealt with
+                    do {
+                        try self.shutdownSocket(mode: mode)
+                        closePromise?.succeed(())
+                    } catch let err {
+                        closePromise?.fail(err)
+                    }
```
Lukasa, line 213: Can we factor the outbound close code out to a common path? This doesn't do everything the other path does, so we should merge the two codepaths to make sure we're self-consistent.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -380,6 +393,28 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
+                case .pending(let eventLoopPromise):
+                    self.outboundCloseState = .readyForClose(eventLoopPromise)
+                case .readyForClose:
+                    assertionFailure("Transitioned from readyForClose to readyForClose, shouldn't we have closed?")
+                case .closed:
+                    ()
+                }
+
+            case .couldNotWriteEverything:
+                ()
```
Lukasa, line 413: This one also feels like it is worth an `assertionFailure`.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,11 +498,34 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
+    ///   - promise: Optionally an `EventLoopPromise` that will be succeeded once all outstanding writes have been dealt with
+    func close(_ promise: EventLoopPromise<Void>?) -> CloseState {
+        assert(self.isOpen)
+
+        if self.isEmpty {
+            switch self.outboundCloseState {
+            case .open:
+                self.outboundCloseState = .readyForClose(promise)
+            case .readyForClose:
+                ()
```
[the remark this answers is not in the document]
rnro: I've changed the API of this object so that the promise is never surfaced after it has been passed in. The promise must be completed by a call to `closeComplete.

Sources/NIOPosix/BaseStreamSocketChannel.swift
rnro: I _think_ I've done this.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -97,12 +97,23 @@ internal enum OneWriteOperationResult {
...
+        static func == (lhs: Self, rhs: Self) -> Bool {
+            switch (lhs, rhs) {
+            case (.writtenCompletely, .writtenCompletely):
+                return true
+            case (.couldNotWriteEverything, .couldNotWriteEverything):
+                return true
+            default:
+                return false
+            }
+        }
```
glbrntt, line 116: I get a bit nervous about ignoring state when comparing. Do we _actually_ need the equality check? Also, it might make more sense to do this on the `CloseState` and compare the identity of its underlying futures where applicable.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,16 +498,82 @@ final class PendingStreamWritesManager: PendingWritesManager {
         return self.didWrite(itemCount: result.itemCount, result: result.writeResult)
     }
 
-    /// Fail all the outstanding writes. This is useful if for example the `Channel` is closed.
+    /// Fail all the outstanding writes.
     func failAll(error: Error, close: Bool) {
         if close {
             assert(self.isOpen)
             self.isOpen = false
+            self.state.removeAll()?.fail(error)
```
glbrntt, line 506: General rule of thumb: update your state before making outcalls. Failing the promise here might lead to calling back into this type again and we haven't fully reconciled our state yet. We should grab the promise and fail it afterwards.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,16 +498,82 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
+            assert(self.state.isEmpty)
+            switch self.outboundCloseState {
+            case .open, .closed:
+                self.outboundCloseState = .closed
+            case .pending(let closePromise), .readyForClose(let closePromise):
+                closePromise?.fail(error)
+                self.outboundCloseState = .closed
```
glbrntt, line 513: Same here, update the state then complete the promise

Sources/NIOPosix/PendingWritesManager.swift
glbrntt, line 506: Not sure if it's just the diff but it looks like this function also failed all the promises if `!close`, that doesn't seem to happen anymore.

rnro: Thanks for the general guidance :)

Sources/NIOPosix/PendingWritesManager.swift
rnro: It turns out we don't need it anymore so I've just deleted it.

rnro: 🤦‍♂️ it's needed in tests. Looking again.

rnro: I think this has exposed that I was exposing the promise in a way which wasn't needed anyway because we just passed it right into `pendingWrites.close` which already held it. I've changed the method to return the `CloseResult` instead.

glbrntt (approved):
This LGTM now but please wait for Cory's review as well.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,18 +487,85 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
+
+        if self.isEmpty {
+            switch self.outboundCloseState {
+            case .open:
+                self.outboundCloseState = .readyForClose(promise)
+            case .readyForClose(var closePromise), .pending(var closePromise):
+                closePromise.setOrCascade(to: promise)
+                self.outboundCloseState = .readyForClose(closePromise)
+            case .closed:
+                promise?.succeed(())
```
Lukasa, line 524: As noted above, it's probably better to let the caller complete the promise.

Lukasa (commented):
Cool, this is looking closer, I think I just have some suggestions around promise hygiene and naming.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,18 +487,85 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
     func failAll(error: Error, close: Bool) {
         if close {
             assert(self.isOpen)
             self.isOpen = false
+            switch self.outboundCloseState {
+            case .open, .closed:
+                self.outboundCloseState = .closed
+            case .pending(let closePromise), .readyForClose(let closePromise):
+                self.outboundCloseState = .closed
+                closePromise?.fail(error)
```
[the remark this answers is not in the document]
rnro: Can you explain the thinking here? My understanding from our previous discussions was that it would be better that once the higher-level caller handed-off the promise it was the responsibility of the `PendingWritesManager` to complete the promise, so passing it in was a one-way street. This is why I separated-out `CloseResult` as its own type, so we have clarity on where the promise is exposed and not exposed. 

Are you proposing that this method returns `EventLoopPromise<Void>?` where the promise is non-nil if the caller needs to do something with it?

Sources/NIOPosix/PendingWritesManager.swift
Lukasa: Or put the promise in the close result as associated data.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,16 +494,93 @@ final class PendingStreamWritesManager: PendingWritesManager {
         return self.didWrite(itemCount: result.itemCount, result: result.writeResult)
     }
 
-    /// Fail all the outstanding writes. This is useful if for example the `Channel` is closed.
-    func failAll(error: Error, close: Bool) {
+    /// Fail all the outstanding writes.
+    func failAll(error: Error, close: Bool) -> CloseResult? {
```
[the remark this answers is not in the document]
rnro: https://github.com/apple/swift-nio/pull/3221

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -463,16 +494,93 @@ final class PendingStreamWritesManager: PendingWritesManager {
...
+            case .pending(var closePromise):
+                closePromise.setOrCascade(to: promise)
+                self.outboundCloseState = .pending(closePromise)
+                closeResult = .pending
+            case .readyForClose:
+                preconditionFailure(
+                    "We are in .readyForClose state but we still have pending writes. This should never happen."
+                )
+            case .closed:
+                preconditionFailure(
```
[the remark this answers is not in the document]
rnro: I've done some broader restructuring to try and make this type of thing easier to reason through

Sources/NIOPosix/PendingWritesManager.swift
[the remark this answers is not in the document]
rnro: Disregard the above PR. With my changes we will always close on a call to this method so I've removed the boolean.

Sources/NIOPosix/BaseStreamSocketChannel.swift
```diff
@@ -194,13 +194,29 @@ class BaseStreamSocketChannel<Socket: SocketProtocol>: BaseSocketChannel<Socket>
...
                     return
                 }
-                try self.shutdownSocket(mode: mode)
-                // Fail all pending writes and so ensure all pending promises are notified
-                self.pendingWrites.failAll(error: error, close: false)
                 self.unregisterForWritable()
-                promise?.succeed(())
 
-                self.pipeline.fireUserInboundEventTriggered(ChannelEvent.outputClosed)
+                let writesCloseResult = self.pendingWrites.closeOutbound(promise)
```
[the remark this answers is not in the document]
rnro: I've created another test for this case which wedged until I resolved this issue.

Sources/NIOPosix/PendingWritesManager.swift
```diff
@@ -97,12 +106,34 @@ internal enum OneWriteOperationResult {
...
         /// Could not write everything. Before attempting further writes the eventing system should send a 'writable' notification.
         case couldNotWriteEverything
+
+        /// The resulting status of a `PendingWritesManager` after a completely-written write
+        ///
+        /// This type is subtly different to `CloseState` so that it only surfaces the close promise when the caller
+        /// is expected to fulfill it
+        internal enum WrittenCompletelyResult: Equatable {
+            case open
+            case pending
```
Lukasa, line 122: This pending state seems unreachable.

Lukasa (commented):
Ok, the product code here is looking a _lot_ better. A small nit.

Tests/NIOPosixTests/StreamChannelsTest.swift
```diff
@@ -261,6 +261,126 @@ class StreamChannelTest: XCTestCase {
...
+            XCTAssertNoThrow(try chan2.syncCloseAcceptingAlreadyClosed())
+        }
+        XCTAssertNoThrow(try forEachCrossConnectedStreamChannelPair(runTest))
+    }
+
+    func testHalfCloseOwnOutputWithWritabilityChange() throws {
+        final class BytesReadCountingHandler: ChannelInboundHandler, Sendable {
+            typealias InboundIn = ByteBuffer
+
+            private let numBytes: NIOLoopBoundBox<Int>
```
Lukasa, line 308: No need to overcomplicate this, you can just use a `NIOLockedValueBox`.

Tests/NIOPosixTests/StreamChannelsTest.swift
```diff
@@ -261,6 +261,126 @@ class StreamChannelTest: XCTestCase {
...
+            }
+        }
+
+        final class BytesWrittenCountingHandler: ChannelInboundHandler, Sendable {
+            typealias InboundIn = ByteBuffer
+
+            public typealias OutboundIn = ByteBuffer
+            public typealias OutboundOut = ByteBuffer
+
+            private let numBytes: NIOLoopBoundBox<Int>
```
Lukasa, line 327: Same here, just use `NIOLockedValueBox`.

Tests/NIOPosixTests/StreamChannelsTest.swift
```diff
@@ -261,6 +261,126 @@ class StreamChannelTest: XCTestCase {
...
+            try chan1.setOption(.autoRead, value: true).wait()
+
+            // wait for the reading-side close to complete
+            try chan1.closeFuture.wait()
+
+            // now the dust has settled all the bytes should be accounted for
+            try chan2.eventLoop.submit {
+                XCTAssertNotEqual(bytesWritten.value, 0)
+                XCTAssertEqual(bytesRead.value, bytesWritten.value)
+            }.wait()
```
Lukasa, line 379: Let's extend these tests to confirm we do receive inputClosed as well, and at the right time.

Lukasa: At the point of receiving inputClosed we should record how many bytes we saw in an optional. We can then add an assertion at the end of the test that this value is equal to the total bytes sent. This asserts that both a) we saw the message (value isn't `nil`) and b) that we saw it at the right time.

rnro: Added, thank you.

Lukasa (approved):
Very nice, let's ship it.

PR #3150, opened 2025-03-19
Update macOS CI workflow

rnro:
### Motivation:

We are still exploring the shape of macOS CI and exploring the system's capabilities.

### Modifications:

* Add visionOS jobs
* Remove Xcode 15.4 as it is no longer on the runners
* Add support for running tests targetting other platforms
* Enable test targeting other platforms on daily jobs to increase coverage without bloating PR runtimes.

###  Result:

More macOS CI coverage

CI run with this enabled https://github.com/apple/swift-nio/actions/runs/13944231147/job/39027647470?pr=3150 - the iOS failure is unrelated to this change

### DIFF
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -5,11 +5,11 @@ on:
     inputs:
       xcode_15_4_enabled:
         type: boolean
-        description: "Boolean to enable the Xcode version 15.4 jobs. Defaults to true."
-        default: true
+        description: "⚠️ Jobs with this version of Xcode are deprecated."
+        default: false
       xcode_15_4_test_arguments_override:
         type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/macos_tests.yml
```diff
@@ -5,11 +5,11 @@ on:
     inputs:
```
czechboy0, line 5: Probably for a separate PR, but should we rename from `macos_tests` to, for example `xcode_tests` or `apple_tests`?

czechboy0 (approved):

.github/workflows/macos_tests.yml
rnro: Yeah I'm not sure - I hovered on `darwin_tests` too. I'll ask around.

PR #3152, opened 2025-03-20
Add make*Future methods to isolated event loop

glbrntt:
Motivation:

Creating a completed future of a non-sendable value when on the current event loop is currently a little tedious. It requires you to create a promise, succeed it using the isolated view, or fail it using the non-isolated view. This is more tedious than it needs to be.

Modifications:

- Add `make{Succeeded,Failed,Completed}Future` methods to the isolated event loop. These don't require the `Success` value to be `Sendable`.

Result:

It's easier to create completed futures of non sendable values from an isolated event loop.

### Motivation:

### Modifications:

### Result:


Lukasa (approved):
Nice one, thanks @glbrntt!

Lukasa (approved):

PR #3156, opened 2025-03-24
fix: Fix NIOCore build for wasm targets

scottmarchant:
Fixes NIOCore compiler errors for the `wasm32` architecture.

### Motivation:

When compiling NIOCore for a SwiftWasm target, several compilation errors such as the following occur:

```
> swift build --swift-sdk wasm32-unknown-wasi --target NIOCore

.../swift-nio/Sources/NIOCore/FileHandle.swift:151:42: error: cannot find type 'TwoUInt32s' in scope
149 | public final class NIOFileHandle: FileDescriptor & Sendable {
150 |     private static let descriptorClosed: CInt = CInt.min
151 |     private let descriptor: UnsafeAtomic<TwoUInt32s>
    |                                          `- error: cannot find type 'TwoUInt32s' in scope
152 | 
153 |     public var isOpen: Bool {
.../swift-nio/Sources/NIOCore/FileHandle.swift:77:8: error: Unknown architecture
 75 | typealias TwoUInt32s = DoubleWord
 76 | #else
 77 | #error("Unknown architecture")
    |        `- error: Unknown architecture
 78 | #endif
 79 | 
 
…
```
The compilation errors were introduced with the following change: 

https://github.com/apple/swift-nio/pull/2598/files#diff-a041d0c0d50a8f4fcda2020eeb4450ea994920ef61653cc06014457ccdb3ba0cR70

### Modifications:

The fix is straight-forward and unlikely to affect any existing compilation targets other than the `wasm32` target.

We simply need to add `arch(wasm32)` to compiler architecture condtions for a new typedef on line, then NIOCore compiles to wasm32 again.

### Result:

Before: NIOCore wasm build fails ❌ 

After: NIOCore wasm build succeeds ✅

### DIFF
--- a/Sources/NIOCore/FileHandle.swift
+++ b/Sources/NIOCore/FileHandle.swift
@@ -67,7 +67,7 @@ extension UInt64 {
         }
     }
 }
-#elseif arch(arm) || arch(i386) || arch(arm64_32)
+#elseif arch(arm) || arch(i386) || arch(arm64_32) || arch(wasm32)
 // 32 bit architectures
 // Note: for testing purposes you can also use these defines for 64 bit platforms, they'll just consume twice as
 // much space, nothing else will go bad.


weissi:
Amazing, thanks so much and sorry for missing wasm!

kateinoigakukun (approved):

Sources/NIOCore/FileHandle.swift
```diff
@@ -67,7 +67,7 @@ extension UInt64 {
         }
     }
 }
-#elseif arch(arm) || arch(i386) || arch(arm64_32)
```
kateinoigakukun, line 70: If this only matters the pointer size, `_pointerBitWidth(_64)` and `_pointerBitWidth(_32)` might be better choice

MaxDesiatov: +1, otherwise it misses plenty of other 32-bit architectures that in theory could be supported in the future.

weissi: That's up to the NIO folks but
- `_pointerBitWidth*` is underscored and we have non-underscored alternatives (but yes, they're annoying)
- mixed platforms like `arm64_32` need to fall into the correct place, which in this case would be true (assuming that `_pointerBitWidth(_32)` is true for `arm64_32`

glbrntt: My gut feeling is to avoid the underscored APIs. If it becomes necessary to support more 32-bit architectures in the future then we should push to get a non-underscored API.

scottmarchant:
> Amazing, thanks so much and sorry for missing wasm!

@weissi No worries at all - pretty hard to catch them all. I meant to mention, I was impressed with that nice `#error("Unknown architecture")` fail mode. Pointed me straight to the right spot and helped me figure out what to do.

Sources/NIOCore/FileHandle.swift
scottmarchant: Sounds like the consensus is stick with ` || arch(wasm32)`. I'll leave as is unless someone voice opinions otherwise. Happy to change this if desired. I tried out adding ` || _pointerBitWidth(_32)` instead of ` || arch(wasm32)`, and that also worked, just as a data point.

weissi: Friendly ping @glbrntt to approve & merge

glbrntt (approved):
Thank you!

kateinoigakukun:
FWIW I just opened a PR to set up CI for Wasm SDK https://github.com/apple/swift-nio/pull/3159

Issue #2736, opened 2024-06-07
We should offer the same `[get/set]MultipleIntegers` as we offer for `[read/write]MultipleIntegers` methods on ByteBuffer


FranzBusch:
Normally we provide all four variants of methods like `get/set/read/write` on the various `ByteBuffer` operations but we seem to missing the `get/set` variants for the `multipleInterger` based methods.

Lukasa:
I think @weissi might make the case that this is a feature not a bug, I know he's been unhappy with the get/set methods for a while. I'm curious what he thinks.

weissi:
> I think @weissi might make the case that this is a feature not a bug, I know he's been unhappy with the get/set methods for a while. I'm curious what he thinks.

I hate get/set's prominence, yes. 99.9% of their uses are either verbose or incorrect, usually both. Particularly

```
// correct: Something(buffer: buffer)
//
// incorrect and often seen in the wild
buffer.getSomething(at: 0, length: buffer.readableBytes)

// correct but verbose, seldomly seen
buffer.getSomething(at: buffer.readerIndex, length: buffer.readableBytes)
```

is both incorrect & verbose.

---

Now, there _are_ some niche use cases for our `get*/`set*`s but most people just use them because `read*/`write*` are mutating. So I think instead of adding `get*`/`set*`s eagerly we should offer `peek*` which doesn't take an `at: Int` parameter. So `peek*` ~= `read*` except that it doesn't read it off.

---

My thinking is, let's invest time in

- Better docs, every single `get*` should be discouraged in its docs
- `peek*` methods for cases where people need to take a part out without reading it out
- if we still have time left: add the missing `get*`/`set*`s with discouraging comments

FranzBusch:
My use-case is the classic `peek` case where I need to look into an incoming packet to decide the kind of it and the send into the right child channel.

weissi:
> My use-case is the classic `peek` case where I need to look into an incoming packet to decide the kind of it and the send into the right child channel.

Makes sense, I think we should add `peek*`s for everything. That's btw also a good starter task because it really just is

```
@inlinable
public func peekX(length: Int) -> X? {
    return self.getX(at: self.readerIndex, length: length)
}
```

or

```
@inlinable
public func peekX(length: Int) -> X? {
    var `self` = self
    return self.readX(length: length)
}
```

FranzBusch:
Yeah I am totally fine rebranding the current `get` methods to peek and avoid doing public `get/set` all together.

[...]

Issue #2034, opened 2022-01-24
Deprecate and rename ByteBuffer.get*/set*


weissi:
`ByteBuffer`'s `get*`/`set*` methods have a very niche use-case. In almost all cases, the users should use `read*`/`write*` or something like `readableBytesView`.

The key issues are stuff like `buffer.getData(at: 0, length: 16)`. This will often work (because many ByteBuffers happen to have `readerIndex == 0`) but it won't always work.

Unfortunately the `get*` APIs are very prominent and new users are pretty much invited to use them because the `read*` APIs require a mutable `ByteBuffer`. So if you (which is common!) have an immutable one (say a function argument), then Xcode won't even complete the `read*` versions. So  of course users use `get*` and then write wrong code.

I'm proposing a set of changes:

- deprecate `get*`/`set*`
- introduce new, scary sounding (long names) methods that do the same thing as the existing `get*`/`set*` ones
- introduce a new family of APIs (maybe `peek*`?) that work like `read*` (ie. from `readerIndex`) but do _not_ move the `readerIndex`

That would mean. All existing uses of `buffer.get*(at: 0, length: ...)` can without other changes be made `buffer.peek*(length: ...)`. And the few uses of `buffer.get*(at: nonZeroArgument)` can use the specialised, scary sounding new methods.

vlm:
+1

Lukasa:
While I think these changes are reasonable, I'm not sure if they merit the churn of a deprecate/replace cycle. In particular, the fact that we cannot unilaterally recommend a single transformation for the `get*` methods means that each call site needs to be audited by hand for the appropriate replacement.

A quick search across the core NIO repositories (swift-nio, swift-nio-ssl, swift-nio-http2, swift-nio-extras, swift-nio-transport-services, swift-nio-ssh) reveals 530 call sites for these methods. Each of these call sites will need to be manually audited and rewritten to use the appropriate invocation.

A related note is that it's not uncommon to _declare_ methods like this as well, so this change has to propagate through the ecosystem. This forces some more churn there too: I count 35 such methods in the core NIO repos listed above.

None of this is the end of the world, but I think it does prompt the question of whether the downside of these APIs justifies the cost of migrating them. I don't have good numbers for how this affects repos outside the core NIO repo, so it's hard to be confident of what the total ecosystem churn is. The result is that I'm on the fence: this change produces a lot of work, but I'm not sure how much it helps.

Some other info:

[...]

PR #3157, opened 2025-03-25
Add peekInteger API to ByteBuffer

RaghavRoy145:
### Motivation:

The existing getInteger API is often used incorrectly, leading to verbose or error-prone code as mentioned in https://github.com/apple/swift-nio/issues/2034 and https://github.com/apple/swift-nio/issues/2736. Users may unintentionally access data without correctly managing the reader index. To address this, I added a peekInteger API that reads integers using the current readerIndex without modifying it.

### Modifications:

* Added a peekInteger method to ByteBuffer, which provides a nonmutating way to read integers using the current readerIndex.
* Implemented peekInteger using the existing getInteger function.
* Added comprehensive unit tests to validate correct behavior, including scenarios for different integer types, insufficient bytes, and multi-write scenarios.

### Result:

* Users now have access to a safer API for inspecting integers in a buffer without modifying state.
* Added tests for integer reading scenarios.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -114,9 +114,9 @@ extension ByteBuffer {
         }
     }
 
-    /// Peeks at the integer at the current reader index without advancing it.
+    /// Returns the integer at the current reader index without advancing it.
     ///
-    /// This method is equivalent to calling `getInteger(at: readerIndex, ...)
+    /// This method is equivalent to calling `getInteger(at: readerIndex, ...)`
     ///
     /// - Parameters:
     ///   - endianness: The endianness of the integer (defaults to big endian).
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -126,7 +126,7 @@ extension ByteBuffer {
     public func peekInteger<T: FixedWidthInteger>(
         endianness: Endianness = .big,
         as: T.Type = T.self
-    ) -> T?  {
+    ) -> T? {
         self.getInteger(at: self.readerIndex, endianness: endianness, as: `as`)
     }
 }
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -126,11 +126,8 @@ extension ByteBuffer {
     public func peekInteger<T: FixedWidthInteger>(
         endianness: Endianness = .big,
         as: T.Type = T.self
-    ) -> T? {
-        guard let result = self.getInteger(at: self.readerIndex, endianness: endianness, as: T.self) else {
-            return nil
-        }
-        return result
+    ) -> T?  {
+        self.getInteger(at: self.readerIndex, endianness: endianness, as: `as`)
     }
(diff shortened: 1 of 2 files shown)


Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -113,6 +113,22 @@ extension ByteBuffer {
             self.setBytes(ptr, at: index)
         }
     }
+
+    /// Peeks at the integer at the current reader index without advancing it.
+    ///
+    /// This method is equivalent to calling `getInteger(at: readerIndex, ...)
```
glbrntt, line 119:
```suggestion
    /// This method is equivalent to calling `getInteger(at: readerIndex, ...)`
```

Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -113,6 +113,22 @@ extension ByteBuffer {
             self.setBytes(ptr, at: index)
         }
     }
+
+    /// Peeks at the integer at the current reader index without advancing it.
```
glbrntt, line 117:
```suggestion
    /// Returns the integer at the current reader index without advancing it.
```

glbrntt (commented):
Couple of doc nits but otherwise this looks great, thank you!

Sources/NIOCore/ByteBuffer-int.swift
RaghavRoy145: Just to echo my comment on one of the issues referred to in the description. Since this kind of a change can be extrapolated to every type that has a get* API, can the scope of this PR increase to consider them or should they separate PRs?
Thanks!

glbrntt: Given that this PR is effectively good to go I'd bias towards merging this and opening a separate PR for other methods.

glbrntt (approved):

Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -113,6 +113,25 @@ extension ByteBuffer {
...
+    /// - Returns: An integer value deserialized from this `ByteBuffer` or `nil` if the bytes are not readable.
+    @inlinable
+    public func peekInteger<T: FixedWidthInteger>(
+        endianness: Endianness = .big,
+        as: T.Type = T.self
+    ) -> T? {
+        guard let result = self.getInteger(at: self.readerIndex, endianness: endianness, as: T.self) else {
+            return nil
+        }
+        return result
```
glbrntt, line 133: We don't need the `guard` here. The formatter was complaining because of the explicit return where it wasn't necessary:

```suggestion
        self.getInteger(at: self.readerIndex, endianness: endianness, as: T.self)
```

RaghavRoy145:
I'm sorry, I need to be more careful with the format checks

glbrntt:
> I'm sorry, I need to be more careful with the format checks. 

No worries! For what it's worth you should be able to run `swift format --parallel --recursive --in-place Sources Tests` to run the formatter locally (assuming you have a Swift 6 toolchain installed). 

> However, there's some unit tests that fail, are those supposed to be taken care of as well?

The failing unit tests are an issue with a nightly build of the compiler, so nothing more you need to do here.

PR #3160, opened 2025-03-29
Add Peek API variants for ByteBuffer

RaghavRoy145:
### Motivation:
Existing get APIs require passing an explicit index and can be misused, leading to verbose and error-prone code. Adding peek variants that automatically use the current readerIndex improves safety and clarity. This aims to address https://github.com/apple/swift-nio/issues/2034 and https://github.com/apple/swift-nio/issues/2736, and is a continuation of PR https://github.com/apple/swift-nio/pull/3157

### Modifications:
- Introduced peekBytes(), peekString(), peekNullTerminatedString(), peekUTF8ValidatedString() and peekDispatchData().
- Added tests for each peek API covering normal, empty, repeated, and partial read scenarios.

### Result:
Developers can now use nonmutating peek APIs to inspect ByteBuffer contents without altering the reader index.

***PS**: I have focussed on the ByteBuffer-aux get\* implementations, so I'm not sure if I am missing any other important/error-prone get\* elsewhere, if so, please point me to them and I'll be happy to add respective peek\*'s!*

### DIFF
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -999,5 +999,18 @@ extension ByteBuffer {
         /// The length of the bytes to copy was negative.
         public static let invalidUTF8: ReadUTF8ValidationError = .init(baseError: .invalidUTF8)
     }
+
+    /// Return a UTF-8 validated String decoded from the bytes at the current reader index.
+    ///
+    /// This is equivalent to calling `getUTF8ValidatedString(at: readerIndex, length: ...)` and does not advance the reader index.
+    ///
+    /// - Parameter length: The number of bytes making up the string.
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
This looks great, thanks @RaghavRoy145!

RaghavRoy145:
> This looks great, thanks @RaghavRoy145!

Thanks for the review! Do let me know if there's get methods that I may have missed. I know there's getData and getSlice for example, but if these are not part of the original plan, I can start working on Multiple Integer instead.

glbrntt:
`getData` and `getSlice` are absolutely worth doing as well!

RaghavRoy145:
> `getData` and `getSlice` are absolutely worth doing as well!

Great! Perhaps its better to create a new PR with the ones I find that are defined outside ByteBuffer-aux.

PR #3162, opened 2025-03-31
Separate build and test steps in `macos_tests.yml`

rnro:
Separate build and test steps in `macos_tests.yml`. Test execution can now also be disabled through an input. This is useful for clarity of output and also in some cases where test execution may not be supported yet.

### DIFF
--- a/.github/workflows/macos_tests.yml
+++ b/.github/workflows/macos_tests.yml
@@ -39,6 +39,10 @@ on:
       build_scheme:
         type: string
         description: "The build scheme used in the Xcode builds."
+      swift_test_enabled:
+        type: boolean
+        description: "Boolean to enable test execution with `swift test`. Defaults to true."
+        default: true
       macos_xcode_build_enabled:
         type: boolean
         description: "Boolean to enable the Xcode build targeting macOS. Defaults to true."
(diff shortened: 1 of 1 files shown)


.github/workflows/macos_tests.yml
```diff
@@ -153,7 +157,11 @@ jobs:
         with:
           persist-credentials: false
           submodules: true
+      - name: Swift build
+        run: |
+          swift build --build-tests
```
glbrntt, line 162: Should there be build argument overrides as well?

.github/workflows/macos_tests.yml
```diff
@@ -96,48 +100,56 @@ jobs:
           cat >> "$GITHUB_OUTPUT" << EOM
           darwin-matrix=$(
             xcode_15_4_enabled="${MATRIX_MACOS_15_4_ENABLED:=true}"
+            xcode_15_4_build_arguments_override="${MATRIX_MACOS_15_4_BUILD_ARGUMENTS_OVERRIDE:=""}"
```
glbrntt, line 103: Do these need to be listed in the `inputs`?

.github/workflows/macos_tests.yml
```diff
@@ -153,7 +165,15 @@ jobs:
         with:
           persist-credentials: false
           submodules: true
+      - name: Swift build
+        run: |
+          if [ -n "${{ matrix.config.build_arguments_override }}" ]; then
+            swift build "${{ matrix.config.build_arguments_override }}"
```
glbrntt, line 171: Just checking that `--build-tests` is purposefully dropped here because the args are being overridden?

rnro: Yeah I thought giving people power to totally customize was more general.

glbrntt (approved):

PR #3167, opened 2025-04-01
Strict concurrency for NIOPerformanceTester and NIOCrashTester

glbrntt:


rnro (approved):

Sources/NIOPerformanceTester/LockBenchmark.swift
```diff
@@ -17,7 +17,9 @@ import NIOConcurrencyHelpers
 import NIOCore
 import NIOPosix
 
-final class NIOLockBenchmark: Benchmark {
+final class NIOLockBenchmark: Benchmark, @unchecked Sendable {
+    // mutable state is protected by the lock
```
gjcairo, line 21: Nit but couldn't we wrap `opsDone` in a `NIOLockedValueBox` instead and remove the `@unchecked`?

Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -114,14 +116,20 @@ final class TCPThroughputBenchmark: Benchmark {
...
         let connectionEstablishedPromise: EventLoopPromise<EventLoop> = self.group.next().makePromise()
 
+        let promise = self.group.next().makePromise(of: NIOLoopBound<ServerHandler>.self)
         self.serverChannel = try ServerBootstrap(group: self.group)
             .childChannelInitializer { channel in
-                self.serverHandler = ServerHandler(connectionEstablishedPromise)
-                return channel.pipeline.addHandler(self.serverHandler)
+                channel.eventLoop.makeCompletedFuture {
+                    let serverHandler = ServerHandler(connectionEstablishedPromise, eventLoop: channel.eventLoop)
+                    promise.succeed(NIOLoopBound(serverHandler, eventLoop: channel.eventLoop))
```
gjcairo, line 124: Do we want this promise succeeded if adding the handler to the pipeline failed?

Sources/NIOPerformanceTester/UDPBenchmark.swift
```diff
@@ -87,7 +90,7 @@ extension UDPBenchmark: Benchmark {
 }
 
 extension UDPBenchmark {
-    final class EchoHandler: ChannelInboundHandler {
+    final class EchoHandler: ChannelInboundHandler, Sendable {
```
gjcairo, line 93: Do we want this to be `Sendable` if we have a `SendableView`?

Sources/NIOPerformanceTester/LockBenchmark.swift
glbrntt: From a `Sendable`-cleanness perspective, yes. However, this is a benchmark for `NIOLock`, so using `NIOLockedValueBox` would somewhat defeat the point of the benchmark!

Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
glbrntt: If adding the handler fails then the bootstrap will `throw` anyway so it doesn't really matter.

Sources/NIOPerformanceTester/UDPBenchmark.swift
glbrntt: The `SendableView` is on a different type.

gjcairo (approved):

PR #3171, opened 2025-04-02
Add support for Xcode 16.3

rnro:
### Motivation:

Xcode 16.3 has been released, we should cover it in CI

### Modifications:

* Add Xcode 16.3 runs by default
* Disable Xcode 16.0 runs by default (soft deprecation)

Note that the Xcode version offering is still a work in progress.

### Result:

macOS CI will test using Xcode 16.3

An example run with this working https://github.com/apple/swift-nio/actions/runs/14220724296/job/39876502770?pr=3171


fabianfett (approved):
LGTM

PR #3174, opened 2025-04-03
Add More Peek API Variants for ByteBuffer

RaghavRoy145:
### Motivation:

Existing get APIs require passing an explicit index and can be misused, leading to verbose and error-prone code. Adding peek variants that automatically use the current readerIndex improves safety and clarity. This aims to address issue #2034 and issue #2736, and is a continuation of PR #3157

### Modifications:

    Introduced peekSlice(), peekLengthPrefixedSlice(), peekData(), peekUUIDBytes() and peekWebSocketErrorCode().
    Added tests for each peek API covering normal, empty and repeated peek scenarios.

### Result:

Developers can now use nonmutating peek APIs to inspect ByteBuffer contents without altering the reader index.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-lengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-lengthPrefix.swift
@@ -165,15 +165,6 @@ extension ByteBuffer {
         endianness: Endianness = .big,
         as integer: Integer.Type
     ) -> ByteBuffer? where Integer: FixedWidthInteger {
-        guard let lengthPrefix = self.getInteger(at: self.readerIndex, endianness: endianness, as: Integer.self),
-            let messageLength = Int(exactly: lengthPrefix),
-            let messageBuffer = self.getSlice(
-                at: self.readerIndex + MemoryLayout<Integer>.size,
-                length: messageLength
-            )
-        else {
-            return nil
-        }
-        return messageBuffer
+        self.getLengthPrefixedSlice(at: self.readerIndex, endianness: endianness, as: integer)
     }
 }
--- a/Sources/NIOCore/ByteBuffer-lengthPrefix.swift
+++ b/Sources/NIOCore/ByteBuffer-lengthPrefix.swift
@@ -149,4 +149,31 @@ extension ByteBuffer {
 
         return messageBuffer
     }
+
+    /// Reads an integer length prefix at the current `readerIndex`, then returns a `ByteBuffer` slice of that length
+    /// without advancing the reader index.
+    ///
+    /// This method is equivalent to calling `getLengthPrefixedSlice(at: readerIndex, ...)`.
+    ///
+    /// - Parameters:
+    ///   - endianness: The endianness of the length prefix (defaults to big endian).
+    ///   - integer: The `FixedWidthInteger` type for the length prefix.
+    /// - Returns: `nil` if the length prefix could not be read, the length prefix is negative, or if there aren't enough
+    ///            bytes for the message after the prefix. Otherwise, a slice of the requested length.
+    @inlinable
+    public func peekLengthPrefixedSlice<Integer>(
+        endianness: Endianness = .big,
+        as integer: Integer.Type
+    ) -> ByteBuffer? where Integer: FixedWidthInteger {
+        guard let lengthPrefix = self.getInteger(at: self.readerIndex, endianness: endianness, as: Integer.self),
+            let messageLength = Int(exactly: lengthPrefix),
+            let messageBuffer = self.getSlice(
+                at: self.readerIndex + MemoryLayout<Integer>.size,
+                length: messageLength
+            )
+        else {
+            return nil
+        }
+        return messageBuffer
+    }
 }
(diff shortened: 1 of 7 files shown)


RaghavRoy145:
Just adding some more peek*s for some more get*s I found, I'm not sure if any more are remaining apart from multiple integers. 

Weirdly, I was unable to get the peekWebSocketErrorCode() Api tests to run, and then I realised there are no tests written for any of WebSocketErrorCode APIs? I could be wrong, but because of that I have skipped adding tests for that.

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
```diff
@@ -149,4 +149,31 @@ extension ByteBuffer {
...
+        guard let lengthPrefix = self.getInteger(at: self.readerIndex, endianness: endianness, as: Integer.self),
+            let messageLength = Int(exactly: lengthPrefix),
+            let messageBuffer = self.getSlice(
+                at: self.readerIndex + MemoryLayout<Integer>.size,
+                length: messageLength
+            )
+        else {
+            return nil
+        }
+        return messageBuffer
```
gjcairo, line 177: Can't we just call `getLengthPrefixedSlice` here?

RaghavRoy145:
> Thanks @RaghavRoy145! Left one comment.
> 
> About the `WebSocketErrorCode` yeah, that's a strange oversight. We should get that fixed up. You could try adding tests just for the peek method if you want, but otherwise we can tackle adding tests for all the related APIs separately.

Got it, however, when I try to write tests for the peek API, it's unable to find the definition of my implementation, and I'm guessing that's because of how imports are handled in this file. I didn't want to add any additional dependencies here so I skipped it seeing how none of the other WebSocketErrorCode APIs are tested, can you help me out with this if its needed?

_swift-nio/Tests/NIOCoreTests/ByteBufferTest.swift:3868:27: error: value of type 'ByteBuffer' has no member 'peekWebSocketErrorCode'_

RaghavRoy145:
> Thanks @RaghavRoy145! Left one comment.
> 
> About the `WebSocketErrorCode` yeah, that's a strange oversight. We should get that fixed up. You could try adding tests just for the peek method if you want, but otherwise we can tackle adding tests for all the related APIs separately.

Got it, however, when I try to write tests for the peek API, it's unable to find the definition of my implementation, and I'm guessing that's because of how imports are handled in this file. I didn't want to add any additional dependencies here so I skipped it seeing how none of the other WebSocketErrorCode APIs are tested, can you help me out with this if its needed?

swift-nio/Tests/NIOCoreTests/ByteBufferTest.swift:3868:27: error: value of type 'ByteBuffer' has no member 'peekWebSocketErrorCode'

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
RaghavRoy145: Yeah I thought about that, but I went with this approach just because of how `getLengthPrefixedSlice` is implemented, it's doing multiple things. It reads an integer prefix, then reads a slice relative to that prefix, fails on some inputs (negative or out-of-bounds), etc. So it didnt seem as safe as say implementing a peek variant of getUUID which is predictable and doesn't have side-effects, so I could just create a wrapper for it.

If however it's more reasonable to simply call `getLengthPrefixedSlice`, I can change it to that. I can see how the current version is just duplicate code that needs to be separately maintained (unless ofcourse get*s will eventually be phased out anyway!)

Lukasa:
I recommend putting those tests into NIOWebSocketTests, as you need the import of NIOWebSocket.

RaghavRoy145:
> I recommend putting those tests into NIOWebSocketTests, as you need the import of NIOWebSocket.

Sure, since there doesn't exist a file meant for ByteBuffer tests for WebSocketAPIs, I'll create one in the directory. This can then be used to add the tests for the other APIs as well that don't exist yet

Sources/NIOCore/ByteBuffer-lengthPrefix.swift
Lukasa: I think this should probably just call `getLengthPrefixedSlice`.

gjcairo: Cory beat me to it but yes I think it's better to just call `getLengthPrefixedSlice`. Copying the code doesn't really make e.g. the input validation any more obvious; and it makes maintainability potentially harder as we now have the same code in multiple places as you rightly pointed out.

gjcairo (approved):

gjcairo (approved):
Thank you @RaghavRoy145 !

RaghavRoy145:
> Thank you @RaghavRoy145 !

Thanks for the review! I _think_ these should be all of the peek\*s left to implement, but if there's more get\*s lurking around that need wrapping up, happy to help

Lukasa (approved):

Lukasa:
Thanks a bunch @RaghavRoy145!

Issue #3180, opened 2025-04-07
Build error on Android: call to `signal` yields "error: function is unused"


marcprux:
https://github.com/apple/swift-nio/pull/3167 introduced a failure in my hourly Android builds of swift-nio (see https://github.com/swift-everywhere/swift-package-builds/actions/runs/14312839641/job/40111885760), which is a stopgap until https://github.com/swiftlang/github-workflows/pull/106 gets approved.

```
/Users/runner/work/swift-package-builds/swift-package-builds/swift-nio/Sources/NIOCrashTester/main.swift:140:5: error: function is unused
138 |         try! group.syncShutdownGracefully()
139 |     }
140 |     signal(SIGPIPE, SIG_IGN)
    |     `- error: function is unused
```

@finagolfin, have you ever seen anything like this?

marcprux:
I guess because signal is [defined](https://android.googlesource.com/platform/bionic/+/master/libc/include/signal.h#71) as:

```
sighandler_t _Nonnull signal(int __signal, sighandler_t _Nullable __handler);
```

and the return value is being treated as a function (which we can presumably ignore), then the solution might be simply:

```
    #if os(Android)
    _ = signal(SIGPIPE, SIG_IGN)
    #else
    signal(SIGPIPE, SIG_IGN)
    #endif
```

finagolfin:
No, never, it is related to the new strict concurrency checking, which was [just enabled for this executable in the pull you linked](https://github.com/apple/swift-nio/pull/3167/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR440), that [just broke my Android CI too](https://github.com/finagolfin/swift-android-sdk/actions/runs/14325649245/job/40150553436#step:19:638) and is reproducible natively in Termux. Your linked fix works for me when I just tried it natively on Android.

Why do you think this only hits on Android: is it that these are C macros in linux's `signal.h` so such type-checking is abandoned there, but is only being performed on Android's more simple C declarations?

PR #3181, opened 2025-04-07
Ignore return value of signal() on Android

marcprux:
Fixes https://github.com/apple/swift-nio/issues/3180 by ignoring return value of signal() on Android

### DIFF
--- a/Sources/NIOCrashTester/main.swift
+++ b/Sources/NIOCrashTester/main.swift
@@ -137,7 +137,13 @@ func main() throws {
     defer {
         try! group.syncShutdownGracefully()
     }
+    #if os(Android)
+    // needed because Android returns the previous signal handler function
+    // https://github.com/apple/swift-nio/issues/3180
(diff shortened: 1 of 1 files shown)


finagolfin:
My Android CI just hit this too, and this pull fixes it when I tried it natively on Android. I don't know why simply defining the type and a temp variable for the output fixes this error, but it does.

glbrntt (approved):
Thanks!

marcprux:
> I don't know why simply defining the type and a temp variable for the output fixes this error, but it does.

I'm guessing that nullability annotations are making the return value non-nil for Android, but nullable on Darwin.

```
func f0() -> (() -> ())? {
    { }
}

f0() // warning: Result of call to 'f0()' is unused

func f1() -> (() -> ()) {
    { }
}

f1() // error: Function is unused
_ = f1() // fine
```

glbrntt:
@marcprux could you rebase your branch on main? GitHub isn't giving me the option for some reason...

PR #3190, opened 2025-04-08
Strict concurrency for NIOPosix SelectorTest

glbrntt:


Tests/NIOPosixTests/SelectorTest.swift
```diff
@@ -121,8 +122,8 @@ class SelectorTest: XCTestCase {
     private static let testWeDoNotDeliverEventsForPreviouslyClosedChannels_numberOfChannelsToUse = 10
     func testWeDoNotDeliverEventsForPreviouslyClosedChannels() {
         /// We use this class to box mutable values, generally in this test anything boxed should only be read/written
-        /// on the event loop `el`.
-        class Box<T> {
+        /// on the event loop `el`..
```
gjcairo, line 125:
```suggestion
        /// on the event loop `el`.
```

gjcairo (approved):

Tests/NIOPosixTests/SelectorTest.swift
```diff
@@ -122,7 +123,7 @@ class SelectorTest: XCTestCase {
     func testWeDoNotDeliverEventsForPreviouslyClosedChannels() {
         /// We use this class to box mutable values, generally in this test anything boxed should only be read/written
         /// on the event loop `el`.
-        class Box<T> {
+        final class Box<T>: @unchecked Sendable {
```
Lukasa, line 126: This sendable annotation is definitely a lie. It should probably have been replaced by something like NIOLoopBoundBox or NIOLockedValueBox.

glbrntt: I'll fix this up

PR #3200, opened 2025-04-14
Enable Swift 6.1 jobs in CI

rnro:
Motivation:

Swift 6.1 has been released, we should add it to our CI coverage.

Modifications:

Add additional Swift 6.1 jobs where appropriate in main.yml, pull_request.yml

Result:

Improved test coverage.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,7 @@ jobs:
       linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
+      linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_nightly_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
+      linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
```
gjcairo, line 19: Looks like we already had 6.1 here

.github/workflows/main.yml
```diff
@@ -16,6 +16,7 @@ jobs:
...
       linux_nightly_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
```
gjcairo, line 20: We should probably have nightly 6.2 in all our repos now, no?

gjcairo (changes requested):

.github/workflows/main.yml
rnro: Sorry about this, I really thought I omitted this repo from my script

.github/workflows/main.yml
rnro: This should say `linux_nightly_next_arguments_override` - I'll fix it up separately

PR #3205, opened 2025-04-15
Strict concurrency for SAL tests

glbrntt:
Motivation:

The syscall abstraction layer tests aren't sendable clean.

Modifications:

- Factor out the setup/teardown of SAL tests into a SALContext
- SALContext a method to run a SAL test with two closues: one on the hooked event loop and another on the calling thread.
  - The EL bound closure is provided with the EL and locked boxes for the user ot kernel and kernel to user boxes.
  - The closure on the calling thread is provided with an 'assertions' object which provides a view over the hooked selector.
- Note that there are still sendable holes in this code.
- Refactored the tests to use these APIs.

Results:

Safer tests.

### DIFF
--- a/Tests/NIOPosixTests/SALChannelTests.swift
+++ b/Tests/NIOPosixTests/SALChannelTests.swift
@@ -18,62 +18,55 @@ import XCTest
 
 @testable import NIOPosix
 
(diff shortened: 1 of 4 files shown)


Lukasa (approved):
Very nice work, this seems like a very meaningful improvement.

PR #3210, opened 2025-04-15
Add errorCaught to NIOTypedHTTPClientUpgradeHandler

adam-fowler:
To handle errors fired by previous ChannelHandlers. eg NIOSSLClientHandler

### Motivation:

Otherwise if a ChannelHandler previous to the upgrade ChannelHandler fails the error returned is `inappropriateOperationForState` and the error fired by the original ChannelHandler is lost.

Added `testReturnsErrorsFromPriorChannelHandlers`. Also had to reorganise `setUpClientChannel` slightly so I could get the result of the upgrade.

### DIFF
--- a/Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
@@ -193,6 +193,7 @@ public final class NIOTypedHTTPClientUpgradeHandler<UpgradeResult: Sendable>: Ch
 
     public func errorCaught(context: ChannelHandlerContext, error: any Error) {
         self.upgradeResultPromise.fail(error)
+        context.fireErrorCaught(error)
     }
 
     private func channelRead(context: ChannelHandlerContext, responsePart: HTTPClientResponsePart) {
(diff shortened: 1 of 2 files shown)


Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
```diff
@@ -191,6 +191,10 @@ public final class NIOTypedHTTPClientUpgradeHandler<UpgradeResult: Sendable>: Ch
         }
     }
 
+    public func errorCaught(context: ChannelHandlerContext, error: any Error) {
```
Lukasa, line 194: We should probably fire the error on as well, to avoid breaking any systems that were relying on this.

Lukasa (commented):
One minor nit, otherwise this looks really good.

Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
adam-fowler: done

adam-fowler:
> Looks like we have some test failures here.

Ah yeah forwarding the error means `writeInbound` throws an error.

PR #3212, opened 2025-04-16
Clarify test server error code

Lukasa:
Motivation

The NIOHTTP1TestServer had a weird codepath where we could hit errors configuring the channel pipeline, but we weren't sure why. I've investigated, and this happens because a Channel that had been accepted earlier was closed while we were waiting to start handling it. This is totally benign: the code works just fine in that situation.

To be clear, though, let's document the behaviour and add a test.

Modifications

- Code comment to explain the behaviour
- Added a regression test

Result

More confidence.

### DIFF
--- a/Tests/NIOTestUtilsTests/NIOHTTP1TestServerTest.swift
+++ b/Tests/NIOTestUtilsTests/NIOHTTP1TestServerTest.swift
@@ -474,7 +474,7 @@ class NIOHTTP1TestServerTest: XCTestCase {
             }
         )
 
-        // Create a second channel now, and again send a request head. We can't wait for the test server to receive it, because it hasn't yet.
+        // Create a second channel now.
         let secondResponsePromise = self.group.next().makePromise(of: String.self)
         let secondChannel = try self.connect(serverPort: testServer.serverPort, responsePromise: secondResponsePromise)
             .wait()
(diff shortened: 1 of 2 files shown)


Tests/NIOTestUtilsTests/NIOHTTP1TestServerTest.swift
```diff
@@ -458,6 +458,55 @@ class NIOHTTP1TestServerTest: XCTestCase {
...
+        // Send a request head and wait for it to be sent, and received at the test server so we know the connection is well underway.
+        let requestHead = HTTPRequestHead(version: .http1_1, method: .POST, uri: "/uri")
+        firstChannel.writeAndFlush(SendableRequestPart.head(requestHead), promise: nil)
+        XCTAssertNoThrow(
+            try testServer.receiveHeadAndVerify { head in
+                XCTAssertEqual(head.uri, "/uri")
+            }
+        )
+
+        // Create a second channel now, and again send a request head. We can't wait for the test server to receive it, because it hasn't yet.
```
gjcairo, line 477: We're not sending a request head here, are we?

Lukasa: Good catch, no.

Lukasa: Handled.

PR #3216, opened 2025-04-18
Update LLHTTP and apply patch to support C++ interop

### LINKED ISSUE #3093: Cxx interop problem
### Expected behavior

Using `swift-nio` package in a project with Cxx interop enabled builds, runs, and is debuggable.

### Actual behavior

Using `swift-nio` package in a project with Cxx interop enabled builds and runs but renders debugger unusable with following error:

```
(lldb) po someVar
error: /Users/some.user/Library/Developer/Xcode/DerivedData/SomeProject-ejsisrcmvvxjfmgjdseiwglvudot/SourcePackages/checkouts/swift-nio/Sources/CNIOLLHTTP/include/c_nio_llhttp.h:3:1: import of C++ module 'std_stdint_h' appears within extern "C" language linkage specification
#include <stdint.h>
^
```

Locally modifying `c_nio_llhttp.h` as follows builds, runs, and is debuggable:

```diff
13a14
> 
15c16,18
< extern "C" {
---
> #include <cstdint>
> #else
> #include <stdint.h>
18c21,23
< #include <stdint.h>
---
> #ifdef __cplusplus
> extern "C" {
> #endif
```

### Steps to reproduce

I have attempted a standalone repro without success so far. The project in which I've enabled Cxx interop is a sizeable legacy project that is a mix of `UIKit` and `SwiftUI`.

At the project level I have added `-cxx-interoperability-mode=default -Xcc -std=c++17` to the `OTHER_SWIFT_FLAGS` build setting. I have one local package that depends on `grpc-swift` which transitively depends on `swift-nio` and this local package does not enable Cxx interop. I have another local package that depends on a remote package that requires Cxx interop and is using `swiftSettings: [.interoperabilityMode(.Cxx)]` on the target definition and `cxxLanguageStandard: .cxx17` on the package definition.

No minimal standalone repro yet.

`2.78.0`

Please provide at the very least your operating system and Swift version
information.

macOS: `14.7.1`
Swift: `5.9`
Xcode: `16.2 (16C5032a)`

Ideally, run

```
scripts/nio-diagnose -o nio-diagnose.md PID_OF_YOUR_NIO_PROGRAM
```

and attach (or paste) the resulting file `nio-diagnose.md` into this bug report
or send it to the SwiftNIO maintainers privately.

Lukasa:
### Motivation

C++ interop support in Xcode has started to dislike the header files from llhttp. While a bug has been filed upstream at https://github.com/nodejs/llparse/issues/71 it doesn't seem as though there is any movement on that fix upstream. As a result, we need to apply a patch.

To keep the code maintainable, we'd like to apply that patch in an automated fashion, so we should do so by amending the "update and patch" script. Along the way, we should update LLHTTP to the latest code.

### Modifications

The modifications are split into separate commits.

1. Modify the script to fix some bugs that have crept in.
2. Run the script to update LLHTTP to latest, without the patch.
3. Fix up the tests to account for the fact that we get some different errors from LLHTTP now.
4. Apply the patch suggested by @postmechanical in #3093
5. Re-run the script, which only applies the patch

### Result

C++ interop should be working again. Resolves #3093.


adam-fowler:
@Lukasa did you remove the workaround implemented for #2274 given there was a fix upstream in LLHTTP

Lukasa:
No I didn't. Good catch though, we should remove that workaround. I think it can be a separate patch though.

glbrntt (approved):

PR #3217, opened 2025-04-18
Fix ByteBufferTests when compiling with Swift 6 but running on macOS 14.

allevato:
Writing `@available(someOS...)` on an XCTest test method only affects whether the compiler allows known-available APIs to be called, but it doesn't actually prevent the test from running on older OSes (XCTest doesn't see the annotation since it's not part of the Objective-C runtime metadata). So if this code is compiled with Swift 6 but run on an older OS, it will segfault when it tries to call an API that doesn't exist on that platform.

These tests need to use an `#available` guard instead to make the decision to skip them at runtime.


Lukasa (approved):
Thanks Tony, much appreciated!

PR #3219, opened 2025-04-22
feat: Support non-Sendable types in `EventLoopFuture.get`

CrownedPhoenix:
Enable using `EventLoopFuture.get()` for non-Sendable `Value` on compilers that support `sending` keyword.

### Motivation:
I am using NIO to prepare non-Sendable reference types. The EventLoop does not maintain ownership of these types after they are prepared and returned to the caller - so there's no reason that they should have to be Sendable.

### Modifications:
Gate use of the `sending` keyword (instead of Sendable) behind an appropriate if-def.

### Result:

Swift compilers >=6 will be able to use `EventLoopFuture.get()` on non-Sendable Value types.

### DIFF
--- a/Sources/NIOCore/AsyncAwaitSupport.swift
+++ b/Sources/NIOCore/AsyncAwaitSupport.swift
@@ -11,7 +11,31 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
-
+#if compiler(>=6)
+extension EventLoopFuture {
+    /// Get the value/error from an `EventLoopFuture` in an `async` context.
+    ///
+    /// - warning: This method currently violates Structured Concurrency because cancellation isn't respected.
+    ///
+    /// This function can be used to bridge an `EventLoopFuture` into the `async` world. Ie. if you're in an `async`
+    /// function and want to get the result of this future.
+    @available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)
+    @preconcurrency
+    @inlinable
+    public func get() async throws -> sending Value {
+        try await withUnsafeThrowingContinuation { (cont: UnsafeContinuation<Value, Error>) in
+            self.whenComplete { result in
+                switch result {
+                    case .success(let value):
+                        cont.resume(returning: value)
+                    case .failure(let error):
+                        cont.resume(throwing: error)
+                }
+            }
+        }
+    }
+}
+#else
 extension EventLoopFuture {
     /// Get the value/error from an `EventLoopFuture` in an `async` context.
     ///
@@ -35,6 +59,7 @@ extension EventLoopFuture {
         }.wrappedValue
     }
 }
+#endif
 
 #if canImport(Dispatch)
 extension EventLoopGroup {
--- a/Sources/NIOCore/EventLoopFuture.swift
+++ b/Sources/NIOCore/EventLoopFuture.swift
@@ -869,7 +869,21 @@ extension EventLoopFuture {
             return CallbackList()
         }
     }
-
+    #if compiler(>=6)
+    /// Adds an observer callback to this `EventLoopFuture` that is called when the
+    /// `EventLoopFuture` has any result.
+    ///
+    /// - Parameters:
+    ///   - callback: The callback that is called when the `EventLoopFuture` is fulfilled.
+    @inlinable
+    @preconcurrency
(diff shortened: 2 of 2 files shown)


Lukasa (changes requested):
Thanks for this patch! However, we can’t accept it: this change is not sound. 

EventLoopFuture cannot return the value `sending` from `get()` because ELF is still holding onto that value. This means it hasn’t met the criteria for `sending`: the value is not leaving the EventLoop, it’s still there. 

As an example of where this could cause trouble, you could pass an ELF through several isolation domains, repeatedly calling `get()`, and end up with the value in all of those domains. That’s not sound, I’m afraid.

CrownedPhoenix:
Thank you for your clarification.
This was the bit of code that I missed:
```swift

// The future is unchecked Sendable following the below isolation rules this is safe
//
// 1. Receiving the value of the future is always done on the EventLoop of the future, hence
// the value is never transferred out of the event loops isolation domain. It only gets transferred
// by certain methods such as `hop()` and those methods are annotated with requiring the Value to be
// Sendable
// 2. The promise is `Sendable` but fulfilling the promise with a value requires the user to
// transfer the value to the promise. This ensures that the value is now isolated to the event loops
// isolation domain. Note: Sendable values can always be transferred

extension EventLoopPromise: Sendable {}

extension EventLoopFuture: @unchecked Sendable {}
```

Am I to understand that NIO and its EventLoopFutures are expected to be inoperable with anything except values that are @Sendable (the obvious answer is 'yes' given that I've already seen the inclusion of `where Value: Sendable` being used in many places)?

Lukasa:
> Am I to understand that NIO and its EventLoopFutures are expected to be inoperable with anything except values that are https://github.com/sendable (the obvious answer is 'yes' given that I've already seen the inclusion of where Value: Sendable being used in many places)?

No, quite the opposite: NIO and EventLoopFutures are highly operable with non-Sendable values, and in fact it's what they're for. We've written [a piece of documentation about this](https://swiftpackageindex.com/apple/swift-nio/2.82.0/documentation/niocore/loops-futures-concurrency), but I can repeat here the core insights of that documentation.

1. Each instance of a NIO `EventLoop` defines a single unitary isolation domain. Within this domain, values of non-`Sendable` type may be passed around freely, as they can never lead to data races.
2. All callbacks passed to an `EventLoopFuture` execute on one `EventLoop`, the `EventLoop` to which that `EventLoopFuture` is bound.
3. Multiple callbacks may be passed to the same `EventLoopFuture`: passing a callback does not consume the future. Similarly, `wait` and `get` may be called multiple times on the same future. Thus, we can conclude that an `EventLoopFuture` holds a reference to its completed value indefinitely.
4. The corollary of (2) and (3) is that it is safe to put a non-`Sendable` value into an `EventLoopFuture` (that is, to succeed an `EventLoopPromise` with a non-`Sendable` value if-and-only-if you are already on the relevant `EventLoop` when you do so. Otherwise, you move the non-`Sendable` value _to_ the `EventLoop` to which the future is bound, which is not safe.
    - Sidebar: we could use `sending` in this case, instead of `Sendable`. We intend to do that when we only support Swift language versions that have `sending `support, i.e. when we drop 5.10 support.
5. To enable (4), we have provided the isolated views on [futures](https://swiftpackageindex.com/apple/swift-nio/2.82.0/documentation/niocore/eventloopfuture/isolated) and [promises](https://swiftpackageindex.com/apple/swift-nio/2.82.0/documentation/niocore/eventlooppromise/isolated). These views perform a runtime check that you are already on the correct `EventLoop`, and then remove the need for the values and closures to be `Sendable`.
6. `EventLoopFuture.wait()` is never safe to call on the same event loop as the one the `EventLoopFuture` is bound to. As a result, `EventLoopFuture.wait()` cannot be used with non-`Sendable` values: by definition you aren't on the `EventLoop, so it is not sound to give you the value.
7. `EventLoopFuture.get()` could only be safe to use with non-`Sendable` values _if_ it was called from the same `EventLoop` as the one to which the future is bound. This is only possible when using an `EventLoop` as a custom executor. Unfortunately, due to some limitations in region-based isolation in Swift 6.0 and 6.1, we can't safely enable this today, because it will allow violations of the concurrency model.

CrownedPhoenix:
Closing as I consider this resolved.
Thank you so much for your help and comments @Lukasa

PR #3221, opened 2025-04-23
PendingDatagramWritesManager split close and failAll steps

rnro:
### Motivation:

PendingDatagramWritesManager split close and failAll steps of the failAll method into two explicit methods. This separates concerns and makes future changes more straightforwards.

### Modifications:

PendingDatagramWritesManager split close and failAll steps of the failAll method into two explicit methods.

### Result:

No behavioral change

### DIFF
--- a/Sources/NIOPosix/PendingDatagramWritesManager.swift
+++ b/Sources/NIOPosix/PendingDatagramWritesManager.swift
@@ -693,13 +693,16 @@ final class PendingDatagramWritesManager: PendingWritesManager {
         }
     }
 
-    /// Fail all the outstanding writes. This is useful if for example the `Channel` is closed.
-    func failAll(error: Error, close: Bool) {
-        if close {
-            assert(self.isOpen)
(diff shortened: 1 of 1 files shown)


Sources/NIOPosix/PendingDatagramWritesManager.swift
```diff
@@ -693,13 +693,16 @@ final class PendingDatagramWritesManager: PendingWritesManager {
...
-        }
+    /// Close the write manager and fail all the outstanding writes. This is useful if for example the `Channel` is closed.
+    func closeAndFailAll(error: Error) {
+        assert(self.isOpen)
+        self.isOpen = false
 
+        self.failAll(error: error)
+    }
+
+    /// Fail all the outstanding writes. This is useful if for example the `Channel` is closed.
```
Lukasa, line 704: It might be useful to clean up these doc comments and explain where we might do one and the other.

PR #3225, opened 2025-04-28
Add unix domain socket based async channel test

danieleggert:
Add unix domain socket based async channel test.

### Motivation:

This adds additional test coverage, and also provides an example for how to setup async-channel based UNIX domain socket channel pairs (server + client).

### Modifications:

Adds a new file with test code.

### Result:

Changes to test only.

### DIFF
--- a/Tests/NIOCoreTests/AsyncChannel/AsynChannelUnixDomainSocketTests.swift
+++ b/Tests/NIOCoreTests/AsyncChannel/AsynChannelUnixDomainSocketTests.swift
@@ -1,0 +1,169 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftNIO open source project
+//
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftNIO project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import NIOPosix
+import Testing
+
+@testable import NIOCore
+
+@Suite
+private enum AsynChannelUnixDomainSocketTests {
+    /// This is a end-to-end async channel based test.
+    ///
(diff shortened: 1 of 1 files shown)


Lukasa (changes requested):
Thanks for this Daniel! In general we're happy to take a test like this, but it will need to be an XCTest, not a swift-testing test, as we still support 5.10.

Tests/NIOCoreTests/AsyncChannel/AsynChannelUnixDomainSocketTests.swift
```diff
@@ -0,0 +1,171 @@
...
+    // nested directory hierarchy.
+    let path = "/tmp/9ac7750dc22a066066871aadf481e31a"
+    let serverChannel = try await makeServerChannel(path: path)
+
+    try await withThrowingDiscardingTaskGroup { group in
+        try await serverChannel.executeThenClose { inbound in
+            group.addTask {
+                // Create a client connection to the server:
+                let clientChannel = try await makeClientChannel(path: path)
+                print("Executing client channel")
```
glbrntt, line 59: Can you remove the prints?

Tests/NIOCoreTests/AsyncChannel/AsynChannelUnixDomainSocketTests.swift
```diff
@@ -0,0 +1,171 @@
...
+        try await channel.executeThenClose { inbound, outbound in
+            for try await message in inbound {
+                print("S: Did receive '\(message)'")
+                guard message != "QUIT" else { return }
+                serverDidReceive.confirm()
+                try await outbound.write(message)
+            }
+            print("S: Bye")
+        }
+    } catch {
```
glbrntt, line 139: Can we just make `handleConnection` throwing and propagate that error up rather than suppressing it?

Tests/NIOCoreTests/AsyncChannel/AsynChannelUnixDomainSocketTests.swift
danieleggert: Even in test code? This is super helpful for diagnosing these when they fail.

Issue #3052, opened 2025-01-13
Remove underbar prefix from _NIOFileSystem 


adam-fowler:
I assume _NIOFileSystem is considered to be a public API now, given NonBlockingFileIO has been deprecated in its favour. Is it not time to remove the underbar prefix from its name. This is the quote from its original release

> The module isn't yet considered public API and is therefore available via the _NIOFileSystem product.

Lukasa:
Yes, we intend to stabilise this API in the fairly near future.

adam-fowler:
For the moment though we have a choice between a deprecated or an unstable API

weissi:
@glbrntt I suppose the holdup here is the problematic `SystemPackage.FilePath` (with is related Darwin-only cousin `System.FilePath`). I'd still recommend `NIOFilePath` that converts from/to `SystemPackage.FilePath` and `System.FilePath` and is of course `StringLiteralConvertible`.

adam-fowler:
What's the problem with FilePath?

weissi:
> What's the problem with FilePath?

The problem is that there are _two_ `FilePath`s on macOS: `System.FilePath` and` SystemPackage.FilePath` which are _not_ compatible. On Linux there's _currently_ only `SystemPackage.FilePath` which means we're stuck between a rock and a hard place.

I think until it's settled what FilePath is the "correct" one on all platforms, NIO should have a `NIOFilePath` which takes `String`, `System.FilePath` and `SystemPackage.FilePath` and is expressible by string literal such that it's compatible with everything.

Otherwise, if NIO standardise on -- say -- `SystemPackage.FilePath`, you'd be in a weird situation on Darwin. If NIO used `System.FilePath` on Darwin and `SystemPackage.FilePath` on Linux, you'd be in a weird situation everywhere as you need a conditional import (`System` vs. `SystemPackage`) and a conditional dependency (on `swift-system` or nothing).

Lukasa:
For what it's worth, this is something that I _think_ a third-party contributor should be able to do. The simplest version just transforms everything to `SystemPackage.FilePath`. A more complex version allows using `System.FilePath` on Darwin as well by storing the element in an enum, but that would be acceptable as an enhancement rather than an initial feature.

Then you just have initializers from the types @weissi listed above.

weissi:
@Lukasa good-first-bug?

MartinLau7:
> Yes, we intend to stabilise this API in the fairly near future.

[...]

PR #3226, opened 2025-04-29
Rename public product to NIOFileSystem; keep implementation target _N…

AnupamKumar-1:
NIOFileSystem

Remove leading underscore from the public NIOFileSystem product

### Motivation:

The `_NIOFileSystem` module has been stable long enough to be considered public API.  Consumers should import it without a leading underscore.  This makes the library name consistent with other SwiftNIO modules and marks it as a stable, supported API.

### Modifications:

- Renamed the library product in `Package.swift` from `_NIOFileSystem` to `NIOFileSystem`.  
- Updated three test targets in `Package.swift` to depend on `NIOFileSystem` instead of `_NIOFileSystem`.  
- Left implementation target name and shim target intact.  The shim (`Sources/_NIOFileSystemExported/Exports.swift`) continues to `@_exported import _NIOFileSystem` so that the public module re-exports the implementation.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -68,7 +68,7 @@ let package = Package(
         .library(name: "NIOFoundationCompat", targets: ["NIOFoundationCompat"]),
         .library(name: "NIOWebSocket", targets: ["NIOWebSocket"]),
         .library(name: "NIOTestUtils", targets: ["NIOTestUtils"]),
-        .library(name: "_NIOFileSystem", targets: ["_NIOFileSystem", "NIOFileSystem"]),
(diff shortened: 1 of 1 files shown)


glbrntt:
Thanks for opening this. Unfortunately there are still ongoing discussions around some of the APIs so we can't yet consider this stable API. See also https://github.com/apple/swift-nio/issues/3052.

I think it's worth explicitly stating that underscored APIs _aren't_ public API and shouldn't be considered stable even if they haven't changed for a long time.

PR #3227, opened 2025-05-01
Removed Docker references as repo moved to Github actions

akbashev:
### Motivation:

A bit confusing to see docker mentioned in Readme file, when in reality you can't run docker compose anymore.

### Modifications:

Updated README and CONTRIBUTING files, removed some scripts.

### Result:

No mentions of docker so users are not confused.


akbashev:
Think dev/alloc-limits-from-test-output and dev/update-benchmark-thresholds.sh are not used anymore, so also removed.

.licenseignore
```diff
@@ -39,7 +39,6 @@ Sources/CNIOLLHTTP/c_nio_llhttp.c
 Sources/CNIOLLHTTP/include/c_nio_llhttp.h
 Sources/CNIOSHA1/c_nio_sha1.c
 Sources/CNIOSHA1/include/CNIOSHA1.h
-dev/alloc-limits-from-test-output
```
Lukasa, line 42: This script _is_ used, I updated it in #2987.

akbashev: huh, ok, will revert

Lukasa (approved):

Lukasa:
Thanks for this!

PR #3229, opened 2025-05-02
Replace almost all public static lets with computed vars

Lukasa:
Motivation:

Public static lets can serve a bunch of roles, but one of them is to store simple constants: integers, and other trivial types, for example. This is a nice pattern and for internal and private static lets it works well, but for public ones it produces some inefficient code.

In particular, it has two downsides. First, it allocates storage for that value. We don't actually need to allocate a few hundred extra megabytes for the various integers we want to store.

Secondly, it forces calling code to access the address and call the dispatch_once code in order to get hold of the value. For trivial types we don't need that cost: they can just know what the value is directly.

Inlinable computed vars avoid all of these costs: they have no size overhead for storage, and they are visible to all clients so their values can be directly assembled.

While I'm here, I added a bunch of other inlinable annotations for a few trivial data types I stumbled onto.

Modifications:

- Added loads of inlinables. Like, loads.
- Swapped many static lets to static vars.

Result:

Better codegen, smaller memory footprints, more attributes.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-hex.swift
+++ b/Sources/NIOCore/ByteBuffer-hex.swift
@@ -32,23 +32,29 @@ extension ByteBuffer {
     /// Can be either xxd output compatible, or hexdump compatible.
     public struct HexDumpFormat: Hashable, Sendable {
 
-        enum Value: Hashable {
+        @usableFromInline
+        enum Value: Hashable, Sendable {
             case plain(maxBytes: Int? = nil)
             case detailed(maxBytes: Int? = nil)
             case compact(maxBytes: Int? = nil)
         }
 
+        @usableFromInline
         let value: Value
+
+        @inlinable
         init(_ value: Value) { self.value = value }
 
         /// A plain hex dump format compatible with `xxd` CLI utility.
-        public static let plain = Self(.plain(maxBytes: nil))
+        @inlinable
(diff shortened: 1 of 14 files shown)


glbrntt (approved):

Sources/NIOCore/ByteBuffer-hex.swift
```diff
@@ -32,23 +32,29 @@ extension ByteBuffer {
...
+        public static var plain: HexDumpFormat { Self(.plain(maxBytes: nil)) }
 
         /// A hex dump format compatible with `hexdump` command line utility.
-        public static let detailed = Self(.detailed(maxBytes: nil))
+        @inlinable
+        public static var detailed: HexDumpFormat { Self(.detailed(maxBytes: nil)) }
 
         /// A hex dump analog to `plain` format  but without whitespaces.
-        public static let compact = Self(.compact(maxBytes: nil))
+        public static var compact: HexDumpFormat { Self(.compact(maxBytes: nil)) }
```
fabianfett, line 57: @Lukasa the above did get an `@inlinable`, which makes sense based on the discussion we had last week. This one didn't. Is this an oversight?

Lukasa: Almost certainly, yes.

Issue #3222, opened 2025-04-23
Errors should have a way of getting a printable message


axelandersson:
Currently logging an error received from a `whenFailure()` callback results in something like this:

```
The operation couldn't be completed. (NIOPosix.NIOConnectionError error 1.)
```

It would be nice if `NIOConnectionError` could provide an error message instead.

The same applies to the `ChannelError` enum.

glbrntt:
This happens because you're using the `localizedDescription` which is provided by `Foundation`. I believe only `NSError`s will produce a "useful" description using `localizedDescription`.

Generally speaking avoid `localizedDescription` unless you know the underlying error is an `NSError`.

glbrntt:
Yeah they should conform to `CustomStringConvertible` -- I'm a little surprised they don't already.

axelandersson:
Fix was merged in https://github.com/apple/swift-nio/pull/3230.

PR #3230, opened 2025-05-06
Make ChannelError and NIOConnectionError conform to CustomStringConvertible

axelandersson:
### Motivation:

`ChannelError` and `NIOConnectionError` should conform `CustomStringConvertible` so that these errors can be easily logged.

As discussed in https://github.com/apple/swift-nio/issues/3222.

### Modifications:

Added conformance to `ChannelError` and wrote error messages for all cases.

Added conformance to `NIOConnectionError`.


Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -70,6 +70,28 @@ public struct NIOConnectionError: Error {
     }
 }
 
+extension NIOConnectionError: CustomStringConvertible {
+    public var description: String {
+        if let dnsError = (dnsAError ?? dnsAAAAError) {
```
glbrntt, line 75: nit: coding style in NIO is to use explicit `self`:

```suggestion
        if let dnsError = (self.dnsAError ?? self.dnsAAAAError) {
```

Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -70,6 +70,28 @@ public struct NIOConnectionError: Error {
...
+             return "DNS error: \(dnsError.localizedDescription)"
```
glbrntt, line 76: Localized description gives bad error messages unless the underlying error is an `NSError`:

```suggestion
             return "DNS error: \(dnsError)"
```

Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -70,6 +70,28 @@ public struct NIOConnectionError: Error {
...
+                     channelError.description
+                 } else {
+                     $0.error.localizedDescription
+                 }
+             }
+
+             return "Connection errors: \(descriptions.joined(separator: ", "))"
+         }
+
+         return "Unknown error"
```
glbrntt, line 91: I think this can still hint that it's a connection error: e.g. `"NIOConnectionError: unknown"`

Sources/NIOPosix/HappyEyeballs.swift
```diff
@@ -70,6 +70,28 @@ public struct NIOConnectionError: Error {
...
+             return "DNS error: \(dnsError.localizedDescription)"
+         }
+
+         if !connectionErrors.isEmpty {
+             let descriptions = connectionErrors.map {
+                 if let channelError = $0.error as? ChannelError {
+                     channelError.description
+                 } else {
+                     $0.error.localizedDescription
+                 }
```
glbrntt, line 85: No need to do any casting here and avoid using `localizedDescription`. Just use `String(describing: $0)` here.

glbrntt (changes requested):
Thanks for opening this @axelandersson -- I left some feedback inline.

Sources/NIOPosix/HappyEyeballs.swift
axelandersson: Thanks - updated.

Sources/NIOPosix/HappyEyeballs.swift
axelandersson: Thanks - updated.

Sources/NIOPosix/HappyEyeballs.swift
axelandersson: Thanks - updated.

Sources/NIOPosix/HappyEyeballs.swift
axelandersson: Thanks - updated.

PR #3248, opened 2025-05-16
Revert "Revert "Remove default setup command for Xcode 16.2 (#3242)" (#3244)"

glbrntt:


rnro (approved):
revert revert final v2

PR #3254, opened 2025-05-26
Embedded channels should set local and remote address always

hamzahrmalik:
Embedded channels should set local and remote address always

### Motivation:

Currently, if connect or bind are called without a promise, the remote/local address does not get set because those were getting set in the whenSuccess of the promise.

### Modifications:

If the user didn't provide a promise, create one ourselves, so we have something to listen for.

### Result:

Calling connect/bind will always result in remote/local address getting set, regardless of whether you pass a promise


Lukasa (approved):
Awesome, this looks really good. Thanks @hamzahrmalik!

PR #3263, opened 2025-06-04
Add `scheduleCallback` APIs to `NIOIsolatedEventLoop`

### LINKED ISSUE #3262: `NIOIsolatedEventLoop` does not offer fast timer API
#2759 introduced a super fast, super cool new timer API (`scheduleCallback(at:handler:)`). Sadly `NIOIsolatedEventLoop` doesn't have an extension to create timers using this new API. Since ChannelHandlers are normally not Sendable, this would be a major quality of live improvement.

ptoffy:
Fixes #3262 by adding the missing APIs.

### Motivation:

As explained by the issue linked above, having a non-`Sendable`-requiring variant of the new `scheduleCallback` APIs on `NIOIsolatedEventLoop` can be useful since we're not always dealing with `Sendable` types.

### Modifications:

This adds the required `scheduleCallback` APIs to `NIOIsolatedEventLoop` by wrapping the non-`Sendable` `NIOScheduleCallbackHandler` in a `NIOLoopBound`-based handler.
### Result:

The two `scheduleCallback`s can be used on `NIOIsolatedEventLoop` too.


Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -170,3 +170,23 @@ extension EventLoop {
         }
     }
 }
+
+@usableFromInline
+struct UnsafeUncheckedScheduledCallbackHandlerWrapper<Handler: NIOScheduledCallbackHandler>: @unchecked Sendable, NIOScheduledCallbackHandler {
```
fabianfett: Instead of using `@unchecked Sendable` we should use a NIOLoopBound here instead.

ptoffy: Done! [e1e581f](https://github.com/apple/swift-nio/pull/3263/commits/e1e581f1e62b358e06c443d3661ee263e50a0eb7). I wonder if that's the case for `scheduleTask` too where `UnsafeTransfer` (which is `@unchecked Sendable`) is used?

Sources/NIOCore/EventLoop.swift
```diff
@@ -458,6 +458,18 @@ public protocol EventLoop: EventLoopGroup {
...
+    func _scheduleCallbackIsolatedUnsafeUnchecked(
+        at deadline: NIODeadline,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback
+
+    @discardableResult
+    func _scheduleCallbackIsolatedUnsafeUnchecked(
+        in amount: TimeAmount,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback
```
fabianfett: @Lukasa do we need those additional hooks on the EL? What was the reasoning behind adding the `_scheduleCallbackIsolatedUnsafeUnchecked`?

Sources/NIOCore/EventLoop.swift
```diff
@@ -533,6 +545,28 @@ extension EventLoop {
...
         }
     }
+
+    @inlinable
+    @discardableResult
+    public func _scheduleCallbackIsolatedUnsafeUnchecked(
+        at deadline: NIODeadline,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback {
+        self.assertInEventLoop()
```
fabianfett: The NIOLoopBound already enforces being on the correct EL. For this reason we don't need to assert here.

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -170,3 +172,23 @@ extension EventLoop {
...
+
+    @usableFromInline
+    func handleScheduledCallback(eventLoop: some EventLoop) {
+        self.box.value.handleScheduledCallback(eventLoop: eventLoop)
+    }
+
+    @usableFromInline
+    func didCancelScheduledCallback(eventLoop: some EventLoop) {
+        self.box.value.didCancelScheduledCallback(eventLoop: eventLoop)
+    }
```
fabianfett, line 193: This type is generic over a user provided `Handler`, to ensure perf, this should not just be `@useableFromInline` it should be `@inlinable` instead.
```suggestion
    @inlinable
    init(wrapping handler: Handler, eventLoop: some EventLoop) {
        self.box = .init(handler, eventLoop: eventLoop)
    }

    @inlinable
    func handleScheduledCallback(eventLoop: some EventLoop) {
        self.box.value.handleScheduledCallback(eventLoop: eventLoop)
    }

    @inlinable
    func didCancelScheduledCallback(eventLoop: some EventLoop) {
        self.box.value.didCancelScheduledCallback(eventLoop: eventLoop)
    }
```

fabianfett (commented):
Thanks so much for the progress so far and for tackling this issue! @Lukasa can you please also give this an early look?

Sources/NIOCore/EventLoop.swift
Lukasa: We needed it because the existing EL hooks _require_ `Sendable`, _and_ they require a specific type. So it isn't possible to pass a `T` through them that isn't `Sendable`, because wrapping it into a `NIOLoopBound` makes it not a `T` anymore. In this case, a `NIOLoopBound<some (NIOScheduledCallbackHandler & Sendable)>` isn't a `some (NIOScheduledCallbackHandler & Sendable)`, so it cannot be passed through there.

Wrapping things in a sufficiently large number of loop bounds can work, but performs poorly, and we don't want to _pessimise_ the performance of the on-loop case. That's what we want to work _best_.

Sources/NIOCore/EventLoopFuture+AssumeIsolated.swift
```diff
@@ -125,6 +125,32 @@ public struct NIOIsolatedEventLoop {
...
+        in amount: TimeAmount,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback {
+        try self._wrapped._scheduleCallbackIsolatedUnsafeUnchecked(in: amount, handler: handler)
+    }
+
+    @inlinable
+    @available(*, noasync)
+    public func cancelScheduledCallback(_ scheduledCallback: NIOScheduledCallback) {
+        self._wrapped.cancelScheduledCallback(scheduledCallback)
```
Lukasa, line 166: The checking strategy here is incoherent. Methods that are marked `unsafeUnchecked` should have asserts. Methods without that need `preconditions`.

Sources/NIOCore/EventLoop.swift
fabianfett: Great, if I understand you correctly adding the above methods is our intend outcome.

fabianfett: Does this mean, we want to work with our own unsafe transfer method in this PR, so that we don't use NIOLoopBound at all? To work around the preconditions?

ptoffy: Cool, so we do want to use assertions (and therefore the `...UnsafeUnchecked` method) rather than the `NIOLoopBound` which provides a precondition

ptoffy: Oh I think we wrote that at the same time 😄

Lukasa: Yeah, and we should also implement a version that doesn't require this fallback for `SelectableEventLoop`, `EmbeddedEventLoop`, and `AsyncTestingEventLoop`.

ptoffy: Cool, done. [`97deac6`](https://github.com/apple/swift-nio/pull/3263/commits/97deac63b621c03df868bdc60f9c692cb5722b83) Should I leave those changes for a different PR?

Sources/NIOCore/EventLoop.swift
```diff
@@ -533,6 +545,28 @@ extension EventLoop {
...
+    @inlinable
+    @discardableResult
+    public func _scheduleCallbackIsolatedUnsafeUnchecked(
+        in amount: TimeAmount,
+        handler: some NIOScheduledCallbackHandler
+    ) throws -> NIOScheduledCallback {
+        self.assertInEventLoop()
+        let unsafeHandlerWrapper = LoopBoundScheduledCallbackHandlerWrapper(wrapping: handler, eventLoop: self)
+        return try self.scheduleCallback(in: amount, handler: unsafeHandlerWrapper)
+    }
```
fabianfett, line 599: I think we might need add a `_cancelScheduledCallbackUnsafeUnchecked` method here as well, @Lukasa?

ptoffy: Why? `NIOScheduledCallback` is `Sendable`

Sources/NIOCore/EventLoop.swift
Lukasa: Sounds reasonable to me.

Sources/NIOCore/EventLoop.swift
Lukasa: Indeed, once you have it you can call the regular API.

ptoffy:
Okay so I think this is ready for a proper review

Sources/NIOCore/NIOScheduledCallback.swift
```diff
@@ -170,3 +170,26 @@ extension EventLoop {
         }
     }
 }
+
+@usableFromInline
+struct UnsafeUncheckedScheduledCallbackHandlerWrapper<Handler: NIOScheduledCallbackHandler>:
```
Lukasa: I don't think this needs to be unsafe. Instead, let's implement this in terms of `NIOLoopBound`. That way we preserve the logical semantics, and avoid an `@unchecked Sendable`.

ptoffy: Didn't we decide on using assertions rather than preconditions a few messages above?

Lukasa: In the unchecked fast path, yes. But the slower fallback is just that: slower. As a result, we can safely re-use things we already have, and avoid proliferating `@unchecked Sendable`.

In general I have a plan to increasingly remove the `@unchecked Sendable`s in NIO because there are far too many, and they can be very damaging. So wherever we can use an existing spelling for this, I'd prefer we did.

fabianfett (approved):
@ptoffy thanks for this PR. I'll wait for @Lukasa final feedback before we can merge.

PR #3268, opened 2025-06-16
Extend NIOEventLoopTickInfo with end-time and loop sleep time

josephnoir:
### Motivation

NIO’s event loop can emit metrics. Currently, this includes the start time (i.e. when the loop started working) and the number of tasks which were executed in that tick.

It would be useful to collect the time the loop tick ended and the length of time the loop was asleep before starting the tick.

### Modifications

The event loop has a new property to track the time between runs. The tick info collects two new values:

* The time the loop finished: the end-time.
* The time since the last run: the sleep time.

For the first run, the sleep time is the time since the event loop was created.

### Result

The new metrics are available for collection.


Tests/NIOPosixTests/EventLoopMetricsDelegateTests.swift
```diff
@@ -49,22 +49,36 @@ final class EventLoopMetricsDelegateTests: XCTestCase {
...
             promise.succeed()
         }
+
         promise.futureResult.whenSuccess {
-            // There are 3 tasks (scheduleTask, whenSuccess, wait) which can trigger a total of 1...3 ticks
-            XCTAssertTrue((1...3).contains(delegate.infos.count), "Expected 1...3 ticks, got \(delegate.infos.count)")
-            // the total number of tasks across these ticks should be either 2 or 3
+            // There are 4 tasks (scheduleTask, scheduleTask, whenSuccess, wait) which can trigger a total of 2...4 ticks
+            XCTAssertTrue((2...4).contains(delegate.infos.count), "Expected 2...4 ticks, got \(delegate.infos.count)")
+            // The total number of tasks across these ticks should be either 2 or 3
```
glbrntt:
```suggestion
            // The total number of tasks across these ticks should be either 3 or 4
```

glbrntt (commented):
Left a fix for a comment but otherwise this looks good

PR #3269, opened 2025-06-16
Adding RawSpan support to writeBytes in ByteBuffer

yigityazicilar:
Motivation:

With the new Swift version (6.2) we get access to spans for safely interacting with the contiguous memory owned by other types. Allow users write bytes to the ByteBuffer with RawSpans.

Modifications:

* ByteBuffer-core: Overload the setBytes, _setBytes, and _setBytesAssumingUniqueBufferAccess to use RawSpan instead of UnsafeRawBufferPointer
* ByteBuffer-aux: Overload the writeBytes to use RawSpan instead of UnsafeRawBufferPointer

Result:

Users can now write the contiguous memory of other types into ByteBuffer using the new RawSpan overloads, removing the need for unsafe pointer handling.

### DIFF
--- a/Sources/NIOCore/ByteBuffer-aux.swift
+++ b/Sources/NIOCore/ByteBuffer-aux.swift
@@ -517,6 +517,22 @@ extension ByteBuffer {
         return written
     }
 
+    #if compiler(>=6.2)
+    /// Write `bytes` into this `ByteBuffer`. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - Parameters:
+    ///   - bytes: A `RawSpan`
+    /// - Returns: The number of bytes written or `bytes.byteCount`.
+    @discardableResult
+    @inlinable
+    @available(macOS 26, iOS 26, tvOS 26, watchOS 26, visionOS 26, *)
+    public mutating func writeBytes(_ bytes: RawSpan) -> Int {
+        let written = self.setBytes(bytes, at: self.writerIndex)
+        self._moveWriterIndex(forwardBy: written)
+        return written
+    }
+    #endif
+
     /// Writes `byte` `count` times. Moves the writer index forward by the number of bytes written.
     ///
     /// - Parameters:
--- a/Sources/NIOCore/ByteBuffer-core.swift
+++ b/Sources/NIOCore/ByteBuffer-core.swift
@@ -508,6 +508,34 @@ public struct ByteBuffer {
         targetPtr.copyMemory(from: bytes)
     }
 
+    #if compiler(>=6.2)
+    @inlinable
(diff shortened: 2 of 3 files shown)


Sources/NIOCore/ByteBuffer-aux.swift
```diff
@@ -517,6 +517,22 @@ extension ByteBuffer {
...
+    #if compiler(>=6.2)
+    /// Write `bytes` into this `ByteBuffer`. Moves the writer index forward by the number of bytes written.
+    ///
+    /// - Parameters:
+    ///   - bytes: A `RawSpan`
+    /// - Returns: The number of bytes written or `bytes.byteCount`.
+    @discardableResult
+    @inlinable
+    @available(macOS 26, iOS 26, tvOS 26, watchOS 26, visionOS 26, *)
+    public mutating func writeBytes(_ bytes: RawSpan) -> Int {
```
FranzBusch, line 529: Should we mark this as `consuming`?

Sources/NIOCore/ByteBuffer-core.swift
```diff
@@ -1035,6 +1063,16 @@ extension ByteBuffer {
         Int(self._setBytes(bytes, at: _toIndex(index)))
     }
 
+    #if compiler(>=6.2)
+    /// Copy `bytes` from a `RawSpan` into the `ByteBuffer` at `index`. Does not move the writer index.
+    @discardableResult
+    @inlinable
+    @available(macOS 26, iOS 26, tvOS 26, watchOS 26, visionOS 26, *)
+    public mutating func setBytes(_ bytes: RawSpan, at index: Int) -> Int {
```
FranzBusch, line 1071: Same here?

Sources/NIOCore/ByteBuffer-aux.swift
Lukasa: There's no need: we don't consume it. We copy the bytes out.

Lukasa: Oh, and also `Span` is copyable, so `consuming` doesn't have much of a practical effect either. It can, in the best case, elide a copy of the `Span` object, but that's just two words so it's not really worth the hassle.

FranzBusch: Yeah `Span` is `Copyable` so this might give the compiler a nudge to optimize it a bit better. At least that's what I was thinking.

weissi: If anything, it should be `borrowing`, right? We're happy to just borrow this -- now as per @Lukasa's point: not much to win here of course.

PR #3276, opened 2025-06-28
Fix API documentation links in README.md

azarovalex:
Point API docs URLs to the latest stable version instead of the unstable main branch.

### Motivation:

For most of developers this README.md is the main entry point when searching online or installing the library, so pointing them to API docs for the latest stable release makes more sense then the main branch, which may contain not-yet-released API or breaking changes.

### Modifications:

This PR removes `main` branch name from API docs URLs, making them always point to the latest stable release instead of the main branch.

### Result:

API docs URLs in README.md point to the latest stable release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ The core SwiftNIO repository will contain a few extremely important protocol imp
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


rnro:
Thanks for contributing this! The approach makes sense to me and looks good.

rnro (approved):

PR #3277, opened 2025-06-29
Support for 64-bit timespec seconds on 32-bit platforms

lhoward:
Modern 32-bit platforms (such as embedded Linux builds) support 64-bit timestamps in `time_t` and `timespec`. Accommodate this.

### DIFF
--- a/Sources/NIOConcurrencyHelpers/lock.swift
+++ b/Sources/NIOConcurrencyHelpers/lock.swift
@@ -253,11 +253,7 @@ public final class ConditionLock<T: Equatable> {
         gettimeofday(&curTime, nil)
 
         let allNSecs: Int64 = timeoutNS + Int64(curTime.tv_usec) * 1000
-        #if canImport(wasi_pthread)
-        let tvSec = curTime.tv_sec + (allNSecs / nsecPerSec)
-        #else
-        let tvSec = curTime.tv_sec + Int((allNSecs / nsecPerSec))
-        #endif
+        let tvSec = curTime.tv_sec + time_t((allNSecs / nsecPerSec))
 
         var timeoutAbs = timespec(
             tv_sec: tvSec,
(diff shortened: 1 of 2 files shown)


Sources/NIOConcurrencyHelpers/lock.swift
```diff
@@ -252,12 +252,16 @@ public final class ConditionLock<T: Equatable> {
...
+	typealias _Seconds = Int
+	#endif
+
         let allNSecs: Int64 = timeoutNS + Int64(curTime.tv_usec) * 1000
-        #if canImport(wasi_pthread)
-        let tvSec = curTime.tv_sec + (allNSecs / nsecPerSec)
-        #else
-        let tvSec = curTime.tv_sec + Int((allNSecs / nsecPerSec))
-        #endif
+        let tvSec = curTime.tv_sec + _Seconds((allNSecs / nsecPerSec))
```
Lukasa: I'm inclined to assume that we could actually just make this `time_t` on all platforms.

Sources/NIOPosix/SelectorGeneric.swift
```diff
@@ -43,10 +43,18 @@ extension Optional {
...
+        typealias _Seconds = time_t
+        #else
+        typealias _Seconds = Int
+        #endif
+
         let nsecPerSec: Int64 = 1_000_000_000
         let ns = amount.nanoseconds
         let sec = ns / nsecPerSec
-        self = timespec(tv_sec: Int(sec), tv_nsec: Int(ns - sec * nsecPerSec))
+        self = timespec(tv_sec: _Seconds(sec), tv_nsec: Int(ns - sec * nsecPerSec))
```
Lukasa: Similarly, I'm inclined to assume we could just make this a `time_t` on all platforms.

Lukasa (commented):
Thanks for filing this! I've left a quick note in the diff.

Lukasa (approved):

Lukasa:
Thanks for this! :sparkles:

lhoward:
Great! Out of interest, what's the release cadence of swift-nio? Curious when we can enable 64-bit timestamps in the Yocto layer.

https://github.com/jeremy-prater/meta-swift/pull/12

cc @xtremekforever

Lukasa:
I expect we'll be able to release this week.

Lukasa:
Shipped as 2.84.0.

PR #3296, opened 2025-07-11
Structured Concurrency compliant MTELG create/shutdown

weissi:
### Motivation:

Structured Concurrency is helpful but so far, MTELG was very difficult to use with it mostly because Swift is still missing `async` in `defer`s.

This builds on
- #3297 
- #3302
- #3304

### Modifications:

- Add MTELG.withELG { ... }

### Result:

MTELG + SC = <3

### DIFF
--- a/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
+++ b/Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
@@ -622,3 +622,43 @@ extension NIODeadline {
         return self - target
     }
 }
+
+extension MultiThreadedEventLoopGroup {
+    #if compiler(>=6.0)
+    /// Start & automatically shut down a new ``MultiThreadedEventLoopGroup``.
+    ///
+    /// This method allows to start & automatically dispose of a ``MultiThreadedEventLoopGroup`` following the principle of Structured Concurrency.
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/MultiThreadedEventLoopGroup.swift
```diff
@@ -587,3 +610,38 @@ extension NIODeadline {
         return self - target
     }
 }
+
+extension MultiThreadedEventLoopGroup {
+    /// Start & automatically shut down a new ``MultiThreadedEventLoopGroup``.
+    ///
+    /// This method allows to start & automatically dispose of a ``MultiThreadedEventLoopGroup`` following the principle of Structured Concurrency.
+    /// The ``MultiThreadedEventLoopGroup`` is guaranteed to be shut down upon return, whether `body` throws or not.
```
Lukasa, line 631: Can we add text here that discourages the use of this outside of testing cases or top-level code, and suggests users should prefer the singleton group? Generally we don't want people doing this too much.

weissi: great idea done

weissi:
I will split this PR in two:
- the NIOThread refactor
- the actual new functionality

PR #3297, opened 2025-07-11
Jw niothread refactor

weissi:
### Motivation:

NIOThread assumed it can hold onto `pthread_t` forever. Unfortunately, on Linux, `pthread_t` is dead when the thread got joined (if not detached) or if it exits (if detached) _which is at a random point in time_.

### Modifications:

- Only assume ownership owner the `pthread_t` if not detached
- `nil` out the `pthread_t` on `join()`

### Result:

- Fewer `pthread_t` lifecycle bugs.


Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -1026,13 +1026,49 @@ internal final class SelectableEventLoop: EventLoop, @unchecked Sendable {
...
+                return """
+                    SelectableEventLoop { \
+                    selector = \(self._selector), \
+                    scheduledTasks = \(self._scheduledTasks.description), \
+                    thread = \(self.thread), \
+                    state = \(self.internalState) \
+                    }
+                    """
+            } else {
+                return self.externalStateLock.withLock {
```
Lukasa, line 1069: Why is the lock-holding concern not as relevant here?

Sources/NIOPosix/Thread.swift
```diff
@@ -42,41 +44,74 @@ protocol ThreadOps {
 /// All methods exposed are thread-safe.
 @usableFromInline
 final class NIOThread: Sendable {
-    internal typealias ThreadBoxValue = (body: (NIOThread) -> Void, name: String?)
+    internal typealias ThreadBoxValue = (body: (NIOThread) -> Void, name: String?, isDetached: Bool)
```
Lukasa, line 48: Should we tweak this API to try to discourage creating detached threads outside of tests?

Sources/NIOPosix/Thread.swift
```diff
@@ -42,41 +44,74 @@ protocol ThreadOps {
...
+
+    static var currentThreadName: String? {
+        ThreadOpsSystem.threadName(.init(handle: pthread_self()))
+    }
+
+    static var currentThreadID: UInt {
+        .init(bitPattern: .init(bitPattern: ThreadOpsSystem.currentThread.handle))
+    }
+
+    func takeOwnership() {
```
Lukasa: Should this return the thread ID?

Sources/NIOPosix/SelectableEventLoop.swift
weissi: 1. it already holds a lock, so I'm not making things worse I think :P 
2. it's `debugDescription` so maybe less likely to be called?

weissi: Basically replicating prior art in the existing impl

Sources/NIOPosix/Thread.swift
weissi: @Lukasa Honestly, I'm planning a follow-up to remove the ability to use detached threads at all, we don't use them outside of tests and we don't need them in tests.

Sources/NIOPosix/Thread.swift
weissi: good idea

Lukasa (approved):

PR #3298, opened 2025-07-12
Run update_and_patch_llhttp.sh script to update to 9.3.0

xtremekforever:
Update llhttp dependency in `CNIOLLHTTP` target ahead of adding a new patch to the `update_and_patch_llhttp.sh` script.

### Motivation:

We need to update the `update_and_patch_llhttp.sh` script to add an additional patch to fix compiling swift-nio with C++ interop enabled in Swift targets. It was recommended to update llhttp separately from the patch update and ahead of time to keep the changes separate from each other.

### Modifications:

Ran the `update_and_patch_llhttp.sh` script in the CNIOLLHTTP target to update the llhttp dependency to the latest version (9.3.0). It appears that the only changes from 9.2.1 (previous version) is a few new includes, and the definition of the `UNREACHABLE` macro that is used instead of the `abort()` call.

### Result:

Everything compiles locally and test suite passes with these changes.

@Lukasa


Lukasa (approved):
Very nice, LGTM.

PR #3300, opened 2025-07-13
[Issue-3295] - Make file handler sendable

ali-ahsan-ali:
Solve Issue-3295

### Motivation:

### Modifications:

### Result:


glbrntt (changes requested):
Thanks @ali-ahsan-ali. I think we can actually make the readable file handle and the directory file handle sendable too. Would you mind adding those too?

glbrntt (approved):

ali-ahsan-ali:
❤️

PR #3301, opened 2025-07-14
inEventLoop benchmark

weissi:
### Motivation:

PR #3297 is changing `inEventLoop` and `inEventLoop` is very much on the perf path. So let's add a benchmark

### Modifications:

- Add `(!)inEventLoop` benchmark
- Fix a guaranteed deadlock in `TCPThroughputBenchmark` to make the benchmarks runnable

### Result:

- More benchmarks
- Fewer hangs

### DIFF
--- a/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
+++ b/Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
@@ -128,8 +128,6 @@ final class TCPThroughputBenchmark: Benchmark {
             .bind(host: "127.0.0.1", port: 0)
             .wait()
 
-        self.serverHandler = try promise.futureResult.wait()
-
         self.clientChannel = try ClientBootstrap(group: group)
             .channelInitializer { channel in
                 channel.eventLoop.makeCompletedFuture {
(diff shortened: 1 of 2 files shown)


Sources/NIOPerformanceTester/TCPThroughputBenchmark.swift
```diff
@@ -140,6 +138,8 @@ final class TCPThroughputBenchmark: Benchmark {
             .connect(to: serverChannel.localAddress!)
             .wait()
 
+        self.serverHandler = try promise.futureResult.wait()
```
weissi, line 141: without this change, this is a guaranteed deadlock. The server `childChannelInitializer` can't run before the client starts connecting.

Sources/NIOPerformanceTester/main.swift
```diff
@@ -838,6 +838,31 @@ measureAndPrint(desc: "future_reduce_into_10k_futures") {
     return try! EventLoopFuture<Int>.reduce(into: 0, futures, on: el1, { $0 += $1 }).wait()
 }
 
+try measureAndPrint(desc: "el_in_eventloop_100M") {
```
weissi, line 841: it's about 0.2s in release mode on my machine, should be alright

PR #3302, opened 2025-07-14
improve inEventLoop checks

weissi:
### Motivation:

`inEventLoop` is very much in the performance path of SwiftNIO, especially these days with Concurrency, `NIOLoopBound` and friends. Previously, we relied on `pthread_equal(pthread_self(), myPthread)`, however, this could cause a number of issues.

1. Holding onto a `pthread_t` after `.join` is actually illegal (fixed in #3297)
2. ABA issues when `pthread_t` pointer are re-used for new pthreads
3. Fix would require a lock around `myPthread` which makes things (2x slower, even without contention)

### Modifications:

- New type `SelectableEventLoopUniqueID` which can be packed into a `UInt`
- Attach them into a C thread local
- Added `testInEventLoopABAProblem()` which does some basic testing (passes now, failed before this PR)

### Result:

- Even faster than the old, incorrect version
  - old: `measuring: el_in_eventloop_100M: 0.257395375, 0.241049208, 0.243188792, 0.259125916, 0.24843225, 0.229690125, 0.244281541, 0.225078834, 0.236395, 0.233305167` 
  - new: `measuring: el_in_eventloop_100M: 0.175561125, 0.187225625, 0.199269375, 0.19740975, 0.1922695, 0.179850958, 0.177612458, 0.17665125, 0.17897475, 0.18038775`
- More correct
- Groundwork to make #3297 not make things slower

### DIFF
--- a/Sources/NIOCore/ByteBuffer-int.swift
+++ b/Sources/NIOCore/ByteBuffer-int.swift
@@ -180,7 +180,7 @@ extension UInt32 {
 
         var n = self
 
-        #if arch(arm) || arch(i386) || arch(arm64_32)
+        #if arch(arm) || arch(i386) || arch(arm64_32) || arch(wasm32)
         // on 32-bit platforms we can't make use of a whole UInt32.max (as it doesn't fit in an Int)
         let max = UInt32(Int.max)
         #else
--- a/Sources/NIOPosix/SelectableEventLoop.swift
+++ b/Sources/NIOPosix/SelectableEventLoop.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import Atomics
+import CNIOPosix
 import DequeModule
 import Dispatch
 import NIOConcurrencyHelpers
@@ -134,6 +135,8 @@ internal final class SelectableEventLoop: EventLoop, @unchecked Sendable {
     internal var _scheduledTasks = PriorityQueue<ScheduledTask>()
     @usableFromInline
     internal var _immediateTasks = Deque<UnderlyingTask>()
+    @usableFromInline
+    internal let _uniqueID: SelectableEventLoopUniqueID
 
     // We only need the ScheduledTask's task closure. However, an `Array<() -> Void>` allocates
     // for every appended closure. https://bugs.swift.org/browse/SR-15872
@@ -247,12 +250,14 @@ internal final class SelectableEventLoop: EventLoop, @unchecked Sendable {
 
     internal init(
         thread: NIOThread,
+        uniqueID: SelectableEventLoopUniqueID,
         parentGroup: MultiThreadedEventLoopGroup?,  // nil iff thread take-over
         selector: NIOPosix.Selector<NIORegistration>,
         canBeShutdownIndividually: Bool,
         metricsDelegate: NIOEventLoopMetricsDelegate?
     ) {
         self.metricsDelegate = metricsDelegate
+        self._uniqueID = uniqueID
         self._parentGroup = parentGroup
         self._selector = selector
         self.thread = thread
@@ -267,6 +272,8 @@ internal final class SelectableEventLoop: EventLoop, @unchecked Sendable {
         let voidPromise = self.makePromise(of: Void.self)
         voidPromise.succeed(())
         self._succeededVoidFuture = voidPromise.futureResult
+        preconditionInEventLoop()
+        precondition(self._uniqueID.matchesCurrentThread)
     }
 
     deinit {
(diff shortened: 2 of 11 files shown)


Sources/NIOCore/ByteBuffer-int.swift
```diff
@@ -180,7 +180,7 @@ extension UInt32 {
 
         var n = self
 
-        #if arch(arm) || arch(i386) || arch(arm64_32)
+        #if arch(arm) || arch(i386) || arch(arm64_32) || arch(wasm32)
```
Lukasa, line 183: It's unfortunate that `pointerBitWidth` didn't get made stable yet.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -1098,3 +1106,78 @@ extension SelectableEventLoop {
...
+    }
+
+    @inlinable
+    init(_loopID: UInt32, groupID: UInt32) {
+        self._loopID = _loopID
+        self._groupID = groupID
+    }
+
+    static func makeNextGroup() -> Self {
+        let groupID = Self._nextGroupID.loadThenWrappingIncrement(ordering: .relaxed)
```
Lukasa, line 1138: Wrapping here is _extremely_ dangerous: we'll start recycling IDs. I appreciate the desire to avoid the checked increment in the atomic, but we should validate that we never see `0` out of this.

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -1098,3 +1106,78 @@ extension SelectableEventLoop {
...
+
+    @inlinable
+    var groupID: Int {
+        Int(self._groupID)
+    }
+
+    @inlinable
+    var loopID: Int {
+        Int(self._loopID)
+    }
```
Lukasa, line 1129: Is this just a long-winded way of saying `@inlinable private(set)`?

Sources/NIOPosix/SelectableEventLoop.swift
```diff
@@ -1098,3 +1106,78 @@ extension SelectableEventLoop {
...
+
+    @inlinable
+    internal var matchesCurrentThread: Bool {
+        let threadUniqueID = c_nio_posix_get_el_id()
+        return threadUniqueID == self.packedEventLoopID
+    }
+
+    @inlinable
+    internal var packedEventLoopID: UInt {
```
Lukasa, line 1155: Probably wants a leading underscore.

Sources/NIOPosix/SelectableEventLoop.swift
weissi: Yes, great point, I took this line from the existing code. Will change

weissi: don't need perf here haha :)

weissi: oh, those atomics do not have a non-wrapping increment. Will use an `UInt64` increment and then `UInt32` it

Sources/NIOPosix/SelectableEventLoop.swift
weissi: It's also `Int` for the "public" API (didn't want to 'leak' the UInt32'd-ness in the guts as we may need to change that in the future, say if we wanted to add randomness and switched to `_UInt24`)

Sources/NIOPosix/SelectableEventLoop.swift
weissi: Was on the fence here, doesn't really hurt to "open" it up internally but I'm happy to add it if you prefer

Sources/NIOCore/ByteBuffer-int.swift
weissi: yes, very annoying.

weissi: happy to use `#if _bitPointerWidth` or whatever it is if the guidance on `_`ed things has changed

Sources/NIOPosix/SelectableEventLoop.swift
weissi: @Lukasa fixed, now a UInt64 that we later cast to UInt32, so we'll crash on the 2^32th ELG :)

Sources/NIOCore/ByteBuffer-int.swift
Lukasa: The guidance hasn't changed, sadly.

Sources/NIOPosix/SelectableEventLoop.swift
Lukasa: :+1: Fantastic

Sources/NIOPosix/SelectableEventLoop.swift
Lukasa: Oh gotcha, missed that.

PR #3303, opened 2025-07-14
Wrap closures stored in the CallbackList

glbrntt:
Motivation:

The compiler is better at optimising closures wrapped in nominal types than raw closures when used as generic parameters. CallbackList is used a lot and stores an optional closure as well as on optional array of closures. Wrapping these internally should save some allocations.

Modifications:

- Wrap the element stored in the `CallbackList`.

Result:

Fewer allocs


Sources/NIOCore/EventLoopFuture.swift
```diff
@@ -23,16 +23,30 @@ import Dispatch
...
+
+        @inlinable
+        init(_ callback: @escaping () -> CallbackList) {
+            self.callback = callback
+        }
+    }
+
     @usableFromInline
-    internal var firstCallback: Optional<Element>
+    internal var firstCallback: Optional<Wrapper>
```
Lukasa: Did we need to rename this? It's called `Element`.

Lukasa (approved):

PR #3304, opened 2025-07-15
NIOThread: remove the detached threads functionality

weissi:
### Motivation:

Detached threads are a bad idea, their lifetime becomes random and NIO doesn't need this functionality (except for in some tests that are easily refactored)

### Modifications:

- Refactor tests to not use detached threads
- Remove detached threads functionality

### Result:

- Cleaner, less complex and more correct code


Sources/NIOPosix/Thread.swift
```diff
@@ -77,8 +77,8 @@ final class NIOThread: Sendable {
     ///
     /// - arguments:
     ///   - handle: The `ThreadOpsSystem.ThreadHandle` that is wrapped and used by the `NIOThread`.
-    internal init(handle: ThreadOpsSystem.ThreadHandle, isDetachedThread: Bool, desiredName: String?) {
-        self.handle = NIOLockedValueBox(isDetachedThread ? nil : handle)
+    internal init(handle: ThreadOpsSystem.ThreadHandle, desiredName: String?) {
+        self.handle = NIOLockedValueBox(handle)
```
Lukasa, line 81: This is being optional-promoted, you can remove the optional on the type definition.

weissi: Alas when we `join()` we need to set it to `nil` to not hold onto the then-deallocated `pthread_t`.

PR #3308, opened 2025-07-17
Support options on AsyncTestingChannel and EmbeddedChannel

### LINKED ISSUE #3305: `NIOAsyncTestingChannel` and `EmbeddedChannel` option handling
Considering that `NIOAsyncTestingChannel` and `EmbeddedChannel` would it make sense to allow more than  `ChannelOptions.Types.AllowRemoteHalfClosureOption`?

I'm trying to test a traceroute implementation which sets `ChannelOptions.socket(IPPROTO_IP, IP_TTL)` but the test crashes because of a [`fatalError`](https://github.com/apple/swift-nio/blob/344db24aaa0eb3a56f47a700a9121639fe5437f9/Sources/NIOEmbedded/AsyncTestingChannel.swift#L580).

Would it make sense to allow any (socket) option? Perhaps storing them in the type so `getOption` would be able to return it as well?

klaaspieter:
Support options on AsyncTestingChannel and EmbeddedChannel

Fixes #3305 

### Motivation:

The channels are intended for testing purposes so while most options have no practical use setting and getting them can be useful for testing.

For example a traceroute implementation will set TTL to the current hop number. Testing such an implementation requires the channels to pretend to support the TTL option.

### Modifications:

Added option storage to AsyncTestingChannel and EmbedededChannel and made `getOptionSync` and `setOptionSync` read and write from that storage.

### Result:

EmbeddedChannel and AsyncTestingChannel support changing their options.

### DIFF
--- a/Sources/NIOEmbedded/AsyncTestingChannel.swift
+++ b/Sources/NIOEmbedded/AsyncTestingChannel.swift
@@ -205,15 +205,19 @@ public final class NIOAsyncTestingChannel: Channel {
     nonisolated(unsafe) var channelcore: EmbeddedChannelCore!
     nonisolated(unsafe) private var _pipeline: ChannelPipeline!
 
-    private struct State {
+    @usableFromInline
+    internal struct State: Sendable {
         var isWritable: Bool
         var localAddress: SocketAddress?
         var remoteAddress: SocketAddress?
+
+        @usableFromInline
         var options: [(option: any ChannelOption, value: any Sendable)]
     }
 
     /// Guards any of the getters/setters that can be accessed from any thread.
-    private let stateLock = NIOLockedValueBox(
+    @usableFromInline
+    internal let _stateLock = NIOLockedValueBox(
         State(isWritable: true, localAddress: nil, remoteAddress: nil, options: [])
     )
 
@@ -230,10 +234,10 @@ public final class NIOAsyncTestingChannel: Channel {
     /// - see: `Channel.isWritable`
     public var isWritable: Bool {
         get {
-            self.stateLock.withLockedValue { $0.isWritable }
+            self._stateLock.withLockedValue { $0.isWritable }
         }
         set {
-            self.stateLock.withLockedValue {
+            self._stateLock.withLockedValue {
                 $0.isWritable = newValue
             }
         }
@@ -242,10 +246,10 @@ public final class NIOAsyncTestingChannel: Channel {
     /// - see: `Channel.localAddress`
     public var localAddress: SocketAddress? {
         get {
-            self.stateLock.withLockedValue { $0.localAddress }
+            self._stateLock.withLockedValue { $0.localAddress }
         }
         set {
-            self.stateLock.withLockedValue {
+            self._stateLock.withLockedValue {
                 $0.localAddress = newValue
             }
         }
@@ -254,29 +258,18 @@ public final class NIOAsyncTestingChannel: Channel {
     /// - see: `Channel.remoteAddress`
     public var remoteAddress: SocketAddress? {
         get {
-            self.stateLock.withLockedValue { $0.remoteAddress }
+            self._stateLock.withLockedValue { $0.remoteAddress }
         }
         set {
-            self.stateLock.withLockedValue {
+            self._stateLock.withLockedValue {
                 $0.remoteAddress = newValue
             }
         }
     }
 
     public var options: [(option: any ChannelOption, value: any Sendable)] {
-        _options
+        self._stateLock.withLockedValue { $0.options }
     }
-
-    @usableFromInline
-    internal var _options: [(option: any ChannelOption, value: any Sendable)] {
-        get {
-            self.stateLock.withLockedValue { $0.options }
-        }
-        set {
-            self.stateLock.withLockedValue { $0.options = newValue }
-        }
-    }
-
     /// Create a new instance.
     ///
     /// During creation it will automatically also register itself on the ``NIOAsyncTestingEventLoop``.
@@ -637,11 +630,15 @@ public final class NIOAsyncTestingChannel: Channel {
     @inlinable
     internal func addOption<Option: ChannelOption>(_ option: Option, value: Option.Value) {
         // override the option if it exists
-        let optionIndex = options.firstIndex(where: { $0.option is Option })
-        if let optionIndex = optionIndex {
-            self._options[optionIndex] = (option, value)
-        } else {
-            self._options.append((option, value))
+        _stateLock.withLockedValue { state in
+            var options = state.options
+            let optionIndex = options.firstIndex(where: { $0.option is Option })
+            if let optionIndex = optionIndex {
+               options[optionIndex] = (option, value)
+            } else {
+                options.append((option, value))
+            }
+            state.options = options
         }
     }
 
--- a/Sources/NIOEmbedded/AsyncTestingChannel.swift
+++ b/Sources/NIOEmbedded/AsyncTestingChannel.swift
@@ -617,7 +617,7 @@ public final class NIOAsyncTestingChannel: Channel {
             return result as! Option.Value
         }
 
-        guard let value = optionValue(for: option) else {
+        guard let value = self.optionValue(for: option) else {
             fatalError("option \(option) not supported")
         }
 
@@ -632,7 +632,7 @@ public final class NIOAsyncTestingChannel: Channel {
     @inlinable
     internal func addOption<Option: ChannelOption>(_ option: Option, value: Option.Value) {
         // override the option if it exists
-        _stateLock.withLockedValue { state in
+        self._stateLock.withLockedValue { state in
             var options = state.options
             let optionIndex = options.firstIndex(where: { $0.option is Option })
             if let optionIndex = optionIndex {
--- a/Sources/NIOEmbedded/AsyncTestingChannel.swift
+++ b/Sources/NIOEmbedded/AsyncTestingChannel.swift
@@ -636,7 +636,7 @@ public final class NIOAsyncTestingChannel: Channel {
             var options = state.options
             let optionIndex = options.firstIndex(where: { $0.option is Option })
             if let optionIndex = optionIndex {
-               options[optionIndex] = (option, value)
+                options[optionIndex] = (option, value)
             } else {
                 options.append((option, value))
             }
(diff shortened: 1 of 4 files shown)


Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -262,6 +263,15 @@ public final class NIOAsyncTestingChannel: Channel {
         }
     }
 
+    public var options: [(option: any ChannelOption, value: any Sendable)] {
```
Lukasa, line 272: This probably doesn't need a public setter, which is good, because this pattern is prone to weird ABA problems.

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -606,7 +616,28 @@ public final class NIOAsyncTestingChannel: Channel {
...
+    }
+
+    @inlinable
+    internal func addOption<Option: ChannelOption>(_ option: Option, value: Option.Value) {
+        // override the option if it exists
+        let optionIndex = options.firstIndex(where: { $0.option is Option })
+        if let optionIndex = optionIndex {
+            self.options[optionIndex] = (option, value)
+        } else {
+            self.options.append((option, value))
```
Lukasa: Let's take the lock here and then do our complete computation under it.

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -788,6 +788,9 @@ public final class EmbeddedChannel: Channel {
     /// - see: `Channel.isWritable`
     public var isWritable: Bool = true
 
+    @usableFromInline
+    internal var options: [(option: any ChannelOption, value: any Sendable)] = []
```
Lukasa: Is it intentional that there aren't any public `getters` for this?

Sources/NIOEmbedded/AsyncTestingChannel.swift
klaaspieter: I keep running into `@inlinable` issues because `getOption` and `setOption` are inlinable. The choices I see are making the property internal with a private setter or have internal storage with a a separate public getter:

# internal getter, private setter

```swift
@usableFromInline
internal private(set) var options: [(option: any ChannelOption, value: any Sendable)] {
```

# public getter, internal setter

```swift

public var options: [(option: any ChannelOption, value: any Sendable)] {
    _options
}

@usableFromInline
internal var _options: [(option: any ChannelOption, value: any Sendable)] {
    get {
        self.stateLock.withLockedValue { $0.options }
    }
    set {
        self.stateLock.withLockedValue { $0.options = newValue }
    }
}
```

I can imagine people wanting to assert the entire set of options so a public getter with an internal setter seems best. Unfortunately it unnecessarily allows other types to change the value.

Sources/NIOEmbedded/AsyncTestingChannel.swift
klaaspieter: I believe that has the same `@inlinable` problem as making the `options` property public. `stateLock` is private and cannot be accessed directly from the inlinable function.

Sources/NIOEmbedded/Embedded.swift
klaaspieter: Nope. I probably ran into the compiler errors with addressing your [AsyncTestingChannel comment](https://github.com/apple/swift-nio/pull/3308#discussion_r2215452213) and decided internal was fine.

That said I can imagine wanting to assert on the complete set of options instead of going through `getOption` so I think it should be public.

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: Oh, I'm saying don't offer a `setter` at all. Instead, take the lock and operate on the state directly. This kind of `get`/`set` pattern on locked objects is almost always a source of bugs, and you should operate directly on the lock instead.

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: That's fine, we can make `stateLock` non-`private`. Make it `@usableFromInline internal` and then prefix its name with `_`.

Sources/NIOEmbedded/Embedded.swift
Lukasa: Agreed :+1:

Sources/NIOEmbedded/AsyncTestingChannel.swift
klaaspieter: Done in https://github.com/apple/swift-nio/pull/3308/commits/20a58e87071c2ffbe861f43b076dbb636ef5b13b

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -606,7 +616,32 @@ public final class NIOAsyncTestingChannel: Channel {
 
             return result as! Option.Value
         }
-        fatalError("option \(option) not supported")
+
+        guard let value = optionValue(for: option) else {
```
Lukasa: Nit: 
```suggestion
        guard let value = self.optionValue(for: option) else {
```

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -606,7 +616,32 @@ public final class NIOAsyncTestingChannel: Channel {
...
+            fatalError("option \(option) not supported")
+        }
+
+        return value
+    }
+
+    @inlinable
+    internal func optionValue<Option: ChannelOption>(for option: Option) -> Option.Value? {
+        self.options.first(where: { $0.option is Option })?.value as? Option.Value
```
Lukasa, line 629: This is probably worth turning into `as!`

Sources/NIOEmbedded/AsyncTestingChannel.swift
```diff
@@ -606,7 +616,32 @@ public final class NIOAsyncTestingChannel: Channel {
...
+    }
+
+    @inlinable
+    internal func addOption<Option: ChannelOption>(_ option: Option, value: Option.Value) {
+        // override the option if it exists
+        _stateLock.withLockedValue { state in
```
Lukasa: Nit:

```suggestion
        self._stateLock.withLockedValue { state in
```

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -788,6 +788,14 @@ public final class EmbeddedChannel: Channel {
     /// - see: `Channel.isWritable`
     public var isWritable: Bool = true
 
+
+    @usableFromInline
+    internal var _options: [(option: any ChannelOption, value: any Sendable)] = []
+
+    /// The `ChannelOption`s set on this channel.
+    /// - see: `Embedded.setOption`
+    public var options: [(option: any ChannelOption, value: any Sendable)] { _options }
```
Lukasa: Nit:

```suggestion
    public var options: [(option: any ChannelOption, value: any Sendable)] { self._options }
```

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -1048,7 +1057,26 @@ public final class EmbeddedChannel: Channel {
...
+        if let optionIndex = self.options.firstIndex(where: { $0.option is Option }) {
+            self._options[optionIndex] = (option: option, value: value)
+        } else {
+            self._options.append((option: option, value: value))
+        }
+    }
+
+    @inlinable
+    internal func optionValue<Option: ChannelOption>(for option: Option) -> Option.Value? {
+        self.options.first(where: { $0.option is Option })?.value as? Option.Value
```
Lukasa, line 1085: This can probably be `as!`

Sources/NIOEmbedded/Embedded.swift
```diff
@@ -1048,7 +1057,26 @@ public final class EmbeddedChannel: Channel {
...
+        guard let value = optionValue(for: option) else {
+            fatalError("option \(option) not supported")
+        }
+
+        return value
+    }
+
+    @inlinable
+    internal func addOption<Option: ChannelOption>(_ option: Option, value: Option.Value) {
+        if let optionIndex = self.options.firstIndex(where: { $0.option is Option }) {
```
Lukasa: Nit: let's consistently use `_options` internally.

Lukasa (commented):
Cool, looking nice. A few tweaks around the edges.

Sources/NIOEmbedded/AsyncTestingChannel.swift
klaaspieter: The way it works now is that it becomes nil and is then caught by `guard let value = self.optionValue` in `getOptionSync` which fatal errors with a more descriptive `option \(option) not supported`. The `fatalError` is what initially made me aware of the underlying issue fixed by this PR and I think it's better than crashing by force unwrapping an optional. What do you think?

Sources/NIOEmbedded/Embedded.swift
klaaspieter: Same as https://github.com/apple/swift-nio/pull/3308#discussion_r2222946597

Sources/NIOEmbedded/AsyncTestingChannel.swift
Lukasa: Gotcha, works for me.

Lukasa:
Mind fixing up the formatting issues that CI has reported?

klaaspieter:
I fixed the formatting. Do you know what's going on with [Linux (nightly-main) failure](https://github.com/apple/swift-nio/actions/runs/16471144748/job/46560489777)? I don't know how to fix that one.

klaaspieter:
@Lukasa anything left for me to do here?

Lukasa:
It looks like we have some tests failing on the nightly branch. Mind cleaning those up?

klaaspieter:
@Lukasa I don't mind but I don't know how to fix them. The errors are `error: call can throw but is not marked with 'try'` or `error: expression is 'async' but is not marked with 'await'` but if I add the `try`/`await` other platforms build with warnings: `No 'async' operations occur within 'await' expression` and `No calls to throwing functions occur within 'try' expression`.

Lukasa:
Interesting. Sounds like this is a bug in nightly Swift. Let me see if I can figure out what's going on.

Lukasa:
Oh, actually that bug has pointed out that this test isn't sound. `EmbeddedChannel` cannot be used in an `async` function like this because it's not actually thread-safe. In principle this can move around threads, which is no good. Can that test be rewritten to not be `async`, and instead use the `wait()` APIs?

klaaspieter:
@Lukasa done!

Lukasa:
For `asyncTestingChannel`, I believe the `await` is actually needed, isn't it?

klaaspieter:
How about bc6fcc972461a2d508e7cb01f8c1242265af3716?

Lukasa (approved):
Very nice, thank you! ✨

PR #3309, opened 2025-07-17
Call channel initializer when using async methods on pipe bootstrap

Lukasa:
Motivation:

For our async initializers on bootstraps, it's imortant that _both_ channel initializers are called, not only the one that comes as an argument on the bind function.

Modifications:

Fix the takeOwnershipOfDescriptor async functions to call the initializer from the bootstrap, and add regression tests.

Result:

Better behaved code.

### DIFF
--- a/Sources/NIOPosix/Bootstrap.swift
+++ b/Sources/NIOPosix/Bootstrap.swift
@@ -2635,9 +2635,9 @@ extension NIOPipeBootstrap {
             return channelOptions.applyAllChannelOptions(to: channel).flatMap {
                 if let bootstrapChannelInitializer {
                     bootstrapChannelInitializer(channel)
-                 } else {
+                } else {
                     channel.eventLoop.makeSucceededVoidFuture()
-                 }
+                }
             }
             .flatMap {
                 _ -> EventLoopFuture<ChannelInitializerResult> in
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
Oops! Nice catch.

Austinpayne:
@Lukasa I'm hitting a precondition failure after bumping from nio 2.84.0 to 2.85.0 and I believe it is due to this change.

I'm adding a `ByteToMessageHandler` in `.channelInitializer` then calling `.takingOwnershipOfDescriptor` (the event loop based API, not the `async` one) and it hits [this precondition in Codec.swift](https://github.com/apple/swift-nio/blob/main/Sources/NIOCore/Codec.swift#L640). I believe this is due to the path through `.takingOwnershipOfDescriptor` calling the channel initializer twice, once [here](https://github.com/apple/swift-nio/blob/main/Sources/NIOPosix/Bootstrap.swift#L2636) and the second time [here](https://github.com/apple/swift-nio/blob/main/Sources/NIOPosix/Bootstrap.swift#L2643) (which for my code path, calls `self.channelInitializer` twice).

I'm happy to submit a reproducer but curious on your thoughts first and what the fix would be. I might be misusing the API to begin with. I'm able to work around it by initializing the channel after taking ownership:

Broken (hits precondition):
Screenshot 2025-07-31 at 3 23 02 PM

Work around:
Screenshot 2025-07-31 at 3 23 22 PM

Lukasa:
Ah, yeah, I see the issue. This new API surface is frustratingly under-tested.

Lukasa:
See #3330.

Issue #3274, opened 2025-06-27
When a WebSocket upgrades fails. No information is provided as to why


adam-fowler:
The `NIOTypedHTTPClientUpgradeConfiguration.notUpgradingCompletionHandler` only includes the channel as a parameter. This makes it very hard to ascertain why a websocket upgrade failed. Ideally the parameters would include the full HTTP response to provide additional information, but including the response header would be a good start.

Lukasa:
This sounds like a reasonable enhancement to me. Are you open to making the change @adam-fowler?

adam-fowler:
> This sounds like a reasonable enhancement to me. Are you open to making the change @adam-fowler?

Why did I know this would be the response 🤣

ali-ahsan-ali:
Hey I am wondering if this is something I could take a look at. Without diving deep into the code at face value it seems simple.

gjcairo:
@ali-ahsan-ali feel free to do so!

PR #3312, opened 2025-07-20
[Issue-3274] - Add Request and Response to notUpgradingHandler

ali-ahsan-ali:
Motivation:

Solve Issue 3274

Modifications:
Allow HTTPResponseHead and HTTPRequestHead to be sent via the notUpgradingCOmpletionHandler

Result:

You can access the request or response when configuring the not upgrading completion handler

### DIFF
--- a/Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
+++ b/Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
@@ -51,12 +51,12 @@ public struct NIOTypedHTTPClientUpgradeConfiguration<UpgradeResult: Sendable> {
 
     /// A closure that is run once it is determined that no protocol upgrade is happening. This can be used
     /// to configure handlers that expect HTTP.
-    public var notUpgradingCompletionHandler: @Sendable (Channel) -> EventLoopFuture<UpgradeResult>
+    public var notUpgradingCompletionHandler: @Sendable (Channel, HTTPResponseHead) -> EventLoopFuture<UpgradeResult>
 
     public init(
         upgradeRequestHead: HTTPRequestHead,
         upgraders: [any NIOTypedHTTPClientProtocolUpgrader<UpgradeResult>],
-        notUpgradingCompletionHandler: @Sendable @escaping (Channel) -> EventLoopFuture<UpgradeResult>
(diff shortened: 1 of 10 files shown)


ali-ahsan-ali:
https://github.com/apple/swift-nio/issues/3274

ali-ahsan-ali:
Feel stuck and unsure how to test this change and if this is even the change we want to solve the issue?

Sources/NIOHTTP1/NIOTypedHTTPClientUpgradeHandler.swift
```diff
@@ -51,12 +51,12 @@ public struct NIOTypedHTTPClientUpgradeConfiguration<UpgradeResult: Sendable> {
 
     /// A closure that is run once it is determined that no protocol upgrade is happening. This can be used
     /// to configure handlers that expect HTTP.
-    public var notUpgradingCompletionHandler: @Sendable (Channel) -> EventLoopFuture<UpgradeResult>
+    public var notUpgradingCompletionHandler: @Sendable (Channel, HTTPResponseHead) -> EventLoopFuture<UpgradeResult>
```
Lukasa, line 54: This isn't the exact shape of the change we'd make. To keep the API compatible, we'd need to have a new property with the new closure API, and then deprecate this one. We can then put them all into the same underlying private property, with just a reabstraction for the original one (which will drop the last argument).

Sources/NIOHTTP1/NIOTypedHTTPClientUpgraderStateMachine.swift
```diff
@@ -160,7 +160,7 @@ struct NIOTypedHTTPClientUpgraderStateMachine<UpgradeResult> {
     @usableFromInline
     enum ChannelReadResponsePartAction {
         case fireErrorCaughtAndRemoveHandler(Error)
-        case runNotUpgradingInitializer
+        case runNotUpgradingInitializer(HTTPResponseHead)
```
Lukasa, line 163: We should probably keep the request head as well.

Lukasa (commented):
The rough shape of this is fine, I left some notes in the diff. For testing, there should already be tests that invoke this code path: you just want to confirm that the request and response heads match the appropriate shape.

ali-ahsan-ali:
Hey, struggling to find the time for this. Happy for someone else to pick it up and use this if it is helpful.

PR #3313, opened 2025-07-21
Move conditional setup logic from bash to actions

madrob:
Instead of checking for a setup command in bash, we use GitHub Actions to determine if there is anything to run

### Motivation:

Slightly cleaner output when looking at logs.

### Modifications:

Moved an if statement from bash to yaml.

### Result:

The setup commands step will generally be skipped instead of executing a no-op


rnro (approved):
Looks good, thanks!

PR #3320, opened 2025-07-25
Fixes all warnings when `-require-explicit-sendable` flag is enabled

rafaelcepeda:
Fixes all warnings when `-require-explicit-sendable` flag is enabled and enables the flag on macOS CI.

### Motivation:

We want to ensure our public API is either explicitly marked as `Sendable` or not.

### Modifications:

Marked appropriate public types as `Sendable`, or explicitly defined their conformance to the `Sendable` protocol as unavailable.

### Result:

We can now enable `-require-explicit-sendable` compiler flag in our codebase.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,3 +74,7 @@ jobs:
       watchos_xcode_test_enabled: true
       tvos_xcode_test_enabled: true
       visionos_xcode_test_enabled: true
+
+  release-builds:
+    name: Release builds
+    uses: apple/swift-nio/.github/workflows/release_builds.yml@main
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -91,3 +91,7 @@ jobs:
     name: Static SDK
     # Workaround https://github.com/nektos/act/issues/1875
     uses: apple/swift-nio/.github/workflows/static_sdk.yml@main
+
+  release-builds:
+    name: Release builds
+    uses: apple/swift-nio/.github/workflows/release_builds.yml@main
--- a/.github/workflows/release_builds.yml
+++ b/.github/workflows/release_builds.yml
@@ -62,8 +62,8 @@ on:
         default: ""
       windows_nightly_next_enabled:
         type: boolean
-        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to true."
-        default: true
+        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to false."
+        default: false
       windows_nightly_next_arguments_override:
         type: string
         description: "The arguments passed to swift test in the Windows matrix job using the nightly build for the next Swift version."
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -102,8 +102,8 @@ on:
         default: ""
       windows_nightly_next_enabled:
         type: boolean
-        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to true."
-        default: true
+        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to false."
+        default: false
       windows_nightly_next_arguments_override:
         type: string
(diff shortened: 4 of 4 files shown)


Sources/NIOConcurrencyHelpers/NIOLock.swift
```diff
@@ -185,6 +185,11 @@ final class LockStorage<Value>: ManagedBuffer<Value, LockPrimitive> {
     }
 }
 
+#if compiler(<6.2)
```
Lukasa, line 190: Why is this compiler-version guarded?

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
```diff
@@ -80,3 +80,6 @@ public struct NIOLockedValueBox<Value> {
 }
 
 extension NIOLockedValueBox: @unchecked Sendable where Value: Sendable {}
+
+@available(*, unavailable)
+extension NIOLockedValueBox.Unsafe: Sendable {}
```
Lukasa: FWIW, the unsafe box _is_ Sendable, so I'm inclined to say that we should probably mark it Sendable to, at least where `Value` is `Sendable`.

Sources/NIOFileSystem/BufferedReader.swift
```diff
@@ -239,3 +239,6 @@ extension ReadableFileHandleProtocol {
         BufferedReader(wrapping: self, initialOffset: initialOffset, capacity: Int(capacity.bytes))
     }
 }
+
+@available(*, unavailable)
+extension BufferedReader: Sendable {}
```
Lukasa: @glbrntt Should this be conditionally `Sendable` based on `Handle`?

Sources/NIOFileSystem/BufferedWriter.swift
```diff
@@ -243,3 +243,6 @@ extension WritableFileHandleProtocol {
         }
     }
 }
+
+@available(*, unavailable)
+extension BufferedWriter: Sendable {}
```
Lukasa: @glbrntt Same question here

Sources/NIOConcurrencyHelpers/NIOLock.swift
rafaelcepeda: Because `ManagedBuffer` is already [declaring](https://github.com/swiftlang/swift/blob/d0cf535664554e5a7c4cd808a5ccb8c7fa6bc2d1/stdlib/public/core/ManagedBuffer.swift#L72) `Sendable` unavailability after 6.1. Without this guard, the CI fail, saying this was already inherited from `ManagedBuffer`.

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
rafaelcepeda: That only works if we use `@unchecked`. Knowing that `LockStorage` is _not_ Sendable, do you still think it is safe for both the Box and the `Unsafe` struct to be `Sendable`, given both hold references to `LockStorage`?

Sources/NIOFileSystem/BufferedReader.swift
glbrntt: Yes

Sources/NIOFileSystem/BufferedWriter.swift
glbrntt: Also a yes, this should be conditionally sendable based on the underlying handle

Sources/NIOConcurrencyHelpers/NIOLock.swift
Lukasa: Can you leave a comment to that effect?

Sources/NIOConcurrencyHelpers/NIOLockedValueBox.swift
Lukasa: Yes, both are absolutely intended to be `Sendable`. The storage is _not_ because it's not the API type and using it correctly requires constraints outside the storage type, but this type does use it correctly.

rafaelcepeda: Got it, thank you.

PR #3323, opened 2025-07-28
Fix typo in `HTTPServerClientTest.swift` comments

MaxDesiatov:
`we read faster then we can write` -> `we read faster than we can write`


glbrntt (approved):
Thanks Max!

PR #3324, opened 2025-07-28
[Windows] IOVector is WSABUF on Windows

fabianfett:
On Windows there is no iovec, because of this we need to make sure, that:

1. We use the typealias `IOVector` as much as possible
2. We add platform specific code to `IOVector` extensions

Also adds some more Windows imports to fix some compiler errors.

### DIFF
--- a/Sources/NIOPosix/BSDSocketAPIWindows.swift
+++ b/Sources/NIOPosix/BSDSocketAPIWindows.swift
@@ -703,7 +703,7 @@ extension NIOBSDSocket {
 }
 
 extension IOVector {
-    // An initializer thats maps `ByteBuffer`` derived pointers to WSABUF. This allows us to use the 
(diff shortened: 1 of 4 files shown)


Sources/NIOPosix/BSDSocketAPIWindows.swift
```diff
@@ -701,4 +701,13 @@ extension NIOBSDSocket {
         throw ChannelError.operationUnsupported
     }
 }
+
+extension IOVector {
+    init(iov_base: UnsafeMutableRawPointer, iov_len: UInt32) {
+        self = WSABUF(
+            len: iov_len,
+            buf: iov_base.assumingMemoryBound(to: Int8.self)
```
Lukasa, line 711: Can we add a documentation comment that we only use this with `ByteBuffer`-derived pointers? Otherwise the assumption here is not guaranteed to be true.

Lukasa (approved):

PR #3325, opened 2025-07-29
Do not run windows nightly CI by default

josephnoir:
### Motivation:

Not everything that relies on these workflows builds on windows. Unit tests and release builds should be explicitly enabled for windows builds.

### Modifications:

Change default of windows nightly to false. Fix flag checks that enable nightly builds.

### Result:

Easier to use CI workflows.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -102,8 +102,8 @@ on:
         default: ""
       windows_nightly_next_enabled:
         type: boolean
-        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to true."
-        default: true
+        description: "Boolean to enable the Windows matrix job using the nightly build for the next Swift version. Defaults to false."
+        default: false
       windows_nightly_next_arguments_override:
         type: string
         description: "The arguments passed to swift test in the Windows matrix job using the nightly build for the next Swift version."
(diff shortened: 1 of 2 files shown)


josephnoir:
I shimmed this branch into two PRs to check the expected behavior:

* [This PR on swift crypto](https://github.com/apple/swift-crypto/pull/388) still correctly builds nightlies for the unit test workflows.
* [This PR on swift-statsd-client](https://github.com/apple/swift-statsd-client/pull/55) now has windows nightly release builds disabled, which were falsely enabled before.

.github/workflows/unit_tests.yml
```diff
@@ -141,7 +141,7 @@ jobs:
           MATRIX_LINUX_6_0_COMMAND_ARGUMENTS: ${{ inputs.linux_6_0_arguments_override }}
           MATRIX_LINUX_6_1_ENABLED: ${{ inputs.linux_6_1_enabled }}
           MATRIX_LINUX_6_1_COMMAND_ARGUMENTS: ${{ inputs.linux_6_1_arguments_override }}
-          MATRIX_LINUX_NIGHTLY_NEXT_ENABLED: ${{ inputs.linux_nightly_6_1_enabled && inputs.linux_nightly_next_enabled }}
+          MATRIX_LINUX_NIGHTLY_NEXT_ENABLED: ${{ inputs.linux_nightly_6_1_enabled || inputs.linux_nightly_next_enabled }}
```
FranzBusch, line 144: Why is this checking for 6.1 in the first place? Same comment for the one below

glbrntt (approved):

.github/workflows/unit_tests.yml
josephnoir: The nightly builds used to be numbered (here 6.1). The new flag does not have that number, but the old flags are still around and marked as deprecated.

PR #3159, opened 2025-03-27
CI: Add build check for WebAssembly SDK

kateinoigakukun:
_[One line description of your change]_

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_

PR #3331, opened 2025-08-02
[wasm][1/3] Add reusable WebAssembly workflow

kateinoigakukun:
Add build check CI for WebAssembly SDK

### Motivation:

Add build check CI for WebAssembly SDK

### Modifications:

Add only reusable GHA workflow from https://github.com/apple/swift-nio/pull/3159 to gracefully merge changes without breaking CI.

After this change being merged into main branch, we can merge the actual PR checks

### Result:

Setting up the CI job makes it easier to detect such incompatibility before merging.

### DIFF
--- a/.github/workflows/wasm_sdk.yml
+++ b/.github/workflows/wasm_sdk.yml
@@ -1,0 +1,27 @@
+name: WebAssembly SDK
+
+on:
+  workflow_call:
+
+jobs:
+  wasm-sdk:
+    name: WebAssembly Swift SDK
+    # Workaround https://github.com/nektos/act/issues/1875
+    uses: apple/swift-nio/.github/workflows/swift_test_matrix.yml@main
+    with:
+      name: "WebAssembly Swift SDK"
+      matrix_string: >-
+        {
+          "config":[
+            {
+              "name":"main Jammy",
+              "swift_version":"main",
+              "platform":"Linux",
+              "runner":"ubuntu-latest",
+              "image":"ubuntu:jammy",
+              "setup_command":"apt update -q && apt install -y -q curl jq tar && curl -s --retry 3 https://raw.githubusercontent.com/apple/swift-nio/main/scripts/install_swift_prerequisites.sh | bash && curl -s --retry 3 https://raw.githubusercontent.com/apple/swift-nio/main/scripts/install_swift_sdk.sh | INSTALL_SWIFT_BRANCH=main INSTALL_SWIFT_ARCH=x86_64 INSTALL_SWIFT_SDK=wasm-sdk bash && hash -r",
+              "command":"curl -s --retry 3 https://raw.githubusercontent.com/apple/swift-nio/main/scripts/swift-build-with-wasm-sdk.sh | bash -s --",
+              "command_arguments":"--target NIOCore"
+            }
+          ]
+        }
--- a/scripts/install_swift_sdk.sh
+++ b/scripts/install_swift_sdk.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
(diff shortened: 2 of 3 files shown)


scripts/install_swift_sdk.sh
kateinoigakukun, line 1: This script was originally copied from the existing `./scripts/install_static_sdk.sh` and modified to support Wasm SDK. `./scripts/install_static_sdk.sh` is left as is in this PR to keep the CI passing. If we were to remove the existing one in this PR and modify `.github/workflows/static_sdk.yml` to point to the new name, `.github/workflows/static_sdk.yml` will fail on this PR because the file doesn't exist on the main branch yet.

MaxDesiatov:
This is potentially superseded by https://github.com/swiftlang/github-workflows/pull/142

kateinoigakukun:
We still need this separate PR to add a reusable workflow based on https://github.com/swiftlang/github-workflows/pull/142

FranzBusch:
> This is potentially superseded by [swiftlang/github-workflows#142](https://github.com/swiftlang/github-workflows/pull/142)

This is not suppressed by it because NIO and all other server repos are not using the workflows provided by the linked repo due to their limited capabilities.

.github/workflows/wasm_sdk.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  wasm-sdk:
+    name: WebAssembly Swift SDK
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    with:
+      enable_wasm_sdk_build: true
+      enable_linux_checks: false
+      enable_windows_checks: false
+      swift_flags: --target NIOCore
+      swift_nightly_flags: --target NIOCore
```
kkebo, line 15: FYI: Thanks to https://github.com/swiftlang/github-workflows/pull/149, these two lines can be shortened into one line.


```suggestion
      wasm_sdk_build_command: swift build --target NIOCore
```

MaxDesiatov: IIRC `swift_nightly_flags` in absence of a defined value is auto-populated from `swift_flags`, so the workflow here should just specify the latter without a need for `wasm_sdk_build_command`.

kkebo: @MaxDesiatov That's incorrect. `swift_nightly_flags` is not inherited from `swift_flags`.

https://github.com/swiftlang/github-workflows/blob/1f5e7cde8ba1fd63b9ec36f624453a6fc4420043/.github/workflows/swift_package_test.yml#L330

MaxDesiatov: I may be missing something, but there's `|| inputs.swift_flags` on the line you've linked to?

kkebo: Yes, I linked to the following line.

```yaml
BUILD_FLAGS: ${{ (contains(matrix.swift_version, 'nightly') && inputs.swift_nightly_flags) || inputs.swift_flags }}
```

kkebo: This is a ternary operator.

MaxDesiatov: I don't think so. If `swift_nightly_flags` is undefined, the first operand of `||` is falsy, which then evaluates to the second operand `swift_flags`. You can also see `swift_flags` respected for nightlies in this workflow, where `swift_nightly_flags` is not defined at all https://github.com/swiftlang/swift-for-wasm-examples/blob/main/.github/workflows/pull_request.yml

kkebo: Sorry, I was wrong. I didn't expect that the empty string was falsy. Thank you.

PR #3322, opened 2025-07-28
Add a new file path type, NIOFilePath, backed by SystemPackage's FilePath type

aryan-25:
Motivation:

Two separate types for representing file paths currently exist: [`SystemPackage`'s `FilePath`](https://swiftpackageindex.com/apple/swift-system/1.6.0/documentation/systempackage/filepath) type and [`System`'s `FilePath`](https://developer.apple.com/documentation/system/filepath) type.

`NIOFileSystem` currently uses `SystemPackage`'s `FilePath`. However, the lack of an API for converting between the two representations means that users whose application also uses `System.FilePath` may find interacting with `NIOFileSystem` difficult.

Modifications:

- Added `NIOFilePath`, and its subcomponents `NIOFilePath.Root`, `NIOFilePath.Component`, and `NIOFilePath.ComponentView`.
- These types mirror the API of `SystemPackage`'s `FilePath` and are internally backed by it.
- `NIOFilePath` can be initialized from either an instance of `SystemPackage.FilePath` or `System.FilePath`.

Result:

Reduces the friction of interacting with `NIOFileSystem` by providing a bridge between `SystemPackage.FilePath` and `System.FilePath`.

PR #3333, opened 2025-08-04
Update FilePath references to NIOFilePath

aryan-25:
Motivation:

Currently, file paths are represented by the `SystemPackage.FilePath` type in `NIOFileSystem`. Following from #3322, we want to use `NIOFilePath` instead.

Modifications:

This change provides `NIOFilePath` accepting/returning variants to all existing public methods that accept/return `SystemPackage.FilePath`.

The bulk of the changes are in `FileSystemProtocol` and `FileHandleProtocol`, and their conforming types. The protocols have been modified to require `NIOFilePath` accepting/returning methods. Default implementations are provided for `SystemPackage.FilePath` accepting/returning methods---these call through to the `NIOFilePath` variants.

Result:

`NIOFilePath` path representations can now be used with the methods for interacting with the file system.

### DIFF
--- a/Sources/NIOFileSystem/Array+FileSystem.swift
+++ b/Sources/NIOFileSystem/Array+FileSystem.swift
@@ -23,9 +23,8 @@ extension Array where Element == UInt8 {
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
     ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
-    @_disfavoredOverload
     public init(
-        contentsOf path: FilePath,
+        contentsOf path: NIOFilePath,
         maximumSizeAllowed: ByteCount,
         fileSystem: some FileSystemProtocol
     ) async throws {
@@ -42,12 +41,13 @@ extension Array where Element == UInt8 {
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
     ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    @_disfavoredOverload
     public init(
-        contentsOf path: NIOFilePath,
+        contentsOf path: FilePath,
         maximumSizeAllowed: ByteCount,
         fileSystem: some FileSystemProtocol
     ) async throws {
-        try await self.init(contentsOf: path.underlying, maximumSizeAllowed: maximumSizeAllowed, fileSystem: fileSystem)
+        try await self.init(contentsOf: .init(path), maximumSizeAllowed: maximumSizeAllowed, fileSystem: fileSystem)
     }
 
     /// Reads the contents of the file at the path using ``FileSystem``.
@@ -55,9 +55,8 @@ extension Array where Element == UInt8 {
     /// - Parameters:
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, as a ``ByteCount``.
-    @_disfavoredOverload
     public init(
-        contentsOf path: FilePath,
+        contentsOf path: NIOFilePath,
         maximumSizeAllowed: ByteCount
     ) async throws {
         self = try await Self(
@@ -72,11 +71,12 @@ extension Array where Element == UInt8 {
     /// - Parameters:
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, as a ``ByteCount``.
+    @_disfavoredOverload
     public init(
-        contentsOf path: NIOFilePath,
+        contentsOf path: FilePath,
         maximumSizeAllowed: ByteCount
     ) async throws {
-        try await self.init(contentsOf: path.underlying, maximumSizeAllowed: maximumSizeAllowed)
+        try await self.init(contentsOf: .init(path), maximumSizeAllowed: maximumSizeAllowed)
     }
 }
 #endif
--- a/Sources/NIOFileSystem/Array+FileSystem.swift
+++ b/Sources/NIOFileSystem/Array+FileSystem.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2024 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -23,6 +23,7 @@ extension Array where Element == UInt8 {
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
     ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    @_disfavoredOverload
     public init(
         contentsOf path: FilePath,
         maximumSizeAllowed: ByteCount,
@@ -35,11 +36,26 @@ extension Array where Element == UInt8 {
         self = Self(buffer: byteBuffer)
     }
 
+    /// Reads the contents of the file at the path.
+    ///
+    /// - Parameters:
+    ///   - path: The path of the file to read.
+    ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
+    ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    public init(
+        contentsOf path: NIOFilePath,
+        maximumSizeAllowed: ByteCount,
+        fileSystem: some FileSystemProtocol
+    ) async throws {
+        try await self.init(contentsOf: path.underlying, maximumSizeAllowed: maximumSizeAllowed, fileSystem: fileSystem)
+    }
+
     /// Reads the contents of the file at the path using ``FileSystem``.
     ///
     /// - Parameters:
     ///   - path: The path of the file to read.
     ///   - maximumSizeAllowed: The maximum size of file which can be read, as a ``ByteCount``.
+    @_disfavoredOverload
     public init(
         contentsOf path: FilePath,
         maximumSizeAllowed: ByteCount
@@ -49,6 +65,18 @@ extension Array where Element == UInt8 {
             maximumSizeAllowed: maximumSizeAllowed,
             fileSystem: .shared
         )
+    }
+
+    /// Reads the contents of the file at the path using ``FileSystem``.
+    ///
+    /// - Parameters:
+    ///   - path: The path of the file to read.
+    ///   - maximumSizeAllowed: The maximum size of file which can be read, as a ``ByteCount``.
+    public init(
+        contentsOf path: NIOFilePath,
+        maximumSizeAllowed: ByteCount
+    ) async throws {
+        try await self.init(contentsOf: path.underlying, maximumSizeAllowed: maximumSizeAllowed)
     }
 }
 #endif
--- a/Sources/NIOFileSystem/DirectoryEntries.swift
+++ b/Sources/NIOFileSystem/DirectoryEntries.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -361,7 +361,7 @@ private struct DirectoryEnumerator: Sendable {
     internal var state: State
 
     /// The path to the directory being enumerated.
-    private let path: FilePath
+    private let path: NIOFilePath
 
     /// The route via which directory entry batches are yielded,
     /// the sourcing end of the `DirectoryEntrySequenceProducer`
@@ -560,7 +560,7 @@ private struct DirectoryEnumerator: Sendable {
                 name = FilePath.Component(platformString: CNIOLinux_dirent_dname(entry))!
                 #endif
 
-                let fullPath = self.path.appending(name)
+                let fullPath = NIOFilePath(self.path.underlying.appending(name))
                 // '!' is okay here: the init returns nil if there is an empty path which we know
                 // isn't the case as 'self.path' is non-empty.
                 entries.append(DirectoryEntry(path: fullPath, type: fileType)!)
--- a/Sources/NIOFileSystem/DirectoryEntries.swift
+++ b/Sources/NIOFileSystem/DirectoryEntries.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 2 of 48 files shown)


Sources/NIOFileSystem/Array+FileSystem.swift
```diff
@@ -35,11 +36,26 @@ extension Array where Element == UInt8 {
...
+    /// - Parameters:
+    ///   - path: The path of the file to read.
+    ///   - maximumSizeAllowed: The maximum size of file which can be read, in bytes, as a ``ByteCount``.
+    ///   - fileSystem: The ``FileSystemProtocol`` instance to use to read the file.
+    public init(
+        contentsOf path: NIOFilePath,
+        maximumSizeAllowed: ByteCount,
+        fileSystem: some FileSystemProtocol
+    ) async throws {
+        try await self.init(contentsOf: path.underlying, maximumSizeAllowed: maximumSizeAllowed, fileSystem: fileSystem)
```
glbrntt: The `NIOFilePath` variant should be the canonical version of the function and the `FilePath` version should call through to it. You'll need to make this change in a bunch of places.

Sources/NIOFileSystem/DirectoryEntries.swift
```diff
@@ -361,7 +361,7 @@ private struct DirectoryEnumerator: Sendable {
     internal var state: State
 
     /// The path to the directory being enumerated.
-    private let path: FilePath
+    private let path: NIOFilePath
```
glbrntt: This is an internal implementation detail so `FilePath` is okay here

Sources/NIOFileSystem/DirectoryEntry.swift
```diff
@@ -21,6 +21,10 @@ public struct DirectoryEntry: Sendable, Hashable, Equatable {
     /// - Precondition: The path must have at least one component.
     public let path: FilePath
 
+    public var filePath: NIOFilePath {
```
glbrntt: Docs please! Also, I think we should store the `NIOFilePath` and have a computed property for `path`.

Sources/NIOFileSystem/FileSystem.swift
```diff
@@ -668,6 +671,38 @@ public struct FileSystem: Sendable, FileSystemProtocol {
...
+    ///   accessed.
+    /// - On Android this returns "/data/local/tmp".
+    /// - On other platforms this returns "/tmp".
+    ///
+    /// - Returns: The path to a temporary directory.
+    public var temporaryDirectoryNIO: NIOFilePath {
+        get async throws {
+            NIOFilePath(try await self.temporaryDirectory)
+        }
+    }
```
glbrntt: I think it's okay to just change the type for these two and not have a `FilePath` variant.

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -108,9 +108,15 @@ public protocol FileSystemProtocol: Sendable {
...
     var currentWorkingDirectory: FilePath { get async throws }
 
+    /// Returns the current working directory.
+    var currentWorkingDirectoryNIO: NIOFilePath { get async throws }
+
     /// Returns the path of the temporary directory.
     var temporaryDirectory: FilePath { get async throws }
 
+    /// Returns the path of the temporary directory.
+    var temporaryDirectoryNIO: NIOFilePath { get async throws }
```
glbrntt: Let's not add the `NIO` variants, we can just change the return type from `FilePath` to `NIOFilePath`.

Sources/NIOFileSystem/Internal/System Calls/Syscall.swift
```diff
@@ -31,8 +31,8 @@ import CNIOLinux
 @_spi(Testing)
 public enum Syscall: Sendable {
     @_spi(Testing)
-    public static func stat(path: FilePath) -> Result<CInterop.Stat, Errno> {
-        path.withPlatformString { platformPath in
+    public static func stat(path: NIOFilePath) -> Result<CInterop.Stat, Errno> {
```
glbrntt: I don't think you really needed to change any of these. The purpose of `NIOFilePath` is to be at the API layer. Internally `FilePath` is an implementation detail so the simplest strategy would be to just drop down to `FilePath` as soon as possible.

Note also that these are `@_spi(Testing)` so not actually `public` API.

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
```diff
@@ -38,14 +38,14 @@ public final class SystemFileHandle: Sendable {
     internal var threadPool: NIOThreadPool { self.sendableView.threadPool }
 
     /// The path used to open this handle.
-    internal var path: FilePath { self.sendableView.path }
+    internal var path: NIOFilePath { self.sendableView.path }
```
glbrntt: Same here; there are a lot of changes in here which aren't necessary, `FilePath` is fine as an implementation detail.

glbrntt (changes requested):

Sources/NIOFileSystem/FileSystem.swift
aryan-25: Agreed. I have removed these two properties now and instead changed the type of `currentWorkingDirectory` and `temporaryWorkingDirectory` to `NIOFilePath`.

Sources/NIOFileSystem/Internal/SystemFileHandle.swift
aryan-25: Agreed. I have kept `NIOFilePath` only in the public layers now.

`SystemFileHandle` (which is annotated with `@_spi(Testing)`) has `NIOFilePath` overloads so that it can continue conforming to `DirectoryFileHandleProtocol`.

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -323,6 +325,29 @@ extension FileSystemProtocol {
     ///       automatically after the closure exits.
     /// - Important: The handle passed to `execute` must not escape the closure.
     /// - Returns: The result of the `execute` closure.
+    public func withFileHandle<Result>(
+        forReadingAt path: NIOFilePath,
+        options: OpenOptions.Read = OpenOptions.Read(),
+        execute: (_ read: ReadFileHandle) async throws -> Result
+    ) async throws -> Result {
+        try await self.withFileHandle(forReadingAt: .init(path), options: options, execute: execute)
```
glbrntt: This one should be in terms of the `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -350,6 +375,29 @@ extension FileSystemProtocol {
     ///       automatically after the closure exits.
     /// - Important: The handle passed to `execute` must not escape the closure.
     /// - Returns: The result of the `execute` closure.
+    public func withFileHandle<Result>(
+        forWritingAt path: NIOFilePath,
+        options: OpenOptions.Write = .newFile(replaceExisting: false),
+        execute: (_ write: WriteFileHandle) async throws -> Result
+    ) async throws -> Result {
+        try await self.withFileHandle(forWritingAt: .init(path), options: options, execute: execute)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -372,8 +420,30 @@ extension FileSystemProtocol {
...
+    ///   - execute: A closure which provides access to the open file. The file is closed
+    ///       automatically after the closure exits.
+    /// - Important: The handle passed to `execute` must not escape the closure.
+    /// - Returns: The result of the `execute` closure.
+    public func withFileHandle<Result>(
+        forReadingAndWritingAt path: NIOFilePath,
+        options: OpenOptions.Write = .newFile(replaceExisting: false),
+        execute: (_ readWrite: ReadWriteFileHandle) async throws -> Result
+    ) async throws -> Result {
+        try await self.withFileHandle(forReadingAndWritingAt: .init(path), options: options, execute: execute)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -403,6 +474,23 @@ extension FileSystemProtocol {
     ///   - execute: A closure which provides access to the directory.
     /// - Important: The handle passed to `execute` must not escape the closure.
     /// - Returns: The result of the `execute` closure.
+    public func withDirectoryHandle<Result>(
+        atPath path: NIOFilePath,
+        options: OpenOptions.Directory = OpenOptions.Directory(),
+        execute: (_ directory: DirectoryFileHandle) async throws -> Result
+    ) async throws -> Result {
+        try await self.withDirectoryHandle(atPath: .init(path), options: options, execute: execute)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -427,6 +515,21 @@ extension FileSystemProtocol {
     /// - Parameters:
     ///   - path: The path of the file to open.
     /// - Returns: A readable handle to the opened file.
+    public func openFile(
+        forReadingAt path: NIOFilePath
+    ) async throws -> ReadFileHandle {
+        try await self.openFile(forReadingAt: .init(path))
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -441,6 +544,21 @@ extension FileSystemProtocol {
     /// - Parameters:
     ///   - path: The path of the directory to open.
     /// - Returns: A handle to the opened directory.
+    public func openDirectory(
+        atPath path: NIOFilePath
+    ) async throws -> DirectoryFileHandle {
+        try await self.openDirectory(atPath: .init(path))
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -454,6 +572,18 @@ extension FileSystemProtocol {
     /// - Parameters:
     ///    - path: The path to get information about.
     /// - Returns: Information about the file at the given path or `nil` if no file exists.
+    public func info(forFileAt path: NIOFilePath) async throws -> FileInfo? {
+        try await self.info(forFileAt: .init(path))
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -475,6 +605,32 @@ extension FileSystemProtocol {
     ///   - sourcePath: The path to the item to copy.
     ///   - destinationPath: The path at which to place the copy.
     ///   - copyStrategy: This controls the concurrency used if the file at `sourcePath` is a directory.
+    public func copyItem(
+        at sourcePath: NIOFilePath,
+        to destinationPath: NIOFilePath,
+        strategy copyStrategy: CopyStrategy = .platformDefault
+    ) async throws {
+        try await self.copyItem(at: .init(sourcePath), to: .init(destinationPath), strategy: copyStrategy)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -603,6 +813,29 @@ extension FileSystemProtocol {
     ///   - path: The path to delete.
     /// - Returns: The number of deleted items which may be zero if `path` did not exist.
     @discardableResult
+    public func removeItem(
+        at path: NIOFilePath
+    ) async throws -> Int {
+        try await self.removeItem(at: .init(path))
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -629,6 +862,34 @@ extension FileSystemProtocol {
     ///       being removed isn't a directory.
     /// - Returns: The number of deleted items which may be zero if `path` did not exist.
     @discardableResult
+    public func removeItem(
+        at path: NIOFilePath,
+        recursively removeItemRecursively: Bool
+    ) async throws -> Int {
+        try await self.removeItem(at: .init(path), recursively: removeItemRecursively)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -651,6 +912,29 @@ extension FileSystemProtocol {
     ///       concurrent scan of the tree at `path` and delete files when they are found.
     /// - Returns: The number of deleted items which may be zero if `path` did not exist.
     @discardableResult
+    public func removeItem(
+        at path: NIOFilePath,
+        strategy removalStrategy: RemovalStrategy
+    ) async throws -> Int {
+        try await self.removeItem(at: .init(path), strategy: removalStrategy)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -672,6 +956,28 @@ extension FileSystemProtocol {
     /// - Parameters:
     ///   - path: The directory to create.
     ///   - createIntermediateDirectories: Whether intermediate directories should be created.
+    public func createDirectory(
+        at path: NIOFilePath,
+        withIntermediateDirectories createIntermediateDirectories: Bool
+    ) async throws {
+        try await self.createDirectory(at: .init(path), withIntermediateDirectories: createIntermediateDirectories)
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -696,6 +1002,32 @@ extension FileSystemProtocol {
...
     ///   - execute: A closure which provides access to the directory and its path.
     /// - Returns: The result of `execute`.
+    public func withTemporaryDirectory<Result>(
+        prefix: NIOFilePath? = nil,
+        options: OpenOptions.Directory = OpenOptions.Directory(),
+        execute: (_ directory: DirectoryFileHandle, _ path: NIOFilePath) async throws -> Result
+    ) async throws -> Result {
+        try await self.withTemporaryDirectory(prefix: prefix.map { FilePath($0) }, options: options) {
+            directory,
+            path in
```
glbrntt: `swift-format` makes bad choices sometimes:

```suggestion
        try await self.withTemporaryDirectory(
            prefix: prefix.map { FilePath($0) }, 
            options: options
        ) { directory, path in
```

Sources/NIOFileSystem/FileSystemProtocol.swift
```diff
@@ -696,6 +1002,32 @@ extension FileSystemProtocol {
...
+            try await execute(directory, .init(path))
```
glbrntt: Same here, should be in terms of `NIOFilePath`

Tests/NIOFileSystemIntegrationTests/FileHandleTests.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5: Looks like none of the changes in this file should be necessary, we can use the `NIOFilePath` versions

Tests/NIOFileSystemIntegrationTests/FileSystemTests.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftNIO open source project
 //
-// Copyright (c) 2023 Apple Inc. and the SwiftNIO project authors
+// Copyright (c) 2025 Apple Inc. and the SwiftNIO project authors
```
glbrntt, line 5: Same for this file, we should be able to use the `NIOFilePath` versions

glbrntt:
The API breakage is expected: we changed a bunch of protocol requirements but added compatibility shims. This is okay, `NIOFileSystem` doesn't yet have a stable API.

PR #3334, opened 2025-08-06
Add Android imports for some new tests in #3308 that require it

finagolfin:
My Android CI [just broke because it couldn't find these newly used C symbols in the NIO tests](https://github.com/finagolfin/swift-android-sdk/actions/runs/16768316992/job/47477878652#step:19:242), this patch [fixed it again](https://github.com/finagolfin/swift-android-sdk/actions/runs/16787226279/job/47540691932). Unsure why this only affects Android, likely the module leaks postulated before.


glbrntt (approved):
Thanks!

PR #3336, opened 2025-08-11
[Windows] Add msghdr extension for Windows

fabianfett:
To be able to compile on Windows eventually, we need to bridge msghdr and WSAMSG. This PR tries to accomplish this by using extensions on those types.

### DIFF
--- a/Sources/NIOPosix/BSDSocketAPIPosix.swift
+++ b/Sources/NIOPosix/BSDSocketAPIPosix.swift
@@ -398,4 +398,16 @@ extension NIOBSDSocket {
         #endif
     }
 }
+
+extension msghdr {
+    var control_ptr: UnsafeMutableRawBufferPointer {
+        set {
+            self.msg_control = newValue.baseAddress
+            self.msg_controllen = numericCast(newValue.count)
+        }
+        get {
+            UnsafeMutableRawBufferPointer(start: self.msg_control, count: numericCast(self.msg_controllen))
+        }
+    }
+}
 #endif
--- a/Sources/NIOPosix/BSDSocketAPIPosix.swift
+++ b/Sources/NIOPosix/BSDSocketAPIPosix.swift
@@ -406,7 +406,7 @@ extension msghdr {
             self.msg_controllen = numericCast(newValue.count)
         }
         get {
-            UnsafeMutableRawBufferPointer(start: self.msg_control, count: numericCast(self.msg_controllen))
+            UnsafeMutableRawBufferPointer(start: self.msg_control, count: Int(self.msg_controllen))
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Sources/NIOPosix/BSDSocketAPIPosix.swift
```diff
@@ -398,4 +398,16 @@ extension NIOBSDSocket {
...
 }
+
+extension msghdr {
+    var control_ptr: UnsafeMutableRawBufferPointer {
+        set {
+            self.msg_control = newValue.baseAddress
+            self.msg_controllen = numericCast(newValue.count)
+        }
+        get {
+            UnsafeMutableRawBufferPointer(start: self.msg_control, count: numericCast(self.msg_controllen))
```
Lukasa: No need for `numericCast` here, this can just be `Int`

Sources/NIOPosix/BSDSocketAPIWindows.swift
```diff
@@ -712,4 +712,61 @@ extension IOVector {
...
+        }
+    }
+
+    var control_ptr: UnsafeMutableRawBufferPointer {
+        set {
+            self.Control.buf = newValue.baseAddress?.bindMemory(to: CHAR.self, capacity: newValue.count)
+            self.Control.len = numericCast(newValue.count)
+        }
+        get {
+            UnsafeMutableRawBufferPointer(start: self.Control.buf, count: numericCast(self.Control.len))
```
Lukasa: Same note here on `numericCast`

Sources/NIOPosix/BSDSocketAPIWindows.swift
```diff
@@ -712,4 +712,61 @@ extension IOVector {
...
+        }
+    }
+
+    var msg_flags: Int {
+        set {
+            self.dwFlags = numericCast(newValue)
+        }
+        get {
+            numericCast(self.dwFlags)
```
Lukasa: Same note here on `numericCast`

Sources/NIOPosix/BSDSocketAPIPosix.swift
fabianfett: Am I right in assuming, that we should use `Int` directly, if the conversion will always succeed?

Lukasa: No, you can use `Int` directly whenever you know you're transforming things to `Int`. The reason to use `numericCast` is if there is a context in which the target type is not necessarily definitively statically known, because you don't have to spell it out. But `numericCast` [is not a complex function:](https://github.com/swiftlang/swift/blob/9d7d3a15cb66e8db4b8f877a15c51c128bb6bc3e/stdlib/public/core/Integers.swift#L3354-L3357)

```swift
@inlinable
public func numericCast<T: BinaryInteger, U: BinaryInteger>(_ x: T) -> U {
  return U(x)
}
```

In cases where we know what `U` is, statically, in all configurations, we should save ourselves one layer of genericism and just write down the type of U instead. Here, that's `Int`.

PR #3339, opened 2025-08-12
Drop all server response parts after an invalid header

### LINKED ISSUE #3326: `NIOHTTPResponseHeadersValidator` should reject further writes are invalid headers
### Expected behavior
`NIOHTTPResponseHeadersValidator` to drop further writes when it detects an invalid header

### Actual behavior
Further writes occur. Because the head is not valid, it's not written, so can result in just the `end` being written for a response which hits an assertion in `HTTPServerPipelineHandler` making it hard to use

### Steps to reproduce

Create a pipeline and send a response with invalid headers

_[anything to help us reproducing the issue]_

2.84.0

macOS 15.5

swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0

glbrntt:
Motivation:

NIOHTTPResponseHeadersValidator drops head/end response parts if they contain invalid header fields. If a head part is dropped and the server allows pipelining then the subsequent body/end part will reach the pipelining handler and result in an assertion failure in debug builds. In release builds, some parts may be incorrectly written out leading to protocol violations.

Modifications:

- Drop all response parts after an invalid response part

Result:

- Fewer bugs
- Resolves #3326

### DIFF
--- a/Sources/NIOHTTP1/HTTPHeaderValidator.swift
+++ b/Sources/NIOHTTP1/HTTPHeaderValidator.swift
@@ -63,27 +63,46 @@ public final class NIOHTTPResponseHeadersValidator: ChannelOutboundHandler, Remo
     public typealias OutboundIn = HTTPServerResponsePart
     public typealias OutboundOut = HTTPServerResponsePart
 
-    public init() {}
+    private enum State {
+        /// Validating response parts.
+        case validating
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice one @glbrntt, thanks.

PR #3340, opened 2025-08-12
Fix a formatting issue

glbrntt:
Motivation:

The pre-release 6.2 formatter picks up an issue that old formatters don't: an extra space between a variable and the method being called on it.

Modifications:

- Fix the formatting issue

Result:

Better formatting


glbrntt:
Ugh... old formatters format this badly.

PR #3343, opened 2025-08-12
[Windows] Map naming conventions Windows -> Posix (ssize_t)

fabianfett:
`ssize_t` doesn't exist in Windows land. Windows loves uppercase naming. Therefore we need a typealias.

### DIFF
--- a/Sources/NIOPosix/ControlMessage.swift
+++ b/Sources/NIOPosix/ControlMessage.swift
@@ -21,6 +21,14 @@ import CNIOLinux
 import CNIOWindows
 #endif
 
+#if os(Windows)
+private let _IPPROTO_IP = WinSDK.IPPROTO_IPV4.rawValue
+private let _IPPROTO_IPV6 = WinSDK.IPPROTO_IPV6.rawValue
+#else
+private let _IPPROTO_IP = IPPROTO_IP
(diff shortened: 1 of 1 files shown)


Sources/CNIOWindows/include/CNIOWindows.h
```diff
@@ -24,6 +24,9 @@
 
 #define NIO(name) CNIOWindows_ ## name
 
+// Map Windows naming conventions to Unix/Posix naming conventions.
+typedef SSIZE_T ssize_t;
```
Lukasa: This is a dangerous typedef to offer. Can we do this at the Swift layer, where the name will be namespaced?

fabianfett: Would you mind expanding why this is dangerous? I'm here to learn.

Lukasa: Because this isn't implemented with `@_implementationOnly`, this header file is included by anything that imports NIO. The result is that this typedef is visible to all users of NIO, and can cause horrifying clashes if someone else also does this. This is why everything else tends to be namespaced under CNIOWindows.

fabianfett: Thanks for providing this context!

PR #3345, opened 2025-08-13
[Windows] Fix deprecation warnings in NIOCore

fabianfett:
- `getenv` and `strerror` are deprecated in Windows land
- replace with undeprecated alternatives
- result: less yellow in the build log


Sources/NIOCore/GlobalSingletons.swift
```diff
@@ -184,7 +184,22 @@ extension NIOSingletons {
...
+            withUnsafeMutablePointer(to: &ptr) { buffer in
+                // according to docs only EINVAL and ENOMEM are possible here.
+                _ = _dupenv_s(buffer, &count, environmentVariable)
+            }
+            ptr?.deallocate()
+            let envVarString: String? = if count > 0, let ptr {
+                String(cString: ptr)
+            } else {
+                nil
+            }
```
Lukasa: Can we factor this out? It makes the code much harder to follow by being one line long.

Sources/NIOCore/IO.swift
```diff
@@ -125,8 +125,18 @@ public struct IOError: Swift.Error {
...
 private func reasonForError(errnoCode: CInt, reason: String) -> String {
-    if let errorDescC = strerror(errnoCode) {
-        return "\(reason): \(String(cString: errorDescC)) (errno: \(errnoCode))"
+    #if os(Windows)
+    let errorDesc = withUnsafeTemporaryAllocation(of: CChar.self, capacity: 256) { ptr -> String? in
+        if strerror_s(ptr.baseAddress, ptr.count, errnoCode) == 0 {
+            return String(cString: UnsafePointer(ptr.baseAddress!))
+        }
+        return nil
+    }
```
Lukasa: Same note here, can we factor this out?

Sources/NIOCore/SystemCallHelpers.swift
```diff
@@ -72,11 +72,23 @@ internal func isUnacceptableErrno(_ code: Int32) -> Bool {
...
+        return
+    }
+
+    #if os(Windows)
+    let errorDesc = withUnsafeTemporaryAllocation(of: CChar.self, capacity: 256) { ptr -> String? in
+        if strerror_s(ptr.baseAddress, ptr.count, err) == 0 {
+            return String(cString: UnsafePointer(ptr.baseAddress!))
+        }
+        return nil
+    }
```
Lukasa: Another reason to factor the above out is that it'll deduplicate here.

Sources/NIOCore/Utilities.swift
```diff
@@ -269,3 +270,30 @@ extension System {
...
+func windows_strerror(_ errnoCode: CInt) -> String? {
+    withUnsafeTemporaryAllocation(of: CChar.self, capacity: 256) { ptr in
+        if strerror_s(ptr.baseAddress, ptr.count, errnoCode) == 0 {
+            return String(cString: UnsafePointer(ptr.baseAddress!))
+        }
+        return nil
+    }
+}
+
+func windows_getenv(_ env: String) -> String? {
```
Lukasa: These are Swift functions, we can use a Swift naming convention for them.

Sources/NIOCore/Utilities.swift
```diff
@@ -269,3 +270,30 @@ extension System {
...
+    var count = 0
+    var ptr: UnsafeMutablePointer<CChar>? = nil
+    withUnsafeMutablePointer(to: &ptr) { buffer in
+        // according to docs only EINVAL and ENOMEM are possible here.
+        _ = _dupenv_s(buffer, &count, env)
+    }
+    defer { ptr?.deallocate() }
```
Lukasa: We can't call `deallocate` here: we _must_ call `free`.

Sources/NIOCore/Utilities.swift
```diff
@@ -269,3 +270,30 @@ extension System {
...
+    return if count > 0, let ptr {
+        String(cString: ptr)
```
Lukasa: The documentation does not promise that this is a null-terminated string, and even if it did, we were given the count. It's more efficient to use that count.

Sources/NIOCore/Utilities.swift
fabianfett: You mean with a `Windows` namespace around them? Happy to do that!

Lukasa: Yeah, and no `snake_case`.

Sources/NIOCore/Utilities.swift
fabianfett: Is this because this is a c pointer and `deallocate` only works for pointers that were acquired in Swift land?

Lukasa: No, it's because you are never allowed to mix allocators. It is possible that Swift's `allocate`/`deallocate` functions use an entirely separate heap from the `malloc`/`free` set. Similarly, C++'s `new` and `delete` may use a different, third allocator.

In this case, the documentation says you must call `free`, so that's the function you must call.

fabianfett: Okay, so c, swift and c++ may use different allocators. Thanks for clarifying!

Sources/NIOCore/GlobalSingletons.swift
```diff
@@ -184,7 +184,11 @@ extension NIOSingletons {
         case 0:  // == 0
             // Not set by user, not yet finalised, let's try to get it from the env var and fall back to
             // `System.coreCount`.
+            #if os(Windows)
+            let envVarString = windows_getenv(environmentVariable)
```
Lukasa: Does this compile? Where has this function come from?

